PR #24, opened 2021-10-19
Remove duplicate yaml files

hanlint:


Averylamp:
Did we want to get rid of the symlinks?  I think its nice to have a symlink to the underlying yaml in the relevant part of the codebase for reference

hanlint:
@Averylamp -- I dont think we are removing any symbolic links in this PR?

I believe the symbolic links were removed by @ravi-mosaicml because they were breaking for the `pip` package, somehow leading to infinite recursions (e.g. pointing to itself). If there's a fix for that, agreed that we should have symbolic files to `/composer/models/resnet50/resnet50.yaml` as that's where users expect them.

PR #48, opened 2021-10-28
Configuration of DDP syncing strategies to resolve `find_unused_parameters` woes

### LINKED ISSUE #9: More efficient microbatch DDP sync behavior when `find_unused_parameters` is True
Ordinarily, when training with gradient accumulation, we only need to do a DDP sync on the final microbatch, because synced gradients aren't needed until the optimizer runs at the end of the batch. However, the `find_unused_parameters` flag indicates that some algorithms (such as stochastic depth) may cause not all gradients to be generated. Critically, the set of unused parameters may vary between microbatches. Syncing on only the last microbatch may cause some parameters used in earlier microbatches but unused in the final microbatch to not be properly synced - resulting in severe quality degradations.

Our current solution to this issue is just to sync all microbatches when the `find_unused_paramaters` flag is set, but this incurs a throughput penalty of about 5%, depending on gradient accumulation setting. We would like to investigate whether it is possible to sync all parameters used in any microbatch, to avoid this throughput penalty.

jbloxham:
Partially resolves #9 , but only in extreme cases.

We've been benchmarking stochastic depth, an algorithm with plays with `find_unused_parameters`, with some extremely high values for `grad_accum`. This stretches thin our current `find_unused_parameters` logic, which causes DDP syncs to happen after every microbatch. This PR introduces a new DDP sync strategy, `FORCED_SYNC`, which causes us to do a manual sync of all gradients at the end of a batch, regardless of which parameters are allegedly being used. This causes a dramatic increase in throughput for high `grad_accum` cases, but can actually decrease throughput in some other cases. Because of this, I've made it configurable.

Experiments:
`single_auto_sync` represents the previous, pre-fix behavior. `multi_auto_sync` is the original behavior introduced to fix `find_unused_parameters`. `manual_sync` (now renamed to `forced_sync`) is the new behavior intended as an alternative to `multi_auto_sync`.

RN50:

ResNet compellingly demonstrates the need for alternatives to `single_auto_sync`, which fails miserably when trained with stochastic depth. However, `manual_sync` is unnecessary here since the `grad_accum` is only 2. `multi_auto_sync` is around 4% slower than `single_auto_sync`, and `manual_sync` is around 5% slower. Accuracy is comparable between `multi_auto_sync` and `manual_sync`

Screen Shot 2021-10-28 at 09 43 18
Screen Shot 2021-10-28 at 09 43 25

GPT-2:

For these experiments, GPT-2 was trained with a `grad_accum` of 64, though not to completion. Accuracy is again comparable between `multi_auto_sync` and `manual_sync`, though `single_auto_sync` is curiously only slightly behind. Here, `manual_sync` is less than a percent slower than `single_auto_sync`, but `multi_auto_sync` is nearly _70%_ slower, owing to the very high `grad_accum`. Note that the dips in throughput are caused by validation happening every 250 steps, in combination with throughput measuring a 100-step window.

Screen Shot 2021-10-28 at 09 49 59
Screen Shot 2021-10-28 at 09 50 13


Landanjs:
LGTM but I'm sure there are people who know better than me. Thanks for the implementation Jamie! This is a big improvement for language models.

Landanjs (approved):

Landanjs:
Could someone from eng (@ajaysaini725, @abhi-mosaic, @hanlint) check this PR as well? I am blocked from running stochastic depth experiments on NLP without this fix.

abhi-mosaic:
This makes sense to me, but for the config I think we should be using `manual_sync` as the default when `find_unused_parameters` is turned on, right?

composer/trainer/ddp.py
```diff
@@ -93,6 +126,10 @@ def __init__(self,
         self.fork_rank_0 = fork_rank_0
         self.processes: List[subprocess.Popen[str]] = []
         self.find_unused_parameters = find_unused_parameters
+        if ddp_sync_strategy is None:
+            self.ddp_sync_strategy = DDPSyncStrategy.SINGLE_AUTO_SYNC if not find_unused_parameters else DDPSyncStrategy.MULTI_AUTO_SYNC
```
abhi-mosaic, line 130: Here, shouldn't we be using `FORCED_SYNC` when find_unused_parameters is enabled

jbloxham: Yeah I think I agree that makes more sense now as well.

PR #50, opened 2021-10-28
Added repeat to the torch profiler

ravi-mosaicml:
Added the repeat argument to the hparams for the torch profiler callback

This makes it possible to profile only N times per epoch. Profiling slows down the training process significantly!


Averylamp (approved):
lgtm

PR #52, opened 2021-10-29
Cleanup dataloader threads to allow termination

### LINKED ISSUE #51: Training run processes do not stop at the end of training
## Environment

mosaicml/research:latest docker container on 3080s.

## To reproduce

Command: 
`python examples/run_mosaic_trainer.py -f composer/yamls/models/resnet50.yaml --loggers wandb --loggers.wandb.entity mosaic-ml --loggers.wandb.project landan-random --callbacks speed_monitor lr_monitor --callbacks.speed_monitor.window_size 100`

I believe Cory saw hanging at the end of the CIFAR-10 benchmark as well, so that may be sufficient to reproduce the bug.

## Expected behavior

All (sub)processes to be killed at the end of training.

## Additional context

Training runs hang at the end of training. This means the processes will continue to run although training is complete.

jbloxham:
Resolves #51 .

Seems that the problem was that the threads initialized by the dataloaders were preventing processes from exiting. I'm a little confused why this only became a problem after launch, but I've been able to confirm that this is definitely the issue.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -642,6 +642,10 @@ class Trainer:
 
         self.engine.run_event(Event.TRAINING_END)
 
+        # Deleting the dataloaders ensures their worker threads get cleaned up.
+        del state.train_dataloader
+        del state.eval_dataloader
+
     def _train_batch(self, microbatches: Sequence[Batch], ddp_sync: bool = True):
         """Run training on a full batch of data.
 


jbloxham:
In retrospect this probably isn't the right fix - one can imagine that someone would still want access to the dataloaders after the `fit` function has returned. The real issue is why the dataloader workers aren't exiting after the dataloader goes out of scope.

PR #53, opened 2021-10-29
Add factorization

dblalock:
This is a port of factorization from the old repo. Components: 

- [X] Core factorization logic for linear case
- [X] Core factorization logic for Conv2d case
- [x] Factorized Conv2d layers
- [x] Factorized Linear layers
- [X] Algorithm + standalone functions to surgery vanilla modules->factorized modules
- [X] Tests for all of the above
- [x] Docs for public API
- [x] Algo card

Not in scope:

- Anything involving intelligently allocating or scheduling the factorization. This is because [DDP](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html) basically just won't let us mess with the parameters after INIT at all, which factorization needs to do. We could get around this by iteratively recreating the DDP wrapper, but that's definitely out of scope for this PR
- Refactoring tests and other somewhat shared code (many algorithms have somewhat copy-pasted implementations and tests right now, including this one).

[Wandb ResNet-50 factorization + baseline runs](https://wandb.ai/mosaic-ml/rn50?workspace=user-dblalock). tl;dr is that, at these layer widths, factorization is slightly slower and way less accurate than the baseline.
[Method card / docs](https://mosaicml-composer--53.com.readthedocs.build/en/53/method_cards/factorize.html)

Changes to existing code:

- Added a few modules to the `SimpleConvModel` fixture so that factorizing it would be meaningful
- Bugfix in the layer freezing tests revealed by the altered `SimpleConvModel`
- `algorithm_registry`, `test_algorithm_registry`, and algorithms `__init__.py` have new entries, as expected for any new algorithm


dblalock:
I still need to finish off the docs + fix some pyright errors, but since next release deadline is closing in, un-draft-ing this to give you guys enough time to look at it (it's pretty big).

A couple parts I'm unsure about:

- The naming conventions for the pairs factorized sub-operations + their variables. I won't tell you what specifically I think is weird so we get independent estimates of whether it stands out, but basically I'm not certain about navigating what looks like clean mathematical variable names vs normal torch/python variable names.
- Using `_channels` for the factorized convolution sizes and `_features` for the factorized linear sizes. This matches the field names for the associated torch classes, but I'm not quite satisfied with them.
- What dumb things I subtly did, since there's always something

hanlint (commented):
Nice work! Did a quick pass on the more visible bits of code

composer/algorithms/factorize/factorize.py
```diff
@@ -0,0 +1,160 @@
...
+        default=_DEFAULT_LATENT_FEATURES,
+    )
+
+    def initialize_object(self) -> 'Factorize':
+        return Factorize(**asdict(self))
+
+
+class Factorize(Algorithm):
+
+    def __init__(self,
```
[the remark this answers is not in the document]
dblalock: Done. Also, in case it's relevant in the future, I checked and sphinx is smart here and puts the values of the constants in the docs instead of their names

ravi-mosaicml:
Can you rebase against the latest version of `dev`? I can then give a review. Thanks!

dblalock:
@ravi-mosaicml: Merged latest dev. Didn't actually rebase because 1) it needed a surprising number of manual steps and I don't want to break stuff, and 2) almost all the changes are self-contained in `composer/algorithms/factorize` + `docs`, and the diffs seem pretty clean. I can make the rebase happen if your life is harder without it though. 

Also, I think @hanlint mentioned being hesitant to merge this (or put it in a release?) without better results. I thought about it more and I think I can hack a way around the DDP limitation that's forcing us to factorize naively (and get bad results) right now, but I'm hesitant to add another big feature to an already-heavy PR. Also, for the models + hardware we're considering, even this isn't going to make factorization an especially good idea.

hanlint:
OK great -- @dblalock defer to @jfrankle on whether to merge this!

composer/algorithms/factorize/factorize.py
```diff
@@ -0,0 +1,192 @@
...
+    have lower thresholds.
+
+    See :func:`~composer.algorithms.factorize.factorize_matrix` and
+    :func:`~composer.algorithms.factorize.factorize_conv2d` for more
+    information about the factorization process. See :class:`~composer.algorithms.factorize.FactorizedConv2d` and :class:`~composer.algorithms.factorize.FactorizedLinear`
+    for more information about the factorized modules used to replace the
+    original modules.
+
+    Args:
+        factorize_convs: whether to try factorizing :class:`torch.nn.Conv2d` modules.
```
ravi-mosaicml, line 118: Can you include the default values in the docstrings?

composer/algorithms/factorize/factorize_core.py
```diff
@@ -0,0 +1,353 @@
...
+    Wa: Optional[torch.Tensor] = None
+    Wb: Optional[torch.Tensor] = None
+    bias: Optional[torch.Tensor] = None
+    rank: int = -1
+    nmse: float = 0
+
+
+def _lstsq(A: torch.Tensor, B: torch.Tensor) -> torch.Tensor:
+    if A.shape[0] != B.shape[0]:
+        raise IndexError("A has different number of rows than B! " f"A.shape = {A.shape}, B.shape = {B.shape}")
```
ravi-mosaicml, line 48:
```suggestion
        raise ValueError("A has different number of rows than B! " f"A.shape = {A.shape}, B.shape = {B.shape}")
```
I think `IndexError` should only be raised when slicing. See https://docs.python.org/3/library/exceptions.html#IndexError

composer/algorithms/factorize/factorize_modules.py
```diff
@@ -0,0 +1,441 @@
...
+from typing import Optional, Tuple, Union, cast
+
+import numpy as np
+import torch
+from torch import nn
+from torch.nn.common_types import _size_2_t
+
+from composer.algorithms.factorize.factorize_core import LowRankSolution, factorize_conv2d, factorize_matrix
+
+FractionOrInt = Union[int, float]
```
ravi-mosaicml, line 14: Would it be possible to hardcode `Union[int, float]` where it is used? I get this simplifies things code-wise, but makes it slightly less inteutivie to follow it, as I had to look up in another file what the allowed types actually are (by `Fraction`, I was originally thinking a https://docs.python.org/3/library/fractions.html)

composer/algorithms/factorize/factorize_modules.py
```diff
@@ -0,0 +1,441 @@
...
+    error_msg = "Can't apply unititalized solution!"
+    assert solution.bias is not None, error_msg
+    assert solution.Wa is not None, error_msg
+    assert solution.Wb is not None, error_msg
+
+    with torch.no_grad():
+        # first op always has no bias since adds no expressivity
+        if module0.bias is not None:
+            assert isinstance(module0.bias, torch.Tensor)
+            module0.bias = torch.nn.parameter.Parameter(torch.zeros(solution.rank, dtype=module0.bias.dtype))
```
ravi-mosaicml, line 63: does this need a device argument?

composer/algorithms/factorize/factorize_modules.py
```diff
@@ -0,0 +1,441 @@
...
+            out_channels: number of channels produced by the convolution
+            kernel_size: size of the convolving kernel
+
+        Returns:
+            latent_channels: the largest allowable number of latent channels
+        """
+        return max_rank_with_possible_speedup(in_features, out_features, kernel_size=kernel_size)
+
+    @staticmethod
+    def from_conv2d(module: torch.nn.Conv2d, module_ix: int = -1, **kwargs) -> 'FactorizedConv2d':
```
ravi-mosaicml, line 317: At the top of the file, you can add a `from __future__ import annotations`. Then, you don't need to use quotes for typing annotations.
```suggestion
    def from_conv2d(module: torch.nn.Conv2d, module_ix: int = -1, **kwargs) -> FactorizedConv2d:
```

ravi-mosaicml (commented):
Overall LGTM!!! Left some minor comments throughout.

I love the class docstrings -- they are very detailed. I would also ask to please add docstrings to all public functions and public class member functions as well (unless if the superclass has applicable docstrings) or to make the functions private. I didn't comment them all but I think that would be helpful.

jfrankle:
I think we should merge this. It'll form the foundation for future factorization experiments, and we should get it in before it gets too far out of date. Given that it includes a method card that acknowledges the strengths and weaknesses of the current method, I think we're in good shape. This is something people frequently ask about when I mention what we're up to, and it's good that we have an answer to the question, even if it isn't an emphatic "you should definitely use this."

composer/algorithms/factorize/factorize_core.py
```diff
@@ -0,0 +1,353 @@
...
+
+import dataclasses
+from typing import Optional, Tuple, Union
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
dblalock: I need it mutable with certain default values. So technically not necessary, but less code than namedtuple. Leaving for now but happy to change it if you want.

dblalock:
@ravi-mosaicml, I think this is good to go.

The one question mark I have is that I wasn't sure which methods / functions you were referring to as far as missing docs; I think the factorize algorithm + modules have docs for everything ([built docs](https://mosaicml-composer--53.com.readthedocs.build/en/53/algorithms.html#id25)). There are some non-leading-underscore functions in the modules though--my current understanding is that we're not considering stuff "public" unless it's imported in the directory's `__init__.py`, although I could totally be wrong about that.

ravi-mosaicml:
Discussed offline, but I think we're going with the convention that a "public" function is one that does not begin with an underscore -- even if it doesn't appear in `__init__.py`. (An exception to this rule is if the file or folder begins with an underscore -- then it is assumed that everything in there is private, regardless if individual functions have an underscore). So, it would be great if these functions could either be renamed to start with an underscore (in which case, no docs required!) or have a proper docstring.

dblalock:
Yeah, my above commit uses this understanding. Everything should be good now.

hanlint (approved):
This LGTM -- please cleanup any linting/jenkins testing errors and we can merge.

PR #57, opened 2021-11-02
Additional`SyntheticDataset` Customization

ajaysaini725:
Right now, `SyntheticDataset` can only represent an epoch that consists of a single batch. This change allows `SyntheticDataset` to emulate a dataset that consists of multiple batches. This functionality is useful for benchmarking workloads that use a `SyntheticDataset`.

Additionally, `SyntheticDataset` only supports classification labels. This PR adds functionality for custom label shapes so that `SyntheticDataset` can be used for tasks such as image segmentation and NLP.

### DIFF
--- a/composer/datasets/synthetic.py
+++ b/composer/datasets/synthetic.py
@@ -126,7 +126,9 @@ class SyntheticDatasetHparams(DatasetHparams):
     memory_format: MemoryFormat = hp.optional("Memory format for the sample pool",
                                               default=MemoryFormat.CONTIGUOUS_FORMAT)
     sample_pool_size: int = hp.optional("Number of samples", default=100)
-    total_dataset_size: int = hp.optional("Total size of the dataset to emulate", default=100)
+    total_dataset_size: int = hp.optional(
+        "Total size of the dataset to emulate. "
+        "If None defaults to sample_pool_size.", default=None)
     drop_last: bool = hp.optional("Whether to drop the last samples for the last batch", default=True)
     shuffle: bool = hp.optional("Whether to shuffle the dataset for each epoch", default=True)
     data_type: SyntheticDataType = hp.optional("Type of synthetic data to create.", default=SyntheticDataType.GAUSSIAN)
--- a/composer/datasets/synthetic.py
+++ b/composer/datasets/synthetic.py
@@ -141,6 +141,7 @@ class SyntheticDatasetHparams(DatasetHparams):
                 memory_format=self.memory_format,
                 sample_pool_size=self.sample_pool_size,
                 data_type=self.data_type,
+                total_dataset_size=self.total_dataset_size,
             ),
             drop_last=self.drop_last,
             shuffle=False,
--- a/composer/datasets/synthetic.py
+++ b/composer/datasets/synthetic.py
@@ -72,9 +72,7 @@ class SyntheticDataset(torch.utils.data.Dataset):
         self.memory_format = getattr(torch, MemoryFormat(memory_format).value)
         self.transform = transform
 
-        _validate_label_inputs(label_type=self.label_type,
-                               num_classes=self.num_classes,
-                               label_shape=self.label_shape)
+        _validate_label_inputs(label_type=self.label_type, num_classes=self.num_classes, label_shape=self.label_shape)
 
         # The synthetic data
         self.input_data = None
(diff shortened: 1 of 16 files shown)


composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
     def __len__(self) -> int:
-        return self.size
+        return self.batch_size * self.batches_in_dataset
 
     def __getitem__(self, idx: int):
+        idx = idx % self.batch_size
+        if idx == 0 and self.input_data is not None and self.input_target is not None:
+            # allocate actual memory on each new batch for more realistic
+            # throughput
+            self.input_data = torch.clone(self.input_data)
```
ravi-mosaicml, line 89: clone the sample, not the entire dataset

composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
-                 data_type: SyntheticDataType = SyntheticDataType.GAUSSIAN):
-        self.size = sample_pool_size
-        self.shape = shape
-        self.input_data = None
-        self.input_target = None
-        self.memory_format = getattr(torch, MemoryFormat(memory_format).value)
-        self.device = device
+                 batch_size: int,
+                 data_shape: Sequence[int],
+                 batches_in_dataset: int = 1,
```
ravi-mosaicml, line 51:
```suggestion
                 dataset_size: int = ...,
```

composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
-                 num_classes: int,
-                 transform: Optional[Callable] = None,
-                 data_type: SyntheticDataType = SyntheticDataType.GAUSSIAN):
-        self.size = sample_pool_size
-        self.shape = shape
-        self.input_data = None
-        self.input_target = None
-        self.memory_format = getattr(torch, MemoryFormat(memory_format).value)
-        self.device = device
+                 batch_size: int,
```
ravi-mosaicml, line 49: The dataloader is expecting the dataset to yield samples, not batches, unless if we're using a custom collate_fn. So I don't think this argument should be needed?

ravi-mosaicml, line 49: The dataloader is expecting the dataset to yield samples, not batches, unless if we're using a custom collate_fn. So I don't think this argument should be needed?

composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
+                 data_shape: Sequence[int],
+                 batches_in_dataset: int = 1,
+                 data_type: SyntheticDataType = SyntheticDataType.GAUSSIAN,
```
ravi-mosaicml, line 52:
```suggestion
                 synthetic_data_distribution: SyntheticDataDistribution = SyntheticDataDistribution.GAUSSIAN,
```

composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
+                 label_type: SyntheticDataLabelType = SyntheticDataLabelType.CLASSIFICATION,
+                 one_hot: Optional[bool] = None,
```
ravi-mosaicml, line 54: Should `one_hot` be implied via the `label_type`? E.g. have `CLASSIFICATION_ONE_HOT`, CLASSIFICATION_INT`

composer/datasets/synthetic.py
```diff
@@ -17,40 +17,78 @@ class SyntheticDataType(StringEnum):
...
+                 num_classes: Optional[int] = None,
+                 label_shape: Optional[Sequence[int]] = None,
+                 device: str = "cpu",
+                 memory_format: Union[str, MemoryFormat] = MemoryFormat.CONTIGUOUS_FORMAT,
+                 transform: Optional[Callable] = None):
+        self.batch_size = batch_size
```
ravi-mosaicml, line 60: To go with the (kinda new convention), can you store all of these variables in `self.hparams = SyntheticDatasetHparams(data_shape=data_shape, ...)`

ravi-mosaicml (changes requested):
:+1: for the new options, though a bit confused on the batch size, since that is usually a dataloader (not dataset) property.

ravi-mosaicml (changes requested):
:+1: for the new options, though a bit confused on the batch size, since that is usually a dataloader (not dataset) property.

ravi-mosaicml (approved):
LGTM

PR #62, opened 2021-11-03
Lazy loading of non-core dependencies

ravi-mosaicml:
Resolved lazy-loading of non-core dependencies by using guarded imports.

Non-core dependencies should be imported either a) in a function, and/or b) if guarded with `TYPE_CHECKING` (we run pyright when installing everything).

Updated the tests to selectively ignore some `ModuleNotFound` errors and instead trigger a pytest.skip.

Tested manually by running tests on an installation of `pip install composer[dev]`. In the future, Jenkins will run this test.


tests/test_load.py
```diff
@@ -22,6 +23,14 @@
 model_names = [name for name in model_names if name not in EXCLUDE_MODELS]
 
 
+def get_model_algs(model_name: str) -> List[str]:
```
hanlint, line 26: Could you say more here about why this is needed?

hanlint (changes requested):

tests/test_load.py
ravi-mosaicml: Alibi requires non-core dependencies, so I did not want to load that when testing algs on image models.

ravi-mosaicml: Also see #76

hanlint: Ah I see. Perhaps we can do a try catch here, similar to your handling of unet? Otherwise users that implement a new image model would need to update this list here.

hanlint (approved):
LGTM

ajaysaini725 (approved):
LGTM 👍

PR #63, opened 2021-11-04
DDP Enhancements

ravi-mosaicml:
1. Added timeout to the hparams for initialize_process_group. The default of 30 minutes was too long for failing tests, which prevents one from getting any meaningful log.
2. In cleanup(), using `os.killpg` to terminate DDP subprocesses instead of just subprocess.kill(). It appears that sometimes zombie processes would be still there (e.g. from ddp / dataloader workers)
3. In cleanup(), attempting a `sigterm` before resorting to a `sigkill()` 5 seconds later. Graceful cleanup is preferred!
4. Directing output from `stdout` and `stderr` to tempfiles instead of `subprocess.PIPE`, which can hang if a subprocess generates significant output


hanlint:
is this going to conflict with #65 ? cc: @jbloxham

jbloxham:
> is this going to conflict with #65 ? cc: @jbloxham

Hmm, I definitely need to incorporate these changes into that PR, but I don't see any reason that shouldn't work. :)

composer/trainer/ddp.py
```diff
@@ -82,16 +85,19 @@ def __init__(self,
                  num_nodes: int,
                  backend: str,
                  fork_rank_0: bool,
+                 timeout: float,
                  find_unused_parameters: bool = False):
+        self.hparams = DDPHparams(
```
ajaysaini725, line 90: Are we for sure using this pattern everywhere? There are some places (ex. `trainer.py`) where we do the opposite (call init from `create_from_hparams`).

composer/trainer/ddp.py
```diff
@@ -287,24 +301,50 @@ def monitor(self) -> None:
...
-                pass
+            if process.returncode is None:
+                logger.info("Killing subprocess %s with SIGTERM", process.pid)
+                self.killed_pids.add(process.pid)
+                try:
+                    os.killpg(process.pid, signal.SIGTERM)
+                except ProcessLookupError:
+                    pass
+        current_time = datetime.datetime.now()
+        while datetime.datetime.now() - current_time < datetime.timedelta(seconds=5):
```
ajaysaini725, line 329: should use a `timeout` variable instead of hardcoding 5 seconds

composer/trainer/ddp.py
```diff
@@ -287,24 +301,50 @@ def monitor(self) -> None:
...
+            all_finished = True
```
ajaysaini725, line 330: `all_finished = all([p.returncode is None for p in self.processes])`

composer/trainer/ddp.py
```diff
@@ -287,24 +301,50 @@ def monitor(self) -> None:
...
+                if process.returncode is None:
+                    all_finished = False
+                    break
+            if all_finished:
+                break
+            time.sleep(0.1)
+
+        for process in self.processes:
+            if process.returncode is None:
+                logger.error("Killing subprocess %s with SIGKILL", process.pid)
```
ajaysaini725, line 341: Change to `Failed to kill subprocess %s with SIGTERM, using SIGKILL instead`?

composer/trainer/ddp.py
ravi-mosaicml: trainer.py is the exception to this rule, just because there is so much logic in there that really belongs in the trainer.py and not the hparams. But otherwise, `create_from_hparams` should call init, which then recreates the hparams.

composer/trainer/ddp.py
ravi-mosaicml: Good point, I'll move it to a constant at the top. CC @jbloxham if we want to make this into a cli arg in #65 (not going to do so here, since #65 will replace this).

ajaysaini725 (approved):
LGTM 👍

PR #64, opened 2021-11-04
Fix Tests and setup.py

ravi-mosaicml:
1. Fixed settings for coverage reports so source code can be included
2. Fixed package_data so yaml files are properly included. 
3. Added `py.typed` to indicate that the repository has typing annotations
4. Printing the `pillow-simd` warning in setup.py to stderr so it does not interfere with `setup.py --version` and other CLI args
5. Logging at the debug level when running tests
6. Increasing timeout for `test_load`, since that sometimes takes longer than 2 seconds.
7. Only bundling the `composer` package in setup.py so the `tests/` folder isn't included as well


ajaysaini725 (approved):
LGTM!

PR #65, opened 2021-11-04
Executable for distributed launching

jbloxham:
Refactors DDP initialization to make use of a dedicated executable for launching multiprocessing, rather than relaunching in the middle of the trainer initialization itself. For now, this executable takes only a single argument, `--nproc`, but this can easily be extended to handle multinode setups, and could also invoke the `deepspeed` exectuable. Most of the previous DDP hparams, like DDP store and `fork_rank_0`, are no more.

I experimented a bit with having this executable wrap `torch.distributed.run`, but I found that script to be buggy, verbose, and lacking good config options (for example, the script doesn't support `nproc=1` at all, and some of the config options get confused if you try passing a 2-digit number for node rank). Instead, I repurposed a lot of the existing DDP code.

Here's a pair of RN50 runs on 4x3080s to demonstrate mathematical equivalence: https://wandb.ai/mosaic-ml/jamie-ddp-launch-regression?workspace=user-jbloxham. I accidentally only ran to 5 epochs, but I think this is enough to be compelling.

Most of the complexity of this PR comes from refactoring the tests so that they are no longer dependent on the previous behavior. The solution I landed on was to introduce a helper function `with_distributed` that runs a function with multiprocessing. This turned out to be fairly versatile - it was convenient for checkpointing tests to be able to invoke this function twice within a single test, for instance, which wasn't possible before.

Admittedly, I still need to add docs, and I swear by my honor that I will not merge this PR before then. The code is final though, and ready for review.


tests/fixtures/ddp_fixtures.py
```diff
@@ -1,70 +1,55 @@
...
+    os.environ['LOCAL_RANK'] = str(rank)
+    os.environ['LOCAL_WORLD_SIZE'] = str(num_procs)
+
+    os.environ['MASTER_ADDR'] = '127.0.0.1'
+    os.environ['MASTER_PORT'] = str(29400)
+
+    target(*args, **kwargs)
+
+
+def with_distributed(num_procs: int,
```
ravi-mosaicml, line 30: I love the simplicity of this....but also wonder what are the potential implication of using `mp.spawn` for tests while we run jobs with the launch script. Here are some questions that come to mind:

* Is this slow? (Especially after running many tests, and then we spawn something?)
* I vaguely remember that some things did not work with mp.spawn (maybe that was never an issue) but if so, curious how we fixed it?
* Do we have any testing for the launch script?
* Are there any issues with using a TCPstore (i.e. port already in use?)
* Does this work with wandb? Mainly thinking if we end up doing long-running regressions with pytest, then we would be using spawn.
* Curious about the function signature -- could it take the targets's `*args` and `**kwargs`, invoke `run_target`,  return None? I think that is more standard than returning a callable.

As an alternative, thoughts about invoking pytest using the launch script? (E.g. `composer -m pytest ....`. Also see the comment in the launch script for the `-m` option). All the CPU tests would execute once per rank (not a big deal). For the 1 GPU tests, we would need to make the 2nd process idle for the duration of the test. It would also be good to have some sort of coordination to make sure that tests start at the same time across all ranks (to ensure we don't hit a timeout stemming from a slow cleanup of a previous test).

tests/fixtures/ddp_fixtures.py
```diff
@@ -1,70 +1,55 @@
...
+                     target: Callable,
+                     initial_timeout: int = 30,
```
ravi-mosaicml, line 32: I think it would be fine to have one timeout parameter and then have `subsequent_timeout_per_device` always be 0....just would need to make sure the initial timeout is sufficiently high.

tests/fixtures/ddp_fixtures.py
```diff
@@ -1,70 +1,55 @@
...
+            processes += [p]
+
+        processes[0].join(initial_timeout)
+        for p in processes[1:]:
+            p.join(subsequent_timeout_per_device)
+
+        for rank, p in enumerate(processes):
+            if p.exitcode is None:
+                p.kill()
+                pytest.fail(f'Process {rank} did not exit', pytrace=False)
```
ravi-mosaicml, line 51: Will we get stdout / stderr output from the sub processes?

tests/trainer/test_trainer.py
```diff
@@ -77,38 +77,40 @@ def test_trainer_validation(mosaic_trainer_hparams: TrainerHparams, invalid_hpar
...
     first_trainer = Trainer.create_from_hparams(mosaic_trainer_hparams)
     first_trainer.fit()
     first_model = first_trainer.state.model.module
     assert isinstance(first_model, BaseMosaicModel)
     assert first_trainer.state.train_dataloader is not None
-    first_loss = get_total_loss(first_model, first_trainer.state.train_dataloader)
+    first_loss = get_total_loss(first_model, first_trainer.state.train_dataloader, first_trainer.ddp)
+
+    # Need to reinitialize some distributed settings in order to train twice in the same process
+    torch.distributed.destroy_process_group()
```
ravi-mosaicml, line 94: Curious why wasn't the process group destroyed properly the first time?

tests/utils/trainer_fit.py
```diff
@@ -9,21 +9,25 @@
...
     with torch.no_grad():
         total_loss = 0
         for batch in dataloader:
             outputs = model(batch)
             loss = model.loss(outputs, batch=batch)
             for l in ensure_tuple(loss):
                 total_loss += l.item()
-        return total_loss
+
+        total_loss_tensor = torch.Tensor([total_loss])
```
ravi-mosaicml, line 28: Nice good catch

ravi-mosaicml (changes requested):
Great work. Love the simplicity of the launch script. Main comment relates to testing and whether it would be possible to use the launch script to run pytest, instead of using mp.spawn. Other than that, most other comments are minor.

composer/cli/launcher.py
```diff
@@ -0,0 +1,141 @@
...
+
+def main():
+    args = parse_args()
+
+    processes = launch_processes(args.nproc, args.training_script, args.training_script_args)
+
+    try:
+        monitor_processes(processes)
+    finally:
+        cleanup_processes(processes)
```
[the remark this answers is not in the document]
jbloxham: I suppose so - it's not strictly necessary because the entry point defined in `setup.py` knows about the `main` function, but I guess a user could try to do something like `python composer/cli/launcher.py ...`?

composer/core/state.py
```diff
@@ -144,17 +142,21 @@ class State(Serializable):
...
+    def world_size(self) -> int:
+        return get_world_size()
 
     @property
     def global_rank(self) -> int:
         return get_global_rank()
 
+    @property
+    def local_world_size(self) -> int:
+        return get_local_world_size()
```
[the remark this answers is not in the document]
jbloxham: Yeah... I've seen a bit of variation wrt which one is used. I marginally prefer `local_world_size` because there's a slim chance we could end up with nodes with different sizes.

tests/callbacks/test_grad_monitor.py
```diff
@@ -39,7 +40,8 @@ def test_grad_monitor_no_layers(mosaic_trainer_hparams: TrainerHparams):
     assert grad_norm_calls == num_train_steps
 
 
-def test_grad_moniter_per_layer(mosaic_trainer_hparams: TrainerHparams):
+@pytest.mark.timeout(60)
```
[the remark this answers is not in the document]
jbloxham: TBH, I'm not really sure where that came from. I've gotten rid of it now.

Averylamp:
This reminds me a lot of Nvidias launching script, which they seem to simplify a little bit
https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/Classification/ConvNets/multiproc.py
A couple differences with their script that may be worth stealing:
- Saving non rank-0 stdout to named `GPU_#.log` files
- Seems like they don't differentiate between sigkill and other errors
- They do run python unbuffered
- For killing they just do `p.terminate()` and don't do the sigkill style.  Not really sure what works better here/if there was a specific reason to do sigkill and other things there?

composer/cli/launcher.py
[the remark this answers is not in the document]
Averylamp: I think because Nvidia examples allows you to execute it as a script it'd probably be good to allow that as well.  I'd also want for it to be a 
```
if __name__ == "__main__":
  exit(main())
```
where the exit code gets passed through the script if possible (sometimes its nice for other automations/tests)

composer/trainer/ddp.py
```diff
@@ -414,60 +245,10 @@ def ddp_sync_context(self, state: State, is_final_microbatch: bool):
...
-    }
-
-    store: StoreHparams = hp.optional(doc="Store", default_factory=TCPStoreHparams)
-    node_rank: int = hp.optional(doc="Node ID for multi-node training", default=0)
-    num_nodes: int = hp.optional(doc="Number of nodes used for training", default=1)
-    fork_rank_0: bool = hp.optional(
-        doc="Whether to fork the local rank 0 process, or use the existing process for rank 0 training.",
-        default=False,
-    )
+    sync_strategy: Optional[str] = hp.optional(doc="Strategy for DDP syncing", default=None)
```
Averylamp, line 253: Would love to have the default be something reasonable or explicit somewhere

tests/fixtures/ddp_fixtures.py
jbloxham: Great questions - here are my thoughts:

- I seem to have broken the tests after rebasing on latest `dev`, but I'll report benchmarks for `pytest` once that's fixed.
- `spawn` does admittedly have a notable issue - it depends on being able to pickle the local state. This wasn't a serious blocker for any of the tests I refactored, but I do see how it could become an issue.
- There's no testing for the launch script at the moment - I'll add a test :)
- Honestly I expected to run into issues with TCPstore reuse, but for some reason that never happened. If that does become an issue down the road somehow, this function can handle some of the logic of searching for an available port.
- I don't see any reason this wouldn't work with WandB. The means of multiprocessing isn't too relevant to logging.
- I think I stand by the function signature. The main advantage it has is it makes it easy to differentiate between arguments for `with_distributed` (especially those optional timeout arguments), and the arguments for the target function.

Averylamp (commented):
Generally LGTM, left a few small comments

tests/utils/trainer_fit.py
jbloxham: Yeah, this was causing a fair amount of flakiness in this test. This change is definitely "correct", but I'm confused why this wasn't causing more flakiness in the previous branch.

composer/trainer/ddp.py
jbloxham: I'll make the comment a bit more descriptive, but the issue is that the default depends on whether `find_unused_parameters` is on.

tests/fixtures/ddp_fixtures.py
jbloxham: `run_long` takes 110s for me before this change; 140s after.

ravi-mosaicml: Hmm kk...30s isn't a killer, but would be concerned if this delta grows with O(# of ddp tests).

This is fine for now, but would like to experiment in a future PR of doing `composer -m pytest ...`.

jbloxham: Yeah, that makes sense to me. I'm hoping at least that the need for N > 2 in multiprocessing tests should be extremely limited.

composer/trainer/ddp.py
```diff
@@ -113,40 +107,34 @@ class DDP:
...
+
+        if "RANK" in os.environ and "WORLD_SIZE" in os.environ:
+            # Assume we can initialize based off of env vars
+            torch.distributed.init_process_group(self.backend, timeout=timeout)
+            return
+
+        warnings.warn("RANK and WORLD_SIZE env vars not set; assuming no parallelization. If "
+                      "this is unexpected, make sure you are running your training script with "
+                      "the composer executable.")
+        store = torch.distributed.HashStore()
```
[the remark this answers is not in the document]
jbloxham: I've readded `TCPStore`. `FileStore` existed primarily for test reasons, but `HashStore` seems to be a totally appropriate alternative for the single-process case.

tests/fixtures/ddp_fixtures.py
jbloxham: Ahh but there needs to be _some_ non-zero value for the `.join` timeout - I'm fairly sure that non-positive values just get ignored. I think the ideal solution would be for this to have more robust monitoring logic but frankly I can't be bothered it implement that right now. I'm just going to move the `subsequent_timeout_per_device` to a constant so that users don't need to think about it.

tests/trainer/test_trainer.py
jbloxham: There's just nowhere for it to be destroyed. I don't think the trainer itself handles destroying its own process group.

tests/fixtures/ddp_fixtures.py
jbloxham: We do!

tests/fixtures/ddp_fixtures.py
jbloxham: Aight scratch that - I rewrote the monitoring logic to be more robust. This helps us fail faster in cases where the non-rank 0 process fails on launch.

composer/cli/launcher.py
```diff
@@ -0,0 +1,201 @@
...
+                                 master_port=args.master_port,
+                                 module_mode=args.module_mode,
+                                 training_script=args.training_script,
+                                 training_script_args=args.training_script_args)
+
+    try:
+        monitor_processes(processes)
+    finally:
+        cleanup_processes(processes)
+        return 0
```
ravi-mosaicml, line 197: If there's an error, let's return an error code rather than 0. Perhaps the code of the first process that returns a nonzero code.

composer/trainer/ddp.py
```diff
@@ -6,22 +6,16 @@
...
 from dataclasses import dataclass
-from threading import Thread
-from typing import Callable, ContextManager, Iterator, List, Optional, Sequence, Set, TypeVar, Union, cast
+from typing import Callable, ContextManager, Iterator, List, Optional, Sequence, TypeVar, cast
 
 import torch
 import torch.distributed
 import torch.utils.data
 import yahp as hp
+from torch.distributed.distributed_c10d import get_world_size
```
ravi-mosaicml, line 18:
```suggestion
from composer.utils.ddp import get_world_size
```

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
+
+def get_local_world_size() -> int:
+    return _get_distributed_config_var(env_var="LOCAL_WORLD_SIZE", human_name="local world size", default=1)
+
+
+def get_local_rank() -> int:
+    return _get_distributed_config_var(env_var="LOCAL_RANK", human_name="local rank", default=0)
 
 
 def is_rank_zero() -> bool:
```
ravi-mosaicml, line 52: `is_rank_set` below should be removed, since that will always return True now.
Any code guarded by it no longer needs to be guarded.

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
+                                default: int,
+                                fetch_fn: Optional[Callable[[], int]] = None) -> int:
     if not dist.is_available():
-        warnings.warn("Torch distributed is not available; returning 0 for global rank")
-        return 0
-    return dist.get_rank()  # raises a runtime error if not yet initialized
+        warnings.warn(f"Torch distributed is not available; returning {default} for {human_name}")
+        return default
+
+    if fetch_fn is not None:
```
ravi-mosaicml, line 18: In an environment where torch.distributed is not available, then I believe this would throw an attributeerror before `_get_distributed_config_var` is called. Instead, perhaps provide the fetch_fn by string and do a `getattr(torch.dist, fetch_fn_name)()`? This should not raise an attributeerror here cause of the check on line 14.

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
-def get_global_rank() -> int:
+def _get_distributed_config_var(env_var: str,
+                                human_name: str,
+                                default: int,
+                                fetch_fn: Optional[Callable[[], int]] = None) -> int:
     if not dist.is_available():
-        warnings.warn("Torch distributed is not available; returning 0 for global rank")
-        return 0
-    return dist.get_rank()  # raises a runtime error if not yet initialized
+        warnings.warn(f"Torch distributed is not available; returning {default} for {human_name}")
```
ravi-mosaicml, line 15:
```suggestion
        warnings.warn(f"DDPDefaultValueWarning: Torch distributed is not available; returning {default} for {human_name}")
```

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
+        return default
+
+    if fetch_fn is not None:
+        try:
+            return fetch_fn()
+        except RuntimeError:
+            pass
```
ravi-mosaicml, line 22:
```suggestion
            # Raises a RuntimeError if DDP is not yet initialized
            pass
```

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
+
+    if not env_var in os.environ:
+        warnings.warn(f"{env_var} env var not set"
```
ravi-mosaicml, line 25:
```suggestion
        warnings.warn(f"DDPDefaultValueWarning: {env_var} env var not set"
```

docs/source/getting_started/distributed.rst
```diff
@@ -0,0 +1,49 @@
...
+
+
+Detailed Usage
+---------------
+
+The following information can be easily accessed by typing ``composer -h`` into a command line.
+
+.. code-block::
+
+    usage: composer [-h] -n NPROC [--world_size WORLD_SIZE] [--base_rank BASE_RANK] [--master_addr MASTER_ADDR] [--master_port MASTER_PORT] [-m] training_script ...
```
ravi-mosaicml, line 28: Is it possible to auto-generate this via a call to `composer --help` during the readthedocs build?

docs/source/getting_started/distributed.rst
```diff
@@ -0,0 +1,49 @@
+Distributed Training (DDP)
+==========================
+
+Distributed data parallel, or DDP, is an essential technique for training large models. It is among the most common ways of parallelizing machine learning models. However, as are many things, parallelization in Python is not well supported. To get around these limitations, we offer a ``composer`` launch script to set up multiprocessing and synchronization. This script is directly analogous to the ``torch.distributed.run``, ``torchrun``, and ``deepspeed`` scripts that users may be familiar with.
```
ravi-mosaicml, line 4: Awesome docs

composer/utils/ddp.py
```diff
@@ -1,15 +1,52 @@
...
+from typing import Callable, Optional
 
 import torch.distributed as dist
 
 
-def get_global_rank() -> int:
+def _get_distributed_config_var(env_var: str,
+                                human_name: str,
+                                default: int,
+                                fetch_fn: Optional[Callable[[], int]] = None) -> int:
```
ravi-mosaicml, line 13: See comment below re: torch.distributed not being available.

```suggestion
                                fetch_fn: Optional[str] = None) -> int:
```

composer/cli/launcher.py
```diff
@@ -0,0 +1,201 @@
...
+    while len(processes) > 0:
+        for process in processes:
+            if process.poll() is None:
+                # the process is still running
+                continue
+            else:
+                # return code of 0 implies clean exit
+                # return code of -9 implies sigkill, presumably from cleanup_processes()
+                if process.returncode not in (0, -9):
+                    print(process)
```
ravi-mosaicml, line 115: Should this print be removed?

ravi-mosaicml (changes requested):
Looking really close...just a final few more comments.

composer/utils/ddp.py
jbloxham: Nice catch!

composer/cli/launcher.py
jbloxham: Woops, yes.

ravi-mosaicml (approved):
:100:

PR #66, opened 2021-11-05
Parameterized Dockerfiles

ravi-mosaicml:
1. Added a `Dockerfile` to build a pytorch image.
2. Parameterized the `Dockerfile` for composer, which extends the `Pytorch` image
3. Updated the `Makefile` to use the new parameterized Dockerfiles


docker/README.md
```diff
@@ -4,10 +4,10 @@ To simplify environment setup for the MosaicML `Composer` library, we provide a
...
 
 All Docker images come preinstalled with the following packages:
-* Ubuntu 18.04
-* Python 3.8.0
-* Nvidia CUDA 11.1.1
-* PyTorch 1.9.0
+* Ubuntu 20.04
+* Python
+* Pytorch
+* Nvidia CUDA (except if using a CPU-only flavor of pytorch)
```
bandish-shah, line 10: Let's document the matrix build here.  We are going to need to support this "legacy" image as well, also looks like we should have a Python 3.7 version of the image we can test with.

docker/pytorch/Dockerfile
```diff
@@ -1,6 +1,99 @@
...
+
+#################
+# Install Pytorch
+#################
+ARG PYTORCH_VERSION
+ARG TORCHVISION_VERSION
+ARG CUDA_VERSION_TAG
 
 RUN pip install --no-cache-dir --upgrade pip && \
+    pip install --no-cache-dir --upgrade pillow_simd==7.0.0.post3 && \
```
bandish-shah, line 82: Inconvenient but this needs to happen *after* uninstalling the torch default pillow.  Check out Installation Instructions here:
https://github.com/uploadcare/pillow-simd

docker/Makefile
```diff
@@ -1,16 +1,28 @@
...
-	IMAGE_TAG = ${BUILD_TYPE}
-endif
+BASE_IMAGE ?= 'nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu20.04'
+PYTHON_VERSION ?='3.9'
+CUDA_VERSION_TAG ?= 'cu113'
+PYTORCH_VERSION ?= '1.10.0'
+TORCHVISION_VERSION ?= '0.11.1'
+PYTORCH_TAG ?= mosaicml/pytorch
+COMPOSER_TAG ?= mosaicml/composer
+COMPOSER_EXTRA_DEPS ?= base
```
bandish-shah, line 8: [nit] Wondering if we should move these versions into a conditional similar to the one removed above?  Ideally we don't support more than a few images but right now already seems like we will have:
1. Initial released image we released with: Ubuntu 18.04, Python 3.8, PyTorch 1.9)
2. New image: Ubuntu 20.04, Python 3.9, PyTorch 1.10
3. Legacy image: Ubuntu 18.04?, Python 3.7, PyTorch 1.9?

docker/pytorch/Dockerfile
```diff
@@ -1,6 +1,99 @@
...
+#########
+# Cleanup
+#########
+RUN rm -rf /tmp/*
+
+#########################
+# Configure non-root user
+#########################
+RUN echo "export PATH=\$PATH:~/.local/bin" >> /etc/skel/.bashrc && \
+    useradd -rm -d /home/ubuntu -s /bin/bash -u 1000 -U -s /bin/bash ubuntu
```
bandish-shah, line 96: [nit] Let's rep MosaicML somehow has the user name!

bandish-shah (changes requested):
Few changes to address

docker/README.md
ravi-mosaicml: Sg, I will put this in table form. It's only a handful of images now.

ravi-mosaicml: Python 3.7 is addressed in #68. But I will update the docs here to reflect that.

ravi-mosaicml: I will wait on adding in the build matrix until we have our Jenkins CI/CD configured and docker images published. Let's discuss which images we want to be "officially supported".

docker/pytorch/Dockerfile
ravi-mosaicml: Sg! Named to user to mosaicml.

docker/Makefile
ravi-mosaicml: I added documentation for these variables. I would prefer to keep the documentation for the build matrix out of the Makefile, but agree this should be in the readme, once we have these images published and released on dockerhub.

bandish-shah (approved):
LGTM thanks for addressing the feedback!

PR #68, opened 2021-11-06
Python 3.7 Compatibility

ravi-mosaicml:
1. Fallbacking to `Protocol = object` in 3.7
2. Fallbacking to `final = lambda x: x` in python 3.7
3. Removed custom warning type for DeferredLogMetricWarning since pytest filterwarnigns errored


hanlint:
Before merging this, can we run some testing with some simple examples on colab?

hanlint:
A few other requests:
* update the docs/README where we mention requirements
* add something bout colab after it tested?

setup.py
```diff
@@ -27,7 +27,7 @@ def package_files(directory):
     "torchvision>=0.9.0",
     "torch>=1.8.1",
     "argparse>=1.4.0",
-    "yahp>=0.0.10",
+    "yahp>=0.0.11",
```
Averylamp, line 30: Don't worry about changing Yahp versions, I'll handle that separately just so I can ensure to track it properly

Averylamp: Otherwise looks good to me, but ditto on hanlins requests

ravi-mosaicml:
```suggestion
    "yahp>=0.0.10",
```

ravi-mosaicml: I'll keep it here just so we don't merge this in until new yahp is released (this diff should disappear once you add that commit in).

Averylamp: Yeah, fyi YAHP is already on `0.0.11`, so that isn't the right number to ensure it updates with YAHP either.  I think I'd rather just track all version numbers independently for now to keep it lower overhead for me.

Averylamp: Marked with a new label, but as this is a backwards compatible change you should be fine to merge into dev

Landanjs (commented):
I can't speak to the code, but it looks like the examples (MNIST and CIFAR) are trainingin [collab](https://colab.research.google.com/drive/1HscE_39ywgGZhI2glt-jF_PNElqh4Pez?usp=sharing)!

Could someone look at the notebook and see if anything else should be tested?

Averylamp (approved):
lgtm

PR #71, opened 2021-11-09
Removed duplicate callbacks

ravi-mosaicml:
The correct pytorch profiler file is `composer/callbacks/pytorch_profiler.py`


Averylamp (approved):
lgtm
file is: `composer/callbacks/torch_profiler.py`

hanlint:
@ravi-mosaicml can you also remove `timimg_monitor.py` which is duplicate of the `benchmarker.py` as well?

ajaysaini725:
> @ravi-mosaicml can you also remove `timimg_monitor.py` which is duplicate of the `benchmarker.py` as well?

Actually can we keep `timing_monitor.py` and remove `benchmarker.py` because of #74 which introduces a class called `Benchmarker` that uses `TimingMonitor`? Just trying to avoid a name conflict - open to suggestions for alternate names

ravi-mosaicml:
@ajaysaini725 The `timing_monitor.py` is out of date (Benchmarker is correct). Open to other names though for the what is currently called the benchmarker...perhaps `FastForwarder`?

PR #73, opened 2021-11-09
Added explicit cli_args=False for `load_model` and `load_algorithms`

ravi-mosaicml:
When using preset models or algorithms (e.g. via `load_model` or `load_algorithms`), CLI args were also being read, which lead to runtime errors in Colab (where the CLI arguments it was started with conflicted with the CLI arguments for the mosaic trainer. Instead, this change disables parsing of CLI arguments when using `load_model` and `load_algorithms`.


Averylamp (approved):
lgtm

PR #74, opened 2021-11-10
Benchmarking client functionality 

ajaysaini725:
Adds the following:
- A `Benchmarker` class which wraps a `Trainer` with a `BenchmarkerCallback` to estimate both throughput and wall clock time of training runs.
- A `run_benchmarker.py` entrypoint to instantiate and run the `Benchmarker` with `BenchmarkerHparams`.
- Utility functions for serializing and deserializing Torchscript model traces. Will be used by the benchmarking jobs which profile models without access to their code. 

wandb runs showing similar throughput for a traced model and a non-traced model: https://wandb.ai/mosaic-ml/ajay-benchmarking-torchscript

Pending: wandb runs showing the output of `run_benchmark.py`


composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+class Benchmarker:
+    """Benchmarker for timing model training workloads with synthetic data.
+
+    Args:
+        model (BaseMosaicModel): The model to profile.
+        data_shape (List[int]): Shape of the tensor for input samples.
+        total_batch_size (int): The batch size to train with.
+        grad_accum (int, optional): The number of microbatches to split a per-device batch into. Gradients
+            are summed over the microbatches per device.
+        data_type (SyntheticDataType, optional), Type of synthetic data to create.
```
jbloxham, line 31: Is this vestigial?

composer/core/tracing.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Args:
+            outputs (Any): The output of the forward pass.
+            batch (~composer.core.types.Batch): The input batch from dataloader.
+
+        Returns:
+            Tensors:
+                A stub of the loss as a ``Tensors`` object.
+        """
+        return torch.ones((1,))
```
jbloxham, line 52: Could instead raise an exception here.

composer/core/tracing.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Args:
+            batch (Batch): The input batch for the forward pass.
+
+        Returns:
+            Tensors:
+                The result that is passed to :meth:`loss` as a ``Tensors``
+                object.
+        """
+        return torch.ones((1,))
```
jbloxham, line 65: Here too.

composer/core/tracing.py
```diff
@@ -0,0 +1,100 @@
...
+                     for more details.
+
+        Args:
+            train (bool, optional): True to return metrics that should be computed
+                during training and False otherwise. (default: ``False``)
+
+        Returns:
+            Metrics: A ``Metrics`` object.
+        """
+        return MetricCollection([])
```
jbloxham, line 83: And here.

composer/core/tracing.py
```diff
@@ -0,0 +1,100 @@
...
+        Args:
+            batch (Batch): The data to perform validation with.
+                Specified as a tuple of tensors (input, target).
+
+        Returns:
+            Tuple[Any, Any]: Tuple that is passed directly to the
+                `update()` methods of the metrics returned by :meth:`metrics`.
+                Most often, this will be a tuple of the form (predictions, targets).
+        """
+        return (1, 1)
```
jbloxham, line 100: And here.

composer/benchmarker/benchmarker_hparams.py
```diff
@@ -0,0 +1,59 @@
...
+        template_default=1,
+        doc=
+        "Determines the number of microbatches to split a per-gpu batch into, used to compensate for low-memory-capacity devices."
+    )
+    data_shape: List[int] = hp.required("Shape of the data tensor.")
+    loggers: List[BaseLoggerBackendHparams] = hp.required(doc="loggers to use")
+    label_type: SyntheticDataLabelType = hp.optional("Type of synthetic label to create.",
+                                                     default=SyntheticDataLabelType.CLASSIFICATION_INT)
+    num_classes: int = hp.optional(
+        "Number of classes. Required if label_type is SyntheticDataLabelType.CLASSIFICATION.", default=2)
```
jbloxham, line 46:
```suggestion
        "Number of classes. Required if label_type is SyntheticDataLabelType.CLASSIFICATION_*.", default=2)
```

examples/run_benchmarker.py
```diff
@@ -0,0 +1,31 @@
...
+import composer
+from composer.benchmarker.benchmarker import Benchmarker
+from composer.benchmarker.benchmarker_hparams import BenchmarkerHparams
+from composer.trainer.trainer import Trainer
+from composer.trainer.trainer_hparams import TrainerHparams
+
+logger = logging.getLogger(__name__)
+
+
+def main() -> None:
```
jbloxham, line 16: Note that this file is gonna need to get ran by some version of the distributed script. The `composer` script should work just fine if you want to use that :)

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+                                        num_classes=num_classes,
+                                        label_shape=label_shape)
+
+        self.dataloader_spec = DataloaderSpec(dataset=self.dataset, shuffle=False, drop_last=True)
+
+        # Default for now - adjust to work with algorithms
+        timing_callback = TimingMonitor(min_steps=50, epoch_list=[0, 1], step_list=[0, 50], all_epochs=False)
+
+        # Use optimal device settings based on what is available
+        device = DeviceCPU(num_cpus=1)
```
jbloxham, line 64: `num_cpus` is an obsolete argument after the DDP refactor. Use the `composer` script instead to invoke multiprocessing.

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+        precision = Precision.FP32
+        if torch.cuda.is_available():
+            num_gpus = torch.cuda.device_count()
+            device = DeviceGPU(prefetch_in_cuda_stream=False, n_gpus=num_gpus)
```
jbloxham, line 68: `num_gpus` is obsolete too.

jbloxham (changes requested):
Mostly minor comments, with the one note that this needs to be reworked a bit to work with the DDP refactor.

hanlint (commented):
At a high-level, do we need to have the class and the hparams, or would functions and `argparse` cli be sufficient here?

The `Benchmarker` class doesn't carry a lot of state, and most of the work is just setup and then run. So an alternate proposal to consider would be:

```
# run_benchmarker.py

args = parser.parse_args()

# do setup code to create trainer
# self.trainer.fit()

# return result

```

one could also use `yahp` if you wanted YAML parsing support:
```
hparams = BenchmarkerHparams.create(args)

# setup trainer
trainer = Trainer(hparams.model, hparams,...)
trainer.fit()
# return result
```

ajaysaini725:
> At a high-level, do we need to have the class and the hparams, or would functions and `argparse` cli be sufficient here?
> 
> The `Benchmarker` class doesn't carry a lot of state, and most of the work is just setup and then run. So an alternate proposal to consider would be:
> 
> ```
> # run_benchmarker.py
> 
> args = parser.parse_args()
> 
> # do setup code to create trainer
> # self.trainer.fit()
> 
> # return result
> ```
> 
> one could also use `yahp` if you wanted YAML parsing support:
> 
> ```
> hparams = BenchmarkerHparams.create(args)
> 
> # setup trainer
> trainer = Trainer(hparams.model, hparams,...)
> trainer.fit()
> # return result
> ```

Yeah I considered doing it the way you're suggesting because right now the `Benchmarker` class is really lightweight. However, in future iterations it is going to contain some non-trivial logic (ex. determining which steps to time based on what algorithms are passed in) so I figured it's best to just separate them now and if down the line it turns out that it's just wrapping training with some additional logic then can wrap everything into `run_benchmark.py` without having the extra `Benchmarker` class. Thoughts?

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,108 @@
...
+        precision = Precision.FP32
+        if torch.cuda.is_available():
+            device = DeviceGPU(prefetch_in_cuda_stream=False)
+            precision = Precision.AMP
+
+        self.trainer = Trainer(
+            model=model,
+            train_dataloader_spec=self.dataloader_spec,
+            eval_dataloader_spec=self.dataloader_spec,
+            max_epochs=2,
```
Averylamp, line 73: The benchmarker callback should modify the `max_epochs` for you so you won't need to do that here (It'll be needed when algos modify  benchmarking behavior).  I believe `BenchmarkerCallback` should also handle the no-validation logic, which seems to not exist in `BenchmarkerCallback` right now.

composer/benchmarker/benchmarker_hparams.py
```diff
@@ -0,0 +1,59 @@
...
+    data_shape: List[int] = hp.required("Shape of the data tensor.")
+    loggers: List[BaseLoggerBackendHparams] = hp.required(doc="loggers to use")
+    label_type: SyntheticDataLabelType = hp.optional("Type of synthetic label to create.",
+                                                     default=SyntheticDataLabelType.CLASSIFICATION_INT)
+    num_classes: int = hp.optional(
+        "Number of classes. Required if label_type is SyntheticDataLabelType.CLASSIFICATION_*.", default=2)
+    label_shape: List[int] = hp.optional(
+        "Shape of the label tensor. Required if label_type is SyntheticDataLabelType.RANDOM_INT.",
+        default_factory=lambda: [1])
+    log_level: str = hp.optional(doc="Python loglevel to use composer", default="INFO")
```
Averylamp, line 50: I believe enums have been integrated so this can now be the `LogLevel` type

Averylamp (approved):
lgtm.  Probably need to test it with the `multiproc` entrypoint

composer/benchmarker/benchmarker.py
ajaysaini725: Oh yeah it's just that `max_epochs` is required so I filled it in with a dummy value.

composer/benchmarker/benchmarker_hparams.py
ajaysaini725: This is a different `LogLevel` - the `LogLevel` enum is for loggers but this is the python one

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+NUM_PROFILING_STEPS = 250  # needs to be high enough to be able to properly measure time
+
+
+class Benchmarker:
+    """Benchmarker for timing model training workloads with synthetic data.
+
+    Args:
+        model (BaseMosaicModel): The model to profile.
+        data_shape (List[int]): Shape of the tensor for input samples.
+        total_batch_size (int): The batch size to train with.
```
ravi-mosaicml, line 29: See #103. `data_shape` may not work for NLP data.

composer/datasets/synthetic.py
```diff
@@ -168,7 +169,7 @@ class SyntheticDatasetHparams(DatasetHparams):
     def validate(self):
         super().validate()
 
-        _validate_label_inputs(label_type=self.label_type, num_classes=self.num_classes, label_shape=self.label_shape)
+        validate_label_inputs(label_type=self.label_type, num_classes=self.num_classes, label_shape=self.label_shape)
```
ravi-mosaicml, line 172: If validation happens in `__init__`, does it need to happen again in `validate()`?

composer/utils/device_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+
+from composer.core.types import Batch, BatchPair, Tensor, Tensors
+from composer.utils.iter_helpers import map_collection
+
+
+def tensors_to_device(x: Tensors, device: torch.device) -> Tensors:
+    return map_collection(x, lambda t: cast(Tensor, t).to(device, non_blocking=True))
+
+
+def move_batch_to_gpu(batch: Batch, device: torch.device) -> Batch:
```
ravi-mosaicml, line 15:
```suggestion
def move_batch_to_device(batch: Batch, device: torch.device) -> Batch:
```

composer/utils/device_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+    """Move data to the GPU device.
+
+    Args:
+        batch (Batch): The data to move the gpu.
```
ravi-mosaicml, line 19:
```suggestion
        batch (Batch): The data to move onto the device.
        device (torch.device): The device to move the batch onto.
```

composer/utils/device_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    if isinstance(batch, Tensor):
+        return cast(Tensor, tensors_to_device(batch, device))
+    if isinstance(batch, (tuple, list)):  # BatchPair
+        # print('DOING BATCH PAIR!!!!!')
```
ravi-mosaicml, line 24:
```suggestion
```

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # Use optimal device settings based on what is available
+        device = DeviceCPU()
+        precision = Precision.FP32
+        if torch.cuda.is_available() and (not torch.distributed.is_initialized() or
+                                          torch.distributed.get_backend() == "nccl"):
+            device = DeviceGPU(prefetch_in_cuda_stream=False)
+            precision = Precision.AMP
+
+        self.trainer = Trainer(
```
ravi-mosaicml, line 71: do we need pay attention to optimizers and/or schedulers?

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+                                        num_classes=num_classes,
+                                        label_shape=label_shape)
+
+        self.dataloader_spec = DataloaderSpec(dataset=self.dataset, shuffle=False, drop_last=True)
+
+        # Default for now - adjust to work with algorithms
+        timing_callback = BenchmarkerCallback(min_steps=50, epoch_list=[0, 1], step_list=[0, 50], all_epochs=False)
+
+        # Use optimal device settings based on what is available
+        device = DeviceCPU()
```
ravi-mosaicml, line 64: can you add the device in as an optional argument? I think this assumption is fine if it's not specified, but during tests, it would be better to be explicit

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+from composer.datasets.hparams import DataloaderSpec
+from composer.datasets.synthetic import SyntheticDataLabelType, SyntheticDataset
+from composer.models.base import BaseMosaicModel
+from composer.trainer.devices import DeviceCPU, DeviceGPU
+from composer.trainer.trainer import Trainer
+
+NUM_PROFILING_STEPS = 250  # needs to be high enough to be able to properly measure time
+
+
+class Benchmarker:
```
ravi-mosaicml, line 23: For ease of usability, would it be possible to have the benchmarker take the same yaml (i.e. hparams) file as is used for training, plus a few additional benchmarker-specific arguments (e.g. num samples, and for the time being until #103 is done, synthetic dataset params). I'm thinking it would be really neat if you could invoke it like `benchmarker --num samples 100 -f path/to/training_hparams.yaml`

composer/core/tracing.py
```diff
@@ -0,0 +1,117 @@
...
+class _EmptyMosaicModel(BaseMosaicModel):
+    """An empty implementation of :class:`~composer.trainer.trainer.BaseMosaicModel`.
+
+    This class contains stub definitions for all the required methods in
+    :class:`~composer.trainer.trainer.BaseMosaicModel`. The intended use is for
+    loading models that have been traced as type BaseMosaicModel.
+    See :meth:`~composer.utils.load_model_trace` for an example.
+    """
+
+    def loss(self, outputs: Any, batch: Batch, *args, **kwargs) -> Tensors:
```
ravi-mosaicml, line 54: curious about the design decision to return defaults instead of raising `NotImplementedError()`?

or alternatively, could we create a subclass of `BaseMosaicModel` called, say `JITTracedMosaicModel`?

examples/run_benchmarker.py
```diff
@@ -0,0 +1,31 @@
...
+    logging.basicConfig()
+    logging.captureWarnings(True)
+
+    parser = argparse.ArgumentParser(parents=[BenchmarkerHparams.get_argparse(cli_args=True)])
```
ravi-mosaicml, line 20: I don't think you should need to import argparse / parse here? yahp takes care of this automatically.

ravi-mosaicml (commented):
Looks great for a v0 prototype, just a few questions throughout.

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,110 @@
...
+from composer.callbacks.benchmarker import Benchmarker as BenchmarkerCallback
+from composer.core.logging.base_backend import BaseLoggerBackend
+from composer.core.precision import Precision
+from composer.datasets.hparams import DataloaderSpec
+from composer.datasets.synthetic import SyntheticDataLabelType, SyntheticDataset
+from composer.models.base import BaseMosaicModel
+from composer.trainer.devices import DeviceCPU, DeviceGPU
+from composer.trainer.trainer import Trainer
+
+NUM_PROFILING_STEPS = 250  # needs to be high enough to be able to properly measure time
```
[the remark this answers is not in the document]
ajaysaini725: It actually needs to be intelligently set once we have algorithms so should leave it as a constant for now

composer/benchmarker/benchmarker.py
ajaysaini725: Yup - only meant to work for vision for now

composer/benchmarker/benchmarker.py
ajaysaini725: In the future there is a chance that this benchmarked will just merge with the trainer but for simplicity I made it a separate class for now

composer/benchmarker/benchmarker.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Args:
+        model (BaseMosaicModel): The model to profile.
+        data_shape (List[int]): Shape of the tensor for input samples.
+        total_batch_size (int): The batch size to train with.
+        grad_accum (int, optional): The number of microbatches to split a per-device batch into. Gradients
+            are summed over the microbatches per device.
+        label_type (SyntheticDataLabelType, optional), Type of synthetic data to create.
+            If `CLASSIFICATION_INT` or `CLASSIFICATION_ONE_HOT` then `num_classes` must be specified.
+            If `RANDOM_INT` then `label_shape` must be specified.
```
jbloxham, line 35: This comment needs double backticks, right?

composer/core/tracing.py
```diff
@@ -0,0 +1,117 @@
...
+def load_model_trace(filename: str) -> BaseMosaicModel:
+    jit_model = torch.jit.load(filename)
+
+    model = _JITTracedMosaicModel()
+    model.forward = jit_model.forward  # type: ignore
+    model.loss = jit_model.loss  # type: ignore
+    return model
+
+
+class _JITTracedMosaicModel(BaseMosaicModel):
```
ravi-mosaicml, line 45: I think this could be public? Wouldn't be opposed to having this be a first-class feature that someone could specify directly via hparams

composer/core/tracing.py
```diff
@@ -0,0 +1,117 @@
...
+    """An empty implementation of :class:`~composer.trainer.trainer.BaseMosaicModel`.
```
ravi-mosaicml, line 46: Update docstrings

composer/core/tracing.py
```diff
@@ -0,0 +1,117 @@
...
+
+    This class contains stub definitions for all the required methods in
+    :class:`~composer.trainer.trainer.BaseMosaicModel`. The intended use is for
+    loading models that have been traced as type BaseMosaicModel.
+    See :meth:`~composer.utils.load_model_trace` for an example.
+    """
+
```
ravi-mosaicml, line 53: Add an `__init__` function take the jit traced model?

Then you can have .loss() and .forward() call the jit traced model

composer/core/tracing.py
```diff
@@ -0,0 +1,117 @@
...
+        torch.jit.save(jit_model, save_file_path)
+
+    return jit_model
+
+
+def load_model_trace(filename: str) -> BaseMosaicModel:
+    jit_model = torch.jit.load(filename)
+
+    model = _JITTracedMosaicModel()
+    model.forward = jit_model.forward  # type: ignore
```
ravi-mosaicml, line 40: I missed this before, would prefer something like `model = JITTracedMosaicModel(jit_model)` rather than monkeypatching.

ravi-mosaicml: Then no need for the type ignores

ravi-mosaicml (approved):
Mostly LGTM, though see the comment re: JITTracedMosaicModel. Other than that I'm happy with it!

jbloxham (approved):
Looks good to me! Thanks for these changes!

hanlint:
Stale -- closing. Will submit new PR.

PR #75, opened 2021-11-10
Optimized trainer initializations

ajaysaini725:
If `device` is not provided the `Trainer` should default to using all GPUs available on a system with GPUs. If `precision` is not provided the `Trainer` should default to using `Precision.AMP` if training on a GPU.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -30,6 +30,7 @@ from composer.trainer.checkpoint import Checkpointer, CheckpointLoader
 from composer.trainer.ddp import DDP, DataloaderMultipleIterationWarning, StoreHparams, TCPStoreHparams
 from composer.trainer.devices.device import Device
 from composer.trainer.devices.device_cpu import DeviceCPU
+from composer.trainer.devices.device_gpu import DeviceGPU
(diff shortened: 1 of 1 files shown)


hanlint:
I actually had the alternate perspective, that if `device` and/or `precision` are not provided, the trainer should default to the naivest simplest thing possible (e.g. CPU with FP32) that consumes the least amount of resources. For example, implicitly using all available GPUs could accidentally disrupt other workloads when multiple users are sharing a GPU box.

ajaysaini725:
> I actually had the alternate perspective, that if `device` and/or `precision` are not provided, the trainer should default to the naivest simplest thing possible (e.g. CPU with FP32) that consumes the least amount of resources. For example, implicitly using all available GPUs could accidentally disrupt other workloads when multiple users are sharing a GPU box.

Okay makes sense. I'm not strongly opinionated about changing the trainer defaults this just came up because I wanted to automatically set the device + precision in #74 to reduce the number of user input parameters in benchmarking so figured it might make sense to just put that logic in the `Trainer` itself.

hanlint:
OK sounds good -- closing this PR, thanks!

PR #80, opened 2021-11-11
Run Directory for Artifacts

ravi-mosaicml:
Added a run directory for storing artifacts (e.g. checkpoints, non-rank-0 logs, torch profiler traces)

Unlike most hparams, the CLI launch script needs to be aware of the run directory, so it can create output files for the non-rank-0 processes. However, it would result in a confusing api to have this parameter specified both via the launch script and via a trainer hparam. 

Following the convention of other launch script arguments, the run directory is set as a CLI argument for the launch script, which then passes it through as a process-local environment variable.

Updated other components to use the run directory. Specifically:
* Updated the wandb logger to automatically log files placed in the run directory as artifacts
* Updated the checkpointer to store checkpoints inside the run directory by default
* Updated the pytorch profiler to store traces in the run directory by default, so they are visible in wandb.


composer/algorithms/alibi/gpt2_alibi.py
```diff
@@ -42,7 +42,7 @@ def _attn(self, query, key, value, attention_mask=None, head_mask=None):
 def enlarge_mask(module: torch.nn.Module, max_sequence_length: int):
     old_mask = module.bias
     new_mask = torch.tril(
-        torch.ones((max_sequence_length, max_sequence_length), dtype=torch.uint8,
+        torch.ones((max_sequence_length, max_sequence_length), dtype=torch.uint8,  # type: ignore
```
jbloxham, line 45: Why are new `type: ignore` statements needed? Are these just to silence warnings?

composer/cli/launcher.py
```diff
@@ -5,14 +5,14 @@
...
+import textwrap
 import time
 from argparse import ArgumentParser
-from typing import Any, List, Set
+from typing import Any, List, Optional, Set
 
 import torch.distributed
 
-CLEANUP_TIMEOUT = datetime.timedelta(seconds=5)
+CLEANUP_TIMEOUT = datetime.timedelta(seconds=30)
```
jbloxham, line 15: Admittedly I remember finding that the processes never seemed to exit from `SIGTERM` within 5 seconds, but I assumed they were just hanging/non-responsive, not that they needed more time. Did this fix it?

composer/cli/launcher.py
```diff
@@ -92,17 +104,18 @@ def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: s
...
 
         if local_rank == 0:
             process = subprocess.Popen(cmd, env=current_env, text=True)
         else:
             process = subprocess.Popen(
                 cmd,
                 env=current_env,
-                stdout=tempfile.TemporaryFile(),
-                stderr=tempfile.TemporaryFile(),
+                stdout=open(os.path.join(logs_dir, f"rank_{global_rank}.stdout.txt"), "x"),
```
jbloxham, line 117: This is awesome - very excited to have clearer log output!

composer/cli/launcher.py
```diff
@@ -92,17 +104,18 @@ def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: s
         current_env["LOCAL_WORLD_SIZE"] = str(nproc)
         current_env["MASTER_ADDR"] = master_addr
         current_env["MASTER_PORT"] = str(master_port)
+        current_env["RUN_DIRECTORY"] = run_directory
 
-        print(f"Launching process for local_rank({local_rank}), global_rank({global_rank})", local_rank, global_rank)
+        print(f"Launching process for local_rank({local_rank}), global_rank({global_rank})")
 
         if local_rank == 0:
             process = subprocess.Popen(cmd, env=current_env, text=True)
```
jbloxham, line 112: Is there a way to have the rank 0 process print to `stdout` _and_ a log file?

composer/loggers/wandb_logger.py
```diff
@@ -36,9 +44,39 @@ def state_dict(self) -> StateDict:
...
+        # On resnet50, _log_artifacts() caused a 22% throughput degradation
+        # wandb.log_artifact() is async according to the docs
+        # (see https://docs.wandb.ai/guides/artifacts/api#2.-create-an-artifact)
+        # so uploads will not block the training loop
+        # slowdown is likely from extra I/O
+        # Hence, logging every n batches instead of every batch
+        if (state.step + 1) % self.log_artifacts_every_n_batches == 0:
+            self._log_artifacts()
+
+    def epoch_end(self, state: State, logger: Logger) -> None:
```
jbloxham, line 58: I guess there's no reason we would need `log_artifacts_every_n_epochs` when `1` is a reasonable default?

jbloxham (approved):
This looks great! Only a couple questions :)

composer/algorithms/alibi/gpt2_alibi.py
ravi-mosaicml: It might be a pyright update, not sure. I was getting some errors.

composer/cli/launcher.py
ravi-mosaicml: If wandb has artifacts to upload on a failed experiment, I want to give it slightly longer than 5 seconds.

Re: never exiting. We forgot to add a `.poll()` in there, so the `returncode` was never updated. So it never appeared to exit :rofl:

jbloxham: Ah I see, that makes sense. XD

composer/loggers/wandb_logger.py
ravi-mosaicml: Correct, I figured every epoch is infrequent enough that it's OK to incur the cost triggering wandb then.

composer/algorithms/alibi/gpt2_alibi.py
ravi-mosaicml: image

composer/cli/launcher.py
ravi-mosaicml: Theoretically yes, though we would have to create our own file writer. My concern about having our own one would be re: buffering and TQDM, which is why I skipped it. But I agree this would be really nice to have.

ravi-mosaicml: (Experimenting with it more, python subprocess require a real FD. So, we'd have to write to a file, and then have a background thread to read from stdout/stderr, and reinterpret TQDM and terminal-like commands. This won't be trivial...)

jbloxham: Ok, that doesn't sound like a good use of our time :)

PR #81, opened 2021-11-12
Implement MosaicMLLoggerBackend

ajaysaini725:
Adds `MosaicMLLoggerBackend` for logging training run data to the MosaicML backend.
- `MosaicMLLoggerBackend` requires an api_key provided by MosaicML to work
- Logging happens in a separate thread to reduce performance overhead due to network I/O
- Buffers log data and flushes by sending data to an HTTP endpoint with a provided frequency


composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,163 @@
...
+                             json={
+                                 "experimentID": job_id,
+                                 "runID": sweep_id,
+                                 "data": data
+                             },
+                             timeout=3)
+    return response
+
+
+class MosaicMLLoggerBackend(RankZeroLoggerBackend, Serializable):
```
ajaysaini725, line 40: Thoughts on making this `Serializable`?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+_JOB_ID_ENV = "MOSAICML_LOGGER_JOB_ID"
+_SWEEP_ID_ENV = "MOSAICML_LOGGER_SWEEP_ID"
+
+_STOP_LOGGING_SIGNAL = "STOP_LOGGING"
+
+log = logging.getLogger(__name__)
+
+
+def _send_data(job_id: str, sweep_id: str, data: JSON):
+    response = requests.post(_MOSAICML_LOGGER_URL,
```
ajaysaini725, line 29: No need to use `aiohttp` here because this has a 3 second timeout and will not be called that often due to log buffering. Adding the async functionality just introduces unnecessary complexity since it is very unlikely that multiple post requests will interleave.

composer/loggers/logger_hparams.py
```diff
@@ -215,3 +216,27 @@ class TQDMLoggerBackendHparams(BaseLoggerBackendHparams):
...
+@dataclass
+class MosaicMLLoggerBackendHparams(BaseLoggerBackendHparams):
+    """:class:`~composer.loggers.mosaicml_logger.MosaicMLLoggerBackend`
+    hyperparameters.
+
+    See :class:`~composer.loggers.mosaicml_logger.MosaicMLLoggerBackend`
+    for documentation.
+    """
+    job_id: int = hp.required("The id of the job to write logs for.")
+    sweep_id: int = hp.required("The id of the sweep to write logs for.")
```
ajaysaini725, line 230: On second thought the logger should not have knowledge of sweeps that are happening instead the backend should look up the sweep based on the `job_id` - will remove this

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+                             timeout=3)
+    return response
+
+
+class MosaicMLLoggerBackend(RankZeroLoggerBackend, Serializable):
+    """Log to the MosaicML backend.
+
+    Args:
+        job_id (str, optional): The id of the job to write logs for.
+        sweep_id (str, optional): The id of the sweep to write logs for.
```
ajaysaini725, line 45: Same here - will remove

composer/loggers/__init__.py
```diff
@@ -8,5 +8,6 @@
 from composer.core.logging.logger import TLogDataValue as TLogDataValue
 from composer.loggers.logger_hparams import BaseLoggerBackendHparams as BaseLoggerBackendHparams
 from composer.loggers.logger_hparams import FileLoggerBackendHparams as FileLoggerBackendHparams
+from composer.loggers.logger_hparams import MosaicMLLoggerBackendHparams as MosaicMLLoggerBackendHparams
```
ravi-mosaicml, line 11: do we have a name for the MosaicML Logger Backend?

composer/loggers/logger_hparams.py
```diff
@@ -215,3 +216,27 @@ class TQDMLoggerBackendHparams(BaseLoggerBackendHparams):
...
+
+@dataclass
+class MosaicMLLoggerBackendHparams(BaseLoggerBackendHparams):
+    """:class:`~composer.loggers.mosaicml_logger.MosaicMLLoggerBackend`
+    hyperparameters.
+
+    See :class:`~composer.loggers.mosaicml_logger.MosaicMLLoggerBackend`
+    for documentation.
+    """
+    job_id: int = hp.required("The id of the job to write logs for.")
```
ravi-mosaicml, line 229: Curious on the design choice to use int's (as opposed to UUIDs) for the job_id / sweep id?

composer/loggers/logger_hparams.py
```diff
@@ -215,3 +216,27 @@ class TQDMLoggerBackendHparams(BaseLoggerBackendHparams):
...
+    sweep_id: int = hp.required("The id of the sweep to write logs for.")
+    creds_file: Optional[int] = hp.optional(
```
ravi-mosaicml, line 231: YAHP has automatic loading of environment variables. It's totally undocumented, but if you have an env variable with the same name as the cli flag, then YAHP will read it.

composer/loggers/logger_hparams.py
```diff
@@ -215,3 +216,27 @@ class TQDMLoggerBackendHparams(BaseLoggerBackendHparams):
...
+        "A file containing the MosaicML api_key. If not provided "
+        "will default to the environment variable MOSAIC_API_KEY",
+        default=None)
+    flush_every_n_batches: int = hp.optional("Flush the log data buffer every n batches.", default=100)
```
ravi-mosaicml, line 235: This is outside the scope of the PR, but we should probably have a global flush interval hparam, so there is a lot of flushing happening in a bunch of places.

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+from threading import Thread
+from typing import Optional
+
+import requests
+
+from composer.core.logging import LogLevel, RankZeroLoggerBackend, TLogData
+from composer.core.serializable import Serializable
+from composer.core.types import JSON, Logger, State, StateDict
+
+_MOSAICML_API_KEY_ENV = "MOSAICML_LOGGER_API_KEY"
```
ravi-mosaicml, line 18: see comment re: yahp reading of env variables

composer/loggers/mosaicml_logger.py
ravi-mosaicml: All Callbacks are serializable, and Loggers inherit callbacks.

composer/loggers/mosaicml_logger.py
ravi-mosaicml: Would it work to move network requests to a background thread? If it's in the main thread, and there's a network issue, then it will crash the training process. Or do we want training to be halted?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+def _send_data(job_id: str, sweep_id: str, data: JSON):
+    response = requests.post(_MOSAICML_LOGGER_URL,
+                             headers={"X-MosaicML-API-key": os.environ.get(_MOSAICML_API_KEY_ENV, "")},
+                             json={
+                                 "experimentID": job_id,
+                                 "runID": sweep_id,
+                                 "data": data
+                             },
+                             timeout=3)
+    return response
```
ravi-mosaicml, line 37: Does the requests library automatically raise for error codes? Or do we need to add in a `response.raise_for_status()` or the like?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+            "sweep_id": self.sweep_id,
+            "epoch": epoch,
+            "step": step,
+            "data": data,
+        }
+        self.buffered_data.append(log_data)
+        if len(self.buffered_data) > self.max_logs_in_buffer:
+            self._flush_buffered_data()
+
+    def _training_start(self, state: State, logger: Logger) -> None:
```
ravi-mosaicml, line 105: can you move this to the `init` callback? #85 will switch to expecting callbacks to be initialized on init, not training start

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+        self._flush_buffered_data()
+
+        self.queue.put_nowait(_STOP_LOGGING_SIGNAL)
+        self.thread.join()
+
+        log.info(f"MosaicML logger thread has exited.")
+
+    def state_dict(self) -> StateDict:
+        # Storing these fields in the state dict to support run resuming in the future.
+        return {"job_id": self.job_id, "sweep_id": self.sweep_id}
```
ravi-mosaicml, line 141: should unflushed data be recorded in the state?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+        while True:
+            data = self.queue.get(block=True)
+            if data == _STOP_LOGGING_SIGNAL:
+                log.info("MosaicML logger thread received stop logging signal.")
+                self.queue.task_done()
+                return
+
+            try:
+                response = _send_data(job_id=self.job_id, sweep_id=self.sweep_id, data=data)  # type: ignore
+                if response.status_code != 200:
```
ravi-mosaicml, line 166: Ah ok, I would move this into the _send_data function.
But should we ignore errors?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+                    # Ignore errors for now for simplicity
+                    warnings.warn("Posting data to MosaicML backend failed with response code "
```
ravi-mosaicml, line 168: By default, this warning.warn emits UserWarnings which implies that the user should do something differently to make the warning go away. I think here it may make sense to use `log.warn` since this is beyond the user's control?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+            return
+
+        data_to_write = self.buffered_data.copy()
+        self.buffered_data = []
+
+        self.queue.put_nowait(data_to_write)
+
+    def _listen_to_queue(self):
+        while True:
+            data = self.queue.get(block=True)
```
ravi-mosaicml, line 158: Instead of uploading each log entry individual (for which there will be a bunch! especially at the batch level), can you send the entire queue (or a large chunk of it) at once? I'd imagine that the network overhead would be pretty significant, and the HTTP headers would be a large % of the total request size (given that individual log entries are quite small).

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+    def _flush_buffered_data(self):
+        if len(self.buffered_data) == 0:
+            return
+
+        data_to_write = self.buffered_data.copy()
+        self.buffered_data = []
+
+        self.queue.put_nowait(data_to_write)
+
+    def _listen_to_queue(self):
```
ravi-mosaicml, line 156: Thoughts about having multiple listening threads for parallel uploads?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+        while True:
+            data = self.queue.get(block=True)
+            if data == _STOP_LOGGING_SIGNAL:
```
ravi-mosaicml, line 159: While this works, I think it'd be cleaner to use a `threading.Event` to indicate stopping (especially if there are multiple listening threads)

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,174 @@
...
+                response = _send_data(job_id=self.job_id, sweep_id=self.sweep_id, data=data)  # type: ignore
+                if response.status_code != 200:
+                    # Ignore errors for now for simplicity
+                    warnings.warn("Posting data to MosaicML backend failed with response code "
+                                  f"{response.status_code} and message {response.json()}.")
+            except requests.exceptions.Timeout as e:
+                warnings.warn(f"MosaicML logger timed out with error {e}.")
+
+            # Mark the task done regardless of error to not block thread termination
+            self.queue.task_done()
```
ravi-mosaicml, line 174: if there's a timeout, should this task be marked as done? or should it be retried?

ravi-mosaicml (changes requested):
I like where this is heading; just left some preliminary thoughts on it!

composer/loggers/logger_hparams.py
ajaysaini725: Lol typo it's supposed to be strings will fix

composer/loggers/mosaicml_logger.py
ajaysaini725: Given that it's only flushing every n logs / every n batches and we have a reasonable timeout set, one thread is probably enough for now

composer/loggers/mosaicml_logger.py
ajaysaini725: Uploading many logs at once is what is happening right now - `self.buffered_logs` stores many log entries and then either every n batches or when the number of buffered logs exceeds some max amount an upload happens where all buffered logs are uploaded in a single network request - see `_flush_buffered_data`

composer/loggers/mosaicml_logger.py
ajaysaini725: Decided to ignore errors for now because custom error handling introduces a lot of code complexity and it's not critical that all logs are uploaded anyway (the only use case so far is benchmarking and the cost of retrying a benchmarking job in the case where the final throughput isn't uploaded is low)

composer/loggers/mosaicml_logger.py
ajaysaini725: See comment about retrying

composer/loggers/mosaicml_logger.py
ravi-mosaicml: (discussed offline, leave as is)

composer/loggers/logger_hparams.py
ajaysaini725: Yup - we can consolidate in a separate PR

composer/loggers/mosaicml_logger.py
ajaysaini725: Right now we catch exceptions + I'll make a change such that the process will exit in the case of a non-transient error

composer/loggers/mosaicml_logger.py
ajaysaini725: Discussed offline - no need for `threading.Event` because there is only one thread

composer/loggers/logger_hparams.py
ravi-mosaicml: should this be an int?

ravi-mosaicml (approved):
LGTM, though I think creds_file should be Optional[str], not Optional[int]

ajaysaini725:
> LGTM, though I think creds_file should be Optional[str], not Optional[int]

Will fix ty 👍

hanlint:
cc: @nqn for a quick check

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,164 @@
...
+                                 json={
+                                     "experimentID": job_id,
+                                     "data": data,
+                                 },
+                                 timeout=5)
+        if response.status_code >= 400:
+            # Terminate the process in the case of a non-transient error
+            log.error("Posting data to MosaicML backend failed with response code "
+                      f"{response.status_code} and message {response.text}.")
+            sys.exit(1)
```
nqn, line 40: You could consider just to drop the batch, instead of killing the training job

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,164 @@
...
+        job_id (str, optional): The id of the job to write logs for.
+        creds_file (str, optional): A file containing the MosaicML api_key. If not provided
+            will default to the environment variable MOSAIC_API_KEY.
+        flush_every_n_batches (int): Flush the log data buffer every n batches. (default: ``100``)
+        max_logs_in_buffer (int): The maximum number of log entries allowed in the buffer before a
+            forced flush. (default: ``1000``)
+    """
+
+    def __init__(self,
+                 job_id: Optional[str] = None,
```
nqn, line 59: It is called experiment id other places - should we align on a single term, or use "jobs" for benchmarks and "experiments" for the rest?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,164 @@
...
+        self.queue = Queue()
+        self.thread = Thread(target=self._listen_to_queue)
+
+    def _log_metric(self, epoch: int, step: int, log_level: LogLevel, data: TLogData):
+        del log_level  # unused
+
+        if self.skip_logging:
+            return
+
+        log_data = {
```
nqn, line 93: An example call would be (with curl):

```
$ curl -X POST -H "Content-Type:application/json" -H "X-MosaicML-API-key:A106A6B8" https://api.mosaicml.com/v0/log/metric -d '{"experimentID": "79A8D3D2", "runID": "DE23902E", "data": {"acc/val": 0.5, "wall_clock_time": 123.0}}'
```
(Modulo, that the data field can be an array)

So we need to provide both experimentID and runID, and get the epoch and step in the data object.

This leads me to think we should team up and try this out real quick, and ensure that it works.

nqn (changes requested):
TL;DR We need to sync real quick to make sure the endpoints and submitted fields are correct.

composer/loggers/mosaicml_logger.py
ajaysaini725: Discussed this with @ravi-mosaicml - we drop the batch in the case of a transient error but thought it was best to just terminate the process in the case of a non-transient error because this usually means that none of the batches will be logged (ex. the user provided an incorrect api key or something).

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,224 @@
...
+        self._flush_buffered_data()
+
+        self.queue.put_nowait(_STOP_LOGGING_SIGNAL)
+        self.thread.join()
+
+        log.info(f"MosaicML logger thread has exited.")
+
+    def state_dict(self) -> StateDict:
+        # Storing these fields in the state dict to support run resuming in the future
+        return {"run_id": self.run_id, "buffered_data": self.buffered_data}
```
ajaysaini725, line 199: Need to persist `run_name` and `experiment_name` here

nqn (approved):

PR #84, opened 2021-11-15
Renamed run directory

ravi-mosaicml:
Renamed to `runs` from `.runs`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,4 +152,4 @@ README.md.bkp
 generated/
 
 # Run directory
-.runs/
+runs/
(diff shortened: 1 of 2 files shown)


hanlint (changes requested):
I believe the default still needs to be set in the code itself. e.g. currently still have:
```
run_directory = os.path.join(".runs", datetime.datetime.now().isoformat())
```

ravi-mosaicml:
Whoops - thanks!

hanlint (approved):

PR #85, opened 2021-11-15
`run_event` for callbacks, removed deferred logging

### LINKED ISSUE #11: Add `callback.run_event()`
Add a helper method in `callback` for `run_event`. This helper method would then call the correct method on callback. It would look something like:

```python3
class Callback:
  def run_event(self, state; State, logger; Logger, event: Event):
    if event == Event.TRAINING_START:
        self.training_start(state, logger)
    if event == Event.BEFORE_FORWARD:
        self.before_forward(state, logger)
    ...
```

Then, the engine would do `callback.run_event(state, logger, event)`.

This would help clean up code in the following places:
* RankZeroCallback: Instead of monkeypatching each callback function, it would simply override run_event.
* RankZeroLogger: No need for a private _training_start method that is different from all of the other callbacks
* Checkpointing tests: The `EventCounterCallback` basically does this, via monkeypatching

### LINKED ISSUE #87: Remove deferred logging
With #65, the global rank is now known when the python process starts. Thus, for rank zero loggers, it is not necessary to wait until training start to initialize the logger. Instead, loggers should initialize on the INIT event, and process all logging calls immediately.
By convention, there will not be any calls to the loggers before the init event.

ravi-mosaicml:
## 1. Added a `run_event` function for callbacks.

Previously, callbacks had one method corresponding to each event name. This format isn't flexible to callbacks that need to run on all events, and to future profiling work that may use callbacks to run on things other than events.

Instead, this PR adds a more generic structure, where a callback can instead implement a `def _run_event(event, state, logger)`. The engine calls `callback.run_event()` which dispatches to `callback._run_event()`. By default, `callback._run_event()` calls the methods named after each event, but individual callbacks can override it for generic functionality. 

Closes #11, cleaned up some of the tests and rank zero callbacks.

## 2. Removed deferred logging

#65 made the rank available upon startup, which made the deferred logging docs out of date and the use case for deferred logging obsolete. Now, callbacks initialize themselves upon `Event.INIT`. Closes #87.


composer/core/callback.py
```diff
@@ -58,246 +46,29 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
         try:
             event_cb = getattr(self, event.value)
```
jbloxham, line 47: I'm not really sure under what circumstances this would actually succeed. We'd need something like:

```python
class MyCallback(Callback):
    def batch_start(...):
        super().run_event(...)
        ...
```
Do you think it's fine to just drop the deprecation warning?

composer/core/callback.py
```diff
@@ -58,246 +46,29 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
-        del state, logger  # unused
-        pass
-
 
 class RankZeroCallback(Callback, abc.ABC):
     """Base class for callbacks that only run on the rank zero process.
+
+    Subclasses should override :meth:`_run_event`
+    (**not** `run_event`) to run in response
+    to given :class:`Event` invocations.
```
jbloxham, line 63: I like this concept, but see no reason to limit it to `RankZeroCallback`. Could it be imposed for `Callback` instead? It would mean that calling `super().run_event` is no longer necessary.

composer/loggers/tqdm_logger.py
```diff
@@ -101,73 +102,37 @@ def _will_log(self, state: State, log_level: LogLevel) -> bool:
...
+        super()._run_event(event, state, logger)
 
     def state_dict(self) -> StateDict:
-        state = {"is_validating": self.is_validating}
-        if self.pbar_train:
-            state["pbar_train"] = self.pbar_train.state_dict()
-        if self.pbar_val:
-            state["pbar_val"] = self.pbar_val.state_dict()
-        return state
+        return {"pbars": {k: v.state_dict() for (k, v) in self.pbars.items()}}
```
jbloxham, line 135: Does the `state_dict` not need to store `active_pbar`? I guess it probably wasn't possible to checkpoint in the middle of validation, but it's totally possible to checkpoint in the middle of a batch and I'm wondering if that can lead to subtle issues when restoring from a checkpoint.

docs/source/core/callback.rst
```diff
@@ -12,8 +12,7 @@ By convention, callbacks should not modify the :class:`State`.
 
 
 Each callback inherits from the :class:`Callback` base class,
-and overrides functions corresponding to the event.
-
+and overrides the :meth:`~Callback.run_event` method.
```
jbloxham, line 15: If you do the `run_event => _run_event` change, just make sure to change this.

composer/core/logging/base_backend.py
```diff
@@ -150,22 +151,11 @@ def log_metric(self, epoch: int, step: int, log_level: LogLevel, data: TLogData)
...
-    @final
-    def training_start(self, state: State, logger: Logger) -> None:
-        self._training_start(state, logger)  # initialize the logger
-        if self._deferred_log_metric_calls is None:
-            raise RuntimeError("_deferred_log_metric_calls should not be None")
-        for epoch, step, log_level, data in self._deferred_log_metric_calls:
-            self._log_metric(epoch, step, log_level, data)
-        self._deferred_log_metric_calls = None
+    def _run_event(self, event: Event, state: State, logger: Logger) -> None:
+        super()._run_event(event, state, logger)
```
jbloxham, line 155: Is this `super()._run_event` actually necessary? The superclass implementation doesn't do anything.

jbloxham (approved):
Looks good! Mainly just one suggestion around going full-send with `run_event => _run_event`.

composer/loggers/tqdm_logger.py
ravi-mosaicml: (Yup, added it back in)

composer/core/callback.py
ravi-mosaicml: I'd like to keep it in for now in case if there are any user-implemented callbacks that are no longer getting triggered.

ravi-mosaicml: The engine calls run_event, so if it's not overridden, then this check would happen (and deprecation warnings would be raised)

composer/core/callback.py
ravi-mosaicml: Super calls no longer necessary.

composer/core/logging/base_backend.py
ravi-mosaicml: Removed

hanlint (changes requested):
This is a pretty major API change to `callbacks`, and away from how folks usually implement callbacks in other ML frameworks. I think this is worth a discussion before we merge.

ajaysaini725 (dismissed):
LGTM!

ravi-mosaicml:
Would it work if we supported both styles for callbacks (e.g. I'll remove the DeprecationWarning and add back in all the stub methods for the events?) That way, if a callback doesn't override _run_event, then the method with the event name would be invoked; if a callback does override _run_event, then one is responsible for triggering any event logic.

composer/loggers/tqdm_logger.py
```diff
@@ -101,73 +102,44 @@ def _will_log(self, state: State, log_level: LogLevel) -> bool:
...
-    def eval_end(self, state: State, logger: Logger) -> None:
-        del state, logger  # Unused
-        assert self.pbar_val is not None
-        self.pbar_val.close()
-        self.pbar_val = None
-        self.is_validating = False
+            assert self.is_train is not None
+            self.pbars[self.is_train].log_metric(data)
+
+    def _run_event(self, event: Event, state: State, logger: Logger) -> None:
```
hanlint, line 110: For readability, let's keep this logger implemented in the method-based approach. Can use common functions to keep this DRY.

composer/loggers/wandb_logger.py
```diff
@@ -67,8 +61,19 @@ def epoch_end(self, state: State, logger: Logger) -> None:
...
             self._upload_artifacts()
 
+    def training_end(self, state: State, logger: Logger) -> None:
+        del state, logger  # unused
+        if self._log_artifacts:
+            self._upload_artifacts()
+
     def _upload_artifacts(self):
         # Scan the run directory and upload artifacts to wandb
+        # On resnet50, _log_artifacts() caused a 22% throughput degradation
```
hanlint, line 71: :(

Any solutions here? Would prefer for something non-critical (e.g. logging artifacts) inadvertently cause a slowdown for users.

hanlint (changes requested):
Overall looks good, requesting some changes:
* LoggerTQDM implementation refactor into the method-based style for readability
* I noticed the ~20% degradation comment. Are there solutions here?

composer/loggers/wandb_logger.py
ravi-mosaicml: Yup! `_upload_artifacts` is called only once every 100 (or a user configurable amount) batches, so this cost is amortized and becomes negligable.

hanlint (approved):
LGTM.

PR #86, opened 2021-11-17
DDP Port Auto Selection; Removal of Test Spawning

ravi-mosaicml:
* Added port auto-selection to the launcher script. Included a warning regarding the possible (but unlikely) race condition.
* In tests, renamed `pytest.mark.n_gpus` to `pytest.mark.world_size`
* Removed spawning in tests. Before, each test was responsible for spawning the processes it needed. Now, the `composer` cli is used to launch pytest for each world size. (So, the same rank 1 pytest process is used for all `world_size=2` tests, rather than spawning a new rank 1 process per tests). This change significantly reduces test time down to ~1 minute from ~4 minutes before (spawning is slow! especially when doing it multiple times)
* Added `./scripts/test.sh` as a helper, since running `pytest` from the command line would only run world-size 1 tests.


composer/cli/launcher.py
```diff
@@ -90,6 +99,13 @@ def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: s
     logs_dir = os.path.join(run_directory, "logs")
     os.makedirs(logs_dir, exist_ok=True)
 
+    if master_port is None:
+        warnings.warn("AutoSelectPortWarning: The DDP port was auto-selected. "
+                      "This may lead to race conditions when launching multiple training processes simultaneously. "
+                      "To eliminate this race condition, explicitely specify a port with --master_port PORT_NUMBER")
```
hanlint, line 105:
```suggestion
                      "To eliminate this race condition, explicitly specify a port with --master_port PORT_NUMBER")
```

composer/cli/launcher.py
```diff
@@ -166,11 +182,6 @@ def monitor_processes(processes: Set[subprocess.Popen]):
 
 
 def cleanup_processes(processes: Set[subprocess.Popen]):
-    if len(processes) == 0:
```
hanlint, line 169: Curious to know the reason we no longer need these cleanup checks

ravi-mosaicml: The launched processes, not the launcher, should call the destroy process group. But since we use the same process group for the entire duration of the python process (regardless of how many times .fit() is called), we do not destroy it anymore.

ravi-mosaicml:
Closing for #91

PR #89, opened 2021-11-19
Add option to disable artifact logging in WandB

ravi-mosaicml:
Added a log_artifacts (default is False) hparam to control whether artifacts are logged to WandB.

### DIFF
--- a/composer/loggers/logger_hparams.py
+++ b/composer/loggers/logger_hparams.py
@@ -82,6 +82,9 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
         name (str, optional): Weights and Biases run name.
         entity (str, optional): Weights and Biases entity name.
         tags (str, optional): Comma-seperated list of tags to add to the run.
+        log_artifacts (bool, optional): Whether to log artifacts. Defaults to False.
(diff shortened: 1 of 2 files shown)


siriuslee (approved):
LGTM. I've added an issue (#90) to be sure we get the multiple artifacts with same ID bug fixed. This is the right change for now.

composer/loggers/logger_hparams.py
```diff
@@ -82,6 +82,9 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
         name (str, optional): Weights and Biases run name.
         entity (str, optional): Weights and Biases entity name.
         tags (str, optional): Comma-seperated list of tags to add to the run.
+        log_artifacts (bool, optional): Whether to log artifacts. Defaults to False.
```
hanlint, line 85: This should say "automatically log artifacts in runs folder" correct? manual logs would still be enabled?

ravi-mosaicml: Theoretically, yes, but there is no api available in our trainer to call `wandb.log_artifact` (since our base logger doesn't expose an artifact API)

PR #91, opened 2021-11-19
DDP Port Auto Selection; Removed spawning in tests

ravi-mosaicml:
* Added port auto-selection to the launcher script. Included a warning regarding the possible (but unlikely) race condition.
* In tests, renamed `pytest.mark.n_gpus` to `pytest.mark.world_size`
* Removed spawning in tests. Before, each test was responsible for spawning the processes it needed. Now, the `composer` cli is used to launch pytest for each world size. (So, the same rank 1 pytest process is used for all `world_size=2` tests, rather than spawning a new rank 1 process per tests). This change significantly reduces test time down to ~1 minute from ~4 minutes before (spawning is slow! especially when doing it multiple times)
* Added `./scripts/test.sh` as a helper, since running `pytest` from the command line would only run world-size 1 tests.

(Replaces #86 so it no longer depends on Jenkins)


ajaysaini725 (approved):
LGTM - only ask is to add documentation for how to run the tests using the new codepath 😄

PR #92, opened 2021-11-19
Create dataloader on trainer __init__()

ravi-mosaicml:
#65 made the global rank available in the process start, so it is no longer necessarry to wait until training_start() to create the dataloader. Instead, dataloaders are now initialized in __init__.

This change will help with dataloader profiling, as now the dataloader will be immediately bound to the state.


composer/core/state.py
```diff
@@ -108,6 +108,10 @@ class State(Serializable):
     # stopping conditions
     max_epochs: int
 
+    # dataloaders
+    train_dataloader: types.DataLoader
+    eval_dataloader: types.DataLoader
```
ajaysaini725, line 113: @anisehsani should be aware of this change since he's working on supporting multiple eval dataloaders

ajaysaini725 (approved):
This doesn't change the external API at all so LGTM 

I think we should have some discussion around whether to have the `__init__` take any information about datasets at all or have the train + eval datasets be passed directly into `fit()` as well (this is a standard among other ML libraries like sklearn and HuggingFace).

composer/core/state.py
anisehsani: Thanks for the ping. This actually simplifies some thing for me!

ravi-mosaicml: Would be down to discuss this, though my initial thoughts are to keep fit as a zero-argument function, which can allow `__init__` to be used as a kind of static analysis check.

Issue #90, opened 2021-11-19
WandB error due to multiple artifacts with the same ID


siriuslee:
When running a baseline resnet50 model on imagenet, I encountered this error:
```
wandb: ERROR Error while calling W&B API: Error 1062: Duplicate entry '6394579-1' for key 'unique_artifact_collection_membership_version' (<Response [409]>)
Exception in thread Thread-7:
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.8/dist-packages/wandb/filesync/step_upload.py", line 50, in _thread_body
    self._handle_event(event)
  File "/usr/local/lib/python3.8/dist-packages/wandb/filesync/step_upload.py", line 79, in _handle_event
    self._maybe_commit_artifact(job.artifact_id)
  File "/usr/local/lib/python3.8/dist-packages/wandb/filesync/step_upload.py", line 161, in _maybe_commit_artifact
    self._api.commit_artifact(artifact_id)
  File "/usr/local/lib/python3.8/dist-packages/wandb/sdk/internal/internal_api.py", line 2235, in commit_artifact
    response = self.gql(mutation, variable_values={"artifactID": artifact_id})
  File "/usr/local/lib/python3.8/dist-packages/wandb/sdk/lib/retry.py", line 102, in __call__
    result = self._call_fn(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/wandb/sdk/internal/internal_api.py", line 147, in execute
    six.reraise(*sys.exc_info())
  File "/usr/local/lib/python3.8/dist-packages/six.py", line 719, in reraise
    raise value
  File "/usr/local/lib/python3.8/dist-packages/wandb/sdk/internal/internal_api.py", line 141, in execute
    return self.client.execute(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/wandb/vendor/gql-0.2.0/gql/client.py", line 52, in execute
    result = self._get_result(document, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/wandb/vendor/gql-0.2.0/gql/client.py", line 60, in _get_result
    return self.transport.execute(document, *args, **kwargs)
[...]

PR #101, opened 2021-11-22
Run Directory Uploader

### LINKED ISSUE #98: Blob Store Uploading for the Run Directory
## 🚀 Feature Request
Add callbacks to upload the run directory to blob stores (s3, gcs)

## Motivation
Currently, the run directory is only saved locally (or, uploaded to WANDB, but we're running into issues with that). When a K8S pod dies, we lose the run directory. We store logs, checkpoints, traces, etc... in the run directory, so this should be persisted.

## [Optional] Implementation

This can be implemented via a callback, quite trivially. It would be best to delegate the directory monitoring / uploading to a subprocess (not sub thread), as not to use GIL time in the main training loop. While network I/O happens outside the GIL, other work related to uploading (e.g. computing file hashes) does occur within the GIL, so it would be best to offload this. However, an initial implementation can use a background thread.

For cross-cloud compatibility, going to use apache libcloud.

ravi-mosaicml:
Run Directory Uploader
    
Added uploading of the run directory to various cloud providers via a callback. Depends on the LibCloud plugin. Did not use s3 as azure blob store is not s3-compatible.
    
Closes #98.

TODO:
- [x] Add in prefix to hparams
- [x] Manual tests on cloud providers (e.g. GCS)
- [x] Merge #105
- [x] Add in DDP barriers (may require attaching the ddp object to state, or making it global)


abhi-mosaic:
QQ: is the intention here that we should always be writing our checkpoints to our personal object store? And not using something like WandB artifacts (https://wandb.ai/wandb/common-ml-errors/reports/How-to-Save-and-Load-Models-in-PyTorch--VmlldzozMjg0MTE#save-as-artifacts)

ravi-mosaicml:
> QQ: is the intention here that we should always be writing our checkpoints to our personal object store? And not using something like WandB artifacts (https://wandb.ai/wandb/common-ml-errors/reports/How-to-Save-and-Load-Models-in-PyTorch--VmlldzozMjg0MTE#save-as-artifacts)

Yes, ideally we would use wandb internally, but we have been running into issues with getting 4xx errors returned by their client / API (e.g. #90). Also external customers may prefer to use a blob store of their own.

composer/callbacks/callback_hparams.py
```diff
@@ -153,3 +155,59 @@ class TorchProfilerHparams(CallbackHparams):
...
+
+
+@dataclass
+class RunDirectoryUploaderHparams(CallbackHparams):
+    """:class:`~composer.callbacks.torch_profiler.RunDirectoryUploader` hyperparameters.
+
+    See :class:`~composer.callbacks.torch_profiler.RunDirectoryUploader` for documentation.
+    """
+
+    # Args:
```
ajaysaini725, line 167: Remove extra comments?

composer/callbacks/callback_hparams.py
```diff
@@ -153,3 +155,59 @@ class TorchProfilerHparams(CallbackHparams):
...
+    #     provider_init_kwargs (Dict[str, Any], optional): Parameters to pass into the constructor for the
+    #         :class:`~libcloud.storage.providers.Provider` constructor. These arguments would usually include the cloud region
+    #         and credentials. Defaults to None, which is equivalent to an empty dictionary.
+    provider: str = hp.required("Cloud provider to use.")
+    container: str = hp.optional("he name of the container (i.e. bucket) to use.", default=None)
```
ajaysaini725, line 172: "he" -> "the"

composer/callbacks/callback_hparams.py
```diff
@@ -153,3 +155,59 @@ class TorchProfilerHparams(CallbackHparams):
...
+    key: Optional[str] = hp.optional(textwrap.dedent(
+        """API key or username to use to connect to the provider. For security. do NOT hardcode the key in the YAML.
+        Instead, please specify via CLI arguments, or even better, environment variables."""),
+                                     default=None)
+    secret: Optional[str] = hp.optional(textwrap.dedent(
```
ajaysaini725, line 177: Should probably just make this a file path rather than rely on the user to not hardcode the key - see #81 for an example

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,250 @@
...
+        run_directory = get_run_directory()
+        if run_directory is None:
+            warnings.warn("NoRunDirectory: The run directory is not set, so the RunDirectoryUploader will be a no-op")
+            return
+
+        if provider_init_kwargs is None:
+            provider_init_kwargs = {}
+        self._provider_init_kwargs = provider_init_kwargs
+        self._upload_every_n_batches = upload_every_n_batches
+        self._object_name_prefix = ""  # TODO ravi. Decide how this will be set. Hparams? Run directory name?
```
ajaysaini725, line 99: Perhaps everything related to the file upload config should just be in a config file that gets read to populate these fields.

Thinking about it from a user perspective, we'd like to avoid people having to hardcode any kind of sensitive info (including blobstore paths)  in their code

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,250 @@
...
+        init_kwargs (Dict[str, Any]): Arguments to pass in to the
+            :class:`~libcloud.storage.providers.Provider` constructor.
+    """
+    from libcloud.storage.providers import get_driver
+    provider_cls = get_driver(provider_name)
+    provider = provider_cls(**init_kwargs)
+    container = provider.get_container(container_name)
+    while True:
+        try:
+            file_path_to_upload = file_queue.get_nowait()
```
ajaysaini725, line 235: Cleaner to do `file_queue.get(block=True, timeout=0.5)` to avoid having to do the `time.sleep()`

ajaysaini725 (commented):
Just some initial comments

composer/callbacks/run_directory_uploader.py
ravi-mosaicml: It's fine to hardcode in everything except the secret. Blobstore paths should not be sensitive. I think it'd be common to hardcode a bucket to use, for example.

composer/callbacks/callback_hparams.py
ravi-mosaicml: Going to leave it as is for now. Would rather not add extra environment variables in everywhere. By January, YAHP will have better inheritance for secret files.

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,319 @@
...
+            worker.join()
+
+    def _trigger_upload(self, logger: Optional[Logger], log_level: Optional[LogLevel]) -> None:
+        # Ensure that every rank is at this point
+        # Assuming only the main thread on each rank writes to the run directory, then the barrier here will ensure
+        # that the run directory is not being modified after we pass this barrier
+        ddp.barrier()
+        new_last_uploaded_timestamp = time.time()
+        # Now, for each file that was modified since self._last_upload_timestamp, copy it to the temporary directory
+        # IMPROTANT: From now, until self._last_upload_timestamp is updated, no files should be written to the run directory
```
ajaysaini725, line 205: IMPROTANT -> IMPORTANT

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,319 @@
...
+        run_directory = get_run_directory()
+        assert run_directory is not None, "invariant error"
+        files_to_be_uploaded = []
+
+        # check if any upload threads have crashed. if so, then shutdown the training process
+        for worker in self._workers:
+            if not worker.is_alive():
+                assert self._finished is not None, "invariant error"
+                self._finished.set()
+                raise RuntimeError("Upload worker crashed unexpectedly")
```
ajaysaini725, line 215: Should add a TODO with an issue for better error handling - ideally we have some retry logic for something like this long term 

Can leave it as-is for now though to keep things simple

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,319 @@
...
+                )
+            except LibcloudError as e:
+                # The S3 driver does not encode the error code in an easy-to-parse manner
+                # So doing something fairly basic to retry on transient error codes
+                if any(x in str(e) for x in ("408", "409", "425", "429", "500", "503", '504')):
+                    if retry_counter < 3:
+                        retry_counter += 1
+                        # exponential backoff
+                        time.sleep(2**(retry_counter - 1))
+                        continue
```
ajaysaini725, line 315: Should include a warning if the retry is maxed out and there is still an error when uploading.

ajaysaini725 (commented):
Just a few notes but otherwise looks good!

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,319 @@
...
+            try:
+                provider.upload_object(
+                    file_path=file_path_to_upload,
+                    container=container,
+                    object_name=object_name_prefix + obj_name,
+                )
+            except LibcloudError as e:
+                # The S3 driver does not encode the error code in an easy-to-parse manner
+                # So doing something fairly basic to retry on transient error codes
+                if any(x in str(e) for x in ("408", "409", "425", "429", "500", "503", '504')):
```
ajaysaini725, line 310: Should 500 be included in transient errors? It probably shouldn't

ravi-mosaicml: 500s are the catchall...would rather retry than not, in case if it is transient like a race

composer/callbacks/run_directory_uploader.py
ravi-mosaicml: I added log.warn for each retry so this will be evident when / if it fails

composer/callbacks/run_directory_uploader.py
ravi-mosaicml: We retry in the handler for transient error codes. Would much rather crash if we are not logging things (and stop the training process) rather than continue training without logging (which can waste $$ / credits). This really should not error here unless if a transient error propegates through.

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,323 @@
...
+        self._last_upload_timestamp = 0.0  # unix timestamp of last uploaded time
+        if upload_staging_folder is None:
+            self._tempdir = tempfile.TemporaryDirectory()
+            self._upload_staging_folder = self._tempdir.name
+        else:
+            self._tempdir = None
+            self._upload_staging_folder = upload_staging_folder
+
+        if num_concurrent_uploads < 1:
+            raise ValueError("num_concurrent_uploads must be >= 1. Blocking uploads are not supported.")
```
Averylamp, line 131:
```suggestion
            raise ValueError("num_concurrent_uploads must be <= 1. Blocking uploads are not supported.")
```

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,323 @@
...
+        files_to_be_uploaded = []
+
+        # check if any upload threads have crashed. if so, then shutdown the training process
+        for worker in self._workers:
+            if not worker.is_alive():
+                assert self._finished is not None, "invariant error"
+                self._finished.set()
+                raise RuntimeError("Upload worker crashed unexpectedly")
+        for root, dirs, files in os.walk(run_directory):
+            del dirs  # unused
```
Averylamp, line 216:
```suggestion
        for root, _, files in os.walk(run_directory):
```

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,323 @@
...
+                if any(x in str(e) for x in ("408", "409", "425", "429", "500", "503", '504')):
+                    if retry_counter < 3:
+                        retry_counter += 1
+                        # exponential backoff
+                        sleep_time = 2**(retry_counter - 1)
+                        log.warn("Request failed with a transient error code. Sleeping %s seconds and retrying",
+                                 sleep_time,
+                                 exc_info=e,
+                                 stack_info=True)
+                        time.sleep()
```
Averylamp, line 318:
```suggestion
                        time.sleep(sleep_time)
```

setup.py
```diff
@@ -36,6 +36,7 @@ def package_files(directory):
 extra_deps['dev'] = [
     'junitparser>=2.1.1',
     'coverage[toml]>=6.1.1',
+    'fasteners>=0.16.3',  # run_directory_uploader tests require fasteners
```
Averylamp, line 39: I think it may make sense to split off all tests dependencies into its own category now that we have a couple of them

Averylamp (approved):
LGTM, just make sure it passes tests before merging.  Also concerned about performance hits/knowing what the performance hit is, so it'd be nice to know that

composer/callbacks/run_directory_uploader.py
```diff
@@ -0,0 +1,323 @@
...
+                             }) for _ in range(self._num_concurrent_uploads)
+        ]
+        for worker in self._workers:
+            worker.start()
+
+    def batch_end(self, state: State, logger: Logger) -> None:
+        if get_run_directory() is None:
+            return
+        if (state.batch_idx + 1) % self._upload_every_n_batches == 0:
+            self._trigger_upload(logger, LogLevel.BATCH)
```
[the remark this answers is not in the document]
ravi-mosaicml: Uploads happen `_upload_every_n_batches`, not every batch. The performance hit is the time it takes to copy the files that have changed since the last upload. I can try to measure it.

composer/callbacks/run_directory_uploader.py
ravi-mosaicml: Google style guide prefers `del` instead of `_`: https://google.github.io/styleguide/pyguide.html#214-decision

setup.py
ravi-mosaicml: We could split out `dev` into `tests` and `docs`? Would prefer to do in a separate PR...

ravi-mosaicml: See #129

composer/callbacks/run_directory_uploader.py
ravi-mosaicml: Good catch, thanks. I think this is what caused pyright and tests to fail.

composer/callbacks/run_directory_uploader.py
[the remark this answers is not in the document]
ravi-mosaicml: Performance implications:
* Negligible if there are very small files to upload (e.g. throughput drops from ~1280 to ~1260, but only sometimes)
* If you have larger artifacts to upload (e.g. 100 MB traces), then throughput drops from ~1280 to ~1040

Note that only new or modified files are uploaded (determined by using the last modified timestamp), and uploads only occur every `self._upload_every_n_batches`. This is equivalent to the wandb artifacts.

If we wanted performance over correctness, we can do non-blocking copies, but then we may be uploading corrupted data. This can be done in a separate PR (and controlled via a hparams flag) if desired. 

Raw data: https://wandb.ai/mosaic-ml/research2-run/runs/vv4atxxz. The drops at epoch 10, 20 and 30 come from the torch profiler saving the trace. The drops at steps 20, 30, and 40 come from the run directory uploader uploading traces. The other spikes every 10 steps are from the uploader uploading log files.

PR #97, opened 2021-11-20
Training Loop Profiler

ravi-mosaicml:
The Training Loop Profiler gives a breakdown of how long is spent on each part of the training loop, and how long is spent on each algorithm and callback for each event. It is implemented through the engine and wraps calls around each event. It also can profile system performance metrics through `psutil` and incorporates the pytorch profiler.

TODO:
- [x] Added in a `MosaicProfiler` class that supports markers for different segments of the code
- [x] Added in Markers in the engine around the different portions we wish to profile
- [x] Added in a JSON Trace Handler that logs trace data into Chrome JSON Trace format. See https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview. These traces can be viewed in chrome at chrome://tracing.
- [x] Added in HParams
- [x] Sample trace on resnet 50. See https://drive.google.com/file/d/1X19lPF9pjTDUrpcVQzXowD96vfUEy1DR/view?usp=sharing
- [x] @ravi-mosaicml Documentation
- [x] @anisehsani Helper script to merge traces from multiple GPUs, ~~start trace viewing web server~~. Automatically starting the trace viewer web server is more involved, as perfetto requires that you post the trace to their site. Will also need to think about automatically pulling traces from buckets.
- [x] @ravi-mosaicml Trace capture scheduling (need to add test cases)
- [x] @bandish-shah Simple Developer Instrumentation API Decorator option for profiling (need to add test cases)
- [ ] @bandish-shah Training loop observation events
- [x] @ravi-mosaicml clean closing (it's actually OK to have a dirty close -- it'll just record that some events did not finish, which is totally fine).
- [x] @ravi-mosaicml Trace uploading to blob stores (#101 does this!)
- [ ] @bandish-shah Test on real training runs 

## Examples of the visualization
image
image

PR #105, opened 2021-11-23
Remove `composer.trainer.ddp`; replace with `composer.utils.ddp`

ravi-mosaicml:
Before #65, composer.trainer.ddp ensured that DDP functionality was accessed only after ddp was initialized. Now, DDP is available from process start, so this class is no longer needed. Moved all the functionality from this class to the global composer.utils.ddp.

This change allows callbacks, algorithms, etc... to use DDP (such as barriers and reductions) as needed. #97 and #101 depend on this functionality.

Also removed DDP from the state, as that is available globally.


composer/utils/ddp.py
```diff
@@ -31,25 +78,182 @@ def _get_distributed_config_var(env_var: str,
...
+        int: The world size
+    """
     return _get_distributed_config_var(env_var="WORLD_SIZE",
                                        human_name="world size",
                                        default=1,
                                        fetch_fn_name="get_world_size")
 
 
 def get_global_rank() -> int:
+    """Returns the global rank of the current process, which is on `[0, WORLD_SIZE - 1]`
```
jbloxham, line 93:
```suggestion
    """Returns the global rank of the current process, which is in `[0, WORLD_SIZE - 1]`
```

composer/utils/ddp.py
```diff
@@ -31,25 +78,182 @@ def _get_distributed_config_var(env_var: str,
...
+    """Returns the local world size, which is the number of processes for the current node.
+
+    Returns:
+        int: The local world size
+    """
     return _get_distributed_config_var(env_var="LOCAL_WORLD_SIZE", human_name="local world size", default=1)
 
 
 def get_local_rank() -> int:
+    """Returns the local rank for the current process, which is on `[0, LOCAL_WORLD_SIZE - 1]`
```
jbloxham, line 111:
```suggestion
    """Returns the local rank for the current process, which is in `[0, LOCAL_WORLD_SIZE - 1]`
```

composer/utils/ddp.py
```diff
@@ -31,25 +78,182 @@ def _get_distributed_config_var(env_var: str,
...
+
+def all_reduce(
+    tensor: torch.Tensor,
+    reduce_operation: str = "SUM",
+) -> None:
+    if dist.is_available():
+        reduce_op = getattr(dist.ReduceOp, reduce_operation.upper())
+        dist.all_reduce(tensor, op=reduce_op)
+    else:
+        raise NotImplementedError("Non-DDP versions of reduce operations are not yet implemented")
```
jbloxham, line 135: This should be correct, right? Since BYOT is now able to hit this function, it's a little more important for us to support non-distributed training runs.

```suggestion
        return
```

jbloxham (approved):
I like this change a lot. The only major suggestion I have is that the file shouldn't be called `composer.utils.ddp`; it should be called `composer.utils.dist`. The only DDP specific code in that file (which could maybe be split out) is the `DDPSyncStrategy` stuff, and the `prepare_module` and `create_dataloader` stuff, which themselves are highly specific to our trainer and could probably live in the `trainer` package still, if you agree.

composer/utils/ddp.py
ravi-mosaicml: Ah yes, correct, assuming that the world size is 1, then we can just return. I updated the guards in the DDP functions to reflect this.

ravi-mosaicml:
> I like this change a lot. The only major suggestion I have is that the file shouldn't be called `composer.utils.ddp`; it should be called `composer.utils.dist`. The only DDP specific code in that file (which could maybe be split out) is the `DDPSyncStrategy` stuff, and the `prepare_module` and `create_dataloader` stuff, which themselves are highly specific to our trainer and could probably live in the `trainer` package still, if you agree.

God point, Would you be able to refactor it as needed in your deepspeed PR? I agree it would make sense to have an abstract API of sorts that then DDP or Deepspeed could implement. However, without knowing what deepspeed's abstractions are, I'm not sure how best to split it out.

jbloxham:
> > I like this change a lot. The only major suggestion I have is that the file shouldn't be called `composer.utils.ddp`; it should be called `composer.utils.dist`. The only DDP specific code in that file (which could maybe be split out) is the `DDPSyncStrategy` stuff, and the `prepare_module` and `create_dataloader` stuff, which themselves are highly specific to our trainer and could probably live in the `trainer` package still, if you agree.
> 
> God point, Would you be able to refactor it as needed in your deepspeed PR? I agree it would make sense to have an abstract API of sorts that then DDP or Deepspeed could implement. However, without knowing what deepspeed's abstractions are, I'm not sure how best to split it out.

Yes, I'm happy to take point on this.

PR #107, opened 2021-11-29
DeepLabv3 + ADE20k benchmark

Landanjs:
Pull request to add the DeepLabv3 model and the ADE20k dataset as a new benchmark for semantic segmentation tasks

Some results with 4 seeds for each experiment:

| Experiment | Final | Best |
| ----------- | ----- | ----- |
| Train for 127 epochs | 44.888 +/- 0.304 | 44.959 +/- 0.307 |
| Train for 64 epochs | 45.051 +/ - 0.328 | 45.323 +/- 0.284 |
| Use batch size 32 instead of 16 | 44.904 +/- 0.181 | 45.089 +/- 0.055 |
| Decoupled SGD (WD = 5e-6 | 45.54 +/- 0.364 | 45.669 +/- 0.284 |

[Here](https://wandb.ai/mosaic-ml/landan-deeplabv3/runs/rqdcd531?workspace=user-landanjs) is a convergence run.

mmsegmentation reports 44.08 mIoU and 45.00 mIoU for 64 and 127 epochs, respectively ([here](https://github.com/open-mmlab/mmsegmentation/tree/master/configs/deeplabv3#ade20k)). I think these numbers are the final results for a single run.

## Before Merging

- [x] Test for mean IoU
- [x] Test for each segmentation transformation


Landanjs:
@hanlint @coryMosaicML @A-Jacobson I want to make one more pass through the code today, but should be ready for review tomorrow!

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,336 @@
...
+        self.ignore_background = ignore_background
+
+    def __call__(self, batch: Batch):
+        xs, ys = batch
+        assert isinstance(xs, Tensor)
+        assert isinstance(ys, Tensor)
+        device = xs.device
+
+        if self.mean is None:
+            self.mean = torch.tensor([0.485 * 255, 0.456 * 255, 0.406 * 255], device=device)
```
ravi-mosaicml, line 38: Can you add a citation to where these magic numbers come from?

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,336 @@
...
+            self.std = self.std.view(1, 3, 1, 1)
+
+        xs = xs.float()
+        xs = xs.sub_(self.mean).div_(self.std)
+        if self.ignore_background:
+            ys = ys.sub_(1)
+        return xs, ys
+
+
+def fast_collate(batch: List[Tuple[Image.Image, Tensor]], memory_format: torch.memory_format = torch.contiguous_format):
```
ravi-mosaicml, line 51: This code looks _really_ similar to what we have for imagenet. Would it be possible to refactor it into a shared helper class and then that for here and for imagenet? (Same thing for the TransformationFn above)

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+
+    def __init__(self, min_scale: float, max_scale: float, base_size: Optional[Tuple[int, int]] = None):
+        super().__init__()
+        self.min_scale = min_scale
+        self.max_scale = max_scale
+        self.base_size = base_size
+
+    def forward(self, sample: Tuple[Image.Image, Image.Image]):
+        image, target = sample
```
ravi-mosaicml, line 93: I'm a bit confused on how a PIL `Image.Image` would be passed through on the forward pass...wouldn't the inputs be tensors?

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,336 @@
...
+        return len(self.image_files)
+
+
+@dataclass
+class ADE20kDatasetHparams(DatasetHparams):
+    """ Defines an instance of the ADE20k dataset for semantic segmentation.
+
+    Args:
+        split (str): the dataset split to use, either train, val, or test.
+        base_size (int): initial size of the image and target before other augmentations.
```
ravi-mosaicml, line 292: Can you include the defaults in the docstrings?

composer/datasets/ade20k.py
ravi-mosaicml: Ah, I see below that the dataset is calling these functions. Thoughts on having it not inherit from torch.nn.Module and instead rename `forward` to `__call__`?

ravi-mosaicml (commented):
Overall LGTM!

1. I don't think Dataset transformation functions should inherit from torch.nn.Module, since they aren't really NN building blocks (and they run on the CPU before the data is sent to the actual model). Instead, would it work to not have them inherit from torch.nn.Module and rename the `forward` function to `__call__`?
2. For the docstrings, they overall look great! Can you add the default values to them, so users will know what's optional?
3. Can you add synthetic dataset support? I'd take a look of how it's done for imagenet. That will allow the tests to check the model.

composer/datasets/ade20k.py
Landanjs: The convention in torchvision is to inherent from torch.nn.Module and use the forward call. I think this is to allow the transforms to be torch scripted. Should we deviate form this standard?

composer/datasets/ade20k.py
Landanjs: Yeah that would make the most sense, I just tried to avoid it 😅 What do you mean by shared helper class? Would it be sufficient to generalize the function to be able to handle both 1D and 2D targets? Would this be placed in `composer/utils/data.py`?

composer/datasets/ade20k.py
Landanjs: Defaults added!

composer/datasets/ade20k.py
Landanjs: I generalized the ADE20k and ImageNet versions and moved them to `utils/data.py`. The first few epochs of of both datasets seem to align with previous runs. I'm not convinced my code was the best, so feel free to provide critiques.

composer/datasets/ade20k.py
ravi-mosaicml: Ah makes sense, didn't realize that's how torchvision did it. Let's keep the standard.

composer/utils/data.py
```diff
@@ -1,11 +1,81 @@
...
+import numpy as np
+import torch
 import torch.utils.data
+from PIL import Image
 from torchvision import transforms
 
-from composer.core.types import Dataset
+from composer.core.types import Dataset, Tensor, Batch
+
+class TransformationFn:
```
ravi-mosaicml, line 14:
```suggestion
class NormalizationTransformationFn:
```

composer/utils/data.py
```diff
@@ -1,11 +1,81 @@
...
+            self.std = self.std.view(1, 3, 1, 1)
+
+        xs = xs.float()
+        xs = xs.sub_(self.mean).div_(self.std)
+        if self.ignore_background:
+            ys = ys.sub_(1)
+        return xs, ys
+
+
+def fast_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]],
```
ravi-mosaicml, line 48:
```suggestion
def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]],
```

composer/utils/data.py
```diff
@@ -1,11 +1,81 @@
...
+                 memory_format: torch.memory_format = torch.contiguous_format):
+    """Constructs a batch for training from individual samples.
```
ravi-mosaicml, line 50: Can you add the args and return to the docstring?

composer/datasets/ade20k.py
ravi-mosaicml: Yup looks good. Thanks!

composer/utils/data.py
Landanjs: What about `NormalizationFn`?

composer/utils/data.py
Landanjs: The fast_collate is a reference to Apex code [here](https://github.com/NVIDIA/apex/blob/master/examples/imagenet/main_amp.py#L28) (I can add a citation). It also aggregates the class labels for ImageNet. Does the name change still make sense?

composer/utils/data.py
ravi-mosaicml: Yup, that sounds good.

composer/utils/data.py
ravi-mosaicml: I would, since the function name and module path would otherwise be too vague....if this function was part of say `composer.utils.data.images` then I'd say it'd be fine to leave it as is, but it's a bit unclear with `composer.utils.data` since it doesn't apply to NLP (I think?)

composer/datasets/synthetic.py
```diff
@@ -100,9 +100,11 @@ def __getitem__(self, idx: int):
                 input_target[:, 0] = 1.0
             elif self.label_type == SyntheticDataLabelType.CLASSIFICATION_INT:
                 assert self.num_classes is not None
-                input_target = torch.randint(0,
-                                             self.num_classes, (self.num_unique_samples_to_create,),
-                                             device=self.device)
+                if self.label_shape:
```
Landanjs, line 103: I added this for synthetic segmentation datasets (i.e. image and labels are the same dimension). Is this okay? Also, is RANDOM_INT setting below still relevant?

composer/cli/launcher.py
```diff
@@ -168,7 +168,7 @@ def monitor_processes(processes: Set[subprocess.Popen]):
                         output=output,
                         stderr=stderr,
                     )
-                    error_msg = [f"Process {process.pid} excited with code {process.returncode}"]
+                    error_msg = [f"Process {process.pid} exited with code {process.returncode}"]
```
ravi-mosaicml, line 171: whoops, thanks

composer/datasets/synthetic.py
ravi-mosaicml: Sure, cool with this so long as it doesn't break the existing tests. Hmm yeah, nothing really uses random_int. I guess for consistency, could you also copy this change down there? Or remove the `RANDOM_INT` option for synthetic data labels?

ravi-mosaicml (approved):
LGTM. Thanks!!

Landanjs:
I'm not able to pass the CPU test due to a timeout error during `test_hparams.py`. I think the timeout occurs because DeepLabV3 by default downloads pre-trained weights which take about 30s to download. Is there a way to avoid this timeout? I don't think we would want to change the defaults for the sake of the CPU test.

ravi-mosaicml:
Would you be able to use monkeypatch to mock out the actual download during the tests (and instead have it return uninitialized weights)?

composer/datasets/synthetic.py
Landanjs: `RANDOM_INT` seemed redundant, so I removed it

Landanjs:
The pretrained weight download is done within torchvision, so I thought the easiest way to avoid the long download was to manually change `is_backbone_pretrained` in `test_hparams.py` and `test_model_registry.py` when DeepLabV3 is used.

hanlint:
cc: @A-Jacobson or @coryMosaicML for a review here (vision team)

A-Jacobson (approved):

examples/run_mosaic_trainer.py
```diff
@@ -16,9 +16,12 @@
 import warnings
 from typing import Type
 
+import composer
 from composer.trainer.trainer import Trainer
 from composer.trainer.trainer_hparams import TrainerHparams
 
+print(composer.__file__)
```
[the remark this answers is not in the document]
Landanjs: Nice catch 😂  Removed!

composer/yamls/models/deeplabv3_ade20k.yaml
```diff
@@ -0,0 +1,52 @@
...
+    drop_last: false
+    shuffle: false
+optimizer:
+  decoupled_sgdw:
+    lr: 0.01
+    momentum: 0.9
+    weight_decay: 5.0e-5
+    dampening: 0
+    nesterov: false
+schedulers:
```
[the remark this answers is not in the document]
Landanjs: I was under the impression warmup is only helpful when large batches are used. Does warmup help across most scenarios?

composer/models/loss.py
```diff
@@ -14,6 +14,52 @@
     from composer.core.types import Tensor
 
 
+class mIoU(Metric):
```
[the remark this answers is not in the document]
Landanjs: Not sure why I did that, I'll change it to `MIoU`

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,300 @@
...
+        image_file = self.image_files[index]
+        image_path = os.path.join(self.image_dir, image_file)
+        image = Image.open(image_path)
+
+        # Load annotation target if using either train or val splits
+        if self.split in ['train', 'val']:
+            target_path = os.path.join(self.datadir, 'annotations', self.split, image_file.split('.')[0] + '.png')
+            target = Image.open(target_path)
+
+            if self.both_transforms:
```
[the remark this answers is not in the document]
Landanjs: It isn't necessary, but it felt weird for the photometric distortions to take a target argument since the target would likely never be used during the transformation. Also, the `PadToSize` function would need to take a separate `fill` arguments for the image and target which feels a bit weird. The changes would decrease downstream code for this specific use case, but would add unnecessary attributes to the transformations.

How strongly do you feel about simplifying the transformation composition? Happy to change the transformations if you think it makes more sense.

composer/models/deeplabv3/deeplabv3.py
```diff
@@ -0,0 +1,76 @@
...
+            self.model.backbone.load_state_dict(backbone.state_dict())
+        if self.hparams.sync_bn:
+            policy = {torch.nn.BatchNorm2d: bn_to_syncbn}
+            replace_module_classes(self.model, policy)
+
+    # Should the forward pass take a batch pair? We shouldn't expect the forward pass to have labels
+    def forward(self, batch: BatchPair):
+        x, _ = batch
+
+        context = contextlib.nullcontext if self.training else torch.no_grad  # should this be added?
```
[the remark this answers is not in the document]
Landanjs: This is something I took from the Unet code... I think that is the concern, but it should be up to the user to set the appropriate context, right?

composer/datasets/ade20k.py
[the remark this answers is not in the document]
A-Jacobson: I do prefer it to changing the dataset interface but I can see the benefits on being more explicit about exactly which transformation is applied. I think this is ok for now but let's think about it more as we add more datasets that need to transform both the input and targets.

composer/yamls/models/deeplabv3_ade20k.yaml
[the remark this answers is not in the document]
A-Jacobson: From https://arxiv.org/pdf/1803.09820.pdf it's not exactly the same schedule but a lr warmup can let you train with a higher overall learning rate. Maybe I should have said I'd be curious about bumping up the learning rate + adding warmup.

composer/models/deeplabv3/deeplabv3.py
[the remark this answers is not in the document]
A-Jacobson: I'd say to remove it but I think you already did!

composer/yamls/models/deeplabv3_ade20k.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - cosine_decay:
+      T_max: "127ep"
+      eta_min: 0
+      verbose: false
+      interval: step
+model:
+  deeplabv3:
+    initializers:
```
A-Jacobson, line 32: Are these initializers being used?

Landanjs: Ah, no... The original PyTorch DeepLabV3 did not initialize the head, so I was hesitant. I the initialization is being used now!

Landanjs:
After some thought and discussion, I think it makes the most sense to make the initial benchmark to be as close as possible to other benchmarks. This means reverting back to the original batch size, no decoupled weight decay, and using a Polynomial LR schedule instead of cosine decay. 

@ravi-mosaicml and/or @A-Jacobson could y'all skim through these recent changes today or tomorrow?

- Added code to use initializers in `deeplabv3.py`

- Added `PolynomialLR` and PolynomialLRHparams` to `schedulers.py` and in the required scheduler tests

- Added `test_loss.py` for mIoU tests (subsequently made a `tests/models` directory and added `test_efficientnet.py`)

- Added `test_segmentation_transforms.py` for tests on segmentation transforms. I could not quickly come up with good ways to test these especially since some of them are random, but I'm not sure how rigorous these tests should be. Be gentle!

It would be amazing if I could get this in tomorrow, thank you!!

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+
+    def __init__(self, crop_size: Tuple[int, int]):
+        super().__init__()
+        self.crop_size = crop_size
+
+    def forward(self, sample: Tuple[Image.Image, Image.Image]):
+        image, target = sample
+        if image.height > self.crop_size[0] or image.width > self.crop_size[1]:
+            i, j, h, w = transforms.RandomCrop.get_params(image, output_size=self.crop_size)  # type: ignore
```
ravi-mosaicml, line 62: can you add comments why these type ignores are required? The error message from pyright will suffice.

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+
+    def __init__(self, probability: float = 0.5):
+        super().__init__()
+        self.probability = probability
+
+    def forward(self, sample: Tuple[Image.Image, Image.Image]):
+        image, target = sample
+        if np.random.random_sample() > self.probability:
+            image = TF.hflip(image)  # type: ignore
```
ravi-mosaicml, line 82: Same thing here re: type ignore comments

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def __init__(self, size: Tuple[int, int], fill: Union[int, Tuple[int, int, int]] = 0):
+        super().__init__()
+        self.size = size
+        self.fill = fill
+
+    def forward(self, image: Image.Image):
+        padding = max(self.size[0] - image.height, 0), max(self.size[1] - image.width, 0)
+        padding = (padding[1] // 2, padding[0] // 2, ceil(padding[1] / 2), ceil(padding[0] / 2))
+        image = TF.pad(image, padding, fill=self.fill)  # type: ignore
```
ravi-mosaicml, line 103: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+        super().__init__()
+        self.brightness = brightness
+        self.contrast = contrast
+        self.saturation = saturation
+        self.hue = hue
+
+    def forward(self, image: Image.Image):
+        if np.random.randint(2):
+            brightness_factor = np.random.uniform(1 - self.brightness, 1 + self.brightness)
+            image = TF.adjust_brightness(image, brightness_factor)  # type: ignore
```
ravi-mosaicml, line 130: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+
+        contrast_mode = np.random.randint(2)
+        if contrast_mode == 1 and np.random.randint(2):
+            contrast_factor = np.random.uniform(1 - self.contrast, 1 + self.contrast)
+            image = TF.adjust_contrast(image, contrast_factor)  # type: ignore
```
ravi-mosaicml, line 135: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if np.random.randint(2):
+            saturation_factor = np.random.uniform(1 - self.saturation, 1 + self.saturation)
+            image = TF.adjust_saturation(image, saturation_factor)  # type: ignore
```
ravi-mosaicml, line 139: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if np.random.randint(2):
+            hue_factor = np.random.uniform(-self.hue, self.hue)
+            image = TF.adjust_hue(image, hue_factor)  # type: ignore
```
ravi-mosaicml, line 143: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if contrast_mode == 0 and np.random.randint(2):
+            contrast_factor = np.random.uniform(1 - self.contrast, 1 + self.contrast)
+            image = TF.adjust_contrast(image, contrast_factor)  # type: ignore
```
ravi-mosaicml, line 147: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+                image, target = self.both_transforms((image, target))
+
+            if self.target_transforms:
+                target = self.target_transforms(target)
+
+        if self.image_transforms:
+            image = self.image_transforms(image)
+
+        if self.split in ['train', 'val']:
+            return image, target  # type: ignore
```
ravi-mosaicml, line 225: ditto

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+                both_transforms = torch.nn.Sequential(
+                    RandomResizePair(min_scale=self.min_resize_scale,
+                                     max_scale=self.max_resize_scale,
+                                     base_size=(self.base_size, self.base_size)),
+                    RandomCropPair(crop_size=(self.final_size, self.final_size)),
+                    RandomHFlipPair(),
+                )
+                image_transforms = torch.nn.Sequential(
+                    PhotometricDistoration(brightness=32. / 255, contrast=0.5, saturation=0.5, hue=18. / 255),
+                    PadToSize(size=(self.final_size, self.final_size),
```
ravi-mosaicml, line 306: can you add a comment to where these magic values came from?

composer/datasets/ade20k.py
```diff
@@ -0,0 +1,331 @@
...
+                both_transforms = None
+                image_transforms = transforms.Resize(size=(self.final_size, self.final_size),
+                                                     interpolation=TF.InterpolationMode.BILINEAR)
+                target_transforms = transforms.Resize(size=(self.final_size, self.final_size),
+                                                      interpolation=TF.InterpolationMode.NEAREST)
+            collate_fn = pil_image_collate
+            device_transform_fn = NormalizationFn(self.ignore_background)
+
+            dataset = ADE20k(
+                datadir=self.datadir,  # type: ignore
```
ravi-mosaicml, line 320:
```suggestion
                datadir=self.datadir,
```

ravi-mosaicml (approved):
Overall LGTM. Two main things:

1. I don't think many of the type ignores are required. I tried to comment how to remove them. If they are required, please add the pyright error message as a comment next to the type ignore. I realize they can get annoying; happy to help debug these!

2. Can you move the imagenet normalization parameters out of the normalization_fn class and to the imagenet dataset file? The normalization function should be dataset-generic.

composer/models/deeplabv3/deeplabv3.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    # Instantiate backbone module
+    if backbone_arch == 'resnet50':
+        resnet.model_urls[backbone_arch] = "https://download.pytorch.org/models/resnet50-f46c3f97.pth"
+    elif backbone_arch == 'resnet101':
+        resnet.model_urls[backbone_arch] = "https://download.pytorch.org/models/resnet101-cd907fc2.pth"
+    else:
+        raise ValueError(f"backbone_arch must be one of ['resnet50', 'resnet101'] not {backbone_arch}")
+    backbone = resnet.__dict__[backbone_arch](pretrained=is_backbone_pretrained,
```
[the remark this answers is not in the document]
Landanjs: Ah, I took this line from pytorch. The link you posted was for `regnet.py`, `resnet.py` is here: https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py

composer/models/deeplabv3/deeplabv3.py
```diff
@@ -0,0 +1,125 @@
...
+                 is_backbone_pretrained: bool = True,
+                 sync_bn: bool = True,
+                 initializers: Optional[List[str]] = None):
+
+        super().__init__()
+        self.num_classes = num_classes
+        self.model = deeplabv3_builder(
+            backbone_arch=backbone_arch,
+            is_backbone_pretrained=is_backbone_pretrained,
+            num_classes=num_classes,  # type: ignore
```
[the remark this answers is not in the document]
Landanjs: Oops, nope!

composer/models/deeplabv3/deeplabv3.py
```diff
@@ -0,0 +1,125 @@
...
+    def forward(self, batch: Batch):
+        x = batch[0]  # type: ignore
+        logits = self.model(x)['out']
+        return logits
+
+    def loss(self, outputs: Any, batch: Batch):
+        """Calculate the specified loss for training.
+        """
+        target = batch[1]  # type: ignore
+        loss = soft_cross_entropy(outputs, target, ignore_index=-1)  # type: ignore
```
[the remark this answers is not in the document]
Landanjs: `Argument of type "Tensors" cannot be assigned to parameter "target" of type "Tensor" in function "soft_cross_entropy"`

composer/models/deeplabv3/deeplabv3_hparams.py
```diff
@@ -0,0 +1,39 @@
...
+        if self.num_classes is None:
+            raise ValueError("num_classes must be specified")
+
+        if self.backbone_arch not in ['resnet50', 'resnet101']:
+            raise ValueError(f"backbone_arch must be one of ['resnet50', 'resnet101']: not {self.backbone_arch}")
+
+    def initialize_object(self):
+        from composer.models.deeplabv3.deeplabv3 import MosaicDeepLabV3
+        return MosaicDeepLabV3(
+            num_classes=self.num_classes,  # type: ignore
```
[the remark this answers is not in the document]
Landanjs: `Argument of type "int | None" cannot be assigned to parameter "num_classes" of type "int" in function "__init__"`

I don't think it makes sense to make `num_classes` optional in `MosaicDeepLabV3`, would be okay to keep the `type: ignore`? Which part of the error should I include?

composer/models/deeplabv3/deeplabv3_hparams.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def validate(self):
+        if self.num_classes is None:
+            raise ValueError("num_classes must be specified")
+
+        if self.backbone_arch not in ['resnet50', 'resnet101']:
+            raise ValueError(f"backbone_arch must be one of ['resnet50', 'resnet101']: not {self.backbone_arch}")
+
+    def initialize_object(self):
+        from composer.models.deeplabv3.deeplabv3 import MosaicDeepLabV3
```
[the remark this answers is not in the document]
Landanjs: Added!

composer/models/deeplabv3/deeplabv3_hparams.py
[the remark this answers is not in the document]
Landanjs: Ah, saw the comment below!

composer/models/loss.py
```diff
@@ -14,6 +14,52 @@
...
+        """Update the state with new predictions and targets.
+        """
+        preds = logits.argmax(dim=1)
+        for pred, target in zip(preds, targets):
+            mask = (target != self.ignore_index)
+            pred = pred[mask]
+            target = target[mask]
+
+            intersect = pred[pred == target]
+            area_intersect = torch.histc(intersect.float(), bins=self.num_classes, min=0, max=self.num_classes - 1)
```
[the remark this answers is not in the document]
Landanjs: TBH I'm not sure, but I think the metrics are usually on the same device as the model? Should I test this somehow? The metric works during training and the other metrics don't seem to mess with devices either

tests/test_hparams.py
```diff
@@ -45,4 +45,8 @@ def test_trainer_initialize(self, hparams_file: str):
         _configure_dataset_for_synthetic(hparams.val_dataset)
         hparams.device = CPUDeviceHparams()
 
+        # Manually set is_backbone_pretrained to false to prevent downloading pretrained weights during test
+        if hasattr(hparams, "model") and hasattr(hparams.model, "is_backbone_pretrained"):
```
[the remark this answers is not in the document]
Landanjs: Ah, I was trying to futureproof for new models that also use pretrained weights, but I can make the change and we cross that bridge later :)

composer/datasets/ade20k.py
Landanjs: Yeah these are annoying... Pytorch transforms accept both PIL images and torch Tensors, but the typing only specifies Tensor. Is there a fix here?

composer/datasets/ade20k.py
Landanjs: Added a check before to avoid type ignore

composer/datasets/ade20k.py
Landanjs: I added the message `transform typing does not include PIL.Image` does this work?

composer/datasets/ade20k.py
Landanjs: Added!

composer/models/loss.py
```diff
@@ -14,6 +14,52 @@
...
+            area_prediction = torch.histc(pred.float(), bins=self.num_classes, min=0, max=self.num_classes - 1)
+            area_target = torch.histc(target.float(), bins=self.num_classes, min=0, max=self.num_classes - 1)
+
+            self.total_intersect += area_intersect
+            self.total_union += area_prediction + area_target - area_intersect
+
+    def compute(self):
+        """Aggregate state across all processes and compute final metric.
+        """
+        return 100 * (self.total_intersect / self.total_union).mean()  # type: ignore
```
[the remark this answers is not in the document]
Landanjs: Ah, this error is `Operator "/" not supported for types "Tensor | Module" and "Tensor | Module"`

composer/utils/data.py
```diff
@@ -1,11 +1,93 @@
...
+
+    def __call__(self, batch: Batch):
+        xs, ys = batch
+        assert isinstance(xs, Tensor)
+        assert isinstance(ys, Tensor)
+        device = xs.device
+
+        # ImageNet normalization values from torchvision: https://pytorch.org/vision/stable/models.html
+        if self.mean is None:
+            self.mean = torch.tensor([0.485 * 255, 0.456 * 255, 0.406 * 255], device=device)
```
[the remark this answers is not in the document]
Landanjs: I moved the mean and std to `IMAGENET_CHANNEL_MEAN` and `IMAGENET_CHANNEL_STD` at the top of `imagenet.py`. I converted the `NormalizationFn` to take `mean` and `std` as inputs, then made them into tensors with the appropriate device in `__call__`. Let me know if I missed anything!

Landanjs:
Okay, I think addressed all of @ravi-mosaicml  comments, but let me know if more is needed!

Also, I'm getting this failure for the cpu test now: 
```
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    torch>=1.9 from https://files.pythonhosted.org/packages/2c/c8/dcef19018d2fe730ecacf47650d3d6e8d6fe545f02fbdbde0174e0279f02/torch-1.10.1-cp39-cp39-manylinux1_x86_64.whl#sha256=cca660b27a90dbbc0af06c859260f6b875aef37c0897bd353e5deed085d2c877 (from mosaicml==0.3.1):
        Expected sha256 cca660b27a90dbbc0af06c859260f6b875aef37c0897bd353e5deed085d2c877
             Got        59289571109939f561c54f7e13ddb31d041c5a9dbd0397414446e4b31440b6b9
```

How do I resolve this?

A-Jacobson (approved):
I'm good with merging this as is. Though I would like to better formalize what constitutes a "baseline" as you've reached/surpassed mmseg's reported IoU with about 3 different recipes.

PR #109, opened 2021-11-29
Deepspeed Integration

jbloxham:
Support for using DeepSpeed instead of Pytorch DDP in the trainer. Somewhat WIP, but good enough to merge for now. The major TODOs are:

- Support for regular checkpointing
- Cleaner support for precision casting
- Config support for DeepSpeed goodies like ZeRO and activation checkpointing
- Quite a bit of testing

I'm waiting on a node to run a quick regression test.


composer/trainer/trainer.py
```diff
@@ -224,6 +235,10 @@ def __init__(
         )
         self.eval_dl_spec = eval_dataloader_spec
 
+        # TODO: handling precision correctly with DeepSpeed is an open problem
```
Averylamp, line 238: I think it'd be nice to leave these as issues as well in the code so that they can be tracked in multiple places.  
```
# TODO(issue #xyz): <explaination>
```
is nice to also be able to reference issues and make sure that comments are removed as issues are solved

composer/trainer/trainer.py
```diff
@@ -274,13 +289,15 @@ def __init__(
...
-        if checkpoint_folder and checkpoint_interval and checkpoint_interval_unit:
+        # TODO: get checkpointing working with DeepSpeed.
+        if not self.deepspeed_enabled and checkpoint_folder and checkpoint_interval and checkpoint_interval_unit:
             self.checkpointer = Checkpointer(checkpoint_folder=get_relative_to_run_directory(checkpoint_folder),
                                              checkpoint_interval=checkpoint_interval,
                                              checkpoint_interval_unit=checkpoint_interval_unit)
 
         self.checkpoint_loader = None
-        if checkpoint_filepath:
+        # TODO: get checkpointing working with DeepSpeed.
```
Averylamp, line 299: Same github issue tracking as above

Averylamp (approved):
LGTM, maybe wait on one more lgtm for the training loop changes

composer/core/precision.py
```diff
@@ -8,8 +8,11 @@ class Precision(StringEnum):
     
     Attributes:
         AMP: Use :mod:`torch.cuda.amp`. Only compatible with GPUs.
+        FP16: Use 16-bit floating-point precision. Currently only
```
ravi-mosaicml, line 11: is this BF16?

composer/trainer/deepspeed.py
```diff
@@ -0,0 +1,15 @@
...
+from dataclasses import dataclass
+
+import yahp as hp
+
+
+@dataclass
+class DeepSpeedHparams(hp.Hparams):
+    """Params for configuration of DeepSpeed."""
+
+    enabled: bool = hp.required("Whether to use DeepSpeed.")
```
ravi-mosaicml, line 12: Why do we need the enabled flag? If the params are specified in the trainer, then wouldn't it be enabled?

composer/trainer/trainer.py
```diff
@@ -174,8 +180,10 @@ def __init__(
 
         self.config = config
 
+        self.deepspeed_enabled = deepspeed_hparams and deepspeed_hparams.enabled
```
ravi-mosaicml, line 183:
```suggestion
        self.deepspeed_enabled = deepspeed_hparams is not None
```

composer/trainer/deepspeed.py
ravi-mosaicml: Or, as an alternative, maybe just add a `use_deepspeed` flag in the trainer hparams if there are hparams?

composer/trainer/deepspeed.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def validate(self):
```
ravi-mosaicml, line 14: Not needed if it jsut calls super

composer/trainer/trainer.py
ravi-mosaicml: Can you reference a link to a stack overflow / GH issue to what is meant here (I'm a bit confused why it's an open problem)

composer/trainer/trainer.py
```diff
@@ -224,6 +235,10 @@ def __init__(
         )
         self.eval_dl_spec = eval_dataloader_spec
 
+        # TODO: handling precision correctly with DeepSpeed is an open problem
+        precision_context = self.device.precision_context if not self.deepspeed_enabled else lambda _: contextlib.nullcontext(
```
ravi-mosaicml, line 239:
```suggestion
        precision_context = self.device.precision_context if not self.deepspeed_enabled else contextlib.nullcontext
```

composer/trainer/trainer.py
```diff
@@ -274,13 +289,15 @@ def __init__(
         self.state.schedulers = ComposedScheduler(schedulers=schedulers)
 
         self.checkpointer = None
-        if checkpoint_folder and checkpoint_interval and checkpoint_interval_unit:
+        # TODO: get checkpointing working with DeepSpeed.
```
ravi-mosaicml, line 292: can you raise a `NotImplementedError` here if someone tries to use checkpointing and deepspeed?

composer/trainer/trainer.py
```diff
@@ -9,6 +9,7 @@
 import warnings
 from typing import Any, Dict, List, Optional, Sequence, Union
 
+import deepspeed
```
ravi-mosaicml, line 12: Should this be a conditional import? Or is deepspeed core?

composer/trainer/trainer.py
```diff
@@ -705,15 +733,26 @@ def _train_batch_inner(self, microbatches: Sequence[Batch]):
                 for loss in ensure_tuple(state.loss):
                     loss.backward(create_graph=self.backwards_create_graph)
 
+                for loss in ensure_tuple(state.loss):
```
ravi-mosaicml, line 736: Should this be before the loss is scaled? Or was that a bug all along?

tests/trainer/test_deepspeed.py
```diff
@@ -0,0 +1,108 @@
...
+from composer.utils import ddp
+from tests.fixtures.models import SimpleBatchPairModelHparams
+from tests.trainer.test_ddp import CheckBatch0Hparams, TrackedDatasetHparams, get_batch_file_path, get_file_path
+
+
+@pytest.mark.timeout(90)
+@pytest.mark.gpu
+@pytest.mark.parametrize("world_size", [pytest.param(1), pytest.param(2, marks=pytest.mark.world_size(2))])
+def test_deepspeed(world_size: int, mosaic_trainer_hparams: TrainerHparams, tmpdir: pathlib.Path) -> None:
+    """Pretty much just copied from ./test_ddp"""
```
ravi-mosaicml, line 24: #110  changes the DDP test fyi

ravi-mosaicml (commented):
I'll take another look once the TODOs are adressed, but looks awesome!

composer/core/precision.py
jbloxham: No. BF16 assigns different amounts of bits to the mantissa and exponent. DeepSpeed does have some BF16 support but it's very limited right now.

composer/trainer/deepspeed.py
jbloxham: When I add support for things like ZeRO, there's going to be a lot more DeepSpeed-relevant hparams, so I definitely at least want this object as a way of organizing those hparams.

composer/trainer/trainer.py
jbloxham: DeepSpeed handles loss scaling on its own, but doesn't do it in place. I've since updated the code to reflect the awkward reality that for non-DeepSpeed, we need to do this scaling before we call `loss.backward`, and for DeepSpeed, we need to scale after the backward pass.

PR #111, opened 2021-11-30
Moved device.prepare() to __init__

ravi-mosaicml:
Since dataloaders are created in `__init__`, the device must also be prepared in `__init__`

### DIFF
--- a/composer/trainer/devices/device.py
+++ b/composer/trainer/devices/device.py
@@ -18,14 +18,10 @@ class Device(Serializable, ABC):
     """
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


Landanjs (approved):
Can't comment on the actual code, but I can run this command now:
 `composer -n 8 examples/run_mosaic_trainer.py -f composer/yamls/models/resnet50.yaml --loggers wandb --loggers.wandb.projects landan-random --loggers.wandb.name resnet50_synthetic_test --datadir /localdisk/ImageNet/`
 
 Thanks!

PR #112, opened 2021-11-30
Replaced `atexit` with cleanup methods

ravi-mosaicml:
When running the trainer multiple times, such as in interactive enviroments, `atexit` does not fire. Instead, replaced it with `.close()` and `.post_close()` hooks on callbacks.

`.close()` can be used to write and flush files.
`.post_close()` can be used to backup the run directory and capture any changes that may have been made on `.close()`

### DIFF
--- a/composer/callbacks/torch_profiler.py
+++ b/composer/callbacks/torch_profiler.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import atexit
 import warnings
 from dataclasses import asdict, dataclass
 from typing import TYPE_CHECKING, Optional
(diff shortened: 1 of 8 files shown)


Averylamp (approved):
lgtm.  I think we still will need to check state to ensure that state can be modified, kept in memory, and have a new `.fit` be called

ravi-mosaicml:
Yes,

> lgtm. I think we still will need to check state to ensure that state can be modified, kept in memory, and have a new `.fit` be called

Agreed, this PR does not fully address multiple `.fit`. However, it does remove one of the blockers for it.

PR #113, opened 2021-11-30
Running callbacks befor algorithms for the INIT event in the engine

ravi-mosaicml:
* For the INIT event, run the callbacks first to initialize the loggers.
* For other events, run the algorithms first, so the callbacks have the state  after algorithms modify it.


anisehsani (approved):
Looks good!

PR #119, opened 2021-12-01
Add Stratify Batches algorithm

dblalock:
This algorithm imposes better class balance within each batch. At present, better can be defined as either "more closely matches the overall class balance" or "samples classes evenly, upsampling infrequent classes as needed".

This is motivated partially by word-of-mouth claims that this can yield a large accuracy lift on some datasets, and partially by a desire to use this to help balance assignments in mixture-of-experts models.

- [x] Core stratification/balance logic
- [x] Algorithm + standalone wrappers
- [x] Tests for core logic
- [x] Tests for wrappers
- [ ] Experiments (probably just a couple ImageNet networks)

The subtle thing we'll have to think about here is how to avoid surprises in the upsampling case. Everything should be fine if the training duration is specified in steps, but the step count will change if it's specified in epochs.

[ResNet50 results so far](https://wandb.ai/mosaic-ml/rn50)


hanlint:
Closing as stale

PR #120, opened 2021-12-01
Multiple Evaluator Datasets

anisehsani:
Allows a model to be evaluated with multiple datasets with metrics specifically relevant to that dataset. For example, the GLUE NLP tasks use different metrics for each dataset/task.
The basic data-structure introduced is the Evaluator which takes three parameters:
- label (str) name of the evaluator - used for logging metrics
- dataloader
- metrics (MetricCollection)

When a user doesn't want to use the evaluator objects to create a trainer, they can still use the val_dataset arguments, but the trainer will wrap the dataloader into an Evaluator object with validation metrics from the model. 

Before Evaluators are created, the Hparams system creates EvaluatorSpecs, which are passed to the Trainer, which then creates Evaluator objects with and stores them in the state. The EvaluatorSpecs are similar to the Evaluators, but contain DataloaderSpecs instead of data loaders.

Here is an example of how you would add evaluator fields to your YAML.

image

The metrics in the YAML should have the same names as the metrics registered in the EvaluatorHparams class. To add a custom metric, you would add it to the metric_registry dictionary in the EvaluatorHparams class. 

However, by using the eval_dataset argument, the trainer wraps the metrics associated with the model in an evaluator, so to use custom metrics, you could also just include them in the model and not in the metric registry. 

#### Metrics and How They are Populated

There are several ways you can add metrics to an evaluator currently. By using the metric_names field, directly including metrics, or by using the model defaults.
- If no metrics are specified in an evaluator, the trainer will add all the validation metrics specified in the model to the evaluator. 
- If Metrics are passed in to the Evaluator, the trainer just looks to see if the metrics are compatible with the model (in the model validation metrics). It gives a warning if it doesn't find the metric in the model validation metrics.
- If no Metrics are directly specified and metric_names IS specified in the Evaluator, the trainer will look through all the model validation metrics and check if there are any metrics with names specified in metric_names. It then adds these metrics to the Evaluator. (Gives a warning if it finds entries in metric_names that do not correspond to metrics in the model validation metrics)

##### Example:

You want to evaluate metric1 on dataset1 and metric2 on dataset2.
You can
- not specify the metrics in your model and pass in instances of metric1 and metric2 when you create the evaluators. 
- list the metrics in your model metrics and pass in the names "metric1" and "metric2" in the metric_names field when you create the evaluators.


moinnadeem:
Should we resolve PR Gate issues before doing reviews, in case the code changes?

composer/core/state.py
```diff
@@ -54,6 +54,9 @@
     "precision",
     "train_dataloader",
     "eval_dataloader",
```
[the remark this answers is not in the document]
moinnadeem: +1 here

composer/datasets/dataset_registry.py
```diff
@@ -0,0 +1,21 @@
...
+    "imagenet": ImagenetDatasetHparams,
+    "cifar10": CIFAR10DatasetHparams,
+    "synthetic": SyntheticDatasetHparams,
+    "mnist": MNISTDatasetHparams,
+    "lm": LMDatasetHparams,
+}
+
+
+def get_dataset_registry():
+    return registry
```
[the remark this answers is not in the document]
moinnadeem: Why do we need to add a dataset registry? Isn't this already in `trainer_hparams.py`?

Can we import it from there?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+
+@dataclass
+class EvaluatorHparams(hp.Hparams):
+    hparams_registry = {  # type: ignore
+        "eval_dataset": get_dataset_registry(),
+    }
+
+    metric_registry = {
+        "Accuracy": Accuracy,
+        "MatthewsCorrcoef": MatthewsCorrcoef,
```
moinnadeem, line 34: Can we remove "coef" from these and just do something like "MatthewsCorr"? It's less verbose

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+        "F1": F1,
+        "PearsonCorrcoef": PearsonCorrcoef,
+        "SpearmanCorrcoef": SpearmanCorrcoef,
+        "CrossEntropyLoss": CrossEntropyLoss,
+    }
+
+    label: str = hp.required(doc="label")
```
moinnadeem, line 41: Can we update the label with a more descriptive docstring? Something like 'Label, to be used in logging output' or related to the functionality

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+    eval_dataset: DatasetHparams = hp.required(doc="Training dataset hparams")
+    metrics: List[str] = hp.required(doc="metrics", template_default=list)
```
moinnadeem, line 43: Can we update the docs to describe if the metrics should be a valid instance of TorchMetrics?

moinnadeem: We should also add a validate() function that checks if the metrics are valid options in the registry ,and a helpful error message if not. Something like "The metric `KL-Divergence` was not found. Valid options are OPTION_1, OPTION_2, OPTION_3, etc"

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+        dataset = self.eval_dataset.initialize_object()
+
+        def resolve_metric_list(metric_names: List[str]) -> List[Metric]:
+            resolved_metrics = []
+            for metric_name in metric_names:
+                if metric_name in self.metric_registry:
+                    metric_instance = self.metric_registry[metric_name]()
+                    resolved_metrics.append(metric_instance)
+                else:
+                    log.warning(f"The metric {metric_name} is not a registered metric. Add it to the metric registry.")
```
moinnadeem, line 55: Should we raise an exception instead of a warning if a metric is not valid? I think I would want my training to crash before I check it the next day, but curious for other people's thoughts as well.

composer/loggers/tqdm_logger.py
```diff
@@ -117,7 +117,8 @@ def init(self, state: State, logger: Logger) -> None:
 
     def _start(self, state: State):
         assert self.is_train is not None, "self.is_train should be set by the callback"
-        total_steps = len(state.train_dataloader) if self.is_train else len(state.eval_dataloader)
+        total_steps = len(state.train_dataloader) if self.is_train else sum(
```
moinnadeem, line 120: For TQDM, does the total_steps include evaluation, or do we want to have evaluation be part of a separate TQDM instance? Curious for @ravi-mosaicml 's thought here too

composer/trainer/trainer_hparams.py
```diff
@@ -142,6 +136,11 @@ class TrainerHparams(hp.Hparams):
         "Determines the number of microbatches to split a per-gpu batch into, used to compensate for low-memory-capacity devices."
     )
     precision: Precision = hp.required(doc="Precision to use for training", template_default=Precision.AMP)
+
+    val_dataset: Optional[datasets.DatasetHparams] = hp.optional(doc="Validation dataset hparams", default=None)
+
+    evaluators: Optional[List[EvaluatorHparams]] = hp.optional(doc="Evaluators", default_factory=list)
```
moinnadeem, line 142: Can we add more descriptive docstrings around the purpose of evaluators? It should help user friendliness

moinnadeem (changes requested):
Great work! I like the structure. Overall, we should probably work on more purposeful docstrings that help the user understand *why* they need to fill out this field -- it'll go a long way towards making our library more user-friendly. 

Otherwise, I am curious on how we want to design our abstraction for TQDM steps. Should evaluation time be part of the TQDM logger? Curious for thoughts from others.

composer/datasets/evaluator.py
ravi-mosaicml: Instead of having to enumerate all the metrics, would it work to have a `metric_name` hparam field, and then do a `metric = getattr(torchmetrics, self.metric_name)`?

composer/datasets/dataset_registry.py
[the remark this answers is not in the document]
ravi-mosaicml: I'm guess this needed to be moved out of the trainer_hparams so it can also be used by the evaluators hparams object?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+    metrics: List[str] = hp.required(doc="metrics", template_default=list)
+
+    def initialize_object(self):
+        dataset = self.eval_dataset.initialize_object()
+
+        def resolve_metric_list(metric_names: List[str]) -> List[Metric]:
+            resolved_metrics = []
+            for metric_name in metric_names:
+                if metric_name in self.metric_registry:
+                    metric_instance = self.metric_registry[metric_name]()
```
ravi-mosaicml, line 52: Do any metrics have custom parameters that need to be specified?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+                    resolved_metrics.append(metric_instance)
+                else:
+                    log.warning(f"The metric {metric_name} is not a registered metric. Add it to the metric registry.")
+            return resolved_metrics
+
+        metric_list = resolve_metric_list(self.metrics)
```
ravi-mosaicml, line 58: NIT: do we need a helper function?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+        "SpearmanCorrcoef": SpearmanCorrcoef,
+        "CrossEntropyLoss": CrossEntropyLoss,
+    }
+
+    label: str = hp.required(doc="label")
+    eval_dataset: DatasetHparams = hp.required(doc="Training dataset hparams")
+    metrics: List[str] = hp.required(doc="metrics", template_default=list)
+
+    def initialize_object(self):
+        dataset = self.eval_dataset.initialize_object()
```
ravi-mosaicml, line 46:
```suggestion
        dataloader_spec = self.eval_dataset.initialize_object()
```

composer/loggers/tqdm_logger.py
ravi-mosaicml: Total steps should be the number of times TQDM is updated. Currently this is done on each batch. IDK, could go either way on this -- don't have a strong preference on the TQDM format.

composer/trainer/trainer.py
```diff
@@ -49,10 +50,15 @@ class Trainer:
     Args:
         model (BaseMosaicModel): The model to train.
         train_dataloader_spec (DataloaderSpec): The dataloader spec for the training data.
-        eval_dataloader_spec (DataloaderSpec): The dataloader spec for the evaluation data.
+        eval_dataloader_spec (Optional[DataloaderSpec]): The dataloader spec for the evaluation
```
ravi-mosaicml, line 53: Please make sure to rebase against dev. I don't think the diff has incorporated the latest dataloader changes.

ravi-mosaicml: (I'll review the training loop changes after the rebase)

composer/trainer/trainer_hparams.py
```diff
@@ -142,6 +136,11 @@ class TrainerHparams(hp.Hparams):
         "Determines the number of microbatches to split a per-gpu batch into, used to compensate for low-memory-capacity devices."
     )
     precision: Precision = hp.required(doc="Precision to use for training", template_default=Precision.AMP)
+
+    val_dataset: Optional[datasets.DatasetHparams] = hp.optional(doc="Validation dataset hparams", default=None)
```
ravi-mosaicml, line 140: Should val dataset be removed?

composer/trainer/trainer_hparams.py
```diff
@@ -195,6 +194,10 @@ def validate(self):
                 f"eval batch size ({self.eval_batch_size}) not divisible by the total number of processes ({world_size}) "
             )
 
+        # There should be at least one structure for validation - evaluators or ordinary validation dataset
+        if self.val_dataset is None and not self.evaluators:
```
ravi-mosaicml, line 198: Hmm interesting point....but should this be an error? Or a warning (if there use cases for training without ever evaluating?)

ravi-mosaicml (commented):
I like where this is headed. Another thought I had is whether we want to allow some evaluators to be run every batch whereas others run every epoch? Or would we want different evaluators to run at different frequencies? Perhaps the `validate_every_n_epochs` and `validate_every_n_batches` parameters could be incorporated into the evaluator haprams.

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class EvaluatorHparams(hp.Hparams):
+    hparams_registry = {  # type: ignore
+        "eval_dataset": get_dataset_registry(),
+    }
+
+    metric_registry = {
```
ajaysaini725, line 32: Ideally, we want to support all metrics listed in https://torchmetrics.readthedocs.io/en/latest/references/modules.html (there are a lot) in addition to custom metrics that the user has defined (see `Dice` in `composer/models/loss.py` as an example).

We shouldn't have to instantiate the metrics in Hparams because each model class that extends `BaseMosaicModel` returns instantiated objects for each metric that can be computed for the model. 

The evaluator only needs to track the names of the metrics so that it can filter out metrics that shouldn't be used for the given eval_dataset. 

For example, we can have `metric_names: List[str]` be a field here and then in the `eval()` loop of the `Trainer` we can check the `metric_names` against the metrics returned by `model.metrics()` and only compute the metrics whose names match the ones for this `Evaluator`.

ajaysaini725 (changes requested):
Great progress - biggest thing is my comment on the `metric_registry`

composer/datasets/evaluator.py
ajaysaini725: Oh also another thing - the list of metric names should be optional so that if not provided it defaults to using all metrics returned by `model.metrics()`

moinnadeem:
Just checking, where do we stand on this?

composer/core/types.py
```diff
@@ -77,6 +78,22 @@ def as_batch_pair(batch: Batch) -> BatchPair:
     return batch
 
 
+def _split_fn(batch: Batch, n_microbatches: int) -> List[Batch]:
```
ravi-mosaicml, line 81: This function was moved to the `DataSpec` class; I presume it was left over from the merge?

composer/core/types.py
```diff
@@ -140,6 +157,32 @@ def __len__(self) -> int:
...
+    dataset.
+
+    Attributes:
+        label (str): Name of the Evaluator
+        dataset (Union[DataLoader, DataSpec]): Dataloader for evaluation data
+        metrics (Metrics): Metrics to use for the dataset
+        validate_every_n_epochs: (int)
+        validate_every_n_batches: (int)
+        metric_names: (Optional[List[str]])
+        device_transform_fn (TDeviceTransformFn, optional):
```
ravi-mosaicml, line 172: Docstring is out of date I think...

composer/core/types.py
```diff
@@ -140,6 +157,32 @@ def __len__(self) -> int:
...
+            A function to modify the data once it has been loaded onto the device (for example, GPU-based batch normalization)
+            This function is invoked with a batch of data after it has been moved onto the device,
+            and it is expected to return a batch.
+    """
+
+    label: str
+    dataset: Union[DataLoader, DataSpec]
```
ravi-mosaicml, line 179:
```suggestion
    dataloader: Union[DataLoader, DataSpec]
```

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,48 @@
...
+        "metric name for torchmetrics and use the classname for custom metrics.",
+        default_factory=list)
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams):
+        dataloader = self.eval_dataset.initialize_object(batch_size=batch_size, dataloader_hparams=dataloader_hparams)
+
+        # Populate the metrics later in the trainer initialization
+        return Evaluator(label=self.label,
+                         dataloader=dataloader,
+                         metrics=MetricCollection([]),
```
ravi-mosaicml, line 45: Could you build the `metrics` here and then `Evaluator` wouldn't need to take in `metric_names`?

composer/core/state.py
```diff
@@ -134,12 +134,13 @@ def __init__(
...
         self.train_data = train_dataloader
-        if not isinstance(eval_dataloader, DataSpec):
-            if isinstance(eval_dataloader, dict):
-                eval_dataloader = DataSpec(**eval_dataloader)
-            else:
-                eval_dataloader = DataSpec(eval_dataloader)
-        self.eval_data = eval_dataloader
+        for evaluator in evaluators:
+            if not isinstance(evaluator.dataset, DataSpec):
+                if isinstance(evaluator.dataset, dict):
```
ravi-mosaicml, line 139: Can you move this logic to `Evaluator.__init__`, rather than here? (That way it also won't need to be duplicated on the setter below)

composer/core/state.py
```diff
@@ -134,12 +134,13 @@ def __init__(
...
+                    evaluator.dataset = DataSpec(**evaluator.dataset)
+                else:
+                    evaluator.dataset = DataSpec(evaluator.dataset)
+        self.evaluator_data = evaluators
```
ravi-mosaicml, line 143: nit: 
```suggestion
        self._evaluators = evaluators
```
An evaluator is more than just the data :)

tests/datasets/test_synthetic_data.py
```diff
@@ -25,6 +25,7 @@ def test_synthetic_data_creation(data_type: SyntheticDataType, label_type: Synth
     else:
         num_classes = 10
         label_shape = (1, 10, 12)
+        # run run
```
ravi-mosaicml, line 28:
```suggestion
```

composer/core/state.py
```diff
@@ -175,12 +176,18 @@ def train_dataloader(self, dataloader: types.DataLoader):
...
 
     @property
-    def eval_dataloader(self):
-        return self.eval_data.dataloader
-
-    @eval_dataloader.setter
-    def eval_dataloader(self, dataloader: types.DataLoader):
-        self.eval_data = DataSpec(dataloader)
+    def evaluators(self):
+        return self.evaluator_data
```
ravi-mosaicml, line 180: I don't think 
```suggestion
        return self._evaluators
```

composer/core/state.py
```diff
@@ -175,12 +176,18 @@ def train_dataloader(self, dataloader: types.DataLoader):
...
+
+    @evaluators.setter
+    def evaluators(self, evaluators: Sequence[types.Evaluator]):
+        for evaluator in evaluators:
```
ravi-mosaicml, line 184:
```suggestion
        self._evaluators[:] = ensure_tuple(optimizers)
```
This will ensure that the evaluators is updated in-place in case if something has a reference to it.

composer/core/types.py
```diff
@@ -138,6 +139,23 @@ def __len__(self) -> int:
 
 Metrics = Union[Metric, MetricCollection]
 
+
+@dataclass
+class Evaluator:
```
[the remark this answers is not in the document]
ravi-mosaicml: I modify my earlier feedback and instead would recommend to make this a plain old python class....that way you can do conversion on `__init__` (like converting the dataloader or a dictionary to a dataspec)

In addition, can you put `eval_subset_num_batches` in here and make that a property for each evaluator?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,48 @@
...
+from composer.datasets.hparams import DatasetHparams
+from composer.datasets import DataloaderHparams
+from typing import Optional
+
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class EvaluatorHparams(hp.Hparams):
+    hparams_registry = {  # type: ignore
```
ravi-mosaicml, line 23: Docstrings...

composer/trainer/trainer_hparams.py
```diff
@@ -222,6 +221,10 @@ def validate(self):
                 f"Eval batch size ({self.eval_batch_size}) not divisible by the total number of processes ({world_size})."
             )
 
+        # There should be at least one structure for validation - evaluators or ordinary validation dataset
+        if self.val_dataset is None and self.evaluators is None:
```
ravi-mosaicml, line 225: Sometimes (e.g. in tests) I'd like to do a run without validation. I think it'd make sense to raise a warning but not error out.
In addition, would it make sense to remove the `val_dataset` hparam since now that should be specified in `evaluators`?

composer/trainer/trainer.py
```diff
@@ -117,7 +121,8 @@ def __init__(
             *,
             model: BaseMosaicModel,
             train_dataloader: Union[DataLoader, DataSpec],
-            eval_dataloader: Union[DataLoader, DataSpec],
+            eval_dataloader: Optional[Union[DataLoader, DataSpec]] = None,
```
ravi-mosaicml, line 124:
```suggestion
```
I'd be for specifying just the evaluators, not the eval dataloader

ravi-mosaicml: Can you also remove the hparams for valiadate_every_n_epochs and validate_every_n_batches?

composer/trainer/trainer.py
```diff
@@ -202,29 +206,58 @@ def __init__(
...
                 self.ddp_sync_strategy = DDPSyncStrategy(ddp_sync_strategy)
 
-        # TODO(#123): DeepSpeed still needs a precision context, but it's not completely clear how to
-        # handle this with our version of Pytorch
-        precision_context = self.device.precision_context if not self.deepspeed_enabled else cast(
-            Callable[..., ContextManager], contextlib.nullcontext)
+        if evaluators is not None:
+            for evaluator in evaluators:
+                if isinstance(evaluator.dataset, DataSpec):
+                    dataloader_spec = evaluator.dataset
```
ravi-mosaicml, line 212: Same comment re: doing the conversion in the evaluator

composer/trainer/trainer.py
```diff
@@ -202,29 +206,58 @@ def __init__(
...
+                else:
+                    dataloader_spec = DataSpec(evaluator.dataset)
+                evaluator.dataset = DDPDataLoader(dataloader_spec.dataloader)
+        self.evaluators = evaluators
+
+        if eval_dataloader is not None:
```
ravi-mosaicml, line 218: Same comment: re remove the eval_dataloader option for the evaluators option

composer/trainer/trainer.py
```diff
@@ -202,29 +206,58 @@ def __init__(
...
-            model=model,
-            grad_accum=grad_accum,
-            precision=precision,
-            precision_context=precision_context,
-            train_dataloader=train_dataloader,
-            eval_dataloader=eval_dataloader,
-        )
+        # do a check here to make sure there is at least one validation set
+        if self.evaluators is None or len(self.evaluators) == 0:
+            raise ValueError('At least one validation set should be used and passed in through ',
```
ravi-mosaicml, line 239: I'd like to allow no evaluators (e.g. in tests), so an error is a bit harsh, but I think a warning would be fine here.

composer/trainer/trainer.py
```diff
@@ -238,11 +271,14 @@ def __init__(
...
         if eval_subset_num_batches is not None:
-            if eval_subset_num_batches > len(self.state.eval_dataloader):
-                warnings.warn(
-                    textwrap.dedent(f"""SubsetNumBatchesWarning: The eval_subset_num_batches({eval_subset_num_batches})
-                        is greater than the number of batches in the evaluation dataloader
-                        ({len(self.state.eval_dataloader)})"""))
+            for evaluator in self.evaluators:
+                assert isinstance(evaluator.dataset, DataSpec)
+                eval_dataspec = evaluator.dataset.dataloader
+                if eval_subset_num_batches > len(eval_dataspec):
```
ravi-mosaicml, line 277: Can you make this an evaluator-specific option (rather than the same value used for all evaluators?)

composer/trainer/trainer.py
```diff
@@ -459,19 +524,84 @@ def _get_metrics_as_collection(self, *, is_train: bool) -> MetricCollection:
...
+        assert isinstance(original_model, BaseMosaicModel)
+        metrics = original_model.metrics(train=False)
+        assert isinstance(metrics, (Metric, MetricCollection)), \
+        "   Error module.metrics() must return a Metric or MetricCollection object."
+        if isinstance(metrics, Metric):
+            # Forcing metrics to be a MetricCollection simplifies logging results
+            metrics = MetricCollection([metrics])
+
+        # Use all the metrics from the model if no metric_names are specified
+        if evaluator.metric_names is None or evaluator.metrics is None or evaluator.metrics:
```
ravi-mosaicml, line 554: This is more of a design question, but should metrics be specified on the model or in the evaluators? I'd be for having just one place for them -- ideally the evaulators calss? I think it can be confusing if they could be specified in multiple places. Would it be ok to remove the `metrics` from the `BaseMosaicModel` and instead require that metrics be specified as `Evaluator`s?

ravi-mosaicml (commented):
Thanks for updating this to the latest training loop, sorry there were like 1000 changes in there.

This is looking better! A few questions:

1. I'd prefer for there to be just one way of defining things. e.g. use the evalautors instead of val_dataset, require metrics to be specified via `evaluators` rather than on the model
2. I think some docstrings are missing; I took a quick pass but didn't mark them all.
3. Do we need any evaluator-specific test cases? It's totally possible that the existing test cases cover this, in which case that's fine.

composer/trainer/trainer.py
ajaysaini725: One issue with that suggestion is that right now we have the requirement that the output of a model's `validate()` can be passed directly into the `update()` function for all metrics that are computed for the model.

If the specification of the `metrics` is completely disjoint from the model (ex. by specifying metrics only in `Evaluator`s), then it opens up the possibility of bugs where the output of a model's `validate()` is not compatible with a metric specified in an `Evaluator`.

composer/trainer/trainer.py
```diff
@@ -361,17 +397,37 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
         train_dataloader = hparams.train_dataset.initialize_object(train_device_batch_size, hparams.dataloader)
 
         eval_device_batch_size = hparams.eval_batch_size // dist.get_world_size()
-        if hparams.val_dataset.shuffle and hparams.eval_subset_num_batches:
-            warnings.warn(
-                textwrap.dedent(f"""SubsetNumBatchesWarning: When specifying eval_subset_num_batches,
-            (set to {hparams.eval_subset_num_batches}), val_dataset.shuffle should be set to False. Otherwise,
-            each evaluation epoch may load a different subset of samples."""))
-        eval_dataloader = hparams.val_dataset.initialize_object(eval_device_batch_size, hparams.dataloader)
+        if hparams.val_dataset is not None:
```
ajaysaini725, line 400: Probably don't need this since there are Evaluators now?

composer/trainer/trainer.py
ajaysaini725: @ravi-mosaicml To add to my comment above:

What makes this tricky is that we want to be able to specify `Evaluator` objects in YAML files, which means that we can only specify `metrics` for the `Evaluator` using strings.

Metrics can either come from the `torchmetrics` library or be a user-defined class that extends `torchmetrics.Metrics`. Also, in the latter case, there may be arbitrary parameters needed to instantiate the metric.

Because there doesn't seem to be a clear way for an Evaluator to instantiate a `metric` object just from the name (and possibly parameter list), the cleanest solution seemed to be to have the model return the `metric` objects that are associated with it and then have each individual `Evaluator` select the relevant `metric` objects from what gets returned by `model.metrics()` using the name of the metric. 

Definitely open to alternative solutions if you have something in mind - these were just my thoughts.

composer/trainer/trainer.py
anisehsani: I had thought maybe we'd leave the option for users to pass in a eval_dataloader and just create an evaluator from it in the trainer. That way, a user wouldn't have to interact with specifying metrics/evaluators at all if they have a simple model with just one evaluation dataset.

composer/core/types.py
```diff
@@ -141,6 +142,58 @@ def __len__(self) -> int:
...
+            modified.
+    """
+
+    label: str
+    dataloader: Union[DataLoader, DataSpec]
+    metrics: Metrics = None
+    validate_every_n_epochs: int = 1
+    validate_every_n_batches: int = -1
+    metric_names: Optional[Sequence[str]] = None
+    eval_subset_num_batches: Optional[int] = None
```
hanlint, line 176: remove this and use the `eval_subset_num_batches` in the Trainer `init` signature

composer/core/types.py
```diff
@@ -141,6 +142,58 @@ def __len__(self) -> int:
...
+    device_transforms: Optional[Callable[[Batch], Batch]] = None
+    _data_spec: Optional[DataSpec] = None
```
hanlint, line 178: Remove these two, as discussed offline.

composer/trainer/trainer.py
```diff
@@ -120,7 +123,8 @@ def __init__(
             *,
             model: BaseMosaicModel,
             train_dataloader: Union[DataLoader, DataSpec],
-            eval_dataloader: Union[DataLoader, DataSpec],
+            eval_dataloader: Optional[Union[DataLoader, DataSpec]] = None,
+            evaluators: Optional[List[Evaluator]] = None,
```
hanlint, line 127: As discussed, let's unify `evaluators` and `eval_dataloaders` into a single `eval_dataloader` argument.

composer/trainer/trainer.py
```diff
@@ -208,18 +211,36 @@ def __init__(
...
             Callable[..., ContextManager], contextlib.nullcontext)
 
         if not isinstance(train_dataloader, DataSpec):
             train_dataloader = DataSpec(train_dataloader)
-        if not isinstance(eval_dataloader, DataSpec):
-            eval_dataloader = DataSpec(eval_dataloader)
 
         self._train_data_spec = train_dataloader
-        self._eval_data_spec = eval_dataloader
+        # self._eval_data_spec = eval_dataloader
```
hanlint, line 243: nit: cleanup

composer/trainer/trainer.py
```diff
@@ -488,19 +544,94 @@ def _get_metrics_as_collection(self, *, is_train: bool) -> MetricCollection:
...
+        assert isinstance(metrics, (Metric, MetricCollection)), \
+        "   Error module.metrics() must return a Metric or MetricCollection object."
+        if isinstance(metrics, Metric):
+            # Forcing metrics to be a MetricCollection simplifies logging results
+            metrics = MetricCollection([metrics])
+
+        # Use all the metrics from the model if no metric_names are specified
+        if evaluator.metrics is None:
+            if evaluator.metric_names is None:
+                evaluator_metrics = copy.deepcopy(metrics)
```
hanlint, line 572: Add a warning if deepcopy fails with that particular error.

composer/trainer/trainer.py
```diff
@@ -488,19 +544,94 @@ def _get_metrics_as_collection(self, *, is_train: bool) -> MetricCollection:
...
+            else:
+                evaluator_metrics = MetricCollection([])
+                for metric_name in evaluator.metric_names:
+                    if metric_name in metrics.keys():
+                        evaluator_metrics.add_metrics(copy.deepcopy(metrics[metric_name]))
+                    else:
+                        warnings.warn(
```
hanlint, line 579: This should also be an error.

ravi-mosaicml (commented):
Sorry, I realize I had some pending comments from before I left. Submitting them now!

composer/loggers/tqdm_logger.py
```diff
@@ -108,9 +108,10 @@ def _start(self, state: State):
         if dist.get_global_rank() != 0:
             return
         assert self.is_train is not None, "self.is_train should be set by the callback"
-        # TODO(anis) -- in #120, len(state.eval_dataloader) is inaccurate, as it does not incorporate
-        # trainer._eval_subset_num_batches. The evaluator spec should fix this.
-        total_steps = state.steps_per_epoch if self.is_train else len(state.eval_dataloader)
+        assert state.evaluators is not None
+        total_steps = state.steps_per_epoch if self.is_train else sum(
+            len(evaluator.dataloader) for evaluator in state.evaluators)
```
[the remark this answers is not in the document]
anisehsani: the state doesn't include eval_subset_num_batches at this point - should I add it to the state? The state doesn't have a trainer at this point either

composer/trainer/trainer.py
```diff
@@ -819,35 +951,46 @@ def eval(self, is_batch: bool):
 
             self.engine.run_event(Event.EVAL_START)
 
-            metrics = self._get_metrics_as_collection(is_train=False)
+            for evaluator in state.evaluators:
+                assert isinstance(evaluator.dataloader, DataLoader)
```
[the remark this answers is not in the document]
anisehsani: The evaluator.dataloader is always a DataSpec. This is to simplify the evaluator so we don't have to copy device_transforms etc into the evaluator.

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,45 @@
...
+        default_factory=list)
+    eval_subset_num_batches: Optional[int] = hp.optional("If specified, evaluate on this many batches.", default=None)
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams):
+        dataloader = self.eval_dataset.initialize_object(batch_size=batch_size, dataloader_hparams=dataloader_hparams)
+
+        # Populate the metrics later in the trainer initialization
+        return Evaluator(
+            label=self.label,
+            dataloader=dataloader,
```
ravi-mosaicml, line 43: Pass eval_subset_num_batches through?

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,45 @@
...
+
+    See the documentation for the :class:`Evaluator`.
+    """
+    hparams_registry = {  # type: ignore
+        "eval_dataset": get_dataset_registry(),
+    }
+
+    label: str = hp.required(doc="Name of the Evaluator object. Used for logging/reporting metrics")
+    eval_dataset: DatasetHparams = hp.required(doc="Evaluator dataset for the Evaluator")
+    metric_names: Optional[List[str]] = hp.optional(
```
ravi-mosaicml, line 31: Should it be a `List[str]`? Curious when it would be none?

ravi-mosaicml: Also should it be required?

composer/profiler/dataloader_profiler.py
```diff
@@ -49,4 +49,5 @@ def init(self, state: State, logger: Logger):
                 textwrap.dedent("""To use the dataloader profiler, state.profiler must be set.
                 Make sure to run composer with the profiler -- i.e. with the `--profiler` CLI flag."""))
         state.train_dataloader = ProfiledDataLoader(state.profiler, state.train_dataloader, "train")
-        state.eval_dataloader = ProfiledDataLoader(state.profiler, state.eval_dataloader, "eval")
+        for evaluator in state.evaluators:
+            evaluator.dataloader = ProfiledDataLoader(state.profiler, evaluator.dataloader, "eval")
```
ravi-mosaicml, line 53: I forget how the profiler works,but should the name be the evaluator name rather than just "eval"?

composer/trainer/trainer.py
```diff
@@ -122,7 +124,7 @@ def __init__(
             *,
             model: BaseMosaicModel,
             train_dataloader: Union[DataLoader, DataSpec],
-            eval_dataloader: Union[DataLoader, DataSpec],
+            eval_dataloader: Union[DataLoader, DataSpec, List[Evaluator]],
```
ravi-mosaicml, line 127: Type annotation is inconsistent with docstring. Can we support an evaluator or sequence of evaluators?

composer/trainer/trainer.py
```diff
@@ -222,6 +223,25 @@ def __init__(
...
 
+        self.evaluators = []
+        if isinstance(eval_dataloader, DataLoader):
+            eval_dataloader = DataSpec(eval_dataloader)
+        if isinstance(eval_dataloader, DataSpec):
+            default_evaluator = Evaluator(label="eval_dataset", dataloader=eval_dataloader, metrics=None)
+            self.evaluators = [default_evaluator]
+        else:
+            self.evaluators = []
+            assert isinstance(eval_dataloader, list)
```
ravi-mosaicml, line 234: Can you convert to a typeerror?

composer/trainer/trainer.py
```diff
@@ -222,6 +223,25 @@ def __init__(
...
+            for evaluator in eval_dataloader:
+                assert isinstance(evaluator, Evaluator)
```
ravi-mosaicml, line 236: I'm a bit confused on why the list needs to be rebuilt...if it's a list then isn't it going to be a list of evaluators?

composer/trainer/trainer.py
```diff
@@ -222,6 +223,25 @@ def __init__(
...
+                self.evaluators.append(evaluator)
+
+        # do a check here to make sure there is at least one validation set
+        if len(self.evaluators) == 0:
+            warnings.warn(f'At least one validation set should be used and passed in through ',
+                          'eval_dataloader or the evaluators',
```
ravi-mosaicml, line 242: I think the qarning message is out of date

composer/trainer/trainer.py
```diff
@@ -246,9 +263,13 @@ def __init__(
             precision=precision,
             precision_context=precision_context,
             train_dataloader=train_dataloader.dataloader,
-            eval_dataloader=eval_dataloader.dataloader,
+            evaluators=self.evaluators,
         )
 
+        if self.evaluators:
```
ravi-mosaicml, line 269: I don't think this should need the if guard? State.evaluators is always a (potentially empty) list?

composer/trainer/trainer.py
```diff
@@ -403,12 +426,26 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
...
+        if hparams.evaluators is not None and len(hparams.evaluators) > 0:
+            eval_dataloader = [
+                evaluator.initialize_object(eval_device_batch_size, hparams.dataloader)
+                for evaluator in hparams.evaluators
+            ]
+            for evaluator in hparams.evaluators:
+                if evaluator.eval_dataset.shuffle and hparams.eval_subset_num_batches:
+                    warnings.warn(
+                        textwrap.dedent(f"""SubsetNumBatchesWarning: When specifying eval_subset_num_batches,
+                    (set to {hparams.eval_subset_num_batches}), evaluator.dataloader.shuffle should be set to False. Otherwise,
```
ravi-mosaicml, line 446: Can you add context here to which evaluator this warning is for? And then remove the break?

ravi-mosaicml (commented):
Looking better! Didn't have time to go through it fully but left some comments. Hopefully it'll be merged in next week!

composer/datasets/evaluator.py
hanlint: Let's freeze the API design for now. We can make subsequent improvements, but I don't want to delay this PR any further. The optional use case here is if the user has multiple datasets and wants to apply all the metrics provided by the model's `metrics() -> MetricCollection` method.

composer/datasets/evaluator.py
hanlint: We decided to remove `eval_subset_num_batches` from Evaluator to simplify this class since its a edge use case.

composer/trainer/trainer.py
```diff
@@ -403,12 +426,26 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
         train_dataloader = hparams.train_dataset.initialize_object(train_device_batch_size, hparams.dataloader)
 
         eval_device_batch_size = hparams.eval_batch_size // dist.get_world_size()
-        if hparams.val_dataset.shuffle and hparams.eval_subset_num_batches is not None:
-            warnings.warn(
-                textwrap.dedent(f"""SubsetNumBatchesWarning: When specifying eval_subset_num_batches,
-            (set to {hparams.eval_subset_num_batches}), val_dataset.shuffle should be set to False. Otherwise,
-            each evaluation epoch may load a different subset of samples."""))
-        eval_dataloader = hparams.val_dataset.initialize_object(eval_device_batch_size, hparams.dataloader)
+        if hparams.val_dataset is not None:
```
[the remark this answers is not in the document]
ravi-mosaicml: Implemented

ravi-mosaicml (commented):
I started to leave a review and then realized it would be easier if I just implemented some of the changes directly, so I added a commit for that. I'll continue the review after lunch.

composer/core/state.py
```diff
@@ -144,7 +145,7 @@ def __init__(
         self.model = model
         self.grad_accum = grad_accum
         self.train_dataloader = train_dataloader
-        self.eval_dataloader = eval_dataloader
+        self.evaluators = evaluators
```
hanlint, line 148: `self.eval_dataloader`, _not_ `self.evaluator` here

anisehsani: As discussed offline - we're going to leave it as self.evaluators and change it in a further PR

composer/core/state.py
```diff
@@ -79,8 +79,9 @@ class State(Serializable):
         grad_accum (int): The number of gradient accumulation steps to use. The size of each microbatch is ``train_batch_size / num_gpus / grad_accum``.
         train_dataloader (types.DataLoader, types.DataSpec, or dict):
             The :class:`types.DataLoader`, :class:`types.DataSpec`, or dict of :class:`types.DataSpec` kwargs to used for training.
-        eval_dataloader (types.DataLoader, types.DataSpec, or dict):
-            The :class:`types.DataLoader`, :class:`types.DataSpec`, or dict of :class:`types.DataSpec` kwargs to used for evaluation.
+        evaluators (Evaluators):
+            The :class:`types.Evaluators` contain the evaluation datasets used for evaluation with specific metrics.
+        max_epochs (int): The maximum number of epochs to train for.
```
hanlint, line 84:
```suggestion
```

composer/trainer/trainer_hparams.py
```diff
@@ -231,6 +228,14 @@ def validate(self):
                 f"Eval batch size ({self.eval_batch_size}) not divisible by the total number of processes ({world_size})."
             )
 
+        # There should be at least one structure for validation - evaluators or ordinary validation dataset
+        if self.evaluators is None or len(self.evaluators) == 0:
+            if self.val_dataset is None:
+                raise ValueError("val_dataset and evaluators shouldn't both be empty")
```
ravi-mosaicml, line 234: We have a warning for this in the trainer. Can you remove this check here? Please keep the one below though.

ravi-mosaicml (commented):
Looking really close! Just a final few comments.

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,91 @@
...
+        doc=textwrap.dedent("""Name of the metrics for the evaluator. Can be a torchmetrics metric name or the
+        class name of a metric returned by model.metrics(). If None (the default), uses all metrics in the model"""),
+        default=None)
+    eval_batch_size: Optional[int] = hp.optional(
```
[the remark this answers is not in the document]
hanlint: Remove from `Evaluator` and just pass in `initialize_object`.

composer/datasets/evaluator.py
```diff
@@ -0,0 +1,91 @@
...
+        doc="batch size to use for each evaluation step",
+        default=None,
+    )
+
+    def initialize_object(self, model: BaseMosaicModel, batch_size: int, dataloader_hparams: DataloaderHparams):
```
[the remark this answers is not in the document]
hanlint: keep this here

composer/trainer/trainer_hparams.py
```diff
@@ -161,6 +154,10 @@ class TrainerHparams(hp.Hparams):
     )
     precision: Precision = hp.required(doc="Precision to use for training", template_default=Precision.AMP)
 
+    val_dataset: Optional[datasets.DatasetHparams] = hp.optional(doc="Validation dataset hparams", default=None)
+
+    evaluators: Optional[List[EvaluatorHparams]] = hp.optional(doc="Evaluators", default_factory=list)
```
hanlint, line 159: cc: @siriuslee , API change in YAML

hanlint (approved):
LGTM, thanks for all the work here!

PR #130, opened 2021-12-02
Fix loss reporting

jbloxham:
Need to add to `total_loss` before call to `scaler.scale()`.


moinnadeem:
Yup, I can confirm that this reports the appropriate value. LGTM!

moinnadeem (approved):

PR #131, opened 2021-12-03
Remove argparse from setup.py

ravi-mosaicml:
Argparse is included in the stdlib for python 3.2+, so no need to install from pip. It generates a warning when using colab that you need to restart the runtime.

image


Averylamp (approved):
Lgtm

PR #132, opened 2021-12-03
Fixed pickling of torch.memory_format objects

ravi-mosaicml:
Torch.memory_format cannot be attached to datasets that are passed into dataloaders, as they are not pickleable. Instaed, storing the enum until it is ready to be used


bandish-shah (approved):
LGTM

PR #141, opened 2021-12-08
ssd-resnet34 on coco

florescl:
current run here https://wandb.ai/mosaic-ml/ssd/runs/3p7840t1/overview?workspace=user-laura.

@Landanjs @A-Jacobson


A-Jacobson:
Hey Laura nice job, sorry for the holiday delay. That said. I need a few questions answered before I can really review this.

1. What's the target performance for this and are we logging it anywhere? The ssd runs i see in w&b are only logging loss. @florescl 

2. How do we deal with adapted code that doesn't fit our style guidelines? @ravi-mosaicml

ravi-mosaicml:
It depends on how much code:

1. If its an entire file, and we are not modifying it at all, then it's fine to exclude it from our formatting rules by adding this to the top:

```python3
# isort: skip_file
# type: ignore
# yapf: disable
```

If its anything less than an entire file, or we are modifying it, then it should be brought up to date with out guidelines.

In both cases, there should be a citation (url is fine) to where the code came from.

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,161 @@
...
+        with torch.no_grad():
+            ploc, plabel = self.module(img)
+            ploc, plabel = ploc.float(), plabel.float()
+
+            for idx in range(ploc.shape[0]):
+                ploc_i = ploc[idx, :, :].unsqueeze(0)
+                plabel_i = plabel[idx, :, :].unsqueeze(0)
+
+                try:
+                    result = encoder.decode_batch(ploc_i, plabel_i, 0.50, 200)[0]
```
A-Jacobson, line 97:
```suggestion
                    result = encoder.decode_batch(ploc_i, plabel_i, overlap_threshold, nms_max_detections)[0]
```

composer/models/ssd/utils.py
```diff
@@ -0,0 +1,510 @@
...
+
+    plt.imshow(img)
+    ax = plt.gca()
+    for (cx, cy, w, h), label in zip(bboxes, labels):
+        if label == "background":
+            continue
+        ax.add_patch(patches.Rectangle((cx - 0.5 * w, cy - 0.5 * h), w, h, fill=False, color="r"))
+        bbox_props = dict(boxstyle="round", fc="y", ec="0.5", alpha=0.3)
+        ax.text(cx - 0.5 * w, cy - 0.5 * h, label, ha="center", va="center", size=15, bbox=bbox_props)
+    plt.show()
```
A-Jacobson, line 510:
```suggestion
```

A-Jacobson: not used, though maybe useful to keep around for vis/debugging

composer/models/ssd/utils.py
```diff
@@ -0,0 +1,510 @@
...
+                                             torch.tensor(labels_out,
+                                                          dtype=torch.long), \
+                                             torch.cat(scores_out, dim=0)
+
+        _, max_ids = scores_out.sort(dim=0)
+        max_ids = max_ids[-max_output:]
+        return bboxes_out[max_ids, :], labels_out[max_ids], scores_out[max_ids]
+
+
+class DefaultBoxes(object):
```
A-Jacobson, line 207: This is a pretty important part of SSD. We should document the box generation better if we're going to stick with this implementation.

composer/models/ssd/utils.py
```diff
@@ -0,0 +1,510 @@
...
+    area1 = delta1[:, :, 0] * delta1[:, :, 1]
+    delta2 = be2[:, :, 2:] - be2[:, :, :2]
+    area2 = delta2[:, :, 0] * delta2[:, :, 1]
+
+    iou = intersect / (area1 + area2 - intersect)
+    return iou
+
+
+# This function is from https://github.com/kuangliu/pytorch-ssd.
+class Encoder(object):
```
A-Jacobson, line 58:
```suggestion
class BoxEncoder(object):
```

A-Jacobson (commented):
Overall... this is a lot. I think you did an awesome job getting it to work and we should use it to test methods, for now. 

Unfortunately, the reference implementation (mlperf?) is sort of a mess. We're getting 3rd and 4th hand code copied from multiple hobby projects through it. Really, makes me wonder if only Ross Girshick and a few other people truly understand how to build these things (only half joking). 

In the long term, I think we should move towards wrapping detectron2, mmdetection, or torchvision rather than investing in truly cleaning this up + maintaining it. This will align us with industry and avoid some of the common pitfalls with these complicated models (there are a lot) so we don't get bottlenecked by non-deeplearning ops such as the looped non-maximum-supression we have here vs something like https://pytorch.org/vision/stable/ops.html

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,161 @@
...
+        train_annotate = os.path.join(data, "annotations/instances_train2017.json")
+        train_coco_root = os.path.join(data, "train2017")
+        val_coco = COCODetection(val_coco_root, val_annotate, val_trans)
+
+        inv_map = {v: k for k, v in val_coco.label_map.items()}
+
+        ret = []
+
+        overlap_threshold = 0.50
+        nms_max_detections = 200
```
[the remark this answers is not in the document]
A-Jacobson:
```suggestion
        overlap_threshold = 0.50
        nms_max_detections = 100
```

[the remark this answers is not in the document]
A-Jacobson: nvidia results are reported with 100 max detections

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,161 @@
...
+                                loc_[1] * htot,
+                                (loc_[2] - loc_[0]) * wtot,
+                                (loc_[3] - loc_[1]) * htot,
+                                prob_,
+                                inv_map[label_]])
+
+        return ret, ret
+
+
+class my_map(Metric):
```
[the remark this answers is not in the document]
A-Jacobson: torchmetrics has a MAP metric already. may be better to use this to avoid depending on COCO since loading it is odd. https://torchmetrics.readthedocs.io/en/latest/references/modules.html#meanaverageprecision

composer/datasets/coco.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def __getitem__(self, idx):
+        img_id = self.img_keys[idx]
+        img_data = self.images[img_id]
+        fn = img_data[0]
+        img_path = os.path.join(self.img_folder, fn)
+
+        from os.path import exists
+
+        if (exists(img_path)):
```
Landanjs, line 196: This if-statement seems sus... Are there image paths that don't exist? If that is the case, I think there is a better way to handle this

composer/models/ssd/base_model.py
```diff
@@ -0,0 +1,166 @@
+## Code adapted from https://github.com/mlcommons/training/tree/master/single_stage_detector/ssd
+
+from collections import OrderedDict
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.autograd import Variable
```
Landanjs, line 8: Should remove this

composer/models/ssd/base_model.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def forward(self, data):
+        layer1_activation = self.layer1(data)
+        x = layer1_activation
+        layer2_activation = self.layer2(x)
+
+        return [layer2_activation]
+
+
+class L2Norm(nn.Module):
```
Landanjs, line 81: I might have missed it, where is this used?

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,173 @@
...
+from composer.models.ssd.utils import DefaultBoxes, Encoder, SSDTransformer
+
+
+class SSD(ComposerModel):
+
+    def __init__(self, hparams: SSDHparams) -> None:
+        super().__init__()
+
+        self.hparams = hparams
+        ln = COCODetection.labelnum
```
Landanjs, line 32: Should this be used by SSD300?

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,173 @@
...
+        self.MAP = MAP()
+
+    def loss(self, outputs: Any, batch: BatchPair) -> Tensors:
+
+        (img, img_id, img_size, bbox, label) = batch
+        trans_bbox = bbox.transpose(1, 2).contiguous()
+
+        ploc, plabel = outputs
+        gloc, glabel = transbbox, label
+        '''
```
Landanjs, line 47: Should delete

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,173 @@
...
+        val_coco = COCODetection(val_coco_root, val_annotate, val_trans)
+
+        inv_map = {v: k for k, v in val_coco.label_map.items()}
+
+        ret = []
+
+        overlap_threshold = 0.50
+        nms_max_detections = 200
+
+        dboxes = dboxes300_coco()
```
Landanjs, line 85: Duplicate of above call, should delete

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,173 @@
...
+        cat = ann['category_id']
+
+        t.append(dict(boxes=torch.Tensor([[x_topleft, y_topleft, bbox_width, bbox_height]]),
+                      labels=torch.Tensor([cat])))
+
+    #
+    return t
+
+
+def dboxes300_coco():
```
Landanjs, line 155: In addition to Austin's comment, is this the same as the dboxes300_coco in the dataset code? If so, delete this duplicate, but apply Austin's comment to the other code.

composer/models/ssd/ssd.py
```diff
@@ -0,0 +1,173 @@
...
+        val_trans = SSDTransformer(dboxes, (input_size, input_size), val=True)
+        data = "/localdisk/coco"
+
+        val_annotate = os.path.join(data, "annotations/instances_val2017.json")
+        val_coco_root = os.path.join(data, "val2017")
+        train_annotate = os.path.join(data, "annotations/instances_train2017.json")
+        train_coco_root = os.path.join(data, "train2017")
+        val_coco = COCODetection(val_coco_root, val_annotate, val_trans)
+
+        inv_map = {v: k for k, v in val_coco.label_map.items()}
```
Landanjs, line 78: I feel like there has to be a better way to get the proper labels than to create a `Dataset` object within the `validate()` function. I think the `__getitem__` function could return whatever value is being extracted from this conversion. We can discuss more in person

composer/datasets/coco.py
```diff
@@ -0,0 +1,597 @@
...
+                    self.transform(img, (htot, wtot), bbox_sizes, bbox_labels)
+            else:
+                pass
+            return img, img_id, (htot, wtot), bbox_sizes, bbox_labels
+        else:
+            print('here')
+            pass
+
+
+class COCO:
```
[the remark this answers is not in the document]
Landanjs: Wait, `pycocotools` is being used below, could that also be used here?

Landanjs (commented):
Wow this is a lot! Like Austin said, it is very impressive you were able to get this put together.

That being said, the code is very complex/ambiguous and a lot of clean up needs to be done given that the code comes from so many places. @A-Jacobson @florescl do y'all have thoughts on if it would be faster and cleaner to import detection models from another library? I'm think we should do a review of what is available.

I was only able to make an initial pass and will do another pass after some comments are addressed if we continue this route

A-Jacobson:
> Wow this is a lot! Like Austin said, it is very impressive you were able to get this put together. I wanted to make an initial pass over a few documents and will do another pass within the next few days.
> 
> 
> 
> @florescl How is the performance as of now? Depending on current performance, do you think it would be easier to import a model from another library? I'm wondering if this would be quicker and cleaner than trying to get the MLperf model to work well.
> 
> 
> 
> 

I looked at torchvision, detectron2, and mmdet when I was working on the first version of my kaggle notebook. 

main blockers I noticed: 
- the standard batch format for coco is a list of dicts which isn't supported by composer yet.
- the models are structured to take images/labels and return a dict of losses in train mode and image -> prediction in eval mode. Makes wrapping them in ComposerModel a bit tricky.

Landanjs:
For 1, I think dictionary outputs from dataloaders is typically handled by `split_batch` functions. Laura has one for the current COCO dataset, so it should be modifiable for other datasets (but we can hold off on that consideration for now)

For 2, HuggingFace models also output a dictionary of logits and loss values, so there is some precedent to handle this. I would expect the detection models to be more complicated, but should still be doable. If anything, it would motivate making the training loop more flexible for arbitrary model outputs or help us define restrictions on what can be returned.

composer/datasets/coco.py
```diff
@@ -0,0 +1,597 @@
...
+                [b[i::nm] for i in range(nm)],
+                [c[i::nm] for i in range(nm)],
+            ))
+
+
+def _isArrayLike(obj):
+    return hasattr(obj, '__iter__') and hasattr(obj, '__len__')
+
+
+class COCODetection(data.Dataset):
```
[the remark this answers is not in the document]
Landanjs: Removing this for now to focus on current implementation, but something to consider for the future

florescl:
closing and reopening another PR with all the fixes and all your comments taken into account.

PR #147, opened 2021-12-10
Apply channels last on init

ravi-mosaicml:
Channels last should be applied BEFORE ddp. DDP wrapping of the module happens after init but before training start.


ajaysaini725 (approved):
LGTM - maybe include a quick comment about why channels_last needs to happen on `INIT` instead of `TRAINING_START` for other people's future reference.

PR #150, opened 2021-12-13
Update Trainer checkpointing documentation

moinnadeem:
Batches is not a validation option, we moved to iterations.


ravi-mosaicml (approved):
LGTM

PR #151, opened 2021-12-13
Address crashes with DDP + Checkpointing

### LINKED ISSUE #149: Checkpointing fails when using the composer command
## Environment 
Running on colo on branch `moin/bert` rebased off of the latest dev: commit hash `be411d3f71feffcbab84da8671dd530fdbbc2f30`.

## To reproduce

1. Checkout `moin/bert` branch
2. Run `composer -n 8 examples/run_mosaic_trainer.py -f composer/yamls/models/glue/sst-2_checkpointing.yaml`

## Expected behavior

If I run `python examples/run_mosaic_trainer.py -f composer/yamls/models/glue/sst-2_checkpointing.yaml` instead of `composer -n 8`, then checkpointing works as expected, and we are able to save a checkpoint successfully and continue training. However, when I run it with `composer -n 8`, we see the following crash (exactly when checkpointing happens): 

```
[BATCH][step=599]: { "lr-AdamW/group0": 0.0000, }
Error in subprocess
----------Subprocess STDOUT----------

----------Subprocess STDERR----------

Command '['/usr/bin/python3.8', '-u', 'examples/run_mosaic_trainer.py', '-f', 'composer/yamls/models/glue/sst-2_checkpointing.yaml']' returned non-zero exit status 1.
Killing subprocess 33667 with SIGTERM
Killing subprocess 33662 with SIGTERM
Killing subprocess 33661 with SIGTERM
Killing subprocess 33669 with SIGTERM
Killing subprocess 33659 with SIGTERM
Waiting 30 seconds for processes to terminate...
```

For reference, checkpointing is set to every 600 iterations. There is no printed output.

## Additional context
This is the next step in getting BERT pre-training working (saving the pre-trained model, and then fine-tuning it on GLUE tasks).

moinnadeem:
Resolves #149 .

The issue was that checkpointing stored `state.state_dict()` before checking if the process was Rank 0. Certain loggers (such as WandB) are only stored on Rank 0 processes, so it would cause a crash on other ranks since the loggers don't exist.

I have verified this works.

Some points of discussion:

- [ ] Should we update checkpointing tests to include DDP?

### DIFF
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -119,7 +119,6 @@ class Checkpointer:
         # to all processes on checkpoint resume.
         # This will be fixed by: https://github.com/mosaicml/composer/issues/12
         state_dict = {
-            'state': state.state_dict(),  # should be the same across all ranks. per-rank state not stored
             'rng': self._get_rng_state(device=device),  # stored across all ranks
             'seed': seed,
         }
@@ -127,6 +126,9 @@ class Checkpointer:
             # only rank 0 saves checkpoints
             # Need the check down here so all the DDP syncs will work for generating the checkpoint
             return
+
+        # we add the state only on rank 0 since other processes don't have loggers to serialize
+        state_dict['state'] = state.state_dict()  # should be the same across all ranks. per-rank state not stored
 
         if config:
             hparams_path = os.path.join(self.checkpoint_folder, "hparams.yaml")


ravi-mosaicml (approved):
LGTM, thanks! FYI checkpointing tests do run with ddp, though not with any rank zero callbacks enabled.

PR #152, opened 2021-12-13
Sudo in the dockerimage

ravi-mosaicml:
Installed and configured sudo in the docker image. Makes it possible to install apt packages in the container that are not included in the image.


bandish-shah (approved):
LGTM

PR #153, opened 2021-12-15
Removed dataclass from state

ravi-mosaicml:
Making state a dataclass was more hassle than it was worth, since dataclasses do not easily allow getters to have a different type signature than setters and the init. Instead, switched to using a plain old python class.

This change allowed for the following cleanup:

* Replaced `train_batch_size` and `eval_batch_size` with getters that read from the dataloader
* Added getters for `state.optimizers` and `state.schedulers` which return a list, regardless of the type passed to the setters or upon init
* Removed epoch, step, loss, batch, and outputs from the init args, as they would be updated by the trainer (and would not be set on init)
* Added setters for `optimizers`, `schedulers`, `callbacks`, and `algorithms` that ensure in-place operations, so references to previous state objects remain valid
* Cleaned up the precision context so it need not be called before being entered / exited.

This cleanup will help with binding trainer init arguments to the trainer via getters / setters, which will allow one to easily change trainer properties when doing interactive development.


Averylamp (approved):
Lgtm

PR #154, opened 2021-12-15
Bind trainer `__init__` arguments as properites/attributes

ravi-mosaicml:
Most arguments that could be passed to the `__init__` of the trainer are now bound to the trainer either as raw attributes or properties. This will enable changing the trainer behavior when doing interactive development.


composer/trainer/checkpoint.py
```diff
@@ -71,22 +72,32 @@ class Checkpointer:
...
+    @property
+    def _save_event(self):
+        if self.checkpoint_interval_unit is None:
+            raise RuntimeError("Checkpointing is diabled")
+        if self.checkpoint_interval_unit.lower() == "ep":
+            return Event.EPOCH_END
+        elif self.checkpoint_interval_unit.lower() == "it":
+            return Event.BATCH_END
+        else:
+            raise RuntimeError(f"Unknown checkpointing interval: {self.checkpoint_interval_unit}")
```
siriuslee, line 99: It's a little strange to move this event outside of the init. This seems like it could delay the error by quite a bit when it could have been checked already.

composer/trainer/trainer.py
```diff
@@ -403,13 +393,187 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
...
+        self.state.callbacks = [*self.logger.backends, *callbacks]
+
+    @property
+    def optimizers(self):
+        """Optimizers"""
+        return self.state.optimizers
+
+    @optimizers.setter
+    def optimizers(self, optimizers: Optimizers):
+        self.state.optimizers = optimizers
```
siriuslee, line 463: should this call `ensure_tuple`?

composer/trainer/trainer.py
```diff
@@ -403,13 +393,187 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
...
+
+    @property
+    def schedulers(self):
+        """Schedulers"""
+        return self.state.schedulers
+
+    @schedulers.setter
+    def schedulers(self, schedulers: Schedulers):
+        self.state.schedulers = schedulers
```
siriuslee, line 472: should this call `ensure_tuple`?

composer/trainer/trainer.py
```diff
@@ -403,13 +393,187 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
...
+
+    @property
+    def device(self):
+        """Device"""
+        # No setter for device since it wouldn't make sense to change it
```
siriuslee, line 477: Might be worth raising an error on setting with a pointer that the user should re-initialize the trainer if they want to switch devices.

siriuslee (commented):
Some small comments. There was a lot in here that didn't have anything to do with the getting and setting. From what I could tell, they looked sensible and fine, but I'll defer to folks that might be more aware of the reasons for those changes. The init stuff mostly looks good to me, though.

composer/trainer/trainer.py
ravi-mosaicml: It is no longer strictly required, as #153 adds a custom setter for `state.optimizers`. This setter does an in-place replacement of `state.optimizers` with whatever is passed in. (This ensures that anything that hard-coded a reference to `state.optimizers` will always have the correct list of optimizers, and not some old, dangling version anymore). Same thing with the `schedulers` below.

composer/trainer/trainer.py
ravi-mosaicml: Similar to https://github.com/mosaicml/composer/pull/154#discussion_r770734434

composer/trainer/checkpoint.py
ravi-mosaicml: This change allows you to change the checkpointing parameters mid-training. Good point though, I'll do a custom setter for checkpoint_interval_unit that performs this validation.

composer/trainer/trainer.py
```diff
@@ -279,27 +267,29 @@ def __init__(
...
+            raise NotImplementedError("Multiple optimizers are not supported.")
         schedulers = [
-            x.initialize_object(optimizer, self.state.steps_per_epoch) for x in ensure_warmup_last(schedulers_hparams)
+            x.initialize_object(optimizers[0], self.state.steps_per_epoch)
+            for x in ensure_warmup_last(list(ensure_tuple(schedulers_hparams)))
         ]
-        self.state.optimizers = optimizer
-        self.state.schedulers = ComposedScheduler(schedulers=schedulers)
+        self.state.optimizers = optimizers
+        self.state.schedulers = [ComposedScheduler(schedulers=schedulers)]
```
Averylamp, line 285: Not sure, but the Composed Scheduler casting maybe could be the underlying schedulers object?  A little opinionated, but can auto set ComposedScheduler in the setter

composer/trainer/trainer.py
```diff
@@ -403,13 +393,198 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
...
+    @property
+    def train_dataloader(self) -> Union[DataLoader, DataloaderSpec]:
+        """The train dataloader"""
+        if self._train_split_fn is not None and self._train_device_transformation_fn is not None:
+            return DataloaderSpec(self.state.train_dataloader, self._train_device_transformation_fn,
+                                  self._train_split_fn)
+        else:
+            return self.state.train_dataloader
+
+    @train_dataloader.setter
```
Averylamp, line 410: I think possibly it'd be nice to go to a TrainerProperties style separation of all properties for ease of reading all properties like PTL?  maybe separate into a separate file

Averylamp (approved):
lgtm, I think I'd very much prefer breaking out properties into a TrainerProperties interface just to keep the Trainer file a little smaller and separate out the bloat

composer/trainer/trainer.py
```diff
@@ -475,13 +477,20 @@ def device(self):
...
+
     @property
     def grad_accum(self):
         """Gradient Accumulation"""
         return self.state.grad_accum
 
     @grad_accum.setter
     def grad_accum(self, grad_accum: int):
+        if self.deepspeed_enabled:
+            raise RuntimeError("Cannot change grad_accum when using deepspeed")
```
siriuslee, line 493: Could you say what should be done instead? Is this another "If you'd like to change it, create a new Trainer" type of thing?

composer/core/logging/logger.py
```diff
@@ -60,11 +60,14 @@ class Logger:
...
-            backends: Sequence[BaseLoggerBackend] = tuple(),
+        self,
+        state: State,
+        backends: Optional[List[BaseLoggerBackend]] = None,
     ):
-        self.backends = backends
+        if backends is None:
+            self.backends = []
+        else:
+            self.backends = backends
```
hanlint, line 70:
```suggestion
        self.backends = [] if backends is None else backends
```

composer/trainer/trainer.py
```diff
@@ -187,38 +191,26 @@ def __init__(
...
         else:
             ddp.initialize_ddp(device.ddp_backend, datetime.timedelta(seconds=ddp_timeout))
-            if ddp_sync_strategy is None:
-                self.ddp_sync_strategy = ddp.DDPSyncStrategy.SINGLE_AUTO_SYNC if not find_unused_parameters else ddp.DDPSyncStrategy.FORCED_SYNC
-            else:
-                self.ddp_sync_strategy = ddp.DDPSyncStrategy(ddp_sync_strategy)
 
         if isinstance(train_dataloader, DataloaderSpec):
-            train_dataloader_spec = train_dataloader
+            self._train_device_transformation_fn = train_dataloader.device_transform_fn
```
hanlint, line 201: These changes don't seem related to binding the trainer arguments as properties?

I also find the previous logic to be more readable, and now the handling of train_dataloader and eval_dataloader are not consistent.

composer/trainer/trainer.py
```diff
@@ -233,20 +225,12 @@ def __init__(
...
-        if train_subset_num_batches is not None:
-            if train_subset_num_batches > self.state.steps_per_epoch:
-                warnings.warn(
-                    textwrap.dedent(
-                        f"""SubsetNumBatchesWarning: The train_subset_num_batches({train_subset_num_batches})
-                        is greater than the number of batches in the training dataloader
-                        ({self.state.steps_per_epoch})"""))
-            else:
-                self.state.steps_per_epoch = train_subset_num_batches
+        self.state.steps_per_epoch = train_subset_num_batches
```
hanlint, line 233: curious to understand why train_subset and eval_subset are now handled differently?

composer/trainer/trainer.py
```diff
@@ -261,13 +245,17 @@ def __init__(
...
         self.logger = Logger(self.state, log_destinations)
         self.state.callbacks = [*log_destinations, *callbacks]
-
         self.engine = Engine(self.state, self.state.algorithms, self.logger, self.state.callbacks)
 
         self.validate_every_n_batches = validate_every_n_batches
         self.validate_every_n_epochs = validate_every_n_epochs
         self.compute_training_metrics = compute_training_metrics
-        self.grad_clip_norm = grad_clip_norm
+        self._grad_clip_norm = grad_clip_norm
```
hanlint, line 253: since this is not a property of state, we can not make this a getter/private attribute

composer/trainer/trainer.py
hanlint: As discussed, we should remove all the setters here

composer/trainer/trainer.py
```diff
@@ -592,31 +762,37 @@ def _ddp_reduce_tensor_sum(tensor: Tensor) -> Tensor:
...
 
                     state.last_batch_size = self._get_batch_size(state.batch)
                     state.batch = self.device.batch_to_device(state.batch)
                     if self._train_device_transformation_fn is not None:
                         state.batch = self._train_device_transformation_fn(state.batch)
 
+                    if self._train_split_fn is None:
+                        split_fn = default_batch_split_fn
+                    else:
+                        split_fn = self._train_split_fn
```
hanlint, line 778:
```suggestion
                    split_fn = default_batch_split_fn if self._train_split_fn is None else self._train_split_fn
```

or.. put this logic closer to where the `self._train_split_fn`  is defined?

composer/trainer/trainer.py
```diff
@@ -279,27 +267,29 @@ def __init__(
...
         # TODO(#121): get checkpointing working with DeepSpeed.
-        if checkpoint_folder and checkpoint_interval and checkpoint_interval_unit:
-            if self.deepspeed_enabled:
-                raise NotImplementedError("Checkpointing is not yet supported with DeepSpeed.")
-            self.checkpointer = Checkpointer(checkpoint_folder=get_relative_to_run_directory(checkpoint_folder),
-                                             checkpoint_interval=checkpoint_interval,
-                                             checkpoint_interval_unit=checkpoint_interval_unit)
+        if checkpoint_interval_unit is not None and self.deepspeed_enabled:
+            raise NotImplementedError("Checkpointing is not yet supported with DeepSpeed.")
+        self._checkpointer = Checkpointer(checkpoint_folder=checkpoint_folder,
```
hanlint, line 290: this need not be private

hanlint (changes requested):

ravi-mosaicml:
Closing this PR as we need to revisit multiple calls to .fit

Issue #138, opened 2021-12-07
Multiple calls to .fit


ravi-mosaicml:
The trainer should support multiple calls to .fit.

Composer is going with the convention of "one run" = "one instance of the Trainer". So, if you want to do this, create a new trainer for each run:
* Pre training and fine tuning
* Sweeps across parameters

Nonetheless, there are valid reasons for calling `.fit()` multiple times, for example:
* When doing interactive development in developing an algorithm, model, etc...
* When you want to change trainer properties in the middle of a run (outside of an algorithm)

To support this, we will allow `.fit()` to _optionally_ take a `training_duration` parameter. If specified, then `.fit` will train for this much time. `.fit()` can be called multiple times, and each time it will train for the specified duration. If the duration not specified, then it will train for `max_epochs`. The trainer will never train _beyond_ `max_epochs`.

To support changing trainer behavior, almost all attributes that are specified upon `__init__` will be bound to the trainer as attributes or properties with proper getters and setters. However, when manually updating attributes in the middle of a `.fit`, then the burden is on the user to make sure that changed attributes are in the correct state (e.g. adding a callback halfway through? make sure that you called `callback.run_event(Event.INIT)` before calling `.fit(training duration)` again).

In pseudocode:
```python3
class Trainer:
    def __init__(model, train_dataloader, max_epochs): ...
         self.state = State(model, train_dataloader, max_epochs)
 
    @property
    def train_dataloader(self):
        return self.state.train_dataloader

    @train_dataloader.setter
    def train_dataloader(self, train_dataloader):
        self.state.train_dataloader = train_dataloader

    def fit(self, duration = None):
        if duration is None:
            # train to end
            ...
        else:
            # train for duration
            ...
```

Todos:
- [ ] Merge #154 (which depends on #153)
- [ ] Add support for `.fit(training_duration)`

siriuslee:
I can see at least two primary ways in which multiple `.fit()` calls would be useful:
1. Interactive development where you want to see how the model develops. In this case, something like `trainer.fit(epochs=1)` would be useful to fit for one extra epoch and then the user can evaluate their model again.
2. Training curricula like pre-training followed by fine-tuning. In this case the user might switch a lot of hyperparameters (e.g. dataset, loss function, optimizer)

With that in mind:
> What hyperparemtets can be changed between calls to .fit()?
Could be quite a lot, but those should mostly be left to the user, in my opinion. They could do things like change dataloaders, loss functions or even small model modifications between calls. 

[...]

Issue #146, opened 2021-12-10
Time Abstraction


ravi-mosaicml:
## Motivation

There are various measures of time during training, and we need a common steppable abstraction to handle conversion between units. In the CV community, it is common to track time in terms of samples in batches. However, in NLP, it is more common to track time in terms of tokens and the duration of the training process. Here, we propose a time tracking solution.

## Implementation

After discussion with @abhi-mosaic  and @moinnadeem, we are leaning towards the following design:

- [x] 1. `Time` objects will simplify time arithmatic. A `Time` object consists of an integer and a unit, which will be one of `epochs`, `batches`, `samples`, or `tokens`. Via overloaded functions, `Time` objects will support comparisons, addition, and subtraction against other Time objects of same units, and for backwards compatibility, raw integers (though, in that case, a `UserWarning` will be emitted). They will also have getters to get the underlying value as an integer and unit.

- [x] 2. A `Timer` object, attached to the trainer's state, will track `epochs`, `batches`, `samples`, and `tokens`. Types are `Time` objects, except for tokens which may be `None` (for non-NLP training jobs). The timer object will have getters for each of these fields and a single update function that the training loop will call to update the timer at the end of every batch -- e.g. `timer.update(samples=X, token=Y)`.

- [x] 3. To determine the number of samples and number of tokens, a dataset can provide `get_batch_size(batch)` and `get_num_tokens(batch)`. If not specified, the default `get_batch_size()` will be used, and tokens will NOT be tracked.

- [x] 4. Datasets can optionally provide `__len__` and `get_num_tokens()`. By pytorch convention, `__len__` should be the number of samples in the dataset. `get_num_tokens` can either return a constant number, perform some sort of computation upon initialization to determine the number of tokens in the dataset, or (by default) return None if the number of tokens is unknown.

- [x] 5. The `max_epochs` property in the trainer hparams will be replaced with `max_duration`, where duration can be specified in terms of `epochs`, `steps`, `samples`, or `tokens`.

- [x] 6. The trainer will have a function `trainer.get_elapsed_duration()` that will query the timer object and return a float on `[0,1]` representing how much of the training process has been completed (relative to the `max_duration` parameter).

- [x] 7. The timing module (NOT the timer object) will have a static method like:
    ```python3
    convert(time_string, desired_unit, dataset_num_samples: Optional[int] = None, dataset_num_tokens: Optional[int] = None, max_training_duration: Optional[str] = None, batch_size: Optional[int] = None):
        pass
[...]

PR #161, opened 2021-12-15
Multiple calls to `.fit`

ravi-mosaicml:
Support multiple calls to .fit for partial training, as discussed in #138. Specifically, this PR:
* Moves most initialization logic to `Trainer.__init__`, so it is invoked only once
* Added `num_epochs` and `num_batches` as optional parameters to `Trainer.fit`
* Significant refactor of the training loop to support the above
* Added test cases (and updated the synthetic dataloader to use a deterministic generator, so the model from the tests would be identical)

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -228,6 +228,7 @@ class Trainer:
             train_dataloader=DDPDataLoader(train_dataloader),
             eval_dataloader=DDPDataLoader(eval_dataloader_spec.dataloader),
         )
+        self.state.train_metrics = self._get_metrics_as_collection(is_train=True)
 
         # Steps per epoch
         self.state.steps_per_epoch = train_subset_num_batches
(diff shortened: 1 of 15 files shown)


composer/trainer/trainer.py
```diff
@@ -565,12 +635,89 @@ def eval_subset_num_batches(self):
     def eval_subset_num_batches(self, eval_subset_num_batches: Optional[int] = None):
         self._eval_subset_num_batches = eval_subset_num_batches
 
-    def fit(self):
-        """Train and evaluate the model on the provided data."""
+    def fit(self, num_batches: Optional[int] = None, num_epochs: Optional[int] = None):
```
A-Jacobson, line 638: I like that we're moving some params to fit, but I'm conflicted about `num_batches`. It doesn't tell me anything about the amount of information I'm feeding to the network. An `epochs` is useful as it represent a full pass through the dataset, `train_fraction` (training on a subset of the dataset) could also be useful for debugging, learning curves, and active learning situations. I know large NLP models like tokens so perhaps @moinnadeem some opinions about this?

ravi-mosaicml: `num_batches` and `num_epochs` will be replaced with a `duration` argument once #146 is implemented. That will support tokens and a fraction of the total training duration.

ravi-mosaicml: (I'd prefer not to add that logic here as #146 is next on my agenda)

ravi-mosaicml:
Closing this PR as we need to revisit multiple calls to .fit

PR #166, opened 2021-12-16
DeepSpeed ZeRO config options

jbloxham:
Exposes configuration options for ZeRO, including optimizer offloading and parameter offloading. There's a lot more options that could be exposed, but through testing I've found these to be the most consistently useful. As a bonus, also includes gradient checkpointing for GPT models.

ZeRO 3 is available, but intentionally under supported at the moment and also for the immediate future. There's an open question around how model initialization will need to change in order to support very large models without overwhelming the CPU. Model checkpointing is also idiosyncratic with ZeRO 3, and I'm not planning to deal with that any time soon. We estimate we can get up to a 13B model with just ZeRO 2.

Here's a sweep demonstrating the impacts of various configurations: https://wandb.ai/mosaic-ml/deepspeed-more-testing?workspace=user-jbloxham.


composer/trainer/trainer.py
```diff
@@ -538,7 +527,9 @@ def _train_loop(self) -> None:
             log.warn('Computing model evaluation metrics during training.'
                      ' This doubles the number of forward passes and may lead'
                      ' to a throughput degradation.')
-        train_metrics = self._get_metrics_as_collection(is_train=True)
+            train_metrics = self._get_metrics_as_collection(is_train=True)
```
jbloxham, line 530: This isn't explicitly required for DeepSpeed, but it was a bug, right?

Averylamp (approved):
lgtm

hanlint (approved):

PR #169, opened 2021-12-16
Add the ability to load a checkpoint without restoring state

moinnadeem:
# Motivation
This pull request enables the ability to load a checkpoint without loading the associated state. It does so via the following YAML changes:
1. Introduces a `CheckpointLoaderHparams` object, with three fields: `checkpoint_filepath`, `load_weights_only`, and `strict`. 
2. If `load_weights_only = False`, then nothing changes and the previous codepaths are used. The `strict` value isn't considered if `load_weights_only = False`, since restoring a checkpoint with state should ensure that the model exactly matches up. YAML validation ensures that the `strict` value cannot be set without `load_weights_only = True`.
3. If `load_weights_only = True`, then it loads the checkpoint and avoids recovering the state via a new codepath. If `strict = False`, it also prints the keys that did not match up for user safety.

It creates the `CheckpointLoader` object when the `Trainer` is created via `create_from_hparams`, and passes the CheckpointLoader in as well.

# Discussion Points
1. Are we happy with the YAML API change?
2. Should we add any tests for these new codepaths?

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -221,10 +221,10 @@ class State(Serializable):
             state_dict (types.StateDict): object returned from call to :meth:`state_dict`.
             strict (bool): whether the keys in the state_dict should perfectly match the keys in the model.
         """
-        if state_dict['state']["_is_model_ddp_wrapped"] and not isinstance(self.model, DistributedDataParallel):
-            torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(state_dict['state']['model'], "module.")
+        if state_dict["_is_model_ddp_wrapped"] and not isinstance(self.model, DistributedDataParallel):
+            torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(state_dict['model'], "module.")
 
-            missing_keys, unexpected_keys = self.model.load_state_dict(state_dict['state']['model'], strict=strict)
+            missing_keys, unexpected_keys = self.model.load_state_dict(state_dict['model'], strict=strict)
             if len(missing_keys) > 0:
                 logger.warning(f"Found these missing keys in the checkpoint: {', '.join(missing_keys)}")
             if len(unexpected_keys) > 0:
@@ -246,7 +246,7 @@ class State(Serializable):
                 serialized_value = state[state_field_name]
 
                 if state_field_name == "model":
-                    self.load_model_state(serialized_value, strict=strict)
+                    self.load_model_state(state, strict=strict)
                 else:
                     for target in ensure_tuple(state_field_value):
                         if target is None:
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -213,7 +213,24 @@ class State(Serializable):
         state_dict["_is_model_ddp_wrapped"] = isinstance(self.model, DistributedDataParallel)
         return state_dict
 
-    def load_state_dict(self, state: types.StateDict):
+    def load_model_state(self, state_dict: types.StateDict, strict: bool):
+        """
+        Loads the model's state from a state_dict.
+
+        Args:
+            state_dict (types.StateDict): object returned from call to :meth:`state_dict`.
+            strict (bool): whether the keys in the state_dict should perfectly match the keys in the model.
+        """
+        if state_dict['state']["_is_model_ddp_wrapped"] and not isinstance(self.model, DistributedDataParallel):
+            torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(state_dict['state']['model'], "module.")
+
+            missing_keys, unexpected_keys = self.model.load_state_dict(state_dict['state']['model'], strict=strict)
+            if len(missing_keys) > 0:
+                logger.warning(f"Found these missing keys in the checkpoint: {', '.join(missing_keys)}")
+            if len(unexpected_keys) > 0:
+                logger.warning(f"Found these unexpected keys in the checkpoint: {', '.join(unexpected_keys)}")
+
+    def load_state_dict(self, state: types.StateDict, strict: bool = False):
         """Loads the state.
 
         Args:
@@ -229,9 +246,7 @@ class State(Serializable):
                 serialized_value = state[state_field_name]
 
                 if state_field_name == "model":
-                    if state["_is_model_ddp_wrapped"] and not isinstance(self.model, DistributedDataParallel):
-                        torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(serialized_value, "module.")
-                    state_field_value.load_state_dict(serialized_value)
+                    self.load_model_state(serialized_value, strict=strict)
                 else:
                     for target in ensure_tuple(state_field_value):
                         if target is None:
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -43,7 +43,7 @@ class CheckpointLoader:
         """
 
         if self.load_weights_only:
-            state.load_model_state(self.state_dict, strict=self.strict_model_weights)
+            state.load_model_state(self.state_dict['state'], strict=self.strict_model_weights)
         else:
             state.load_state_dict(self.state_dict["state"])
         self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -17,7 +17,6 @@ from composer.utils import ddp, seed_all
 
 log = logging.getLogger(__name__)
 
-
 class CheckpointLoader:
     """Manager for initializing state and restoring RNG state from existing checkpoints.
 
@@ -25,8 +24,10 @@ class CheckpointLoader:
         checkpoint_filepath (str): The path to an existing checkpoint file.
     """
 
-    def __init__(self, checkpoint_filepath: str):
+    def __init__(self, checkpoint_filepath: str, load_weights_only: bool = False, strict: bool = False):
         self.state_dict = torch.load(checkpoint_filepath, map_location='cpu')
+        self.load_weights_only = load_weights_only
+        self.strict = strict
 
     def load_checkpoint(self, state: State):
         """Initialize state from the loaded checkpoint's data.
@@ -38,8 +39,16 @@ class CheckpointLoader:
             The seed that was loaded from the checkpoint if it exists otherwise `None`.
         """
 
-        state.load_state_dict(self.state_dict["state"])
-        self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
+        if self.load_weights_only:
+            missing_keys, unexpected_keys = state.model.load_state_dict(self.state_dict['state']['model'],
+                                                                        strict=self.strict)
+            if len(missing_keys) > 0:
+                log.warning(f"Found these missing keys in the checkpoint: {', '.join(missing_keys)}")
+            if len(unexpected_keys) > 0:
+                log.warning(f"Found these unexpected keys in the checkpoint: {', '.join(unexpected_keys)}")
+        else:
+            state.load_state_dict(self.state_dict["state"])
+            self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
 
         if "seed" in self.state_dict:
             world_size = ddp.get_world_size()
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -46,18 +46,18 @@ class CheckpointLoader:
             state.load_model_state(self.state_dict['state'], strict=self.strict_model_weights)
         else:
             state.load_state_dict(self.state_dict["state"])
-        self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
+            self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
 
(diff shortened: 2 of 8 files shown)


composer/trainer/checkpoint_hparams.py
```diff
@@ -0,0 +1,40 @@
...
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class CheckpointLoaderHparams(hp.Hparams):
+    """Hparams for the :class:`CheckpointLoader`.
+
+    See the documentation for the :class:`CheckpointLoader`.
+    """
+    checkpoint_filepath: Optional[str] = hp.optional(doc="Path to the serialized state_dict to recover state from.",
```
ravi-mosaicml, line 28: I think this should be required. The `CheckpointLoaderHparams` in the `TrainerHparms` can be none.

composer/trainer/checkpoint_hparams.py
```diff
@@ -0,0 +1,40 @@
...
+                                                    default=False)
+    strict: Optional[bool] = hp.optional(
+        doc="Ensure that the set of weights in the checkpoint and model must exactly match.", default=True)
+
+    def initialize_object(self) -> CheckpointLoader:
+        if self.checkpoint_filepath:
+            return CheckpointLoader(checkpoint_filepath=self.checkpoint_filepath,
+                                    load_weights_only=self.load_weights_only,
+                                    strict=self.strict)
+        return None
```
ravi-mosaicml, line 40: This `return None` is probably causing a pyright bug.

composer/trainer/checkpoint_hparams.py
moinnadeem: Agreed, changing this to be required.

composer/trainer/checkpoint.py
```diff
@@ -25,14 +26,29 @@ class CheckpointLoader:
...
+        self.strict = strict
 
     def load_checkpoint(self, state: State):
         """Initialize state from the loaded checkpoint's data.
         """
 
-        state.load_state_dict(self.state_dict["state"])
+        if self.load_weights_only:
+
+            if self.state_dict['state']["_is_model_ddp_wrapped"] and not isinstance(state.model,
```
ravi-mosaicml, line 40: Can you refactor some of this logic out of state.load_state_dict so it's not duplicated? Perhaps create a method on the state class called `def load_model_state(self, state_dict: Dict[str, Any], strict: bool)`. Then the `load_state_dict` and state and this logic can call that.

composer/trainer/checkpoint.py
```diff
@@ -25,14 +26,29 @@ class CheckpointLoader:
         checkpoint_filepath (str): The path to an existing checkpoint file.
     """
 
-    def __init__(self, checkpoint_filepath: str):
+    def __init__(self, checkpoint_filepath: str, load_weights_only: bool = False, strict: bool = False):
         self.state_dict = torch.load(checkpoint_filepath, map_location='cpu')
+        self.load_weights_only = load_weights_only
+        self.strict = strict
```
ravi-mosaicml, line 32: How would the strict parameter be used if `load_weights_only is False`?

composer/trainer/checkpoint.py
```diff
@@ -25,14 +26,29 @@ class CheckpointLoader:
...
 
     def load_checkpoint(self, state: State):
         """Initialize state from the loaded checkpoint's data.
         """
 
-        state.load_state_dict(self.state_dict["state"])
+        if self.load_weights_only:
```
ravi-mosaicml, line 38: To avoid duplicated logic, can you modify `State.load_state_dict` to take a `keys_to_load` parameter?

ravi-mosaicml (commented):
Looks pretty good overall, just some minor changes and comments.

composer/trainer/trainer.py
```diff
@@ -386,7 +384,7 @@ def create_from_hparams(cls, hparams: TrainerHparams) -> Trainer:
             callbacks=tuple(callbacks),
 
             # Checkpointing hparams
-            checkpoint_filepath=hparams.checkpoint_filepath,
+            checkpoint_loader=checkpoint_loader,
```
hanlint, line 387: I find this API a bit confusing .. checkpoint_loader is an hparams file that has a `checkpoint_filepath`, but the root trainer hparams also has `checkpoint_folder`?

For consistency, I would suggest not having a spearate nested hparams object, and keeping all the settings flat.

moinnadeem: Yes, this is because `CheckpointLoader` and `Checkpointer` are currently two different classes, where they should be one. So `checkpoint_folder` is for saving checkpoints, and `checkpoint_filepath` is for loading a checkpoint. 

To be clear, this is the previous API design, but I agree with your comments. 

@ravi-mosaicml thoughts if we should merge the hparams for saving and loading into a unified interface? 

This would broaden the scope of the PR (my goal was to get a blocker for BERT out of the way), but it probably is the correct thing to do.

moinnadeem: > For consistency, I would suggest not having a spearate nested hparams object, and keeping all the settings flat

Do we think this adds too many hparams to the base YAML? I would almost prefer something like:
```
load_checkpoint: 
   filepath: str
   load_weights_only: bool
   strict: bool
save_checkpoint:
   interval_unit: str
   folder: str
   interval: int
```

If we keep them flat, then the `folder` and `filepath` can become confusing, whereas one requires a folder (for saving multiple checkpoints) and the other requires the link to a single file (for loading a single checkpoint). This also doesn't clutter the main Trainer interface.

Additionally, `save_checkpoint.interval` should use the `Time` abstraction when it is ready.

ravi-mosaicml: I would prefer to have a `checkpoint` section in the haparms where you can then have that above yaml structure.

moinnadeem: Yes, that is my thought. If both of you agree, I am happy to make the Trainer required changes. If we do that, can someone take care of addressing tests? I don't want to focus on this PR to the detriment of getting BERT ready in time.

ravi-mosaicml: The save checkpoint won't require any additional tests. For load checkpoint with strict=False or load_weights_only=True, I think it would be trivial to add that in...I would be OK with a functional test (that just makes sure it runs, and doesn't necessarily assert that that warnings are raised or that only the model is loaded).

hanlint: I'm OK with this approach! Yes maybe @ravi-mosaicml if you aren't overloaded can help get this merged?

composer/trainer/checkpoint_hparams.py
moinnadeem: This has been addressed!

composer/trainer/checkpoint_hparams.py
moinnadeem: Taken care of!

composer/trainer/checkpoint.py
moinnadeem: Done!

composer/trainer/checkpoint.py
moinnadeem: I added validation, so `strict` cannot be used if `load_weights_only = False`

composer/trainer/trainer.py
moinnadeem: @hanlint @ravi-mosaicml Awesome, I moved the checkpoint saving features into a `save_checkpoint` hyperparameter class. We now have two main level hparams: `load_checkpoint` and `save_checkpoint`, which have their respective hyperparameters.

moinnadeem:
Cool, addressed all feedback so far!

re: the PyRight issues, they're not problems in practice, should we add a manual ignore? In more detail:
```
 /home/runner/work/composer/composer/composer/trainer/checkpoint.py:47:60 - error: Argument of type "bool | None" cannot be assigned to parameter "strict" of type "bool" in function "load_model_state"
    Type "bool | None" cannot be assigned to type "bool"
      Type "None" cannot be assigned to type "bool" (reportGeneralTypeIssues)
```
The argument can't be None, because the method signature enforces a default. It seems as if PyRight isn't catching onto this?

```
/home/runner/work/composer/composer/composer/trainer/trainer.py
  /home/runner/work/composer/composer/composer/trainer/trainer.py:304:52 - error: "load_checkpoint" is not a known member of "None" (reportOptionalMemberAccess)
```
Before this line, we check if the `checkpoint_loader` is not None`, so `load_checkpoint` can never be run on an object of NoneType. Should we manually ignore this?

ravi-mosaicml:
For the pyright issues where it's complaining about optional variables, you need to add one of these options before the line it's complaining about:

1. `assert x is not None` (if it's an invariant violation)
2. `if x is None: raise ValueError(f"x is None, but it shouldn't be because ..."}` (if it's a user error)

ravi-mosaicml (commented):
Code path looks great! Please a some test to verify it. Thinking something like this:
1. Train, save checkpoint
2. Load checkpoint with a different optimizer and scheduler with `weights_only=True` 
3. Assert that the weights are the same as the first trainer, but that the optimizer is the new one

composer/trainer/checkpoint_hparams.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    filepath: str = hp.required(doc="Path to the serialized state_dict to recover state from.")
+    load_weights_only: Optional[bool] = hp.optional(doc="Whether to only load the weights from the model.",
+                                                    default=False)
+    strict: Optional[bool] = hp.optional(
+        doc="Ensure that the set of weights in the checkpoint and model must exactly match.", default=True)
+
+    def validate(self):
+        if not self.load_weights_only and self.strict:
+            raise ValueError(
```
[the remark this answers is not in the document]
moinnadeem: Ah, good catch! I wrote this code on the plane, so it could have been a little better ;)

composer/trainer/checkpoint_hparams.py
```diff
@@ -0,0 +1,60 @@
...
+@dataclass
+class CheckpointLoaderHparams(hp.Hparams):
+    """Hparams for the :class:`CheckpointLoader`.
+
+    See the documentation for the :class:`CheckpointLoader`.
+    """
+    filepath: str = hp.required(doc="Path to the serialized state_dict to recover state from.")
+    load_weights_only: Optional[bool] = hp.optional(doc="Whether to only load the weights from the model.",
+                                                    default=False)
+    strict: Optional[bool] = hp.optional(
```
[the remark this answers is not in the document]
moinnadeem: @ravi-mosaicml How do you feel about `strict_weights`?

moinnadeem:
Cool, I've addressed all feedback and added the tests that Ravi requested. Hanlin also requested that, instead of passing the Checkpoint{Loader, Saver} object to the Trainer, I pass the hparams directly to make it more BYOT friendly. I agree there, so that has also been reflected.

I've addressed all PyRight issues on my changed files, but it seems as if PyRight is complaining about a few extra files that are outside of the scope of this PR. Namely, `composer/algorithms/augmix/augmix.py`, `composer/algorithms/randaugment/randaugment.py`, and `composer/datasets/brats.py`. What should we do about these?

ravi-mosaicml:
Can you merge in the latest from dev and see if that fixes those files?

moinnadeem:
@ravi-mosaicml Just did -- didn't help for some reason. Any clue why this is happening?

ravi-mosaicml:
Probably a pyright update...can you fix what's it complaining about in
those files if they're small changes?

ravi-mosaicml (approved):
LGTM

composer/core/state.py
```diff
@@ -213,7 +213,24 @@ def state_dict(self) -> types.StateDict:
         state_dict["_is_model_ddp_wrapped"] = isinstance(self.model, DistributedDataParallel)
         return state_dict
 
-    def load_state_dict(self, state: types.StateDict):
+    def load_model_state(self, state_dict: types.StateDict, strict: bool):
```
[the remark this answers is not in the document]
ravi-mosaicml: NVM, this doesn't make sense, since the function is only for the model.

tests/trainer/test_checkpoint.py
```diff
@@ -109,6 +138,100 @@ def inject_stateful_callback_hparams(monkeypatch: MonkeyPatch):
...
+    # load only model weights
+    second_trainer_hparams.load_checkpoint = CheckpointLoaderHparams(filepath=checkpoint_a_file_path,
+                                                                     load_weights_only=True,
+                                                                     strict_model_weights=True)
+    # setup a new optimizer
+    second_trainer_hparams.optimizer = AdamWHparams()
+
+    # setup a new LR scheduler
+    scheduler_options = [ConstantLRHparams(), CosineAnnealingLRHparams(T_max=f"{second_trainer_hparams.max_epochs}ep")]
+    second_trainer_hparams.schedulers = [random.choice(scheduler_options)]
```
jbloxham, line 219: @moinnadeem just ran into this now - it's very strange to use randomness in a test in this way since it can potentially cause flaky tests. Was there a reason you wanted this randomness here?

moinnadeem: Ah, I see the rationale. I just wanted to make sure that we test several schedulers, but didn't think it was worth the time to test all of them. In hindsight, we should either pick the more difficult one, or do both. I agree with you.

jbloxham: I'm modifying this line in a PR for a different purpose anyways, so I'll fix it here.

moinnadeem: Thanks Jamie!

Issue #128, opened 2021-12-02
Investigate and prune any verbose logging


hanlint:
From usability standpoint, we'd want to assess and reduce any logging verbosity, as folks are using our codebase in both (1) CLI environment, and (2) notebook environment.

ravi-mosaicml:
Closing since #170 is merged

PR #170, opened 2021-12-16
Print help on `run_mosaic_trainer.py`, cleaned up verbosity.

### LINKED ISSUE #88: run_mosaic_trainer.py to print help text when invoked without arguments
## 🚀 Feature Request

Instead of getting a ValueError (and stack trace) when running `run_mosaic_trainer.py` without arguments, it might be a bit more friendly to print out the help text from `-h`.

## Motivation

Getting a good first impression (and not feeling accused) as a CLI user is good practice, and we can print out the help text easily.

ravi-mosaicml:
1. Moved the `datadir` onto trainer hparams to avoid an UnusedArgumentWarning
2. Setting the default loglevel to warning
3. Cleaned up warnings to not print a useless extra second line showing the source code of the warning.warn call
4. Cleaned up the launch script to be not-very-verbose by default; but added a -v option for verbosity.
5. If `run_mosaic_trainer.py` is run without args, then the help is printed.

Closes #88, #128

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -1,6 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
 import datetime
+import logging
 import os
 import signal
 import socket
@@ -13,6 +14,8 @@ from argparse import ArgumentParser
 from typing import Any, List, Optional, Set
 
 CLEANUP_TIMEOUT = datetime.timedelta(seconds=30)
+
+log = logging.getLogger(__name__)
 
 
 def get_parser():
@@ -56,6 +59,7 @@ def get_parser():
                         "--module_mode",
                         action="store_true",
                         help="If set, run the training script as a module instead of as a script.")
+    parser.add_argument("-v", "--verbose", action="store_true", help="If set, print verbose messages")
     parser.add_argument("training_script",
                         type=str,
                         help="The path to the training script used to initialize a single training "
@@ -90,7 +94,7 @@ def parse_args():
 def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: str, master_port: Optional[int],
                      module_mode: bool, run_directory: Optional[str], training_script: str,
                      training_script_args: List[Any]) -> Set[subprocess.Popen]:
-    print(f"Starting DDP on local node for global_rank({base_rank}-{base_rank+nproc-1})")
+    log.info("Starting DDP on local node for global_rank(%s-%s)", base_rank, base_rank + nproc - 1)
     processes = []
 
     if run_directory is None:
@@ -104,7 +108,7 @@ def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: s
                       "This may lead to race conditions when launching multiple training processes simultaneously. "
                       "To eliminate this race condition, explicitely specify a port with --master_port PORT_NUMBER")
         master_port = get_free_tcp_port()
-    print(f"DDP Store: tcp://{master_addr}:{master_port}")
+    log.info("DDP Store: tcp://%s:%s", master_addr, master_port)
 
     for local_rank in range(nproc):
         global_rank = base_rank + local_rank
@@ -122,7 +126,7 @@ def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: s
         current_env["MASTER_PORT"] = str(master_port)
         current_env["RUN_DIRECTORY"] = run_directory
 
-        print(f"Launching process for local_rank({local_rank}), global_rank({global_rank})")
+        log.info("Launching process for local_rank(%s), global_rank(%s)", local_rank, global_rank)
 
         if local_rank == 0:
             process = subprocess.Popen(cmd, env=current_env, text=True)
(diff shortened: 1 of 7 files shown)


composer/cli/launcher.py
```diff
@@ -90,7 +94,7 @@ def parse_args():
 def launch_processes(nproc: int, world_size: int, base_rank: int, master_addr: str, master_port: Optional[int],
                      module_mode: bool, run_directory: Optional[str], training_script: str,
                      training_script_args: List[Any]) -> Set[subprocess.Popen]:
-    print(f"Starting DDP on local node for global_rank({base_rank}-{base_rank+nproc-1})")
+    log.info("Starting DDP on local node for global_rank(%s-%s)", base_rank, base_rank + nproc - 1)
```
jbloxham, line 97: Is there a reason to prefer C-style string formatting over f-strings?

composer/cli/launcher.py
```diff
@@ -164,15 +168,20 @@ def monitor_processes(processes: Set[subprocess.Popen]):
...
                     )
-                    error_msg = [
-                        "Error in subprocess",
-                        "----------Subprocess STDOUT----------",
-                        exc.output,
-                        "----------Subprocess STDERR----------",
-                        exc.stderr,
-                    ]
+                    error_msg = [f"Process {process.pid} excited with code {process.returncode}"]
+                    if output is not None:
```
jbloxham, line 172: I have a preference for making it very explicit when stdout and stderr are empty, even if that looks like
```
----------Begin subprocess STDOUT----------
----------End subprocess STDOUT----------
```

Otherwise, sometimes I'm not sure as a user whether the program isn't giving me stdout, or if it's just empty.

composer/cli/launcher.py
```diff
@@ -185,14 +194,14 @@ def cleanup_processes(processes: Set[subprocess.Popen]):
...
-            print(f"Killing subprocess {process.pid} with SIGTERM")
+            log.info("Killing subprocess %s with SIGTERM", process.pid)
             try:
                 os.killpg(process.pid, signal.SIGTERM)
             except ProcessLookupError:
                 pass
 
     current_time = datetime.datetime.now()
-    print(f"Waiting {CLEANUP_TIMEOUT.seconds} seconds for processes to terminate...")
+    print(f"Waiting up to {CLEANUP_TIMEOUT.seconds} seconds for all training processes to terminate...")
```
jbloxham, line 204: I'm still not sure what the cause of your crusade against f-strings is, but you missed one here :wink:

examples/run_mosaic_trainer.py
```diff
@@ -12,33 +12,26 @@
...
+    # From https://stackoverflow.com/questions/26430861/make-pythons-warnings-warn-not-mention-itself
+    return f'{category.__name__}: {message} (source: {filename}:{lineno})\n'
 
 
 def main() -> None:
-    logging.basicConfig()
-    logging.captureWarnings(True)
+    warnings.formatwarning = warning_on_one_line
 
-    parser = argparse.ArgumentParser(parents=[TrainerHparams.get_argparse(cli_args=True)])
```
jbloxham, line 29: This was unnecessary the whole time? 😳

jbloxham (approved):
Looks good - just some questions!

composer/cli/launcher.py
Averylamp: also agree here that I think fstrings are much more understandable and readable

composer/cli/launcher.py
ravi-mosaicml: `log` calls should not use f-strings because they are lazily constructed. But it's fine for print statements.

examples/run_mosaic_trainer.py
ravi-mosaicml: Nah, only after we moved the `datadir` to trainer hparams.

composer/cli/launcher.py
ravi-mosaicml: I agree, but log calls should use %s because they're lazily evaluated. Granted it doesn't make a big difference, but it is a pylint rule (that's next on my agenda to add pylint to the repo...)

Averylamp (approved):
lgtm.  I also have a preference for fstrings instead of old python2 format.  Even if wanting to put args at the end, I think I'd still prefer
`this is a string with: {argname}`.format(argname=item)

composer/cli/launcher.py
ravi-mosaicml: It will be None if it's rank 0, not if it's empty. It should still print if it's empty because `None is not ""`

ravi-mosaicml:
> lgtm. I also have a preference for fstrings instead of old python2 format. Even if wanting to put args at the end, I think I'd still prefer

FYI here is the convention: https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

Averylamp:
> > lgtm. I also have a preference for fstrings instead of old python2 format. Even if wanting to put args at the end, I think I'd still prefer
> 
> FYI here is the convention: https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

I don't think we need to follow this convention although it does save one evaluation call.  %s does lead to bugs when you are printing tuples and items such as that.  Becomes more confusing in terms of readability as well

Issue #12, opened 2021-10-13
Proper seeding for DDP


ajaysaini725:
If the seed is not set in hparams, it is randomly selected in `__init__`. Each DDP process, when it starts up, gets a different random seed.

The seed from the rank 0 process is saved in checkpoints

When resuming from a checkpoint, the seed from the rank 0 process is restored across all DDP processes.
This leads to inconsistent behavior, since the non-rank-0 process now resume with a different seed than they first trained with.

To fix: add the `seed` to the RNG state, and sync across all DDP processes

PR #173, opened 2021-12-17
DDP Seeding Across Processes

ajaysaini725:
This change addresses issue #12 and introduces the following:
- Even when the user provides a seed, each process will have a different seed which is `user_provided_seed + global_rank`. The reason for this is that for algorithms like mixup which modify training batches, we want the modifications to be different on batches on different processes.
- When checkpointing, we store and load seeds using all processes, not just the seed from rank 0.

Test Plan:
- The existing distributed checkpointing tests cover the new code paths introduced here.

### DIFF
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -36,11 +36,15 @@ class CheckpointLoader:
         self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
 
         if "seed" in self.state_dict:
-            assert ddp.get_world_size() == len(self.state_dict["seed"]), \
-                f"invariant violation: if the seed is being restored, then" \
-                "the world size should be the same as in the checkpoint."
-
-            seed_all(self.state_dict["seed"][ddp.get_global_rank()])
+            world_size = ddp.get_world_size()
+            checkpointed_world_size = len(self.state_dict["seed"])
+            if world_size != checkpointed_world_size:
+                warnings.warn(f"Current world size {world_size} does not match the checkpointed world size "
+                              f"{checkpointed_world_size}. The seed will not be restored.")
+                return
+            seed_to_restore = self.state_dict["seed"][ddp.get_global_rank()]
+            seed_all(seed_to_restore)
+            return seed_to_restore
 
     def restore_checkpoint_rng_state(self, state: State, device: Device):
         """Restore the state of all RNG objects in this context from the loaded checkpoint's data.
@@ -112,6 +116,7 @@ class Checkpointer:
 
         Args:
             state (State): The current State of the trainer.
+            seed (int): The seed used for random number generation.
             device (Device): The Device in use by this process.
             ddp (DDP): The DDP engine in use by this trainer.
             config (Optional[Dict[str, Any]]): The hparams used to initialize this trainer, if any.
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -30,6 +30,12 @@ class CheckpointLoader:
 
     def load_checkpoint(self, state: State):
         """Initialize state from the loaded checkpoint's data.
+
+        Args:
+            state (`~composer.core.State`): The state to load the checkpoint into.
+
+        Returns:
+            The seed that was loaded from the checkpoint if it exists otherwise `None`.
         """
 
         state.load_state_dict(self.state_dict["state"])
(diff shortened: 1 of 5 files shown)


composer/trainer/checkpoint.py
```diff
@@ -36,7 +36,11 @@ def load_checkpoint(self, state: State):
         self.checkpoint_rng_state = self._get_checkpoint_rng_state(state, self.state_dict["rng"])
 
         if "seed" in self.state_dict:
-            seed_all(self.state_dict["seed"])
+            assert ddp.get_world_size() == len(self.state_dict["seed"]), \
```
ravi-mosaicml, line 39: I would make this into a warning. It's possible that we're restoring on different hardware

composer/trainer/trainer.py
```diff
@@ -179,10 +179,12 @@ def __init__(
         self.device = device
 
         if not seed:
-            # Set a deterministic seed in the hparams
-            # This seed will be dumped in the hparams that are saved with checkpoints
             seed = get_random_seed()
             log.info(f"Seed was None. Setting seed to random value: {seed}")
+
+        # Assure that each process has a different seed, necessary if a seed is passed to init
+        seed += ddp.get_global_rank()
```
ravi-mosaicml, line 186: When resuming from a checkpoint, would the global rank effectively be added twice (once upon the first training run, and again when resuming?)

ajaysaini725: Nope because when restoring from checkpoint the seed is loaded from the save file in the `load_checkpoint` function and then `seed_all` is called with that seed so this codepath is nullified.

composer/trainer/checkpoint.py
ajaysaini725: The issue here is that you won't be able to restore the seeds correctly if the `world_size` doesn't match. If the new `world_size` is smaller but if it is larger then you won't be able to set seeds for the excess processes because there won't be any stored. 

Also, in general if the new `world_size` isn't the same then there can't be any deterministic guarantees when resuming from checkpoint in the case where a seed is set.

Once thing I can do is if the `world_size` doesn't match, then just log a warning and don't restore any seeds - does that sound good @ravi-mosaicml ?

composer/trainer/trainer.py
```diff
@@ -308,7 +310,10 @@ def __init__(
             if self.deepspeed_enabled:
                 raise NotImplementedError("Checkpointing is not yet supported with DeepSpeed.")
             self.checkpoint_loader = CheckpointLoader(checkpoint_filepath=checkpoint_filepath)
-            self.checkpoint_loader.load_checkpoint(state=self.state)
+            restored_seed = self.checkpoint_loader.load_checkpoint(state=self.state)
+            if restored_seed is not None:
```
ajaysaini725, line 314: This was necessary to handle the case:
- Train a model and save a checkpoint which includes the seed
- Resume from checkpoint and use the saved seed to seed all rng objects
- Save a checkpoint again (in this case `self.seed` is stored in the checkpoint so it must be the correct (i.e. originally checkpointed) seed
- Resume from checkpoint

composer/trainer/checkpoint.py
ravi-mosaicml: Yup, this is fine. I believe the existing behavior is to log a warning and not restore seeds.

composer/trainer/checkpoint.py
```diff
@@ -36,7 +36,15 @@ def load_checkpoint(self, state: State):
...
-            seed_all(self.state_dict["seed"])
+            world_size = ddp.get_world_size()
+            checkpointed_world_size = len(self.state_dict["seed"])
+            if world_size != checkpointed_world_size:
+                warnings.warn(f"Current world size {world_size} does not match the checkpointed world size "
+                              f"{checkpointed_world_size}. The seed will not be restored.")
+                return
+            seed_to_restore = self.state_dict["seed"][ddp.get_global_rank()]
+            seed_all(seed_to_restore)
+            return seed_to_restore
```
ravi-mosaicml, line 47: Can you update the docstring for the function?

composer/trainer/trainer.py
ravi-mosaicml: This is really helpful...I'd add this as a comment in the code

ravi-mosaicml (approved):
LGTM, though would really like to see some tests just verifiying this functionality...I don't think it should be too hard to extend the checkpointer tests with another parameterize.

PR #178, opened 2021-12-18
Updated the `DataSpec` for the timing abstraction (#146) parts 3 and 4

ravi-mosaicml:
For the timing abstraction (#146), the `DataloaderSpec` needed two addition functions -- `get_num_samples_in_batch` and `get_num_tokens_in_batch`.

Moved the `DataSpec` class to `composer.core`, as the `DataSpec` is now bound directly to the state. #120 will also need this change.

This PR implements part 3 and 4 of the timing abstraction (#146). The implementation differs from the GH issue by adding `num_tokens`, `num_samples`, `get_batch_size`, and `get_num_tokens` to the new `DataSpec` rather than the pytorch dataset class.


composer/core/data_spec.py
```diff
@@ -0,0 +1,135 @@
...
+        else:
+            if isinstance(dataloader.dataset, collections.abc.Sized):
+                try:
+                    self.num_samples = len(dataloader.dataset)
+                except (TypeError, NotImplementedError):
+                    self.num_samples = None
+            else:
+                self.num_samples = None
+
+    def device_transformation_fn(self, batch: Batch):
```
hanlint, line 48: rename to: `device_transforms` ?

composer/core/state.py
```diff
@@ -89,15 +98,14 @@ def __init__(
 
             # data configurations
             grad_accum: int,
-            train_dataloader: types.DataLoader,
-            eval_dataloader: types.DataLoader,
+            train_data: Union[types.DataLoader, types.DataSpec],
```
hanlint, line 101: I would prefer to keep it as `train_dataloader` even though one can pass in a dataSpec

composer/core/data_spec.py
```diff
@@ -0,0 +1,135 @@
...
+
+import torch
+
+from composer.utils.iter_helpers import ensure_tuple
+
+if TYPE_CHECKING:
+    from composer.core.types import Batch, DataLoader
+
+
+class DataSpec:
```
hanlint, line 16: I think users would prefer to pass the functions directly in `__init__`, or as a dictionary with some default functions, rather than have to learn a new class and its logic.

hanlint (changes requested):

composer/core/state.py
hanlint: From user standpoint, I think the desired interface would be to pass in either a `dataloader` or a config, which would be a `dict` with keys for the `device_transforms`, `batch_split`, etc.

ravi-mosaicml: I will `DataSpec` to allow passing of function pointers in init. The API would thus look like:

`state = State(train_dataloader=DataSpec(num_tokens=100, device_transforms=lambda x: x, ...)`

We don't use dicts for configs anywhere since they aren't typed, but I think the above API is quite similar to that.

hanlint: Let's use dicts instead as that is the most light-weight datatype that accomplishes what we need. With the above approach, the user still needs to directly import `DataSpec` in their code to even create our `State`. So then `train_dataloader` would be either the pytorch dataloader or a `dict` config. From usability perspective, we want to replace these `Spec`-holding classes to native dictionary types as much as possible.

composer/core/state.py
```diff
@@ -89,15 +98,14 @@ def __init__(
 
             # data configurations
             grad_accum: int,
-            train_dataloader: types.DataLoader,
-            eval_dataloader: types.DataLoader,
+            train_dataloader: Union[types.DataLoader, types.DataSpec],
```
hanlint, line 101: `State` should only keep the pytorch dataloader; the rest of the details can be in the trainer itself?

composer/core/data_spec.py
hanlint: Per my suggestion around taking `dict` as input, is it then possible to remove the `DataSpec` class entirely? I believe this was make the code much more readable and simplified.

composer/trainer/trainer.py
```diff
@@ -628,17 +602,16 @@ def _ddp_reduce_tensor_sum(tensor: Tensor) -> Tensor:
                             self.checkpoint_loader.restore_checkpoint_rng_state(self.state, self.device)
                         continue
 
-                    state.last_batch_size = self._get_batch_size(state.batch)
+                    state.last_batch_size = state.train_data.get_num_samples_in_batch(state.batch)
```
hanlint, line 605: as mentioned above, we can remove these from state, and store them in the trainer itself.

hanlint (commented):
Made a few suggestions-- I know this PR was just to update the `DataSpec`, but I think we should take this opportunity to remove `DataSpec` completely for better usability.

hanlint (approved):

ravi-mosaicml:
As discussed offline, we will leave dataspec for now and leave it attached to the state. A later PR will remove it. For the time being, however, dataspecs can be initialized using a dictionary which serves as kwargs for the class constructor.

composer/core/state.py
ravi-mosaicml: Discussed offline that we'll leave this for now. I did update the constructor to allow for dicts so one doesn't have to import DataSpec directly.

composer/core/data_spec.py
ravi-mosaicml: Same as above comment.

composer/core/state.py
ravi-mosaicml: Discussed offline that it'll remain attached to state -- algorithms might need the batch split function, for example.

composer/trainer/trainer.py
ravi-mosaicml: Same as above.

Averylamp (approved):
lgtm

PR #182, opened 2021-12-22
Dockerfile enhancements

### LINKED ISSUE #180: Add venv into docker image to enable editable `pip install`
When trying to install composer with `pip install -e .` from within the docker image, we are seeing this error:
```
  File "/usr/bin/composer", line 33, in <module>
    sys.exit(load_entry_point('mosaicml', 'console_scripts', 'composer')())
  File "/usr/bin/composer", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/lib/python3.8/importlib/metadata.py", line 445, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 169, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: mosaicml
```
This seems to be remedied by running the `pip install` from within a virtualenv. Can we bake a virtualenv into the docker image as a workaround?

ravi-mosaicml:
Fixed the dockerfile

1. Installing git to speed up jenkins checkouts, so jenkins doesn't need to download its own version of git every time.
2. Upgrading pip and setuptools since old pip and setuptool versions have bugs in them that prevent a `pip install -e` without root
3. Re-enable the hack to support A100s on google cloud
4. Manually configure the path for python to 1) configure the user-local bin and 2) avoid breaking system packages (which was occuring with upgrade-alternatives)

Closes #180


hanlint:
Is this:
> pip install -e does not properly install console_scripts outside of a virtualenv.

An underlying issue we can fix, rather than do the virtualenv workaround?

ravi-mosaicml:
> An underlying issue we can fix, rather than do the virtualenv workaround?

Maybe? But it would be a rabbit hole to dig down. I suspect it's a bug in pip, though it could be an issue with our setup.py.

docker/pytorch/Dockerfile
```diff
@@ -102,10 +84,22 @@ RUN pip install --no-cache-dir --find-links https://download.pytorch.org/whl/tor
 #########
 RUN rm -rf /tmp/*
 
+
+################################################
+# Create a virtualenv and activate it by default
+################################################
+RUN python${PYTHON_VERSION} -m venv --system-site-packages ~/mosaicml-venv && \
+    python${PYTHON_VERSION} -m venv --upgrade ~/mosaicml-venv && \
+    echo "source ~/mosaicml-venv/bin/activate" >> ~/.bashrc
```
siriuslee, line 93: Will this trigger if you are not running a command under `/bin/bash`? I've usually seen people just do modify `PATH` to include `<path-to-venv>/bin`

ravi-mosaicml: Ah kk, how would I do that in a shell-independent manner?

ravi-mosaicml: Switched to this technique. But I am modifying the /etc/bashrc file as that is sourced by default(adding it to /etc/profile unfortunately doesn't seem to work)

docker/pytorch/Dockerfile
```diff
@@ -3,12 +3,16 @@ ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 ARG DEBIAN_FRONTEND=noninteractive
 
+# remove a bad symlink
+RUN rm -f /usr/local/cuda-11.3/cuda-11.3
```
bandish-shah, line 7: What's going on here?  Is this something vestigial from the Nvidia CUDA container or is it conflicting with the NCCL workaround below?

docker/pytorch/Dockerfile
```diff
@@ -102,11 +106,30 @@ RUN pip install --no-cache-dir --find-links https://download.pytorch.org/whl/tor
...
+####################################################
+# Create a default virtualenv and add it to the path
+####################################################
+RUN python${PYTHON_VERSION} -m venv --system-site-packages /mosaicml-venv && \
+    /mosaicml-venv/bin/python -m pip install --upgrade pip setuptools && \
+    # Set the path for non-interactive shells
+    sed -i 's/PATH=\"/PATH=\"~\/.local\/bin:\/mosaicml-venv\/bin:\/usr\/local\/nvidia\/bin:\/usr\/local\/cuda\/bin:/g' /etc/environment && \
+    echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/profile && \
+    # And for interactive shells
+    echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/bash.bashrc
```
bandish-shah, line 124: Not completely sure if this is needed but I think we also need to set the `VIRTUAL_ENV` environment variable to be safe as some packages use it to check if they're running within an environment.  Found the following article on this, see the end but it says set `VIRTUAL_ENV` and then add the `$VIRTUAL_ENV/bin` path to `$PATH`:
https://pythonspeed.com/articles/activate-virtualenv-dockerfile/

docker/pytorch/Dockerfile
ravi-mosaicml: It's a bad file from the nvidia image

docker/pytorch/Dockerfile
```diff
@@ -116,12 +118,19 @@ SHELL ["/bin/bash", "-c"]
...
     # Set the path for non-interactive shells
     sed -i 's/PATH=\"/PATH=\"~\/.local\/bin:\/mosaicml-venv\/bin:\/usr\/local\/nvidia\/bin:\/usr\/local\/cuda\/bin:/g' /etc/environment && \
+    echo "export VIRTUAL_ENV=/mosaicml-venv" >> /etc/environment && \
     echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/profile && \
+    echo "export VIRTUAL_ENV=/mosaicml-venv" >> /etc/profile && \
     # And for interactive shells
-    echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/bash.bashrc
+    echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/bash.bashrc && \
+    echo "export VIRTUAL_ENV=/mosaicml-venv" >> /etc/bash.bashrc && \
+    echo "export PATH=~/.local/bin:/mosaicml-venv/bin:$PATH" >> /etc/zshenv && \
```
bandish-shah, line 130: [nit] Use `$VIRTUAL_ENV` in all the path prepends instead of the resolved `/mosaicml-venv` path?

docker/pytorch/Dockerfile
```diff
@@ -116,12 +118,19 @@ SHELL ["/bin/bash", "-c"]
...
+    echo "export VIRTUAL_ENV=/mosaicml-venv" >> /etc/zshenv
+
+ENV VIRTUAL_ENV=/mosaicml-venv
```
bandish-shah, line 133: [nit] I think if you set this before the previous `RUN` you won't need to set the `export VIRTUAL_ENV` in the shell config files and can just invoke `$VIRTUAL_ENV` in the exports for `$PATH`, cleans things up a bit.

ravi-mosaicml: I found that the `ENV` doesn't always persist depending on what type of shell is launched, so it's safer to set it via the shell files. But yes, putting it before the run helps clean things up

docker/pytorch/Dockerfile
```diff
@@ -102,11 +108,41 @@ RUN pip install --no-cache-dir --find-links https://download.pytorch.org/whl/tor
...
+    ln -s $(which pip${PYTHON_VERSION}) ${COMPOSER_PYTHON_BIN}/pip && \
+    ln -s $(which pip${PYTHON_VERSION}) ${COMPOSER_PYTHON_BIN}/pip3 && \
+    ln -s $(which pip${PYTHON_VERSION}) ${COMPOSER_PYTHON_BIN}/pip${PYTHON_VERSION} && \
+    # Include this folder, and the local bin folder, on the path 
+    echo "export PATH=~/.local/bin:$COMPOSER_PYTHON_BIN:$PATH" >> /etc/profile && \
+    echo "export PATH=~/.local/bin:$COMPOSER_PYTHON_BIN:$PATH" >> /etc/bash.bashrc && \
+    echo "export PATH=~/.local/bin:$COMPOSER_PYTHON_BIN:$PATH" >> /etc/zshenv
+
+# Ensure that non-interactive shells load /etc/profile
+ENV BASH_ENV=/etc/profile
```
bandish-shah, line 140: Can we use pyenv and avoid doing all the above as well as simplify the python install and version management?  I can set that up if you need a break :-)

ravi-mosaicml: I totally agree that py-env would be cleaner, but there are a few hesitations I have about it:

1. It builds python from source, rather than using a pre-built python. This will require us to use staged docker builds to avoid polluting the docker image with all of the python build dependencies.
2. py-env seems to be a user-level install. We'd need to make sure that every user gets a properly configured env. These hacks configure it for all users, including root and those without real home directories, etc...

bandish-shah: 1.  It looks like pyenv has different download mirrors but I'm not sure if these are precompile binaries or just source that's being downloaded and compiled but I agree it makes more sense to have a `python` image to stage on if we go down this route, this is originally why we had a separate `python` image to provide this logical separation.
2. Definitely a valid concern, do we expect our users to switch from the `mosaicml` user or add users when working with the docker image?  Regardless being able to set this globally would be ideal so makes sense to keep this as is.

bandish-shah (approved):
LGTM

docker/pytorch/Dockerfile
ravi-mosaicml: Sounds good, I'll leave it as is. I think that it's possible for users to run the image as something other than `mosaicml` (not sure why someone would, but I wouldn't put it past me).

PR #183, opened 2021-12-22
Run the run directory uploader test only when libcloud is installed

ravi-mosaicml:
Run the run directory uploader test only when libcloud is installed


ravi-mosaicml:
Closing for #185

PR #184, opened 2021-12-22
Propagate processes that were sigkilled

ravi-mosaicml:
The launcher script suppressed processes that were terminated with sigkill, which meant that tests that OOM'ed were reported as passing. This was a relic from when we used DDP spawn and we deliberately sigkilled child processes. This change ensures that sigkilled error codes are propagated.


jbloxham (approved):
LGTM

PR #189, opened 2021-12-27
Added helper methods for torch.distributed.boradcast

ravi-mosaicml:


composer/utils/ddp.py
```diff
@@ -141,6 +141,49 @@ def all_reduce(
...
+        tensor (torch.Tensor): Data to be sent if ``src`` is the rank of current process,
+            and tensor to be used to save received data otherwise.
+        src (int, optional): Source rank (default: ``0``)
+    """
+    if dist.is_available() and dist.is_initialized():
+        dist.broadcast(tensor, src)
+    world_size = get_world_size()
+    if world_size == 1:
+        return
+    raise RuntimeError(f"Since the world_size({world_size}) > 1, please configure DDP to use ddp.broadcast(). "
```
jbloxham, line 160: Shouldn't this error be raised at the start of the function?

composer/utils/ddp.py
```diff
@@ -141,6 +141,49 @@ def all_reduce(
...
+                       "The mosaic trainer will automatically do this for you.")
```
jbloxham, line 161: Will it? Do you mean the `composer` CLI?

composer/utils/ddp.py
```diff
@@ -141,6 +141,49 @@ def all_reduce(
...
+    """
+    if dist.is_available():
+        dist.broadcast_object_list(object_list, src)
+        # torch.distributed will replace the None's in obj_gather_list with the gathered objects on rank 0
+        # or will just be None on non-rank-0
+    world_size = get_world_size()
+    if world_size == 1:
+        return
+    raise RuntimeError(f"Since the world_size({world_size}) > 1, please configure DDP to use ddp.all_gather_object(). "
+                       "The mosaic trainer will automatically do this for you.")
```
jbloxham, line 184: Same comments here regarding the error message.

composer/utils/ddp.py
```diff
@@ -141,6 +141,49 @@ def all_reduce(
                        "The mosaic trainer will automatically do this for you.")
 
 
+def broadcast(tensor: torch.Tensor, src: int = 0) -> None:
```
jbloxham, line 144: Defaulting `src` to 0 seems a little odd when the Pytorch equivalent has no default value.

composer/utils/ddp.py
```diff
@@ -141,6 +141,49 @@ def all_reduce(
...
+    if dist.is_available() and dist.is_initialized():
+        dist.broadcast(tensor, src)
+    world_size = get_world_size()
+    if world_size == 1:
+        return
+    raise RuntimeError(f"Since the world_size({world_size}) > 1, please configure DDP to use ddp.broadcast(). "
+                       "The mosaic trainer will automatically do this for you.")
+
+
+def broadcast_object_list(object_list: List[Any], src: int = 0) -> None:
```
jbloxham, line 164: Ditto on default for `src`.

jbloxham (approved):

composer/utils/ddp.py
ravi-mosaicml: It's only an issue if there is no ddp and there's more than one node, so the check is down here.

composer/utils/ddp.py
ravi-mosaicml: Yup, the trainer will configure ddp for you. The launch script sets the env variables. (If no env variables are set, then it'll default to a world size of 1). This error would only be raised if someone is using these functions outside of our trainer, or if we attempt to access DDP before it's configured.

PR #191, opened 2021-12-28
Additional Tests

ravi-mosaicml:
* Added checkpointing tests for resnet50 and gpt2 
* Added test to ensure that pillow-simd is installed


Averylamp (approved):
lgtm

PR #198, opened 2022-01-03
Fixed pyright issues

ravi-mosaicml:
1. Updated pyright version to 1.1.203
2. Update python for github tests to 3.9
2.  Made the pyproject.toml slightly stricter. Added warnings for untyped function signatures and implicit string concat, which should later be converted to errors.
3. Removed unused variables and made those errors.
4. Set all instance variables on `__init__`
5. Fixed implicit None in signatures
6. Pin numpy to 1.21.5 since 1.22 introduces typing bugs

### DIFF
--- a/.github/actions/pyright/action.yaml
+++ b/.github/actions/pyright/action.yaml
@@ -12,7 +12,7 @@ runs:
       working-directory: ${{ inputs.working_directory }}
       run: |
         set -ex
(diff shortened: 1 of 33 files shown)


moinnadeem:
LGTM!

moinnadeem (approved):

PR #199, opened 2022-01-05
Checkpointing & DeepSpeed

jbloxham:
Support for model checkpointing while using DeepSpeed. We're lucky in that DeepSpeed provides a pretty good API for creating checkpoints. With this design, we entrust DeepSpeed to serialize the model and optimizer states, and we handle everything else as before.

One of the biggest changes in this PR is that a "checkpoint" is no longer a file, but a folder. This is necessary because it's how DeepSpeed does it - its checkpoints often comprise several files.

In practice, whereas before our structure was something like:
```
checkpoints_folder/
  hparams.yaml
  ep1.pt
  ep2.pt
  ep3.pt
```
We now have:
```
checkpoints_folder/
  hparams.yaml
  ep1/
    mosaic_states.pt
    <any DeepSpeed files>
  ep2/
    ...
  ep3/
    ...
  ...
```

DDP (and DeepSpeed) initialization has been moved into `Trainer.__init__` - this was necessary because DeepSpeed cannot load a checkpoint until DeepSpeed itself has initialized. This seemed reasonable to me - I think the trainer initialization functions should contain all the initialization code that only runs once. This does, however, have the notable consequence that `Trainer.__init__` will block if called from less than `WORLD_SIZE` different processes. This is why I had to remove the `is_rank_zero` check from some of the tests.

The eagle-eyed might note that DeepSpeed's checkpointing API supports a `client_state` param for any non-DeepSpeed state that might want to be checkpointed. I opted not to use this functionality, because I wasn't sure how it would interact with tensors (since it didn't seem to be using `torch.load`). We also just didn't need it - we already have code for checkpointing our own state.

There's a few additional features (like run directory sanitation) that will likely be necessary for this to be useful in practice - owing to the very large size of DeepSpeed checkpoints. I'll tackle those in a subsequent PR.

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -43,6 +43,7 @@ STATE_DICT_SERIALIZATION_FIELDS = [
 ]
 
 # These fields will be serialized using .state_dict(), but will be skipped if DeepSpeed is enabled.
+# When DeepSpeed is being used, model and optimizer states are serialized directly by the DeepSpeed engine.
 STATE_DICT_SERIALIZATION_FIELDS_SKIP_DEEPSPEED = [
     "model",
     "_optimizers",
@@ -197,9 +198,12 @@ class State(Serializable):
         """Returns the state as a :class:`dict`."""
         state_dict: types.StateDict = {}
 
-        # The "proper" form of this check would be something like ``isinstance(self.model, DeepSpeedEngine)``,
-        # but we don't want to need to import DeepSpeed when it's not needed.
-        deepspeed_enabled = self.model.__class__.__qualname__ == "DeepSpeedEngine"
+        deepspeed_enabled = False
+        try:
+            import deepspeed
+            deepspeed_enabled = isinstance(self.model, deepspeed.DeepSpeedEngine)
+        except ImportError:
+            pass
 
         for state_field_name, state_field_value in self.__dict__.items():
             if state_field_name in SKIP_SERIALIZATION_FIELDS:
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import logging
 import warnings
 from typing import TYPE_CHECKING, Callable, ContextManager, Optional, Sequence, Union
-from deepspeed.runtime.engine import DeepSpeedEngine
 
 import torch
 import torch.nn.modules.utils
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import warnings
 from typing import TYPE_CHECKING, Callable, ContextManager, Optional, Sequence, Union
+from deepspeed.runtime.engine import DeepSpeedEngine
 
 import torch
 import torch.nn.modules.utils
@@ -199,6 +200,9 @@ class State(Serializable):
                 continue
             elif state_field_name in STATE_DICT_SERIALIZATION_FIELDS:
                 if state_field_name == "model":
+                    if isinstance(state_field_value, DeepSpeedEngine):
+                        state_dict["_is_model_deepspeed"] = True
+                        continue
                     # Save model directly instead of by class name, since model may be wrapped by DistributedDataParallel
                     serialized_value = state_field_value.state_dict()
                 else:
@@ -222,6 +226,8 @@ class State(Serializable):
         """
         for state_field_name, state_field_value in self.__dict__.items():
             if state_field_name in SKIP_SERIALIZATION_FIELDS:
+                continue
+            elif state_field_name == "model" and state["_is_model_deepspeed"]:
                 continue
             elif state_field_name in DIRECT_SERIALIZATION_FIELDS:
                 setattr(self, state_field_name, state[state_field_name])
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -229,6 +229,8 @@ class State(Serializable):
                 continue
             elif state_field_name == "model" and state["_is_model_deepspeed"]:
                 continue
+            elif state_field_name == "_optimizers" and state["_is_model_deepspeed"]:
+                continue
             elif state_field_name in DIRECT_SERIALIZATION_FIELDS:
                 setattr(self, state_field_name, state[state_field_name])
             elif state_field_name in STATE_DICT_SERIALIZATION_FIELDS:
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -43,6 +43,12 @@ STATE_DICT_SERIALIZATION_FIELDS = [
     "scaler",
 ]
 
+# These fields will be serialized using .state_dict(), but will be skipped if DeepSpeed is enabled.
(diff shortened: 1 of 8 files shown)


composer/core/state.py
```diff
@@ -42,6 +42,12 @@
     "scaler",
 ]
 
+# These fields will be serialized using .state_dict(), but will be skipped if DeepSpeed is enabled.
```
ravi-mosaicml, line 45: Curious where is the model serialized if not here?

composer/trainer/checkpoint.py
```diff
@@ -18,6 +18,26 @@
 log = logging.getLogger(__name__)
 
 
+def parse_checkpoint_filepath(checkpoint_filepath: str):
```
ravi-mosaicml, line 21: Discussed offline to tarball the checkpoint. This will help ensure 1) a checkpoint doesn't get corrupted in transit (e.g. you somehow download some of the files but not all) and 2) object stores use a flat directory structure, and it can get expensive to list all files that begin with a prefix.

composer/trainer/checkpoint.py
```diff
@@ -43,24 +68,32 @@ def load_checkpoint(self, state: State):
...
                                   f"{checkpointed_world_size}. The seed will not be restored.")
-                    return
-                seed_to_restore = self.state_dict["seed"][ddp.get_global_rank()]
-                seed_all(seed_to_restore)
-                return seed_to_restore
+                else:
+                    seed_to_restore = self.state_dict["seed"][ddp.get_global_rank()]
+                    seed_all(seed_to_restore)
+
+        if state.model.__class__.__qualname__ == 'DeepSpeedEngine':
```
ravi-mosaicml, line 89: same thing here re: conditional import

composer/trainer/checkpoint.py
```diff
@@ -139,9 +172,24 @@ def save_checkpoint(self, state: State, seed: int, device: Device, config: Optio
...
+            raise ValueError(f"Invalid checkpoint event: {self.save_event}")
+
+        if state.model.__class__.__qualname__ == 'DeepSpeedEngine':
+            state.model.save_checkpoint(self.checkpoint_folder, tag)  # type: ignore
+
         if ddp.get_global_rank() != 0:
             # only rank 0 saves checkpoints
             # Need the check down here so all the DDP syncs will work for generating the checkpoint
+
+            # Sync before exiting so that even the non rank 0 processes cannot exit before the
```
ravi-mosaicml, line 190: Good catch, thanks! Generally, I'd prefer this style of short circuiting early to reduce indentation. But, since all processes need to have the same # of ddp barriers, I personally find it easier to follow if there is just one line to `ddp.barrier` that is shared by both the rank-0 and non rank-0 processes...e.g. something like:

```python3
if ddp.get_global_rank() != 0:
        checkpoint_filepath = get_mosaic_checkpoint_filepath(self.checkpoint_folder, tag)
        if not os.path.exists(os.path.join(self.checkpoint_folder, tag)):
            os.makedirs(os.path.join(self.checkpoint_folder, tag))
        with open(checkpoint_filepath, 'xb') as f:
              ....
ddp.barrier()  # <- same line is used for all code paths
```

composer/trainer/trainer.py
```diff
@@ -300,32 +300,49 @@ def __init__(
...
+            assert self.deepspeed_hparams is not None
+            deepspeed_config = self.deepspeed_hparams.initialize_object(self.state, self.grad_clip_norm)
+            optimizer = ensure_tuple(self.state.optimizers)[0]
+            (self.state.model, self.state.optimizers, _, _) = deepspeed.initialize(
+                config=deepspeed_config,
+                model=self.state.model,
+                optimizer=optimizer,
+            )
+
+            if self.checkpoint_loader:
```
ravi-mosaicml, line 331: nit: unindent this block, replace line 335 with `if not self.deepspeed_enabled:`, and remove the duplicated code on lines 336-339.

tests/conftest.py
```diff
@@ -124,14 +124,26 @@ def subfolder_run_directory(tmpdir: pathlib.Path, monkeypatch: MonkeyPatch) -> N
 @pytest.fixture(autouse=True)
 def configure_ddp(request: pytest.FixtureRequest):
     backend = None
+    use_deepspeed = False
     for item in request.session.items:
-        marker = item.get_closest_marker('gpu')
-        if marker is not None:
+        gpu_marker = item.get_closest_marker('gpu')
+        deepspeed_marker = item.get_closest_marker('deepspeed')
```
ravi-mosaicml, line 130: can you raise an error if a test is marked as both gpu and deepspeed? I think they should be mutually exclusive?

tests/trainer/test_checkpoint.py
```diff
@@ -247,10 +252,18 @@ def test_load_weights(
     (0, 1),
     (1, 0),
 ])
-@pytest.mark.parametrize("model_name", [None, "resnet50_synthetic", "gpt2_52m"])
+@pytest.mark.parametrize(
+    "model_name",
+    [
+        None,
+        # "resnet50_synthetic",
```
ravi-mosaicml, line 259: curious why are these commented out?

tests/trainer/test_checkpoint.py
```diff
@@ -223,20 +224,24 @@ def test_load_weights(
...
+    )
 
 
 @pytest.mark.timeout(90)
 @pytest.mark.parametrize("world_size", [
     pytest.param(1),
     pytest.param(2, marks=pytest.mark.world_size(2)),
 ])
+@pytest.mark.parametrize("deepspeed_enabled,zero_stage", [
+    pytest.param(False, None, id="ddp"),
```
ravi-mosaicml, line 240: would it make sense to merge this parameterize with the one below? I don't think deepspeed works with cpu?

ravi-mosaicml (commented):
Overall looks great! Just the tarball change + and otherwise small thoughts.

composer/trainer/trainer_hparams.py
```diff
@@ -195,6 +195,8 @@ class TrainerHparams(hp.Hparams):
                                          default=None)
 
     def validate(self):
+
+        print(self.deepspeed)
```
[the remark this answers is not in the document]
jbloxham: Gah I'd been wondering where that text was coming from.

composer/core/state.py
ravi-mosaicml: NVM, I missed it.

tests/trainer/test_checkpoint.py
jbloxham: True, I have a `pytest.skip` in the test to prevent testing with DeepSpeed on CPU, but the triple parametrizing is cleaner.

tests/trainer/test_checkpoint.py
jbloxham: They were flaky. I checked and they're flaky on dev too. We hadn't noticed because they're only use on GPU testing, which we don't automate.

ravi-mosaicml (approved):
LGTM, thanks for making the changes!

PR #200, opened 2022-01-05
Load checkpoints from cloud storage

### LINKED ISSUE #192: Enable loading a checkpoint from a blob storage
For multi-cloud training setups, it would be useful to be able to download the checkpoint from a cloud blob storage, and load it from there instead of from local disk.

ravirahman:
Added support to load checkpoints stored in object storage (rather than just on the local disk) and from URLs. Closes #192.

- Refactored the run directory uploader to separate out object store related utilities to `composer.utils.object_store` (and added test coverage).
- Updated the checkpointer loader and hparams to support checkpoints stored in URLs and in object stores. Chunk sizing and whether to show tqdm bars are hparams.
- Refactored how ddp checkpoints are loaded and stored to avoid an `os.listdir` and writing temporary files to the run directory
- Updated the `Trainer.__init__` to propagate through this change.


ravi-mosaicml:
Going to re-open this PR once #199  is merged in and this PR is rebased on top of it.

moinnadeem:
I've tested this and it works! Great job Ravi!

moinnadeem (approved):

PR #201, opened 2022-01-05
Distinguish between `dist` and DDP

jbloxham:
The distributed runtime and the DDP engine are distinct entities, but our code has been treating them almost as synonyms. This is already causing some confusion in parts of the DeepSpeed integration, and it will only get worse if we experiment with other parallelism techniques like model and pipeline parallelism. The purpose of this PR is to separate out DDP-specific code from anything that just deals with the distributed runtime in general.

This is purely a refactor of something that was making me unhappy.


ravi-mosaicml (approved):
LGTM, thanks for doing this!

PR #203, opened 2022-01-06
Fix deterministic mode (and use it for tests); simplify checkpointing tests

ravi-mosaicml:
1. Deterministic mode on GPUs requires the `CUBLAS_WORKSPACE_CONFIG`. Setting this variable when configuring deterministic mode.
2. Renamed composer.utils.determinism to composer.utils.reproducibility to match the pytorch naming in the docs
3. Configured tests to use deterministic mode by default.
4. Configured checkpoints to not compress by default, since compression is slow
5. Made the gpu and deepspeed markers mutually exclusive, since they require different process initialization

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -86,11 +86,11 @@ class Trainer:
             (default: ``5.0``)
         seed (int, optional): The seed used in randomization. When not provided a random seed
             will be created. (default: ``None``)
-        log_destinations (List[BaseLoggerBackend], optional): The destinations to log training information to.
-            (default ``[TQDMLoggerBackend()]``).
         deterministic_mode (bool, optional): Run the model deterministically. Experimental. Performance
             degradations expected. Certain Torch modules may not have deterministic implementations,
             which will result in a crash. (default: ``False``)
+        log_destinations (List[BaseLoggerBackend], optional): The destinations to log training information to.
+            (default ``[TQDMLoggerBackend()]``).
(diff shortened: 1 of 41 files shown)


composer/trainer/trainer.py
```diff
@@ -87,9 +86,6 @@ class Trainer:
             (default: ``5.0``)
         seed (int, optional): The seed used in randomization. When not provided a random seed
             will be created. (default: ``None``)
-        deterministic_mode (bool, optional): Run the model deterministically. Experimental. Performance
-            degradations expected. Certain Torch modules may not have deterministic implementations,
-            which will result in a crash. (default: ``False``)
```
Averylamp, line 92: I don't believe we should remove it from calling the trainer directly as well.  Ideally we can have it both in the launcher script and the trainer creation args

Averylamp (changes requested):
I don't think we should remove deterministic as an arg to Trainer, but other than that looks fine

Averylamp (approved):
lgtm, 
Deferring to Jamie on deepspeed tests. 
Also am noticing that our root `conftest.py` is getting large again, when we do have our own fixtures subfolder, so bonus points for moving deepspeed fixtures into its own deepspeed fixtures folder

composer/trainer/trainer.py
ravi-mosaicml: Figured out that we needed to configure it before the first test, not before the process. So resolved :)

PR #210, opened 2022-01-08
Integrate the timer into the training loop

ravi-mosaicml:
- Updated the `TrainerHparams`, `Trainer`, and `State`, to take max_duration instead of max_epochs. However, for the time being, it is required that `max_duration` be in epoch units. This will be required until `state.max_epochs` is no longer used by anything.
- Updated the state to store an instance of `Timer`. Added helper getters to the state for backwards compatibility with the previous API. Added deprecation warnings
- Updated the Timer to allow comparisons directly between the timer object and instances of Time. These statements are now equivalent  `(timer <=other) == (timer.get(other.unit) <= other)`
- Updated the Timer to track the number of batches, samples, and tokens in the current epoch. Tracking batches is required for `state.batch_idx`; and it'd be good to provide equivalent functions for samples and tokens
- Updated the Time to allow for integer comparisons. The same units will be assumed. These statements are now equivalent  `(time <= 10) == (time <= Time(10, time.unit)`
- Updated the rest of the codebase -- mainly tests and hparam files -- to use these new APIs
- Removed the old Time parsing and conversion logic from the composer.optim.scheduler. Instead, replaced it with the timer conversion code that is shelved on https://github.com/mosaicml/composer/tree/ravi/timer_convert


Averylamp (approved):
lgtm

hanlint (approved):

hanlint:
Ah, that was meant to be a Comment not an Approve -- defer to @jbloxham and @abhi-mosaic for approvals on this PR

composer/core/time.py
```diff
@@ -223,28 +231,28 @@ def __gt__(self, other: object):
...
-    def __sub__(self, other: object):
+    def __sub__(self, other: object) -> Time[TValue]:
         other = self._parse(other)
         if self.unit != other.unit:
             raise RuntimeError(f"Cannot subtract {other} from {self} since they have different units.")
         return Time(self.value - other.value, self.unit)
 
-    def __rsub__(self, other: object):
+    def __rsub__(self, other: object) -> Time[TValue]:
         return (-self) + other
```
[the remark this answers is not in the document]
ravi-mosaicml: Yeah, it gets annoying that `a + b` invokes a different operation than `b + a`, but that's what all these methods are for...

composer/core/state.py
```diff
@@ -166,6 +171,47 @@ def __init__(
...
+
+    @property
+    def max_duration(self):
+        return self._max_duration
+
+    @max_duration.setter
+    def max_duration(self, max_duration: Union[str, Time[int]]):
+        if isinstance(max_duration, str):
+            max_duration = cast(Time[int], Time.from_timestring(max_duration))
+        if max_duration.unit != TimeUnit.EPOCH:
```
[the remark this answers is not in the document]
ravi-mosaicml: Using the setter for `max_duration` in `__int__` now, thanks!

composer/core/time.py
```diff
@@ -298,7 +306,7 @@ def from_timestring(cls, timestring: str) -> Time:
         value = float(value)  # always parsing first as float b/c it could be scientific notation
         if unit != TimeUnit.DURATION:
             value = int(value)
-        return Time(value, unit)
+        return cls(value, unit)
```
jbloxham, line 309: Similar to my other question - is this necessary to preserve the generic type annotation?

composer/optim/scheduler.py
```diff
@@ -291,42 +193,28 @@ class CosineAnnealingLRHparams(SchedulerHparams):
...
-
 
 @dataclass
 class CosineAnnealingWarmRestartsHparams(SchedulerHparams):
     """Hyperparameters for the ``CosineAnnealingWarmRestarts` <https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CosineAnnealingWarmRestarts.html#torch.optim.lr_scheduler.CosineAnnealingWarmRestarts>`_
     scheduler.
     """
 
-    T_0: Time = hp.required("Number of iterations for the first restart.")
+    T_0: str = hp.required("Number of iterations for the first restart.")
```
jbloxham, line 210: Docstring could use an update.

composer/optim/scheduler.py
```diff
@@ -291,42 +193,28 @@ class CosineAnnealingLRHparams(SchedulerHparams):
     scheduler.
     """
 
-    T_max: Time = hp.required(doc="Maximum number of iterations.")
+    T_max: str = hp.required(doc="Maximum number of iterations.")
```
jbloxham, line 196: Ditto (see below)

composer/optim/scheduler.py
```diff
@@ -336,7 +224,7 @@ class LinearLRHparams(SchedulerHparams):
 
     start_factor: float = hp.optional("Number to multiply learning rate at the start.", default=1.0 / 3)
     end_factor: float = hp.optional("Number to multiply learning rate at the end .", default=1.0)
-    total_iters: Time = hp.optional("Number of linear decay steps. Default: 5 iterations.", default="5ba")
+    total_iters: str = hp.optional("Number of linear decay steps. Default: 5 iterations.", default="5ba")
```
jbloxham, line 227: DItto

composer/optim/scheduler.py
```diff
@@ -351,7 +239,7 @@ class WarmUpLRHparams(SchedulerHparams):
     """
 
     warmup_factor: float = hp.optional("Number to multiply learning rate at start.", default=1.0 / 3)
-    warmup_iters: Time = hp.optional("Number of warmup step. Default: 5 iterations.", default="5ba")
+    warmup_iters: str = hp.optional("Number of warmup step. Default: 5 iterations.", default="5ba")
```
jbloxham, line 242: Ditto

tests/test_time_convert.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
```
jbloxham, line 1: Thanks for writing this test!

jbloxham (approved):
Looks good to me - just minor comments. Only thing I would add is it seems a little dubious to me that `DURATION` is a valid time unit, given all the special logic around it. Is anything actually dependent on being able to represent `DURATION` and `EPOCH`/`BATCH`/etc. with the same object?

composer/core/time.py
```diff
@@ -195,6 +197,12 @@ def _parse(self, other: object) -> Time:
         raise NotImplementedError(f"Cannot convert type {other} to {self.__class__.__name__}")
 
     def _cmp(self, other: object) -> int:
+        # When doing comparisions, and other is an integer (or float), we can safely infer
+        # the unit from self.unit
+        # E.g. calls like this should be allowed: if batch < 42: do_something()
+        # This eliminates the need to call .value everywhere
+        if isinstance(other, (int, float)):
+            other = type(self)(other, self.unit)
```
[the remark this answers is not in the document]
ravi-mosaicml: I prefer this style should someone subclass the `Time` -- then it will return the same type as the subclass, rather than the parent class. Granted this depends on the constructor being the same...

composer/core/time.py
```diff
@@ -223,28 +231,28 @@ def __gt__(self, other: object):
     def __ge__(self, other: object):
         return self._cmp(other) >= 0
 
-    def __add__(self, other: object):
+    def __add__(self, other: object) -> Time[TValue]:
```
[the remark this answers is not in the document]
ravi-mosaicml: `Time[float]` is only used for durations, and `Time[int]` is used for everything else, so mixed types wouldn't be added together.

composer/core/time.py
ravi-mosaicml: Yeah, same thing here re: subclassing.

ravi-mosaicml:
Yeah, Duration isn't like the rest of the timing units. It's nice to have it be part of the Time class as it supports similar operations, like loading from a time string, and addition / comparison of like units, even though division and multiplication behave differently.

PR #211, opened 2022-01-08
LM Streaming Dataset

abhi-mosaic:
This is a WIP but should work for some basic use cases. 

Successfully used streaming C4 dataset for GPT2 pretraining, confirmed that multi-epoch works, and visually inspected tensors for MLM modeling. Will defer BERT+C4 testing to @moinnadeem.

I added a file to `tests/datasets/...` but it probably needs to be refactored / removed for now. I think we'll need a pretty cohesive set of tests for StreamingLM in the future.

Known issues:
* determining the length of a new dataset is not implemented (right now info for C4 is cached)
* Epoch end handling
  * if `world_size` > `n_shards` and `max_samples` is not set, there could be a bug at the end of the epoch
  * `drop_last` is currently required to be True, until this bug is fixed
  * Basically, use `max_samples` to be safe, future updates will remove this requirement.
* If there are multiple devices, and we are not iterating over the whole dataset, then there are very few guarantees about the exact subset of data that we will see. A single GPU run would see shard 1 only, but an 8 GPU run would see 1/8th of 8 shards. This could be a problem for reproducability... this is why I also included the `max_shards` parameter as a way to deterministically subsample a dataset.
* I am quite nervous about how shuffling is handled overall, and whether the streaming dataset's shard order will / will not get reshuffled at the end of each epoch, if `max_samples` is set. I am also worried about how shuffling is handled when multiple devices are looking at the same shard... they should be careful to subsample first then shuffle or else data could be duplicated. These concerns can be addressed in future PRs, right now I think this class will mostly be used for large single-epoch pretraining.

### DIFF
--- a/composer/trainer/deepspeed.py
+++ b/composer/trainer/deepspeed.py
@@ -83,7 +83,7 @@ class DeepSpeedHparams(hp.Hparams):
         elif state.precision == Precision.FP16:
             deepspeed_config["fp16"] = {
                 "enabled": True,
-                "initial_scale_power": 16,
+                "initial_scale_power": 0,
                 "loss_scale_window": 2000,
             }
 
--- a/composer/yamls/models/gpt2_125m_streaming.yaml
+++ b/composer/yamls/models/gpt2_125m_streaming.yaml
@@ -1,0 +1,110 @@
+# GPT2-125m with streaming C4 dataset
+
+train_dataset:
+  streaming_lm:
+    dataset_name: c4
+    dataset_config_name: en
+    split: train
+    max_shards: -1
+    max_samples: 7168000
+    max_seq_len: 1024
+    group_method: concat
+    tokenizer_name: gpt2
+    use_masked_lm: false
+    seed: 17
+    shuffle: true
+    drop_last: true
+val_dataset:
+  streaming_lm:
+    dataset_name: c4
+    dataset_config_name: en
+    split: validation
+    max_shards: -1
+    max_samples: 128000
+    max_seq_len: 1024
+    group_method: concat
+    tokenizer_name: gpt2
+    use_masked_lm: false
+    seed: 17
+    shuffle: false
+    drop_last: true
+
+model:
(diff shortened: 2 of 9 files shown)


composer/yamls/models/gpt2_125m_streaming.yaml
```diff
@@ -0,0 +1,110 @@
...
+    split: validation
+    max_shards: -1
+    max_samples: 128000
+    max_seq_len: 1024
+    group_method: concat
+    tokenizer_name: gpt2
+    use_masked_lm: false
+    seed: 17
+    shuffle: false
+    drop_last: true
```
abhi-mosaic, line 30: Just want to highlight what the new YAML looks like.

abhi-mosaic: For @moinnadeem you should use `group_method: truncate` for MLM.

moinnadeem:
To be clear, do we want to actually merge this, or is this a "draft PR"?

abhi-mosaic:
> To be clear, do we want to actually merge this, or is this a "draft PR"?

Good call just made it a draft :) Needs a little bit more work.

composer/trainer/deepspeed.py
```diff
@@ -83,7 +83,7 @@ def initialize_object(self, state: State, grad_clip_norm: Optional[float]):
         elif state.precision == Precision.FP16:
             deepspeed_config["fp16"] = {
                 "enabled": True,
-                "initial_scale_power": 16,
+                "initial_scale_power": 0,
```
jbloxham, line 86: For reference, this is the change that was causing issues training models last week. An initial scale power of 0 means we multiply gradients by 2^0 - i.e. loss scaling is effectively disabled. 32 is DeepSpeed's default; 16 is consistent with PyTorch's defaults and seems to work well for the scale of model we work with.

abhi-mosaic:
Closing, newer PR here: #489

PR #213, opened 2022-01-10
Add larger GPT models

jbloxham:
Time to go bigger :)

Specs copied from https://arxiv.org/pdf/2005.14165.pdf. Only fields changed from the existing GPT models are:
`model.gpt2.n_embd`
`model.gpt2.n_head`
`model.gpt2.n_inner`
`model.gpt2.n_layer`
`optimizer.adamw.lr`
`train_batch_size`
`grad_accum`

### DIFF
--- a/composer/yamls/models/gpt2_1,3b.yaml
+++ b/composer/yamls/models/gpt2_1,3b.yaml
@@ -1,0 +1,97 @@
+train_dataset:
+  lm:
+    split: train
+    datadir: /datasets/openwebtext_saved
+    tokenizer_name: gpt2
+    seed: 17
+    shuffle: false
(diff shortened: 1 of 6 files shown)


moinnadeem:
What machine is `grad_accum` tuned for?

jbloxham:
> What machine is `grad_accum` tuned for?

It's just assuming a microbatch size of 1 for any 8-GPU machine. I'm of the opinion that providing pre-tuned `grad_accum` is counterproductive.

abhi-mosaic (approved):
I think these are fine for now for the purpose of performance benchmarking, in the near future we can update / replace them with GPT3 convergence run YAMLs.

PR #214, opened 2022-01-10
Adding use_masked attr to new GPT-2 YAMLs

moinnadeem:
We just merged in some new GPT-2 YAMLs, but they lack the `use_masked_lm` property. This serves to add this to the YAMLs, so BERT can get merged in.


jbloxham (changes requested):
Can this change be combined with the BERT PR? There's no context here as to why this change is necessary.

moinnadeem:
Ah, I just started rebasing and a ton of problems appeared, so this seemed quicker

Basically, adding the GPT-2 YAMLs made PR Gate fail all of a sudden (due to missing a value), and it was quicker to add this PR rather than rebase off of dev + fix it in the BERT PR!

PR #215, opened 2022-01-11
Make the run directory rank-local; fix checkpoints saving and restoring

ravi-mosaicml:
Fixed the run directory and deepspeed checkpointing:

- Creating a run directory by default, if one was not set via an environment variable
- Removed `run_directory.get_relative_to_run_directory(...)`; replaced that with `os.path.join(run_directory.get_run_directory(...))`
- Having one run directory per node (i.e. sharing the run directory across local ranks) won't work in multi-node training. This change makes the run directory rank-local (so each rank sees its own run directory)
- Added a helper method `run_directory.get_node_run_directory()`. This should not really be used directly, but rather 
- Switched the run directory uploader and wandb to run on every rank instead of just rank zero. This ensures that all artifacts will be properly stored.
- Fixed deepspeed checkpointing:
  
  Specifically, when using deepspeed with zero-1+, each rank wrote to the folder specified. Only rank 0's data was being stored, meaning that we were losing rank 1+'s state (i.e. the optimizer state). Deepspeed checkpointing tests were broken. Now, each rank's data is being stored by the rank-local run directory uploader (see above bullet). Meaning, there are N files per checkpoint now.

  To restore a checkpoint, you need the global rank zero checkpoint (which contains the model) and the rank n checkpoint (which contains the optimizer state). To support this, the checkpoint loader takes a checkpoint path that is parameterized by the rank, so each rank can fetch the files it needs. As an optimization, the LOCAL rank zero on each node is responsible for downloading the GLOBAL rank zero checkpoint (it would be redundant to download this file multiple times per node).

TODO:

- [x] Do a wandb run to see what groups look like

### DIFF
--- a/composer/loggers/logger_hparams.py
+++ b/composer/loggers/logger_hparams.py
@@ -12,6 +12,7 @@ import yahp as hp
 from composer.core.logging import BaseLoggerBackend, LogLevel
 from composer.core.types import JSON
 from composer.loggers.mosaicml_logger import RunType
+from composer.utils import dist
 
 if TYPE_CHECKING:
     from composer.loggers.file_logger import FileLoggerBackend
@@ -92,7 +93,7 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
     """
 
     project: Optional[str] = hp.optional(doc="wandb project name", default=None)
-    name: Optional[str] = hp.optional(doc="wandb run name", default=None)
+    name: Optional[str] = hp.optional(doc="wandb group name", default=None)
     entity: Optional[str] = hp.optional(doc="wandb entity", default=None)
     tags: str = hp.optional(doc="wandb tags comma separated", default="")
     log_artifacts: bool = hp.optional(doc="Whether to log artifacts", default=False)
@@ -190,7 +191,8 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
 
         init_params = {
             "project": self.project,
-            "name": self.name,
+            "name": f"Rank {dist.get_global_rank()}",
+            "group": self.name,
             "entity": self.entity,
             "tags": tags,
         }
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import logging
 import os
 import random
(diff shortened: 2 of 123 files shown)


composer/utils/dist.py
```diff
@@ -78,6 +78,16 @@ def get_local_rank() -> int:
     return local_rank
 
 
+def get_node_id() -> int:
```
jbloxham, line 81: This would generally be called the "node rank".

composer/utils/dist.py
```diff
@@ -78,6 +78,16 @@ def get_local_rank() -> int:
...
+    """Returns the node ID for the current rank. For example, if there are 2 nodes, and 2 ranks per node, then
```
jbloxham, line 82: Can you add a big caveat to this comment that this assumes an equal number of processes per node?

composer/trainer/checkpoint.py
```diff
@@ -135,25 +157,78 @@ def load_checkpoint(self, state: State):
...
+                    with tarfile.open(rank_n_checkpoint_archive_filepath) as tarball:
+                        tarball.extractall(extracted_checkpoint_folder)
+
+            # Wait for all checkpoints on the node to finish downloading
+            # technically, for this barrier, we only need to wait for the
+            # global rank zero checkpoint, but putting the barrier here
+            # allows for other ranks to begin downloading their checkpoints
+            dist.barrier()
+
+            # Now, all ranks load the checkpoint that local rank zero downloaded
```
jbloxham, line 223: This function is big enough that it's very difficult to read and keep track of indentation. Could you break it up into smaller subfunctions? It seems like downloading/unpacking the checkpoints and loading the checkpoints could belong to separate functions.

composer/trainer/checkpoint.py
```diff
@@ -135,25 +157,78 @@ def load_checkpoint(self, state: State):
...
 
-        with tempfile.TemporaryDirectory() as checkpoint_folder:
+        # download the checkpoint to the node-local run directory
+        # since the run directory is sub-foldered by rank,
+        # the node-local run directory is just the parent folder
+        tempdir_ctx = tempfile.TemporaryDirectory() if dist.get_local_rank() == 0 else contextlib.nullcontext(None)
+        with tempdir_ctx as tempdir:
+            local_node_zero = dist.get_local_world_size() * dist.get_node_id()
+            tmpdir_list = [tempdir]
+            dist.broadcast_object_list(tmpdir_list, src=local_node_zero)
```
jbloxham, line 167: This won't work for multinode - processes will be broadcasting across node boundaries.

The "correct" way to do this would be with [distributed process groups](https://pytorch.org/docs/stable/distributed.html#groups), but this is not supported for NCCL.

Barring that, I think you'll have to use `all_gather_object`, with the non-local-rank-0 nodes passing in `None`.

jbloxham (changes requested):
There's a major issue with the intra-node communication in multinode settings, and I'll want to take another look after that's fixed. Other than that, it's looking good!

composer/loggers/logger_hparams.py
```diff
@@ -190,7 +191,8 @@ def get_flattened_dict(data: Dict[str, Any], _prefix: List[str] = []) -> Dict[st
 
         init_params = {
             "project": self.project,
-            "name": self.name,
+            "name": f"Rank {dist.get_global_rank()}",
+            "group": self.name,
```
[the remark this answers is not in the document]
ravi-mosaicml: Not yet but will do!

composer/trainer/checkpoint.py
ravi-mosaicml: Good point; refactored the function for readability.

composer/trainer/checkpoint.py
ravi-mosaicml: Sounds good, refactored to use `all_gather_object`

composer/loggers/logger_hparams.py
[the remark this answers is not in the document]
ravi-mosaicml: https://wandb.ai/mosaic-ml/wandb-grouping-test?workspace=user-ravimosaicml

jbloxham (approved):
Looks good now! Thanks for cleaning all that up!

Issue #218, opened 2022-01-12
Checkpoint at the end of training, regardless of checkpoint interval


moinnadeem:
# Motivation
A user might have training running for an odd number of steps which isn't evenly divisible by some `checkpoint_interval`. They are fine with checkpoints running every `3500` steps. For reference, training consists of `68,796` steps. However, this implies that the last checkpoint is captured at `66,500` steps. 

This leaves additional training that wasn't captured. We should provide the expressivity to enable checkpointing to happen at a step frequency, but still capture the final model, regardless of what that interval was.

hanlint:
Is it fair to say that as long as checkpointing is enabled (at any frequency), we always checkpoint the last model?

moinnadeem:
I think that's fine with me, but do we want this to be a hyperparameter? 

Honestly, I can't find a motivating case for when the user would want to checkpoint, but wouldn't want the final model. The only one that I can think of is some scenario when the model is extremely large, and checkpointing to disk adds an extreme time + storage overhead. This seems an extreme edge case.

Thoughts @hanlint? Should I just go ahead and add checkpointing at the end of training if checkpointing is enabled?

moinnadeem:
Closed by #219

PR #219, opened 2022-01-12
Adding checkpointing at the end of training

moinnadeem:
Addresses #218


ravi-mosaicml (approved):
LGTM! Would be probably run a short, manual test just to make sure it's saved before doing a long run in case if there's something that we missed.

moinnadeem:
Yup, already did that before pushing the commit! Running a few lasts tests before merging

PR #220, opened 2022-01-12
Adding conditional branching on data_collator

moinnadeem:
BERT broke the GPT-2 models, because the GPT-2 tokenizer doesn't have a padding token. We need to conditionally branch on this to determine which data collator to use.


dskhudia (approved):

moinnadeem:
Verified this works!

PR #221, opened 2022-01-12
 Remove old timing calls from selective backprop

ravi-mosaicml:
Updated the selective backprop API to  use the timing abstraction, as detailed in #146


composer/algorithms/selective_backprop/selective_backprop.py
```diff
@@ -205,10 +205,10 @@ def match(self, event: Event, state: State) -> bool:
             return False
 
         is_chosen = do_selective_backprop(
-            epoch=state.epoch,
-            batch_idx=state.batch_idx,
-            start_epoch=int(state.max_epochs * self.hparams.start),
-            end_epoch=int(state.max_epochs * self.hparams.end),
+            current_duration=state.get_elapsed_duration().value,
```
Landanjs, line 208: `.value` should be removed, right? In the other code, this was wrapped with `float()`, is that necessary here?

Oh, maybe these are equivalent?

Landanjs (approved):
LGTM besides that one comment!

composer/algorithms/selective_backprop/selective_backprop.py
ravi-mosaicml: Yup, they're equivalent, but I'll go with `float()` for consistency

PR #223, opened 2022-01-13
Cleaned Up State

ravi-mosaicml:
1. Remove the dataspec from state! Instead, the trainer sets `batch_num_samples`, `batch_num_tokens`, `microbatches`, and `microbatch_idx` so these fields are accessible to algorithms that need to know the pertinent information that the data spec provided.
2. Removed last_batch_size from state. Replaced it with `dist.all_reduce(state.batch_num_samples)` where it was used.
3. Removed `train_batch_size` and `eval_batch_size` from state, as algorithms should not depend on constant batch sizing. Replaced it with `state.train_dataloader.batch_size * dist.get_world_size()` in the few places where it was used
4. Added a helper function to `get_device_of_batch`, which is required for part 3 (since `dist.all_reduce` requires tensors to be placed on the device it torch.dist was initialized with)
5. Fixed the type annotations for `ensure_tuple` to support `get_device_of_batch`

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -6,11 +6,14 @@ import time
 from collections import deque
 from typing import Deque, Optional
 
+import torch
+
 from composer.callbacks.callback_hparams import SpeedMonitorHparams
 from composer.core import Logger, State
 from composer.core.callback import RankZeroCallback
 from composer.core.types import StateDict
 from composer.utils import dist
+from composer.utils.data import get_device_of_batch
 
 
 class SpeedMonitor(RankZeroCallback):
@@ -39,6 +42,7 @@ class SpeedMonitor(RankZeroCallback):
         self.batch_num_samples: Deque[int] = deque(maxlen=window_size)  # rolling list of num samples in batch.
         self.hparams = SpeedMonitorHparams(window_size=window_size)
         self.loaded_state: Optional[StateDict] = None
+        self.current_batch_num_samples = torch.tensor(0)
 
     def state_dict(self) -> StateDict:
         current_time = time.time()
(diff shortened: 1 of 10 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -39,6 +42,7 @@ def __init__(self, window_size: int):
         self.batch_num_samples: Deque[int] = deque(maxlen=window_size)  # rolling list of num samples in batch.
         self.hparams = SpeedMonitorHparams(window_size=window_size)
         self.loaded_state: Optional[StateDict] = None
+        self.current_batch_num_samples = torch.tensor(0)
```
jbloxham, line 45: This looks vestigial?

composer/core/state.py
```diff
@@ -107,8 +122,8 @@ def __init__(
 
             # data configurations
             grad_accum: int,
-            train_dataloader: Union[types.DataLoader, types.DataSpec, Dict[str, Any]],
-            eval_dataloader: Union[types.DataLoader, types.DataSpec, Dict[str, Any]],
+            train_dataloader: types.DataLoader,
+            eval_dataloader: types.DataLoader,
```
jbloxham, line 126: 😌

composer/trainer/trainer.py
```diff
@@ -214,6 +217,9 @@ def __init__(
             eval_dataloader = DataSpec(eval_dataloader)
         eval_dataloader.dataloader = DDPDataLoader(eval_dataloader.dataloader)
 
+        self._train_data_spec = train_dataloader
+        self._eval_data_spoc = eval_dataloader
```
jbloxham, line 221: Typo on `_eval_data_spoc`.

composer/trainer/trainer.py
```diff
@@ -580,7 +588,12 @@ def _train_loop(self) -> None:
 
                     self.engine.run_event(Event.AFTER_DATALOADER)
 
-                    microbatches = state.train_data.split_batch(state.batch, state.grad_accum)
+                    num_samples_in_batch = self.device.tensor_to_device(
+                        torch.tensor([state.batch_num_samples], dtype=torch.int))
+                    num_tokens_in_batch = self.device.tensor_to_device(
+                        torch.tensor([state.batch_num_tokens], dtype=torch.int))
+                    dist.all_reduce(num_samples_in_batch, reduce_operation="SUM")
+                    dist.all_reduce(num_tokens_in_batch, reduce_operation="SUM")
```
jbloxham, line 596: Food for thought: this pattern of converting an int to a tensor, reducing the tensor, then converting back to an int recurs quite a lot. Maybe add a new helper function to `dist`? Feel free to ignore.

jbloxham (approved):
Looks good to me! Glad to see the state getting simplified.

composer/trainer/trainer.py
ravi-mosaicml: Should it be a spac? :P

composer/callbacks/speed_monitor.py
ravi-mosaicml: Good catch, thanks!

composer/trainer/trainer.py
ravi-mosaicml: I think I'll leave it for now as dist.all_reduce operates in place, which is why it requires it to be a tensor. I could see a signature like `all_reduce(input, op) -> output`, but I think that would make it unclear that `input` would be modified in place (in the case that it's a tensor).

PR #231, opened 2022-01-13
Fixes apt sources bug fix

### LINKED ISSUE #222: Docker apt Package Sources Configured Multiple Times
Configuration in `mosaicml/pytorch_internal:latest`

W: Target Packages (Packages) is configured multiple times in /etc/apt/sources.list:50 and /etc/apt/sources.list.d/cuda.list:1

Averylamp:
1. Removed the duplicated sources, since they're included in the nvidia base image
2. Skipping the hack on the cpu build, since 1) the sources are no longer set, and 2) it's not needed!

Closes #222


ravi-mosaicml (changes requested):
Can you add this to the pytorch internal docker image under research2? The pytorch image here should not have this preconfigured.

ravi-mosaicml:
The changes for wget and the sources are fine; I'll merge it in cause I need to build the image manually.

Averylamp:
> Can you add this to the pytorch internal docker image under research2? The pytorch image here should not have this preconfigured.

I'm not sure what you're referring to here?

ravi-mosaicml:
> > Can you add this to the pytorch internal docker image under research2? The pytorch image here should not have this preconfigured.
> 
> I'm not sure what you're referring to here?

https://github.com/mosaicml/research2/blob/dev/docker/pytorch_internal/Dockerfile

Averylamp:
> > > Can you add this to the pytorch internal docker image under research2? The pytorch image here should not have this preconfigured.
> > 
> > 
> > I'm not sure what you're referring to here?
> 
> https://github.com/mosaicml/research2/blob/dev/docker/pytorch_internal/Dockerfile

Why isn’t this in the docker-files repo?
fyi Research2 is out of date and deprecated.  Just was a temporary repo for launch. It also doesn’t use git flow and just merges to main

ravi-mosaicml:
Yeah, this should be in dockerfiles....
It's on my todo list
We should mark research2 as archive

PR #232, opened 2022-01-14
Require `pip install -e` be `pip install --user -e` when running as root

ravi-mosaicml:
`pip install -e` doesn't work as root. `pip install --user -e` does.


bandish-shah (approved):
LGTM

PR #233, opened 2022-01-14
Clean up the tests to make them work on jenkins

ravi-mosaicml:
1. Override dataloader hparams to use 0 workers and be cpu-only (no memory pinning)
2. Lower the batch size of the dummy model so tests run faster.
3. Upgrade yapf
4. Fix the conditional import of `transformers`, `libcloud`, and `deepspeed` so `pip install composer[dev]` works
5. Disable wandb in the wandb test
6. Speed up the nlp metric tests by skipping larger inputs
7. Increased timeouts where needed
8. Fix a deepspeed test that was incorrectly marked as gpu


tests/fixtures/distributed_fixtures.py
```diff
@@ -43,35 +41,25 @@ def configure_dist(request: pytest.FixtureRequest):
...
+        if "RANK" in os.environ and "WORLD_SIZE" in os.environ:
+            torch.distributed.init_process_group(backend, timeout=DIST_TIMEOUT)
+        else:
+            store = torch.distributed.HashStore()
+            torch.distributed.init_process_group(backend, timeout=DIST_TIMEOUT, store=store, world_size=1, rank=0)
 
 
 @pytest.fixture(autouse=True)
-def wait_for_all_procs(subfolder_run_directory: None):
+def wait_for_all_procs(subfolder_run_directory: None, configure_dist: None):
```
jbloxham, line 57: Is it necessary to explicitly include `configure_dist` when it has `autouse=True`?

jbloxham (approved):
Looks good to me, as long as Jenkins is happy!

tests/fixtures/distributed_fixtures.py
ravi-mosaicml: Maybe not strictly, but this helps set the ordering with the autouse -- so dist will be configured before this fixture is called.

Averylamp (approved):
lgtm

PR #236, opened 2022-01-14
Remove custom W&B config changes

siriuslee:
Removes custom 'algorithms' and 'models' sections from W&B

### DIFF
--- a/composer/loggers/logger_hparams.py
+++ b/composer/loggers/logger_hparams.py
@@ -116,7 +116,7 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
 
         if config is not None:
             if "algorithms" in config:
-                algos_dict_list = config.get("algorithms", [])
+                algos_dict = config.get("algorithms", {})
                 algorithm_names = []
                 for algo_dict in algos_dict_list:
                     if len(algo_dict.keys()) != 0:
--- a/composer/loggers/logger_hparams.py
+++ b/composer/loggers/logger_hparams.py
@@ -115,31 +115,6 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
         tags = list(set([x.strip() for x in self.tags.split(",") if x.strip() != ""]))
 
         if config is not None:
-            if "algorithms" in config:
-                algos_dict = config.get("algorithms", {})
-                algorithm_names = []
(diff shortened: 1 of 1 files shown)


ravi-mosaicml (approved):
lgtm

Averylamp (changes requested):
lgtm, but also remove the custom flattening in the next 60 or so lines

hanlint:
@siriuslee could you provide more motivation/context for this change, and how it impacts users of w&b?

siriuslee:
> @siriuslee could you provide more motivation/context for this change, and how it impacts users of w&b?

This removes some custom W&B work we did to make organizing our data collection efforts easier as the underlying composer library changed. A lot of these changes are now unnecessary or would be better made explicitly by the user. One of these changes (replacing model config data with the model name) actually deleted important configuration data. 

As for the flattening, W&B will partially flatten the `config` dictionary that it's given for easier visualization and filtering runs. A nested key, after flattening, can be queried in W&B as `"a.b.c"`. This doesn't work for objects that are lists, however, like `schedulers`. Our custom flattener takes care of that. After talking a bit more with @Averylamp , I've added a switch `flatten_hparams` that can optionally flatten the config before uploading if the user desires. I've defaulted it to `False` but am open to `True` if we think that is a more common use case.

Averylamp (approved):
lgtm thanks just make sure checks pass

Averylamp:
> > @siriuslee could you provide more motivation/context for this change, and how it impacts users of w&b?
> 
> This removes some custom W&B work we did to make organizing our data collection efforts easier as the underlying composer library changed. A lot of these changes are now unnecessary or would be better made explicitly by the user. One of these changes (replacing model config data with the model name) actually deleted important configuration data.
> 
> As for the flattening, W&B will partially flatten the `config` dictionary that it's given for easier visualization and filtering runs. A nested key, after flattening, can be queried in W&B as `"a.b.c"`. This doesn't work for objects that are lists, however, like `schedulers`. Our custom flattener takes care of that. After talking a bit more with @Averylamp , I've added a switch `flatten_hparams` that can optionally flatten the config before uploading if the user desires. I've defaulted it to `False` but am open to `True` if we think that is a more common use case.

This was a hack I introduced during data collection to make Niklas’ data querying more consistent

PR #238, opened 2022-01-18
Fix the speed monitor

ravi-mosaicml:
Since the speed monitor calls `dist.all_reduce`, it must run on all ranks now.


Landanjs (approved):
LGTM!

PR #240, opened 2022-01-18
Fixed loggers and callbacks

### LINKED ISSUE #239: Remove `RankZeroCallback` and `RankZeroLogger`
The `RankZeroCallback` and `RankZeroLogger` cannot make any calls to `composer.utils.dist`. This has led to bugs where the entire training process hangs. Instead, it would be much cleaner to guard the rank-zero code directly in the few places where the `RankZeroCallback` and `RankZeroLogger` are used, and then these base classes can be removed.

ravi-mosaicml:
1. Removed rank zero callbacks and loggers, since these hid complexity and led to infinitely-blocking code when using distributed functions. Closes #239.
2. Incrementing `state.timer`  _before_ calling `.eval()` in the trainer. This helps ensure that the batch count is consistent for both batch-wise and epoch-wise evaluators. This batch is printed in the logs.
3. Fixed the TQDM logger so it works properly with gradient accumulation.
4. Removed `LogLevel.ALGORITHM`, `LogLevel.MICROBATCH`, and `LogLevel.VERBOSE` since these were rarely being used. Instead, the built-in python logger should probably be used for anything that is verbose (since it really wouldn't be a useful metric), MICROBATCH should use BATCH (since a MICROBATCH is like another gpu), and ALGORITHM should use batch or epoch, depending where it is being run.,
5. Updated the file logger to take a `log_interval` instead of `log_every_n_epochs` and `log_every_n_batches`, and a `flush_interval` instead of `flush_every_n_batches`.
6. Switched the default logger in all yamls to tqdm.

### DIFF
--- a/composer/core/callback.py
+++ b/composer/core/callback.py
@@ -8,7 +8,6 @@ import abc
 from typing import TYPE_CHECKING
 
 from composer.core.serializable import Serializable
-from composer.utils import dist
 
 try:
     from typing import final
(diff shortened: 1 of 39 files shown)


jbloxham (approved):
All looks good to me! Thanks for nixing the rank zero stuff!

PR #245, opened 2022-01-19
Remove the DDP DataLoader

### LINKED ISSUE #244: Remove the `DDPDataLoader`
Let's remove the `DDPDataLoader` and instead move its functionality directly into the training loop -- e.g.

```python3

for epoch in epochs:
  if isinstance(state.train_dataloader.sampler, DistributedSampler):
    state.train_dataloader.sampler.set_epoch(int(state.timer.epoch))
```

ravi-mosaicml:
Closes #244


jbloxham (approved):
Looks good - but I do wonder if down the road we could remove some of the repetition by repurposing the callbacks system in the trainer.

composer/trainer/trainer.py
```diff
@@ -807,6 +806,11 @@ def eval(self, is_batch: bool):
 
             metrics = self._get_metrics_as_collection(is_train=False)
 
+            if isinstance(self.state.eval_dataloader.sampler, torch.utils.data.DistributedSampler):
+                # using the batch as the epoch, as the evaluator may be running every batch
+                # instead of every epoch
+                self.state.eval_dataloader.sampler.set_epoch(int(self.state.timer.batch))
+
```
abhi-mosaic, line 813: What is this section of code meant for? The eval dataloader's sampler should probably not matter. And setting the epoch to the current batch seems a bit strange?

ravi-mosaicml: Discussed offline; updated comment.

ravi-mosaicml:
> Looks good - but I do wonder if down the road we could remove some of the repetition by repurposing the callbacks system in the trainer.

We discussed this a while back (I think before you started) and decided against this design so it becomes explicit in the training loop what's happening where.

PR #246, opened 2022-01-19
[WIP] Benchmarks Folder

A-Jacobson:
@coryMosaicML, @Landanjs, @growlix 

I just naively moved everything related to resnet50 and deep lab out of the main composer folder. Obviously none of this runs yet but you can clearly see the problems we have to address. 

1. All resnets pull from `resnets.py` for each benchmark to be entirely self contained we would have to redefine the architecture for each script.
2. We dont have a way to actually run these in this configuration since the entry point has been `examples/run_mosaic_trainer.py`. perhaps each benchmark needs it's own entrypoint?
3. our yamls aren't just configurations (hparams) they also define dataset and architecture objects. I'm not sure how we woul dmake this work with the `examples/adding_custom_models.py` style run script if that's the way we wanted to go.
4. Hparams interfaces?

Edit:
I tried to write a rough imagenet training script with composer. It's in `benchmarks/resnet50_i1k_script` ideally it could be called with `python train.py` or `composer train.py`. While working on this i realized moving `resnet.py` actually broke stochastic depth because that algorithm was pulling a module form that architecture. I hacked around stuff like that and deregistered all models from the trainer in this branch as well.

You'll notice the script version is not as flexible as the Hparams version in some sense (Even if i made it argparse friendly) as we can only change configurations such as lr, duration, ect. and not the defined objects (resnet, SGD, schedulers, datasets).


Averylamp:
Is this one stale?

A-Jacobson:
I'll be picking this up again this week or early next week.

PR #249, opened 2022-01-19
Optimizer Surgery

ravi-mosaicml:
* Added in optimizer surgery by porting the code from https://github.com/mosaicml/mosaicml/pull/284
* Added the optimizer as an optional argument into all functional algorithms that perform surgery, and passing through the state.optimizer via the class-based forms
* Updated `Trainer.__init__` to take `Optimizers` and `Schedulers` instead of `OptimizerHparams` and `SchedulerHparams`. Constructing the optimizers and schedulers before `Event.INIT`. (TODO: Merge `Event.INIT` and `Event.TRAINING_START`). Refactored some ancillary code around the `State` to support passing the optimizers and schedulers in upon State initialization. Moved construction of optimizers and schedulers from the hparams to `Trainer.create_from_hparams`.
* Updated `WarmUpLR` to default to epoch-wise scheduling, to match the `WarmUpLRHparams` which defaulted to epoch-wise scheduling. Otherwise, this would have been a breaking change in the trainer, which previously was taking the hparams.
* Updated the `SchedulerHparms` to return a `Scheduler` rather than a `(Scheduler, interval)`, so the result from this can be passed into `Trainer.__init__`. The `interval` is already bound as an instance variable `scheduler.interval`, so it was redundant to pass it as a tuple (and, the type annotations were incorrect all along I think).

Shoutout to @dblalock for the original optimizer surgery code

TODO:
- [x] Address remaining PR feedback
- [x] Run another wandb run


ravi-mosaicml:
Ran a wandb run to confirm that surgery is working. See https://wandb.ai/mosaic-ml/ravi_resnet50_test_blurpool_post_surgery?workspace=user-ravimosaicml

tests/algorithms/test_blurpool_algorithm.py
```diff
@@ -94,18 +96,39 @@ def test_blurpool(state, blurpool_instance, dummy_logger):
...
+    # old params removed
+    assert not surgery._tensor_in(orig_conv.weight, param_list)
+
+    # new params added
+    new_conv2d = new_conv.conv
+    assert isinstance(new_conv2d, torch.nn.Module)
+    new_weight = new_conv2d.weight
+    assert new_weight is not orig_conv.weight
+    assert isinstance(new_weight, torch.Tensor)
+    assert surgery._tensor_in(new_weight, param_list)
```
dblalock, line 134: Probably not worth prioritizing right now, but would be nice to split this into a helper func that could work for any surgery method and a method-specific test for each surgery method.

composer/core/surgery.py
```diff
@@ -142,3 +187,113 @@ def count_module_instances(model: torch.nn.Module, module_class: Type[torch.nn.M
...
+
+    Args:
+        param (torch.nn.parameter.Parameter): The parameter to search for.
+        optimizer (torch.optim.Optimizer): The optimizer to search within.
+    
+    Returns:
+        int: The index within `opt.param_groups` of the first group containing ``param``.
+
+    Raises:
+        RuntimeError: If the ``param`` is not in the ``opt`.
```
dblalock, line 222: Probably better to return -1 instead of throwing here; more consistent with e.g., string `find()`. Doesn't matter much since it's private, but it made reading the `test_optimizer_surgery_removed_params_gone` a little confusing. Although maybe there was some reason I (or you?) did it this way that I'm missing

tests/test_surgery.py
```diff
@@ -109,3 +116,56 @@ def from_linear(module: torch.nn.Module, module_index: int = -1):
             ret.weight.copy_(module.weight)  # type: ignore
             ret.bias = module.bias  # same param object
         return ret
+
+
+@pytest.fixture
+def optimizer_surgury_state():
```
dblalock, line 122: surgury -> surgery

dblalock (commented):
Just a few local nits. Only two really touch logic:

1. Lower priority: Maybe have `surgery._find_param_in_optimizer` return -1 when not found instead of throwing. (which I think was originally my weird design choice?)
2. Higher priority: I'm worried about calling `surgery.update_params_in_optimizer` in trainer `__init__` after moving the model to the device. I think this will make it impossible to pass in an optimizer with multiple param groups, although I'm not certain.

composer/trainer/trainer.py
```diff
@@ -340,7 +328,17 @@ def __init__(
...
         if not self.deepspeed_enabled:
+            host_model_params = self.state.model.parameters()
             self.state.model = self.device.module_to_device(self.state.model)
+            device_model_params = self.state.model.parameters()
+
+            # use surgery to update the parameters of the optimizers, now that the model is on the device
+            # see https://pytorch.org/docs/stable/optim.html#constructing-it
+            surgery.update_params_in_optimizer(old_params=host_model_params,
+                                               new_params=device_model_params,
+                                               optimizers=self.state.optimizers)
```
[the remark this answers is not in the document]
hanlint: Hmm yeah I'd prefer a more elegant solution to this, if possible.

tests/fixtures/models.py
```diff
@@ -34,11 +34,15 @@ def __init__(self, in_shape: Tuple[int, ...], num_classes: int) -> None:
...
+
+        self.fc2 = torch.nn.Linear(5, num_classes)
+
         self.net = torch.nn.Sequential(
             torch.nn.Flatten(),
-            torch.nn.Linear(in_features_flattened, 5),
+            self.fc1,
             torch.nn.ReLU(),
-            torch.nn.Linear(5, num_classes),
+            self.fc2,
```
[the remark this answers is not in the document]
ravi-mosaicml: Really good point, thanks!

tests/test_surgery.py
ravi-mosaicml: Whoops thanks!

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Really good point, I think it's fine to rip out the parameter groups and just add them back in as @dblalock suggested.

composer/core/surgery.py
ravi-mosaicml: Yup, I did refactor it -- I prefer to raise exceptions than to use return codes, but just switched it back -- agree it makes sense to be similar to `find`

composer/trainer/trainer.py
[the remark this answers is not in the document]
jbloxham: I believe I've fixed this issue now with a new function for swapping out the parameters.

jbloxham:
New WandB run in the same project: https://wandb.ai/mosaic-ml/ravi_resnet50_test_blurpool_post_surgery?workspace=user-ravimosaicml

Looking identical so far.

dblalock (commented):
Code basically LGTM; just a few suggested tweaks to polish it. 

But, from the wandb results, it looks like we have a 1.4% accuracy regression with blurpool; [77% when I first debugged Hanlin's port](https://github.com/mosaicml/mosaicml/pull/299) down to [75.6% now](https://wandb.ai/mosaic-ml/ravi_resnet50_test_blurpool_post_surgery?workspace=user-dblalock), which is 1% lower than our baseline.

This is probably not because of this PR based on Ravi's earlier experiment, but I launched some more rn50 debugging runs with configs I trust to independently make sure. I don't think I have the roadmap context to trade off possible regression chance vs getting this in sooner, so I'll defer to @hanlint on whether to go ahead and approve this.

hanlint:
@dblalock : The 75.6% vs 77% is because the runs above are using `ssr=0.3`, whereas the 77% was at `ssr=1`, so not an issue here.

dblalock:
ah, didn't notice that. Awesome! Then we have no problem

composer/core/surgery.py
```diff
@@ -142,3 +188,159 @@ def count_module_instances(model: torch.nn.Module, module_class: Type[torch.nn.M
...
+    """
+    for i, group in enumerate(optimizer.param_groups):
+        param_list: List[torch.nn.parameter.Parameter] = group['params']
+        if _tensor_in(param, param_list):
+            return i
+
+    return -1
+
+
+def update_params_in_optimizer(old_params: Iterable[torch.nn.parameter.Parameter],
```
ravi-mosaicml, line 231: Thoughts about replacing this function with something that takes the output of `replace_module_classes`? I lm thinking then we would have a new function for explicitley adding parameters E.g. something like:

```python3
def update_params_in_optimizer(old_params_to_replaced_params: Dict[nn.Parameters, Optional[nn.Parameters]])
  ...

def add_paramaters_to_optimizer(param_group_of: nn.Parameter ,new_parameters: Iterable[parameters])
  ...
```

The first function could handle replqcements / removals (if the value of a dict is none, thrn that would act as a removal) across parameters groups, and the second one could add in new parameters. For the 2nd function, you would pass in a parameters that is used to infer the parameter group.

dblalock: I instinctively like the idea of having something directly take in the output of `replace_module_classes`. I'm not sure when we'd call it though.

I'm mildly skeptical of having a public function to support adding parameters to a specific param group, as opposed to only adding them to "the same group as these other params I'm giving you". Mostly because:

1. I don't think the caller should have to keep track of param groups, especially since we don't return them (and I wouldn't be surprised if many users didn't even know about param_groups).
2. Conditioned on the user not managing param_groups, they should probably be calling `add_param_group` when adding params to stay within the optimizer's public interface.

Brainstorming a little bit, I think (?) we have two use cases:

1. Making module replacement not ruin the set of tensors being optimized
2. Making moving params from one device to another not ruin the set of tensors being optimized

(2) could be handled in the same manner as (1), but it might be good to just replace the groups wholesale to depend a bit less on implementation details. If there are other use cases, that would definitely suggest building more general functionality.

So...I don't actually know the ideal solution, but that's my 2c.

hanlint: Agreed with @dblalock that a public function to `add_param_group` is duplicative of the optimizer's public interface. Going to close this API discussion, can file an issue if someone feels strongly here.

composer/core/surgery.py
```diff
@@ -299,3 +299,47 @@ def update_params_in_optimizer(old_params: Iterable[torch.nn.parameter.Parameter
...
+    
+    This differs from `update_params_in_optimizer` in that this method is capable
+    of replacing parameters spanning multiple param groups. To accomplish this,
+    this function assumes that parameters in `new_params` should inherit the
+    param group of the corresponding parameter from `old_params`. Thus, this
+    function also assumes that `old_params` and `new_params` have the same length.
+    Args:
+        old_params: Existing parameters of this optimizer.
+        new_params: New parameters of the optimizer, given in the same order as
+            `old_params`.
```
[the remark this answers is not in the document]
jbloxham: Yeah I'm struggling a bit with phrasing this clearly, but I hope that the rest of the docstring will make clear that this is supposed to be ALL the new params, not just a diff.

composer/core/surgery.py
```diff
@@ -299,3 +299,47 @@ def update_params_in_optimizer(old_params: Iterable[torch.nn.parameter.Parameter
     new_param_list += list(added_params)
     log.info(f'adding {len(added_params)} new parameters to parameter group #{group_idx}')
     param_group['params'] = new_param_list
+
+
+def replace_params_in_optimizer(old_params: Iterable[torch.nn.parameter.Parameter],
+                                new_params: Iterable[torch.nn.parameter.Parameter], optimizers: Optimizers) -> None:
+    """Fully replaces parameters from an optimizer.
```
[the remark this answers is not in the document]
jbloxham: Fair point - I could move this into the trainer-specific code. I like keeping it here just because it's semantically related to surgery, and it seems likely we may refactor surgery some day so that `update_params_in_optimizer` and `replace_params_in_optimizer` can be unified someday. And who knows, maybe someone will write an algorithm some day that actually requires `replace_params_in_optimizer`.

hanlint (approved):
LGTM

PR #257, opened 2022-01-20
Fix broken tests

ravi-mosaicml:
1. Removed `test_mnist_real_dataset` since it's slow (downloads the entire mnist) and doesn't add much value
2. Refactored `_convert_time_fields` in the scheduler so it doesn't modify the dataclass in place. The issue with in-place modification was that checkpoint saver was saving the hparams dataclass after the fields have been converted. Then, when resuming from checkpoint, the already-converted fields were no longer valid time strings, so it crashed.
3. Added `requests` to the base dependency (it is used by the checkpoint loader)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ install_requires = [
     "torchvision>=0.9.0",
     "torch>=1.9",
     "yahp>=0.0.14",
+    "requests>=2.26.0",
     "numpy==1.21.5",
 ]
 extra_deps = {}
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -50,6 +50,7 @@ def package_files(directory: str):
     "torchvision>=0.9.0",
     "torch>=1.9",
     "yahp>=0.0.14",
+    "requests>=2.26.0",
```
ajaysaini725, line 53: 🎉

ajaysaini725 (approved):
LGTM!

PR #258, opened 2022-01-20
Fix bug with AFTER_DATALOADER event; remove microbatches from state

ravi-mosaicml:
1. #223 introduced a bug where algorithms that run on the AFTER_DATALOADER and did a not-in-place modification of state.batch did not also update state.microbatches (which was used for training), so these algorithms were effectively ignored. Fixed this bug by computing the microbataches AFTER the Event.AFTER_DATALOADER event.

2. Removed the `microbatches` and `microbatch_idx` from the state. Instead, algorithms that need to run on smaller batch sizes should use the Event.BATCH_START event instead of Event.AFTER_DATALOADER, since Event.BATCH_START will get the forward-pass sized batch.

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -57,8 +57,6 @@ SKIP_SERIALIZATION_FIELDS = [
     "batch",
     "batch_num_samples",
     "batch_num_tokens",
-    "microbatches",
-    "microbatch_idx",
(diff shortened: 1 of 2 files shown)


coryMosaicML (approved):
LGTM

PR #262, opened 2022-01-21
Timm support

A-Jacobson:
- Timm model wrapper with MosaicClassifier
- Timm Hparams interface
- Timm resnet50 yaml
- working W&B run

@hanlint , @Landanjs


hanlint (commented):
Overall, this design looks good. We might want a common schema across huggingface and TIMM though. So for hugging-face, our models would then be:
```
model
  hf:
     model_name: '<name_here>'
```

Would that work for you @moinnadeem ?

A-Jacobson:
Just checked out HF, looks like they have some model factories such as `AutoModelForMaskedLM` but they're divided by task. Timm is all image classification on imagenet so it's more straightforward. We'd probably have to have a wrapper for each factory so hf may look like 

masked language modeling
```
model
  hf-maskedlm:
     model_name: '<name_here>'
```
question answering
```
model
  hf-qa:
    model_name
 ```
and so on.

**Comment by hanlint** [comment]
Does `model_name` always map to a unique model factory type in HF? If so, we could just require `hf, model_name` and auto-infer the factory type.

**Comment by A-Jacobson** [comment]
I don't think so. I'm seeing "Bert-base" show up in multiple examples. Though, I'm not sure of the differences between tasks. It may just be different weights or a different classifier head.

File: composer/models/timm/model.py [hunk 1]
```diff
@@ -0,0 +1,44 @@
+from typing import Optional
+
+from composer.models.base import MosaicClassifier
+
+
+class Timm(MosaicClassifier):
+    """A wrapper around timm.create_model() used to create mosaic classifiers from timm models
+    Args:
+        model_name (str) timm model name e.g:'resnet50'list of models can be found at https://github.com/rwightman/pytorch-image-models
```
Landanjs, line 9: Nit:
```suggestion
        model_name (str): timm model name e.g. 'resnet50', list of models can be found at https://github.com/rwightman/pytorch-image-models.
```

composer/models/timm/model.py
```diff
@@ -0,0 +1,44 @@
...
+        pretrained (bool) imagenet pretrained. default: False
+        num_classes (int) The number of classes.  Needed for classification tasks. default: 1000
+        drop_rate (float) dropout rate. default: 0.0
+        drop_path_rate (float) drop path rate (model default if None). default: None
+        drop_block_rate (float) drop block rate (model default if None). default: None
+        global_pool (str) Global pool type, one of (fast, avg, max, avgmax, avgmaxc). Model default if None. default: None
+        bn_momentum (float) BatchNorm momentum override (model default if not None). default: None
+        bn_eps (float) BatchNorm epsilon override (model default if not None). default: None
```
Landanjs, line 17: You might need a colon after all types e.g. `(float):`. Not sure if this effects docstrings...

composer/models/timm/model.py
```diff
@@ -0,0 +1,44 @@
...
+        self,
+        model_name: str,
+        pretrained: bool = False,
+        num_classes: int = 1000,
+        drop_rate: float = 0.0,
+        drop_path_rate: Optional[float] = None,
+        drop_block_rate: Optional[float] = None,
+        global_pool: Optional[str] = None,
+        bn_momentum: Optional[float] = None,
+        bn_eps: Optional[float] = None,
```
Landanjs, line 30: What do you think about adding `**kwargs` argument to capture the various arguments across models? I'm not sure if yahp could use this right now, but seems good for futureproofing.

Landanjs (commented):
LGTM! 2 comments on docstrings and a question on if we should include a `**kwargs` argument.

@mosaicml/research-engineering does anyone know if we could have a `kwargs`-like argument in yahp? Maybe defining a dict hparam `kwargs: dict = hp.optional(default={})`, then using it in a yaml like: `kwargs.image_size: 512`?

jbloxham:
> LGTM! 2 comments on docstrings and a question on if we should include a `**kwargs` argument.
> 
> @mosaicml/research-engineering does anyone know if we could have a `kwargs`-like argument in yahp? Maybe defining a dict hparam `kwargs: dict = hp.optional(default={})`, then using it in a yaml like: `kwargs.image_size: 512`?

YAHP supports JSON types, at least to some extent. I don't have any experience using it.

moinnadeem:
> I don't think so. I'm seeing "Bert-base" show up in multiple examples. Though, I'm not sure of the differences between tasks. It may just be different weights or a different classifier head.


Yeah, the `bert-base-uncased` is a single set of weights, but doing `AutoModelForClassification.from_pretrained("bert-base-uncased")` will create some randomly initialized weights for the classifier head. 

In other words, there is a bijection between model names and weights.

tests/test_hparams.py
```diff
@@ -34,6 +34,7 @@ def _configure_dataset_for_synthetic(dataset_hparams: DatasetHparams) -> None:
 
 @pytest.mark.parametrize("hparams_file", walk_model_yamls())
 class TestHparamsCreate:
+    pytest.importorskip("timm")  # yapf: disable
```
hanlint, line 37: Based on the docs for `pytest.importorskip`, i think this test will be now be skipped for every model, not just the ones that require `timm`.

tests/test_load.py
```diff
@@ -39,6 +39,7 @@ def get_model_algs(model_name: str) -> List[str]:
 @pytest.mark.parametrize('model_name', model_names)
 @pytest.mark.timeout(15)
 def test_load(model_name: str):
+    pytest.importorskip("timm")
```
hanlint, line 42: Same as above.

tests/test_model_registry.py
```diff
@@ -8,6 +8,8 @@
 
 @pytest.mark.parametrize("model_name", model_registry.keys())
 def test_model_registry(model_name, request):
+    pytest.importorskip("timm")
```
hanlint, line 11: Same as above. Perhaps instead we do:
```
if model_name in ['timm']:
    pytest.importorskip("timm")
if model_name in ['unet']:
   pytest.importorskip("monai")
```
We can then skip the try, except below.
```

hanlint (commented):
Looks good, a few suggested changes. Do you have a w&b run with the TIMM resnet50 to verify convergence?

tests/test_hparams.py
A-Jacobson: you're right! I missed that. fixing.

A-Jacobson:
https://wandb.ai/mosaic-ml/timm-imagenet/reports/Shared-panel-22-01-30-20-01-86--VmlldzoxNTAzMTky

resnet50 convergence run to 76.89 accuracy. note on the yaml said `_quality = '76.51'` so it seems we're within the margin for error.

hanlint (approved):
LGTM

PR #263, opened 2022-01-21
Rename `Event.TRAINING_START` to `Event.FIT`; remove `Event.TRAINING_END`

ravi-mosaicml:
- Renamed `Event.TRAINING_START` to `Event.FIT_START` to denote that it is called on every call to `.fit()`, not just at the beginning of training (i.e. when the timer is 0)
- Moved algorithms currently running on `Event.TRAINING_START` (i.e. those modifying the model, optimizers, or schedulers) that should instead run on `Event.INIT` to `Event.INIT`. Renamed the others (that modify the dataloader) to run on `Event.FIT_START`
- Removed `Event.TRAINING_END`. It is not guaranteed that this event will fire (e.g. if the training process crashes). Nor is training really ever finished -- especially with pre-fitting and fine-tuning. In addition, for callbacks, all events in callbacks running on training end were redundant with `.close()` or `.post_close()`. Refactored SWA, the only algorithm using this event, to instead use `state.get_elapsed_duration() >= 1.0`.

TODO:
- [x] Create a separate PR to handle algorithm haparm fixes that replace required params that have template_default with optional parameters (#369)
- [x] Create a seperate PR to handle test fixes unrelated to idempotency (#371)
- [x] This PR should be left with the event rename and idempotency changes


ajaysaini725 (approved):
Logger + Event changes LGTM 👍

ravi-mosaicml:
(Let's hold on this one for now -- want to discuss with @ajaysaini725 how this would work with the trainer.eval)

ravi-mosaicml:
With #328, it will be safe to call the INIT event at the beginning of each call of `.fit` and `.eval`. This should resolve the concerns with trainer.eval.

ravi-mosaicml:
After discussion with @dblalock, there is a need to have an event that is run exactly once per trainer (i.e. for model surgery) vs for every call to fit. It does not make sense to combine these, as that would put the onus on the algorithm or callback developer to ensure idempotency.

Adjusted this PR to rename `Event.TRAINING_START` to `Event.FIT_START`, and leaving `Event.INIT` as-is to run once per trainer.

PR #264, opened 2022-01-21
add ability to specify custom run name, with rank auto-appended

dblalock:
As discussed verbally, this PR makes it possible to specify a custom wandb run name, instead of having it always set to the DDP rank. This is convenient for inspecting runs, and makes it feasible to use groupby only for splitting runs based on custom attributes.

[Wandb run](https://wandb.ai/mosaic-ml/c10r56/runs/2u23v77z/overview?workspace=user-dblalock) with this working.


composer/loggers/logger_hparams.py
```diff
@@ -186,10 +188,12 @@ def get_flattened_dict(data: Dict[str, Any], _prefix: List[str] = []) -> Dict[st
                 self.extra_init_params["config"] = {}
             self.extra_init_params["config"].update(flattened_config)  # type: ignore
 
+        name_suffix = f"Rank{dist.get_global_rank()}"
```
ravi-mosaicml, line 191:
```suggestion
        name_suffix = f"Rank {dist.get_global_rank()}"
```

composer/loggers/logger_hparams.py
```diff
@@ -89,7 +90,8 @@ class WandBLoggerBackendHparams(BaseLoggerBackendHparams):
     """
 
     project: Optional[str] = hp.optional(doc="wandb project name", default=None)
-    name: Optional[str] = hp.optional(doc="wandb group name", default=None)
+    group: Optional[str] = hp.optional(doc="wandb group name", default=None)
```
ravi-mosaicml, line 93: If the group is not set and `rank_zero_only` is True, could you set the group to the value of `name`? This will help make it backwards compatible with the existing YAMLs, so all ranks of the same job will be auto-grouped together.

ravi-mosaicml (approved):
LGTM -- see comment re: defaulting the group to the name if it is not set. Thanks!

composer/loggers/logger_hparams.py
dblalock: I actually wanted no space to match what custom run names usually look like, but I'll add it back in

PR #267, opened 2022-01-21
Fix Jenkins to work on PRs from Forks

ravi-mosaicml:
Fix jenkins to work on PRs from forks


nlsapp (approved):
`/head` is going to take the tip of the PR's branch, while I think `/merge` is going to take their tip and merge it into the destination branch

Not sure which one we care about most, but this LGTM

PR #272, opened 2022-01-25
Dramatically increase default dist_timeout

jbloxham:
Every single time I've seen `init_process_group` fail, it's just been because something in initialization is slow with high variance across ranks (usually some sort of network operation). This change will make our lives a bit easier. For reference, DeepSpeed's default is also 30 minutes, which is what I went with here.


hanlint (approved):
LGTM

PR #277, opened 2022-01-26
Allow `str` and `dict` in Trainer `init` signature

hanlint:
The Trainer `__init__` signature should minimize the need to import custom classes or enums from our codebase. This PR allows for:
* specifying device with 'cpu' or 'gpu' string instead of needing to import and supply `DeviceCPU` or `DeviceGPU`
* for precision, allows the use of `fp32`, `fp16`, and `amp`
* For `DeepspeedHparams`, allows passing in of `dict` and then initializing the dataclass using that dictionary

Also disabling the distributed backend check (@coryMosaicML ) to unblock notebook creators when running on GPU with the `gloo` backend.

### DIFF
--- a/composer/algorithms/__init__.py
+++ b/composer/algorithms/__init__.py
@@ -18,6 +18,8 @@ from composer.algorithms.cutout import CutOut as CutOut
 from composer.algorithms.cutout import CutOutHparams as CutOutHparams
 from composer.algorithms.dummy import Dummy as Dummy
 from composer.algorithms.dummy import DummyHparams as DummyHparams
+from composer.algorithms.factorize import Factorize as Factorize
+from composer.algorithms.factorize import FactorizeHparams as FactorizeHparams
 from composer.algorithms.ghost_batchnorm import GhostBatchNorm as GhostBatchNorm
 from composer.algorithms.ghost_batchnorm import GhostBatchNormHparams as GhostBatchNormHparams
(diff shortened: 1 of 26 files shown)


coryMosaicML (approved):
Looks good to me. Confirmed this does seem to fix the colab issue for me.

ravi-mosaicml:
@hanlint @coryMosaicML what was the error caused when training on gpu in a notebook? We do not reinitialize torch.distributed after it was initialized by the first training run, so if someone did a cpu run before switching to gpu, then they must restart the training process (i.e. notebook). This was because reinitializing the process group sometimes caused the training process to crash.

I think this change to initialize_dist would simply reuse the cpu backend even if the user specified gpu (or visa versa) on the second training run, without notifying the user. I'd prefer to keep the error here, or fix the underlying issue with switching backends (if possible).

The other changes look fine to me, though 
deepspeed_hparams needs an updated docstring for the dict option.

hanlint:
@ravi-mosaicml thanks! I made it a warning for now. The issue was that `torch.distributed` was `gloo`, and @coryMosaicML was using GPU which is hardcoded to use `nccl`.

ravi-mosaicml:
Can you elaborate why it is a warning instead of an error? I think this will now cause errors later when torch.distributed attempts to do dist functions on gpu tensors, since the gloo backend inst supported there.

PR #269, opened 2022-01-24
DeYAHPificiation

jbloxham:
WIP

Depends on https://github.com/mosaicml/yahp/pull/99

PR #281, opened 2022-01-26
Rename checkpoint API

hanlint:
With this change, checkpointing `Hparams` objects are no longer needed in the Trainer `__init__` signature. Instead, the checkpoint saving/loading can be enabled via these arguments:

```
def __init__(
            ...

            # load checkpoint
            load_path: Optional[str] = None,
            load_object_store_config: Optional[dict] = None,
            load_weights_only: bool = False,
            load_strict: bool = False,
            load_chunk_size: int = 1_048_576,
            load_progress_bar: bool = True,

            # save_checkpoint
            save_interval: int = 1,
            save_interval_unit: str = 'ep',

            ...)
```

This differs from the `CheckpointHparams` in a few ways:
(1) previously, the `CheckpointLoaderHparams` has the ambigious field `checkpoint`, which here is renamed to `load_path`. 
(2) all the arguments above have `load_` and `save_` prepended for separation.

This is WIP, seeking comments on the API and signature change from the reviewers before implementing the logic.


jbloxham:
I'm not sure I like this API. I like organizing Hparams into dataclasses where appropriate, because:
1) It allows us to express required fields within an optional dataclass, making it clear which fields you have to specify if you, say, want to use checkpointing
2) It reduces bloat within the trainer init signature

There's precedent in libraries like HF for using objects in this way.

For context, my plan for the deYAHPificiation of objects like `DeepSpeedHparams` and `LoadCheckpointHparams` was to just strip away all the YAHP-specific code, leaving them as plain dataclasses, removing the need for users to interact with YAHP.

hanlint:
One key principle of the `__init__` signature is that it must strive not to require users to import additional classes or objects. In the previous case, we require:
```
from composer.trainer.checkpoint_hparams import CheckpointLoaderHparams, CheckpointSaveHparams
from composer.utils.object_store import ObjectStoreHparams

trainer = Trainer(
    ...
    checkpoint_loader=CheckpointLoaderHparams(path=..., object_store_Hparams=ObjectStoreHparams(...), ...)
    checkpoint_save=CheckpointSaverHparams(folder=..,...)
    ...
)
```

This places extra burden on the user to remember where to import these dataclasses from (and there would be many.. for deepspeed, checkpoint_saver, etc.), and search through the docs at a different page to find the arguments.

HF has one long list of arguments in a separate class to reduce init args bloat (2 above) -- (https://huggingface.co/docs/transformers/v4.15.0/en/main_classes/trainer#transformers.TrainingArguments). I'm OK with this approach if the user just has to import one top-level dataclass for many of the training arguments, but would otherwise prefer flat and bloated to hierarchical but scattered.

ravi-mosaicml:
My hesitation with this design is that everything now becomes a training loop change. A new argument in the checkpoint loader now touches trainer.create_from_hparams and trainer.init. in addition, we now need to manually make sure that variable names do not conflict. As an alternative, would the "dict or data class" approach work instead?
(We could even use a typeddict to preserve typing and docstrings, without requiring another user import)

Nonetheless, if we went with this, I'd prefer to have the trainer init match the trainer hparams, so if we flatten it in init, then I'd also prefer to flatten it in trainer hparams and remove the checkpoint loader hparam classes and checkpoint saver hparam classes. Trainer.init can construct a cehckpointloader and checkpointsaver directly. Fewer things to keep in sync :)


Agree with @jbloxham and would love to see where #269 leads and see of that could also simplify things here.

hanlint:
After some discussion, we are going to go with usability (e.g. a flat API) and take the hit on maintainability. I'll create a separate issue for cleaning up the hparams, which may be taken care of #269 .

hanlint:
OK this is ready for review on code implementation cc: @jbloxham @ravi-mosaicml when you have a chance, thanks!

composer/trainer/checkpoint.py
```diff
@@ -324,22 +324,22 @@ class CheckpointSaver:
...
 
     Args:
-        checkpoint_folder (str): The path to store checkpoints in.
-        checkpoint_interval (int): The amount of time units to wait between checkpoints.
-        checkpoint_interval_unit (str): The unit (`"ep"` or `"it"`) that
-            `checkpoint_interval` should be measured in.
+        save_folder (str): The path to store checkpoints in.
+        interval (int): The amount of time units to wait between checkpoints.
+        interval_unit (str): The unit (`"ep"` or `"it"`) that
+            `interval` should be measured in.
```
growlix, line 330: Is there a reason this takes an `int interval` and `str interval_unit` instead of a single `Union[str, Time]` à la [`composer.Trainer(max_duration="5ep")`](https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L126)? Seems to me like we should try to keep the interface for time specification consistent.

composer/trainer/trainer.py
```diff
@@ -151,9 +153,18 @@ def __init__(
             log_destinations: Optional[Sequence[BaseLoggerBackend]] = None,
             callbacks: Sequence[Callback] = tuple(),
 
-            # Checkpoint hparams
-            checkpoint_loader: Optional[CheckpointLoaderHparams] = None,
-            checkpoint_saver: Optional[CheckpointSaverHparams] = None,
+            # load checkpoint
```
growlix, line 156: Docstring should be updated to reflect change in in checkpointing arguments. It still describes `checkpoint_loader` and `checkpoint_saver`. (lines 103-106).

jbloxham (approved):
Looks good to me!

composer/trainer/trainer.py
hanlint: Done!

composer/trainer/checkpoint.py
hanlint: Agreed, but will defer this change to later.

PR #284, opened 2022-01-27
Streaming vision datasets

knighton:


composer/datasets/cifar.py
```diff
@@ -0,0 +1,252 @@
...
+    train_shards: int = hp.optional('Training split shards', default=16)
+    val_shards: int = hp.optional('Validation split shards', default=8)
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams) -> DataLoader:
+        cifar10_mean = 0.4914, 0.4822, 0.4465
+        cifar10_std = 0.247, 0.243, 0.261
+
+        if self.is_train:
+            split = 'train'
+            size = 50_000
```
abhi-mosaic, line 89: Interesting... I didn't realize you could include '_' tokens.

composer/datasets/cifar.py
```diff
@@ -0,0 +1,252 @@
...
+                    transforms.ToTensor(),
+                    transforms.Normalize(cifar10_mean, cifar10_std),
+                ])
+            else:
+                transform = transforms.Compose([
+                    transforms.ToTensor(),
+                    transforms.Normalize(cifar10_mean, cifar10_std),
+                ])
+
+            urls = ['/datasets/web_cifar10/%s_%05d.tar' % (split, i) for i in range(n_shards)]
```
abhi-mosaic, line 120: I'm assuming this will be updated to accept an S3 bucket rather than local paths?

composer/datasets/cifar.py
```diff
@@ -0,0 +1,252 @@
...
+            size_per_device = size // dist.get_world_size()
+            dataset = WebDataset(urls, cache_dir=self.dataset_cache_dir,
+                                 cache_verbose=self.dataset_cache_verbose)
+            dataset = dataset.decode('pil').map_dict(jpg=transform).to_tuple('jpg', 'cls')
+            dataset = dataset.with_epoch(size_per_device).with_length(size_per_device)
```
abhi-mosaic, line 125: TODO for @abhi-mosaic: confirm that this works as expected, ideally with tests.

composer/datasets/imagenet.py
```diff
@@ -86,3 +88,142 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+    resize_size: int = hp.optional("resize size. Set to -1 to not resize", default=-1)
+    crop_size: int = hp.optional("crop size", default=224)
+
+    train_shards: int = hp.optional('Training split shards', default=1024)
+    val_shards: int = hp.optional('Validation split shards', default=128)
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams) -> DataLoader:
+        if self.is_train:
+            split = 'train'
+            size = 1_281_167
```
abhi-mosaic, line 173: Similar for all datasets, but I'm wondering if this should be set (or updated somewhere below) with the true size of the dataset (num_shards * samples_per_shard), otherwise there might be some inconsistency in how the Trainer measures time?

abhi-mosaic (commented):
Looks good so far! Left a few comments, looking forward to some end-to-end testing with streaming from S3 before approving this PR. Quick reminder that for the big datasets, we probably don't want to test more than 1 or 2 epochs because it will cost 

We can also postpone support for cache-ing to a future PR, since it is currently still broken in `webdataset`. Context: https://github.com/webdataset/webdataset/issues/131

composer/datasets/webdataset.py
```diff
@@ -0,0 +1,70 @@
...
+        samples = tqdm(samples, total=n_samples, leave=False)
+    for sample in samples:
+        out.write(sample)
+    out.close()
+    create_webdataset_meta(split_dir, n_samples, n_shards)
+
+
+def download_webdataset_meta(dataset_name: str, split: str) -> bytes:
+    '''Download a WebDataset meta file from S3.'''
+    url = f's3://mosaicml-internal-dataset-{dataset_name}/{split}/meta.json'
```
abhi-mosaic, line 44: I think this URL input needs to be lifted up to the `DatasetHparams` level, because the bucket path will end up being customer specific.

composer/datasets/webdataset.py
```diff
@@ -0,0 +1,70 @@
...
+    cmd = 'aws', 's3', 'cp', url, '-'
+    return subprocess.run(cmd, capture_output=True).stdout
+
+
+def load_webdataset(dataset_name: str, split: str, cache_dir: Optional[str] = None,
```
abhi-mosaic, line 49: Since this loading function is being used in all of our examples, it might be nice to add some asserts here, like `n_shards` is divisible by the number of CPU workers, that kind of thing. Or if you think there's a better place to do that check that would also be good. But I think it's important, otherwise with a poor setting of `num_workers`, we might lose shards / have some dataloaders hanging, etc.

ravi-mosaicml:
@knighton curious about the status of this PR?

knighton:
@ravi-mosaicml webdataset is giving weird issues

composer/datasets/hparams.py
```diff
@@ -80,3 +85,85 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+                    transforms.RandomHorizontalFlip(),
+                    transforms.ToTensor(),
+                    transforms.Normalize(self.channel_means, self.channel_stds),
+                ])
+            else:
+                split = 'val'
+                transform = transforms.Compose([
+                    transforms.ToTensor(),
+                    transforms.Normalize(self.channel_means, self.channel_stds),
+                ])
```
abhi-mosaic, line 158: It looks like this `JpgClsWebDatasetHparams` class is going to force some tyubg of augmentation steps for the vision datasets that use it: `CIFAR10/20/100` and `TinyImageNet200`. Could you check with the vision team (maybe @coryMosaicML) about whether this tying is OK? 

If we need more dataset-specific behavior maybe we can remove the abstraction and duplicate the code so each vision dataset is a standalone. I think it's only ~50 lines or so in the `initialize_object`.

composer/datasets/hparams.py
```diff
@@ -80,3 +85,85 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+            dataset, meta = load_webdataset(self.dataset_s3_bucket, self.dataset_name, split, self.webdataset_cache_dir,
+                                            self.webdataset_cache_verbose)
+            if self.shuffle:
+                dataset = dataset.shuffle(512)
```
abhi-mosaic, line 162: Can we have an hparam for the shuffle buffer size? Same for all the other webdatasets too. I think it may also be important to let the user know that the buffer size is **per worker**. 

I'm also curious how much this slows down the training init time? If each worker has to wait for 512 samples to arrive, that could be a good chunk of the dataset that needs to be downloaded before a single sample is returned. Just food for thought, and another reason why it could be good to have it as an hparam.

composer/datasets/webdataset.py
```diff
@@ -0,0 +1,124 @@
...
+        text = download_webdataset_meta(dataset_s3_bucket, split)
+    meta = json.loads(text)
+    max_shard = meta['n_shards'] - 1
+    shards = f'{{{0:05d}..{max_shard:05d}}}.tar'
+    urls = f'pipe: aws s3 cp s3://{dataset_s3_bucket}/{split}/{shards} -'
+    dataset = WebDataset(urls, cache_dir=cache_dir, cache_verbose=cache_verbose)
+    return dataset, meta
+
+
+def size_webdataset(dataset: WebDataset, n_shards: int, samples_per_shard: int, n_devices: int, workers_per_device: int,
```
abhi-mosaic, line 84: We will have to instruct users to use this function when they construct their custom webdatasets... I'm wondering if we can also do some auto detection in the `Trainer` just to catch mistakes. Currently the `Trainer` will probably try to call `len(dataloader)` at some point and it will fail on a non-sized Webdataset, but the warning will be uninformative. Curious if we can do better and print something like `Detected unsized webdataset... please use "size_webdataset(...)..."`

setup.py
```diff
@@ -101,6 +101,10 @@ def package_files(directory: str):
     'deepspeed>=0.5.5',
 ]
 
+extra_deps['webdataset'] = [
+    'awscli>=1.22.60', 'wurlitzer>=3.0.2', 'webdataset @ git+https://github.com/mosaicml/webdataset.git@dev'
```
abhi-mosaic, line 105: Looks good! Just for posterity, I discussed with @hanlint last week that we'll use this Github url -based install for our webdataset fork, no official PyPI yet.

composer/datasets/cifar.py
```diff
@@ -65,3 +66,48 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
                                                     batch_size=batch_size,
                                                     sampler=sampler,
                                                     drop_last=self.drop_last)
+
+
+@dataclass
+class CIFAR10WebDatasetHparams(JpgClsWebDatasetHparams):
+    """Defines an instance of the CIFAR-10 WebDataset for image classification."""
+
+    dataset_s3_bucket = 'mosaicml-internal-dataset-cifar10'
```
abhi-mosaic, line 75: I think these MosaicML-specific bucket paths should probably be removed, or replaced with a more abstract 'your-s3-bucket' string.

abhi-mosaic (commented):
Overall this is looking pretty good to me! Left some comments for `shuffle_buffer_size` hparams, but nothing major.

I think the next step would be to coordinate with Vision team, maybe @coryMosaicML, and see what kind of convergence tests we should do on all these webdatasets? E.g. ResNet50 + WebImageNet1k vs. our current baseline.

I expect there will be none-to-very-small differences in convergence, but the runs' x-axis will look slightly different because these streaming datasets are slightly smaller than the original size (constraints of sharding). In a way the switch from `localdisk` -> `webdataset` will be a "reset" of our benchmarks, but I hope it won't be a major disruption to the research we have underway.

composer/datasets/ade20k.py
```diff
@@ -345,3 +347,126 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
                                                                         collate_fn=collate_fn,
                                                                         drop_last=self.drop_last),
                         device_transforms=device_transform_fn)
+
+
+@dataclass
+class ADE20kWebDatasetHparams(WebDatasetHparams, SyntheticHparamsMixin):
```
hanlint, line 353: I dont think these need to support the `SyntheticHparamsMixin`. We already have a normal dataset with synthetic support correct?

composer/datasets/create/ade20k.py
```diff
@@ -0,0 +1,68 @@
+from argparse import ArgumentParser, Namespace
```
hanlint, line 1: defer to @ravi-mosaicml how we want to handle these dataset conversion scripts.

composer/datasets/cifar.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
```
hanlint, line 1: is there a reason we renamed `cifar10` -> `cifar`?

composer/datasets/cifar.py
```diff
@@ -0,0 +1,168 @@
...
+                                                    sampler=None,
+                                                    drop_last=self.drop_last)
+
+
+@dataclass
+class CIFAR10WebDatasetHparams(CIFARWebDatasetHparams):
+    """Defines an instance of the CIFAR-10 WebDataset for image classification."""
+
+    dataset_s3_bucket = 'mosaicml-internal-dataset-cifar10'
+    dataset_name = 'cifar10'
```
hanlint, line 131: need to use `hp.required` and `hp.optonal` with `import yahp as hp` for these to work. See the other hprams objects for reference.

composer/yamls/models/web/classify_mnist.yaml
```diff
@@ -0,0 +1,41 @@
+train_dataset:
```
hanlint, line 1: I dont think we need all these yamls, as it'll be difficult to main both web and non-web versions.

hanlint: Maybe just keep _one_ for reference.

knighton: Okay, removed web/ directory, keeping the Imagenet example.

composer/datasets/cifar.py
knighton: I trivially added webdataset versions of cifar20 and cifar100 while converting cifar10. This is designed as base class CIFARWebDatasetHparams with three subclasses with only different class variables defining remote bucket, n_classes, etc.

Might as well have all three? I think having webdataset cifar10 is primarily useful for us as a guide for external users looking to create their own webdatasets. Alternatively we could remove cifar20 and cifar100 without pain (and rename back to cifar10.py), or remove all the webdataset cifar stuff due to their small size.

composer/datasets/cifar.py
knighton: Sorry, what do you do when they're meant to be constants?

composer/datasets/cifar.py
hanlint: ah ok makes sense!

hanlint (commented):
This is looking good! One major comment, but we can defer this to another PR --
for those folks who are not using YAHP, we need a way to directly instanitate the webdatset object for each dataset.

This essentially means moving much of the logic of `initialze_object` into it's own object. e.g.
```
class ImageNetWebDataset(..)
```
and then having:

```
def initialze_object(..):
    return ImageNetWebDataset(..)
```

See the C4Dataset code for an example. Again this can be deferred to a subsequent PR if needed.

composer/datasets/ade20k.py
knighton: Done (78bae50bbf903ed01d858cc4562073963476798d)

composer/datasets/hparams.py
knighton: Done

composer/datasets/cifar.py
knighton: Done

composer/datasets/webdataset.py
knighton: Redesigned to wrap loadwds/sizewds calls together in order to prevent this unpleasantness

Landanjs:
I started to test this and talked to James about these comments, but I thought I should add them here:
1. `pip install --user -e .[all]` failed due to conflicting dependencies: `ERROR: Cannot install awscli==1.22.67, mosaicml and mosaicml[all]==0.4.0 because these package versions have conflicting dependencies.`
2. If a download fails, then is retried, it will use the folder from the failed attempt and fail again. James might have adjusted this one?
3. It appears the calculation for number of iterations is off. The training ends in ~900 fewer iterations (~1%) and the LR schedule is also adjusted:

image

The pink line is webdatasets and the blue line is my original runs. The performance seems to be a bit worse, but it is possible that it is within noise. I would need to do more runs, but I can try ResNet50 ImageNet first.

knighton:
Yup, item 2 is fixed in commit 25a3dd55ca2e0f20362b04de322c40d643b07886

It now errors out with this message and does not alter the cache dir

> AssertionError: Download failed, check your credentials?

knighton:
item 1 is fixed in commit b6ead1a3ec628a5388a427be11677bd6bc4ae106

for item 3, some minor differences are expected behavior. these sharded streaming datasets are new datasets

abhi-mosaic:
> I started to test this and talked to James about these comments, but I thought I should add them here:
> 
> 1. `pip install --user -e .[all]` failed due to conflicting dependencies: `ERROR: Cannot install awscli==1.22.67, mosaicml and mosaicml[all]==0.4.0 because these package versions have conflicting dependencies.`
> 2. If a download fails, then is retried, it will use the folder from the failed attempt and fail again. James might have adjusted this one?
> 3. It appears the calculation for number of iterations is off. The training ends in ~900 fewer iterations (~1%) and the LR schedule is also adjusted:
> 
> image
> 
> The pink line is webdatasets and the blue line is my original runs. The performance seems to be a bit worse, but it is possible that it is within noise. I would need to do more runs, but I can try ResNet50 ImageNet first.

@Landanjs Just want to confirm that the # samples is expected to be lower, this is due to the constraints of sharded datasets. Basically we need an equal number of samples per shard, so `n_samples % n_shards == 0` and to accomplish this we truncate the dataset. So we have lost a few examples for each vision dataset but I think @knighton was careful to keep the loss small to around ~1%.

This probably means that baselines need to be re-run, for all vision tasks. I would love to know if the results are meaningfully different or within the noise! I'm hoping the latter :)

tests/test_dataset_registry.py
```diff
@@ -66,6 +66,9 @@
 
 @pytest.mark.parametrize("dataset_name", dataset_registry.keys())
 def test_dataset(dataset_name: str, dummy_dataloader_hparams: DataloaderHparams) -> None:
+    # Skip WebDatasets.
+    if dataset_name.startswith('wds_'):
+        return
```
hanlint, line 71:
```suggestion
        pytest.skip("Skipping test for webdatasets")
```

tests/test_dataset_registry.py
```diff
@@ -66,6 +66,9 @@
 
 @pytest.mark.parametrize("dataset_name", dataset_registry.keys())
 def test_dataset(dataset_name: str, dummy_dataloader_hparams: DataloaderHparams) -> None:
+    # Skip WebDatasets.
+    if dataset_name.startswith('wds_'):
```
hanlint, line 70: instead of by name, maybe check on whether the hparams is a subclass of the webdataset hparams?

composer/datasets/webdataset.py
```diff
@@ -0,0 +1,191 @@
...
+    for sample in samples:
+        out.write(sample)
+    out.close()
+    create_webdataset_meta(split_dir, n_samples, n_shards)
+
+
+def init_webdataset_meta_from_s3(remote: str, split: str) -> bytes:
+    """Read a WebDataset meta file from S3."""
+    url = f'{remote}/{split}/meta.json'
+    cmd = 'aws', 's3', 'cp', url, '-'
```
hanlint, line 72: instead of doing aws s3, perhaps use their AWS SDK (boto) instead.

tests/test_dataset_registry.py
knighton: Done

tests/test_dataset_registry.py
knighton: Done

hanlint (approved):

PR #285, opened 2022-01-27
Profiler fixes and guide

bandish-shah:
* Fix bug where profilers were not getting registered as callbacks
* Add user guide/tutorial documentation


bandish-shah:
Refactoring in #422

PR #289, opened 2022-01-28
Added ResNet20

growlix:
Carbon-copy of CIFAR10 ResNet56, but with 9 residual blocks instead of 27. Hparams are also duplicated from ResNet56 and not optimized.


A-Jacobson (approved):
LGTM, super straight forward.

Issue #265, opened 2022-01-21
Run `Event.BATCH_END` and `Event.EPOCH_END` after the timer is incremented


ravi-mosaicml:
We currently increment the timer (e.g. batch count, num samples, and num tokens trained) after `Event.BATCH_END`. This has the property that the timer in the before event will be the same as the timer in the after event. However, this property makes things more challenging,  since:
* Algorithms and callbacks cannot easily measure how much training has elapsed between the before and after event
* Algorithms and callbacks have to manually adjust the timer to know how much training actually occurred for the `batch_end` and `epoch_end` events.

Instead, the timer should be incremented before `Event.BATCH_END` and `Event.EPOCH_END`. While this would be a breaking change, it would significantly help clean up the code and align more with the expected behavior of the timer, such as in the following places:

* The speed monitor and benchmarker use `composer.utils.dist.all_reduce` on `Event.BATCH_END` to get the number of samples trained in the batch. (https://github.com/mosaicml/composer/blob/4ec0544002b7eaeddbae24716fe6fbde01e8d761/composer/callbacks/speed_monitor.py#L82-L90) With this change, it could instead read the timer in the `before_batch` event, and subtract this value from the timer in the `after_batch` event. This would avoid an extra communication sync  (since the timer is already invoking a distributed reduction) and be cleaner code to follow, since callbacks should ideally not have to be DDP aware.
* SWA would like to run on the training end, which SHOULD be equivalent to when `state.get_elapsed_duration() == 1` (https://github.com/mosaicml/composer/blob/4ec0544002b7eaeddbae24716fe6fbde01e8d761/composer/algorithms/swa/swa.py#L141). However, no event (after #263 is implemented) fires then. However, with this change, SWA could check for `state.get_elapsed_duration() == 1` at `Event.EPOCH_END`.
* Callbacks that run on a frequency, like the run directory uploader, file logger, etc... use code like `if (state.timer.batch + 1) % self.interval == 0: do_something()` to fire on the end events. Instead, with this change, the `+1` would no longer be needed, since the timer would accurate reflect the training progress.

PR #310, opened 2022-01-31
Run `Event.BATCH_END` and `Event.EPOCH_END` after the timer is increm…

ravi-mosaicml:
Implements issue #265.

We currently increment the timer (e.g. batch count, num samples, and num tokens trained) after `Event.BATCH_END`. This has the property that the timer in the before event will be the same as the timer in the after event. However, this property makes things more challenging,  since:
* Algorithms and callbacks cannot easily measure how much training has elapsed between the before and after event
* Algorithms and callbacks have to manually adjust the timer to know how much training actually occurred for the `batch_end` and `epoch_end` events.

Instead, the timer should be incremented before `Event.BATCH_END` and `Event.EPOCH_END`. While this would be a breaking change, it would significantly help clean up the code and align more with the expected behavior of the timer, such as in the following places:

* The speed monitor and benchmarker use `composer.utils.dist.all_reduce` on `Event.BATCH_END` to get the number of samples trained in the batch. (https://github.com/mosaicml/composer/blob/4ec0544002b7eaeddbae24716fe6fbde01e8d761/composer/callbacks/speed_monitor.py#L82-L90) With this change, it could instead read the timer in the `before_batch` event, and subtract this value from the timer in the `after_batch` event. This would avoid an extra communication sync  (since the timer is already invoking a distributed reduction) and be cleaner code to follow, since callbacks should ideally not have to be DDP aware.
* SWA would like to run on the training end, which SHOULD be equivalent to when `state.get_elapsed_duration() == 1` (https://github.com/mosaicml/composer/blob/4ec0544002b7eaeddbae24716fe6fbde01e8d761/composer/algorithms/swa/swa.py#L141). However, no event (after #263 is implemented) fires then. However, with this change, SWA could check for `state.get_elapsed_duration() == 1` at `Event.EPOCH_END`.
* Callbacks that run on a frequency, like the run directory uploader, file logger, etc... use code like `if (state.timer.batch + 1) % self.interval == 0: do_something()` to fire on the end events. Instead, with this change, the `+1` would no longer be needed, since the timer would accurate reflect the training progress.


composer/core/callback.py
```diff
@@ -200,6 +200,11 @@ def after_train_batch(self, state: State, logger: Logger) -> None:
     def batch_end(self, state: State, logger: Logger) -> None:
         """Called on the :attr:`~Event.BATCH_END` event.
 
+        .. note::
+
+            The :attr:`state.timer.batch` is incremented between `:attr:`~Event.BATCH_START` and
+            :attr:`~Event.BATCH_END`.
```
jbloxham, line 206: Maybe, incremented just before `~Event.BATCH_END`? Makes it a bit more clear when it is incremented.

composer/core/callback.py
```diff
@@ -211,6 +216,12 @@ def batch_end(self, state: State, logger: Logger) -> None:
     def epoch_end(self, state: State, logger: Logger) -> None:
         """Called on the :attr:`~Event.EPOCH_END` event.
 
+
+        .. note::
+
+            The :attr:`state.timer.epoch` is incremented between `:attr:`~Event.EPOCH_START` and
+            :attr:`~Event.EPOCH_END`.
```
jbloxham, line 223: Ditto.

composer/trainer/trainer.py
```diff
@@ -666,16 +666,16 @@ def _train_loop(self) -> None:
...
-
                     state.timer.on_batch_complete(
                         samples=int(num_samples_in_batch.item()),
                         tokens=int(num_tokens_in_batch.item()),
                     )
 
+                    self.engine.run_event(Event.BATCH_END)
+
+                    for scheduler in state.schedulers:
+                        scheduler.step(interval='batch')  # type: ignore
```
jbloxham, line 677: Looks like batch-level schedulers are incremented after the `BATCH_END` event, but epoch-level schedulers are incremented before `EPOCH_END`? Is that an oversight? Not sure why this change is necessary.

jbloxham (approved):

composer/trainer/trainer.py
ravi-mosaicml: Good point; schedulers should be incremented after the timer is incremented. Schedulers should step before the end event so any checkpoint will capture the correct scheduler state.

PR #311, opened 2022-01-31
Don't init dist when world_size is 1

jbloxham:
The distributed process group doesn't necessarily need to be initialized when we're doing single-process training, and this can help avoid crashes in certain environments. Note that the DDP initialization code already is a no-op when the distributed process group is not initialized.

Tested by running models with various world sizes (and no explicit world size), and using a print statement to verify whether the model was the original model class or a DDP wrapper.

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -236,9 +236,9 @@ def initialize_dist(backend: str, timeout: datetime.timedelta):
 
     if dist.is_initialized():
         if dist.get_backend() != backend.lower() and get_world_size() > 1:
-            warnings.warn(f"The requested backend ({backend}) differs from the backend "
-                          f"of the current process group ({dist.get_backend()})."
-                          "If you wish to change backends, please restart the python process.")
+            raise RuntimeError(f"The requested backend ({backend}) differs from the backend "
+                               f"of the current process group ({dist.get_backend()}). If you "
+                               "wish to change backends, please restart the python process.")
         return
 
     if "RANK" not in os.environ or "WORLD_SIZE" not in os.environ:
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -74,7 +74,6 @@ def get_local_rank() -> int:
         int: The local world size
     """
     local_rank = _get_distributed_config_var(env_var="LOCAL_RANK", human_name="local rank", default=0)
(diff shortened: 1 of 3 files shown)


composer/utils/dist.py
```diff
@@ -74,7 +74,6 @@ def get_local_rank() -> int:
         int: The local world size
     """
     local_rank = _get_distributed_config_var(env_var="LOCAL_RANK", human_name="local rank", default=0)
-    assert local_rank == get_global_rank() % get_local_world_size(), "invariant violation"
```
jbloxham, line 77: This change is unrelated to the main purpose of this PR, but I didn't feel like creating a new PR for it. Essentially, we shouldn't be assuming that all nodes have the same amount of processes.

ravi-mosaicml: If we are removing this assumption, we should also add in a `NODE_RANK` env variable, and update `get_node_rank` to use that.

composer/utils/dist.py
```diff
@@ -235,23 +234,18 @@ def initialize_dist(backend: str, timeout: datetime.timedelta):
         return
 
     if dist.is_initialized():
-        if not dist.get_backend() == backend.lower():
+        if dist.get_backend() != backend.lower() and get_world_size() > 1:
             warnings.warn(f"The requested backend ({backend}) differs from the backend "
```
ravi-mosaicml, line 238: Can you convert this back to a RuntimeError? Another PR made it a warning :disappointed:

ravi-mosaicml (approved):
LGTM

composer/utils/dist.py
jbloxham: I'm game.

composer/utils/dist.py
```diff
@@ -235,23 +236,18 @@ def initialize_dist(backend: str, timeout: datetime.timedelta):
         return
 
     if dist.is_initialized():
-        if not dist.get_backend() == backend.lower():
-            warnings.warn(f"The requested backend ({backend}) differs from the backend "
-                          f"of the current process group ({dist.get_backend()})."
-                          "If you wish to change backends, please restart the python process.")
+        if dist.get_backend() != backend.lower() and get_world_size() > 1:
```
ravi-mosaicml, line 239: This should apply whenever dist is initialized

```suggestion
        if dist.get_backend() != backend.lower():
```

PR #312, opened 2022-01-31
Scheduler defaults to step-wise instead of epoch-wise

### LINKED ISSUE #287: Scheduler hparams interval default value should be 'step' instead of 'epoch'
## 🚀 Feature Request
All of the [`SchedulerHparams` classes](https://github.com/mosaicml/composer/blob/main/composer/optim/scheduler.py) have `epoch` as the default value for `interval`. The default value seems like it should be `step`, because stepwise LR updates yield superior performance.

hanlint:
Fixes #287


growlix (approved):
lgtm, pending checks passing after update

PR #316, opened 2022-01-31
Copy `extra_init_params` to get rid of recursive config dicts

siriuslee:
After #236, if you specify the `extra_init_params` dict as part of the `config` for the run, you would get a recursive dictionary by inserting the `config` dict into it directly. Added a copy to prevent that.


ravi-mosaicml (approved):
LGTM

composer/loggers/logger_hparams.py
```diff
@@ -176,6 +175,8 @@ def get_flattened_dict(data: Dict[str, Any], _prefix: List[str] = []) -> Dict[st
                 )
             self.extra_init_params["config"].update(config)
 
+
+
```
hanlint, line 179:
```suggestion
```

PR #317, opened 2022-01-31
Remove explicit YAHP constructs from algorithms

jbloxham:
Hparams initialize objects, so it's weird that those objects sometimes themselves initialize hparams. I know of at least one Mosaic engineer who has been confused by this. Let's just avoid using Hparams where we can.


ravi-mosaicml (approved):
LGTM

hanlint (approved):
LGTM

PR #319, opened 2022-02-01
Composer Style Guide

ravi-mosaicml:
First draft at a style guide


STYLE_GUIDE.md
```diff
@@ -0,0 +1,174 @@
...
+pyright .
+```
+
+Our pyright configuration is stored in [pyproject.toml](pyproject.toml).
+
+While pyright warnings will not cause a build to fail, they should be reviewed as they can identify potential bugs.
+
+### Debugging
+
+Here are some suggestions to deal with pyright errors:
```
jbloxham, line 41: It would be nice to have an example of using `assert` for type checking here. :)

STYLE_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+
+```python
+if parameter is None:
+    raise ValueError("parameter must be specified and cannot be None")
+```
+
+
+## 5. Imports and `__init__.py`
+
+All imports in composer should be absolute -- that is, they do not being with a period.
```
jbloxham, line 160:
```suggestion
All imports in composer should be absolute -- that is, they do not begin with a period.
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+
+If an import is not in `install_requires` (and instead appears in `extra_deps`), then this import
+MUST be conditionally imported in the code -- e.g. like this:
+
+```python
+def unet():
+    try:
+        import monai
+    except ImportError:
+        raise ImportError("monai is not installed. Please run `pip install composer[unet])
```
jbloxham, line 173:
```suggestion
        raise ImportError("monai is not installed. Please run `pip install composer[unet]`")
```

jbloxham (approved):
Looks mostly good - I agree with all the content! Just spotted a couple typos.

PR #320, opened 2022-02-01
Updated add_dataset_transform() to have flexible insertion point

growlix:
`utils.data.add_dataset_transform()` by default appends transformations to `dataset.tranform.transforms`. This is how algorithms such as `randaugment` and `augmix` insert themselves. Some datasets already have `to_tensor()` in the list of transforms (e.g. Cifar10), however `randaugment` and `augmix` need to be added *before* `to_tensor()`. This PR updates `add_dataset_transform()` to have an optional `location` argument that allows insertion before `to_tensor()`.


hanlint:
Nice! Do we care to build a more generalized system? e.g.

```
before=ToTensor(), after=None
```

and allow someone to say.. insert this before `Normalize`?

composer/utils/data.py
```diff
@@ -95,26 +95,38 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
 
-def add_dataset_transform(dataset, transform):
+def add_dataset_transform(dataset, transform, location="end"):
     """Flexibly add a transform to the dataset's collection of transforms.
 
     Args:
         dataset: A torchvision-like dataset
         transform: Function to be added to the dataset's collection of transforms
+        location [str]: Where to insert the transform in the sequence of transforms. "end"
+        will append to the end, "before_totensor" will insert before ToTensor().
```
hanlint, line 105:
```suggestion
        location (str): Where to insert the transform in the sequence of transforms. "end"
        will append to the end, "before_totensor" will insert before ToTensor(). Default: "end".
```

hanlint (commented):
LGTM, but defer to @coryMosaicML for a review.

composer/utils/data.py
```diff
@@ -95,26 +95,38 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
 
     Returns:
         The original dataset. The transform is added in-place.
     """
 
     if not hasattr(dataset, "transform"):
         raise ValueError(f"Dataset of type {type(dataset)} has no attribute 'transform'. Expected TorchVision dataset.")
+    assert location in ["end", "before_totensor"]
```
ravi-mosaicml, line 113: Can you do an if check and raise a valueerror instead of using assert? Asserts should not be used for data validation

ravi-mosaicml:
> Nice! Do we care to build a more generalized system? e.g.
> 
> ```
> before=ToTensor(), after=None
> ```
> 
> and allow someone to say.. insert this before `Normalize`?

Let's leave this for now. But since this parameter would only have two values, could we make it a boolean - i.e. `is_tensor_transformation: bool = True`?

growlix:
Made generalized as per @hanlint's request and idempotent as per @ravi-mosaicml's request

composer/utils/data.py
```diff
@@ -95,26 +96,61 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
         dataset: A torchvision-like dataset
         transform: Function to be added to the dataset's collection of transforms
+        location (str, torchvision transform), optional: Where to insert the transform in
+            the sequence of transforms. "end" will append to the end or "start" will
+            insert at index 0, a transform (e.g. torchvision.transforms.ToTensor) will
+            insert pre- or post- the first instance of `location`, determined by the
+            `pre_post` argument. Default = "end". If `location` is not in list of
+            transforms, will append to end.
+        pre_post (str), optional: Whether to insert before ("pre") or after ("post") the
+            transform string passed in 'location'. Ignored if transform = "start" or "end". Default = "pre".
```
hanlint, line 112:
```suggestion
            transform string passed in 'location'. Ignored if location = "start" or "end". Default = "pre".
```

composer/utils/data.py
```diff
@@ -95,26 +96,61 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
 
     Args:
         dataset: A torchvision-like dataset
         transform: Function to be added to the dataset's collection of transforms
+        location (str, torchvision transform), optional: Where to insert the transform in
+            the sequence of transforms. "end" will append to the end or "start" will
+            insert at index 0, a transform (e.g. torchvision.transforms.ToTensor) will
+            insert pre- or post- the first instance of `location`, determined by the
+            `pre_post` argument. Default = "end". If `location` is not in list of
+            transforms, will append to end.
```
hanlint, line 110: Would it be safer to error out if `location` is not in the list?

composer/utils/data.py
```diff
@@ -95,26 +96,61 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+            textwrap.dedent(
+                f"Invalid value combination for argument `pre_post`: `{pre_post} and `location`: {location}. If location is not one of ['start', 'end'], `pre_post` must be one of ['pre', 'post']."
+            ))
 
     if dataset.transform is None:
         dataset.transform = transform
-    elif hasattr(dataset.transform, "transforms"):  # transform is a Compose
-        dataset.transform.transforms.append(transform)
+    # Check if dataset.transform is of type transforms.Compose and ensure idempotency by not adding transform if it's already present
+    elif isinstance(dataset.transform,
```
hanlint, line 132: This is becoming sufficiently complex that perhaps we want to add some simple testing around this.

hanlint (approved):
LGTM, although we probably want to add a few tests.

composer/utils/data.py
growlix: Not every dataset has the same list of transforms, nor do they all have `ToTensor()` in the list of transforms. An error would require the user to provide `location` as a hyperparameter for the algorithm, which in turn would require them to know which transforms are in their dataset. This seems like a bit much. Perhaps a `log.warning()` displaying the list of transforms before the function returns?

hanlint: i'm ok with a warning, would prefer an error to avoid silent fail (e.g. was supposed to run before Normalize, but since no Normalize, it's now silently running at the end).

composer/utils/data.py
```diff
@@ -93,26 +94,61 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+                    transforms.Compose) and type(transform) not in [type(t) for t in dataset.transform.transforms]:
+        if location == "end":
+            dataset.transform.transforms.append(transform)
+        elif location == "start":
+            dataset.transform.transforms.insert(0, transform)
+        else:
+            insertion_index = len(dataset.transform.transforms)
+            for i, t in enumerate(dataset.transform.transforms):
+                if type(t) == location:
```
ravi-mosaicml, line 139: Curious when would this be true?

composer/utils/data.py
```diff
@@ -93,26 +94,61 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
     return image_tensor, target_tensor
 
 
-def add_dataset_transform(dataset, transform):
+def add_dataset_transform(dataset, transform, location="end", pre_post: str = "pre") -> Dataset:
```
ravi-mosaicml, line 97: type annotation for dataset, transform, location

composer/utils/data.py
ravi-mosaicml: I find this API a bit confusing. Thoughts about replace the `location` and `pre_post` with two lists of transforms?

1. `pre_transformtions: Transformations`: Any transformations that, if present on `dataloader.transform`, will come before this new transformation
2. `post_transformations: Transformations`: Any transformations that, if present on `dataloader.transform`, will come this new transformation

If this is not possible (i.e. a post_transform already appears before a pre_transform), then a ValueError would be raised.

ravi-mosaicml: Alternatively, I think we may be over engineering this. Thoughts about a singular `is_tensor_transform: bool` argument? If it is True, then it should be inserted after ToTensor; if it is False, then it should be inserted before ToTensor. If there is no `ToTensor` present, then it's inserted at the end.

Do our algorithms depend on a specific ordering (other than for the tensors)?

composer/utils/data.py
ravi-mosaicml: Also, can this line be broken into multiple lines?

composer/utils/data.py
hanlint: I'm also OK with `before=ToTensor()` and/or `after=end` but maybe that makes the input validation a bit difficult. @growlix you have the design decision on the API, so make a decision based on these inputs and we'll go with it!

composer/utils/data.py
growlix: Not sure if you mean the `else` (L136) or the `if` (L139). L136 and below execute when `location` is not `"start"` or `"end"` (expected usage is that `location` is a `torchvision.transform`). L139 is true when the item in the list of transforms is of the same type as `location`, e.g. when they're both `transforms.ToTensor`.

growlix:
I thought it'd be nice to have a flexible function that can insert a transform at location relative to other transforms, but you might be right about over-engineering. For now all we need is to insert before or after `ToTensor()` and to insert the transform somewhere else (default to end, or even always to end) if `ToTensor()` is not present.

composer/utils/data.py
```diff
@@ -93,27 +97,86 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+                f"Invalid value combination for argument `pre_post`: `{pre_post} and `location`: {location}. If location is not one of ['start', 'end'], `pre_post` must be one of ['pre', 'post']."
+            ))
+    if not hasattr(dataset, "transform"):
+        raise AttributeError(textwrap.dedent(f"Argument `dataset` must have attribute `transform`."))
+
+    added_transform = 0
+
+    def added_transform_warning():
+        log.warning(
+            f"Transform {transform} added to dataset. Dataset now has the following transforms: {dataset.transform}")
```
growlix, line 143: How do I deal with all the 
```
Cannot access member "transform" for type "Dataset[Unknown]"
Member "transform" is unknown
```
errors from the static analysis?

hanlint: Hmm guess the static typechecker doesn't understand the guard you have upfront.

I'm not a typing expert, but I think you would use `Protocol` to implement a new type that has the attribute `transform`, and use that instead of `torch.utils.Dataset`, e.g. see [PEP 544](https://www.python.org/dev/peps/pep-0544/). 

But maybe defer to @ravi-mosaicml on this one.

composer/utils/data.py
```diff
@@ -93,27 +97,86 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        pre_post (str), optional: Whether to insert before ("pre") or after ("post") the
+            transform string passed in 'location'. Ignored if transform = "start" or "end". Default = "pre".
 
     Returns:
         The original dataset. The transform is added in-place.
     """
 
     if not hasattr(dataset, "transform"):
-        raise ValueError(f"Dataset of type {type(dataset)} has no attribute 'transform'. Expected TorchVision dataset.")
+        raise AttributeError(textwrap.dedent(f"""Dataset must have attribute `transform`."""))
```
hanlint, line 127: do we need the `textwrap.dedent` here and below when the errors are relatively short?

ravi-mosaicml: Nope, textwrap.dedent is only for multiline errors.

composer/utils/data.py
ravi-mosaicml: I think I missed that `t` could be a string or transform.

composer/utils/data.py
ravi-mosaicml: Can you add an `assert isinstance(dataset, VisionDataset)`?

composer/utils/data.py
ravi-mosaicml: Can we use check to ensure the dataset is a subclass of `torchvision.datasets.VisionDataset` (https://pytorch.org/vision/main/generated/torchvision.datasets.VisionDataset.html)?

e.g.

```python3
if not isinstance(dataset, torchvision.datasets.VisionDataset):
    raise TyepError(...)
```

composer/utils/data.py
```diff
@@ -93,27 +97,86 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        return 1
 
     if dataset.transform is None:
         dataset.transform = transform
-    elif hasattr(dataset.transform, "transforms"):  # transform is a Compose
-        dataset.transform.transforms.append(transform)
-    else:  # transform is some other basic transform, join using Compose
-        dataset.transform = transforms.Compose([dataset.transform, transform])
-
+        added_transform = added_transform_warning()
```
ravi-mosaicml, line 148: For readabiity, can you remove the helper function and split it into two lines? Makes it more readable what's happening.

```suggestion
        log.warn(LOG_MESSAGE)
        added_transform = True
```
Same thing in the other places...

composer/utils/data.py
hanlint: Since we want to support custom datasets that have the `tranforms` attribute but not be properly inheriting from `VisionDataset`, we may not want to prematurely narrow the checks here.

ravi-mosaicml: Any custom dataset can inherit from `VisionDataset`. This function was designed around the structure of `VisionDataset`, so I think it should be limited to datasets of that type.

growlix: I don't think so, because Composer's synthetic datasets are of type `torch.utils.data.Dataset`, which would make them incompatible with this. But I think it could be reasonable to ensure that the dataset is a subclass of `torch.utils.data.Dataset`, which iirc is the most generic `dataset` class.

composer/utils/data.py
```diff
@@ -93,27 +97,86 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+            ))
+    if not hasattr(dataset, "transform"):
+        raise AttributeError(textwrap.dedent(f"Argument `dataset` must have attribute `transform`."))
+
+    added_transform = 0
+
+    def added_transform_warning():
+        log.warning(
+            f"Transform {transform} added to dataset. Dataset now has the following transforms: {dataset.transform}")
+        return 1
```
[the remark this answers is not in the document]
growlix: To set a flag for whether the transform was actually added, but probably best not to combine those functions.

composer/utils/data.py
ravi-mosaicml: Correct, our synthetic datasets do not support torchvision transforms, since the yield tensors directly. See https://github.com/mosaicml/composer/issues/346

growlix:
I simplified `add_dataset_transform()` such that the only placement-related parameter is `is_tensor_transform`, and the new transform is placed before/after `ToTensor()` if `ToTensor()` is present in `dataset.transform`, and at the end if not. I also added a `SyntheticPILDataset` class that's nearly identical to the `SyntheticBatchPairDataset`, except that it generates/returns PIL images instead of tensors.

Also tests.

composer/utils/data.py
```diff
@@ -93,27 +97,86 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+                    insertion_index = i
+                    break
+            if pre_post == "post":
+                insertion_index += 1
+            dataset.transform.transforms.insert(insertion_index, transform)
+        added_transform = added_transform_warning()
+    elif not isinstance(dataset.transform, transforms.Compose):
+        # Transform is some other basic transform, join using Compose
+        # Ensure idempotency by not adding transform if it's already present
+        if not type(dataset.transform) == type(transform):
```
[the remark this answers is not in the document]
growlix: No worries!

composer/core/types.py
```diff
@@ -164,3 +164,8 @@ class MemoryFormat(StringEnum):
     CHANNELS_LAST = "channels_last"
     CHANNELS_LAST_3D = "channels_last_3d"
     PRESERVE_FORMAT = "preserve_format"
+
+
+class DatasetWithTransform(Protocol):
```
ravi-mosaicml, line 169: Curious is this class being used? If not let's remove it?

composer/datasets/synthetic.py
```diff
@@ -125,3 +128,104 @@ def __getitem__(self, idx: int):
...
+                 *,
+                 total_dataset_size: int,
+                 data_shape: Sequence[int] = [64, 64],
+                 num_unique_samples_to_create: int = 100,
+                 data_type: Union[str, SyntheticDataType] = SyntheticDataType.GAUSSIAN,
+                 label_type: Union[str, SyntheticDataLabelType] = SyntheticDataLabelType.CLASSIFICATION_INT,
+                 num_classes: Optional[int] = None,
+                 label_shape: Optional[Sequence[int]] = None,
+                 transform: Optional[Callable] = None):
+        self.total_dataset_size = total_dataset_size
```
ravi-mosaicml, line 158: Can you add a call to `super.__init__(root="", transform=transform)`

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+                          transform: Callable,
+                          is_tensor_transform: bool = False) -> torch.utils.data.Dataset:
+    """Add a transform to a dataset's collection of transforms. Inserts the transform
+    before or after torchvision.transforms.ToTensor(), or at the end of the collection of
+    ToTensor() is not present.
 
     Args:
-        dataset: A torchvision-like dataset
-        transform: Function to be added to the dataset's collection of transforms
+        dataset (torch.utils.data.Dataset): A torchvision-like dataset
```
ravi-mosaicml, line 109: Docstring 
```suggestion
        dataset (VisionDataset): A torchvision-like dataset
```

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        transform (Callable): Function to be added to the dataset's collection of
+            transforms
+        is_tensor_transform (bool): Whether `transform` acts on data of the type
+        torch.Tensor. If `True`, and torchvision.transforms.ToTensor() is present in
```
ravi-mosaicml, line 113: For sphinx to parse it properly, lines 113-117 need to be indented.

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+    before or after torchvision.transforms.ToTensor(), or at the end of the collection of
+    ToTensor() is not present.
 
     Args:
-        dataset: A torchvision-like dataset
-        transform: Function to be added to the dataset's collection of transforms
+        dataset (torch.utils.data.Dataset): A torchvision-like dataset
+        transform (Callable): Function to be added to the dataset's collection of
+            transforms
+        is_tensor_transform (bool): Whether `transform` acts on data of the type
```
ravi-mosaicml, line 112: RST uses double ticks for code literal blocks (I made this mistake wayy to many times, then the docs come out in italics :D) Not going to mark all, but please double check.

```suggestion
        is_tensor_transform (bool): Whether ``transform`` acts on data of the type
```

composer/utils/data.py
ravi-mosaicml, line 113: For auto-highlighting of methods, use these sphinx tags like `:meth:`

The leading `~` says to only show the last part (i.e. `ToTensor`) in the docs rather than the whole name (i.e. `torchvision.transforms.ToTensor`)

```suggestion
        torch.Tensor. If ``True``, and :meth:`~torchvision.transforms.ToTensor` is present in
```

Not going to comment on all, but let's enclose these in double-ticks so they will be formatted as code. Ideally we should reference it using the `:meth:` notation but can address that in a separate PR

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        `dataset`'s transforms, will insert the transform after ToTensor(). If `False` and
+        ToTensor() is present, will insert the transform before ToTensor(). If ToTensor()
+        is not present, the transform will be appended to the end of collection of
+        transforms. Default = `False`.
 
     Returns:
         The original dataset. The transform is added in-place.
     """
 
     if not hasattr(dataset, "transform"):
```
ravi-mosaicml, line 123: Since the type annotation is a `VisionDataset`, this hasattr check is no longer needed.

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        insertion_index = len(dataset.transform.transforms)
+        for i, t in enumerate(dataset.transform.transforms):
+            if isinstance(t, transforms.ToTensor):
+                insertion_index = i
+                break
+        if is_tensor_transform:
+            insertion_index += 1
+        dataset.transform.transforms.insert(insertion_index, transform)
+        log.warning(transform_added_logstring)
+    elif not isinstance(dataset.transform, transforms.Compose):
```
ravi-mosaicml, line 143: Let's keep the comment, it's helpful :) 

Also, should this be an else?

```suggestion
    else: # transform is some other basic transform, join using Compose
```

composer/utils/data.py
```diff
@@ -93,26 +98,60 @@ def pil_image_collate(batch: List[Tuple[Image.Image, Union[Image.Image, Tensor]]
...
+        if isinstance(dataset.transform, transforms.ToTensor) and not is_tensor_transform:
+            dataset.transform = transforms.Compose([transform, dataset.transform])
+            log.warning(transform_added_logstring)
+        else:
+            dataset.transform = transforms.Compose([dataset.transform, transform])
+            log.warning(transform_added_logstring)
+    else:
```
ravi-mosaicml, line 150: I don't think this else would ever trigger?

docs/source/callbacks.rst
```diff
@@ -9,7 +9,7 @@ Composer contains built-in callbacks, which can be added via the ``--callbacks``
 
 .. code-block::
 
-    python examples/run_mosaic_trainer.py -f my_model.yaml --callbacks lr_monitor grad_monitor
+    python examples/run_composer_trainer.py -f my_model.yaml --callbacks lr_monitor grad_monitor
```
ravi-mosaicml, line 12: Whoops, thanks!

composer/datasets/synthetic.py
```diff
@@ -125,3 +128,104 @@ def __getitem__(self, idx: int):
             return self.transform(self.input_data[idx]), self.input_target[idx]
         else:
             return self.input_data[idx], self.input_target[idx]
+
+
+class SyntheticPILDataset(VisionDataset):
```
ravi-mosaicml, line 133: Thanks for writing this! This pretty implements #346 :)

Jotting this down for another issue/PR, unless if you have the bandwidth to implement it tmrw :):

A lot of this logic seems very similar to the `SyntheticBatchPairDataset`.

Could the `SyntheticPILDataset` internally use a `SyntheticBatchPairDataset` to generate the tensors, and then wrap it in a PIL image (and apply the transformations) on `__getitem__`.? I think that would clean up the code.

ravi-mosaicml (approved):
Overall LGTM! See comments.

Can you try a docs build and make sure the docstrings format as expected? I tried to comment on some of the rst syntax errors but probably missed some. I don't want this to hold up the PR, though, since we're going to do a heavy docs scrub next week.

composer/datasets/synthetic.py
growlix: Oh wow, nice to have inadvertently resolved an outstanding issue :)
I'm not quite sure how to internally use a `SyntheticBatchPairDataset` to generate the data for `SyntheticPILDataset` because `SyntheticBatchPairDataset` generates the tensors upon the first call to `__getitem__`, and `SyntheticPILDataset` would need to override `__getitem__`. It might be possible to change `SyntheticBatchPairDataset` to generate the data on init, but the comment for `__getitem__` in `SyntheticBatchPairDataset` makes it seem as though the data need to be generated upon the first call to `__getitem__` so that they're on the correct gpu.

PR #322, opened 2022-02-01
Configure DeepSpeed with an ordinary DeepSpeed config dict

jbloxham:
Completely removes `DeepSpeedHparams`, in favor of instead accepting an ordinary DeepSpeed config dict, formatted exactly as their docs allow.

Down the road, it would be nice for _any_ valid DeepSpeed config dict to work with our trainer (especially the `"optimizer"` field), but that's a refactor that can wait for a future Composer version.

Tested by trying out some configurations locally. Admittedly, it became clear that YAHP doesn't have great support for parsing JSON-type fields, but that's a problem that can be tackled separately.


composer/trainer/deepspeed.py
```diff
@@ -1,98 +1,121 @@
...
-                "enabled": True,
-                "initial_scale_power": 16,
-                "loss_scale_window": 2000,
-            }
-
-        if grad_clip_norm:
-            deepspeed_config["gradient_clipping"] = grad_clip_norm
-
-        return deepspeed_config
+def parse_batch_settings(config: Dict[str, Any], state: State):
```
ravi-mosaicml, line 15:
```suggestion
def _add_batch_sizing_to_config(config: Dict[str, Any], state: State):
```

composer/trainer/deepspeed.py
```diff
@@ -1,98 +1,121 @@
...
+    if "optimizer" in config:
+        raise ValueError("The DeepSpeed configuration specifies an optimizer, but the Mosaic "
+                         "trainer will override this setting.")
+
+    if "scheduler" in config:
+        raise ValueError("The DeepSpeed configuration specifies a scheduler, but the Mosaic "
+                         "trainer will override this setting.")
+
+
+def parse_precision_settings(config: Dict[str, Any], state: State):
```
ravi-mosaicml, line 64:
```suggestion
def _add_precision_to_config(config: Dict[str, Any], state: State):
```

composer/trainer/deepspeed.py
```diff
@@ -1,98 +1,121 @@
...
+                      "This causes DeepSpeed to reject certain Mosaic optimizers that are known to "
+                      "work well with DeepSpeed.")
+
+    config["zero_allow_untested_optimizer"] = True
+
+
+def parse_deepspeed_config(config: Dict[str, Any],
+                           state: State,
+                           grad_clip_norm: Optional[float] = None) -> Dict[str, Any]:
+    new_config = copy.deepcopy(config)
```
ravi-mosaicml, line 114: 1. Docstring
2. Would it make sense to modify the config in place rather than deep-copying?

composer/trainer/deepspeed.py
```diff
@@ -1,98 +1,121 @@
...
+            config["fp16"] = cast({"enabled": True}, Dict[str, Any])
+        fp16_config = config["fp16"]
+        assert isinstance(fp16_config, dict)
+
+        # For equivalence with the non-DeepSpeed defaults of the Mosaic trainer.
+        fp16_config.setdefault("initial_scale_power", 16)
+        fp16_config.setdefault("loss_scale_window", 2000)
+
+
+def parse_misc_settings(config: Dict[str, Any], grad_clip_norm: Optional[float]):
```
ravi-mosaicml, line 92:
```suggestion
def _add_other_settings_to_config(config: Dict[str, Any], grad_clip_norm: Optional[float]):
```

composer/trainer/trainer.py
```diff
@@ -176,7 +176,7 @@ def __init__(
             eval_subset_num_batches: Optional[int] = None,
 
             # DeepSpeed
-            deepspeed_hparams: Optional[Union[dict, DeepSpeedHparams]] = None,
+            deepspeed_config: Optional[dict] = None,
```
ravi-mosaicml, line 179:
```suggestion
            deepspeed_config: Optional[Dict[str, Any]] = None,
```

composer/trainer/trainer_hparams.py
```diff
@@ -211,14 +210,11 @@ def validate(self):
 
         if self.deepspeed is not None:
 
-            if self.precision == Precision.FP16:
-                raise ValueError("FP16 precision is only supported when training with DeepSpeed.")
-
             if isinstance(self.device, CPUDeviceHparams):
                 raise ValueError("Training on CPUs is not supported with DeepSpeed.")
 
-            if self.deterministic_mode and self.deepspeed.zero_stage > 0:
```
ravi-mosaicml, line 220: Can you keep this error?

composer/trainer/trainer.py
ravi-mosaicml: Also docstring updates

ravi-mosaicml (commented):
LGTM, though docstrings!

composer/trainer/deepspeed.py
```diff
@@ -1,98 +1,121 @@
...
+            raise ValueError("Provided DeepSpeed configuration specifies per-GPU microbatch size="
+                             f"{ds_per_gpu_microbatch_size}, but the Mosaic trainer has been "
+                             f"configured with per-GPU microbatch size={per_gpu_microbatch_size}.")
+
+    config["train_batch_size"] = train_batch_size
+    config["gradient_accumulation_steps"] = grad_accum
+    config["train_micro_batch_size_per_gpu"] = per_gpu_microbatch_size
+
+
+def parse_unsupported_settings(config: Dict[str, Any]):
```
ravi-mosaicml, line 54:
```suggestion
def _ensure_no_optimizer_scheduler_in_config(config: Dict[str, Any]):
```

ravi-mosaicml (approved):

composer/trainer/deepspeed.py
jbloxham: The config is potentially user provided, so I'd rather not modify it if possible.

I guess we violate this principle for other things used to initialize the trainer, like the model and optimizers, but any reasonable person would _expect_ those to be modified. I'm less sure about something like a DeepSpeed config.

PR #325, opened 2022-02-01
Fully flatten checkpoint params

### LINKED ISSUE #318: Simplify `ObjectStoreHparams` and `CheckpointLoader`
For ease of use, we should:
* `ObjectStoreHparams` and `ObjectStoreProvider` should have the same arguments
* `CheckpointLoader`'s constructor should not require an `hparams` object, but rather possibly a `dict` of the configuration

ravi-mosaicml:
1. Remove the CheckpointLoaderHparams and CheckpointSaverHparams; merged into TrainerHparams
2. CheckpointLoader and trainer take an `ObjectStoreProvider` rather than `ObjectStoreProviderHparams`
3. Simplifed the `saver_interval` and `save_interval_unit` to use a time string.

Closes #318


composer/yamls/models/glue/mnli-m.yaml
```diff
@@ -61,12 +61,7 @@ validate_every_n_batches: 2300
 validate_every_n_epochs: 1
 callbacks:
   - lr_monitor: {}
-save_checkpoint:
```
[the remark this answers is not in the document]
moinnadeem: thanks for the heads up!

composer/trainer/checkpoint.py
```diff
@@ -86,14 +88,14 @@ def __init__(
 
         checkpoint_uri_parsed = urllib.parse.urlparse(path)
         if checkpoint_uri_parsed.scheme != "":
-            if object_store_hparams is not None:
+            if object_store is not None:
                 raise ValueError(
                     textwrap.dedent("""\
                         When specifying `object_store_hparams`,
```
jbloxham, line 94:
```suggestion
                        When specifying `object_store`,
```

composer/trainer/checkpoint_hparams.py
```diff
@@ -1,87 +0,0 @@
...
-    progress_bar: bool = hp.optional(doc=textwrap.dedent("""\
-        Whether to show a progress bar when downloading checkpoints.
-        Ignored if the checkpoint is a local file path"""),
-                                     default=True)
-
-    def validate(self):
-        if not self.load_weights_only and self.strict_model_weights:
-            raise ValueError(
-                "Strict cannot be used when load_weights_only is true. Restoring a checkpoint from previous state assumes that the checkpoint should perfectly match the model."
-            )
```
jbloxham, line 51: Can you move this check either to `CheckpointLoader.__init__` or `TrainerHparams.validate`?

composer/trainer/trainer_hparams.py
```diff
@@ -177,8 +177,34 @@ class TrainerHparams(hp.Hparams):
...
+        Ignored if the checkpoint is a local file path"""),
+        default=True)
+
+    save_folder: Optional[str] = hp.optional(doc=textwrap.dedent(f"""If set, folder to save checkpoint files.
+        Relative to the run directory. Defaults to None, meaning checkpoints will not be saved."""),
+                                             default=None)
+    save_interval: str = hp.optional(doc=textwrap.dedent("""The time string interval representing how
+        frequently checkpoints should be saved. For example, set to "1ep" to save checkpoints every epoch, or "10ba"
+        to save checkpoints every 10 batches.. """),
+                                     default="1ep")
```
jbloxham, line 207: The thing I don't like about this is as a user, it's now unclear what the difference in configuration is between a config that loads a checkpoint and a config that does not load a checkpoint. "Do you have to specify all of the `load_*` hparams if you want to turn on checkpoint loading?" is a question I imagine many users would have.

Since we're committed to flattening, could we add a `load_checkpoint: Optional[bool]` field? If `load_checkpoint` is true, we attempt load a checkpoint, otherwise we don't. In `.validate()` for the `TrainerHparams`, we would then also validate that `load_path` is required if `load_checkpoint` is true, and that all of the `load_*` must be null if `load_checkpoint` is false.

And likewise for `save_checkpoint`.

hanlint: Would it be natural for users to think that save_folder and load_path serve these purposes?

jbloxham: @ravi-mosaicml and I just discussed offline and we think, since we're committed to flattening configs, we're indeed best off substituting static typing for copious documentation. Probably not going to add unnecessary fields to the hparams.

composer/trainer/checkpoint_hparams.py
ravi-mosaicml: Chucking this check, as discussed offline. Don't understand what it's doing...

composer/trainer/trainer_hparams.py
ravi-mosaicml: Added comments to the docstrings to make it clear that secondary parameters have no effect unless if the paths are set.

hanlint (approved):

PR #326, opened 2022-02-01
Add functional/standalone API tutorial notebook

dblalock:
Attempting to check in an exported version of [this notebook](https://colab.research.google.com/drive/1HIxLs61pyf0ln7MlnrGYvkNHq1uVbNWu#scrollTo=DSzcOUUDOkow). We'll see how it looks in the docs...

Incorporated feedback from @siriuslee, @kobindra, @hanlint, and now have:

1. Code for a decent number of methods included, but
2. A simpler subset actually used.

This is achieved by having functions with the option to add a few different methods, but only one selected. Which I *think* reconciles most of our feedback. That said, I don't trust my taste in tutorial notebooks, so still definitely open to feedback.

If/when we clean up the functional API, a few lines in this notebook may need to change. We'll probably also need an editing pass through all the tutorial notebooks just to make them stylistically consistent (e.g., do we put the install command before the Heading1 like huggingface, or after it?)


dblalock:
^ jenkins failure appears to just be a timeout ([pipeline](https://ci.int.mosaicml.com/job/internal/job/mosaicml/job/composer/job/PR-326/23/flowGraphTable/))

notebooks/Composer_Functional.ipynb
```diff
@@ -0,0 +1,306 @@
...
+    "train_and_eval(model, trainloader, testloader)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Looks like adding squeeze-excite gives us another few percentage points of accuracy, and does so with little decrease in the iterations per second. Great!\n",
+    "\n",
+    "Of course, this is a toy model and an unrealistically short training duration, so almost nothing should be read into these results. But hopefully this served as a useful illustration of how to use Composer's algorithms without the Composer trainer, and with minimal change to your code. If you hit any problems or have questions, feel free to [open an issue](https://github.com/mosaicml/composer/issues/new/) or reach out to us [on Slack](https://join.slack.com/t/mosaicml-community/shared_invite/zt-w0tiddn9-WGTlRpfjcO9J5jyrMub1dg)."
```
hanlint, line 284: suggest removing this qualifer, i think its understood!

also can you link to the method cards overview? each one will have code to show how to use it the functional form.

notebooks/Composer_Functional.ipynb
```diff
@@ -0,0 +1,306 @@
...
+    "import torch.nn.functional as F\n",
+    "import torchvision\n",
+    "import torchvision.transforms as transforms\n",
+    "\n",
+    "datadir = './data'\n",
+    "batch_size = 1024\n",
+    "torch.manual_seed(42)\n",
+    "\n",
+    "\n",
+    "def make_dataloaders():\n",
```
hanlint, line 43: should we un-function-ify this (ironic because its the functional notebook 😂 ) so users can poke around in the variables interactively?

notebooks/Composer_Functional.ipynb
```diff
@@ -0,0 +1,306 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from typing import Sequence\n",
+    "import composer.functional as cf\n",
+    "import functools\n",
+    "import sys\n",
+    "import composer\n",
+    "\n",
+    "def make_augmented_dataloaders(which_transforms: Sequence[str]):\n",
```
hanlint, line 178: I dont think we need `which_transforms`, and can just hardcode it since this is a notebook, and not meant for production configuratable code.

hanlint: also, since this is only used once.. un-function-ify?

notebooks/Composer_Functional.ipynb
```diff
@@ -0,0 +1,306 @@
...
+    "Now let's try using some algorithms that modify the model. We're going to keep things simple and just add a [Squeeze-and-Excitation](https://docs.mosaicml.com/en/latest/method_cards/squeeze_excite.html) module after the larger of the two conv2d operations in our model. We show examples of how to use other algorithms as well as a reference."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def modify_model(model: torch.nn.Module, which_algorithms: Sequence[str]):\n",
```
hanlint, line 248: same thing.. maybe better to un-functionify things so its more interactive.

hanlint (commented):
Looking good -- one major suggestion to break out the functions, if you agree.

notebooks/Composer_Functional.ipynb
dblalock: removed the qualifier.

Tried to link to the methods overview (`method_cards/methods_overview.html`) but it isn't live on the website, and we don't support `/en/latest/`, so I couldn't find a way to do it that didn't hardcode the library version. Happy to add post-release, or if you know a reliable url to use.

notebooks/Composer_Functional.ipynb
dblalock: done

notebooks/Composer_Functional.ipynb
dblalock: done

notebooks/Composer_Functional.ipynb
dblalock: done

hanlint (approved):

Issue #40, opened 2021-10-26
`eval_only` flag


hanlint:
## 🚀 Feature Request

For post-hoc measurements on different datasets, we want to be able to load a checkpoint and run `--eval_only`. 

## [Optional] Implementation

Add a `--eval_only` flag that loads from checkpoint and only runs eval. User would need to specify a new dataset/dataloader that differs from the checkpointed hparams.

ravi-mosaicml:
Depends on #120

hanlint:
adding @ajaysaini725 here to help with design ideas

ajaysaini725:
Gave this some thought - in order to finalize a design for this we need to decide exactly how to handle placement of `train_dataloader` and `Evaluators` across `Trainer.__init__`, `Trainer.fit()`, and `Trainer.eval()`.

This is very connected to the discussion and design that happened in #161 . 

The high-level decision looks like the following:

- Do we plan on keeping `train_dataloader` and `eval_dataloader` as required arguments to the `Trainer.__init__` or are we okay moving them to `Trainer.fit()`?
    - Reason being that it doesn’t make sense to force the user to pass in a train_dataloader in order to instantiate a Trainer to load the checkpoint when running eval_only mode
    - If we decide to keep the `Trainer.__init__` as-is, then it would make the most sense to have the eval_only mode be either:
        - A separate class that takes in the below input and provides a predict() function
        - A staticmethod of the Trainer class that takes in the below input
    - If we decide to move the dataloader arguments out of `Trainer.__init__` and into `Trainer.fit()`, then we can have the `Trainer.__init__` take an eval_only flag, have `eval()` take Evaluators as an argument, and have a `predict()` function within the Trainer class that takes a `test_dataloader` as an argument
    
In order to support interactive development in addition to having a clean `eval_only` interface, it seems like it makes the most sense to move the `train_dataloader` and`Evaluators` out of `__init__` (or at least make them optional) and make them parameters to `fit()` and `eval()`. Once we do this, we can easily add a `predict()` function that is cohesive with the rest of the `Trainer` design as well.

Thoughts?

ravi-mosaicml:
One of the main issues about moving arguments out of init is that create_from_hparams would need to store the arguments that would be passed into fit somewhere. We could also refactoring the yaml and have something else call trainer.init and trainer.fit for jobs created from hparams. I think if we went this route we'd need to resolve the multiple calls to fit issue.

[...]

PR #328, opened 2022-02-01
Ensure INIT is Idempotent

ravi-mosaicml:
As we implement #138 and #40, the init event be called multiple times, potentially in the same training run. As such, it will need to be idempotent wherever it is used.

List of places where `Event.INIT` is used:

Algorithms:
- [x] alibi
- [x] augmix
- [x] blurpool
- [x] channels last
- [x] colout
- [x] cutmix
- [x] factorize
- [x] ghost_batchnorm
- [x] mixup
- [x] no_op_model
- [x] randaugment
- [x] sam
- [x] scale_schedule
- [x] selective_backprop
- [x] seq_length_warmup
- [x] squeeze_excite
- [x] stochastic_depth

Callbacks:
- [x] benchmarker
- [x] run directory uploader
- [x] file_logger
- [x] tqdm_logger
- [x] wandb_logger
- [x] dataloader_profiler (in the profiler)
- [x] json_trace (in the profiler)
- [x] system_profiler (in the profiler)
- [x] torch_profiler (in the profiler)

Most of the changes to each of the algorithms / callbacks are quite small, but there are a lot.

This PR also added in a test that call init multiple times and ensures that the model can still train. Because we previously didn't have tests that attempted to train with certain algorithms, the addition of this test identified other issues with our dummy model, which unfortunately made this PR significantly larger than it should have been.


ravi-mosaicml:
Discussed with Davis; going to separate out Event.INIT / Event.FIT so INIT is called once upon trainer.__init__ (And should be used only for model transformations); will move dataloader transformations to Event.FIT and keep those idempotent.

PR #333, opened 2022-02-01
Fail fast if any step fails

ravi-mosaicml:
If any build fails, immediately kill all other builds with the failFast setting. This should help avoid running useless GPU tests.


hanlint (approved):
Thanks!

PR #335, opened 2022-02-01
Replace most instances of "Mosaic" with "Composer"

jbloxham:
For sake of more consistent branding:

`BaseMosaicModel` -> `ComposerModel`
`MosaicTranformer` -> `ComposerTransformer`
`MosaicClassifier` -> `ComposerClassifier`
"Mosaic(ML) trainer" -> "Composer trainer"

Notably, this also renames `run_mosaic_trainer` to `run_composer_trainer`. This will break everyone's workflows, so it's a good thing that everyone is just writing documentation right now anyways.

One of the only references that persists is the "MosaicML logger". To be honest, I have no idea what the hell this thing is, but I'm assuming it interacts with other Mosaic services, so preserving the "Mosaic" name may be justified.


ravi-mosaicml (approved):
:bomb:

PR #336, opened 2022-02-01
Add compression flag

mvpatel2000:
Adds an option to support compression for checkpoints.

Test plan:
- Updated test to check compression

Predictably, compression is slow with the heavier methods (eg lzma). If this becomes important, we could hide the zipping latency by doing it in parallel with training since it doesn't actually use much CPU (just a lot of serial ops).


jbloxham (commented):
Im afraid you've got a meaty merge conflict to deal with - the checkpoint hparams have all been flattened out directly into trainer hparams.

mvpatel2000:
Hm... getting something weird where now it's not ever checkpointing.... might be calling it wrong? looking into it and then will re-request review

mvpatel2000:
Ah nvm, it seems logging is just not on by default anymore? In either case, it saves w compression :)

composer/trainer/checkpoint_hparams.py
```diff
@@ -75,13 +75,20 @@ class CheckpointSaverHparams(hp.Hparams):
     folder: str = hp.optional(doc="Folder in which to save checkpoint files. Relative to the run directory, if set."
                               "Defaults to `checkpoints`.",
                               default="checkpoints")
+    compression: str = hp.optional(doc="Compression algorithm to run on checkpoints. Must be one of `None`, `gzip`, `bzip2`, or `lzma`.",
```
ravi-mosaicml, line 78: Let's use the empty string to represent no compression -- it's a bit weird to have strings of `None`.

```suggestion
    compression: str = hp.optional(doc="Compression algorithm to run on checkpoints. Must be blank or one of `gzip`, `bzip2`, or `lzma`.",
```

composer/trainer/trainer.py
```diff
@@ -110,6 +110,8 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
         save_folder (Optional[str]): Folder path to save checkpoints
         save_interval (int): How often to save checkpoints
         save_interval_unit (str): Unit of ``save_interval``. Can be ``ep`` or ``steps``. (default: ``ep``).
+        save_compression (str): Compression algorithm to run on checkpoints. Can be `None`, `gzip`, `bzip2`,
```
ravi-mosaicml, line 113:
```suggestion
        save_compression (str): Compression algorithm to run on checkpoints. Can be `gzip`, `bzip2`,
```

composer/trainer/trainer.py
```diff
@@ -174,6 +176,7 @@ def __init__(
             save_folder: Optional[str] = None,
             save_interval: int = 1,
             save_interval_unit: str = 'ep',
+            save_compression: str = 'None',
```
ravi-mosaicml, line 179:
```suggestion
            save_compression: str = '',
```

ravi-mosaicml (commented):
Great first PR!!! :rocket: 

One minor comment to use the empty string rather than "None" to denote no compression, since we don't use the string of "None" elsewhere in the code. Otherwise, looks good!

mvpatel2000:
> One minor comment to use the empty string rather than "None" to denote no compression, since we don't use the string of "None" elsewhere in the code. 

Updated!

>Would also be a good idea to add a simple test case in just to verify that the loading/saving works. I don't think we need to test each compression as that's handled by python's tarfile.

Modified tests to also run `gzip`

composer/trainer/checkpoint.py
```diff
@@ -335,7 +335,7 @@ class CheckpointSaver:
         interval (Time or str): The amount of time units to wait between checkpoints.
     """
 
-    def __init__(self, save_folder: str, interval: Union[Time, str]):
+    def __init__(self, save_folder: str, interval: Union[Time, str], compression: str):
```
ravi-mosaicml, line 338:
```suggestion
    def __init__(self, save_folder: str, interval: Union[Time, str], compression: str = ""):
```

Also docstring update

ravi-mosaicml (approved):
LGTM, just see the one comment. Feel free to merge afterwards!

composer/trainer/checkpoint.py
hanlint: Going to commit the suggestion (pretty innocent), and then merge with red button (no need to rerun all of jenkins for this one change)

Issue #324, opened 2022-02-01
Unify time arguments for SchedulerHparams and Scheduler objects


growlix:
## 🚀 Feature Request
Currently, duration (e.g. `warmup_iters` or `T_max`) and resolution (steps or epochs) are fully dissociated for SchedulerHparams, and duration is specified as a string, and the magical `composer.Time` handles the time conversions under the hood. PR #249 moved us away from using `*Hparam` objects and towards using Scheduler/Optimizer objects, but unfortunately the Scheduler objects don't use `Time` under the hood for duration, nor do they dissociate update resolution from duration. It would be great if they did.

## Motivation

1. Composer's `Time` handling is magical
2. Duration and resolution of LR schedules should be dissociated
3. We should try to keep the interfaces for `*Hparams` their associated objects consistent.

## [Optional] Implementation

Looks like a lot of Schedulers are pulling from pytorch directly, so I'm not sure what the cleanest way to do this is, but if someone wants to point me in the right direction I'm happy to do it.

Thoughts @hanlint, @ravi-mosaicml, @jbloxham

jbloxham:
Can you elaborate on what the dissociation of duration and resolution should look like?

growlix:
For example, you can instantiate a `WarmUpLRHparams` object like:
```
warmup = composer.optim.WarmUpLRHparams(
    warmup_iters="2ep", # Warmup duration
    warmup_method="linear", # Linear warmup
    warmup_factor=1e-4, # Initial LR = LR * warmup_factor
    interval="step" # Update LR with stepwise resolution for superior results
)

hanlint:
Based on discussions in #259 , I think we want to redo how we treat schedulers entirely. This would involve:
* Removing `ComposerScheduler`
* Implementing some common schedulers as `LambdaLR` stateless schedulers (e.g. `warmup_with_cosine_decay`)

Issue #294, opened 2022-01-28
In Memory Logger


growlix:
## 🚀 Feature Request
There currently does not seem to be a way to directly access logged values inside the execution environment. One has to write to a file then read the file. It would be great to be able to log to a dict.

## Motivation
When writing a notebook explaining how to train with algorithms, I wanted to be able to collect and plot training curves to compare baseline vs. algorithm-enhanced training. More generally, I think a lot of exploratory workflows do similar things and people will want easy access to logged values.

PR #340, opened 2022-02-02
Up and Running with Composer and Speedup Algorithms Demo Notebook

growlix:
This walks through training a ResNet56 on CIFAR10 using Composer's trainer, then speeding up training with speed-up algorithms. There are a few things that make this notebook less than ideal:

1. Limited and variable Colab instances means it takes anywhere from 1-3 minutes for a single training epoch, depending on the instance. I try to avoid this by only training for three epochs, but this still takes 3-10 minutes.
2. Instance variability and variability during early training both contribute to highly variable results: sometimes we get no speedup, and sometimes we get a 2.5x speedup, which makes it hard to demonstrate the utility of speedup algorithms.
3. There's no easy way of plotting results (#294)
4. It's not possible to use Composer's lovely time handling to specify scheduler duration and interval (i.e. have to manually convert from epochs to steps; #324)
5. Related to the above: No way to demo cosine decay, because it's not possible to instantiate the Scheduler object with `step` interval (in contrast to the Hparams object).

I think we're in a bit of an awkward point in Composer's development for `Hparams`-free usability, and this combines with the highly variable and resource-constrained Colab instances to make it challenging for this notebook to demonstrate some of Composer's big strengths.


hanlint:
Yeah agreed, if really need to clean up our `schedulers` objects. Not sure if you have bandwidth to help there @growlix but happy to huddle tomorrow if you do.

moinnadeem:
@growlix Does this notebook exist as a Colab link rather than on GitHub?

Issue #299, opened 2022-01-28
`ghost_batchnorm` with a single normalization op


ajaysaini725:
## 🚀 Feature Request

An implementation that made only one call to `F.batch_norm` or `F.layer_norm` would be more performant than the one we have now. Before implementing the change, some investigation / benchmarking should be done to verify this on current Composer benchmarks.

dblalock:
Some subtleties here learned from spending a couple hours on #342:

- Need to ensure identical behavior to vanilla batchnorm for inference, so that onnx export doesn't get messed up. Mostly just means calling F.batchnorm2d, etc, instead of layernorm. This is easy to do by subclassing _BatchNorm and overriding forward to start with:
```
if not self.train:
    return super().forward(...)
```
- Getting consistent behavior with variable batch sizes is hard. And variable batch sizes happen anytime `drop_last=False` in the dataloader. You can reshape the inputs in the common case that the layout is NCHW and the ghost batch size is a factor of the batch size, but you need a clean story for handling straggler elements and/or NHWC.

Bearnardd:
Hi @dblalock! Is this still a valid issue? If so I would love to give it a try.

dblalock:
Yes. Go for it!

Bearnardd:
Hi @dblalock! I have had a little time to take a look at this issue. I am currently facing the problem with updating  optimizer parameters with `update_params_in_optimizer` in some test cases. I am not familiar with `MagicMock` but I have done some debugging and in the previous version `len(remove_params)` and `len(added_params)` equaled 0 which was expected with the older version. But with the new implementation the memory addresses differ which results in going further into `update_params_in_optimizer` method. My question is regarding validity of the current tests using mocked optimizer. There is some weird behavior that in this function we have both `added_parms` and `removed params` in the optimizer which would suggest that the optimizer itself has some `param_groups` (I have checked the the test for module_surgery itself, namly `test_update_params_in_optimizer` and that was the case). But in the test cases for `ghost_batchnorm` the `param_groups` for the mocked optimizer is empty which would suggest that it has not been initalized with any parameters, hence I am confused why the added and removed params are also not empty. Could you clarify it for me a bit?

dblalock:
Hi @Bearnardd,

[...]

PR #342, opened 2022-02-02
Revised ghost batchnorm

dblalock:
Addresses #299. Will hopefully yield higher accuracy and greater speed. Won't actually use layernorm because I can't find a good way to get layernorm to update the running stats.

I'm skeptical that there will be much benefit, but going to give it a quick try based on user feedback.


dblalock:
Closing because 1) this is low priority and I don't know when I'll get back to it, and 2) there are a lot of subtleties around making this change in the presence of drop_last=False or non-uniform batch sizes more generally.

PR #348, opened 2022-02-02
Fix cutmix and mixup reliance on num_classes model attribute

Landanjs:
Cutmix and mixup use `state.model.num_classes` to determine the length of one-hot labels, but the torchvision ResNet does not have a `num_classes` attribute.

Instead, I added `num_classes` as a required argument to both algorithms.

Alternatively, I can check if `state.model.num_classes` exists if `num_classes` is not provided to the algorithm, but I think it is unlikely for the model to have `num_classes` and adds a little complexity.


siriuslee (approved):
I like this solution. LGTM

PR #351, opened 2022-02-03
Get rid of `create_from_hparams`

jbloxham:
There's nothing preventing all the logic from being moved into `initialize_object`. Let's do that, so that `trainer.py` is marginally more readable for non-YAHP users.


composer/trainer/trainer_hparams.py
```diff
@@ -265,8 +269,160 @@ def validate(self):
                 "val_dataset and evaluators shouldn't both be specified. Only one can be passed in to the trainer.")
 
     def initialize_object(self) -> Trainer:
-        from composer.trainer.trainer import Trainer
-        return Trainer.create_from_hparams(hparams=self)
+        from composer.trainer import Trainer
```
ravi-mosaicml, line 272: can the import be at the top?

will probably need to be `from composer.trainer.trainer import Trainer` to avoid a circular import error

ravi-mosaicml (approved):
LGTM; didn't read the refactor closely but I presume there aren't other changes tucked in there.

composer/trainer/trainer_hparams.py
jbloxham: Fair enough. It's "standard practice" in other hparams that we import the object at the start of `initialize_object`, but it's just a given that the `Trainer` is gonna need to be imported sooner or later anyways.

ravi-mosaicml: For context, we did this because the object itself imported the hparams class. But since we're not doing this anymore the import can be at the top

PR #352, opened 2022-02-03
Added In Memory Logger, Timestamp Object

### LINKED ISSUE #294: In Memory Logger

ravi-mosaicml:
1. Added a Timestamp object to `composer.core.time`. The Timestamp object is Tuple that represents a snapshot of state.timer
2. Updated the logging API to pass the Timestamp object rather than (epoch, step).
3. Added an In memory logger that records all logging calls, with easy helpers to the most recent values for each key. This new feature required `Timestamp` objects so it is possible, say, to create a plot of samples or tokens vs a logged value. It also cleans up the API compared to storing (epoch, batch, sample, token) as a non-named tuple.

Closes #294

### DIFF
--- a/composer/loggers/file_logger.py
+++ b/composer/loggers/file_logger.py
@@ -10,6 +10,7 @@ import yaml
 
 from composer.core.logging import BaseLoggerBackend, Logger, LogLevel, TLogData, format_log_data_value
 from composer.core.state import State
+from composer.core.time import Timestamp
 from composer.utils import run_directory
 
 
@@ -89,11 +90,11 @@ class FileLoggerBackend(BaseLoggerBackend):
             return self.is_batch_interval
         raise ValueError(f"Unknown log level: {log_level}")
 
-    def log_metric(self, epoch: int, step: int, log_level: LogLevel, data: TLogData):
+    def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
         data_str = format_log_data_value(data)
         if self.file is None:
             raise RuntimeError("Attempted to log before self.init() or after self.close()")
-        print(f"[{log_level.name}][step={step}]: {data_str}", file=self.file, flush=False)
+        print(f"[{log_level.name}][step={timestamp.batch}]: {data_str}", file=self.file, flush=False)
 
     def init(self, state: State, logger: Logger) -> None:
         del state, logger  # unused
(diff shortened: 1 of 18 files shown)


composer/loggers/file_logger.py
```diff
@@ -89,11 +90,11 @@ def will_log(self, state: State, log_level: LogLevel) -> bool:
             return self.is_batch_interval
         raise ValueError(f"Unknown log level: {log_level}")
 
-    def log_metric(self, epoch: int, step: int, log_level: LogLevel, data: TLogData):
+    def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
         data_str = format_log_data_value(data)
         if self.file is None:
             raise RuntimeError("Attempted to log before self.init() or after self.close()")
-        print(f"[{log_level.name}][step={step}]: {data_str}", file=self.file, flush=False)
+        print(f"[{log_level.name}][step={timestamp.batch}]: {data_str}", file=self.file, flush=False)
```
ravi-mosaicml, line 97:
```suggestion
        print(f"[{log_level.name}][step={int(timestamp.batch)}]: {data_str}", file=self.file, flush=False)
```

composer/loggers/wandb_logger.py
```diff
@@ -52,12 +56,15 @@ def __init__(self,
...
         self._init_params = init_params
 
-    def log_metric(self, epoch: int, step: int, log_level: LogLevel, data: TLogData):
-        del epoch, log_level  # unused
+    def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
+        import wandb
+        del log_level  # unused
         if self._enabled:
-            wandb.log(data, step=step)
+            wandb.log(data, step=timestamp.batch)
```
ravi-mosaicml, line 63:
```suggestion
            wandb.log(data, step=int(timestamp.batch))
```

growlix (approved):
LGTM!!

PR #357, opened 2022-02-03
Replace DataSpec with ComposerDataLoader

jbloxham:
Part one of a two-part series in which I hope to make the API for dataloaders more like models. In this part, I move `DataSpec` from `composer.core` to `composer.datasets`, make it extend rather than encapsulate a `torch.DataLoader`, and rename it to `ComposerDataLoader`.

In my opinion, `DataSpec`/`ComposerDataLoader` is a necessary evil that we should continue to try to find ways to remove fields from. Moving it out of core reflects its status as niche. Having it extend `torch.DataLoader` lets us pretend it doesn't exist in much of our type signatures and docs (though maybe some would argue that's a bad thing).

However, the real underlying motivation for this is that in a follow-up PR I want to create subclasses of `ComposerDataLoader` for each of the datasets we support in `composer.datasets`. Currently, there's an awful lot of logic for setting up dataset transformations and synthetic data that is only available in `initialize_object` for YAHP users. This is also making code-driven initialization of our minimal Composer examples a lot more verbose than feels necessary. See [here](https://github.com/mosaicml/composer/blob/dev/examples/adding_custom_models.py) - the dataset initialization is half the length of the example, even with a custom model.


composer/trainer/trainer.py
```diff
@@ -950,7 +950,9 @@ def eval(self, is_batch: bool):
             self.engine.run_event(Event.EVAL_START)
 
             for evaluator in state.evaluators:
-                dataloader = evaluator.dataloader.dataloader
+                dataloader = evaluator.dataloader
+                if not isinstance(dataloader, ComposerDataLoader):
+                    dataloader = ComposerDataLoader(dataloader)
```
jbloxham, line 955: I'm not a huge fan of this cast, but we can't do it in the evaluator because that's in core and `ComposerDataLoader` is not.

However, I'm hoping that in a nearby release we can drop the uses of `ComposerDataLoader`-specific fields from the below.

ravi-mosaicml (commented):
Overall, code looks good, except for one comment re: re-wrapping. I think after fixing this, it could make the code more less intuitive than the current `DataSpec` approach, since you'll always have to check for wrapping and can't use a typical isinstance check (as you could with `DataSpec`, since `DataSpec` would never be wrapped)

First, before working on this more, can you run it by @hanlint ? In other PRs, he commented how the `DataSpec` should ideally become a dict. The current `Trainer()` allows you to pass in a `dict` for the `DataSpec`; this would remove that feature. While I don't think this change would violate the style guide since I would consider a `DataSpec` to be a power-user feature (and thus be allowed another import), I would like to hear his thoughts on it.

In addition, one other concern I have is that if a user has a dataset where they will need to implement, say, a custom batch split function, then they will need to know about the `ComposerDataLoader` class. The current `Union[DataLoader, DataSpec]` typing signature at least hints that there is another option available. I think this can be solved through a kickass docstring, though.

hanlint:
I agree that `DataSpec` is a necessary evil, and would like to minimize its fields. I am of similar opinion regarding `Evaluator` in the longer-term. However, I am not convinced by the `ComposerDataLoader` approach, for a few reasons beyond what @ravi-mosaicml had mentioned above:
1. `ComposedDataLoader` is more complex and unintuitive regarding our intentions. We'd like to attach metadata, but users would interpret that we are/may modify the `__iter__` and other core dataloading logic, making integration with their own datasets difficult. `DataSpec` clearly delinates a boundary between the pytorch object and the metadata.
2. Discoverability is harder in the trainer API signature in this case.
3. I suspect with all the various datatypes out there, `split_batch_fn` will be commonly required, and eventually i'd like for users to pass in a `dict` with this config without needing to import. 

I do agree with moving `DataSpec` from core to Datasets. 

An alternate approach to address the apt observation on `initialize_object` is to offer `create_imagenet_dataset` methods, rather than creating dataset-specific objects? Would that work @jbloxham ?

Side note: multiple places in our API we need to attach metadata to a pytorch-native objects (e.g. `dataloader`, `schedulers` (step or interval), `eval_dataloader` (metrics), `model` (many)). While the complexity of the metadata dictates our implementation, perhaps we can standardize on a consistent set of rules here? For example:
1. When we do not anticipate needing to override the pytorch object methods, we use `dict` or a `___Spec` (or `__Config` might be less surprising to data scientists). We strive to minimize the number of fields here.
2. Otherwise, we extend the class. For example, `ComposerModel` is required because the `torch.nn.Module.forward()` needs to be overridden.

There could be exceptions of course for complexity or consistency with previous practice in the community.

ravi-mosaicml:
This is going on a tangent now, but I actually think it would be simplifier if we switched the model class to be a `ModelSpec`, similar to dataloader spec. There is no need to subclass it, other than that's what PTL does. We always have to "unwrap" the model to retrieve the original loss function, metrics, etc... after it has been wrapped by DeepSpeed or DDP. It would definitely simplify things if we had a `ModelSpec`, where only the model attribute was wrapped. So, from a style guide perspective, I would prefer the config/spec/dict approach.

ravi-mosaicml:
> An alternate approach to address the apt observation on initialize_object is to offer create_imagenet_dataset methods, rather than creating dataset-specific objects? Would that work @jbloxham ?

Can you elaborate? Sorry, a bit confused on what this would look like and what this would accomplish?

composer/trainer/trainer.py
```diff
@@ -275,11 +275,11 @@ def __init__(
         if isinstance(precision, str):
             precision = Precision(precision)
 
-        if not isinstance(train_dataloader, DataSpec):
-            train_dataloader = DataSpec(train_dataloader)
+        if not isinstance(train_dataloader, ComposerDataLoader):
```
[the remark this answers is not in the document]
jbloxham: Hmm, in the event of `dataloader = ProfiledDataLoader(ComposerDataLoader(TorchDataLoader))`, wouldn't `dataloader._split_batch_fn`, for example, no longer work, anyways?

Also, looking at the code for how `ProfiledDataLoader` is used, there shouldn't be any risk of `_train_data_spec` being wrapped in a `ProfiledDataLoader` - only `state.train_dataloader` gets wrapped.

composer/trainer/trainer.py
jbloxham: The trouble with that is that `Evaluator` is in core, but it can't import `ComposerDataLoader` since that's been moved to `datasets`. I defend `ComposerDataLoader` being moved out of core, since its logic is only relevant to the Composer Trainer.

jbloxham:
> This is going on a tangent now, but I actually think it would be simplifier if we switched the model class to be a `ModelSpec`, similar to dataloader spec. There is no need to subclass it, other than that's what PTL does. We always have to "unwrap" the model to retrieve the original loss function, metrics, etc... after it has been wrapped by DeepSpeed or DDP. It would definitely simplify things if we had a `ModelSpec`, where only the model attribute was wrapped. So, from a style guide perspective, I would prefer the config/spec/dict approach.

I like the idea of this, but I'm not sure what it would look like in practice. The `validate`, `loss`, and `metrics` functions are all critical for the Composer Trainer, and generally lack reasonable defaults. I guess if `ModelSpec` were a dict, that would be one less object for users to import, but come on, that alone doesn't meaningfully reduce the complexity of creating a custom model that can be used with the Composer Trainer.

ravi-mosaicml:
Going to close this PR for now. Will create another one to move `DataSpec` from core to datasets.

PR #359, opened 2022-02-03
Fix Checkpoints

### LINKED ISSUE #355: Add save_compression to trainer hparams
The save_compression flag is missing in trainer_hparams. Surprised pyright didn't catch this in the test case...

### LINKED ISSUE #358: Remove `config` from `Trainer(config: ...)`
Let's remove the `config` argument from the `Trainer` constructor. This is a very yahp-centric field which was added to save the config to the checkpoint folder.

Unlike one might expect from its name, it does not actually affect the configuration of the trainer.
In addition, it assumes that the user has the config in the form of a dictionary. If one is using a notebook, there likely isn't any config. And if one stores config outside of composer (and, say, calls composer through the CLI), then this field is irrelevant.

Instead, the entrypoint should handle saving the config to the checkpoint folder. After all, the entrypoint (not the trainer) is responsible for reading the dumped config from the checkpoints folder.

ravi-mosaicml:
1. Remove config from trainer; fixes #358
2. Add save_compression to trainer_hparams; fixes #355. Fixed checkpointing test.

### DIFF
--- a/composer/trainer/checkpoint.py
+++ b/composer/trainer/checkpoint.py
@@ -34,8 +34,15 @@ _COMPOSER_STATES_FILENAME = "composer_states.pt"
 _DEEPSPEED_TAG = "deepspeed"  # always tag with the same, deterministic name. We'll rename the tarball to the appropriate name.
 
 
-def get_composer_checkpoint_filepath(checkpoint_folder: str):
-    return os.path.join(checkpoint_folder, _COMPOSER_STATES_FILENAME)
+def _format_path_with_rank(path: str, rank: int):
+    """Returns the path with ``{{RANK}}`` substituted with the ``rank`` argument.
+    See the :class:`CheckpointLoader` docs for a description of how this is used.
+
+    Args:
+        path (str): Path to format
+        rank (int): The rank
+    """
+    return path.format(RANK=rank)
 
 
 class CheckpointLoader:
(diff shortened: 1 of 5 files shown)


composer/trainer/checkpoint.py
```diff
@@ -383,15 +384,14 @@ def should_checkpoint(self, state: State, event: Event) -> bool:
...
 
-    def save_checkpoint(self, state: State, seed: int, device: Device, config: Optional[Dict[str, Any]] = None) -> None:
+    def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
         """Save the current state to a a new checkpoint file.
 
         Args:
             state (State): The current State of the trainer.
             seed (int): The seed used for random number generation.
             device (Device): The Device in use by this process.
             ddp (DDP): The DDP engine in use by this trainer.
```
jbloxham, line 394: Tangential, but this looks obsolete.

composer/trainer/checkpoint.py
```diff
@@ -166,9 +165,9 @@ def _download_checkpoint(self, node_checkpoint_folder: str) -> Tuple[str, Option
                 :meth:`deepspeed.DeepSpeedEngine.load_checkpoint`.
         """
         checkpoint_archive_name = self.path.split(os.path.sep)[-1]
-        rank_zero_checkpoint_archive_name = "rank_0." + checkpoint_archive_name.format(rank=0)
+        rank_zero_checkpoint_archive_name = "rank_0." + checkpoint_archive_name.format(RANK=0)
```
jbloxham, line 168: Nit: The formatting logic for checkpoint names is specific enough that it ought to be in its own function.

jbloxham (approved):

composer/trainer/checkpoint.py
ravi-mosaicml: We really need a doc linter lol...thanks!

composer/trainer/checkpoint.py
ravi-mosaicml: Good point, refactored it. Thanks!

PR #371, opened 2022-02-03
Testing trainer.fit on each algorithm, callback, logger, and profiler

ravi-mosaicml:
Added a test that ensures we can train on each algorithm, callback, logger, and profiler individually

Addressed issues identified throughout our test cases and the algorithms


bandish-shah (approved):
LGTM

Issue #329, opened 2022-02-01
Multiple Evaluator Improvements


ravi-mosaicml:
List of improvements to clean up for multiple evaluators:

1. Remove `validate_every_n_epochs` and `validate_every_n_batches` from the trainer hparams and init signature, instead add a `validate_interval` to the Evaluator class
2. Remove `val_dataset` and `eval_batch_size` from the trainer hparams. Require hparam users to use the `evaluators` field
3. Add in `eval_subset_num_batches` as a property for each evaluator

_Originally posted by @ravi-mosaicml in https://github.com/mosaicml/composer/pull/120#discussion_r796858797_

ravi-mosaicml:
Something to work out is how TQDM will work, since it needs to know the length of evaluation to show the progress bar. If the evaluators are removed from state, then TQDM will no longer be able to determine this.

We need to make TQDM work with datasets of unknown # of batches anyways (i.e. for streaming dataloaders). One thought I have is to add `train_num_samples`, `train_num_tokens`, `eval_num_samples`, and `eval_num_tokens` to the state. The trainer would set these properties on init (they are available as the `DataSpec.num_samples` and `DataSpec.num_tokens`). Then, we would have the TQDM bar `total` parameter be in terms of samples (not batches), and on each step, we would update it with the number of samples trained. This should be done in a separate issue before implementing #329.

PR #374, opened 2022-02-04
removed byo_trainer

anisehsani:
Don't think we are supporting byo_trainer anymore and instead are focusing on using the functional forms of algorithms.

This example makes it a bit more difficult difficult to implement( #329 ).

Finally, don't think there are any tests on this either.


ravi-mosaicml (approved):
:+1:

PR #375, opened 2022-02-04
Updated Model Cards

ajaysaini725:
Introduces the following:
- New standardized format for all model cards + updated all model cards to reflect the new format
- Fixes to links on the models summary page
- Removed hparams from the UNet `__init__` signature (was necessary to cleanly document UNet + we are removing hparams from primary codepaths anyway)

To reviewers: I just added everyone who was relevant to the model docs I updated


docs/source/model_cards/cifar_resnet.md
```diff
@@ -39,15 +40,33 @@ The biggest differences between CIFAR ResNet models and ImageNet ResNet models a
...
-    - Momentum: 0.9
-    - Weight decay: 1e-4
-- Batch size: 1024
-- LR Schedulers
-    - Linear warmup for 5 epochs
-    - Multistep decay by 0.1 at epochs 80 and 120
-- Number of epochs: 160
\ No newline at end of file
+```yaml
+optimizer:
```
ravi-mosaicml, line 46: Could we provide a functional code example that someone could run with these values? I.e. that does not depend on yahp?
Ideally the code is in a separate python file that has a test case to ensure that the trainer initializes (and maybe trains a batch on dummy data)

docs/source/model_cards/efficientnet.md
```diff
@@ -43,7 +40,48 @@ Tan and Le included 8 members in their model family. The goal was for each famil
 
 ## Default Training Hyperparameters
 
-Our default hyperparameters are identical to the Nvidia Deep Learning Examples except:
+We use the following default hyperparameters from the [Nvidia Deep Learning Examples](https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/Classification/ConvNets/efficientnet):
+
+```yaml
+optimizer:
+  rmsprop:
```
ravi-mosaicml, line 47: Same thing here

ravi-mosaicml (approved):
I didn't take a thorough read through the docs, but from a glance looks fine. I would prefer that all examples are actual code snippets that are stored in python files, covered by test cases, and then dynamically embedded in the docs.

docs/source/model_cards/cifar_resnet.md
ajaysaini725: Oh this was not supposed to be a YAHP-related example at all. I just used yaml format because it was the cleanest way to show a list of hyperparameters used for training (kind of like how NVIDIA DL Examples shows config files for all their models). 

The point of this example is just to show user's what values for each parameter they can use when training with the model (not necessarily how to use the model with our trainer) in a relatively clean UI. There are plenty of non-yahp examples in the upcoming new `Using the Trainer` section of the docs.

ajaysaini725:
> I didn't take a thorough read through the docs, but from a glance looks fine. I would prefer that all examples are actual code snippets that are stored in python files, covered by test cases, and then dynamically embedded in the docs.

There will be a lot of code examples in the new incoming docs - would you like me to do that now for this PR or should we wait until all docs are in and then do that for all docs at once? The latter seems easier

PR #376, opened 2022-02-04
Fix sample SD inference multiplication

Landanjs:
Fixes bug in sample-wise stochastic depth. The fix appears to have a slight (~0.1%) accuracy improvement over the buggy implementation. [Fixed implementation wandb](https://wandb.ai/mosaic-ml/landan-test/groups/resnet50_baseline_og_sample_sd/workspace?workspace=user-landanjs) and [bug implementation wandb](https://wandb.ai/mosaic-ml/landan-test/groups/og_sample_sd/workspace?workspace=user-landanjs)


ravi-mosaicml (approved):
I trust that this was a bug (and am approving the PR) but for my curiosity, can you elaborate why this was a bug? Shouldn't we be running the algorithm only when training?

Landanjs:
The original stochastic depth method randomly drops sets of residual blocks. The authors claim that this scales activation magnitude in proportion to the stochastic depth probability. I don't think they actually verify this, but it likely based on a similar concept with dropout. 

So, to ensure the activation magnitude is the same after using stochastic depth, the authors recommend multiplying by the probability during evaluation. Another formulation (supposedly equally valid) is to divide by the probability when training. It may be good to test if these are the same in the future.

Issue #344, opened 2022-02-02
Have defaults for functional API function signatures whenever applicable


dblalock:
Right now, some of our functions have default values, while others do not, even within a class of methods (e.g., there are different choices across different data augmentation methods). We should have a consistent choice across all similar methods (in this case, to always have sane defaults).

Note that Alibi and Stochastic Depth can't have sane defaults at present since they need to know network-specific module names. They are therefore out of scope.

PR #377, opened 2022-02-04
Add defaults to functional API / share defaults across interfaces

dblalock:
Addresses #344.

Right now most of the ways to use most of our algorithms have default values for their hparams wherever possible. But some algorithms don't have defaults for some interfaces. In particular, there are various algorithms where the AlgorithmHparams have defaults but the functional form does not.

This PR:

1. Propagates defaults across all interfaces, meaning that the Algorithm, AlgorithmHparams, and functional API share the same default hparams.
2. Adds two defaults that weren't present before. First:

    1. `mixup_batch` now has a default interpolation lambda of 0.9, which is a reasonable value that a Beta(.2,.2) might generate. @coryMosaicML 
    1. stochastic depth now defaults to `stochastic_method='block'`. @Landanjs 

3. Sometimes giving an argument a default necessitated moving it earlier in the function signature, so that args without defaults precede all args with defaults. In this case, I reordered the arguments at most call sites as well. I think this only affects stochastic depth and its tests.


ravi-mosaicml:
Currently we are defining default values in multiple places:
* Hparams
* `__init__` signature
* For the cases of algorithms, in the functional form
* Docstrings for all of the above

To reduce the change of errors, I'd really prefer to have the default value defined in code in just one place.  This will make sure it's consistent and doesn't get out of date anywhere if it changes later.

Then, wherever we need the default value, we can use the variable.

For example:

```python3
_DEFAULT_ALPHA = 0.15

def my_awesome_alg_functional(alpha: float = _DEFAULT_ALPHA):
  f'''best docstring ever
      args:
        alpha (float, optional): alpha param (default: ``{_DEFAULT_ALPHA}``)'''
  ...

class MyAwesomeAlgorithm(Algorithm):
  f'''best docstring ever
      args:
        alpha (float, optional): alpha param (default: ``{_DEFAULT_ALPHA}``)'''
    def __init__(self, alpha: float = _DEFAULT_ALPHA):
        self.alpha = alpha
        ...
class MyAwesomeAlgHparams(AlgHparams):
  alpha: float = hp.optional("alpha param", default=_DEFAULT_ALPHA)
```

I believe this would work with sphinx / rst, we should do a test build before converting everything over.

dblalock:
@ravi-mosaicml  Defining defaults as constants in one place is also what I would prefer, but @hanlint and (I think?) @ajaysaini725 wanted to inline the defaults to make each usage look more self-contained.

dblalock:
Also, yes, this works fine with RST. It just shows the value, not the variable name. 

Related, we actually can't use format strings as docstrings, but we can do what torch does and set `__doc__  = f"""blah blah, my_var: default = {my_var}"""`

dblalock:
Comment more likely to result in resolution: @hanlint are you up for revisiting this based on ravi's concerns, or should we stick with replicating default values everywhere?

hanlint:
Replicate defaults everywhere -- e.g. see duplication in multiple k_means methods and between the functional and the class in [sci-kit learn](https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/cluster/_kmeans.py) or between [conv2d](https://pytorch.org/docs/stable/generated/torch.nn.functional.conv2d.html#torch.nn.functional.conv2d) and [Conv2d](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) in pytorch. 

If desired, we can add a signature check test that makes sure all three have matching defaults as an automated way to maintain equality.

dblalock:
Jenkins is now appeased. This should be ready to re-review.

hanlint (approved):
LGTM, perhaps ask:
* @growlix and @coryMosaicML to look over the data augmentation dfeaults
* @Landanjs  -- stochastic depth default to block?

PR #378, opened 2022-02-04
Remove requirement for `validation` and `metrics`

hanlint:
Users who just want to run training, and not validate on an external dataset should not need to supply validation and metrics as part of their `ComposerModel`. Especially now that the `eval_dataloader` is an optional argument to our trainer. This PR removes the abstract marker for the two methods in question.

h/t: @A-Jacobson

### DIFF
--- a/composer/models/base.py
+++ b/composer/models/base.py
@@ -45,7 +45,6 @@ class ComposerModel(torch.nn.Module, abc.ABC):
         """
         pass
 
-    @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


A-Jacobson (approved):
Thanks! you beat me to it.

ravi-mosaicml:
Can you have these functions raise a `NotImplementedError` instead of just a pass? Will help avoid nasty errors if a user doesn't implement something but we still call the method. Otherwise LGTM

Issue #364, opened 2022-02-03
Standardized + clearer functional API names


dblalock:
In accordance with functional API discussion + design doc, we're going to make function names 1) verb_noun, and 2) standardized as much as possible, with respect to both each other and their associated Algorithm classes.

Examples:
`augment_and_mix` -> `augmix_image`
`cutmix` -> `cutmix_batch`
`scale_scheduler` -> `scale_schedule`
`factorize_conv2d_modules` -> `apply_factorization`

PR #380, opened 2022-02-04
Functional API renaming part 1

dblalock:
Addresses #364. Just renames primary data augmentation algorithm functional API functions. Helper functions out of scope since we should probably make most of these private.

Changed names:
`augment_and_mix -> augmix_image`
`colout -> colout_image`, `batch_colout -> colout_batch`
`cutmix -> cutmix_batch`
`cutout() -> cutout_batch()`
`randaugment() -> randaugment_image()`

Splitting this into 2 PRs to keep each one nice and lightweight, since renames for different algorithms are independent.


ravi-mosaicml (approved):
I presume this is just a refactor / docs change; lgtm.

PR #384, opened 2022-02-04
SWA update

growlix:
`TrainerHparams` doesn't account for `train_subset_num_batches` when calculating quantities such as `samples_per_epoch`, `tokens_per_epoch`, and `steps_per_epoch`. This wrecks things like LR schedules when using `train_subset_num_batches`.


growlix:
btw the name of the branch is because this was originally supposed to be part of a fix for swa and the schedulers (which stopped playing nicely together at some point), but this bug surfaced during testing. Not sure if it makes sense to split them into separate PRs

growlix:
Added a potential fix for `ComposedScheduler`, which now has attribute `_last_lr` and as such won't crash when is called `get_last_lr()` is called

ravi-mosaicml (approved):
LGTM, thanks

composer/trainer/trainer_hparams.py
```diff
@@ -343,13 +343,22 @@ def initialize_object(self) -> Trainer:
...
             steps_per_epoch = None
 
         batch_size = None
         if train_dataloader.batch_size is not None:
             batch_size = train_dataloader.batch_size * dist.get_world_size()
 
+        # Adjust tokens per epoch based on train_subset_num_batches
+        if self.train_subset_num_batches is not None and batch_size is not None and steps_per_epoch is not None:
+            if tokens_per_epoch is not None:
+                tokens_per_epoch = batch_size * self.train_subset_num_batches
```
ravi-mosaicml, line 358: I don't think this is correct, since each batch will have a different number of tokens. But nothing is using this right now...

growlix: @moinnadeem, is there an easy solution to this?

hanlint:
We sycned offline -- likely this is breaking the checkpoining tests because those tests use `train_subset_num_batches`. This is a low prioirty PR, so going to defer this merge until after v0.4

growlix:
This PR now only addresses SWA. `train_subset_num_batches` will be moved into a separate PR.

hanlint:
Closing this for #555

PR #386, opened 2022-02-05
Docs Refactor

ravi-mosaicml:
This PR significantly refactors the docs to support an API reference. Specifically,

1. It has just one **indexed** `autosummary` for all of composer. This adds an API reference section. All other uses of autosummary were removed. (Well, it will have one just indexed autosummary. We need to refactor the documentation for `composer.types` so none if its entries are indexed, but that's for next week).
2. It adds a custom hook to add summary tables that contain an entry for each exported class, function, attribute, or exception. It takes special care to ensure that it links to the original definition of each item to avoid duplicated, indexed sphinx entries.
3. It merged the `composer.core` docs and `composer` docs into a "Composer Walkthrough", and adds links to the API reference at the bottom of each page.
4. It remove the dummy algorithm and algorithm template, since that doesn't belong in core composer code (and it looks funny to document that next to real algorithms). We have enough algorithms now that these aren't needed.
5. It guards the gpt scaling laws generator with `if __name__ == "__main__"` so it isn't called by the sphinx build

This PR, however, leaves plenty of work to be done:

1. `__all__` needs to be defined on every public module. Otherwise, sphinx does not know what was an explicit reimport (like we do in an `__init__.py` vs what was an import that is used by the code). Currently, it's documenting thrid-party modules that we do not want. Using `__all__` will not make the code any more verbose, as you no longer need to do `from XX import YY as YY`; you can just do `from XX import YY` so long as `YY` is included in `__all__`.
2. All public modules need docstrings


Averylamp (approved):
lgtm

Issue #383, opened 2022-02-04
Single GPU models don't load weights from checkpoint


siriuslee:
** To reproduce

```
import torch
from composer.trainer import Trainer
from composer import models

model = models.CIFAR10_ResNet56()
trainer = Trainer(model=model, train_dataloader=train_dataloader, eval_dataloader=None, max_duration="1ep", save_folder="checkpoints")
trainer.checkpoint_saver.save_checkpoint(trainer.state, trainer.seed, trainer.device)
new_model = models.CIFAR10_ResNet56()
new_trainer = Trainer(model=new_model, train_dataloader=train_dataloader, eval_dataloader=None, max_duration="1ep", load_path=os.path.join(trainer.checkpoint_saver.checkpoint_folder, "ep1.tar"))
assert torch.allclose(model.module.conv.weight, new_model.module.conv.weight)
```

## Expected behavior

The new model should have received the checkpoint's weights during the init of `new_trainer`. It doesn't and the reason seems to be due to here: https://github.com/mosaicml/composer/blob/dev/composer/core/state.py#L296-L302. If DDP is not enabled, the `state_dict` is never restored for the model.

ravi-mosaicml:
Fixed in #389

Issue #367, opened 2022-02-03
Script to run the tests locally such that they all pass


dblalock:
## Motivation

If a user downloads composer and wants to either:

1. make sure everything installed okay, or
2. try modifying it without breaking everything (e.g., to add a new method), 

it's really helpful to be able to run the full test suite and know whether failures are because of your code. I know this is helpful (at least for case 2) based on my experience developing methods.

Also, having a command to run the tests post-install is just a pretty common element of installation documentation.

## [Optional] Implementation

Below is equivalent to what I run to get everything passing locally. I say "equivalent" because I don't setup the environment and reinstall pillow-simd regularly; I just run the last command.

The sketchy part here is getting pillow-simd installed so that test_pillow_simd passes. Might be more robust to just have it skip that test.
```
# this block is so that test_pillow_simd passes
pip uninstall pillow
pip uninstall pillow-simd 
pip install pillow-simd  # needs certain deps; might choke

# this block is to avoid a few torch.distributed failures in test_checkpoint, test_ddp, and a couple other files
export NODE_RANK=0
export LOCAL_RANK=0
export LOCAL_WORLD_SIZE=1

# actual testing command
pytest -s --tb=short --disable-pytest-warnings --test_duration all tests/
```

## Additional context

`bash scripts/test.sh` doesn't currently allow the desired behavior :

```
~/code/composer $ bash scripts/test.sh                                              
+ JUNIT_PREFIX=build/output/composer
++ dirname build/output/composer
+ mkdir -p build/output
+ python -m composer.cli.launcher -n 1 --master_port 26000 -m coverage run -m pytest --junitxml build/output/composer.n1.junit.xml
/Users/davis/.pyenv/versions/3.8.6/bin/python: No module named coverage
Waiting up to 30 seconds for all training processes to terminate...
Process 1566 excited with code 1
ERROR:__main__:Subprocess 1566 exited with code 1
```

ravi-mosaicml:
Great points raised here! Can you add this to the `CONTRIBUTING.md` document?

If you are developing on composer, one should install `pip install mosaicml[dev]`

> # this block is to avoid a few torch.distributed failures in test_checkpoint, test_ddp, and a couple other files
> export NODE_RANK=0
> export LOCAL_RANK=0
> export LOCAL_WORLD_SIZE=1

@jbloxham can you create a PR to fix these tests? I think we'll need to guard torch.distributed initialization unless if all these variables are set.

ravi-mosaicml:
[...]

PR #387, opened 2022-02-05
Use `Makefile` instead of scripts; enable easier testing

hanlint:
We should use `Makefile` instead of scripts, as they are more configurable and centralized. This PR replaces `lint.sh` with `make style` and `make lint`, and when completed, should also replace `test.sh` with `make test-all` and its variants. 

Jenkins should also call the make commands directly instead of customized pytest commands, as they are difficult to reproduce locally if debugging needs arise.

Additionally, to enable local running of all pytests, as noted in #367:
* runs basic tests with `make test`
* runs gpu tests and deepspeed tests with `make test-gpu` and `make test-deepspeed`
* runs DDP-based tests with the `composer` launcher with `make test-ddp`
* runs all tests with `make test-all`
* when running single GPU tests _without_ the composer launcher, disables `test_ddp`. We should be running this configuration in jenkins as well to catch errors such as #383 . 
* when running on a non-Linux system, disables `test_pillow_simd`

Some additional changes:
* shorten `--test-duration` -> `--duration`
* remove `composer.ipynb` and tests in advance of major notebooks upgrade
* fix `memory_monitor_cpu` test 

Remaining TODOs:

- [x] enabling passing in of junit configuration
- [ ] replace the jenkinfile with calls to `make`


setup.py
```diff
@@ -72,6 +72,7 @@ def package_files(directory: str):
     'jupyter>=1.0.0',
     'yamllint>=1.26.2',
     'pytest-timeout>=1.4.2',
+    'pyright>=0.0.13',
```
ravi-mosaicml, line 75: neat!

tests/trainer/test_ddp.py
```diff
@@ -138,7 +138,8 @@ def test_ddp(device: DeviceHparams, world_size: int, composer_trainer_hparams: T
     We assert that each of these tensors are different to ensure that 1) random seeding works properly,
     and 2) each ddp process is indeed getting different data.
     """
-    del world_size  # unused. Set via env variables
+    if world_size == 1 and any([e not in os.environ for e in ['NODE_RANK', 'LOCAL_RANK', 'LOCAL_WORLD_SIZE']]):
```
ravi-mosaicml, line 141: The DDP tests should work even when DDP isn't configured, since they should default to just using a world_size = 1, node_rank = 0, local_rank = 0, local_world_Size = 1. So this test shouldn't be skipped; there's actually a bug somewhere.

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	pytest tests/ -m deepspeed --test_duration all
+
+# run all tests, including mgpu tests
+# uses the composer launcher script to properly configure
+# mgpu cases
+test-ddp:
+	python -m composer.cli.launcher -n 1 -m pytest --test_duration all
+	python -m composer.cli.launcher -n 2 -m pytest --test_duration all
+
+test-all: test test-gpu test-ddp
```
ravi-mosaicml, line 34: This will skip long-running cpu only tests. 

Instead of hardcoding test_duration, can you put that as a makefile argument, and have it default to `all`?
Then, if someone wanted to do short tests, they could do `make test TEST_DURATION=short`

Also,  can you mark all of these targets as `.PHONY`.?

tests/trainer/test_ddp.py
hanlint: I think I found the issue -- we initialize `dist` even when the world_size is 1 in our trainer.

Since we don't support using DDP without using our `composer` launcher, perhaps we shouldn't be testing that combination anyway?

ravi-mosaicml: Yup, that is one of the issues, and there are a few other issues that we have to address. This PR runs the tests without DDP, so it should fail until we fully iron out the non-ddp path. https://github.com/mosaicml/composer/pull/389
That PR should address any DDP issues in this PR.

Makefile
hanlint: done

tests/trainer/test_ddp.py
hanlint: removed the skip

hanlint:
Should be ready for a re-review. Addressed all changes, and made a few additional
* `make style` now autoformats the code, where as `make lint` just does the check. This makes it easy for developers to autoformat there code to the pass the linting (except for typing)
* reduced the verbosity of the output for `isort`.

Makefile
```diff
@@ -0,0 +1,39 @@
+# several pytest settings
+DURATION ?= all  # pytest duration, one of short, long or all
+WORLD_SIZE ?= 1  # world size for DDP tests
+EXTRA_ARGS ?=  # additional arguments
+
+EXTRA_ARGS := --duration $(DURATION) $(EXTRA_ARGS)
```
dblalock, line 6: Can we add `--tb=short --disable-pytest-warnings` also? Test failures create super long and hard-to-localize traces without `--tb=short`. For warnings, almost all of them are from deep in the bowels of third-party libraries. It would in theory be good practice to read them, but it's such a wall of 95% useless text that I doubt anyone will (I almost never do).

Makefile
```diff
@@ -0,0 +1,39 @@
...
+	pytest tests/ -m deepspeed $(EXTRA_ARGS)
+
+# run all tests, including multi-gpu tests
+# uses the composer launcher script
+test-with-ddp:
+	python -m composer.cli.launcher -n ${WORLD_SIZE} -m pytest $(EXTRA_ARGS)
+
+test-all: test test-gpu test-deepspeed test-with-ddp
+
+.PHONY: test test-gpu test-with-ddp test-deepspeed
```
dblalock, line 39: `style` and `lint` also PHONY?

tests/conftest.py
```diff
@@ -34,13 +34,12 @@
...
-        choices=["short", "long", "all"],
-        help=
-        "Whether to run short tests (the default), long tests, or all tests. A test is considered short if its timeout is "
-        "less than that specified in the config.")
+    parser.addoption("--duration",
+                     default="short",
+                     choices=["short", "long", "all"],
+                     help="""Duration of tests, one of short, long, or all.
+                             Tests are short if their timeout < 2 seconds
+                             (configurable threshold). Default: short.""")
```
dblalock, line 42: would be nice to have default ("short") match default in the makefile ("all").

tests/test_pillow_simd.py
```diff
@@ -1,7 +1,14 @@
...
+import platform
+
 import PIL
+import pytest
 
 
+@pytest.mark.skipif(
+    'Linux' not in platform.system(),
+    reason="Pillow-SIMD not available on non-Linux systems.",
+)
```
dblalock, line 12: I have it on my intel Mac, so technically this comment isn't quite correct; but it's hard enough to install that I don't mind skipping the test.

dblalock:
These are likely out of scope for this PR (last two fixed by #389?), but I hit three issues when running `make test`:

1. Two tests in `test_trainer.py` time out (`test_multiple_calls_to_fit[WandBLoggerBackendHparams]` and `test_fit_on_all_callbacks_loggers_algs_profilers[WandBLoggerBackendHparams]`)
2. If you `unset NODE_RANK; unset LOCAL_RANK; unset LOCAL_WORLD_SIZE` before running it, there are 16 additional failures.
3. The test command hangs seemingly indefinitely after tests finish running. This makes `test-all` just not work right now. It runs the `test` and never gets to `test-gpu`, etc.

Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+test-gpu:
+	pytest tests/ -m gpu $(EXTRA_ARGS)
+
+test-deepspeed:
+	pytest tests/ -m deepspeed $(EXTRA_ARGS)
+
+# run all tests, including multi-gpu tests
+# uses the composer launcher script
+test-with-ddp:
```
dblalock, line 34: maybe just `test-ddp`? Slight empirical evidence that this is a more natural name, in that this is what I assumed it was + tried to type into my shell after I ran `test-gpu` when trying out each target.

dblalock (approved):
Approving to avoid another round-trip since I had only small nitpicks. (Assuming that the failures + hang I noted are out of scope for this PR.)

hanlint:
@dblalock Thanks for testing. Oddly, I am not many any of the above issues testing on my local mac, nor in the composer docker. I'll discuss with you offline and maybe we can do some debugging.

Makefile
hanlint: I'm going to default to the same verbosity as `pytest` itself. However, users can adjust that directly, e.g.:
```
make test EXTRA_ARGS=--tb-short --disable-pytest-warnings
```

We've already filtered out most of the warnings that can be safely ignored (see pyproject.toml), so there shouldn't be as many now.

Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+test-gpu:
+	pytest tests/ -m gpu $(EXTRA_ARGS)
+
+test-deepspeed:
+	pytest tests/ -m deepspeed $(EXTRA_ARGS)
+
+# run all tests, including multi-gpu tests
+# uses the composer launcher script
+test-ddp:
```
ravi-mosaicml, line 34: nit: `ddp` refers to a specific distributed backend, so we started to call it `dist` instead

```suggestion
test-dist:
```

Makefile
```diff
@@ -0,0 +1,39 @@
...
+lint:
+	isort $(dirs) -c
+	yapf -dr $(dirs)
+	docformatter -r --wrap-summaries 120 --wrap-descriptions 120 $(dirs)
+	pyright $(dirs)
+
+test:
+	pytest tests/ $(EXTRA_ARGS)
+
+test-gpu:
```
ravi-mosaicml, line 26: We will want to run the gpu and deepspeed tests with and without the launcher script, since some of the gpu and deepspeed tests have world sizes of 2. I'm wondering whether it would make sense to have gpu or deepspeed be a flag like we have with world_size?

Makefile
```diff
@@ -0,0 +1,39 @@
...
+	pytest tests/ -m deepspeed $(EXTRA_ARGS)
+
+# run all tests, including multi-gpu tests
+# uses the composer launcher script
+test-ddp:
+	python -m composer.cli.launcher -n ${WORLD_SIZE} -m pytest $(EXTRA_ARGS)
+
+test-all: test test-gpu test-deepspeed test-ddp
+
+.PHONY: test test-gpu test-ddp test-deepspeed lint, style
```
ravi-mosaicml, line 39:
```suggestion
.PHONY: test test-gpu test-ddp test-deepspeed test-all lint style
```

composer/callbacks/memory_monitor.py
```diff
@@ -63,8 +63,13 @@ def get_memory_report() -> Dict[str, Union[int, float]]:
...
         log.debug("Cuda is not available. The memory report will be empty.")
         return {}
-    device_stats = torch.cuda.memory_stats()
-    memory_report = {}
-    for torch_stat_name, stat_alias in _MEMORY_STATS.items():
-        memory_report[stat_alias] = device_stats[torch_stat_name]
+    memory_stats = torch.cuda.memory_stats()
+
+    if len(memory_stats) == 0:
+        log.debug("No GPU memory was used, returning empty.")
```
ravi-mosaicml, line 69: Could we raise an exception rather than return an empty dict if no gpu is available? An empty dict will clutter the logs.

The memory monitor should only be used if on GPU.

tests/test_pillow_simd.py
```diff
@@ -1,7 +1,13 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+import os
+import platform
+
 import PIL
+import pytest
 
 
+@pytest.mark.skipif('composer-python' not in os.environ["PATH"] or 'Linux' not in platform.system(),
```
ravi-mosaicml, line 10: Hmm, could we use a different check than `'composer-python' not in os.environ["PATH"]` to determine if we're in the composer docker? This test should not be linked to the python path of the dockerfile.

Instead, would it work if we checked to see if we were running in a CI environment?
e.g. `if os.environ.get("CI") != "true"`  See: https://ci.int.mosaicml.com/env-vars.html/
Granted one may be running the docker image outside of CI/CD, in which case this test would not run, but I think that would be fine since it would still run for our builds?

I would also be OK with removing this test and instead having a test script to run when building the docker image (and putting a test like this in there). (OK with punting this change for now, until we have more jenkins cycles)

ravi-mosaicml (commented):
Overall LGTM. Will re-review once jenkins is passing.

Makefile
hanlint: Unfortunately, `dist` is usually the shorthand for `torch.distributed`, which is synonymous with `ddp` so not sure the right name here.

composer/callbacks/memory_monitor.py
hanlint: We do raise an exception if no GPU is available (see a few lines above). This is for the case where a GPU is available, but since no GPU memory was allocated, the memory stats is empty.

Makefile
ravi-mosaicml: Would it be possible for the command that jenkins runs to be something (or somethings) in the makefile? Would prefer not to hide that we run distributed GPU tests like this in the jenkinsfile.

Makefile
ravi-mosaicml: Both DDP and Deepspeed are distributed strategies, and both use the launcher script, so I think `dist` would work here?

composer/callbacks/memory_monitor.py
ravi-mosaicml: Ahhh ok, my bad. In this case, then I think it would make sense to return `None` and make any caller code handle missing memory stats, but will create a separate issue to clean this up later....

growlix:
FWIW I encounter the exact same issues as @dblalock when using the `mosaicml/pytorch_internal` image. Using the `mosaicml/pytorch` image fixes the ddp-related environment variable issues (though nothing else).

PR #390, opened 2022-02-05
Cleanup `conftest.py`

hanlint:
This PR should introduce no functional changes, but cleans up the `conftest.py` code for readability and verbosity. 

* `world_size` is now also a configurable parameter that defaults to either to `$WORLD_SIZE` envvar, or 1. This surfaces the world_size configuration that was previously hidden inside one of the helper functions.
* deselected items are now properly reported using the `pytest_deselected` hook

To verify, I ran all the test combinations and compared the number of tests between `dev` and this PR:

This PR (using `--collect-only` as those numbers are now accurate)
duration all, world_size 1, `1324/1487 tests collected (163 deselected)`
duration short, world_size 1, `1248/1487 tests collected (239 deselected)`
duration long, world_size 1, `76/1487 tests collected (1411 deselected)`
duration all, world_size 2, `23/1487 tests collected (1464 deselected)`
duration short, world_size 2, `1/1487 tests collected (1486 deselected)`
duration long, world_size 2, `22/1487 tests collected (1465 deselected)`

`dev`:
duration all, world_size 1, `9 failed, 1101 passed, 113 skipped, 82 xfailed, 19 xpassed` = 1324 tests
duration short, world_size 1, `1064 passed, 85 skipped, 80 xfailed, 19 xpassed`= 1248 tests
duration long, world_size 1, `9 failed, 37 passed, 28 skipped, 2 xfailed,`= 76 tests
duration all, world_size 2, `10 failed, 11 skipped, 2 xfailed` = 23 tests
duration short, world_size 2, `1 failed` = 1 tests
duration long, world_size 2, `9 failed, 11 skipped, 2 xfailed` = 22 tests


tests/conftest.py
```diff
@@ -33,82 +28,67 @@
...
+
+    conditions = [
+        lambda item: _get_world_size(item) == world_size,
+    ]
+
+    # separate tests by whether timeout is < or > threshold.
+    if duration == 'short':
+        conditions += [lambda item: _get_timeout(item) < timeout_threshold]
+    elif duration == 'long':
+        conditions += [lambda item: _get_timeout(item) > timeout_threshold]
```
ravi-mosaicml, line 73: Minor edge case: If the timeout is 0.0, then the test doesn't have a timeout. I think these tests should be classified as long, not short.

From the docs: Setting a timeout to 0 seconds disables the timeout, so if you have a global timeout set you can still disable the timeout by using the mark (https://pypi.org/project/pytest-timeout/)

tests/conftest.py
```diff
@@ -33,82 +28,67 @@
...
+    return item.get_closest_marker("timeout", default=_default).args[0]
+
+
+def _get_world_size(item: pytest.Item):
+    """Returns the world_size of a test, defaults to 1."""
+    _default = pytest.mark.world_size(1).mark
+    return item.get_closest_marker("world_size", default=_default).args[0]
+
+
+def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item]) -> None:
```
ravi-mosaicml, line 59: Thanks for cleaning this up -- much cleaner now!

tests/conftest.py
```diff
@@ -33,82 +28,67 @@
...
+    timeout_threshold = getattr(config, "_env_timeout", 2.0)
+    duration = config.getoption("duration")
+    world_size = config.getoption("world_size")
+
+    conditions = [
+        lambda item: _get_world_size(item) == world_size,
+    ]
+
+    # separate tests by whether timeout is < or > threshold.
+    if duration == 'short':
```
ravi-mosaicml, line 70: Can you validate that the `duration` is one of `short`, `long`, or `all`? I don't think the argparse validation would work if one were to set the duration int a config file.

tests/conftest.py
```diff
@@ -33,82 +28,67 @@
...
+                     type=int,
+                     choices=WORLD_SIZE_OPTIONS,
+                     help="""Number of devices. Filters the tests based on their
+                            requested world size. Defaults to 1, and can also
+                            be set by the WORLD_SIZE environment variable.""")
+
+
+def _get_timeout(item: pytest.Item):
+    """Returns the timeout of a test, defaults to 0."""
+    _default = pytest.mark.timeout(0).mark
```
ravi-mosaicml, line 49: If a timeout is not specified via a marker, then it should use the default. I believe that is available via `getattr(config, "_env_timeout", 2.0)`. A timeout of 0 means no timeout, not the default timeout.

ravi-mosaicml (commented):
Thanks for cleaning this up -- much simpler now! Overall looks fine, though left some comments throughout.

tests/conftest.py
```diff
@@ -33,82 +28,67 @@
...
-    items_to_remove: List[pytest.Item] = []
-    env_world_size = int(os.environ.get('WORLD_SIZE', '1'))
+def pytest_addoption(parser: pytest.Parser) -> None:
+    parser.addoption("--duration",
+                     default="short",
+                     choices=["short", "long", "all"],
+                     help="""Duration of tests, one of short, long, or all.
+                             Tests are short if their timeout < 2 seconds
+                             (configurable threshold). Default: short.""")
+    parser.addoption("--world-size",
```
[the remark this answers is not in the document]
hanlint: This option is just to filter the tests by world_size, and is useful for development purposes (e.g. no world_size envvar set, but would like to know what tests are collecting & running. Since we filter by world_size and duration, prefer to have both of them enumerated at the top rather than being hidden inside one of the functions. 

I clarified the 'help' and also added a guard for then `world_size != os.environ["WORLD_SIZE"]`

tests/conftest.py
hanlint: Thanks, I added logic to account for the timeout=0.0 special case.

tests/conftest.py
```diff
@@ -33,82 +31,83 @@
...
+
+
+def _get_world_size(item: pytest.Item):
+    """Returns the world_size of a test, defaults to 1."""
+    _default = pytest.mark.world_size(1).mark
+    return item.get_closest_marker("world_size", default=_default).args[0]
+
 
+def _validate_world_size(world_size: int):
+    if "WORLD_SIZE" in os.environ and int(os.environ["WORLD_SIZE"]) != world_size:
```
ravi-mosaicml, line 65: Can you add a check that errors if the CLI world size is greater than 1 but the os.environ world size is not set? In the error message, we can refer people to use the composer launch script to run pytest (or the makefile).

tests/conftest.py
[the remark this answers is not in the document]
ravi-mosaicml: I see. If this flag is for documentation purposes only (with the validation function below), I can live with that. I guess I would still prefer not to have it unless it serves a functional purpose. (For the collect-only case, I presume one can still filter tests by world size via an env variable -- e.g `WORLD_SIZE=2 pytest --collect-only`)?

tests/conftest.py
```diff
@@ -33,82 +31,83 @@
...
+                             Defaults to 1, and can also be set by the WORLD_SIZE
+                             environment variable. For world_size>1, please launch
+                             with the composer launcher.""")
+
+
+def _get_timeout(item: pytest.Item, default: float):
+    """Returns the timeout of a test, defaults to -1 """
+    _default = pytest.mark.timeout(default).mark
+    timeout = item.get_closest_marker("timeout", default=_default).args[0]
+    return 999 if timeout == 0 else timeout  # timeout(0) means no timeout restrictions
```
ravi-mosaicml, line 55:
```suggestion
    return float('inf') if timeout == 0 else timeout  # timeout(0) means no timeout restrictions
```

ravi-mosaicml (approved):

Issue #385, opened 2022-02-04
pytests should use `trainer.__init__()` not `hparams` where feasible


hanlint:
Tests are good code snippets for how to use our trainer. However, most of our tests of trainer functions (e.g. checkpointing, DDP) all rely on the hparams system, which makes it difficult for external users who are mostly non-yahp-based to grok our code. We should migrate all the tests to using `trainer=Trainer(..)` approaches.

mvpatel2000:
Hparams is gone. Congrats @hanlint

PR #391, opened 2022-02-06
Pytest Refactor Part 1

hanlint:
This PR is the first in a series to address #385 and also perform cleanup of our testing suite, including:
* Using constructor creation instead of `hparams` initialize whenever possible
* Deduplicate and reduce the number of global fixtures. Also move away from `dummy_` naming convention for many fixtures in favor of more informative names. 

During this transition, the new fixtures and common objects will be stored in `tests/fixtures/new_fixtures.py` and `tests/common.py` files, respectively. A new flag `include_deprecated_fixtures=True` in the conftest toggles off the old fixtures to expose tests that still need to be transitioned.

This first PR effects the above to the `tests/algorithms` test suite, and makes a few additional changes:
* adds a `algorithm_test_template.py` that includes a suggested structure for algorithm-based tests
* imposes a more consistent structure across the algorithm tests
* improves readability by grouping like-tests and associated fixtures under test classes
* uses minimallly functional dataloaders, to reduce test complexity
* renames `test_blurpool -> test_blurpool_layers.py` and `test_blurpool_algorithm.py` -> `test_blurpool.py` for consistency with the rest of the tests.


tests/algorithms/test_blurpool.py
```diff
@@ -1,96 +1,159 @@
...
-from tests.utils.trainer_fit import train_model
+from composer.algorithms import BlurPool, BlurPoolHparams
+from composer.algorithms.blurpool import apply_blurpool
+from composer.algorithms.blurpool.blurpool_layers import BlurConv2d, BlurMaxPool2d
+from composer.core import Event, State, surgery
+from composer.core.types import Logger, Tensors
+from composer.models import ComposerClassifier
+
+
+class ConvModel(torch.nn.Module):
```
hanlint, line 22: This model was specifically designed to test the blurpool algorithms, so scoping it to just this test, instead of making it globally available.

tests/algorithms/test_colout.py
```diff
@@ -107,134 +106,130 @@ def fake_image_batch(H: int, W: int, C: int) -> torch.Tensor:
...
 
-    assert torch.allclose(new_image_1, new_image_2)
+        assert torch.allclose(new_image_1, new_image_2)
 
 
-def test_reproducibility_batch(fake_image_batch: torch.Tensor, p_row: float, p_col: float):
-    """Test that batch augmentation is reproducible given the same seed."""
-    transform_1 = functools.partial(colout_batch, p_row=p_row, p_col=p_col)
-    transform_2 = functools.partial(colout_batch, p_row=p_row, p_col=p_col)
+class TestColOutFunctional:
```
hanlint, line 148: tests of the functional API and those of the algorithm class are now separately grouped for better readability.

tests/algorithms/test_stochastic_depth.py
```diff
@@ -1,254 +1,252 @@
...
+def test_stochastic_depth_hparams(stochastic_method: str, target_layer_name: str):
+    hparams = StochasticDepthHparams(
+        stochastic_method=stochastic_method,
+        target_layer_name=target_layer_name,
+    )
+    algorithm = hparams.initialize_object()
+    assert isinstance(algorithm, StochasticDepth)
+
+
+class TestSampleBernoulli:
```
hanlint, line 80: Stochastic Depths tests are now sorted and grouped by the function they test, and the fixtures are minimally scoped to the class if possible.

tests/common.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        super().__init__(module=net)
+
+        # bind these to class for access during
+        # surgery tests
+        self.conv1 = conv1
+        self.conv2 = conv2
+
+
+class RandomClassificationDataset(Dataset):
```
hanlint, line 85: We opt to use test suite's simpler datasets compared to `SyntheticDatasets` to make the tests simpler to read and debug.

ravi-mosaicml:
Thanks for taking a stab at this! Would it be possible to defer the major test refactors until next week? I think as we write docstrings we will be refactoring code (and potentially changes in test cases too)

hanlint:
@ravi-mosaicml let's discuss. if possible, i'd like to get this merged so that the docstrings/ test refactor are on this refactor.

The only drawback is that the refactor is mid-flight, and so we'll have both the deprecated_fixtures and new_fixtures coexisting. But I think that is a reasonable state for the codebase. Since this PR doesn't change anything functional about the tests, but mostly reorganizes them. Other drawbacks for merging we need to consider?

tests/algorithms/test_blurpool.py
```diff
@@ -1,96 +1,157 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Test the blurpool algorithm.
```
hanlint, line 3: The `diffs` are a bit wrong here because I renamed `test_blurpool_algorithm.py` -> `test_blurpool.py`, and `test_blurpool.py` -> `test_blurpool_layers.py`, but github is comparing the wrong files.

tests/fixtures/new_fixtures.py
```diff
@@ -0,0 +1,30 @@
...
+def minimal_state():
+    """Most minimally defined state possible.
+
+    Tests should configure the state for their specific needs.
+    """
+    return State(model=Mock(),
+                 precision=Precision.FP32,
+                 grad_accum=1,
+                 train_dataloader=Mock(__len__=lambda x: 100),
+                 evaluators=Mock(),
```
ravi-mosaicml, line 23:
```suggestion
                 evaluators=[],
```

tests/conftest.py
```diff
@@ -23,13 +23,21 @@
 # see composer.utils.reproducibility.configure_deterministic_mode
 os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
 
+# during the pytest refactor transition, this flag
+# indicates whether to include the deprecated fixtures
+include_deprecated_fixtures = True
```
ravi-mosaicml, line 28: I don't think having this as a flag is needed (let's include deprecated fixtures right now), but if we want to keep it as a flag, can it be moved to a CLI argument?

tests/common.py
ravi-mosaicml: Can you add this GH comment into the code?

ravi-mosaicml: Also, do we need new tests for the synthetic datasets (are any tests still using them?)

tests/fixtures/new_fixtures.py
```diff
@@ -0,0 +1,30 @@
...
+from unittest.mock import Mock
+
+import pytest
+
+from composer.core import Logger, State
+from composer.core.types import Precision
+
+
+@pytest.fixture
+def minimal_state():
```
ravi-mosaicml, line 14: I would ideally like the`minimal_state` be something that behaves as similarly to the real state, so that way tests only need to set what is needed, and not worry about replacing the mock parameters. Having partially defined mocks in global fixtures hide a pretty significant difference in the fixture construction, leading to harder-to-debug tests.

As an alternative of having just one `minimal_state`, could we have one state fixture for each model type? E.g. `simple_model_state` and `simple_conv_model_state`? Each state could then have a dataset of the appropriate type.
Then, could we update `State.__init__` to default `grad_accum`  to 1, precision to FP32, and evaluators to None or an empty collection (to match the Trainer's init signature)?

I think these changes combined would allow us to remove the mocks from the fixtures while ensuring that it would be something trainable.

ravi-mosaicml (approved):
I didn't review the algorithm test changes, as the diffs were too complicated for me to follow, but I trust that they are correct. For the common test fixtures, I would prefer to avoid using mocks there -- see comments. Otherwise, LGTM.

scripts/lint.sh
```diff
@@ -8,13 +8,13 @@ cd $(dirname $0)/..
...
 # Run isort (installed via setup.py)
-python -m isort . -c -v
+python -m isort . -c
 
 # Run yapf (installed via setup.py)
 python -m yapf -dr . # not using -p since that can lead to race conditions
 
 # Run docformatter
-python -m docformatter -rc --wrap-summaries 120 --wrap-descriptions 120 composer tests examples
+python -m docformatter -r --wrap-summaries 120 --wrap-descriptions 120 composer tests examples
```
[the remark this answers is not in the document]
hanlint: This is outdated -- `lint.sh` no longer exists :)

tests/fixtures/new_fixtures.py
hanlint: Great feedback, thanks! I did all of the above, but instead of having `simple_model_state` and `simple_conv_model_state`, just made `minimal_state == simple_model_state`. We can introduce a `conv_model_state` if the tests need that many times, but right now its more niche applications that require a little tinkering to the dataset configurations, etc.

tests/conftest.py
hanlint: It's more for internal dev purposes, so that we can switch the mode to `False`, and know what is broken once we deprecate the old fixtures. I'll make this a private variable, would that work?

PR #392, opened 2022-02-07
Move `world_size` guard to trainer

hanlint:
Previously, `prepare_ddp` and `initialize_dist` will be no-op calls if `world_size=1`. This makes the trainer code confusing because at first glance, we always initialize distributed and prepare ddp. We often also want to (for pytest) exercise the ddp path even with a world size of 1 since mostly we work on single GPU machines.  This is required for some future trainer tests.

Instead, this PR moves the `world_size==1` guard from those two separate functions to a single control point in the trainer code:
```
if dist.get_world_size() > 1:
    dist.initialize_dist(...)
```

And then the others just guard on dist being initialized.
```
if dist.is_initialized():
    model = prepare_ddp(model, ..)
```

The` prepare_ddp` and `initialize_dist` methods now always attempt their respective actions, and fail out if not configured properly. I am not an expert on the distributed code here, so cc: @jbloxham and @ravi-mosaicml for any potential pitfalls of this approach.


composer/trainer/trainer.py
```diff
@@ -8,6 +8,8 @@
 import logging
 import textwrap
 import warnings
+from collections.abc import Sized
+from re import I
```
ravi-mosaicml, line 12: unused import?

composer/trainer/trainer.py
```diff
@@ -285,27 +279,19 @@ def __init__(
                     To fix, please do not iterate over the dataloader before passing it into
                     the trainer."""))
 
-        if eval_subset_num_batches is not None:
```
ravi-mosaicml, line 288: Curious why this was removed?

composer/trainer/trainer.py
```diff
@@ -226,32 +226,33 @@ def __init__(
...
         find_unused_parameters = any(map(lambda x: x.find_unused_parameters, algorithms))
-
         self.find_unused_parameters = find_unused_parameters
 
+        # optionally initialize distributed backends
         if self.deepspeed_enabled:
             import deepspeed
             deepspeed.init_distributed()
-        else:
+        elif dist.get_world_size() > 1:
```
ravi-mosaicml, line 238: Should we have a code path for our trainer to support dist initialization when the world size is 1? 

I'm thinking perhaps we can add a helper method to `composer.utils.dist` called something like `should_use_dist` that returns True if  a) all dist env variables are set, or b) the world size > 1. I'm thinking this code path could be helpful during development if someone wants to see how distributing wrapping affects a model / algorithm without having to run a multi-process test.

ravi-mosaicml (approved):
LGTM, agree this check belongs in the trainer itself. Thanks!

composer/trainer/trainer.py
ravi-mosaicml: Discovered that we need this code path in #389 for deepspeed.

composer/trainer/trainer.py
hanlint: It's a pretty verbose check for a narrow issue. I think we can clarify in the docs instead of throw a warning.

Issue #362, opened 2022-02-03
Fix `CutMix`, `AugMix`, and `LabelSmoothing` on pytorch 1.9


ravi-mosaicml:
Example build that is failing: https://ci.int.mosaicml.com/job/scratch/job/command/2113/testReport/

Verify this is fixed by removing the following importskip: https://github.com/mosaicml/composer/blob/f14e6e59337f73b04ecd312b55b70755b63a6d45/tests/trainer/test_init_idempotency.py#L68-L71

This may be an issue with the models rather than with the algorithm itself. But, something did change between torch 1.9 and torch 1.10

ravi-mosaicml:
Closing the issue as PyTorch 1.9 is two releases old, and we will likely drop support for it soon.

PR #393, opened 2022-02-07
Trainer test refactor (pytest refactor phase 2)

hanlint:
This Trainer test refactor is also ready for merging. Functionally preserving from the previous tests, with the addition of a new test to make sure that we are passing the augmented batch to the `model.forward()`, a bug we've had repeatedly.

This PR sits on top of #391


pyproject.toml
```diff
@@ -54,7 +54,7 @@ addopts = "--strict-markers -m 'not daily_regression and not weekly_regression a
 
 # by default, tests should be fast.
 # for slower tests, use @pytest.mark.timeout with a higher timeout (specified in seconds)
-timeout = 2
+timeout = 5
```
ravi-mosaicml, line 57: Curious why the timeout increase? Are tests failing with a default timeout of 2?

tests/fixtures/new_fixtures.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.fixture
+def empty_logger(minimal_state: State) -> Logger:
+    """Logger without any output configured."""
+    return Logger(state=minimal_state, backends=[])
+
+
+@pytest.fixture
+def init_process_group(monkeypatch):
```
ravi-mosaicml, line 38: I beleive that we already configure dist globally -- see https://github.com/mosaicml/composer/blob/85b60cddfb722e0ecda93464afecf704673dc254/tests/fixtures/distributed_fixtures.py#L17. Since the fixture has autouse, it is called before any test runs. Curious which tests needed this?

tests/common.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    # convert cpu-fp32 and gpu-fp32 to cpu, gpu
+    if not precision and any(['-' in arg for arg in args]):
+        raise ValueError('-fp32 and -amp tags must be removed if precision=False')
+    args = [arg.replace('-fp32', '') for arg in args]
+
+    # use lambdas to construct DeviceGPU only when gpu requested
+    if precision:
+        devices = {
+            'cpu': lambda: pytest.param('cpu', Precision.FP32, id="cpu-fp32"),
```
ravi-mosaicml, line 37: Are lambdas needed here? I don't think it's taking any args on line 49.

tests/common.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Returns the parameter "device", or if precision=True,
+    also returns the parameter "precision".
+    """
+    # convert cpu-fp32 and gpu-fp32 to cpu, gpu
+    if not precision and any(['-' in arg for arg in args]):
+        raise ValueError('-fp32 and -amp tags must be removed if precision=False')
+    args = [arg.replace('-fp32', '') for arg in args]
+
+    # use lambdas to construct DeviceGPU only when gpu requested
```
ravi-mosaicml, line 34: Can you elaborate? I don't think DeviceGPU needs to get registered for the decorator?

tests/conftest.py
```diff
@@ -23,13 +23,21 @@
 # see composer.utils.reproducibility.configure_deterministic_mode
 os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
 
+# during the pytest refactor transition, this flag
+# indicates whether to include the deprecated fixtures
+include_deprecated_fixtures = True
```
ravi-mosaicml, line 28: Same comment as in the other pr

tests/fixtures/new_fixtures.py
```diff
@@ -0,0 +1,54 @@
...
+import pytest
+import torch.distributed as dist
+
+from composer.cli.launcher import get_free_tcp_port
+from composer.core import Logger, State
+from composer.core.types import Precision
+
+
+@pytest.fixture
+def minimal_state():
```
ravi-mosaicml, line 18: Same comment as in the other pr

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+            'eval_dataloader': DataLoader(dataset=RandomClassificationDataset()),
+            'max_duration': '2ep',
+        }
+
+    def test_init(self, config):
+        trainer = Trainer(**config)
+        assert isinstance(trainer, Trainer)
+
+    def test_model_ddp_wrapped(self, config, init_process_group):
+        # dist initialized first with init_process_group fixture
```
ravi-mosaicml, line 41: This fixture should not be required since dist is configured globally here: https://github.com/mosaicml/composer/blob/06ccbb2830581debe1c06451e6c3a277646e2ea2/tests/fixtures/distributed_fixtures.py#L17

Is the test erroring without it?

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+        with pytest.raises(AssertionError):
+            self.assert_models_equal(trainer.state.model, self.reference_model)
+
+
+class AssertDataAugmented(Callback):
+    """Helper callback that asserts test whether the augmented batch was passed to the model during the forward pass.
+    The original batch is passed through the model and we assert that the outputs are not the same.
+
+    Assumes batch_size=1, and no gradient accumulation.
```
ravi-mosaicml, line 177: Could you add these as asserts in the `init()` event?

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+    """
+
+    def __init__(self, dataset):
+        self.dataset = dataset
+
+    def after_forward(self, state, logger):
+        batch_idx = state.timer.batch_in_epoch.value
+        batch_size = state.train_dataloader.batch_size
```
ravi-mosaicml, line 185: state.batch_num_samples
```suggestion
        batch_size = state.batch_num_samples
```

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+class AssertDataAugmented(Callback):
+    """Helper callback that asserts test whether the augmented batch was passed to the model during the forward pass.
+    The original batch is passed through the model and we assert that the outputs are not the same.
+
+    Assumes batch_size=1, and no gradient accumulation.
+    """
+
+    def __init__(self, dataset):
+        self.dataset = dataset
```
ravi-mosaicml, line 181: I'm not convinced that this callback works as intended, as dataset transforms are run before `__getitem__` is returned, so I think the test would always pass? I'm not sure if it would work to deepcopy the dataset (I presume so?) This would assume all (random) data has already been generated and that dataloader shuffling is disabled. Could be worth adding asserts to ensure these conditions.

```suggestion
        self.dataset = copy.deepcopy(dataset)
```

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+        name, hparams = request.param
+        skip_list = {
+            'swa': 'SWA not compatible with composed schedulers.',
+            'alibi': 'Not compatible with conv model.',
+            'seq_length_warmup': 'Not compatible with conv model.',
+            'randaugment': 'Requires PIL dataset to test.',
+            'augmix': 'Required PIL dataset to test.',
+            'stochastic_depth': 'Only applies to ResNets.',
+            'no_op_model': 'Not compatible with this model.'
```
ravi-mosaicml, line 273: Curious why the no-op model doesn't work?

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+    @pytest.fixture(params=algorithms_registry.items(), ids=algorithms_registry.keys())
+    def algorithm(self, request):
+
+        name, hparams = request.param
+        skip_list = {
+            'swa': 'SWA not compatible with composed schedulers.',
+            'alibi': 'Not compatible with conv model.',
+            'seq_length_warmup': 'Not compatible with conv model.',
+            'randaugment': 'Requires PIL dataset to test.',
```
ravi-mosaicml, line 270: We have the PIL dataset now, so these one should now work!

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+            'augmix': 'Required PIL dataset to test.',
+            'stochastic_depth': 'Only applies to ResNets.',
+            'no_op_model': 'Not compatible with this model.'
+        }
+
+        # skip any tests incompatible with this config
+        if name in skip_list:
+            pytest.skip(skip_list[name])
+        elif name in ('cutmix, mixup, label_smoothing'):
+            pytest.importorskip("torch", minversion="1.10", reason="Pytorch 1.10 required.")
```
ravi-mosaicml, line 280: Can you add more context / perhaps link to https://github.com/mosaicml/composer/issues/362 in a comment to why they require 1.10?

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+    def test_loggers_multiple_calls(self, config, logger):
+        if isinstance(logger, FileLogger):
+            pytest.xfail("Known idempotency issue.")
+        config['loggers'] = [logger]
+        trainer = Trainer(**config)
+        self._test_multiple_fits(trainer)
+
+    def _test_multiple_fits(self, trainer):
+        trainer.fit()
```
ravi-mosaicml, line 373: I think the private method for the first call to fit should be used so callbacks aren't closed.
(Once we officially support multiple calls to fit, then we won't need to use the private method)

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+
+    @pytest.fixture(params=callback_registry.items(), ids=callback_registry.keys())
+    def callback(self, request, tmpdir, monkeypatch):
+        name, hparams = request.param
+
+        if name == 'benchmarker':
+            pytest.skip('Deprecated callback.')
+
+        # create callback
+        if name == 'run_directory_uploader':
```
ravi-mosaicml, line 298: For wandb, can you ensure that it is given the `disabled` argument? See https://github.com/mosaicml/composer/blob/b40ef9f3f984f9c6cb89854eaf86a5b41b48a303/tests/test_logger.py#L126

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+    def test_callbacks_multiple_calls(self, config, callback):
+        if isinstance(callback, RunDirectoryUploader):
+            pytest.xfail("Known idempotency issue.")
+        config['callbacks'] = [callback]
+        trainer = Trainer(**config)
+        self._test_multiple_fits(trainer)
+
+    def test_loggers_multiple_calls(self, config, logger):
+        if isinstance(logger, FileLogger):
+            pytest.xfail("Known idempotency issue.")
```
ravi-mosaicml, line 367: It's not an issue with the logger, but rather, that the trainer automatically closes all callbacks after the first call to .fit. This can be fixed by using the private method `trainer._train_loop()` on line 373. Would rather have the file logger tested and call a private method than expect it to fail.

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,375 @@
...
+        _ = next(dataloader.__iter__())
+
+        config['train_dataloader'] = dataloader
+        with pytest.raises(ValueError, match="active iterator"):
+            Trainer(**config)
+
+
+@world_size(1, 2)
+@device('cpu', 'gpu', 'gpu-amp', precision=True)
+class TestTrainerEquivalence():
```
ravi-mosaicml, line 92: This is so much easier to follow -- thanks!

ravi-mosaicml (commented):
Overall LGTM; but didn't review the algorithm tests since the diffs were large.

The trainer tests are much easier to follow; thanks!

See comments: re helper fixtures.

tests/fixtures/new_fixtures.py
hanlint: Yeah I want to eventually move toward removing this `autouse`, since the Trainer itself already initializes the `dist`. For more isolated tests, better to specifically request the fixture when needed, otherwise we can have subtle bugs.

If this hypothesis doesn't work as we port more tests over, we can move this fixture over.

tests/common.py
hanlint: Ah yes that's a residual, will cleanup, thanks!

tests/common.py
```diff
@@ -0,0 +1,199 @@
...
+from torchvision.datasets import VisionDataset
+
+from composer.core.precision import Precision
+from composer.models import ComposerClassifier
+
+# syntactic sugar below
+# decorators for common parameterizations and marks
+
+
+def device(*args, precision=False):
```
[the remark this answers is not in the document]
hanlint: I am hesitant here because that would hide the settings being tested to the test builder/debugger -- they would need to see the decorators to understand what is going on, which defeats the purpose of the syntactic sugar, which is to reduce syntax but keep needed information.

I agree that it is doing too much. I'll file an issue to clean this up, perhaps split into two decorators:
```
@device('gpu', 'cpu')
```
```
@device_precision('gpu,fp32', 'gpu,amp')
```

The decorator names should also convey the fixture names (e.g. device, and device_precision). Would that work?

tests/trainer/test_trainer.py
hanlint: If we aim to support multiple calls to fit, better for the test to support exactly code path and `xfail` to highlight specific issues. Otherwise, users may read this test and think that calling the private method is the recommended approach for multiple calls.

tests/trainer/test_trainer.py
hanlint: See comment above. Since we already test the `FileLogger` within the single call to fit, and multiple calls to fit is not an officially supported feature, prefer to keep the test code to be the expected public usage in the future.

pyproject.toml
hanlint: I might have done that accidentally -- reverting.

tests/trainer/test_trainer.py
hanlint: That fixture will be deprecated, so I am testing without that fixture active, and introducing a much more localized one. We can migrate it over if its needed as we continue the refactor.

The issue with the global `dist` configuration is that it inits process groups outside of our `Trainer`. Ideally we want a clean environment for the trainer tests to run in. e.g. if the Trainer's own dist init functions are broken, that won't be caught since this global fixture is working.

tests/trainer/test_trainer.py
hanlint: Ah, so the algorithm being tested is `CutOut`, which runs on `AFTER_DATALOADER`.

For example, see the control test `test_data_not_augmented`, which fails when no data augmentation is being applied.

tests/trainer/test_trainer.py
hanlint: Yes, if its OK that takes a bit of work to enable. Maybe defer from here, as I am just meeting functional equivalence.

tests/common.py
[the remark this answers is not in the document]
ravi-mosaicml: Agreed, that makes sense. Thinking about it more, I'm wondering if we need to hide the underlying parameterization logic, since that makes an assumption about what the field name is called. Would something like this instead work?

# common.py
```python3
device_cpu = pytest.param(DeviceCPU())
device_gpu = pytest.param(DeviceGPU(), mark=pytest.mark.gpu)
```
Then, the test case would do something like:
```python3
@pytest.mark.paramtereize(device, [common.device_cpu, common.device_gpu])
def test_something(device: Device):
    ...
```

or for device/precision:
```python3
@pytest.mark.paramtereize("device,precision", [(common.device_gpu, Precision.FP32), (common.device_gpu, Precision.AMP])
def test_something_else(device: Device, precision: Precision):
    ...
```

tests/trainer/test_trainer.py
ravi-mosaicml: OK, makes sense.

tests/fixtures/new_fixtures.py
ravi-mosaicml: Hmm ok, I am now forgetting why we are configuring dist globally. It might not be needed any longer. Testing right now to see if we can rip out the autoused fixture...

tests/trainer/test_trainer.py
```diff
@@ -1,247 +1,380 @@
...
+        trainer = Trainer(**config)
+        assert isinstance(trainer, Trainer)
+
+    def test_model_ddp_wrapped(self, config, init_process_group):
+        # dist initialized first with init_process_group fixture
+        trainer = Trainer(**config)
+        assert isinstance(trainer.state.model, DistributedDataParallel)
+
+    def test_model_ddp_not_wrapped(self, config):
+        torch.distributed.destroy_process_group()
```
ravi-mosaicml, line 46: Ah, I missed this the first time, but I've run into issues before where destroying the process group was buggy. I don't think it always freed the port, but am not quite sure. I guess if it's reliably working then that is fine.

hanlint: Yeah actually I tried to put in a `autouse=True` to always destroy process groups at the end of a test, but it didn't always free the port, and since the port is encoded as an `envvar` by the composer launch process, it never finds a new free one.

tests/trainer/test_trainer.py
ravi-mosaicml: Could we replace this test and the one below it with one test like the following?

The underlying reason is that I remember there being issues with destroying process groups. (Never traced down the underlying issue, so perhaps it is no longer a thing anymore.) Assuming that is an issue, we don't destory process groups after a test. The first trainer test is the one that initializes the process group, and every other trainer test uses that same progress group (and doesn't attempt to re-initialize it). (Technically, we initialize this with the configure_dist fixture before any trainer tests run, but I don't think that's no longer necessary). Because we don't destroy process groups, we have to run pytest multiple times -- once without distributed and once with rank-one distributed.

Instead, could we condition this test (and the one that follows) on which pytest environment it is in with something like the following?

```
    def test_model_ddp_wrapped(self, config):
        # dist initialized first with init_process_group fixture
        trainer = Trainer(**config)
        should_be_ddp_wrapped = "WORLD_SIZE" in os.environ  # if WORLD_SIZE and all other dist environment variables are set, the the trainer will configure dist. Otherwise, it shouldn't, unless if deepspeed is being used.
        assert isinstance(trainer.state.model, DistributedDataParallel) == should_be_ddp_wrapped
```

tests/fixtures/new_fixtures.py
ravi-mosaicml: Just tested it and I think we can rip out `distributed_fixtures.py`. I replaced it with something like the following, and this works for me:

```python3

import pytest

@pytest.fixture(autouse=True)
def wait_for_all_procs():
    yield
    dist.barrier()
```

tests/common.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def decorator(test):
+        if not parameters:
+            return test
+        return pytest.mark.parametrize(name, parameters)(test)
+
+    return decorator
+
+
+def world_size(*args):
```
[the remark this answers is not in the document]
ravi-mosaicml: Similar comment to the above -- perhaps add syntax sugar for world size like this?

```python3
# common.py
world_size_1 = pytest.param(1)
world_size_2 = pytest.param(2, mark=pytest.mark.world_size(2))
```

tests/trainer/test_trainer.py
ravi-mosaicml: Ah kk, I missed that. Could we rename it to `AssertDataAugmentedAfterDataloader` then?

ravi-mosaicml (approved):

Issue #409, opened 2022-02-07
Reasonable Defaults: Add Reasonable Defaults to DWD


moinnadeem:
We should reflect that we copy PyTorch's SGD defaults

PR #410, opened 2022-02-07
Clarify DWD defaults

abhi-mosaic:
Addresses #409


moinnadeem:
Thanks for getting to this so quickly! 

Two thoughts:
1. The current approach still requires the user to Google the current defaults used in PyTorch. I would strongly prefer to just have defaults in our docstring, rather than pass that indirection to our users.
2. If we MUST pass that indirection to our users, we should hyperlink the SGD / AdamW:

`Argument defaults are copied from `torch.optim.SGD`.` in order to make it convenient. 

I have a reasonable preference for #1, but would be willing to hear an argument for #2.

abhi-mosaic:
I'm a bit confused, doesn't our docstring already include the defaults, as well as hyperlinks to torch's SGD and AdamW?

hanlint:
I think what is confusing @moinnadeem is that the first method is not `__init__`, where the defaults are.

hanlint:
Updating and merging after tests pass.. @moinnadeem feel free to reopen if the change, and the comments don't address the issue. But the DWD do seem documented in both the code (`__init__`) and the docstrings.

Issue #300, opened 2022-01-29
Surgery and StochasticDepth Doc Updates


ajaysaini725:
- `surgery.replace_module_classes` docs should explicitly state that model weights are not copied over. Once #297 is implemented the statement can be amended to include that model weights are copied over in the case of a 1-1 mapping
-  Until #253 is implemented, `StochasticDepth` docs need to explicitly state that it only works on Composer's ResNet model.

hanlint:
As #297 is not in `v0.4`, moving this out of `v0.4`. We'll be sure to add documentation to StochasticDepth around the models it supports.

ajaysaini725:
Addressed by #413

PR #413, opened 2022-02-07
Address Doc Fixes for Surgery and StochasticDepth

ajaysaini725:
Addresses documentation fixes in #300
- `replace_module_classes` needs to specify tensor values are not copied
- `StochasticDepth` needs to specify that it only works on `ResNet` models
- Small fix to remove Functional API section from trainer docs


ravi-mosaicml (approved):
LGTM

PR #417, opened 2022-02-08
Remove the `walkthrough` section of the docs; replace with module-level docstrings

### LINKED ISSUE #411: Move Composer Walkthrough docs into docstrings
Discussed offline with Ajay and Jamie, but the current "Composer Walkthrough" docs should be moved into module-level docstrings.

### LINKED ISSUE #425: Fix the docstring autosummary tables to support private members imported and re-exported by public functions
The autosummary tables (built using https://github.com/mosaicml/composer/blob/fef33aa9fdcde3873597a5ab108e9d26fca433ca/docs/source/conf.py#L157) will not work properly for private members that are imported and re-exported by a public module.
To fix this, instead of using `{item.__module__}.{item.__name__}`, it should first be checked whether `__module__` is private. If so, then each parent module should be checked (recursively), starting from the innermost to the outermost, until the first public module that contains an export for `item` is found. That module's `__module__` should be used instead of `item.__module__`. A public module should exist, as otherwise the item should not appear in an autosummary (there is an edge case if an item is not re-imported by a module in the hierarchy, but that should never happen. OK to error in that case)

ravi-mosaicml:
Replaced the Walkthrough sections of the docs with module-level docstrings; closes #411.

Fixed #425 in this PR as well.


composer/algorithms/__init__.py
```diff
@@ -1,5 +1,45 @@
...
+* :meth:`Algorithm.apply`: performs an in-place modification of the given
+  :class:`State`
+
+For example, a simple algorithm that shortens training:
+
+.. code-block:: python
+
+    from composer import Algorithm, State, Event, Logger
+
+    class ShortenTraining(Algorithm):
```
ajaysaini725, line 29: Lol

ajaysaini725, line 29: Lol

composer/callbacks/__init__.py
```diff
@@ -1,13 +1,63 @@
...
 from composer.callbacks.callback_hparams import SpeedMonitorHparams as SpeedMonitorHparams
-from composer.callbacks.lr_monitor import LRMonitor as LRMonitor
-from composer.callbacks.run_directory_uploader import RunDirectoryUploader as RunDirectoryUploader
-from composer.callbacks.speed_monitor import SpeedMonitor as SpeedMonitor
+from composer.callbacks.grad_monitor import GradMonitor
+from composer.callbacks.lr_monitor import LRMonitor
+from composer.callbacks.run_directory_uploader import RunDirectoryUploader
+from composer.callbacks.speed_monitor import SpeedMonitor
+
+__all__ = [
```
ajaysaini725, line 57: Is the reason for not including this in `Algorithms` that the list would be too long so instead it's a table?

composer/core/state.py
```diff
@@ -67,13 +70,26 @@
 
 
 class State(Serializable):
-    """The class used to store the state of the trainer.
+    """The state of the trainer.
```
ajaysaini725, line 73: The top of the state docs look like the below - is that right? Seems like those tables shouldn't be there
Screen Shot 2022-02-07 at 5 44 40 PM
:

composer/core/types.py
```diff
@@ -2,7 +2,27 @@
 
 """Reference for common types used throughout our library.
 
-See :doc:`/core/types` for documentation.
+TODO: This attributes list is incomplete.
```
ajaysaini725, line 5: Is this `TODO` intentional?

composer/functional/__init__.py
```diff
@@ -1,15 +1,18 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-"""Functional API.
+"""Functional API for apply algorithms.
```
ajaysaini725, line 3:
```suggestion
"""Functional API for applying algorithms.
```

composer/trainer/__init__.py
```diff
@@ -1,5 +1,97 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Train models!
+
+The trainer supports models with :class:`Algorithm` instances.
```
ajaysaini725, line 5:
```suggestion
The trainer supports models with :class:`ComposerModel` instances.
```

composer/trainer/__init__.py
```diff
@@ -1,5 +1,97 @@
...
+.. code-block:: python
+
+    from composer.trainer import TrainerHparamms
+
+    # Create a trainer from hparams and train train the model
+    trainer = hparams.initialize_object()
+    trainer.fit()
+
+
+Trainer Hparams
```
ajaysaini725, line 70: Should prob delete the Hparams example

composer/callbacks/__init__.py
ravi-mosaicml: Yea, haven't gotten to it for algorithms yet. that'll be seperate.

composer/core/state.py
ravi-mosaicml: These tables will disappear once `__all__` is properly defined on state.

composer/callbacks/__init__.py
```diff
@@ -1,13 +1,63 @@
...
+    .. code-block:: python
+
+        from composer import Callback
+
+        class MyCallback(Callback)
+
+            def _run_event(self, event: Event, state: State, logger: Logger):
+                if event == Event.EPOCH_START:
+                    print(f'Epoch {state.epoch}/{state.max_epochs}')
+"""
```
Averylamp, line 43: Discussed offline with @ravi-mosaicml, this call should likely be a public method

Averylamp (approved):
lgtm

composer/callbacks/__init__.py
Averylamp: https://github.com/mosaicml/composer/issues/430

PR #418, opened 2022-02-08
Un-deprecate `steps_per_epoch`

jbloxham:
Ultimately, it's difficult to implement stepwise schedulers without being able to convert between epochs and batches. We'd discussed having a fixed samples/epoch or tokens/epoch, but this is insufficient to handle warmup. Here's an example of how this can go wrong:

Total duration: 1ep
Cosine annealing scheduler, step-wise enabled, warmup of 100ba

At say, 101ba, need to calculate `(1dur - 100ba) / (101ba - 100ba)` for the scheduler calculation, but we lack a way to convert between duration units and batch units.*

At the end of the day, the reason for deprecating `steps_per_epoch` was that we anticipate wanting to support dynamic batch sizes. However, this is not on the roadmap. I believe we should punt this problem down the road, and demand that we have a timing abstraction that can deal with dynamically changing timing ratios before we abandon convenient static ratios like `steps_per_epoch`.

*One potential way around this is to take note of the timer state when the warmup ends, but this means that the scheduler is no longer stateless.


abhi-mosaic:
I agree with the choice for now. Basically we are requiring epoch length to be known at launch time, and so `samples_per_epoch`, `tokens_per_epoch`, `batches_per_epoch` should all be known. 

In the future, we can attempt to change this but it will probably be some kind of branch-based thing like "you don't have to specify an epoch length as long as your `max_duration` is not in units `EPOCH`.

ravi-mosaicml:
Good points here. Going to offer a few alternatives; curious if these would work:

1. Would it work to require that the warmup period be in the same unit as max duration (or be a percentage of it)? I.e. if the max duration is 10ep, then warmup can be either `0.1dur` or `1ep` or the like. (Note: if we statically convert epochs to tokens/samples, then the warmup could also be specified in one of those units)
2. Alternatively, do we want to add a `Period` abstraction to the timer? Then you could have a warmup, training, and cooldown period. Each period could have its own max duration in its own unit.

abhi-mosaic:
> Good points here. Going to offer a few alternatives; curious if these would work:
> 
> 1. Would it work to require that the warmup period be in the same unit as max duration (or be a percentage of it)? I.e. if the max duration is 10ep, then warmup can be either `0.1dur` or `1ep` or the like. (Note: if we statically convert epochs to tokens/samples, then the warmup could also be specified in one of those units)
> 2. Alternatively, do we want to add a `Period` abstraction to the timer? Then you could have a warmup, training, and cooldown period. Each period could have its own max duration in its own unit.

Re. 1, we would still need a epoch->batch conversion in order to handle the post-warmup scheduler (think a cosine decay with `dur:0.9`, how do we step through that with batch granualrity, without knowing the equivalent number of batches?)

Re. 2, I would like to have the concept of `Period`s :) but not sure if it's in scope for release 0.4.

ravi-mosaicml:
> > Good points here. Going to offer a few alternatives; curious if these would work:
> > 
> > 1. Would it work to require that the warmup period be in the same unit as max duration (or be a percentage of it)? I.e. if the max duration is 10ep, then warmup can be either `0.1dur` or `1ep` or the like. (Note: if we statically convert epochs to tokens/samples, then the warmup could also be specified in one of those units)
> > 2. Alternatively, do we want to add a `Period` abstraction to the timer? Then you could have a warmup, training, and cooldown period. Each period could have its own max duration in its own unit.
> 
> Re. 1, we would still need a epoch->batch conversion in order to handle the post-warmup scheduler (think a cosine decay with `dur:0.9`, how do we step through that with batch granualrity, without knowing the equivalent number of batches?)
> 
> Re. 2, I would like to have the concept of `Period`s :) but not sure if it's in scope for release 0.4.

For 1., this would require that, if max_duration is specified in epochs, we convert it into samples or tokens using the DataSpec properties. Once warmup finishes, we can calculate the remaining training length by looking at the (converted) max_duration - state.timer.samples (or state.timer.tokens).

For 2., makes sense :)

jbloxham:
> > Good points here. Going to offer a few alternatives; curious if these would work:
> > 
> > 1. Would it work to require that the warmup period be in the same unit as max duration (or be a percentage of it)? I.e. if the max duration is 10ep, then warmup can be either `0.1dur` or `1ep` or the like. (Note: if we statically convert epochs to tokens/samples, then the warmup could also be specified in one of those units)
> > 2. Alternatively, do we want to add a `Period` abstraction to the timer? Then you could have a warmup, training, and cooldown period. Each period could have its own max duration in its own unit.
> 
> Re. 1, we would still need a epoch->batch conversion in order to handle the post-warmup scheduler (think a cosine decay with `dur:0.9`, how do we step through that with batch granualrity, without knowing the equivalent number of batches?)
> 
> Re. 2, I would like to have the concept of `Period`s :) but not sure if it's in scope for release 0.4.

1 would not be a problem in terms of implementation - the scaled elapsed duration calculation could be something like `(1dur - 0.1dur) / ({X}sa - 0.1dur)`, and the duration <=> sample conversion is possible. I'm not a huge fan of it from a user  intuitiveness standpoint though.

hanlint (approved):

Issue #416, opened 2022-02-08
`Trainer` init should not require `eval_dataloader`


siriuslee:
The `eval_dataloader` argument to `Trainer` is required but can be set to `None`. Make it optional so it doesn't have to be provided if not used.

PR #419, opened 2022-02-08
Trainer Docs + Param ordering + Alibi Export

ajaysaini725:
Random fixes including:
- Making `max_duration` before `eval_dataloader` in the `Trainer` parameter list since `max_duration` is required and `eval_dataloader` is optional
- Updating `Trainer` docstrings
- Alibi was not exported in the algorithms `__init__` - added this

Addresses #416


hanlint:
Could you possible take care of #416 as well, and set = None. And just to be more readable, maybe actually do: `model`, `max_duration`, `train_dataloader`, `eval_dataloader=None`?

ravi-mosaicml (approved):
LGTM

Issue #256, opened 2022-01-20
Add wall clock time tracking into the Timer


ravi-mosaicml:
The `composer.core.Timestamp` should also track wall clock training time.

1. Add an `elapsed_time: datetime.timedelta` field to the `Timestamp` class.
2. `timestamp.to_next_batch` and `timestamp.to_next_epoch` should also take in an `elapsed_time: timedelta = timedelta(seconds=0)` argument. This argument should override (rather than increment) the current `elapsed_time` value.
3. The trainer should start tracking time when `.fit()` is called. Note that if `reset_time` is False, then the start time should be offset by the current `state.timestamp.elapsed_time`, so this previous time information will not be discard.
4. Then, when calling `to_next_batch` and `to_next_epoch` are invoked, the trainer should `dist.broadcast` the elapsed time from the rank zero process to all processes, and that reduced value should be passed into `to_next_batch` and `to_next_epoch`.

_Originally posted by @ravi-mosaicml in https://github.com/mosaicml/composer/issues/146#issuecomment-993016660_

PR #422, opened 2022-02-08
Refactor Profiler init()

### LINKED ISSUE #306: Remove `ProfilerHparams` from `Trainer.__init__`
For ease of use, the init signature should never require someone to import a custom class -- must be python/pytorch native. The only exception is `model`, which has to inherit from `MosaicModel`.

bandish-shah:
- [x] Resolves #306 
- [x] Flattened profilers into profiler object, made torch profiler optional
- [ ] Docstring updates


composer/profiler/profiler.py
```diff
@@ -167,20 +152,45 @@ class Profiler:
...
-        state: State,
-        event_handlers: Sequence[ProfilerEventHandler],
-        skip_first: int = 0,
-        wait: int = 0,
-        warmup: int = 1,
-        active: int = 4,
-        repeat: int = 1,
-        merged_trace_file: str = "merged_profiler_trace.json",
-    ) -> None:
+    def __init__(self,
```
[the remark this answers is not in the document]
ravi-mosaicml: In addition, if the trainer init signature to take the list of profiler arguments, could the Trainer's `__init__` construct the profiling callbacks (e.g. SystemProfiler, TorchProfiler, DataLoaderProfiler) from these parameters and append those callbacks directly? Then, I don't think the `Profiler` class doesn't need to take any of the parameters after `merged_trace_file`.

It feels a bit weird to me that the profiler is constructing these callbacks only for the trainer to read them.

composer/profiler/profiler.py
```diff
@@ -214,11 +224,11 @@ def get_action(self, batch_idx: int) -> ProfilerAction:
         return ProfilerAction.ACTIVE
 
     @property
-    def event_handlers(self):
+    def handlers(self):
```
ravi-mosaicml, line 227: If the trainer is directly constructing the profiler callbacks, then I think this can be kept as `event_handlers`.

composer/profiler/profiler.py
```diff
@@ -214,11 +224,11 @@ def get_action(self, batch_idx: int) -> ProfilerAction:
...
         """Profiler event handlers."""
-        return self._event_handlers
+        return self._event_handlers + self._profilers
 
-    def merge_traces(self):
+    def merge_traces(self, callbacks: Callback):
```
ravi-mosaicml, line 231: If `state` is being passed in on `__init__`, then the callbacks need not be passed in here.

composer/profiler/profiler.py
```diff
@@ -252,6 +262,7 @@ def merge_traces(self):
     def marker(
         self,
         name: str,
+        state: State,
```
ravi-mosaicml, line 265: Commented on this above, but could we not require the state to be passed in on the `marker` API? I presume this would be one of the only user-facing APIs of the profiler and would like to keep the required arguments as minimal as possible.

composer/profiler/profiler.py
```diff
@@ -450,9 +464,9 @@ def start(self) -> None:
             raise RuntimeError(
                 f"Attempted to start profiler event {self.name}; however, this marker is already started")
 
-        epoch = self.profiler.state.epoch
-        step = self.profiler.state.step
-        batch_idx = self.profiler.state.batch_idx
+        epoch = self.state.epoch
```
ravi-mosaicml, line 467: Following up on the comment to change the ProfilerEventHandler to take a `Timestamp` object instead of `epoch` and `step`, can these lines be replaced with a call to `state.timer.get_timestamp()`?

composer/profiler/profiler.py
```diff
@@ -23,20 +23,6 @@
...
-    def initialize_object(self) -> ProfilerEventHandler:
-        """Constructs and returns an instance of the :class:`ProfilerEventHandler`.
-
-        Returns:
-            ProfilerEventHandler: The event handler.
-        """
-        pass
-
-
 class ProfilerEventHandler(Callback, abc.ABC):
```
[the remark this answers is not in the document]
ravi-mosaicml: In addition, can we update the signatures for `process_duration_event`, `process_instant_event`, and `process_counter_event` to take `Timestamp` objects instead of `epoch` and `step`?

https://github.com/mosaicml/composer/blob/fef33aa9fdcde3873597a5ab108e9d26fca433ca/composer/core/time.py#L496)

composer/profiler/profiler.py
```diff
@@ -276,6 +287,7 @@ def marker(
             self._names_to_markers[name] = Marker(
                 self,
                 name,
+                state=state,
```
ravi-mosaicml, line 290: Ditto

composer/profiler/profiler.py
```diff
@@ -450,9 +464,9 @@ def start(self) -> None:
             raise RuntimeError(
                 f"Attempted to start profiler event {self.name}; however, this marker is already started")
 
-        epoch = self.profiler.state.epoch
-        step = self.profiler.state.step
-        batch_idx = self.profiler.state.batch_idx
+        epoch = self.state.epoch
+        step = self.state.step
+        batch_idx = self.state.batch_idx
```
ravi-mosaicml, line 469: `state.batch_idx` is deprecated

```suggestion
        batch_idx = int(self.state.timer.batch_in_epoch)
```

composer/profiler/profiler.py
```diff
@@ -508,9 +522,9 @@ def finish(self) -> None:
 
     def instant(self) -> None:
         """Record an instant event."""
-        epoch = self.profiler.state.epoch
-        step = self.profiler.state.step
-        batch_idx = self.profiler.state.batch_idx
+        epoch = self.state.epoch
```
ravi-mosaicml, line 525: ditto

composer/profiler/profiler.py
```diff
@@ -484,8 +498,8 @@ def finish(self) -> None:
 
         if self._action_at_start in self.actions:
             wall_clock_time = time.time_ns()
-            epoch = self.profiler.state.epoch
-            step = self.profiler.state.step
+            epoch = self.state.epoch
```
ravi-mosaicml, line 501: ditto

composer/profiler/profiler.py
```diff
@@ -523,7 +537,7 @@ def instant(self) -> None:
 
     def counter(self, values: Dict[str, Union[float, int]]) -> None:
         """Record a counter event."""
-        batch_idx = self.profiler.state.batch_idx
+        batch_idx = self.state.batch_idx
```
ravi-mosaicml, line 540: ditto

composer/profiler/profiler_hparams.py
```diff
@@ -8,21 +8,31 @@
...
 
 if TYPE_CHECKING:
-    from composer.profiler.dataloader_profiler import DataloaderProfiler
     from composer.profiler.json_trace import JSONTraceHandler
-    from composer.profiler.system_profiler import SystemProfiler
-    from composer.profiler.torch_profiler import TorchProfiler
 
 
 @dataclasses.dataclass
 class ProfilerCallbackHparams(CallbackHparams, abc.ABC):
```
ravi-mosaicml, line 18: I don't think this class is being used by anything anymore?

composer/profiler/torch_profiler.py
```diff
@@ -96,6 +95,7 @@ def _scheduler_fn(self, profiler_step: int, state: State) -> TorchProfilerAction
         assert state.profiler is not None, "composer profiler should be defined"
         composer_profiler_action = state.profiler.get_action(next_batch_in_epoch)
         next_composer_profiler_action = state.profiler.get_action(next_batch_in_epoch + 1)
+        from composer.profiler.profiler import ProfilerAction
```
ravi-mosaicml, line 98: ditto

composer/profiler/profiler.py
```diff
@@ -167,20 +152,45 @@ class Profiler:
...
+                 warmup: int = 1,
+                 active: int = 4,
+                 repeat: int = 1,
+                 merged_trace_file: str = "merged_profiler_trace.json",
+                 profile_cpu: bool = True,
+                 profile_sys_memory: bool = False,
+                 profile_disk: bool = False,
+                 profile_net: bool = False,
+                 stats_thread_interval_seconds: float = 0.5,
+                 torch_profiling: bool = True,
```
ravi-mosaicml, line 168: Can you remove this parameter, and instead make `tensorboard_trace_handler_dir` optional and default to None? If the `tensorboard_trace_handler_dir` is None, then no torch profiling -- otherwise, if it is set, then traces will be stored there.

ravi-mosaicml (commented):
Looking good! Main pieces of feedback are to further flatten the parameters into the trainer_hparams and trainer init signature (see [style guide](https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#3-public-apis) section 3.3), which I then think will help clean up some of the arguments for the profiler and the logic for constructing the callbacks.

I can take a stab at the imports later this week to see if it's possible to do top-level ones without getting circular errors.

composer/profiler/profiler.py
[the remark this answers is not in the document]
bandish-shah: Just a thought, does it make sense to add `wall_clock_time_ns` to the `Timestamp` object?  Makes sense that wall clock time would also be tracked as part of our time abstraction.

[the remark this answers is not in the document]
ravi-mosaicml: Yup, see #256.

composer/core/engine.py
```diff
@@ -95,14 +94,15 @@ def run_event(
         event = Event(event)
 
         if self.state.profiler is not None:
+            from composer.profiler import ProfilerAction
```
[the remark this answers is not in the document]
bandish-shah: It doesn't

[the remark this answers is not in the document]
bandish-shah: Ok with the complete refactor it works

composer/profiler/dataloader_profiler.py
```diff
@@ -22,9 +23,9 @@ class ProfiledDataLoader(WrappedDataLoader):
         name (str): The name for the dataloader.
     """
 
-    def __init__(self, profiler: Profiler, dataloader: DataLoader, name: str) -> None:
+    def __init__(self, state: State, dataloader: DataLoader, name: str) -> None:
         super().__init__(dataloader)
-        self._marker = profiler.marker(f"dataloader/{name}", categories=["dataloader"])
+        self._marker = state.profiler.marker(f"dataloader/{name}", categories=["dataloader"])
```
ravi-mosaicml, line 28: I'd be surprised if this is passing pyright since profiler is an optional member of state.

composer/profiler/event_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        and then again with ``is_start = False``. Interleaving events are not permitted.
+        Specifically, for each event (identified by the ``name``), a call with ``is_start = True`` will be followed
+        by a call with ``is_start = False`` before another call with ``is_start = True``.
+
+        Args:
+            name (str): The name of the event.
+            categories (List[str] | Tuple[str, ...]): The categories for the event.
+            is_start (bool): Whether the event is a start event or end event.
+            timestamp (Timestamp): Snapshot of the training time.
+            epoch (int): The epoch corresponding to the event.
```
ravi-mosaicml, line 47:
```suggestion
```

composer/profiler/json_trace.py
```diff
@@ -141,18 +144,17 @@ def process_duration_event(
...
         process_id: int,
         thread_id: int,
     ) -> None:
         ph = "B" if is_start else "E"
         args = {}
-        if epoch is not None:
-            args["epoch"] = epoch
-        if step is not None:
-            args["step"] = step
+        if timestamp.epoch is not None:
```
ravi-mosaicml, line 154: If the timestamp is required, then this check can be removed. Not sure why we made it optional in the first place since it's required in the base class...

composer/profiler/json_trace.py
```diff
@@ -141,18 +144,17 @@ def process_duration_event(
         name: str,
         categories: Union[List[str], Tuple[str, ...]],
         is_start: bool,
-        epoch: Optional[int],
-        step: Optional[int],
+        timestamp: Optional[Timestamp],
```
ravi-mosaicml, line 147:
```suggestion
        timestamp: Timestamp,
```

composer/profiler/json_trace.py
```diff
@@ -167,17 +169,16 @@ def process_instant_event(
         self,
         name: str,
         categories: Union[List[str], Tuple[str, ...]],
-        epoch: Optional[int],
-        step: Optional[int],
+        timestamp: Optional[Timestamp],
```
ravi-mosaicml, line 172: Same thing here

composer/profiler/json_trace.py
```diff
@@ -167,17 +169,16 @@ def process_instant_event(
...
         wall_clock_time_ns: int,
         process_id: int,
         thread_id: int,
     ) -> None:
         args = {}
-        if epoch is not None:
-            args["epoch"] = epoch
-        if step is not None:
-            args["step"] = step
+        if timestamp.epoch is not None:
```
ravi-mosaicml, line 178: ditto

composer/profiler/event_handler.py
```diff
@@ -0,0 +1,99 @@
...
+            step (int): The step corresponding to the event.
+            wall_clock_time_ns (int): The :meth:`time.time_ns` corresponding to the event.
+            process_id (int): The process id corresponding to the event.
```
ravi-mosaicml, line 50: Thoughts about renaming the `process_id` and `thread_id` arguments to `global_rank` and `pid` throughout the profiler to be more descriptive of the values that are actually being passed through? I think we called it `process_id` and `thread_id` to match the Chrome JSON trace format, but I don't think we need to follow that convention in our profiler design.

composer/profiler/profiler_hparams.py
ravi-mosaicml: (I think we used this when each callback had separate hparams, but it looks like these classes are removed below?)

composer/profiler/system_profiler.py
```diff
@@ -32,73 +32,75 @@ def __init__(self,
...
-                    Make sure to run composer with the profiler -- i.e. with the `--profiler` CLI flag."""))
+                    To use the dataloader state.profiler, state.profiler must be set.
+                    Make sure to run composer with the state.profiler -- i.e. with the `--profiler` CLI flag."""))
 
         # Start the stats thread
-        threading.Thread(target=self._stats_thread, daemon=True, args=[state.profiler]).start()
+        threading.Thread(target=self._stats_thread, daemon=True, args=[state]).start()
 
-    def _stats_thread(self, profiler: Profiler):
+    def _stats_thread(self, state: State):
```
ravi-mosaicml, line 61: I think just the profiler needs to be passed through? (No need for the entire state since the markers no longer require it?)

composer/trainer/trainer.py
```diff
@@ -177,7 +209,25 @@ def __init__(
         save_compression: str = '',
 
         # Profiling
-        profiler: Optional[ProfilerHparams] = None,
+        profiling: bool = False,
+        prof_merged_trace_file: str = "merged_profiler_trace.json",
```
ravi-mosaicml, line 213: See docstring comment above
```suggestion
        prof_merged_trace_file: Optional[str] = None,
```

composer/trainer/trainer.py
```diff
@@ -177,7 +209,25 @@ def __init__(
         save_compression: str = '',
 
         # Profiling
-        profiler: Optional[ProfilerHparams] = None,
+        profiling: bool = False,
```
ravi-mosaicml, line 212:
```suggestion
```

composer/trainer/trainer.py
```diff
@@ -315,10 +365,32 @@ def __init__(
...
+                                           event_handlers=prof_event_handlers,
+                                           skip_first=prof_skip_first,
+                                           wait=prof_wait,
+                                           warmup=prof_warmup,
+                                           active=prof_active,
+                                           repeat=prof_repeat,
+                                           merged_trace_file=prof_merged_trace_file)
             self.state.callbacks.extend(self.state.profiler.event_handlers)
+            self.state.callbacks.append(DataloaderProfiler())
+            self.state.callbacks.append(
```
ravi-mosaicml, line 380: Can we condition adding in the system profiler `if sys_prof_cpu or sys_prof_memory or sys_prof_disk or sys_prof_net`? I doubt that having the polling thread in the background would have much overhead, but would still not like to have it running if nothing is using it.

composer/trainer/trainer_hparams.py
```diff
@@ -234,7 +234,31 @@ class TrainerHparams(hp.Hparams):
         it will override train_dataset.datadir and val_dataset.datadir"""),
                                          default=None)
 
-    profiler: Optional[ProfilerHparams] = hp.optional(doc="Profiler hparams", default=None)
+    #profiler: Optional[Profiler] = hp.optional(doc="Profiler", default=None)
```
ravi-mosaicml, line 237:
```suggestion
```

composer/trainer/trainer_hparams.py
```diff
@@ -234,7 +234,31 @@ class TrainerHparams(hp.Hparams):
...
+    profiling: bool = hp.optional(doc="Enable training profiling", default=False)
+    prof_merged_trace_file: str = hp.optional(doc="Name of the Mosaic+Torch merged traced file",
+                                              default="merged_profiler_trace.json")
+    prof_event_handlers: List[ProfilerEventHandlerHparams] = hp.optional(
```
ravi-mosaicml, line 241: I think this will need an entry in the hparams_registry

setup.py
```diff
@@ -89,11 +89,13 @@ def package_files(directory: str):
     'myst-parser>=0.15.2',
     'pylint>=2.12.2',
     'docformatter>=1.4',
+    'psutil>=5.8.0',
```
ravi-mosaicml, line 92: The error messages mention to install composer with perf...could we defer the refactoring of the imports to another PR?
(I agree we have wayyyyy tooo many extra dep sections)

ravi-mosaicml (commented):
Overall LGTM! Just a few minor comments throughout.

I believe that the profiler test cases will also have to be refactored. Will approve once Jenkins is happy.

composer/profiler/event_handler.py
bandish-shah: I agree, given that it's a Chrome trace specific thing that mapping is only really relevant at the trace generation level.  having the arguments be `global_rank` and `pid` in the code would be much more clear.

setup.py
bandish-shah: Yea for sure but we need to refactor this to have 2-3 dependencies max.  There is way too many dependencies for a user to track, this doesn't fit well under the usability umbrella.  I only mention this because even I have gotten caught up by this and it's just a nuisance.

llm_flops.py
```diff
@@ -0,0 +1,7 @@
+from composer import Trainer, algorithms, trainer
```
ravi-mosaicml, line 1: I don't think this file should be committed?

composer/trainer/trainer.py
```diff
@@ -117,6 +123,43 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
         save_interval_unit (str): Unit of ``save_interval``. Can be ``ep`` or ``steps``. (default: ``ep``).
         save_compression (str): Compression algorithm to run on checkpoints. Can be `gzip`, `bzip2`,
             `lzma`, or left blank for no compression.  (default: ``""`` for no compression).
+        profiler_trace_file (str, optional): Name of the trace file, relative to the run directory.
+            Must be specified to activate the profiler. (default: ``None``).
+        prof_event_handlers (List[ProfilerEventHandler], optional): Trace event handler.
+            Ignored if ``profiler_trace_file`` is not specified. (default: ``[JSONTraceHandler()]``).
+        prof_skip_first (int, optional): Number of batches to skip at epoch start.
+            Ignored if ``profiler_trace_file`` is not specified. (default: ``optional``).
```
ravi-mosaicml, line 131:
```suggestion
            Ignored if ``profiler_trace_file`` is not specified. (default: ``0``).
```

Same thing below...

composer/trainer/trainer.py
```diff
@@ -45,6 +49,8 @@
 if TYPE_CHECKING:
     import deepspeed
 
+    from composer.profiler import ProfilerEventHandler
```
ravi-mosaicml, line 52: does this need to be a conditional import?

composer/profiler/torch_profiler.py
```diff
@@ -7,16 +7,18 @@
...
 
-from composer.core import Callback, Logger, State
-from composer.core.profiler import ProfilerAction
-from composer.profiler.profiler_hparams import TorchProfilerHparams
+from composer.core import Callback
+from composer.profiler.profiler import ProfilerAction
 from composer.utils import dist, run_directory
 
+if TYPE_CHECKING:
+    from composer.core import Logger, State
```
ravi-mosaicml, line 20: Could these imports be combined with the one on line 15?

composer/profiler/system_profiler.py
```diff
@@ -32,28 +33,28 @@ def __init__(self,
...
+                "Please install composer with pip install composer[perf] to use the state.profiler") from e
 
     def init(self, state: State, logger: Logger):
         del logger  # unused
         if state.profiler is None:
             raise RuntimeError(
                 textwrap.dedent("""\
-                    To use the dataloader profiler, state.profiler must be set.
-                    Make sure to run composer with the profiler -- i.e. with the `--profiler` CLI flag."""))
+                    To use the dataloader state.profiler, state.profiler must be set.
```
ravi-mosaicml, line 56:
```suggestion
                    To use the dataloader profiler, state.profiler must be set.
```

ravi-mosaicml (approved):
Overall LGTM; nitty comments on imports and docstrings. Let's merge it afterward!

composer/profiler/system_profiler.py
```diff
@@ -32,28 +33,28 @@ def __init__(self,
...
             import psutil
             del psutil
         except ImportError as e:
-            raise ImportError("Please install composer with pip install composer[perf] to use the profiler") from e
+            raise ImportError(
+                "Please install composer with pip install composer[perf] to use the state.profiler") from e
 
     def init(self, state: State, logger: Logger):
         del logger  # unused
         if state.profiler is None:
```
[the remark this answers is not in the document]
bandish-shah: I'd rather leave the check for `psutil` in for now until we redo our install dependencies.  I've been bitten by this a few times, just bc the trainer is constructing the System Profiler doesn't mean `psutil` is installed.  We could move the import to the top and have it fail if that's the case but I actually like the error message with the pointer on how to resolve a common issue.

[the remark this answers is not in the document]
bandish-shah: Whoops sorry for some reason I addressed the wrong code.  Yes agreed on the assert will add this in.

PR #431, opened 2022-02-08
Added :ignore-module-all: to docs

ravi-mosaicml:
We need to use `:ignore-module-all:` so sphinx does not document the same module twice, even if it is reimported.

For reimports that should be documented somewhere other than where they are defined, the re-imported item's `__module__` should be manually overridden to where the documentation should be defined. We can do this override if in a sphinx build, so we don't mess with python's internals when running code for real.

Example before (see how the FileBackendLogger is documented [here](https://mosaicml-composer--417.com.readthedocs.build/en/417/api_reference/composer.loggers.html#composer.loggers.FileLoggerBackend) and [here](https://docs.mosaicml.com/en/latest/api_reference/composer.loggers.file_logger.html#composer.loggers.file_logger.FileLoggerBackend).)

Example after (see how the FileBackendLogger is NOT documented [here](https://mosaicml-composer--431.com.readthedocs.build/en/431/api_reference/composer.loggers.html#module-composer.loggers) but is only documented [here](https://mosaicml-composer--431.com.readthedocs.build/en/431/api_reference/composer.loggers.file_logger.html#composer.loggers.file_logger.FileLogger).)


bandish-shah (approved):
LGTM

PR #432, opened 2022-02-08
Configured doctest

### LINKED ISSUE #412: Configure doctests
Let's configure https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html so we can have tested examples in our documentation

ravi-mosaicml:
1. Added a doctest fixtures file. These fixtures are available for all doctests.
2. Updated the test script to run doctests. Doctests must pass before PRs can be merged.
3. Updated all existing doctests to be correct.
4. Removed the scheduler doctests because they were incorrect and #324 will completely change the schedulers. Easier to blow them away than correct them, especially if they will be removed in a few days.

Closes #412


bandish-shah (approved):
LGTM

PR #433, opened 2022-02-08
add bfloat16 support

nikhilsardana:
Add bfloat16 support to composer.

Bfloat16 is supported in Pytorch >= 1.10 on GPU and (some operations) on CPUs. Because some major CPU operations are [not yet supported](https://github.com/pytorch/pytorch/issues/55374#issuecomment-926636577), we only enable bfloat for GPUs in this PR.

The implementation is relatively simple (see files changed), however, good performance is not guaranteed.
Ex 1: [ResNet-50 Bfloat16 vs. AMP on A100 GPUs.](https://wandb.ai/mosaic-ml/composer-examples/reports/ResNet50-Bfloat16-vs-AMP--VmlldzoxNTY3ODc5) ~27x worse training speed for Bfloat16 (similar accuracy).
Ex 2: [GPT-2 Bfloat16 vs. AMP on 3090s/A100 GPUs.](https://wandb.ai/mosaic-ml/composer-examples/reports/GPT-2-Bfloat16-vs-AMP--VmlldzoxNTY3ODg0) Bfloat16 and AMP have very similar training speed and performance.

We hypothesize that certain convolutional operations have not been optimized for Bfloat16 precision, and the fallback CUDA operations have high overhead, which causes models like ResNet-50 to see large performance drops using Bfloat16 compared to mixed precision. Transformer-based models (without convolutions) see similar performance using Bfloat16 vs. AMP.


composer/trainer/devices/device_gpu.py
```diff
@@ -33,14 +34,21 @@ def tensor_to_device(self, tensor: Tensor) -> Tensor:
     @contextmanager
     def precision_context(self, precision: Union[str, Precision]) -> Generator[None, None, None]:
         precision = Precision(precision)
+        enabled = False
         if precision == Precision.FP32:
             enabled = False
         elif precision == Precision.AMP:
             enabled = True
+        elif precision == Precision.BF16 and version.parse(torch.__version__) >= version.parse("1.10"):
```
hanlint, line 42: maybe something like:
```
elif precision == Precision.BF16:
    if version.parse(torch.__version__) < version.parse("1.10"):
        raise ValueError("BF16 precision requires torch > 1.10, got version {torch.__version__})
    enabled=True
```

Otherwise, users on torch 1.9 would get a "BF16 not supported for a GPU" error which isn't actually true.

composer/trainer/devices/device_gpu.py
```diff
@@ -33,14 +34,21 @@ def tensor_to_device(self, tensor: Tensor) -> Tensor:
...
+            enabled = True
         else:
             raise ValueError(f"Precision {precision} not supported for a GPU")
-        with torch.cuda.amp.autocast(enabled):  #type: ignore
-            yield
+        if precision == Precision.BF16:
+            with torch.cuda.amp.autocast(True, dtype=torch.bfloat16):
```
hanlint, line 47: maybe for reduced verbosity, we add a `dtype` variable defined in the above if/else statements, and just do:
```
with torch.cuda.amp.autocast(enabled, dtype=dtype)
    yield
```

hanlint (commented):
LGTM, we will need some help sussing out the performance degradations for CNNs, though

composer/trainer/devices/device_cpu.py
```diff
@@ -28,6 +30,11 @@ def precision_context(self, precision: Union[str, Precision]) -> Generator[None,
         precision = Precision(precision)
         if precision == Precision.FP32:
             yield
+        elif precision == Precision.BF16:
```
hanlint, line 33: Consider not enabling this yet, as looks like bf16 for CPU is not fully baked in pytorch just yet: https://github.com/pytorch/pytorch/issues/55374

hanlint:
One other bit -- we should add BF16 into our testing as well.

tests/test_precision.py
```diff
@@ -34,9 +34,11 @@ def run_and_measure_memory(precision: Precision) -> int:
     return torch.cuda.max_memory_allocated()
 
 
-@pytest.mark.timeout(60)
+@pytest.mark.timeout(90)
 @pytest.mark.gpu
-def test_fp16_mixed():
+def test_precision():
```
hanlint, line 39: we usually prefer to have each test only test/assert one condition, so could you use parameterize to run both precisions? e.g.
```
@pytest.mark.parametrize(precision, ['Precision.AMP', 'Precision.BF16'])
def test_precision_memory(precision):
     ...
```

This would run the test twice.

Side question: would test just fail on V100 GPUs that don't wupport bfloat16? we may want to find the CUDA compatibility number for the device and skip if bf16 not supported?

nikhilsardana: I've updated the test to check the PyTorch version number. If it's less than 1.10, it skips the bfloat16 test. This way, the test should not fail when the user is running an old version of PyTorch.
Answer to side question: https://discuss.pytorch.org/t/bfloat16-native-support/117155. It looks like CUDA & PyTorch don't provide a way to see if the GPU hardware natively supports bfloat16. I believe that PyTorch > 1.10 will fallback when operations don't have native support, so it shouldn't fail. This can be tested.

tests/test_precision.py
```diff
@@ -35,8 +36,10 @@ def run_and_measure_memory(precision: Precision) -> int:
 
 
 @pytest.mark.timeout(60)
-@pytest.mark.gpu
-def test_fp16_mixed():
+@pytest.mark.parametrize("precision", ['Precision.AMP', 'Precision.BF16'])
+def test_precision_memory(precision):
+    if version.parse(torch.__version__) < version.parse("1.10"):
+        pass
```
hanlint, line 42:
```suggestion
        pytest.skip("Test required torch >= 1.10")
```

hanlint (approved):
LGTM, with one nit.

abhi-mosaic (approved):
Hey! Sorry for delay, I don't have much to add here except great work @nikhilsardana! `bfloat16` could end up being a big help for our LLM training stability.

I wanted to bring up one "silent" integration point that doesn't show up in the PR diff, that `bfloat16` does not require grad scaling while `amp` does. This logic ends up being handled here: https://github.com/mosaicml/composer/blob/2c40fcfc37ce554de5c788cd8e06210cb666db6d/composer/trainer/trainer.py#L1177 where we have an explicit check against `Precision.AMP`, which will return false for `Precision.BF16`. So it's handled correctly, but just wanted to point it out.

Also (in a future PR) I think it would be good to rename some precision things, as bfloat16 technically is using Pytorch AMP... so the names should probably be:

```
    AMP_FP16 = "amp_fp16"
    AMP_BF16 = "amp_bf16"
    APEX_FP16 = "apex_fp16"
    APEX_BF16 = "apex_bf16"  # I think this exists now
    FP32 = "fp32"
```
rather than

```
    AMP = "amp"
    FP16 = "fp16"
    FP32 = "fp32"
    BF16 = "bf16"
```

PR #437, opened 2022-02-09
Removed dead code from the public library; refactored some imports.

ravi-mosaicml:
1. Moved the benchmarker, mosaicml logger, scaling law generators, and table generators to https://github.com/mosaicml/scratch. This code is not production ready and should not be in our public library. (Also, fewer docstrings to write!)
2. Moved `surgery` from `composer.core.surgery` to `composer.utils.module_surgery`; updated imports.
3. Moved `composer.utils._time_conversion` to `composer.optim._time_conversion` since it's only the schedulers that use it.


hanlint:
I would prefer to keep the `benchmarker` and the `table generators`, as the former is to be transparent with our performance measurements, and the latter is a private utility that we use to auto-populate the tables, so need not be docstring'd.

composer/algorithms/factorize/factorize.py
```diff
@@ -44,9 +45,9 @@ def _maybe_replace_conv2d(module: torch.nn.Module, module_index: int) -> Optiona
             return FactorizedConv2d.from_conv2d(module, module_index, latent_channels=latent_channels)
         return None  # not enough rank reduction to be worth it
 
-    ret = surgery.replace_module_classes(model,
-                                         optimizers=optimizers,
-                                         policies={torch.nn.Conv2d: _maybe_replace_conv2d})
+    ret = module_surgery.replace_module_classes(model,
+                                                optimizers=optimizers,
```
ajaysaini725, line 49: Is this extra whitespace from tabs that should be removed? Can't really tell

ajaysaini725 (approved):
LGTM!

ajaysaini725:
> I would prefer to keep the `benchmarker` and the `table generators`, as the former is to be transparent with our performance measurements, and the latter is a private utility that we use to auto-populate the tables, so need not be docstring'd.
> 
> The rest SGTM



> I would prefer to keep the `benchmarker` and the `table generators`, as the former is to be transparent with our performance measurements, and the latter is a private utility that we use to auto-populate the tables, so need not be docstring'd.
> 
> The rest SGTM

Re: the `Benchmarker` - That callback is really outdated to the point where we should just re-write it from scratch. It doesn't work with the new time abstraction and also doesn't leverage any of the new functionality in the trainer for doing partial runs with `subset_num_batches`.

hanlint:
OK makes sense, thanks @ajaysaini725

composer/algorithms/factorize/factorize.py
ravi-mosaicml: I think it's from renaming `surgery` to `module_surgery`

ravi-mosaicml:
Ah ok, table code makes sense. I'll work on getting the tables to print in the docs build in a separate PR.

PR #438, opened 2022-02-09
Deprecate scale scheduler algorithm and move to trainer

jbloxham:
The upcoming schedulers refactor is easier to do if the trainer treats SSR as a first-class feature. Thus, this PR adds a new `scale_schedule_ratio` param to the trainer. The existing algorithm is still supported, but should eventually be removed.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -82,7 +82,9 @@ class Trainer:
         compute_training_metrics (bool, optional): True to compute metrics on training data and False to not.
             (default: ``False``)
         precision (str or Precision, optional): Numerical precision to use for training, one of 'fp32', 'fp16'
-            for 'amp' (recommended). (default: ``Precision.FP32``).
+            or 'amp' (recommended). (default: ``Precision.FP32``).
+        scale_schedule_ratio (float, optional): Ratio by which to scale the training duration and learning rate
+            schedules. (default: ``0``),
         dist_timeout (float, optional): Timeout, in seconds, for initializing the distributed process group.
             (default: ``15.0``)
         ddp_sync_strategy (str or DDPSyncStrategy, optional): The strategy to use for synchronizing gradients.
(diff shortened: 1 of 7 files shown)


composer/trainer/trainer.py
```diff
@@ -80,7 +82,9 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
         compute_training_metrics (bool, optional): True to compute metrics on training data and False to not.
             (default: ``False``)
         precision (str or Precision, optional): Numerical precision to use for training, one of 'fp32', 'fp16'
-            for 'amp' (recommended). (default: ``Precision.FP32``).
+            or 'amp' (recommended). (default: ``Precision.FP32``).
+        scale_schedule_ratio (float, optional): Ratio by which to scale the training duration and learning rate
+            schedules. (default: ``0``),
```
hanlint, line 87:
```suggestion
            schedules. (default: ``10``),
```

hanlint (commented):
LGTM, one typo.

composer/trainer/trainer.py
```diff
@@ -190,9 +195,24 @@ def __init__(
         # self._use_grad_scaling() will raise a RuntimeError if grad scaling is not available when it is required
         warnings.filterwarnings(action="ignore", message="torch.cuda.amp.GradScaler")
 
+        # ScaleSchedule is a deprecated algorithm, but if it is used, updated SSR with its ratio.
+        # TODO(#434): Remove this completely.
+        for algorithm in algorithms or []:
+            if isinstance(algorithm, ScaleSchedule):
+                scale_schedule_ratio = algorithm.ratio
```
ravi-mosaicml, line 202: Would it make sense to rip out the ScaleSchedule algorithm from the algorithms list?

i.e. `algorithms.remove(algorithm)`?

ravi-mosaicml (approved):
Overall LGTM, though I'm curious on the decision to keep it as a (deprecated) algorithm rather than ripping it out completely? If it is left as deprecated, then when do we expected to completely remove it (0.5 I presume?)

PR #439, opened 2022-02-09
WIP - Docstrings for `composer.utils`

ravi-mosaicml:
Added docstrings to `composer.utils`


composer/utils/module_surgery.py
```diff
@@ -71,41 +65,62 @@ def replace_module_classes(
...
+        ... nn.Linear(5408, 128),
+        ... nn.ReLU(),
+        ... nn.LogSoftmax(dim=1),
+        )
+        >>> policies = {
+        ...    nn.Conv2d: lambda x, idx: nn.Linear(16, 32),
+        ...    nn.Linear: lambda x, idx: nn.Linear(16, 64) if x.in_features == 32 else None
+        ... }
+        >>> replace_module_classes(module, policies)
+        {Conv2d(1, 32, kernel_size=(3, 3), stride=(1, 1)): Linear(in_features=16, out_features=32, bias=True), ...}
```
growlix, line 94: What happens if you have a replacement policy such that the new/replacement module in one policy is the old/module to be replaced in another policy? For example
```
policies = {
    nn.Conv2d: lambda x, idx: nn.Linear(),
    nn.Linear: lambda x, idx: nn.LayerNorm()
}
```
Would all the `Conv2d` layers be `LayerNorm` layers? I assume so. We should probably note this explicitly in the documentation. Something like "replacements are performed sequentially, so the order of replacement policies matters" and then provide an example like the above where the desired behavior (`Conv2d` -> `Linear`) isn't happening and `Conv2d` -> `LayerNorm`) is happening instead.

I suppose if ordering is important than the policy should also be an OrderedDict?

composer/utils/run_directory.py
```diff
@@ -1,5 +1,42 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""The run directory stores artifacts created during training.
+
+Callbacks, loggers, the checkpointer, and the profiler can write files to and read files from the run
+directory. For example,
+
+* The :class:`~composer.trainer.checkpoint.CheckpointSaver` can save checkpoints inside the run directory.
+* The :class:`~composer.loggers.file_logger.FileLogger` can save log files inside the run directory.
+* The :class:`~composer.profiler.profiler.Profiler` can save profiling traces inside the run directory.
```
growlix, line 10: I assume the path should be `~composer.core.profiler.profiler`, and not one of the profilers in`~composer.profiler` (e.g. `dataloader_profiler`, `system_profiler`)

composer/utils/run_directory.py
```diff
@@ -1,5 +1,42 @@
...
+the run directory can be specified using the ``--run_directory`` command line flag. For example:
+
+.. console::
+
+    composer --run_directory /path/to/run/directory -n 8 path/to/training/script.py
+
+Otherwise, if launching python directly, the run directory can be specified via the ``COMPOSER_RUN_DIRECTORY`` environment variable.
+(The :ref:`composer launch script <distributed-training>` sets this variable for you.)
+
+If not specified, then the run directory defaults to ``f"./runs/{datetime.datetime.now().isoformat()}"``,
```
growlix, line 28: this renders weirdly. not sure if there's a way to start the path on a new line.

composer/algorithms/utils/augmentation_primitives.py
```diff
@@ -0,0 +1,346 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
```
growlix, line 1: This is fantastic. Thank you so much for cleaning this up!

tests/datasets/test_add_dataset_transform.py
```diff
@@ -30,25 +30,29 @@ def generate_composition_no_tensor():
 @pytest.mark.parametrize("is_tensor_transform,index", [(False, 1), (True, 2)])
 def test_pre_post_to_tensor_compose(is_tensor_transform, index):
     dataset = generate_synthetic_dataset(generate_default_transforms())
-    dataset = add_dataset_transform(dataset, transforms.RandomAutocontrast(), is_tensor_transform=is_tensor_transform)
+    add_vision_dataset_transform(dataset,
```
growlix, line 33: Linting error on this

tests/datasets/test_add_dataset_transform.py
```diff
@@ -30,25 +30,29 @@ def generate_composition_no_tensor():
...
+                                           transforms.RandomAutocontrast(),
+                                           is_tensor_transform=is_tensor_transform)
     assert type(dataset.transform.transforms[index]) == transforms.RandomAutocontrast  # type: ignore
 
 
 @pytest.mark.parametrize("is_tensor_transform,index", [(False, 0), (True, 1)])
 def test_pre_post_to_tensor(is_tensor_transform, index):
     dataset = generate_synthetic_dataset(transforms.ToTensor())
-    dataset = add_dataset_transform(dataset, transforms.RandomAutocontrast(), is_tensor_transform=is_tensor_transform)
+    add_vision_dataset_transform(dataset,
```
growlix, line 42: also linting error on this

growlix (commented):
This is a lot of good work.

Most of my comments were regarding documentation clarity.

docs/source/trainer/detailed_usage.rst
```diff
@@ -18,10 +18,12 @@ Numerics
 Time
 ----
 
+.. _distributed-training:
 
 Distributed Training
 --------------------
 
+
 Distributed data parallel, or DDP, is an essential technique for training large models. It is among the most common ways of parallelizing machine learning models. However, as are many things, parallelization in Python is not well supported. To get around these limitations, we offer a ``composer`` launch script to set up multiprocessing and synchronization. This script is directly analogous to the ``torch.distributed.run``, ``torchrun``, and ``deepspeed`` scripts that users may be familiar with.
```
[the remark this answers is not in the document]
ravi-mosaicml: Good point; down to call it the launcher.

composer/utils/dist.py
```diff
@@ -1,5 +1,34 @@
...
+* ``MASTER_PORT``: The port for the rank-zero process.
+
+If none of these environment variables are set, this module will safely assume a single-rank configuration, where::
+
+    RANK=0
+    LOCAL_RANK=0
+    NODE_RANK=0
+    WORLD_SIZE=1
+    LOCAL_WORLD_SIZE=1
+"""
```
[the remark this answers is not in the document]
ravi-mosaicml: Yup, I didn't follow my own advice :D

composer/utils/run_directory.py
ravi-mosaicml: Yup, I wrote this in anticipation of Bandish's PR. Need to validate that the link is correct...

ravi-mosaicml: Fixed

composer/utils/module_surgery.py
ravi-mosaicml: Good points....I am updating the example to show recursion, and yes, policies should be an `OrderedDict`, since policies are applied in order for each module.

ravi-mosaicml: Well, technically, each module is a unique type so it should only match on one key. But, with inheritance, it is possible for a module to match on multiple keys, in which case yes ordering is important.

PR #445, opened 2022-02-09
Skip saving of direct serialization fields

### LINKED ISSUE #441: Loading a checkpoint overwrites `max_duration`
Loading a checkpoint overwrites `max_duration`. This makes it pretty difficult to resume training from a late-in-training checkpoint. To do so, you'd have to manually extract optimizer and scheduler states, at least, and then create a new trainer with `load_only_weights=True`. I'm not exactly sure what the right solution is, but this workaround is too cumbersome to be reasonable.

ravi-mosaicml:
When the Trainer is constructed, the user passes in a max_duration, grad_accum, and precision in `Trainer.__init__`. These values should take precedence over whatever was in a checkpoint, so these values should not be saved in the checkpoint anyways.

Since these were the only direct serialization attributes, also cleaned up the state logic to just require that the attributes to be serialized are listed. No need to list which fields are not getting serialized. Also renamed `fields` to `attrs` to reflect that State is no longer a dataclass, since fields are a dataclass concept.

Also cleaned up the state serialization by refactoring the deepspeed logic out of the state and into the trainer next to where deepspeed is initialized. The state should not need to know whether the model is deepspeed or not.

Closes #441.


composer/core/state.py
```diff
@@ -130,6 +91,20 @@ class State(Serializable):
     loss: types.Tensors
     outputs: types.Tensors
 
+    # These attributes will be serialized using .state_dict(), and loaded with .load_state_dict()
+    # All other attributes will not be serialized.
+    # For simplicity, the leading underscore for private attributes with public getters/setters
```
ajaysaini725, line 96: Which underscore is this referring to? Shouldn't `serialized_attributes` be `_serialized_attributes` because it is private?

composer/core/state.py
```diff
@@ -268,38 +243,21 @@ def state_dict(self) -> types.StateDict:
...
-                    serialized_value = state_field_value.state_dict()
-                else:
-                    serialized_value = {
-                        obj.__class__.__qualname__: obj.state_dict()
-                        for obj in ensure_tuple(state_field_value)
-                        if obj is not None
-                    }
-                state_dict[state_field_name] = serialized_value
-
+            if state_field_name == "model":
```
ajaysaini725, line 249: Shouldn't we skip model serialization if deepspeed is enabled?

ajaysaini725 (approved):
Couple questions otherwise LGTM

composer/core/state.py
ravi-mosaicml: I decided to make this public since the trainer modifies it to remove the model and optimizer from the serialized fields when using deepspeed.

ravi-mosaicml: Done

hanlint:
I agree on `max_duration` since it's a required input to the `Trainer`, but less convinced about not saving `grad_accum` and `precision`. Those seem like ones we would want to persist in the checkpoint?

Ancillary question is -- would we actually want logic where the checkpoint is loaded, _then_ any user-provided arguments through the __init__ are overridden? Right now, from what I understand, they are silently ignored.

ravi-mosaicml:
I think grad accum is very hardware-specific ..e.g. start training on 3080s, then resume on 3090s, so I don't think that should be persisted in the checkpoint.

Though I think precision is also hardware specific (e.g. some hardware has support for bf16; others support only fp16 or amp), I could go either way on it theoretically. However, since `precision` is passed in on `init` (perhaps implicitly, though a default), I think whatever is passed to the trainer should be used. This does not affect YAHP uses since that is serialized in the hparams. (https://github.com/mosaicml/composer/blob/8b625ad167f9838580e907c7b411b63f539080c3/examples/run_composer_trainer.py).

This change also makes it consistent with deepspeed, since we don't save the deepspeed config (including precision) in the checkpoint.

It also ensures that all arguments in the trainer init (except for seed) are used. All other parameters are used either to construct the classes into which the state is loaded, or are not saved.

hanlint:
Hmm OK -- might be good to get the original issue requester and user cc: @siriuslee and maybe @A-Jacobson to opine on the expected user experience here.

hanlint:
The use case for saving `precision` especially s that users will want to know whether the precision that the checkpoint model was trained with -- has implications for what precision they should deploy the model in for inference.

ravi-mosaicml:
Sounds good, if precision and/or grad_accumt should be persisted, then we need to update the trainer's init signature to be `grad_accum: Optional[int] = None` where `None` implies to use the value in the checkpoint, or if no checkpoint, then use the current default. Can implement in a separate PR if desired.

ravi-mosaicml:
Created https://github.com/mosaicml/composer/issues/451 to track this.

Issue #343, opened 2022-02-02
Unify functional API function signatures


dblalock:
Many methods have similar behavior, like data augmentation methods and surgery-based methods, but their function signatures do not necessarily match. We should make all of their function signatures as similar as possible without changing the logic.

dblalock:
Resolved (enough) by #715. I apparently didn't type the write string in #715 to make it auto-close this, so closing it manually.

PR #449, opened 2022-02-09
Hide gen_interpolation_lambda in mixup like in cutmix and augmix

dblalock:
Right now, users have to call `gen_interpolation_lambda` before they can call the functional form of `mixup`. This both adds another step the user has to take, and is inconsistent with the behavior of `augmix` and `cutmix`, which auto-generate the interpolation parameter if not provided.

This PR makes `mixup` auto-generate the interpolation lambda if not provided.

Partially addresses #343 .


composer/algorithms/mixup/mixup.py
```diff
@@ -38,7 +38,8 @@ def gen_mixup_interpolation_lambda(alpha: float) -> float:
 def mixup_batch(x: Tensor,
                 y: Tensor,
                 n_classes: int,
-                interpolation_lambda: float = 0.9,
+                interpolation_lambda: Optional[float] = 0.9,
```
ravi-mosaicml, line 41: Should the default be None instead of 0.9?

composer/algorithms/mixup/mixup.py
```diff
@@ -57,6 +58,9 @@ def mixup_batch(x: Tensor,
         y: target tensor of shape (B, f1, f2, ..., fm), B is batch size, f1-fn
             are possible target dimensions.
         interpolation_lambda: amount of interpolation based on alpha.
```
ravi-mosaicml, line 60: Can you elaborate on the docstring that it can be set to None / what that means if it's set to None?

ravi-mosaicml (commented):
LGTM, see comments re docstrings / defaults

ravi-mosaicml (approved):

Issue #366, opened 2022-02-03
Improve algorithm file code organization


dblalock:
Right now, most algorithms have an Algorithm, an AlgorithmHparams, their functional API, and various helper functions scattered throughout one file. Instead, we decided to have:

- public functional API at the top of the file
- Algorithm class after that
- other helper functions later
- AlgorithmHparams all within one shared `algorithms/hparams.py` file

This change should make editor autocomplete more useful, facilitate de-YAHP-ification, and make our source files more readable for users.

PR #452, opened 2022-02-10
Move all AlgorithmHparams classes to shared file

dblalock:
Partially addresses #366 by moving the associated `AlgorithmHparams` for each `Algorithm` to a shared file. This is to reduce clutter in the algorithm file for users reading our code and to facilitate de-YAHP-ification.

Almost all of the changes consist of:

- Cut-and-pasting the `AlgorithmHparams` subclass from an algorithm's main implementation file to `composer/algorithms/hparams.py`
- Cleaning up the imports in accordance with this change in {the original file, the algorithm's `__init__.py`, the new `hparams.py` file, `composer/algorithms/__init__.py`, `composer/algorithms/algorithm_registry.py`}.

The only deviations from this were:

- swa had its hparams in a separate `swa/hparams.py`, which is now removed.
- I moved the SWA import in SWAHparams from within `initialize_object` to the top of the file.
- I deleted a stray `pass` in ChannelsLastHparams
- `tests/test_load.py` and `tests/test_state.py` needed one modified algorithm import each
- `tests/algorithms/test_{cutmix,ghost_batchnorm,label_smoothing,selective_backprop}` needed their associated `AlgorithmHparams`  imported from `composer.algorithms` instead of their old locations.
- A few method cards (mostly mine) listed the associated `AlgorithmHparams` in the code section. I just removed these mentions for consistency rather than updating the paths.

Basically, nothing that meaningfully changes the logic, but figured I'd highlight the not-100%-repetitive parts for convenience.

I can also split this into a couple PRs if you guys want, but be warned that the imports are really hideous when they're a hybrid between the old and new hparam locations (which is why I coalesced all these changes).


ravi-mosaicml (approved):
LGTM; didn't give it a close read but seems fine given the PR description (assuming it's just a copy-and-paste refactor).

Issue #421, opened 2022-02-08
Replace the run directory with a `log_file` API


ravi-mosaicml:
The run directory is used as a staging area for callbacks, loggers, and the checkpointer to write data. Other callbacks can periodically read from this folder, for example to upload files to an object store.

This design poses some challenges:
1. Impossible to clean while training: Once a file is written to the run directory, it cannot be deleted while training. There is no guarantee when callbacks that read the run directory will fire
2. Additional user complexity. The user must pay attention to where the run directory is stored. All paths that a user provides (such as the checkpoint folder or file logger file) are relative to the run directory, which is confusing.
3. Performance degradation: Files in the run directory must be copied before being uploaded. If the run directory is on a slow disk, this blocks the training process.

As an alternative, we can expose a `loggger.log_file` API. Design proposal here: https://www.notion.so/Artifact-Logging-API-d3754be3ec874233910cb0c4f23c9d32

The advantages of this approach are that:
* Files would not need to be kept around in a local directory once uploaded
* Users would not need to worry about this temporary staging directory

ajaysaini725:
We still need to keep files separated per-rank so that the files written by each rank don't overwrite each other right? Like if a user specifies `filename.txt` for `file_logger` then each rank will be writing to a file called `filename.txt` unless we somehow split it by rank.

Also, what happens in the case where the user wants to save files locally and not upload them? Without a `run_directory` we can assume all paths provided by the user are local to `cwd` but this opens up the possibility of files from the current run overwriting previous results - which can be okay we can just throw an error if the file exists.

ravi-mosaicml:
> We still need to keep files separated per-rank so that the files written by each rank don't overwrite each other right? Like if a user specifies `filename.txt` for `file_logger` then each rank will be writing to a file called `filename.txt` unless we somehow split it by rank.

Good point, the implementation for the log file API would need to include `dist.get_global_rank()` somewhere in the object key name. For flexibility, perhaps the user can specify a format string.

> Also, what happens in the case where the user wants to save files locally and not upload them? Without a `run_directory` we can assume all paths provided by the user are local to `cwd` but this opens up the possibility of files from the current run overwriting previous results - which can be okay we can just throw an error if the file exists.

[...]

Issue #414, opened 2022-02-08
Checkpoint usability issues


siriuslee:
There are a number of usability issues that have surfaced while documenting checkpoints. 

- [x] `CheckpointLoader`, `CheckpointSaver` and `RunDirectoryUploader` all have pretty out of date docstrings (Pain level: moderate)
- [ ] `save_folder` that is a relative path should be relative to `CWD`, not the run directory. (Pain level: high)
- [ ] Saving to object store requires the `RunDirectoryUploader` callback to be provided as well. This is not obvious and probably shouldn't be required (Pain level: mild)
- [x] Checkpoints should be saved as pytorch checkpoint files, not tarballs (Pain level: high)

Additional notes:
**Save folder**
Since the user is specifying the `save_folder`, it really should be respected. If they used a relative path, assume they meant relative to the CWD, not relative to some run directory. If we are concerned about checkpoint files overwriting others, we can either error or use the timestamp to ensure the filename is unique. 

**Saving to object stores**
This is not particularly high priority, but is a usability issue. To use the object store, we should do as much as possible to just specify the URI as the `save_folder` and not require a separate callback.

ravi-mosaicml:
> There are a number of usability issues that have surfaced while documenting checkpoints.
> 
> * [ ]  `CheckpointLoader`, `CheckpointSaver` and `RunDirectoryUploader` all have pretty out of date docstrings (Pain level: moderate)
Agreed; this will be fixed by #396 
> * [ ]  `save_folder` that is a relative path should be relative to `CWD`, not the run directory. (Pain level: high)
OK, this can be changed.
> * [ ]  Saving to object store requires the `RunDirectoryUploader` callback to be provided as well. This is not obvious and probably shouldn't be required (Pain level: mild)
For this one and the above one, we should probably remove the run directory, and instead just have an object store provider pool that is attached directly to the state. Callbacks that currently write to the run directory would write directly to the object store provider. (A local directory could also function as an object store provider) 
> * [ ]  Checkpoints should be saved as pytorch checkpoint files, not tarballs (Pain level: high)
Could you elaborate on the use case for accessing a checkpoint directly? I don't think there is a standard format for how to checkpoint the training state.
> 
> Additional notes: **Save folder** Since the user is specifying the `save_folder`, it really should be respected. If they used a relative path, assume they meant relative to the CWD, not relative to some run directory. If we are concerned about checkpoint files overwriting others, we can either error or use the timestamp to ensure the filename is unique.
> 
[...]

PR #455, opened 2022-02-10
Checkpoint Usability Initial Changes

ajaysaini725:
Addresses the following in #414
- General docstring cleanup
- The default file format for checkpoints is `.pt` rather than `.tar` except when DeepSpeed is being used in which case the default is `.tar`
- Previous checkpoint file paths are accessible via a trainer property. If files are being uploaded via the `RunDirectoryUploader` then the paths will be storage provider object paths instead.

Also changed saving logic so that checkpoint tarballs are only saved on rank 0 unless deepspeed is being used.

The remaining points in #414 regarding:
1. checkpoint paths being relative to `cwd` instead of the `run_directory`
2. Not requiring the `RunDirectoryUploader` to upload checkpoints (or being more clear about it)
will be addressed by #421 in a later release.


composer/trainer/trainer.py
```diff
@@ -439,6 +441,25 @@ def __init__(
     def deepspeed_enabled(self):
         return self.deepspeed_config is not None
 
+    @property
+    def saved_checkpoint_filepaths(self):
```
ravi-mosaicml, line 445: I'm hesitant about adding this logic into the trainer itself, as a) the trainer should not know about what callbacks are being used, b) all the other checkpointing logic lives in the checkpointer class, and c) it would not be possible for the user to differentiate between getting a local path or the remote path. In addition, these this as written would not work for multi-rank training, since each rank would have it's own checkpoint saver.

Intead, would it work to split this functionality across the checkpoint saver and run directory uploader?

I'm thinking that the checkpoint saver could have a method that returns the paths to all checkpoints. It would perform a `dist.all_gather` to gather any checkpoint files from other ranks, and then return this as a dictionary of {timestamp: list of checkpoint files for that timestamp}. For example:
```python3
def get_saved_checkpoints(self) -> Dict[Timestamp, List[pathlib.Path]]
   """Returns a list of (timestamp, List[filepath]) tuples corresponding to when a checkpoint was saved and all files from all ranks corresponding to this checkpoint. This function performs a distributed reduction to gather all checkpoint files from all ranks"""
   ...
```

Then, could the run directory uploader have a method like this?

```python3
def get_uri_for_uploaded_file(self, local_filepath: Union[pathlib.Path, str]) -> str:
    """"Returns the URI to where ``local_filepath`` was uploaded or would be uploaded. ``local_filepath`` must be in the run directory, or a ValueError would be raised"""
    ...
```

Then, the user would be able to:
```python3

run_directory_uploader = RunDirectoryUploader(...)
trainer = Trainer(callbacks=[run_directory_uploader], checkpoint_folder="checkpoints")
trainer.fit()
for checkpoint_timestamp, checkpoint_filepaths in self.checkpoint_saver.get_saved_checkpoints():
    print("Local filepaths", checkpoint_filepaths)
    print("Uploaded filepaths", [run_directory_uploader.get_uri_for_uploaded_file(x) for x in checkpoint_filepaths])
```

composer/callbacks/run_directory_uploader.py
```diff
@@ -202,6 +211,11 @@ def _trigger_upload(self, logger: Optional[Logger], log_level: Optional[LogLevel
             logger.metric(log_level, {"run_directory/uploaded_files": files_to_be_uploaded})
 
 
+def get_obj_name_for_local_file(object_name_prefix: str, local_filepath: str) -> str:
```
ravi-mosaicml, line 214:
```suggestion
def _get_obj_name_for_local_file(object_name_prefix: str, local_filepath: str) -> str:
```

ravi-mosaicml, line 214: Curious why this change?

composer/trainer/checkpoint.py
```diff
@@ -333,17 +339,43 @@ def _get_checkpoint_rng_state(self, checkpoint_rng_state: StateDict) -> Optional
...
 class CheckpointSaver:
-    """Manager for saving state to checkpoint files.
+    """Manager for saving trainer state to checkpoint files.
 
     Args:
         save_folder (str): The path to store checkpoints in.
         interval (Time or str): The amount of time units to wait between checkpoints.
-        compression (str): Compression algorithm to run on checkpoints. Can be `gzip`, `bzip2`,
-            `lzma`, or left blank for no compression.  (default: ``""`` for no compression).
+        compression (str, Optional): Compression algorithm to run on checkpoints. Can be `gzip`, `bzip2`,
```
ravi-mosaicml, line 374:
```suggestion
        compression (str, optional): Compression algorithm to run on checkpoints. Can be `gzip`, `bzip2`,
```

composer/trainer/checkpoint.py
```diff
@@ -355,20 +387,16 @@ def __init__(self, save_folder: str, interval: Union[Time, str], compression: st
...
-        elif compression == "bzip2":
-            self.write_mode = "w:bz2"
-            self.file_extension = ".tar.bz2"
-        elif compression == "lzma":
-            self.write_mode = "w:xz"
-            self.file_extension = ".tar.lzma"
-        else:
-            raise ValueError(f"Unknown encryption mode: {compression}")
+        self.file_extension, self.write_mode = _format_from_compression(compression=compression)
+        self._save_paths = []
```
ravi-mosaicml, line 391: I think it'd be fine if this list was public; then no need for the `@property`. Just make sure to have the docstring in the `attributes` secition.

composer/trainer/checkpoint.py
```diff
@@ -394,6 +422,10 @@ def should_checkpoint(self, state: State, event: Event) -> bool:
     def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
         """Save the current state to a a new checkpoint file.
 
+        The default is to save checkpoints in a `.pt` file unless DeepSpeed is being used to train the model
```
ravi-mosaicml, line 425: Tick marks should probably be double ticks.

composer/trainer/checkpoint.py
```diff
@@ -427,11 +462,17 @@ def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
...
-            checkpoint_archive_filepath = os.path.join(self.checkpoint_folder, f'{tag}{self.file_extension}')
-            with tarfile.open(checkpoint_archive_filepath, self.write_mode) as tarball:
-                tarball.add(tmpdir, arcname="")  # add files flat to the tarball
-
-            log.info(f'Trainer checkpoint saved to {checkpoint_archive_filepath}')
+            checkpoint_filepath = os.path.join(self.checkpoint_folder, f'{tag}{self.file_extension}')
+            if not _is_pt_file(checkpoint_filepath):
+                with tarfile.open(checkpoint_filepath, self.write_mode) as tarball:
+                    # add files flat to the tarball with the specified compression
+                    tarball.add(tmpdir, arcname="")
```
ravi-mosaicml, line 469: Can you set the `arcname` to be something like the checkpointing timestamp and see if that fixes the permission issues we've been seeing?

composer/trainer/checkpoint.py
ravi-mosaicml: Can you also keep track of the `state.timer.get_timestamp()` at which `save_checkpoint` was called?

tests/trainer/test_checkpoint.py
```diff
@@ -114,19 +114,22 @@ def checkpointing_trainer_hparams(composer_trainer_hparams: TrainerHparams) -> T
 def assert_checkpoints_equivalent(hparams_a: TrainerHparams, checkpoint_file_a: str, hparams_b: TrainerHparams,
                                   checkpoint_file_b: str) -> None:
 
+    def _load_checkpoint(checkpoint_dir: str, filename: str):
```
ravi-mosaicml, line 117: Can you move this file outside to a top-level function? Not a fan of functions defined in functions...

composer/trainer/checkpoint.py
```diff
@@ -333,17 +339,43 @@ def _get_checkpoint_rng_state(self, checkpoint_rng_state: StateDict) -> Optional
...
+        file_extension = ".tar.gz"
+        write_mode = "w:gz"
+    elif compression == "bzip2":
+        file_extension = ".tar.bz2"
+        write_mode = "w:bz2"
+    elif compression == "lzma":
+        file_extension = ".tar.lzma"
+        write_mode = "w:xz"
+    else:
+        raise ValueError(f"Unknown encryption mode: {compression}")
```
ravi-mosaicml, line 356:
```suggestion
        raise ValueError(f"Unknown compression mode: {compression}. Valid options are ...")
```

composer/callbacks/run_directory_uploader.py
ajaysaini725: I wanted to lift the logic for creating the object name by ripping out the run directory ad attaching the prefix out so that it could be imported by the trainer as well

composer/trainer/trainer.py
ajaysaini725: Yup this sounds good!

composer/trainer/checkpoint.py
ajaysaini725: I tried it didn't :/

composer/trainer/checkpoint.py
```diff
@@ -415,23 +443,39 @@ def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
...
                 with open(composer_states_filepath, 'xb') as f:
                     torch.save(state_dict, f)
 
-            checkpoint_archive_filepath = os.path.join(self.checkpoint_folder, f'{tag}{self.file_extension}')
-            with tarfile.open(checkpoint_archive_filepath, self.write_mode) as tarball:
-                tarball.add(tmpdir, arcname="")  # add files flat to the tarball
-
-            log.info(f'Trainer checkpoint saved to {checkpoint_archive_filepath}')
+            checkpoint_filepath = os.path.join(self.checkpoint_folder, f'{tag}{self.file_extension}')
+            if _is_pt_file(checkpoint_filepath):
```
ajaysaini725, line 462: I wanted to avoid nested if statements but intentionally using them here for clarity that there is a unique codepath for `_is_pt_file` being true

ajaysaini725:
Seems like the `test_file_logger` test might need a higher timeout. It failed with a timeout error in Jenkins but I ran it again and the whole pipeline succeeded (see here: https://ci.int.mosaicml.com/blue/organizations/jenkins/scratch%2Fcommand/detail/command/4664/pipeline/24)

composer/trainer/checkpoint.py
```diff
@@ -415,23 +443,39 @@ def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
...
+                    # move the file out of tmpdir to the user-specified location
+                    shutil.move(composer_states_filepath, checkpoint_filepath)
+            else:
+                # using compression, deepspeed, or both
+                with tarfile.open(checkpoint_filepath, self.write_mode) as tarball:
+                    # add files flat to the tarball with the specified compression
+                    tarball.add(tmpdir, arcname="")
+
+            timestamp = state.timer.get_timestamp()
+            paths = dist.all_gather_object(checkpoint_filepath) if is_module_deepspeed(
```
ravi-mosaicml, line 474: Instead of using a distributed reduction here, I'm wondering if it might be better if I backtrack on my original comment, and instead have each checkpoint saver keep track only of the files it wrote. I'm now realizing that this could be misleading to the user in a multi-node environment, since this function would store paths to files on other nodes. Do you think it would be better if we offloaded this responsibility to the user -- i.e. if you want all checkpoint file paths, perform your own distributed reduction?

Otherwise, I think we should make this extremely clear in the docstring -- perhaps using `.. note::` notation -- and explain that the list index corresponds to the rank on which that the filepath is valid?

composer/trainer/checkpoint.py
ravi-mosaicml: Should the rank zero check be conditioned on whether one is using deepspeed, not whether the checkpoint is ending in  `.pt`? Thinking that if one is using compression and not deepspeed, the only rank zero should make a checkpoint tarball.

ravi-mosaicml (commented):
LGTM, though see comments. Can you also add a test for `get_uri_for_uploaded_file`?

composer/trainer/checkpoint.py
ajaysaini725: We should make it as easy as possible for the user to get the checkpoint paths so ideally they don't have to do the reduction themselves. I'll make it extremely clear in the docstring.

composer/trainer/checkpoint.py
ajaysaini725: If someone is using compression you still want to tarball the files though because the original file is written as a .pt file regardless. It is true that with compression + without deepspeed all ranks will be trying to make the tarball but that doesn't do anything bad in terms of performance and makes the code cleaner. 

I thought about making the `else` an `elif is_module_deepspeed(state.model) or dist.get_global_rank() == 0` but that just seemed more confusing from a readability perspective.

I'm not super strongly opinionated about that though so lmk if you want it changed

ravi-mosaicml: Agree it's not a performance hit, but I think it would be confusing if someone enabled compression and now started to see a bunch of tarballs showing up on non-zero ranks. They may not realize that these can safely be ignored. I know this PR didn't introduce this behavior, but it would be great if we could clean it up here.

composer/trainer/checkpoint.py
```diff
@@ -227,9 +246,9 @@ def _download_checkpoint(self, node_checkpoint_folder: str) -> Tuple[str, Option
             # will detect the process crash and terminate the other ranks
             dist.barrier()
 
-        return composer_checkpoint_filepath, extracted_checkpoint_folder
+        return composer_checkpoint_filepath, extracted_checkpoint_folder, extracted_rank_n
 
-    def _restore_checkpoint(self, state: State, composer_checkpoint_filepath: str,
+    def _restore_checkpoint(self, state: State, composer_checkpoint_filepath: str, extracted_rank_n: bool,
```
ravi-mosaicml, line 251: extracted_rank_n needs a docstring update

ravi-mosaicml (approved):
LGTM, though missing one docstring. Thanks for fixing this up!

PR #456, opened 2022-02-10
Random doc fixes

ravi-mosaicml:
1. Added the `orphan` tag to the template files so sphinx doesn't warn about missing entries from toctree
2. Added in the cutmix model card
3. Fixed the summary tables so they respect `__all__` and autolink properly
4. Fixed a few broken links
5. Hiding typing annotations from method signatures as they add wayyy too much clutter


bandish-shah (approved):
LGTM

PR #457, opened 2022-02-10
Matthew/docstrings update

growlix:
Updated docstrings in accordance with #405.

## Notes
- Module-level docstrings contain a brief description, attribution, and links to primary source(s) and method card. This often duplicates the first sentence of the algorithm class docstring.
- algorithm file (e.g. `alibi.py`) docstrings seem uninspired, not sure if there's relevant detail that can be added.
- Code examples for NLP methods are awaiting test fixtures, eta Tuesday.


composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
     This algorithm modifies the model and runs on Event.INIT. This algorithm
```
ravi-mosaicml, line 103:
```suggestion
    This algorithm modifies the model and runs on :attr:`~composer.core.event.Event.INIT`. This algorithm
```

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
     should be applied before the model has been moved to accelerators.
```
ravi-mosaicml, line 104:
```suggestion
    is applied before the model has been moved to accelerators.
```

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
 
+    Example: Waiting on language model test fixtures
+
     Args:
-        heads_per_layer: number of attention heads per layer
-        max_sequence_length: maximum sequence length that the
+        heads_per_layer (int): number of attention heads per layer
+        max_sequence_length (int): maximum sequence length that the
             model will be able to accept without returning an error
```
ravi-mosaicml, line 111: error, like in OOM?

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
             position embeddings are "transformer.wpe".
-        attention_module_name: module/class that will have
-            its self-attention function replaced. For example,
-            in HuggingFace's GPT, the self-attention module is
-            "transformers.models.gpt2.modeling_gpt2.GPT2Attention".
-        attr_to_replace: attribute that self-attention function will
+        attention_module (torch.nn.Module): module/class that will have its
+            self-attention function replaced. For example, in
+            HuggingFace's GPT, the self-attention module is
+            transformers.models.gpt2.modeling_gpt2.GPT2Attention.
```
ravi-mosaicml, line 118:
```suggestion
            ``transformers.models.gpt2.modeling_gpt2.GPT2Attention``.
```

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
+        attr_to_replace (str): attribute that self-attention function will
             replace. For example, in HuggingFace's GPT2, the
             self-attention function is "_attn".
```
ravi-mosaicml, line 121:
```suggestion
            self-attention function is ``_attn``.
```

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
-        alibi_attention: Path to new self-attention function in which ALiBi is
-            implemented. Used to replace "{attention_module}.{attr_to_replace}".
-        mask_replacement_function: Path to function to replace model's
-            attention mask. This is sometimes necessary for evaluating on
-            sequence lengths longer than the model was initialized to
-            accommodate.
+        alibi_attention (Callable): new self-attention function in which
+            ALiBi is implemented. Used to replace
+            "{attention_module}.{attr_to_replace}".
```
ravi-mosaicml, line 124:
```suggestion
            ``{attention_module}.{attr_to_replace}``.
```

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
+        mask_replacement_function (Union[Callable, None]): function to replace model's
+            attention mask. This is sometimes necessary for evaluating
+            on sequence lengths longer than the model was initialized to
+            accommodate. Takes positional arguments ``module`` and ``max_sequence_length``.
         train_sequence_length_scaling: Amount by which to scale
             training sequence length. One batch of training data will be
             reshaped from size (sequence_length, batch) to
```
ravi-mosaicml, line 131: would math notation work for the equation below?

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
-            attention mask. This is sometimes necessary for evaluating on
-            sequence lengths longer than the model was initialized to
-            accommodate.
+        alibi_attention (Callable): new self-attention function in which
+            ALiBi is implemented. Used to replace
+            "{attention_module}.{attr_to_replace}".
+        mask_replacement_function (Union[Callable, None]): function to replace model's
+            attention mask. This is sometimes necessary for evaluating
+            on sequence lengths longer than the model was initialized to
+            accommodate. Takes positional arguments ``module`` and ``max_sequence_length``.
```
ravi-mosaicml, line 128: I think it would be OK to remove the docstrings for ``apply`` and ``match`` below, or mark them as `:meta private:` (see https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html), since we don't expect the user to call them.

composer/algorithms/alibi/alibi.py
```diff
@@ -29,32 +33,33 @@ def apply_alibi(
...
+    according as per :class:`~composer.algorithms.alibi.alibi.Alibi`.
 
-    <https://arxiv.org/abs/2108.12409>`_.
+    Example: Waiting on language model test fixtures
 
     Args:
-        model: model to transform
-        heads_per_layer: number of attention heads per layer
-        max_sequence_length: maximum sequence length that the
+        model (torch.nn.Module): model to transform
```
ravi-mosaicml, line 42: For the functional form of alibi, same comments regarding formatting of parameters. I think it would also be OK for each parameter that shares the same docstring to say something like `` see :class:`Alibi` `` to reduce duplication, but also cool repeating it. Since you're the first one doing algorithm docs, you should pick the convention :D

composer/algorithms/alibi/alibi.py
```diff
@@ -29,32 +33,33 @@ def apply_alibi(
...
-        attr_to_replace: attribute that self-attention function will
+        attr_to_replace (str): attribute that self-attention function will
             replace. For example, in HuggingFace's GPT2, the
             self-attention function is "_attn".
-        alibi_attention: new self-attention function in which
+        alibi_attention (Callable): new self-attention function in which
             ALiBi is implemented. Used to replace
             "{attention_module}.{attr_to_replace}".
-        mask_replacement_function: function to replace model's
+        mask_replacement_function (Union[Callable, None]): function to replace model's
```
ravi-mosaicml, line 59: Can you update the annotation to be `Optional[Callable[[torch.nn.Module, int], torch.nn.Module]]` (assuming that it is supposed to return a module? In addition, can you add what it means to leave it with the default value of ``None``?

composer/algorithms/alibi/alibi.py
```diff
@@ -29,32 +33,33 @@ def apply_alibi(
...
+        attention_module (torch.nn.Module): module/class that will have its
             self-attention function replaced. For example, in
             HuggingFace's GPT, the self-attention module is
             transformers.models.gpt2.modeling_gpt2.GPT2Attention.
-        attr_to_replace: attribute that self-attention function will
+        attr_to_replace (str): attribute that self-attention function will
             replace. For example, in HuggingFace's GPT2, the
             self-attention function is "_attn".
-        alibi_attention: new self-attention function in which
+        alibi_attention (Callable): new self-attention function in which
```
ravi-mosaicml, line 56: Can you update the type annotation to include the input type / output type (in both the function signature and the docstring)? It's a bit unclear what I would specify here, but perhaps the (soon to be) example will demonstrate this.

composer/algorithms/alibi/alibi.py
```diff
@@ -98,26 +103,29 @@ class Alibi(Algorithm):
...
-            sequence lengths longer than the model was initialized to
-            accommodate.
+        alibi_attention (Callable): new self-attention function in which
+            ALiBi is implemented. Used to replace
+            "{attention_module}.{attr_to_replace}".
+        mask_replacement_function (Union[Callable, None]): function to replace model's
+            attention mask. This is sometimes necessary for evaluating
+            on sequence lengths longer than the model was initialized to
+            accommodate. Takes positional arguments ``module`` and ``max_sequence_length``.
         train_sequence_length_scaling: Amount by which to scale
```
ravi-mosaicml, line 129: What is the type of this parameter? I assume a float?

ravi-mosaicml: Also curious why this is only on the class form and not the functional form? Should it be on both?

composer/algorithms/augmix/augmix.py
```diff
@@ -83,29 +118,49 @@ def forward(self, img: ImageType) -> ImageType:
...
-    the combined augmented image and the original image is drawn from a
+    """AugMix (`Hendrycks et al., 2020 <http://arxiv.org/abs/1912.02781>`_) creates
+    ``width`` sequences of ``depth`` image augmentations, applies each sequence with
+    random intensity, and returns a convex combination of the ``width`` augmented images
+    and the original image.  The coefficients for mixing the augmented images are drawn
+    from a uniform ``Dirichlet(alpha, alpha, ...)`` distribution. The coefficient for
+    mixing the combined augmented image and the original image is drawn from a
     ``Beta(alpha, alpha)`` distribution, using the same ``alpha``.
 
     This algorithm runs on on :attr:`Event.FIT_START` to insert a dataset transformation. It is a no-op if this algorithm already
```
ravi-mosaicml, line 129: hmm, for some reason these references are not auto-linking. might need to specify the full path -- e.g. `~composer.core.event.Event.FIT_START`

composer/algorithms/augmix/augmix.py
```diff
@@ -83,29 +118,49 @@ def forward(self, img: ImageType) -> ImageType:
...
+        depth (int, optional): number of augmentations per sequence. -1 enables stochastic
+            depth sampled uniformly from [1, 3]. Default = ``-1``.
+        alpha (float, optional): pseudocount for Beta and Dirichlet distributions. Must be
+            > 0.  Higher values yield mixing coefficients closer to uniform weighting. As
+            the value approaches 0, the mixing coefficients approach using only one
+            version of each image. Default = ``1.0``.
+        augmentation_set (str, optional): must be one of the following options:
 
             * ``"augmentations_all"``
                 Uses all augmentations from the paper.
```
ravi-mosaicml, line 166: The description for `augmentations_original` could use some formatting for the equations.

composer/algorithms/augmix/augmix.py
```diff
@@ -116,7 +171,16 @@ class AugMix(Algorithm):
...
-                ``"color"``, ``"contrast"``, ``"sharpness"``, and ``"brightness"``.
+                ``"color"``, ``"contrast"``, ``"sharpness"``, and ``"brightness"``. The
+                original implementations have an intensity sampling scheme that samples a
+                value bounded by 0.118 at a minimum, and a maximum value of intensity*0.18
+                + .1, which ranges from 0.28 (intensity = 1) to 1.9 (intensity 10). These
+                augmentations have different effects depending on whether they are < 0 or
+                > 0 (or < 1 or > 1). "augmentations_all" uses implementations of "color",
+                "contrast", "sharpness", and "brightness" that account for diverging
+                effects around 0 (or 1).
+
```
ravi-mosaicml, line 182: Same comment for removing the docstrings for match and apply. Going to update the template not to document inherited members by default.

ravi-mosaicml (commented):
Looking great! The examples that are there help a bunch. I'll continue taking a read tomorrow. At a high level:
* Can we print the equations with `.. math::` (or some other style) of formatting?
* Can you update the docstrings so they begin with capital letters for each parameter? (I know, being a real nitpick here)
* For overridden inherited members like `match` and `apply`, I'd be OK with omitting the docstrings since the user wouldn't need to call these functions (and that way they won't be documented in the webpage). If we did have a docstring, then I would prefer to have a note saying that the engine is responsible for calling this so the user doesn't think they have to call these methods manually
* Can you double check the reference linkings? For example Event doesn't always link properly to the event class. It would be nice if they linked to the Event class documentation, which explains exactly when the events fire in the training loop.

composer/algorithms/alibi/alibi.py
growlix: More tokens than the attention matrix is wide; the input has more elements than the module can operate on. I should just remove the "error" language. I think it's safe to assume anyone doing NLP will be familiar with it.

ravi-mosaicml:
Finished taking a read through! Overall looking great! In addition to the general comments above, here are some more ones:

1. In the docstrings for each algorithm module (and perhaps the functional and class forms), can we include a link back to the method card?
2. In the method cards for these algorithms, do we want to print the documentation inline or link to the api reference? I think it's kinda nice as-is to have the code right there with the method card, in which case there's nothing to change, but that does mean that the code is documented in two different places. (Not necessarily a bad thing)
3. For the algorithms, I am wondering whether we need the sub-moduling. For example, we document `Alibi` on `composer.algorithms.alibi.alibi.Alibi`. Instead, would it make sense to document `alibi` on `composer.algorithms.alibi.Alibi`? (e.g. this page: https://mosaicml-composer--457.com.readthedocs.build/en/457/api_reference/composer.algorithms.alibi.html#module-composer.algorithms.alibi). The advantage being this is one fewer link the user needs to click to get to the source. To do this, then `alibi.py` would need to be renamed to `_alibi.py`, and the `__init__.py` would need to contain a line like `from composer.algorithms.alibi._alibi import Alibi`. We should be consistent in how we do this for all algorithms.
 good

Below are some more specific suggestions:
- [ ] For the `augmix_image` functional form, can you include the ``image`` in the doctest example (e.g. probably some PIL load command)?
- [ ] Many of the same comments from augmix apply to randaugment -- e.g. formatting of equations, autolinking of event references
- [ ] For alibi, should `_gpt2_alibi` and `_attn` be public and documented be documented? In the hyperparameters, we mention "composer.algorithms.alibi._gpt2_alibi._attn". I think it's a bit confusing to recommend to users to import a private class.
- [ ] The answer to this is probably no, but do we want to make the randaugment and augmix algorithms be on the scale 0-1 instead of 0-10? (This is outside the scope of docs for sure, but I'm commenting it here in the theme of usability)
- [ ] For `SeqLengthWarmup`, can you elaborate why step size should be a multiple of 8 for GPUs? Does this assume 8-gpu training (in which case it should be a multiple of n-gpus?)
- [ ] "Variable input lengths can create CUDA OOM errors. To avoid this, we follow PyTorch notes and pre-allocate the memory with a blank forward and backward pass." -- is there a link we can add to underlying notes?

composer/algorithms/alibi/alibi.py
growlix: Hopefully clarified

composer/algorithms/alibi/alibi.py
growlix: The functional form is sort of like graph surgery, in that it operates once. It's a bit different from graph surgery because instead of replacing a module, it mucks around with a module's attributes. The class form does the "graph surgery" and also reshapes data batches on `Event.AFTER_DATALOADER`.

growlix:
I think I was able to address everything except the `_gpt2_alibi.py` question, which is discussed immediately below. Other topics are addressed further below.

> For alibi, should _gpt2_alibi and _attn be public and documented be documented? In the hyperparameters, we mention "composer.algorithms.alibi._gpt2_alibi._attn". I think it's a bit confusing to recommend to users to import a private class.

ALibi changes the self-attention computation. The self-attention implementation and model structure (i.e. module/attribute tree) are totally inconsistent across libraries (and even within libraries), which makes the Alibi implementation a bit complicated—it needs to be bespoke for each model class.

For example, in HuggingFace's GPT-2 (which is what we use for our GPT-2), self-attention is a method named `_attn` of a `GPT2Attention` module (i.e. `transformers.models.gpt2.modeling_gpt2.GPT2Attention._attn`). The Alibi-enhanced attention for our GPT-2 is the code for `transformers.models.gpt2.modeling_gpt2.GPT2Attention._attn` with a small modification that implements Alibi. To ensure that `GPT2Attention` behaves as expected, the Alibi-enhanced attention has the same name as the attention method it replaces (`_attn`).

I agree it's weird to have users import a private class, but it's necessary given the current Alibi implementation. I dunno if it makes it any better, but they shouldn't be expected to use anything in `composer.algorithms.alibi._gpt2_alibi` for anything other than passing into `apply_alibi`. Even the `Alibi` algorithm takes strings pointing to `..._gpt2_alibi._attn` and `..._gpt2_alibi.enlarge_mask` instead of the actual functions.

Maybe a nicer implementation would infer what to use based on the model class? Happy to file an issue for that and get to it post-release, unless you think it should be addressed immediately.

> In the docstrings for each algorithm module (and perhaps the functional and class forms), can we include a link back to the method card?

Links now in every relevant docstring. All links all the time.

> 2. In the method cards for these algorithms, do we want to print the documentation inline or link to the api reference? I think it's kinda nice as-is to have the code right there with the method card, in which case there's nothing to change, but that does mean that the code is documented in two different places -- both in the method card and in the api reference. (Not necessarily a bad thing)

I opted to go with printing the code in the Method Card. I feel like maybe we should just link to the API to make the method card less cluttered, but I suppose it's not too hard to change later.

> For SeqLengthWarmup, can you elaborate why step size should be a multiple of 8 for GPUs? Does this assume 8-gpu training (in which case it should be a multiple of n-gpus?)

Everything in multiples of 8 for nvidia gpus. Added link to relevant nvidia docs in docstring.

hanlint:
@growlix looking good! There still seems to be a formatting issue with your links, e.g. for `RandAugment`

https://mosaicml-composer--457.com.readthedocs.build/en/457/api_reference/composer.algorithms.randaugment.html#module-composer.algorithms.randaugment

The link is being broken because of the newline in your code.

Is this being done by the `doc-formatter`? If so, we'll need to find a more systematic solution.

ravi-mosaicml:
> @growlix looking good! There still seems to be a formatting issue with your links, e.g. for `RandAugment`
> 
> https://mosaicml-composer--457.com.readthedocs.build/en/457/api_reference/composer.algorithms.randaugment.html#module-composer.algorithms.randaugment
> 
> The link is being broken because of the newline in your code.
> 
> Is this being done by the `doc-formatter`? If so, we'll need to find a more systematic solution.

I believe this is the case.

ravi-mosaicml:
I just pushed an example that shows how to have long links that play nicely with docformatter. The trick is to start the url before the linebreak, and then use a `\\` to continue the link on the next line. I don't think this is getting reformatted by docformatter anymore.

For example:

```
    """Randomly applies a sequence of image data augmentations (`Cubuk et al. 2019. <https://openaccess.\\
    thecvf.com/content_CVPRW_2020/papers/w40/Cubuk_Randaugment_Practical_Automated_\\
    Data_Augmentation_With_a_Reduced_Search_Space_CVPRW_2020_paper.pdf>`_) to an image. See
    :class:`~composer.algorithms.randaugment.randaugment.RandAugment` or the :doc:`Method Card
    </method_cards/rand_augment>` for details.
```

hanlint:
Oh i see, neat. Out of curiosity, is it possible to do something like:
```
__doc__ = """
Randomly applies a sequence of image data augmentations (`Cubuk et al. 2019. <{link}>`_) to an image. See
    :class:`~composer.algorithms.randaugment.randaugment.RandAugment` or the :doc:`Method Card
    </method_cards/rand_augment>` for details.
""".format(
link='https://openaccess.thecvf.com/content_CVPRW_2020/papers/w40/Cubuk_Randaugment_Practical_Automated_Data_Augmentation_With_a_Reduced_Search_Space_CVPRW_2020_paper.pdf')
```

ravi-mosaicml:
Yup, either of those would work. For long links we can also use this notation: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

hanlint (approved):
LGTM, thank you!

PR #459, opened 2022-02-10
Add default hparam values for cutout

dblalock:
tl;dr: looks like I missed cutout in #344.

Adding these defaults required a slight semantic change to make them consistently sensible: instead of only accepting an int as the cutout side length, cutout now also accepts a float in (0, 1). If given as a float, it's interpreted as a fraction of min(H, W). This is the same behavior as factorize `latent_channels`, but with `(H, W)` instead of `(in_channels, out_channels)`.

This is necessary because any fixed integer would either be too large sometimes (e.g., cifar10 is only 32x32) or too small sometimes (high-res images of 1024x1024, or even 224x224 cropped ImageNet images). Whereas a fraction of the resolution can be reasonable in all cases.

I added only basic test coverage since the logic is super simple, but could add more if desired. I'm also tempted to add nice error messages for unacceptable side lengths, but we already had none of these and I'm leaning towards adding these throughout the functional API as a separate PR.

On the critical path for #326 (although so are a bunch of other functional API changes).


composer/algorithms/cutout/cutout.py
```diff
@@ -12,7 +12,9 @@
 log = logging.getLogger(__name__)
 
 
-def generate_mask(mask: Tensor, width: int, height: int, x: int, y: int, cutout_length: int) -> Tensor:
+def generate_mask(mask: Tensor, width: int, height: int, x: int, y: int, cutout_length: Union[int, float]) -> Tensor:
+    if 0 < cutout_length < 1:
```
hanlint, line 16: The dual definition of `cutout_length` may be confusing. e.g. if 0 < 1 is a fraction, if >1 is an actual pixel. Since this is a helper function, I think its OK to keep this as pixels, and push the fraction vs pixel spec to the calling object.

composer/algorithms/cutout/cutout.py
```diff
@@ -27,13 +29,14 @@ def apply_cutout(X: Tensor, mask: Tensor):
...
 
-def cutout_batch(X: Tensor, n_holes: int, length: int) -> Tensor:
+def cutout_batch(X: Tensor, n_holes: int, length: Union[int, float]) -> Tensor:
     """See :class:`CutOut`.
 
     Args:
         X (Tensor): Batch Tensor image of size (B, C, H, W).
         n_holes: Integer number of holes to cut out
-        length: Side length of the square hole to cut out.
+        length: Side length of the square hole to cut out. If ``0 < length < 1``,
```
hanlint, line 38: Agree that fraction is a more intuitive approach. Rather than dual-using the `length` argument, which is a bit clunky, how about something more explicit:

```
length_fraction: Optional[float]
length_pixels: Optional[int]
```
```

Comment (dblalock) [reply composer/algorithms/cutout/cutout.py:36]: So you actually hit a bunch of subtle issues if they aren't combined into one parameter. Basically, if `length_fraction` has a default value, then if I want to specify a length in pixels, I can't just write:
```cutout_batch(length_pixels=64)```.
If I do, this will have to do one of:

1. Raise an error because the length was specified in two conflicting ways
2. Silently ignore the length_pixels, which is really bad.
3. Silently ignore the length_fraction, which is okay in this case, but will be really bad if a user ever passes in both by accident (and then suddenly realizes they've been running something different than what they thought the whole time with no error message).

To avoid these cases, one has to instead write:
```cutout_batch(length_pixels=64, length_fraction=None)```.

So I only wanted to provide one argument, but now I have to specify two. This is already not very pythonic, but it also messes up sweeps. For example, for factorize and squeeze-excite latent_channels, I've written in yamls:
```
latent_channels:
    - 0.25
    - 0.5
    - 64
    - 128
```
If I have to specify both params jointly, I have to split this into two completely different sweeps to get the desired behavior.

ravi-mosaicml: If the same variable is used, then let's make the docstrings rock-solid. Anther option is to have a `length_value` parameter and a `length_unit` parameter. I think this would solve the default value problem, since the default would be clear about what the unit is?

hanlint: Ah yes my proposal doesn't work when we want to have good defaults. I was thinking of setting both to `None`, and requiring one of them be set.

dblalock: @ravi-mosaicml I expanded the `length` docstring to be super explicit as you suggested.

I don't really like having a separate `length_unit` parameter in this case since the inputs are already unambiguous; i.e., one would never pass in `cutout_batch(length=0.5, length_unit='pixels'`, or `cutout_batch(length=64, length_unit='fraction'`, since neither of these make sense. And I would personally be kind of annoyed as a user if I had to pass in a second parameter when my intent is already clear from just one.

(Also it hits the same issue with sweeps, although that's not too important.)

hanlint (approved):
LGTM

Issue #365, opened 2022-02-03
Algorithm files should only have public functions that users are expected to call


dblalock:
In a given `my_algorithm/my_file.py`, all functions should either be part of the public interface and included in the relevant `__init__.py` files + documented, or prefixed with an underscore.

PR #460, opened 2022-02-10
Make algorithm functions either public or prefixed with "_"

dblalock:
Addresses #365. No real semantic changes--just makes various functions that don't show up in an algorithm's `__init__.py` prefixed with an underscore.

There are also two cases where functions are documented and suitable for public usage, but not currently present in `__init__.py` or the docs:

- `factorize.factorizing_could_speedup`
- `blurpool.blurmax_pool2d`

These functions are now present in their associated `__init__.py` files and method cards. I didn't add them to `composer.functional` since that currently only has the "main" function for a given algorithm. I see what do with auxiliary functions + classes as an open API question.

@growlix I saw that the alibi tests (and algorithm registry?) relied on functions that otherwise seemed private, sometimes provided as strings. I *think* these changes are okay, but would be great to get your input.


growlix:
The alibi-related changes are already part of #457, so unless we're both wrong this seems like good idea :P

growlix (approved):
The alibi-related changes are already part of #457, so unless we're both wrong this seems like good idea :P

ravi-mosaicml (approved):
Lgtm!

PR #461, opened 2022-02-10
Set WandB default to log rank zero only

abhi-mosaic:
This PR returns our default WandB logging to rank_zero_only. To generate unique names for groups I added the `coolname` package, see comment below. Please let me know if including this extra package is a bad idea.

In the future I'd like to consolidate some of the logger features like `rank_zero_only`, `log_interval` into a base class so that all logger backends will have them in a consistent interface, but that can be a future PR. Right now our WandB logger logs at every batch no matter what, which is kind of silly.

TODO: @abhi do some manual tests with this branch and make sure it works as expected.

### DIFF
--- a/composer/loggers/logger_hparams.py
+++ b/composer/loggers/logger_hparams.py
@@ -8,6 +8,7 @@ from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
+import coolname
 import yahp as hp
 
 from composer.core.logging import LoggerCallback, LogLevel
@@ -95,7 +96,7 @@ class WandBLoggerHparams(LoggerCallbackHparams):
     tags: str = hp.optional(doc="wandb tags comma separated", default="")
     log_artifacts: bool = hp.optional(doc="Whether to log artifacts", default=False)
     log_artifacts_every_n_batches: int = hp.optional(doc="interval, in batches, to log artifacts", default=100)
-    rank_zero_only: bool = hp.optional("Whether to log on rank zero only", default=False)
+    rank_zero_only: bool = hp.optional("Whether to log on rank zero only", default=True)
     extra_init_params: Dict[str, JSON] = hp.optional(doc="wandb parameters", default_factory=dict)
     flatten_hparams: bool = hp.optional(
         doc=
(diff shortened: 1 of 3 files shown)


composer/loggers/logger_hparams.py
```diff
@@ -8,6 +8,7 @@
 from dataclasses import asdict, dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
+import coolname
```
ravi-mosaicml, line 11: can you move the coolname import into ``initialize_object``, since it is a conditional install?

composer/loggers/logger_hparams.py
```diff
@@ -175,9 +176,17 @@ def get_flattened_dict(data: Dict[str, Any], _prefix: List[str] = []) -> Dict[st
                 )
             self.extra_init_params["config"].update(config)
 
-        name_suffix = f"Rank {dist.get_global_rank()}"
-        name = f"{self.name}_{name_suffix}" if self.name else name_suffix
-        group = self.name if (not self.group and self.rank_zero_only) else self.group
+        # If name=None, wandb.init(..) will automatically produce a unique run name
+        # But for multi-rank grouped runs, we want to provide a group name ahead of time to link the runs
+        # So let's explicitly default the run name here in a consistent way for single-rank and multi-rank
+        self.name = self.name if self.name else coolname.generate_slug(2)
```
ravi-mosaicml, line 182: Does this work for multi-rank training if the user does not set a random seed? I forget how we do this in the trainer init.

ravi-mosaicml: If we wanted to ensure a random name for each run (but the same random name across all ranks), then I think we'd need to move this logic into the `init()` event of the wandb callback. I think it would be possible to 1) capture the state of python's RNG, 2) re-seed python's RNG based on, say, the unix timestamp, 3) generate a coolname, 4) share this coolname with a ` dist.broadcast` , and 5) restore python's RNG state. Perhpas this is overkill, in which case I'm fine leaving this as-is.

ravi-mosaicml (approved):
LGTM

composer/loggers/logger_hparams.py
abhi-mosaic: If the user does not set a random seed, I thought that Composer uses a default? Otherwise we wouldn't have random seed consistency across ranks, which seems like it could break dataset handling and/or model initialization...

abhi-mosaic: OK after reading through `trainer.py`, I think I didn't understand how random seeds are handled. Actually, I'm now more confused as to how the WandB ranks are getting the same random name if each device's seed is `seed += dist.get_global_rank()`. Debugging...

hanlint: The `WandbLogger` get constructed before the trainer seeding is done. The flow is usually:
```
logger = WandBlogger(). # coolname defined here (or through wandb_hparams.initialize_object())
trainer = Trainer(loggers=[logger]) # seed += rank happens here
```

PR #462, opened 2022-02-11
Fixes empty log lines if no algorithms are run

siriuslee:
Fixes issue with empty log lines when events were triggered with no algorithms present.


ravi-mosaicml (approved):
LGTM! Thanks for fixing

PR #466, opened 2022-02-11
bug in train metrics

A-Jacobson:
I believe we need train metrics here, this should be returning val metrics


anisehsani (approved):
lgtm - thanks!

PR #468, opened 2022-02-11
notebook tests

hanlint:
For fast testing of notebooks, we need to patch the trainer attributes `train_subset_num_batches` and `max_duration` to train for just a few batches.

To do so, we require that all notebooks (1) tag the cell where they call `trainer.fit` with `trainer_fit` metadata, and (2) `trainer` must be defined separately from the `fit` function, and (3) The trainer must be named `trainer`.

The second one is quite inconvenient, but we need somewhere to inject code to modify the trainer attributes before execution. Roughly, we do:

```
# notebook cell 1
trainer = Trainer(...)

# test injects this code
trainer.state.max_duration = Time.from_timestring("1ep")
trainer.state.train_subset_num_batches = 5

# notebook cell 2
trainer.fit()
```

I am open to other approaches to quickly test notebooks and impose minimal requirements on the notebook writer. For example, we could ask all trainers define upfront those two variables in a cell with a tag, and ensure that those variables are not modified throughout. But that seems to impose an odd pattern to the all the notebooks.

This could would simplified if `max_duration` could be batches, cc: @jbloxham


pyproject.toml
```diff
@@ -66,6 +66,7 @@ markers = [
     # Specify -m weekly_regression to run weekly_regression tests.
     "deepspeed",
     # Whether the test needs to initialize the DeepSpeed engine.
+    "notebooks",
```
ravi-mosaicml, line 69: I'm curious why we need a seperate marker for notebooks? Would it work to mark them as `pytest.mark.timeout(15)` and/or `pytest.mark.gpu`?

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+        try:
+            # patch the trainer to only train for 5 batches
+            tb.inject("""
+                from composer.core import Time
+                trainer.state.max_duration = Time.from_timestring('1ep')
+                trainer.state.train_subset_num_batches = 5
+            """)
+            trainer = tb.ref("trainer")
+            assert trainer.state.train_subset_num_batches == 5
+        except Exception as e:
```
ravi-mosaicml, line 67: Can a more specific exception be caught? Probably an `AttributeError`?

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+    for stop in seek(tb, tag="trainer_fit"):
+        # run until right before the next trainer.fit call
+        tb.execute_cell(range(start, stop))
+
+        try:
+            # patch the trainer to only train for 5 batches
+            tb.inject("""
+                from composer.core import Time
+                trainer.state.max_duration = Time.from_timestring('1ep')
+                trainer.state.train_subset_num_batches = 5
```
ravi-mosaicml, line 63: make the tests run faster :)

```suggestion
                trainer.state.train_subset_num_batches = 2
```

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+    start = 0
+    for stop in seek(tb, tag="trainer_fit"):
+        # run until right before the next trainer.fit call
+        tb.execute_cell(range(start, stop))
+
+        try:
+            # patch the trainer to only train for 5 batches
+            tb.inject("""
+                from composer.core import Time
+                trainer.state.max_duration = Time.from_timestring('1ep')
```
ravi-mosaicml, line 62: 2 epochs will catch issues introduced at the end of epoch 1 that prevent a second epoch from training.

```suggestion
                trainer.state.max_duration = Time.from_timestring('2ep')
```

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+                trainer.state.train_subset_num_batches = 5
+            """)
+            trainer = tb.ref("trainer")
+            assert trainer.state.train_subset_num_batches == 5
+        except Exception as e:
+            traceback.print_exc()
+            print("""
```
ravi-mosaicml, line 69: can the print and raise be converted into a `raise AttributError(textwrap.dedent("""Test failed to patch the trainer's max_duration....""")) from e`?

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    for cell in tb.cells:
+        if cell['cell_type'] == 'code' and 'trainer.fit' in cell['source']:
+            metadata = cell['metadata']
+            if 'tags' not in metadata or 'trainer_fit' not in metadata['tags']:
+                raise ValueError("In cell that starts with '{}', trainer.fit was found "
+                                 "but the trainer_fit tag was not in metadata.".format(cell['source'][:40]))
+
+
+@pytest.mark.timeout(120)
```
ravi-mosaicml, line 48: Can we shorten the notebook timeout tests to a max of 15 seconds? 2 minutes is kinda long...

tests/test_notebooks.py
```diff
@@ -0,0 +1,82 @@
...
+            # patch the trainer to only train for 5 batches
+            tb.inject("""
+                from composer.core import Time
+                trainer.state.max_duration = Time.from_timestring('1ep')
+                trainer.state.train_subset_num_batches = 5
+            """)
+            trainer = tb.ref("trainer")
+            assert trainer.state.train_subset_num_batches == 5
+        except Exception as e:
+            traceback.print_exc()
```
ravi-mosaicml, line 68: with the re-raise, it should not be necessary to manually print the exception
```suggestion
```

tests/test_notebooks.py
ravi-mosaicml, line 63: Will we want to monkeypatch the dataloader to load synthetic data? Perhaps in each notebook we can have a cell that defines `synthetic = False` by default, but it has code-paths for `synthetic = False` and `synthetic = True`? I could also see a user wanting to run on synthetic data if they don't want to download imagenet for example.

pyproject.toml
hanlint: One notebook still takes ~30-60 seconds, so making them a separate mark for now. It does add length runtime. Some of it is downloding the CIFAR10 dataset..

tests/test_notebooks.py
ravi-mosaicml: Another option could be manually override the `dataset` argument in the test case before the dataloader is constructed, if we don't want to expose the synthetic data option to the user.

ravi-mosaicml: I also think we may need to override the dataloader `num_workers` to be `0` and `pin_memory` to be false. I remember running into issues when configuring jenkins where the test pod would appear to randomly crash, but setting the dataloader num workers to 0 fixed it.

hanlint: We could mock the dataloader, but since some notebooks may want to plot the output data, or do other types of analysis, this is a slippery slope of features and complexity that I do not think we need just yet. Overriding the dataset would need knowledge of each individual notebooks.

After this investigation, my conclusion that we will have to live with daily integration tests of the notebooks and potentially flaky notebooks on `dev` branch. Public release notebooks will always work, as they are tied to a specific tag of `composer` and tested during release.

PR #443, opened 2022-02-09
docstrings for callbacks

dskhudia:
Docstring addition/cleanup for callbacks 

Fixes https://github.com/mosaicml/composer/issues/404

Issue #400, opened 2022-02-07
Docstrings cleanup! `composer.loggers`


ravi-mosaicml:
The `composer.loggers` needs docstrings. As you work through adding docstrings, please:

## Things to pay attention to
1. Ensure that each public module has a module-level docstring.
   i. _Need to make a module private? Rename it so the file (or folder) begins with an underscore._ 
2. On each public module (e.g. `__init__.py`, as well as individual public files), ensure that `__all__` is set to the members that should be re-exported. This is necessary so our API docs will properly list the members of that file, and not all imported members (such as from python's stdlib imports).
2. Ensure that each public class has a class-level docstring
  i. _Need to make a class private? Rename it so the class name begins with an underscore._
3. Ensure that each public function (except for `__init__`, which is handled by the class-level docstring) has a docstring.
  i. _Need to make a function private? Rename it so the function name begins with an underscore._
4. Double-check cross references. Double-tick marks (e.g. ``` ``code`` ```) should ONLY be used for values or function parameter names; never for attributes, functions, classes, or modules. We have [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) configured, so this applies to imports as well. Instead, for a cross reference, use the following format:
  i. `` :attr:`MyClass.attribute` `` for attributes
  i. `` :meth:`MyClass.my_method` `` for methods
  i `` :class:`MyClass` `` for classes
  i: `` :mod:`composer.utils` `` for modules
5. Any public functions that we expect a user to invoke should have an example. Please use [doctest](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html) to structure the example codes. Doctest ensures that the examples actually run and works and do not get out-of-date.

## Previewing docs locally

```bash
cd docs
python3 -m http.server --directory _build/html/ &  # run a web server on http://localhost:8000
rm -rf source/api_reference && make clean && make html  # clean the api reference docstrings, and rebuild the docs folder
```

Then, go to http://localhost:8000/api_reference.html to preview the api reference

PR #470, opened 2022-02-11
docstring for callbacks

dskhudia:
Docstring addition/cleanup for callbacks. Addressed comments in https://github.com/mosaicml/composer/pull/443

Fixes https://github.com/mosaicml/composer/issues/404

Replaces https://github.com/mosaicml/composer/pull/443


ravi-mosaicml:
Looking good!

I think some more examples would be helpful. For every callback, can we show how to construct the trainer with it? No need to train anything. (I hope this doesn't take too long; I think after writing the first example, it'll mainly be copy and paste)

Now, I'm being a real nitpick:

- [ ] "# trainer will run your event as follows" -> "# The trainer will run this callback whenever the EPOCH_START event is triggered, like this:"
- [ ] Can you move the examples from here (https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.callbacks.html#module-composer.callbacks) into the docstring for the ``Callback`` class (https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.core.callback.html#composer.core.callback.Callback).
- [ ] "Override _run_event() (not run_event()) to run in response to all events. If this method is overridden, then the individual methods corresponding to each event name will not be automatically called (however, the subclass implementation can invoke these methods as it wishes.)" sounds confusing to me (I know I wrote it originally!) Can you take a stab at cleaning this up? It'd probably be easier to take a read at the source code and write a new docstring for this.
- [ ] For https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.callbacks.callback_hparams.html, can we add a link back to https://docs.mosaicml.com/en/latest/tutorials/adding_models_datasets.html?highlight=yahp#trainer-with-yahp?
- [ ] For the callbacks that log data, the tables are awesome -- makes it much cleaner what is being logged! Can you elaborate in the docstring that data is being logged via the Composer Logger (and include a reference to https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.core.logging.logger.html#composer.core.logging.logger.Logger). I know this page is a stub right now, but #400 and #403 will fill it in.
- [ ] For each callback, can you move the docstring from the underlying callback method and instead into the class docstring? Then, each underlying callback method will not be documented. For example, on https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.callbacks.grad_monitor.html, I think it is slightly confusing that `after_train_batch` is documented since we do not expect the user to ever call this function manually. Instead, after the table, I am thinking we can say something like "This callback runs on the AFTER_TRAIN_BATCH event. It computes the L2 norm of gradients.....`, (In general, I think we probably do not want to include docstrings in concrete classes overridden methods, though there will definitely be exceptions to this!)
- [ ] For the memory monitor, can you enumerate exactly which statistics are recorded (one table line per statistic?)
- [ ] For https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.callbacks.run_directory_uploader.html#module-composer.callbacks.run_directory_uploader, can you link to https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.utils.run_directory.html#module-composer.utils.run_directory?
- [ ] For https://mosaicml-composer--470.com.readthedocs.build/en/470/api_reference/composer.callbacks.run_directory_uploader.html#module-composer.callbacks.run_directory_uploader, can "While all uploads happen in the background, here are some additional tips for minimizing the performance impact:" and the bullets below it be in a `.. note::`? In addition, can we remove the separate background from the bullets?
- [ ] "since data from the last upload may be lost." -> "since data written after the last upload may be lost"
- [ ] "comparing and computing" -> "computing and comparing"
- [ ] For the example, can you move the "do_not_validate" to the `.. testsetup::` section, and perhaps monkeypatch the class `RunDirectoryUploader._validate_credentials = do_not_validate`? Then, in the example, we can construct the hparams, construct the callback. Instead of appending the callback directly to the state, can you construct a trainer instance that uses this callback? (No need to show the monkeypatching, train anything, or show that any uploads happened in the docstring).
- [ ] "RunDirectoryUploader blocks the training loop to copy files from the run directory to upload_staging_folder and to queue these files to the upload queues of the workers." -> "This callback blocks the training loop to copy files from the :mod:`~composer.utils.run_directory` to the ``upload_staging_folder``."
- [ ] "Uploads are always queued at the end of the epoch." -> "In addition, uploads are queued at the end of the epoch."''

I think after this it'll be ready to merge in!

dskhudia:
@ravi-mosaicml. Thanks for the comments. 

> For every callback, can we show how to construct the trainer with it? 

Do you mean a doctest? or just trainer construction?

ravi-mosaicml:
> @ravi-mosaicml. Thanks for the comments.
> 
> > For every callback, can we show how to construct the trainer with it?
> 
> Do you mean a doctest? or just trainer construction?

A doctest example that constructs the callback, and then constructs a trainer by appending the callback via the callbacks argument.

dskhudia:
@ravi-mosaicml I addressed all the comments except "can we remove the separate background from the bullets?"

I am not sure how to do it. May be it's theme dependent or there is an option that can used with bullets but I didn't find such an option.

hanlint (approved):

Issue #448, opened 2022-02-09
Embed the algorithm and model metadata directly in the docs


ravi-mosaicml:
#437 removed the README.md files in the model and algorithms folder (since we use docstrings for module-level docs). Moved the table generation source to docs/source/tables. Need to refactor these files and add the tables into the sphinx docs.

ravi-mosaicml:
Depends on #446 and #447, so going to wait on it.

hanlint:
Ah, I missed the `README` getting deleted. I think we definitely want `README` for the `algorithms` and `model` folders, as they show up when users are navigating github. It's OK if they are duplicated in the module-level docs.

mvpatel2000:
Closing since it's done :)

Issue #447, opened 2022-02-09
Update algorithm metadata


dblalock:
In each algorithm's `__init__.py` file, there is a metadata block along the lines of:
```
_name = 'SqueezeExcite'
_class_name = 'SqueezeExcite'
_functional = 'apply_se'
_tldr = 'Replaces eligible layers with Squeeze-Excite layers'
_attribution = 'Hu et al, 2017'
_link = 'https://arxiv.org/abs/1709.01507'
_method_card = ''
```

Because of #446 along with API changes like #426 , these metadata fields are often out of date. Also, it appears that many of them:

1. never got links to the method cards in the first place, or
2. have only a subset of their functional API listed.

And, because of #446, factorization has no metadata at all.

ravi-mosaicml:
Instead of putting this in `__init__`.py, thoughts about putting it as class-level attributes on each Algorithm? For example

```python3
class SqueezeExcite(Algorithm):
    """Replaces eligible layers with Squeeze-Excite layers
    
    exxtended docstring"""
    name = "My Algorithm"
    attribution = "Hu et al, 2017"
    link = "https://arxiv.org/abs/1709.01507"
    functional = apply_se

```

If we went with this approach, then we wouldn't need the classname or tldr fields anymore (the tldr should be the first line of the docstring).

We need to update and embed the tables in the sphinx doc build (see #448), so I'll wait for resolution on this issue before working on that.

hanlint:
Completed via #471 and metadata updated.

Issue #446, opened 2022-02-09
Algorithm metadata causes pyright reportUnusedVariable errors


dblalock:
In each algorithm directory's `__init__.py`, there is metadata along the lines of:

```
_name = 'SqueezeExcite'
_class_name = 'SqueezeExcite'
_functional = 'apply_se'
_tldr = 'Replaces eligible layers with Squeeze-Excite layers'
_attribution = 'Hu et al, 2017'
_link = 'https://arxiv.org/abs/1709.01507'
_method_card = ''
```

Each of these fields causes pyright to raise a `reportUnusedVariable` error. This both clutters the pyright output during linting and prevents updating the metadata (without `type: ignore` annotations) due to our PR gates.

I propose to do one of the following in each algorithm's `__init__.py`:

1. add `type: ignore[reportUnusedVariable]` to each line,
2. configure pyright to ignore these `__init__.py` files entirely, or
3. Wrap the metadata fields in a `MetaData` class decorated with `@no_type_check` (and update `composer.tables` [logic](https://github.com/mosaicml/composer/blob/d6668bda864ed1b4819cb0154311a6d44023a142/composer/utils/tables.py#L34) appropriately.

1 and 3 both seem fine to me and I don't think this matters enough to have a whole design discussion about it. Ideally we'd use some annotation to ignore a particular error for that whole block, but that doesn't seem to be an option in pyright (or PEP 484).

Alternatively, we could centralize the metadata somewhere else, have yaml files instead, or various other options. Curious what @hanlint thinks since I don't have great perspective on the downstream usage.

ravi-mosaicml:
Cross-post from #446:

---

Instead of putting this in `__init__`.py, thoughts about putting it as class-level attributes on each Algorithm? For example

```python3
class SqueezeExcite(Algorithm):
    """Replaces eligible layers with Squeeze-Excite layers
    
    exxtended docstring"""
    name = "My Algorithm"
    attribution = "Hu et al, 2017"
    link = "https://arxiv.org/abs/1709.01507"
    functional = apply_se

```
This approach wouldn't require the classname or tldr fields anymore (the tldr should be the first line of the docstring).

hanlint:
I like the `MetaData` class. Embedding directly in the class-level attributes would clutter the namespace with attributes not really used by developers.

hanlint:
An alternative approach is to have a `metadata.json` file in each folder that contains this information. Then, in the docstrings, we can do:
```
[...]

PR #471, opened 2022-02-11
Store `metadata` in json files for `algorithms`

hanlint:
With discussions in #446 , #447 , and #448 , it's clear we need to rethink how we store metadata. Currently, we store them in the module level `__init__.py` as private attributes. This PR moves them into `metadata.json` files in each folder, e.g.
```
{
    "alibi": {
        "class_name": "Alibi",
        "functional": "apply_alibi",
        "tldr": "Replace attention with AliBi",
        "attribution": "(Press et al, 2021)",
        "link": "https://arxiv.org/abs/2108.12409v1"
    }
}
```

We chose this approach instead of a `MetaData` class or embedding in the class attribute themselves because:
(1) a separate json for metadata doesn't clutter the module namespace with non-code bits
(2) the json serialization allows access from other systems (e.g. our website's method cards) without needing run python. it's stricter than YAML, which makes it an appropriate serializaiton format for this metadata.

TODO:

- [x] update the table generators
- [ ] add back the `README` into `algorithms` folder.


ravi-mosaicml:
Would the website load directly from github repo, or would the metadata be preprocessed and stored in a database that the website reads? If the latter, then I would prefer to use YAML to be consistent with our other configs.

nqn:
> Would the website load directly from github repo, or would the metadata be preprocessed and stored in a database that the website reads? If the latter, then I would prefer to use YAML to be consistent with our other configs.

Think we should cook it into the `make html` command in the docs folder, so it won't live in a database exactly

nqn (approved):

PR #473, opened 2022-02-12
Profiler docstrings

### LINKED ISSUE #397: Docstrings cleanup! `composer.profiler`
The `composer.profiler` needs docstrings. As you work through adding docstrings, please:

## Things to pay attention to

## Previewing docs locally

```bash
cd docs
```

bandish-shah:
Resolves #397


ravi-mosaicml:
I took a first pass through the preview of the docs build (didn't look at any code changes).

While this is a good start, I think this PR could benefit from a second pass through it. All public modules will need `__all__` to be defined along with a module-level docstring. In addition, it looks like there are still some incomplete docstrings and others that are misformatting. I think we should also think about what classes and/or modules are public vs what should be private and not part of the user-facing API.

Below are some specific comments:

- [ ] For https://mosaicml-composer--473.com.readthedocs.build/en/473/api_reference/composer.profiler.html#module-composer.profiler:
  - [ ] The formatting under "profilers include" seems off -- specifically where line breaks are placed and how  subsequent lines for each bullet are indented.
  - [ ] Can you include an example of what the user needs to do to use the profiler? Probably an example with the trainer init?
  - [ ] Can we include a screenshot of the trace and a tutorial on how to visualize the trace?
- [ ] Would it make sense to make the torch profiler, dataloader profiler, and system profiler modules private, and to not re-export the underlying callbacks? Since users are no longer constructing these classes directly, I don't think we need to expose the underlying implementation of the profiler in the docs (also fewer docstrings to write lol)
- [ ] https://mosaicml-composer--473.com.readthedocs.build/en/473/api_reference/composer.profiler.event_handler.html#module-composer.profiler.event_handler
  - [ ] Missing a module level docstring.
  - [ ] In addition, the __init__ docstring doesn't mention `process_counter_event`
  - [ ] References are not auto-linking
 - [ ] https://mosaicml-composer--473.com.readthedocs.build/en/473/api_reference/composer.profiler.json_trace.html#module-composer.profiler.json_trace could use an example on how to visualize the JSON format
 - [ ] https://mosaicml-composer--473.com.readthedocs.build/en/473/api_reference/composer.profiler.profiler_hparams.html#module-composer.profiler.profiler_hparams: Can you add a YAML example for the profiler hparams?

ravi-mosaicml (commented):
Left feedback offline

composer/profiler/torch_profiler.py
```diff
@@ -31,7 +42,7 @@ class TorchProfiler(Callback):
         pip install tensorbaord torch_tb_profiler
         tensorboard --logdir tensorboard_trace_handler_dir
 
-    Also see https://pytorch.org/docs/stable/profiler.html.
+    Also see :doc:`profiler`.
```
ravi-mosaicml, line 45: Missing a "the" in the note below...it should read...

When record_shapes is specified, THE profiler will temporarily hold...

ravi-mosaicml: Also docstring below is missing a tick mark around torch_profiler

composer/profiler/_profiler.py
```diff
@@ -320,7 +350,17 @@ def __init__(self, profiler: Profiler, name: str, actions: Sequence[ProfilerActi
         self._action_at_start = None
 
     def start(self) -> None:
-        """Record the start of a duration event."""
+        """Record the start of a duration event.
+
+        To record the duration of an event, invoke :meth:`Marker.start` followed by :meth:`Marker.finish`\\:
+
+        .. code-block:: python
```
ravi-mosaicml, line 357: doctest

composer/profiler/_profiler.py
```diff
@@ -374,10 +417,19 @@ def finish(self) -> None:
         self._started = False
 
     def instant(self) -> None:
-        """Record an instant event."""
+        """Record an instant event.
+
+        To record an instant event:
+
+        .. code-block:: python
```
ravi-mosaicml, line 424: doctest

composer/profiler/_profiler.py
```diff
@@ -386,10 +438,21 @@ def instant(self) -> None:
             )
 
     def counter(self, values: Dict[str, Union[float, int]]) -> None:
-        """Record a counter event."""
+        """Record a counter event.
+
+        To record a counter event:
+
+        .. code-block:: python
```
ravi-mosaicml, line 445: doctest

composer/profiler/_profiler.py
```diff
@@ -9,43 +9,46 @@
...
-__all__ = ["Profiler", "Marker"]
+log = logging.getLogger(__name__)
 
 
 class Profiler:
-    """The Profiler produces a trace of the training graph.
+    """Records the duration of Trainer :class:`.Event` using the :class:`.Marker` API.
 
-    Specifically, it records:
+    The :class:`Profiler` is instantiated by the Composer :class:`.Trainer`\\.
```
ravi-mosaicml, line 29: `..note ::` notation?

composer/profiler/json_trace.py
```diff
@@ -16,10 +18,15 @@
...
+
 
 class JSONTraceHandler(ProfilerEventHandler):
-    """Records trace events in `JSON trace format <https://\\
-    docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview>`_.
+    """Records trace events in `JSON trace format
+    <https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview>`_.
+
+    Traces are output to ``output_directory``.  Traces can be visualized using the Chrome Trace Viewer.
+    To view in a Google Chrome browser, navigate to ``chrome://tracing`` and load the JSON trace file.
```
ravi-mosaicml, line 29: Can chrome://tracing be in link format?

composer/profiler/__init__.py
```diff
@@ -1,5 +1,52 @@
...
+
+It is required to specify an output ``profiler_trace_file`` during :class:`.Trainer` initialization to enable profiling.
+The ``profiler_trace_file`` will contain the profiling trace data once the profiling run completes.  By default, the :class:`.Profiler`,
+:class:`.DataloaderProfiler` and :class:`.SystemProfiler` will be active.  The :class:`.TorchProfiler` is **disabled** by default.
+
+To activate the :class:`.TorchProfiler`, the ``torch_profiler_trace_dir`` must be specified *in addition* to the ``profiler_trace_file`` argument.
+The ``torch_profiler_trace_dir`` will contain the Torch Profiler traces once the profiling run completes.  The :class:`.Profiler` will
+automatically merge the Torch traces in the ``torch_profiler_trace_dir`` into the ``profiler_trace_file``, allowing users to view a unified trace.
+
+The complete traces can be viewed by in a Google Chrome browser navigating to ``chrome://tracing`` and loading the ``profiler_trace_file``.
```
ravi-mosaicml, line 28: Can chrome://tracing be in link format?

composer/profiler/json_trace_merger.py
```diff
@@ -59,16 +61,13 @@ def _get_rank_to_clock_syncs(trace_files: Tuple[str, ...]) -> Dict[int, int]:
...
-    Each trace file must contain ``global_rank`` in the metadata, and
-    one trace file per rank must contain ``clock_sync_timestamp_us`` in the metadata.
+    """Merge profiler output JSON trace files together.
 
     This function will update the trace events such that:
 
     - The ``pid`` will be set to the global rank.
     - The ``ts`` is syncronized with that of the rank 0 process.
-    - Ensures that the backward process is one below the forward process
+    - Ensures that the backward process appears below the forward process
```
ravi-mosaicml, line 70:
```suggestion
    - The backward pass process appears below the forward process.
```

ravi-mosaicml (commented):
Last round of comments; all nits. Looks awesome!!

composer/profiler/__init__.py
bandish-shah: No I originally tried this but it doesn't actually open the trace viewer via a hyperlink.  Not sure if there's some magic to enable which is why I explicitly state how to open the trace viewer in all relevant areas.

composer/profiler/json_trace.py
bandish-shah: Nope

ravi-mosaicml (approved):
:rocket:

PR #474, opened 2022-02-13
Remove `dist` warnings

hanlint:
The `JSONTraceHandler` needs access to the `run_directory`, which then calls into `dist` to find the proper rank, leading to warning being emitted whenever composer is imported.
```
>>> import composer
/mnt/cota/hanlin/composer/composer/utils/dist.py:93: UserWarning: DistributedDefaultValueWarning: RANK env var not set and Torch distributed not initialized; returning 0 for global rank.
  warnings.warn(f"DistributedDefaultValueWarning: {env_var} env var not set and Torch "
/mnt/cota/hanlin/composer/composer/utils/dist.py:93: UserWarning: DistributedDefaultValueWarning: RANK env var not set and Torch distributed not initialized; returning 0 for global rank.
  warnings.warn(f"DistributedDefaultValueWarning: {env_var} env var not set and Torch "
```

This PR removes the offending line in `Trainer` that creates the `JSONTraceHandler` upon import. However, this doesn't solve the underlying issue of any calls to getting the rank leading to warnings even when the user is clearly not using distributed training (most debug, development, etc.). 

So this PR also removes the warning entirely, as ideally the warning should be scoped to the users intended use case (e.g. distributed training) and not trigger otherwise.


jbloxham (approved):
I'm OK with this change - you're absolutely right that the dist warnings have been too noisy. I'm convinced this makes the codebase better for the v0.4 release.

bandish-shah (approved):
LGTM

PR #478, opened 2022-02-15
Bert validation refactor: wip

anisehsani:
Refactoring BERT model to use Evaluators, getting rid of some messy logic in the model to specify which evaluation datasets to use. 

Removed training metrics (mostly used for debugging) since not all training metrics are compatible with all evaluation tasks.


moinnadeem:
Overall, LGTM!

Suggested changes:
The BERT Python code looks good! However, the training YAML should only run Wikipedia+BooksCorpus pretraining, and Wikipedia and BooksCorpus validation.

The YAML that needs to be merged is MNLI, where we want to train on MNLI-matched, and validate on MNLI-matched and MNLI-mismatched. 

In other words, we need to match these two YAMLs: [MNLI](https://github.com/mosaicml/composer/blob/dev/composer/yamls/models/glue/mnli.yaml) and [MNLI-mismatched](https://github.com/mosaicml/composer/blob/dev/composer/yamls/models/glue/mnli-m.yaml) to evaluate on both MNLI and MNLI-m validation splits, while training on MNLI.

Does that make sense? Let me know if I could have phrased something more clearly

moinnadeem (changes requested):
See above comment!

moinnadeem:
LGTM!

moinnadeem:
@anisehsani Did you get a chance to create WandB runs for this PR?

moinnadeem:
@anisehsani Let me know if you need a hand running WandB runs for this PR!

anisehsani:
Thanks @moinnadeem will ping you for help

moinnadeem:
Everything LGTM, other than the test duration -- @ravi-mosaicml, should we mark this as a long test? We probably don't want to update the entire `test_hparams` to be 70 seconds?

anisehsani:
@moinnadeem tests with evaluators hadn't used synthetic data - just added - currently experimenting to find the new time it takes Jenkins to run

moinnadeem:
It should be close to 10s if it was 10s before evaluators, so I'm curious where the differences are coming from @anisehsani

moinnadeem (approved):
LGTM! Let's try and move the test timing into a GH issue and figure out where the time is going there.

PR #479, opened 2022-02-15
Get functional API running enough for notebook

dblalock:
Fixing issues that stop the functional API from being used in a reasonable way. E.g., colout and cutout can't be added to a dataset's list of transforms because they *only* work when their input is a batch (not a single image, whether a PIL image or tensor). On the critical path for #326.

Update: [looks like](https://colab.research.google.com/drive/1HIxLs61pyf0ln7MlnrGYvkNHq1uVbNWu#scrollTo=oXfFDF_VUTY6) everything's working now.

### DIFF
--- a/composer/algorithms/colout/colout.py
+++ b/composer/algorithms/colout/colout.py
@@ -76,8 +76,8 @@ def colout_batch(X: torch.Tensor, p_row: float = 0.15, p_col: float = 0.15) -> t
     """
 
     # Get the dimensions of the image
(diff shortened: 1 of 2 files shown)


hanlint (approved):
LGTM, would be good to add a quick test for both single and batched images to guard against future regressions.

dblalock:
@hanlint sounds good re: regression testing. Planning on adding this to ongoing PR for #343; unified interface to data augmentation functions should also allow unified testing.

PR #482, opened 2022-02-15
Clean up Imports and Tests

ravi-mosaicml:
1. Running the doctests on the linting worker so they are only run once, and not
on a dev install or gpu workers.
2. Skipping single-rank tests without distributed
3. Removed the daily and weekly regression test stubs, since we don't have any support for regression testing currently.
4. Removed some unused imports
5. Skipping single-rank tests with distributed. We will run single-rank tests without distributed, and dual-rank tests with distributed. This should speed tests up by ~33-50%, since all single-rank tests were effectively being executed twice. Any distributed errors will be caught by the dual-rank tests.
6. Removed the deepspeed marker, since that's no longer needed. Deepspeed and gpu tests can share the same distributed environment.
7. Changed the trainer to DDP wrap modules only if the world size is greater than 1, not just if distributed is configured. DDP is ONLY needed if the world size is greater than 1; if distributed is configured and the world size is 1 (possible if someone uses deepspeed and then doesn't use deepspeed in the same training process, or in a notebook), then the module should not be wrapped.
8. Cleaned up https://github.com/mosaicml/composer/pull/393#discussion_r803110337


setup.py
```diff
@@ -81,36 +82,29 @@ def package_files(directory: str):
...
-extra_deps['perf'] = ['torch-tb-profiler>=0.3.1', 'psutil>=5.8.0', 'tensorboard>=2.7.0']
-
-extra_deps['nlp'] = [
+extra_deps['extra'] = [
+    # All conditional imports should go here
+    'deepspeed>=0.5.5',
+    'wandb>=0.12.2',
+    'apache-libcloud>=3.4.1',
+    'torch-tb-profiler>=0.3.1',
+    'psutil>=5.8.0',
```
bandish-shah, line 99: I'm thinking `torch-tb-profiler` and `psutil` should go in mosaicml[dev].  Profiling implies dev work and these modules are relatively light.  We can leave `tensorboard` as an `extra` depepndency.

bandish-shah (approved):
Overall LGTM, minor nit about profiler dependencies.

setup.py
```diff
@@ -81,36 +82,29 @@ def package_files(directory: str):
...
     'pylint>=2.12.2',
     'docformatter>=1.4',
 ]
 
-extra_deps['logging'] = ['wandb>=0.12.2', 'apache-libcloud>=3.4.1']
-
-extra_deps['perf'] = ['torch-tb-profiler>=0.3.1', 'psutil>=5.8.0', 'tensorboard>=2.7.0']
-
-extra_deps['nlp'] = [
+extra_deps['extra'] = [
```
hanlint, line 93: I'm OK with consolidating logging and perf together, for example, but domain-specific package should be kept separate, e.g. `nlp`, `vision`, and `deepspeed`. Users like to keep their environments clean and minimal. We always have the `all` option available for users.

tests/trainer/test_trainer.py
```diff
@@ -37,18 +37,10 @@ def test_init(self, config):
         trainer = Trainer(**config)
         assert isinstance(trainer, Trainer)
 
-    def test_model_ddp_wrapped(self, config, init_process_group):
-        # dist initialized first with init_process_group fixture
+    def test_model_ddp_wrapped(self, config):
```
hanlint, line 40: I dont see the corresponding change to the `Trainer` code being pushed?

hanlint (changes requested):
one requested change with imports

tests/trainer/test_trainer.py
ravi-mosaicml: https://github.com/mosaicml/composer/blob/5cc81464f5ba7277a26180f436aefcdce02d9be9/composer/trainer/trainer.py#L525

setup.py
ravi-mosaicml: I agree that it is a best practice to keep environments clean and minimal.

I updated the error messages to be extremely explicit about what packages that a user needs to install, instead of recommending `pip install mosaicml[extra]` (The `extra` option will be for jenkins, or users who do indeed want everything). For example, the `wandb` error message now says:

> This callback requires the `wandb` pip package. To install, run `pip install wandb`.


Here are a list of issues with the current setup.py approach:


1. The current categories don't offer much flexibility in keeping the dependencies minimal. For example, `wandb` and `apache-libcloud` are both in the logging category, but some users may only want `wandb` and others may only want `apache-libcloud`. Instead, by listing the package to install explicitly in the error message, the user can install exactly what they need.
3. `setup.py` hides what is missing and will be downloaded. Error messages that explicitly list what package must be installed make this information visible.
4. I'm not sure how familiar users are with the extra deps notation, especially when installing multiple extra deps. Suppose a user had a requirements file where they wanted the `unet` and `wandb` dependencies. This would require the user to write `mosaicml[unet,wandb]>=0.4` (I had to google how to do this). I think it is more intuitive to have three separate lines -- one for mosaicml, one for unet, and one for wandb.
5. Dependencies could have been listed in multiple places in the setup.py, leading to inconsistent versioning. Now, each dependency will be listed in at most one place in setup.py -- either core, dev, or extra.

setup.py
ravi-mosaicml: `dev` should only be used for those developing composer, since it includes a lot of extra bloat for sphinx and docs. I wouldn't want someone who is debugging performance to have to install those dependencies.

I would be OK with moving `torch-tb-profiler` and `psutil` to core, if they should always be installed. Thoughts?

bandish-shah: Yes in that light moving to `core` makes sense to me.

setup.py
hanlint: Explicit error messages create a try/fail loop for the users that would be unnecessary if we documented ahead of time what they needed. For example, in the unet, wandb case, users may possibly fail multiple times before they install the right imports.

Since we always have the `all` tag to install everything, I don't see the harm in allowing more options/tag for users. We are now effectively forcing them to have bloated env/dockers, or manually experimenting with the codebase (e.g. `monai` is a pretty esoteric requirement), by taking away these options.

I'm OK with consolidating most of the utility extras, but packages like `huggingface, timm, monai, and deepspeed` imho should be separated by domain. The current `extra` approach also isn't scalable -- if we add say FFCV integration, or pytorch-lightning, or fast-ai... . 

I'd rather set the pattern but reduce the amount of tags. Maybe so users need to indicate 2 to get most of their work done? e.g. `pip install mosaicml[dev, nlp]` or `pip install[dev, vision]`. 

Or, they can always just do `pip install mosaicml[all]` if they are OK with bloat. Let's offer these options, and let them decide.

A-Jacobson: A few thoughts from the peanut gallery on imports:

current tags: `logging`, `perf`, `nlp`, `vision`, `unet`, `deepspeed`

1.  looking at these, `nlp`, `vision`, `unet` are all model related. They're modelzoo/benchmarking things that aren't related to the base functionality of the trainer or algorithms. I wouldn't want to bundle them with anything anyone needs to speed up THEIR model. Some of this stuff may eventually even leave the repo.. which would solve this problem on it's own.

2. `logging`, `perf`, and `deepspeed` could arguably just be included in the base install. unless one or all of these is a known troublemaker, platform specific, or particularly heavy. in which case we should make it it's own tag.

my 2 cents, though @hanlint knows I dislike tags =)

hanlint: #1 above from @A-Jacobson is why I think `extras` is problematic. we are forcing users trying to speed up their models to install all these model-specific libraries if they want anything else beyond the base install.

composer/profiler/system_profiler.py
```diff
@@ -44,7 +44,7 @@ def __init__(self,
             import psutil
             del psutil
         except ImportError as e:
-            raise ImportError("Please install composer with pip install composer[perf] to use the profiler") from e
+            raise ImportError("Please install composer with `pip install mosaicml[extra]` to use the profiler") from e
```
bandish-shah: Can remove this if we move the `psutil` install to `core` deps.

setup.py
bandish-shah: Yes agreed with keeping environments clean and minimal, however we need to strike a balance between that and having many build targets which ultimately degrades the UX so I'm in favor of being able to consolidate as much as possible.

Having a clear error message to catch failed imports is very useful, however my only concern with a message like `pip install wandb` is that there is no explicit version specification so the user may instal lthe latest version of the package.  Many of these packages (e.g., deepspeed, monai, timm, wandb) are pre-1.0 so we should consider the possibility of things breaking due to backwards incompatible API changes.  Unfortunately that means the only reasonable solution is to include these dependencies as `extra_requires` for the time being, at least for packages that are pre-1.0.

ravi-mosaicml: I think there is consensus to move `perf`, and to some extent, `logging`, into core. I'm not thrilled about moving wandb into core, but would prefer that than to have a seperate wandb tag.

> Explicit error messages create a try/fail loop for the users that would be unnecessary if we documented ahead of time what they needed. For example, in the unet, wandb case, users may possibly fail multiple times before they install the right imports.

Ageed; let's make it clear in the docstring (maybe via `.. note::` notation) what optional dependencies must be installed.


I guess I still don't quite follow why we need tags / what problems that solves, vs having error messages and docstrings that explicitly list the imports. There will still be try/fail loops whether one is using tags or explicitly pip installing (the only difference is whether docs and error message will say to install `pip install mosaicml[unet]` vs `pip install unet`). We can print multiple dependencies in the error / check for multiple dependencies at once -- e.g. see the LM datasets.

```python
        try:
            import datasets
            import transformers
        except ImportError as e:
            raise ImportError('LM datasets require the `transformers` and `datasets` pip packages.'
                              'To install, run: `pip install datasets transformers`.') from e
```

hanlint: woo lots of discussion, awesome!

For me, the use case for tags is to specify versions to install that have been tested, so I prefer them to install via our controlled `setup.py` than whatever their `pip` will pull.

A-Jacobson: >I guess I still don't quite follow why we need tags / what problems that solves, vs having error messages and docstrings that explicitly list the imports. 


to me a tag has an advantage of being able to work with a group of related packages and a descriptive name related to the function. if we have a tag for just one package it's probably not that useful?

setup.py
ravi-mosaicml: Moving `psutil` to core. I realize that none of our code depends on `torch-tb-profiler` or `tensorboard`, so going to leave those out of the requirements.

setup.py
ravi-mosaicml: > to me a tag has an advantage of being able to work with a group of related packages and a descriptive name related to the function. if we have a tag for just one package it's probably not that useful?

For context, ALL of the extra deps are just one or two packages. Here is the complete list:

```
extra_deps["deepspeed"] = [
    'deepspeed>=0.5.5',
]

extra_deps["wandb"] = [
    'wandb>=0.12.2',
]

extra_deps["unet"] = [
    'monai>=0.7.0',
    'scikit-learn>=1.0.1',
]

extra_deps["timm"] = [
    'timm>=0.5.4',
]

extra_deps["nlp"] = [
    'transformers>=4.11.3',
    'datasets>=1.14.0',
]
```

> > pip install mosaicml[unet] vs pip install unet
> 
> these arent the same thing. `mosaicml[unet]` installs monai and some other stuff. there is no `pip install unet` or at least that's not a package we use.

Sorry, in my example I meant `pip install monai scikit-learn` vs `pip install mosaicml[unet]`. I realize these aren't the same thing.


> even if we do "extras" we probably shouldn't call it "extra" because it's not clear what's in it or when you need that particular install, until you use something and get an error.

Sure, open to a better name. What do you propose?

The purpose of "extra", as mentioned [here](https://github.com/mosaicml/composer/pull/482#discussion_r807429432), is to support jenkins, which needs to know of all extra dependencies.

ravi-mosaicml: > woo lots of discussion, awesome!

> For me, the use case for tags is to specify versions to install that have been tested, so I prefer them to install via our controlled `setup.py` than whatever their `pip` will pull.

I think this is valid. We could pin the version in the error message, or yes, attempt to enforce it via a setup.py option.

ravi-mosaicml:
Given the discussion, I added back the extra dependency targets. @hanlint could you please unblock this PR?

hanlint (approved):

PR #484, opened 2022-02-15
Add back benchmarker callback

siriuslee:
We still use the benchmarker callback internally, so I don't think we should remove it unless it actually doesn't work anymore (or we have a proper replacement). It's been fine so far in my experience, though I haven't used it in all cases.

### DIFF
--- a/composer/callbacks/__init__.py
+++ b/composer/callbacks/__init__.py
@@ -41,6 +41,8 @@ Callbacks can be implemented in two ways:
                 if event == Event.EPOCH_START:
                     print(f'Epoch {state.epoch}/{state.max_epochs}')
 """
+from composer.callbacks.benchmarker import Benchmarker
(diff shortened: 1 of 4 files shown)


ravi-mosaicml:
Can you help justify why the benchmarker should be in the public release? It was removed because it is incompatible with the new time tracking and is completely untested. See https://github.com/mosaicml/composer/pull/437#issuecomment-1033245476

Before we add it back in, would it be possible to a) convert it to an algorithm (since it is currently callback anti-pattern of modifying the state), b) update the public API to take time strings instead of epochs, c) add test cases, and d) fixes the docstrings? (I tried working on them a while back, but was very confused with what the arguments do, so not sure how correct they are). Happy to review a PR that makes it release-ready.

For now, it is available internally at https://github.com/mosaicml/scratch/blob/main/benchmarker.py.

siriuslee:
We're going to keep this on a public branch for now while we bring it up to date.

PR #485, opened 2022-02-15
Dynamic Shrinking Microbatches

mvpatel2000:
Instead of crashing with CUDA out of memory, we dynamically scale the number of microbatches so we can fit the logical batchsize irrespective of the underlying hardware. Right now, we only decrease if we crash. 

As of now, this is _not_ set as default and we emit a warning if it is used. This causes errors if used with methods which change memory usage (as of now, this is just seq_length_warmup)

[Related Design Doc](https://www.notion.so/Dynamic-Gradient-Accumulation-6da926ac1a324d29b79615b00bead698)
 
*Test Log*:
- Resnet 56 on CIFAR with batchsize initially set to 8000
```
Epoch 0:   0%|          | 0/6 [00:02<?, ?it/s]                                                                                                               
1 -> 2
Epoch 0:   0%|          | 0/6 [00:02<?, ?it/s]                                                                                                               
2 -> 4
Epoch 0:  83%|████████▎ | 5/6 [00:08<00:01,  1.40s/it, loss/train=3.2449]               
```
We are able to successfully increase grad_accum until it fits on GPU
- Resnet 56 on CIFAR with batchsize initially set to 8000 + progressive resizing
```
Epoch 39: 100%|██████████| 6/6 [00:04<00:00,  1.25it/s, loss/train=1.4595]                                                                                   
Epoch 40, Batch 240 (val): 100%|██████████| 10/10 [00:00<00:00, 17.03it/s, accuracy/val=0.2500]                                                              
Epoch 40:   0%|          | 0/6 [00:01<?, ?it/s]                                                                                                              
1 -> 2
Epoch 40: 100%|██████████| 6/6 [00:05<00:00,  1.10it/s, loss/train=1.5751]                                                                                   
Epoch 41, Batch 246 (val): 100%|██████████| 10/10 [00:00<00:00, 17.10it/s, accuracy/val=0.2736]   
```
We are able to successfully run with 8000 batchsize until images start scaling up, at which point we increase grad_accum
- GPT2_52m with default params
```
Epoch 0:   0%|          | 0/9000 [00:02<?, ?it/s]                                                                                                            
1 -> 2
Epoch 0:   0%|          | 0/9000 [00:02<?, ?it/s]                                                                                                            
2 -> 4
4 -> 8
Epoch 0:   0%|          | 0/9000 [00:02<?, ?it/s]                                                                                                            
8 -> 16
Epoch 0:   0%|          | 1/9000 [00:03<7:39:21,  3.06s/it, loss/train=10.9392]                                                                              
16 -> 32
Epoch 0:   0%|          | 2/9000 [00:04<4:50:48,  1.94s/it, loss/train=10.9292]                                                                              
32 -> 64
Epoch 0:   0%|          | 3/9000 [00:04<3:19:16,  1.33s/it, loss/train=10.6871]                                                                              
64 -> 128
Epoch 0:   0%|          | 4/9000 [00:05<2:20:50,  1.06it/s, loss/train=10.4537]                                                                              
128 -> 256
Epoch 0:   0%|          | 11/9000 [01:38<35:01:09, 14.02s/it, loss/train=9.6154]     
```
- Multi-GPU GPT2 (on 2 GPUs)
image

```
Epoch 0:   0%|          | 0/9000 [00:03<?, ?it/s]                                                                                                            
1 -> 2
Epoch 0:   0%|          | 0/9000 [00:03<?, ?it/s]                                                                                                            
2 -> 4
Epoch 0:   0%|          | 0/9000 [00:04<?, ?it/s]                                                                                                            
4 -> 8
Epoch 0:   0%|          | 1/9000 [00:05<13:06:39,  5.24s/it, loss/train=10.9348]                                                                             
8 -> 16
Epoch 0:   0%|          | 2/9000 [00:06<7:29:42,  3.00s/it, loss/train=10.9314]                                                                             
16 -> 32
Epoch 0:   0%|          | 35/9000 [03:28<15:42:19,  6.31s/it, loss/train=7.1741]   
```

_Regression Tests_
Ran some tests to confirm results are identical. With resnet 50 on imagenet 8 GPUs, we get basically the same curves (theyre overalpping) when comparing
- dev vs. this branch + no auto grad_accum vs. this branch + auto grad_accum
- same as above except including blurpool + label smoothing + mixup + channels last

Note that grad_accum is set to 2 since it was run on coda, so the auto grad_accum did actually have to activate and change grad_accum

image


composer/trainer/trainer.py
```diff
@@ -90,6 +90,8 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
             or 'amp' (recommended). (default: ``Precision.FP32``).
         scale_schedule_ratio (float, optional): Ratio by which to scale the training duration and learning rate
             schedules. See :func:`scale_schedule` for details. (default: ``1.0``)
+        adaptive_grad_accum (boolean, optional): Dynamically scale down minibatch size and use gradient
```
ajaysaini725, line 93: Can just have the `grad_accum` parameter being set to `-1` used to signal autotuning

composer/trainer/trainer.py
```diff
@@ -680,25 +685,44 @@ def _train_loop(self) -> None:
...
+                                for optimizer in state.optimizers:
+                                    if use_grad_scaling:
+                                        state.scaler.step(optimizer)
+                                    else:
+                                        optimizer.step()
+                        except RuntimeError as e:
+                            if self.adaptive_grad_accum and "CUDA out of memory" in str(e):
+                                rerun_train_batch = True
+                                # Raise runtime error if we can't train even one sample at a time
+                                if state.grad_accum >= num_samples_in_batch:
```
ajaysaini725, line 717: `grad_accum = num_samples_in_batch` should be okay since this will result in a micro batch size of 1

composer/trainer/trainer.py
```diff
@@ -710,7 +734,10 @@ def _train_loop(self) -> None:
                         # total_loss can be None if gradient scaling failed
                         dist.all_reduce(total_loss, reduce_operation="SUM")
                         full_loss = total_loss.cpu().item()
-                        self.logger.metric_batch({'loss/train': full_loss / dist.get_world_size()})
+                        self.logger.metric_batch({
+                            'loss/train': full_loss / dist.get_world_size(),
+                            'trainer/grad_accum': state.grad_accum
```
ajaysaini725, line 739: Only need to log if the grad_accum was actually changed

composer/trainer/trainer.py
```diff
@@ -90,6 +90,8 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
             or 'amp' (recommended). (default: ``Precision.FP32``).
         scale_schedule_ratio (float, optional): Ratio by which to scale the training duration and learning rate
             schedules. See :func:`scale_schedule` for details. (default: ``1.0``)
+        adaptive_grad_accum (boolean, optional): Dynamically scale down minibatch size and use gradient 
+            accumulation if train_batch_size is too large for GPU. (default: ``True``)
```
jbloxham, line 94: I think I'd prefer for this behavior to be gated behind the `grad_accum` parameter. The behavior could be something like "If -1, set `grad_accum` to the per-device batch size and enable dynamic adjustments.

composer/trainer/trainer.py
jbloxham, line 717: `num_samples_in_batch` is the total across GPUs. I think you want `state.batch_num_samples` instead.

composer/trainer/trainer.py
```diff
@@ -680,25 +685,44 @@ def _train_loop(self) -> None:
...
                     total_loss = None
-                    microbatches = self._train_data_spec.split_batch(state.batch, state.grad_accum)
-                    if self.deepspeed_enabled:
-                        total_loss = self._train_batch(microbatches)
-                    elif self._use_closures():
-                        for optimizer in state.optimizers:
-                            if use_grad_scaling:
-                                total_loss = state.scaler.step(
-                                    optimizer, closure=lambda **kwargs: self._train_batch(microbatches, **kwargs))
+                    while rerun_train_batch:
```
ajaysaini725, line 691: This should be tested on the existing composer benchmarks (see the yaml files in composer/yamls for models to run).

In general this will work in the majority of cases, however there are a few problematic edge cases we want to verify: :
- Sometimes memory usage differs between ranks - will this result in different ranks having different values for  grad_accum? The majority of the time it will not but we want to make sure it doesn't cause problems in cases where we are very close to max memory usage (for example, different processes having different values for grad accum will result in our profiler results being different per process).
- If one process crashes but others do not, will it result in the `loss.backwards()` hanging?


Just as a point of consideration - one other way to implement this is to only have rank0 do the grad_accum search and then broadcast the result to other ranks

jbloxham, line 691: Code style-wise, this is a lot of indentation now. Can you refactor some of the inner code out into a new function?

ajaysaini725 (commented):
Overall looks good - this will likely work in the majority of cases but we want to test it on all existing Composer benchmarks to make sure

Try running this with:
- At least one of ResNet50 or ResNet101
- GPT2 or BERT

Methods that change memory usage during training:
- progressive resizing (vision)
- sequence length warmup (NLP)

Verify that the training results are as expected in the above cases

jbloxham (changes requested):
Left a few comments for now. Note that it's gonna be vital to test this on multiple GPUs.

jbloxham:
Also, could you add a check to raise an error if someone tries to use this when training on CPU?

composer/trainer/trainer.py
mvpatel2000: This is saying if micro batch size is 1 and we get CUDA OOM, give up. I'll make the comment more clear

composer/trainer/trainer.py
mvpatel2000: Yep, will move the whole while loop into a new fn

mvpatel2000:
I'm fine doing -1 to enable dynamic, but one issue is that you can't set a starting point for the search. For example, if I want to use it with a large GPT model, it's going to take 6-8 failures before it ramps up to a good grad accum size. Probably fine since its all at start and not too long...?

jbloxham:
> I'm fine doing -1 to enable dynamic, but one issue is that you can't set a starting point for the search. For example, if I want to use it with a large GPT model, it's going to take 6-8 failures before it ramps up to a good grad accum size. Probably fine since its all at start and not too long...?

Yeah, it's not perfect, but it's a lot more approachable from the user's perspective. I'm OK with not letting the user configure the starting point for now - we can always add it later.

My hope is that dynamic grad accum is going to become the de facto standard - that there will be essentially no reason not to use it. It will take a bit of time to validate that, but if so, we'll probably want to change the API around grad accum anyways.

mvpatel2000:
Updated based on all the comments :) Will now start testing

composer/yamls/models/classify_mnist.yaml
```diff
@@ -41,5 +41,5 @@ dataloader:
   persistent_workers: true
   num_workers: 8
 validate_every_n_epochs: 1
-grad_accum: 1
+grad_accum: -1
```
ajaysaini725, line 44: Can we actually just make grad_accum an `Optional[int]` type in the trainer and have it default to `None` (meaning do autotuning). 

That way we can remove it as a parameter from the YAML files entirely rather than having to specify -1 everywhere.

mvpatel2000: Hm, right now it is an optional int defaulting to -1, so I can just get rid of the -1 in all the yamls. 

I'm personally stylistically in favor of -1 over None -- in general, None seems to specify to me don't do this whereas -1 is closer to saying it's not user specified. I'm fine with it either way though if there's a more consistent practice in the codebase

hanlint: Since this is an experimental feature, I'm wary of making it the default until we've had it bake in the trainer for a bit. So perhaps we should defualt still to `grad_accum = 1`. btw, grad_accum='auto' is also an often-used pattern.

mvpatel2000: > Since this is an experimental feature, I'm wary of making it the default until we've had it bake in the trainer for a bit. So perhaps we should defualt still to `grad_accum = 1`. 

It sounds to me like we all would prefer it default since it's one less thing for a user to think about (if this isn't the case, happy to discuss and switch). In that case, it seems to me you're saying we shouldn't push this as default until we're confident it works. I'm happy to switch it non-default, but it'd be helpful to understand what criteria we need to hit to make it default. My sense is pushing it in dev now will mean we get more internal testing before any release, whereas if it isn't default, it's going to be pretty rare for people to switch it on and we won't actually be more confident with it as time passes. Still picking up best practices with composer development, so feel free to correct me / share more detail :) 

> btw, grad_accum='auto' is also an often-used pattern.

This seems like a cleaner pattern -- will switch unless anyone else has other ideas

mvpatel2000:
Talked w Jamie a bit about deadlocks. TDLR: There is a slight risk of deadlock with multiple GPUs, but it's very low and seems very hard to mitigate

We currently hide DDP latency behind the computation of the backward pass. We deadlock if 1) one GPU finishes the forward and backward pass with no issues but 2) one GPU crashes midway through the backward pass after sending at least some of its gradients to the GPU which finished. In this case, the second GPU will wait on weights from the first one it already received before crashing.

I think this is probably low prob because (to my knowledge, could be wrong) pytorch allocates the memory it needs for graph for backward pass at start. This is also extra low prob if grad_accum > 1 since the graph is already populated at start of last pass, so theoretically, it shouldn’t need more allocation (but of course, theres always random stuff happening). This is probably fine as a transient error since it happens at the start of training (usually, unless ur doing something that increases in memory through training) and u can just restart

At least experimentally, I haven’t been able to trigger this. I ran GPT 2 a few times on 2 GPUs with only 24/4268 MiB free per nvidia-smi, and no issues (see screenshot at top of PR).

We could solve this by not hiding the latency, but this seems like a steep performance cost. I think the best course is to just slap a warning on when someone uses auto. Also, note that this is an issue if people manually set grad_accum as well

composer/yamls/models/classify_mnist.yaml
mvpatel2000: One way to test could be to make it `auto` in config yamls but by default 1 (as before) in the actual codebase, so we get a fair amount of internal testing still

hanlint: Yeah the feature is awesome, but want to let it bake for a bit to flush out any odd issues before changing the defaults. Here's what I suggest:

1. keeping the default on 1 in both the function signature and the YAML configs in `composer` (all public facing). 
2. we run our release testing sweeps with "auto"  + add some additional YAML configs that stress test this feature
3. we make `grad_accum=auto` the default in our internal `mcli` tools so internally we are testing it.

We let it bake for a month or two until we set the default with a major increment. We will likely get some additional feedback on the design in this burn-in period. (e.g. if i _know_ grad accum needs to be around 50, why wait all this latency while the script tried to reset itself?)

mvpatel2000:
I think this is in a good shape and tests seem to be fine. I'd love some input on https://github.com/mosaicml/composer/pull/580, and then I'll merge that here and rerun all tests again

composer/yamls/models/classify_mnist.yaml
mvpatel2000: Got it, thanks for clearing up the process. Will adjust configs accordingly

composer/trainer/trainer.py
```diff
@@ -997,6 +993,79 @@ def _train_loop(self) -> None:
...
+                        # data and if so print a warning that metrics may return unexpected results
+                        outputs, targets = self._original_model.validate(eval_microbatch)
+                        results.append((outputs, targets))
+                    for outputs, targets in results:
+                        train_metrics.update(outputs, targets)
+                rerun = False
+            except RuntimeError as e:
+                self._handle_cuda_oom(e)
+
+    def _train_and_compute_loss(self, use_grad_scaling: bool):
```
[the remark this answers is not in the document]
jbloxham: `_do_batch_fwd_bwd_pass` is a kinda awkward name though, so some other suggestions:
- `_do_batch_pass`
- `_batch_autograd_pass`
- `_compute_batch_loss_and_grad`

composer/trainer/trainer.py
```diff
@@ -997,6 +993,79 @@ def _train_loop(self) -> None:
...
+                    for eval_microbatch in self._train_data_spec.split_batch(self.state.batch, self.state.grad_accum):
+                        # TODO: Detect if self.run_event(Event.AFTER_DATALOADER) changes the training
+                        # data and if so print a warning that metrics may return unexpected results
+                        outputs, targets = self._original_model.validate(eval_microbatch)
+                        results.append((outputs, targets))
+                    for outputs, targets in results:
+                        train_metrics.update(outputs, targets)
+                rerun = False
+            except RuntimeError as e:
+                self._handle_cuda_oom(e)
```
jbloxham, line 1032: We can get by without wrapping this for the time being. It's very unlikely an OOM would be caused under the `torch.no_grad` context.

composer/trainer/trainer.py
```diff
@@ -568,6 +569,22 @@ def __init__(
         if num_optimizers != 1:
             raise NotImplementedError(f"Only one optimizer is supported; found {num_optimizers} optimizers")
 
+        # Set initial grad_accum to 1 if using adaptive
+        self.adaptive_grad_accum = isinstance(grad_accum, str) and grad_accum == "auto"
+        if self.adaptive_grad_accum:
```
jbloxham, line 574: Let's also add a check in here that throws an error if the per-GPU microbatch size is not a power of 2.

composer/trainer/trainer.py
```diff
@@ -1032,8 +1101,7 @@ def _train_batch_inner(self, microbatches: Sequence[Batch]):
         current_batch_size = sum([self._train_data_spec.get_num_samples_in_batch(batch) for batch in microbatches])
 
         for microbatch_idx, self.state.batch in enumerate(microbatches):
-            self.state.batch_num_tokens = self._train_data_spec.get_num_tokens_in_batch(self.state.batch)
-            self.state.batch_num_samples = self._train_data_spec.get_num_samples_in_batch(self.state.batch)
```
jbloxham, line 1036: @ravi-mosaicml is it fine to get rid of these state updates?

composer/trainer/trainer_hparams.py
```diff
@@ -257,10 +257,12 @@ class TrainerHparams(hp.Hparams):
...
 
     # training hparams
-    grad_accum: int = hp.optional(textwrap.dedent("""\
+    grad_accum: Union[int, str] = hp.optional(textwrap.dedent("""\
         Determines the number of microbatches to split a per-gpu batch into,
-        used to compensate for low-memory-capacity devices."""),
-                                  default=1)
+        used to compensate for low-memory-capacity devices. If set to auto, 
+        dynamically increases number of microbatch size if train_batch_size is
+        too large for GPU. Defaults to ``1``"""),
```
jbloxham, line 264: Same comment here as before for this doc.

composer/trainer/trainer_hparams.py
```diff
@@ -453,6 +455,10 @@ def validate(self):
         if self.scale_schedule_ratio <= 0:
             raise ValueError("scale_schedule_ratio must be a positive value.")
 
+        if isinstance(self.grad_accum, str) and self.grad_accum != "auto" or isinstance(self.grad_accum,
+                                                                                        int) and self.grad_accum < 1:
```
jbloxham, line 459: I, for one, have no idea what the relative operational priority of `and` and `or` is. Could you put parens around the `and` expressions?

Unless you already did, and YAPF got rid of them. In which case this is just another reason we should use Black instead of YAPF.

composer/trainer/trainer_hparams.py
```diff
@@ -453,6 +455,10 @@ def validate(self):
...
+            raise ValueError("grad_accum must be ``auto`` or an int greater than or equal to 1")
```
jbloxham, line 460:
```suggestion
            raise ValueError('grad_accum must be "auto" or an int greater than or equal to 1')
```

jbloxham (commented):
Looking very close - well done! Alas, this PR is starting to expose some of the tech debt in the train loop that we'll have to address sooner or later. @hanlint @ravi-mosaicml I left some suggestions on how to approach this; let us know your thoughts!

composer/trainer/trainer.py
mvpatel2000: Why does it have to be a power of 2? I think microbatch size can even be different per GPU (eg say I want a logical batch size of 2001 on 2 GPUs. It seems to work fine right now)

composer/trainer/trainer.py
mvpatel2000: Will leave unresolved for now, but I cleared with him ahead of time. I think the idea was to expose microbatch size for algorithms in the future, but the naming makes it unclear what it is and causes problems with grad_accum so Ravi was fine with removing

composer/trainer/trainer_hparams.py
mvpatel2000: adjusted. What's the stylistic rule here for ' vs " vs. `? I've seen in other parts """ + `` so that's what I tried emulating

composer/trainer/trainer.py
mvpatel2000: This happens before `_train_and_compute_loss()`, so if enabled, I think we'd see OOM here on the very first pass. I agree it's not the bottleneck step because of `torch.no_grad`, but I think it's required because it happens first

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: I agree with renaming -- will wait until others chime in before doing it though. I'm more in favor of `_compute_batch_loss_and_grad` because we're also calling `loss.backward()` and this isn't just a forward pass

composer/trainer/trainer.py
jbloxham: Talked with @ajaysaini725, and I think there's a number of reasons why we shouldn't wrap this:

Since the memory requirements of this step are lower than that of an ordinary training fwd/bwd pass, it doesn't actually make sense to use `state.grad_accum` at all to chunk the eval microbatches here. The optimal grad_accum for use here is different than what should be used for the main training loop.

Because of this, suppose an initial OOM in this step leads to some grad accum tuning to make this function succeed. Then suppose that on the first full train batch size, the grad accum is further increased so that _that_ can pass. That means that when this function next runs, the grad accum will have changed, and it won't need as much memory as it did before. That might sound like a good thing, but it can cause memory fragmentation issues. Furthermore, we'll be running this function less efficiently than we otherwise could (and computing train metrics can sometimes impose as much as a 20% slowdown, so that's a big deal).

Anyways, @ajaysaini725 had the insight that the microbatch size we should use, in theory, should be exactly the same as the per-GPU batch size we use for regular eval. So in theory, the wrapper for this should really be a part of our eventual eval_batch_size tuning, not our grad_accum tuning.

In summary, I think there's a lot of complexity going on here that would best be tackled in a separate PR. Ultimately, that PR isn't urgent, since training metrics just aren't used that frequently by us anyways.

composer/trainer/trainer_hparams.py
jbloxham: The one common rule is that if your string itself contains double quotes or single quotes, you should use the opposite type of quotes to encapsulate the string.

jbloxham: Here, I wanted you to use quotes to surround "auto" just to indicate in the printed message that it represents a string.

composer/trainer/trainer.py
mvpatel2000: Got it, I'm convinced :) 

Will gut the `while try except` part but keep the function decomposition

composer/trainer/trainer_hparams.py
mvpatel2000: Got it, makes sense! Thanks for the explanation

composer/trainer/trainer.py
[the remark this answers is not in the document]
hanlint: Prefer to keep the names shorter if possible. How about:
`_train_and_compute_loss` -> `_train_batch`
`_train_batch` -> `_train_batch_inner`
`_train_batch_inner` -> `_train_microbatches`

composer/trainer/trainer.py
```diff
@@ -156,8 +156,9 @@ class Trainer:
             .. seealso:: :mod:`composer.optim.scheduler` for the different schedulers built into Composer.
         device (str or Device, optional): The device to use for training. Either ``cpu`` or ``gpu``.
             (default: ``cpu``)
-        grad_accum (int, optional): The number of microbatches to split a per-device batch into. Gradients
-            are summed over the microbatches per device. (default: ``1``)
+        grad_accum (Union[int, str], optional): The number of microbatches to split a per-device batch into. Gradients
```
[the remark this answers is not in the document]
mvpatel2000: Originally, it was two parameters. -- I think @jbloxham was in favor of one param with the idea being if you set it to auto, you shouldn't need to reason about batch size at all

composer/trainer/trainer.py
```diff
@@ -997,6 +990,72 @@ def _train_loop(self) -> None:
...
                 self._checkpoint_saver.save_checkpoint(state=self.state, seed=self._seed, device=self._device)
 
+    def _handle_cuda_oom(self, e: RuntimeError):
+        """Handles CUDA Out of Memory and rescales if using adaptive grad_accum."""
+        if self.adaptive_grad_accum and "CUDA out of memory" in str(e):
+            # Raise runtime error if training 1 sample at a time still resulted in CUDA out of memory
+            if self.state.grad_accum == self.state.batch_num_samples:
+                raise RuntimeError("CUDA out of memory. Train loop failed with an internal microbatch of size 1")
+            else:
+                self.state.grad_accum = min(2 * self.state.grad_accum, self.state.batch_num_samples)
```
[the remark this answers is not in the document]
mvpatel2000: Hm... 2 questions:
1) Why is it a problem if different nodes have different sized grad accum?
2) If we set it in a finally block, could this lead to slower speeds? Since now you're forcing sync in between batches

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: I used Hanlin's, except I did `_train_batch -> _train_batch_with_context`. I also decomposed `_train_microbatches` more and added `_train_microbatch`

mvpatel2000:
There's been a lot going on here, so to recap on current situation:

It currently works and is functional! We're throwing a warning for using it, and it's not compatible with certain (just one, sequence length warmup) algorithms. 

Theoretically, we can also run into issues of deadlock if just 1 rank OOMs. Ravi and I discussed this last week, and there's some ideas on how to propagate errors across ranks. This feels like a more useful thing in general, so I'll be playing around with that and possibly generalizing to a decorator we might use more broadly.

hanlint:
OK great, I think we are close to getting this merged. I re-requested some reviews. Given the scope of this refactor of the trainer loop itself, we should also touch off some regression runs. Right before merge, let's run a full sweep of regressions.

ravi-mosaicml (commented):
Took a read through and left some comments regarding mostly usability. I didn't completely review it as it looks like there was a messy training loop merge (see some of the comments). Let me know when the training loop is cleaned up, and I'll take another read through.

In addition, I'm curious about our whiteboard discussion re: handling OOMs across processes and synchronizing 1) whether there was a cuda oom error, and 2) the grad accum value across states? Mainly concerned about the rare-but-nearly-impossible-to-debug edge case where rank zero cuda OOMs but other ranks continue fine, and then the processes are no longer synchronized. (@jbloxham can attest to how difficult it is to track distributed errors that occur on one rank but not another)

composer/trainer/trainer.py
```diff
@@ -1083,7 +1076,74 @@ def _train_loop(self) -> None:
...
+
+    def _train_batch(self, use_grad_scaling: bool):
+        """Compute loss by training on a full batch of data. Adaptively change microbatch size if enabled to maximize
+        GPU usage.
+
+        Args:
+            use_grad_scaling (bool): Enables gradient scaling
+        """
+        rerun = True
+        while rerun:
```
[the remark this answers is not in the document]
mvpatel2000: Will rewrite to while True. Recursive is messy because it blows up the call stack to be very deep

composer/trainer/trainer.py
```diff
@@ -1118,78 +1182,89 @@ def _train_batch_inner(self, microbatches: Sequence[Batch]):
...
+            is_final_microbatch (bool): If current microbatch is the last one.
+        """
+        assert self.state.scaler is not None
 
-                # We always want to scale loss by the grad_accum before the backwards pass and
-                # also for sake of metrics. Complicating matters, the DeepSpeed engine does its
-                # own scaling when we call `.backward`, but this isn't in place so we still need
-                # to scale for sake of metrics after the `.backward` call.
+        microbatch_num_samples = self._train_data_spec.get_num_samples_in_batch(self.state.batch)
+        sync_context = contextlib.nullcontext() if self.deepspeed_enabled else _ddp_sync_context(
```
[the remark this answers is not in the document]
mvpatel2000: Can you give some motivation for why you think it's helpful? It seems a bit odd to me to refactor into a function a one line statement

composer/trainer/trainer.py
```diff
@@ -997,6 +990,72 @@ def _train_loop(self) -> None:
...
+                    total_loss = self._train_batch(microbatches)
+                    for optimizer in self.state.optimizers:
+                        if use_grad_scaling:
+                            self.state.scaler.step(optimizer)
+                        else:
+                            optimizer.step()
+                rerun = False
+                return total_loss
+            except RuntimeError as e:
+                self._handle_cuda_oom(e)
```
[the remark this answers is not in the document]
mvpatel2000: Ok hm this is actually necessary for other parts, so I'll go ahead and do it

composer/trainer/trainer.py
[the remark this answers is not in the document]
hanlint: Yeah agreed, probably OK to leave as-is unless its used elsewhere?

mvpatel2000:
@ravi-mosaicml rerequesting review! I've wrapped the train loop to sync if any rank fails and rerun. I propose throwing out any progress and restarting the batch if anything fails. In the scenario where only one rank fails, the rest of the ranks either 
1. successfully complete and have to redo some work (which is fine since it only happens once) or
2. they throw an error (say one rank errors, so everyone else times out at a barrier because it never reaches a gradient sync), which is caught. At the end, we see another rank OOMed, so we retry training on all ranks after updating grad_accum

mvpatel2000:
Ran some tests to confirm results are identical. With resnet 50 on imagenet 8 GPUs, we get basically the same curves (theyre overalpping) when comparing
- dev vs. this branch + no auto grad_accum vs. this branch + auto grad_accum
- same as above except including blurpool + label smoothing + mixup + channels last

Note that grad_accum is set to 2 since it was run on coda, so the auto grad_accum did actually have to activate and change grad_accum

image

composer/trainer/trainer.py
```diff
@@ -1106,98 +1182,108 @@ def _train_batch(self, microbatches: Sequence[Batch], ddp_sync: bool = True):
...
-        if not self.deepspeed_enabled and self._grad_clip_norm is not None:
-            torch.nn.utils.clip_grad_norm_(
-                parameters=self.state.model.parameters(),
-                max_norm=self._grad_clip_norm,
-            )
+                self.state.deepspeed_model.backward(self.state.loss)
+
+                # This is the same loss scaling and reporting we skipped earlier.
+                for loss in ensure_tuple(self.state.loss):
+                    loss.mul_(self.state.batch_num_samples / current_batch_size)
```
ravi-mosaicml, line 1277: is `batch_num_samples` still being updated on state? I see that old lines 1129 and 1130 were removed.
I'd be OK with making the num samples and num tokens local to only the trainer, in which case could you remove these variables from the state?

ravi-mosaicml (commented):
Overall LGTM! Just a few easy to fix comments. Then it should be good to merge.

composer/trainer/trainer.py
```diff
@@ -1091,8 +1081,94 @@ def _train_loop(self) -> None:
...
+                    for optimizer in self.state.optimizers:
+                        if use_grad_scaling:
+                            self.state.scaler.step(optimizer)
+                        else:
+                            optimizer.step()
+            except RuntimeError as e:
+                if self._is_cuda_oom(e):
+                    should_handle_cuda_oom = True
+                else:
+                    caught_error = e
```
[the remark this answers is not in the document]
mvpatel2000: Here's the scenario I'm thinking about: 
1. rank 0 cuda OOMs. `should_handle_cuda_oom` is set to True
2. the sync strategy is set to sync every microbatch. Rank 1 does not crash and waits for 0 at end of microbatch. Eventually, it times out (since 0 never reaches there) and throws an error)
3. Rank 1 has `catch_error` set to be `e`, which is something related to a timeout at a barrier. It ignores this since someone else got an OOM (the root cause of the issue) and everyone reruns with higher `grad_accum`

composer/trainer/trainer.py
```diff
@@ -1091,8 +1081,94 @@ def _train_loop(self) -> None:
...
+                            total_loss = optimizer.step(
+                                closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs).item())
+                else:
+                    total_loss = self._train_microbatches(microbatches)
+                    for optimizer in self.state.optimizers:
+                        if use_grad_scaling:
+                            self.state.scaler.step(optimizer)
+                        else:
+                            optimizer.step()
+            except RuntimeError as e:
```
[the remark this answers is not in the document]
mvpatel2000: I'm not super familiar with this part, so correct me if I'm wrong. My understanding is it would hit a timeout (set with `init_process_group`) and surface an error. When this error is surfaced, it's caught across all ranks. We then hit the `except` here and recognize it was really driven by an OOM, which we appropriately handle. So, the situation is still handled

composer/trainer/trainer.py
mvpatel2000: It's still set each iteration and used in a few places. We could pass it around as parameters, but imo fine to leave in `state` -- still "feels" right to be part of state

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: As a side, I'm fine adding logging, but note that it will fire every time we hit an auto `grad_accum` update. Personally think it's better to not log that often with potential issues since it drowns out stuff.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Ah, I see. In this case, can we explicitly catch the timeout errors?

composer/trainer/trainer.py
ravi-mosaicml: I must be blind, but which lines is it being updated on?

ravi-mosaicml: Ah yes, I see on lines 1002 and 1003.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Ah yes -- I missed the logic with the timeout. If rank 1 times out, then yup -- agreed it would hit here, and handle it appropriately.

Maybe could you add this as a `log.info` or `log.debug`? I think by default python only prints `log.warn`, but if a user runs into an issue, then we'll ask them to include the log output when running on debug mode, so it would be helpful to include this in the stack trace. I don't think there will be too many grad accum updates -- just a few at the beginning of training?

(In general, we need to add more logging statements to the training loop! Don't have enough IMO).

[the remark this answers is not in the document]
mvpatel2000: Got it -- will add more logging :)

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: yes, very good point, will definitely add a few comments to make this more clear because its subtle

ravi-mosaicml (approved):
Lgtm

jbloxham (approved):
This looks great now! Thanks a ton for dealing with all of our comments - I'm really excited to see this get some use :)

hanlint (approved):

PR #486, opened 2022-02-16
Remove colons from run directory timestamps

### LINKED ISSUE #442: Checkpoints stored in run directory cannot have ":" in their name
By default, the run directory contains a timestamp with ":" in its name. If the run directory is uploaded to an object store, any checkpoints within it cannot be used by `CheckpointLoader` because it parses the `load_path` as a URL: https://github.com/mosaicml/composer/blob/dev/composer/trainer/checkpoint.py#L93-L99

ravi-mosaicml:
> By default, the run directory contains a timestamp with ":" in its name. If the run directory is uploaded to an object store, any checkpoints within it cannot be used by CheckpointLoader because it parses the load_path as a URL: https://github.com/mosaicml/composer/blob/dev/composer/trainer/checkpoint.py#L93-L99

Closes #442

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -60,7 +60,7 @@ def get_parser():
                         default=None,
                         help=textwrap.dedent("""\
                             Directory to store run artifcats. 
-                            Defaults to runs/{datetime.datetime.now().isoformat()}/""")),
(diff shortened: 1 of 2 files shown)


Averylamp (approved):
lgtm

Issue #480, opened 2022-02-15
Docs cleanup


ravi-mosaicml:
- [ ] Pin the sidebar so it automatically scrolls to the currently open page
- [x] Remove documenting of inherited members that don't have new docstrings (addressed in #488)
- [x] Show inheritance and base classes (addressed in #488)
- [x] Ensure that all doctests are passing on jenkins (addressed in #488)
- [ ] Skip documentation of modules that contain no public members

ravi-mosaicml:
Closing for #498 

Not going to address skipping documentation of modules with no public members -- those should be renamed to be private members.

PR #488, opened 2022-02-16
Ravi/docs cleanup 2

ravi-mosaicml:
Addresses some issues in #480:

> Pin the sidebar so it automatically scrolls to the currently open page

Not easily do-able; this will require custom javascript.

> Remove documenting of inherited members that don't have new docstrings

Done; removed the inherited members flag

> Show inheritance and base classes

Done; added the show base classes flag

> Ensure that all doctests are running on jenkins

Fixed; see https://ci.int.mosaicml.com/job/scratch/job/command/5280/console

> Skip documentation of modules that contain no public members

Cannot be done easily; prototyped different solutions which used autodoc events / hooks; these were too flaky. The recommended approach is to make the module private.

> Skip documentation of members that are explicitly included in `__all__` in parent modules

Done; still using `ignore-module-all` for documentation.


docs/source/conf.py
```diff
@@ -72,17 +71,6 @@
 #
 html_theme = "furo"
 
-html_theme_options = {
```
ravi-mosaicml, line 75: This dictionary was being overridden on line 103 below

bandish-shah (approved):
LGTM

PR #489, opened 2022-02-16
Add C4 Streaming dataset

abhi-mosaic:
This PR adds an adapter for HF streaming datasets, so that we can train with DDP and sharding. There is also a small QoL fix for DeepSpeed.

I've also added a new GPT3-125m YAML which can be the start of our new GPT3 family of benchmarks. I think we should deprecate the GPT2 collection of YAMLs (-52m, -83m -125m) in the near future, say Composer v0.5.

Some TODOs, which can be follow-up PRs:

* Add the rest of the GPT3 family YAMLs
* Update the GPT2 model with QoL fixes
* Update the BERT-base benchmark to use C4
* Deprecate / combine the old `LMDataset` object

### DIFF
--- a/composer/datasets/c4.py
+++ b/composer/datasets/c4.py
@@ -1,0 +1,360 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
+
+import logging
+from dataclasses import dataclass
+from functools import partial
+from itertools import chain
+from typing import List
+
+import yahp as hp
+from torch.utils.data import IterableDataset, get_worker_info
+
+from composer.core.types import Batch, DataSpec
+from composer.datasets.dataloader import DataloaderHparams
+from composer.datasets.hparams import DatasetHparams
+from composer.utils import dist
+
+log = logging.getLogger(__name__)
+
+__all__ = ["C4Dataset", "C4DatasetHparams"]
+
+
+def _split_dict_fn(batch: Batch, n_microbatches: int) -> List[Batch]:
+    if isinstance(batch, dict):
+        chunked = {k: v.chunk(n_microbatches) for k, v in batch.items()}
+        for k, v in chunked.items():
+            if len(v) != n_microbatches:
+                raise ValueError(
+                    f"Unable to split batch into microbatches. "
+                    f"Key '{k}' has chunked list: {v} with length {len(v)}, but expected length {n_microbatches}. ")
+        microbatches = []
+        for idx in range(n_microbatches):
+            mb = {k: v[idx] for k, v in chunked.items()}
+            microbatches.append(mb)
+        return microbatches
+    else:
+        raise ValueError(f'Expected batch to be of type Dict[str, Tensor], but got {type(batch)}')
+
+
+@dataclass
+class C4DatasetHparams(DatasetHparams):
+    """Builds a DataSpec for the C4 (Colossal Cleaned CommonCrawl) dataset.
+
+    Parameters:
+        split (str): What split of the dataset to use. Either `train` or `validation`.
+        max_samples (int): Max number of post-processed token samples, used to set epoch size of the IterableDataset.
+        tokenizer_name (str): The name of the HuggingFace tokenizer to preprocess text with.
+        max_seq_len (int): The max sequence length of each token sample.
+        group_method (str): How to group text samples into token samples. Either `truncate` or `concat`.
+        mlm (bool): Whether or not to use masked language modeling. (Default: `False`)
+        mlm_probability (float): If `mlm=True`, the probability that tokens are masked. (Default: `0.15`)
+        shuffle (bool): Whether to shuffle the samples in the dataset. Currently, shards are assigned and consumed with deterministic per-device shard order, but shuffling affects the order of samples via (per-device) shuffle buffers. (Default: `False`)
+        shuffle_buffer_size (int): If `shuffle=True`, samples are read into a buffer of this size (per-device), and randomly sampled from there to produce shuffled samples. (Default: `10000`)
+        seed (int): If `shuffle=True`, what seed to use for shuffling operations. (Default: `5`)
+        drop_last (bool): Whether to drop the last samples for the last batch. (Default: `True`)
+    Returns:
+        A :class:`~composer.core.DataSpec` object
+    """
+
+    split: str = hp.optional("What split of the dataset to use. Either `train` or `validation`.", default=None)
+    max_samples: int = hp.optional(
+        "Max number of post-processed token samples, used to set epoch size of the IterableDataset.", default=None)
+    tokenizer_name: str = hp.optional("The name of the HuggingFace tokenizer to preprocess text with.", default=None)
+    max_seq_len: int = hp.optional("The max sequence length of each token sample.", default=None)
+    group_method: str = hp.optional("How to group text samples into token samples. Either `truncate` or `concat`.",
+                                    default=None)
+    mlm: bool = hp.optional("Whether or not to use masked language modeling.", default=False)
+    mlm_probability: float = hp.optional("If `mlm=True`, the probability that tokens are masked.", default=0.15)
+    shuffle: bool = hp.optional(
+        "Whether to shuffle the samples in the dataset. Currently, shards are assigned and consumed with deterministic per-device shard order, but shuffling affects the order of samples via (per-device) shuffle buffers.",
+        default=True)
+    shuffle_buffer_size: int = hp.optional(
+        "If `shuffle=True`, samples are read into a buffer of this size (per-device), and randomly sampled from there to produce shuffled samples.",
+        default=10000)
+    seed: int = hp.optional("If `shuffle=True`, what seed to use for shuffling operations.", default=5)
+    drop_last: bool = hp.optional("Whether to drop the last samples for the last batch.", default=True)
+
+    def validate(self):
+        if self.split not in ["train", "validation"]:
+            raise ValueError(f"Unknown split: '{self.split}'")
+        if self.max_samples is None or self.max_samples <= 0:
+            raise ValueError(f"Must provide 'max_samples' > 0")
+        if self.tokenizer_name is None:
+            raise ValueError(f"Must provide 'tokenizer_name'")
+        if self.max_seq_len is None or self.max_seq_len <= 0:
+            raise ValueError(f"Must provide 'max_seq_len' > 0")
+        if self.group_method not in ["truncate", "concat"]:
+            raise ValueError(f"Unknown group_method: '{self.group_method}'. Must be 'truncate' or 'concat'")
+        if self.mlm and self.mlm_probability <= 0:
+            raise ValueError("Must provide a positive 'mlm_probability' when using masked language modeling.")
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams) -> DataSpec:
+        try:
+            import transformers
+        except ImportError:
+            raise ImportError('HuggingFace transformers not installed. '
+                              'Please install with `pip install composer[nlp]`')
+
+        # Get C4 dataset
+        c4_dataset = C4Dataset(split=self.split,
+                               max_samples=self.max_samples,
+                               tokenizer_name=self.tokenizer_name,
+                               max_seq_len=self.max_seq_len,
+                               group_method=self.group_method,
+                               shuffle=self.shuffle,
+                               shuffle_buffer_size=self.shuffle_buffer_size,
+                               seed=self.seed)
+
+        # Get collate_fn
+        collate_fn = transformers.DataCollatorForLanguageModeling(tokenizer=c4_dataset.tokenizer,
+                                                                  mlm=self.mlm,
+                                                                  mlm_probability=self.mlm_probability)
+        # Return DataSpec
+        return DataSpec(dataloader=dataloader_hparams.initialize_object(
+            dataset=c4_dataset,
+            batch_size=batch_size,
+            sampler=None,
+            drop_last=self.drop_last,
+            collate_fn=collate_fn,
+        ),
+                        split_batch=_split_dict_fn)
+
+
+class C4Dataset(IterableDataset):
+    """Builds a streaming, sharded, sized IterableDataset for the C4 (Colossal Cleaned CommonCrawl) dataset. Used for
+    pretraining autoregressive or masked language models. Text samples are streamed directly from the cloud using
+    HuggingFace's C4 Dataset with streaming backend (See https://huggingface.co/datasets/c4 for more details). The text
+    samples are then shuffled, tokenized, and grouped on-the-fly.
+
+    Parameters:
+        split (str): What split of the dataset to use. Either `train` or `validation`.
+        max_samples (int): Max number of post-processed token samples, used to set epoch size of the IterableDataset.
+        tokenizer_name (str): The name of the HuggingFace tokenizer to preprocess text with.
+        max_seq_len (int): The max sequence length of each token sample.
+        group_method (str): How to group text samples into token samples. Either `truncate` or `concat`.
+        shuffle (bool): Whether to shuffle the samples in the dataset. Currently, shards are assigned and consumed with deterministic per-device shard order, but shuffling affects the order of samples via (per-device) shuffle buffers. (Default: `False`)
+        shuffle_buffer_size (int): If `shuffle=True`, samples are read into a buffer of this size (per-device), and randomly sampled from there to produce shuffled samples. (Default: `10000`)
+        seed (int): If `shuffle=True`, what seed to use for shuffling operations. (Default: `5`)
+    Returns:
+        A :class:`torch.utils.data.IterableDataset` object
+    """
+
+    def __init__(self,
+                 split,
+                 max_samples,
+                 tokenizer_name,
+                 max_seq_len,
+                 group_method,
+                 shuffle=False,
+                 shuffle_buffer_size=10000,
+                 seed=5):
+        try:
+            import datasets
+            import transformers
+        except ImportError:
+            raise ImportError('HuggingFace transformers and datasets are not installed. '
+                              'Please install with `pip install composer[nlp]`')
+
+        self.split = split
+        self.max_samples = max_samples
+        self.tokenizer_name = tokenizer_name
+        self.max_seq_len = max_seq_len
+        self.group_method = group_method
+        self.shuffle = shuffle
+        self.shuffle_buffer_size = shuffle_buffer_size
+        self.seed = seed
+
+        # Metadata
+        c4_metadata = {
+            "train": {
+                "num_shards": 1024,
+                "approx_samples_per_shard": 356317,
+            },
+            "validation": {
+                "num_shards": 8,
+                "approx_samples_per_shard": 45576,
+            }
+        }
+        if self.split in c4_metadata:
+            self.num_shards = c4_metadata[self.split]["num_shards"]
+            self.approx_samples_per_shard = c4_metadata[self.split]["approx_samples_per_shard"]
+        else:
+            raise ValueError(f"Unknown split={self.split}, expected one of {list(c4_metadata.keys())}.")
+
+        # Set dataset size
+        self.world_size = dist.get_world_size()
+        self.rank = dist.get_global_rank()
+        self.max_samples_per_device = self.max_samples // self.world_size
+        if self.max_samples % self.world_size != 0:
+            new_max_samples = self.max_samples_per_device * self.world_size
+            log.warning(
+                f"Max samples will be truncated from {max_samples}->{new_max_samples} to maintain divisibility across {self.world_size} devices."
+            )
+            self.max_samples = new_max_samples
+
+        # Try and detect if max_samples is too large
+        # TODO: For safety, repeat original dataset if max_samples is larger than original size
+        original_approx_samples = self.num_shards * self.approx_samples_per_shard
+        if self.max_samples > original_approx_samples and self.group_method == "truncate":
+            raise ValueError(
+                f"Max samples was set to {self.max_samples} with group_method 'truncate' but split '{split}' has only {original_approx_samples}.)"
+            )
+        if self.group_method == "concat":
+            log.warning(
+                f"When using group_method 'concat', sequential token samples are concatenated and chunked into fixed-length samples of size max_seq_len={self.max_seq_len}. "
+                f"In general we cannot detect ahead-of-time if your setting of max_samples={self.max_samples} is safe, "
+                f"so please be cautious to not request more tokens than the size of the original dataset.")
+
+        # Build tokenizer
+        self.tokenizer = transformers.AutoTokenizer.from_pretrained(tokenizer_name)
+        if self.tokenizer.pad_token is None:
+            # Some tokenizers (e.g. GPT2 tokenizer) have no padding token which causes bugs
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+
+        # Load and shard dataset
+        text_dataset = datasets.load_dataset(path="c4", name="en", split=split, streaming=True)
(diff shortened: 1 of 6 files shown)


composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,246 @@
...
+            result = {
+                k: [v[i:i + self.max_seq_len] for i in range(0, num_tokens, self.max_seq_len)
+                   ] for k, v in concat_tokens.items()
+            }
+            result["labels"] = result["input_ids"].copy()
+            return result
+        else:
+            raise ValueError(f"Unknown group_method: '{self.group_method}'")
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams) -> DataSpec:
```
hanlint, line 167: is there a way we can move all of this logic into a separate class? 

the motivation being that external users not using YAHP should be able to create their StreamingDataset and pass it to our trainer directly?

abhi-mosaic: Yes, I could definitely move some of the first few steps like sharding, and the overwritten __len__ into a class, but if users want to build their own custom dataloader, they would still have to build the collate functions and pass those correctly to their dataloader right? I'm also wondering if the preprocessing steps like (shuffle, tokenize, group) would belong in the class, or if again it is user-specific. I guess I'm skeptical how much I can abstract away, haha.

Looking at Vision, I think there are classes like `ImageFolder(..) -> Dataset` that you can start from, but building a dataloader is still pretty task-specific.

For users that want to build their own streaming dataloaders, I think the upcoming Vision Webdatasets PR would be the better example to follow, and that PR comes with helper classes too. 

WDYT? I'm happy to take a pass at making a light HFStreamingDataset class, or punt it to a future PR?

hanlint: Update: we discussed offline, and will refactor a standalone dataset class for say notebook or script users.

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+from composer.datasets.hparams import DatasetHparams
+from composer.utils import dist
+
+log = logging.getLogger(__name__)
+
+
+def _split_dict_fn(batch: Batch, n_microbatches: int) -> List[Batch]:
+    if isinstance(batch, dict):
+        chunked = {k: v.chunk(n_microbatches) for k, v in batch.items()}
+        num_chunks = len(list(chunked.values())[0])
```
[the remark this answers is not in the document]
abhi-mosaic: Apparently `torch.chunk` is not guaranteed to return the number, if for some reason we request more chunks than the length of the list. I'm not sure if we have any asserts anywhere for this... but I can throw a ValueError if that makes sense?

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+
+@dataclass
+class HFStreamingDatasetHparams(DatasetHparams):
+    """Defines a generic dataset class for autoregressive and masked language models."""
+
+    name: str = hp.optional("Name of the dataset to load.", default=None)
+    config_name: Optional[str] = hp.optional(
+        "If required, the specific configuration of the dataset that you would like to use.", default=None)
+    split: str = hp.optional("What split of the dataset to use (e.g. 'train' or 'validation' or 'test')", default=None)
+    max_shards: int = hp.optional("Max number of shards, used to deterministically reduce dataset size.", default=-1)
```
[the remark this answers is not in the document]
abhi-mosaic: I think `max_shards` can be deleted as we're converting this to a C4-specific dataset class.

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+    max_seq_len: int = hp.optional("The max sequence length of each token sample.", default=None)
+    group_method: str = hp.optional("How to group text samples into token samples.", default=None)
+    use_masked_lm: bool = hp.optional("Whether the dataset shoud be encoded with masked language modeling or not.",
+                                      default=None)
+    mlm_probability: float = hp.optional("If using masked language modeling, the probability to mask tokens with.",
+                                         default=0.15)
+    shuffle_buffer_size: int = hp.optional("How many incoming data samples should be kept in a buffer to shuffle.",
+                                           default=10000)
+    seed: int = hp.optional("Which seed to use to generate train and validation splits.", default=5)
+    shuffle: bool = hp.optional("Whether to shuffle the dataset for each epoch.", default=True)
```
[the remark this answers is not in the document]
abhi-mosaic: Yep! Will fix.

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+    name: str = hp.optional("Name of the dataset to load.", default=None)
+    config_name: Optional[str] = hp.optional(
+        "If required, the specific configuration of the dataset that you would like to use.", default=None)
+    split: str = hp.optional("What split of the dataset to use (e.g. 'train' or 'validation' or 'test')", default=None)
+    max_shards: int = hp.optional("Max number of shards, used to deterministically reduce dataset size.", default=-1)
+    max_samples: int = hp.optional(
+        "Max number of post-processed samples, note that the subset will depend on seed and world size.", default=-1)
+    tokenizer_name: str = hp.optional("The name of the tokenizer to preprocess text with.", default=None)
+    max_seq_len: int = hp.optional("The max sequence length of each token sample.", default=None)
+    group_method: str = hp.optional("How to group text samples into token samples.", default=None)
```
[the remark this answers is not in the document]
abhi-mosaic: Yes, will add a list of choices. The choices are not HF-related, it's just two modes that we have defined for ourselves to determine how to treat the incoming stream of text samples (merge samples to produce full max_sequence_length samples, or keep samples separate and pad)

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+        try:
+            if self.max_samples > 0:
+                return self.max_samples // dist.get_world_size()
+            else:
+                key_config_name = self.config_name if self.config_name else "default"
+                n_shards, samples_per_shard = CACHED_DATASET_SIZES[self.name][key_config_name][self.split]
+                n_shards = self.max_shards if self.max_shards > 0 else n_shards
+                return n_shards * samples_per_shard // dist.get_world_size()
+        except:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
abhi-mosaic: Will refactor this so it is C4 specific.

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+
+CACHED_DATASET_SIZES = {"c4": {"en": {"train": (1024, 356317), "validation": (8, 45576)}}}
+
+
+@dataclass
+class HFStreamingDatasetHparams(DatasetHparams):
+    """Defines a generic dataset class for autoregressive and masked language models."""
+
+    name: str = hp.optional("Name of the dataset to load.", default=None)
+    config_name: Optional[str] = hp.optional(
```
ravi-mosaicml, line 37: Can you give examples or link to HF docs for what a configuration name would be?

composer/datasets/hf_streaming.py
ravi-mosaicml, line 40: What would -1 mean?

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+    max_samples: int = hp.optional(
+        "Max number of post-processed samples, note that the subset will depend on seed and world size.", default=-1)
+    tokenizer_name: str = hp.optional("The name of the tokenizer to preprocess text with.", default=None)
+    max_seq_len: int = hp.optional("The max sequence length of each token sample.", default=None)
+    group_method: str = hp.optional("How to group text samples into token samples.", default=None)
+    use_masked_lm: bool = hp.optional("Whether the dataset shoud be encoded with masked language modeling or not.",
+                                      default=None)
+    mlm_probability: float = hp.optional("If using masked language modeling, the probability to mask tokens with.",
+                                         default=0.15)
+    shuffle_buffer_size: int = hp.optional("How many incoming data samples should be kept in a buffer to shuffle.",
```
ravi-mosaicml, line 50: Can you add in saying this parameter has an effect only if shuffle is True?

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+                                           default=10000)
+    seed: int = hp.optional("Which seed to use to generate train and validation splits.", default=5)
+    shuffle: bool = hp.optional("Whether to shuffle the dataset for each epoch.", default=True)
+    drop_last: bool = hp.optional("Whether to drop the last samples for the last batch.", default=False)
+
+    def validate(self):
+        assert self.name is not None, "Must provide a dataset name."
```
ravi-mosaicml, line 57: For validation, can you using ValueErrors? assert should not be used for data validation (see https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#4-use-of-assert)

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHparams) -> DataSpec:
+        assert dataloader_hparams.num_workers == 1, "LM Streaming Dataloader only supports num_workers=1"
+
+        try:
+            import datasets
+            import transformers
+        except ImportError:
+            raise ImportError('HuggingFace transformers and datasets are not installed. '
+                              'Please install with `pip install mosaicml-composer[nlp]`')
```
ravi-mosaicml, line 171:
```suggestion
                              'Please install with `pip install composer[nlp]`')
```

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+        self.tokenizer = transformers.AutoTokenizer.from_pretrained(self.tokenizer_name)  #type: ignore (thirdparty)
```
ravi-mosaicml, line 172: What is the type error?

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+        self.config = transformers.AutoConfig.from_pretrained(self.tokenizer_name)  #type: ignore (thirdparty)
```
ravi-mosaicml, line 173: Same

composer/datasets/hf_streaming.py
```diff
@@ -0,0 +1,243 @@
...
+        if self.tokenizer_name in ["gpt2"]:
+            # Annoying but GPT2 tokenizer has no padding token which causes bugs
+            collate_fn = transformers.default_data_collator
+        else:
+            collate_fn = transformers.DataCollatorForLanguageModeling(tokenizer=self.tokenizer,
+                                                                      mlm=self.use_masked_lm,
+                                                                      mlm_probability=self.mlm_probability)
+        # Return DataSpec
+        return DataSpec(dataloader=dataloader_hparams.initialize_object(
+            dataset=sized_iterable_dataset,
```
ravi-mosaicml, line 224: Can you pass num_samples and num_tokens into the DataSpec?

composer/trainer/trainer_hparams.py
```diff
@@ -277,38 +277,39 @@ class TrainerHparams(hp.Hparams):
...
+        Estimate flops for operators.
         Ignored if ``torch_profiler_trace_dir`` and `profiler_trace_file` are not specified."""),
                                               default=True)
 
     def validate(self):
         super().validate()
 
         if self.deepspeed is not None:
-            zero_stage = cast(int, self.deepspeed.get("zero_stage", 0))
+            self.deepspeed["zero_stage"] = cast(int, self.deepspeed.get("zero_stage", 0))
```
ravi-mosaicml, line 309: Are these casts necessary?

abhi-mosaic: Probably not, carried over from previous code. Will remove and see what happens.

composer/datasets/hf_streaming.py
abhi-mosaic: I'm not sure if we are quite ready yet... there's still some discussion with NLP team as to what constitutes a sample / a token when there is padding and concatenation involved. I can bring it up with relevant team members and make a follow up PR.

composer/datasets/hf_streaming.py
abhi-mosaic: Code leftover from previous implementation, removing the ignore.

composer/datasets/hf_streaming.py
abhi-mosaic: Removing `max_shards` as we are making this example C4 specific.

composer/datasets/hf_streaming.py
abhi-mosaic: Removing configuration name for now.

composer/trainer/trainer_hparams.py
abhi-mosaic: Actually turns out they are... because the `self.deepspeed` config is of type `[str, JSON]`, and so the values we want to default have to be cast to int before the validation that happens below.

abhi-mosaic:
This PR has been refactored, so instead of a general HF streaming adapter (which is a bit unfeasible, given that HF datasets are user-generated and do not have a consistent implementation of shards or sample dicts), this PR now adds a specific `C4Dataset` that happens to be backed by HF Datasets.

The intent is for `C4Dataset` to be an example for users who want to train with their own HF datasets, but they will have to construct their `CustomDataset` class independently. This is similar to how our `ImageNet` dataset is an example of a vision dataset that happens to be backed by `torchvision.datasets.ImageFolder`.

In the future, we can try to abstract some things away, and/or coordinate with HF upstream to build better sharding support, but right now `C4Dataset` is just a standalone example.

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+        self.split = split
+        self.max_samples = max_samples
+        self.tokenizer_name = tokenizer_name
+        self.max_seq_len = max_seq_len
+        self.group_method = group_method
+        self.shuffle = shuffle
+        self.shuffle_buffer_size = shuffle_buffer_size
+        self.seed = seed
+
+        # Metadata
```
hanlint, line 165: just to check, this is specific for C4 correct?

can you encode it in a single `dict`? e.g.
```
c4_metadata = {"train": {"num_shards" = 1024, .. }, "validation": {"num_shards": 8, ..}}
```

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+            self.approx_samples_per_shard = 45576
+        else:
+            raise ValueError(f"Unknown {split=}, expected 'train' or 'validation'.")
+
+        # Set dataset size
+        self.world_size = dist.get_world_size()
+        self.rank = dist.get_global_rank()
+        self.max_samples_per_device = self.max_samples // self.world_size
+        new_max_samples = self.max_samples_per_device * self.world_size
+        if new_max_samples != self.max_samples:
```
hanlint, line 180: might be more readable to change this to test for divisibility directly:
```
if self.max_samples % self.world_size != 0:
     warning!
     self.max_samples = self.max_samples_per_device * self.world_size
```

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+            truncation = True
+            padding = 'max_length'
+            max_length = self.max_seq_len
+        else:
+            truncation = False
+            padding = False
+            max_length = None
+        return self.tokenizer(text_batch["text"], truncation=truncation, padding=padding, max_length=max_length)
+
+    def _group_tokens(self, token_batch):
```
hanlint, line 305: as some docs for what this is actually doing? input/outputs etc?

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+    def __len__(self):
+        return self.max_samples_per_device
+
+    def _subsample(self, device_offset, text_batch):
+        # Only return the i-th item out of N sequential items
+        for k, v in text_batch.items():
+            text_batch[k] = v[device_offset:device_offset + 1]
+        return text_batch
+
+    def _shard_dataset(self, dataset):
```
hanlint, line 258: even those these are private functions, i think we should have good docstrings because:
1) any step of the data preprocessing is important for users to understand
2) users may want to extend this to other datasets themselves

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+        # Determine how to allocate devices to shards
+        devices_per_shard = 1
+        if self.num_shards < self.world_size:
+            log.warning(
+                f"Not enough unique shards ({self.num_shards}) for world size ({self.world_size}). Splitting shards among devices."
+            )
+            if self.world_size % self.num_shards != 0:
+                raise ValueError(f"Cannot evenly split {self.num_shards} shards among {self.world_size} devices")
+            devices_per_shard = self.world_size // self.num_shards
+        elif self.num_shards % self.world_size != 0:
```
hanlint, line 273: since `num_shards` is defined within the functioni (e.g. `1024`), does that mean that using the streaming dataset with 3 or 6 or 7 GPUs is not supported?

hanlint (commented):
Did mostly a code style/ readability review. Will have to rely on more knowledgable (e.g. @moinnadeem ) to check the functional correctness.

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+    max_samples: int = hp.optional(
+        "Max number of post-processed token samples, used to set epoch size of the IterableDataset.", default=None)
+    tokenizer_name: str = hp.optional("The name of the HuggingFace tokenizer to preprocess text with.", default=None)
+    max_seq_len: int = hp.optional("The max sequence length of each token sample.", default=None)
+    group_method: str = hp.optional("How to group text samples into token samples. Either `truncate` or `concat`.",
+                                    default=None)
+    mlm: bool = hp.optional("Whether or not to use masked language modeling.", default=False)
+    mlm_probability: float = hp.optional("If `mlm=True`, the probability that tokens are masked.", default=0.15)
+    shuffle: bool = hp.optional(
+        "Whether to shuffle the samples in the dataset. Currently, shards are assigned deterministically to devices, but shuffling affects the order of shards (per-device), and the order of samples via (per-device) shuffle buffers.",
```
moinnadeem, line 67: Why would shuffling affect the order of the shards, rather than just shuffle within a shard?

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+            )
+        if self.group_method == "concat":
+            log.warning(
+                f"When using group_method 'concat', sequential token samples are concatenated and chunked into fixed-length samples of size {self.max_seq_len=}. "
+                f"In general we cannot detect ahead-of-time if your setting of {self.max_samples=} is safe, "
+                f"so please be cautious to not request more tokens than the size of the original dataset.")
+
+        # Build tokenizer
+        self.tokenizer = transformers.AutoTokenizer.from_pretrained(tokenizer_name)
+        if self.tokenizer_name in ["gpt2"]:
```
moinnadeem, line 201: Can we just check `if self.tokenizer.pad_token_id is None` rather than checking if the tokenizer name matches a very specific tokenizer?

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+        # Shuffle dataset
+        # Per-device shard list is shuffled at the start of each epoch
+        # Samples are read into and randomly sampled from per-device shuffle buffer
+        if self.shuffle:
+            text_dataset = text_dataset.shuffle(buffer_size=self.shuffle_buffer_size, seed=self.seed)
+
+        # Map text samples to token samples
+        # NOTE: Mapping is executed in batched mode for better CPU utilization,
+        # but the returned dataset is still an iterable over tokenized samples
+        text_sample_batch_size = 1000
```
moinnadeem, line 220: Can we make this an hparam of the dataset?

composer/datasets/c4.py
```diff
@@ -0,0 +1,333 @@
...
+            batch_size=text_sample_batch_size,
+        )
+
+        if self.group_method != "truncate":
+            # Map variable-length token samples to fixed-length token samples
+            # NOTE: Mapping is executed in batched mode for better CPU utilization,
+            # but the returned dataset is still an iterable over tokenized samples.
+            # NOTE: Depending on the 'group_method', this step may alter the number of
+            # token samples in the dataset, and may mix neighboring token samples together.
+            token_sample_batch_size = 1000
```
moinnadeem, line 233: Can we make this an hparam of the dataset?

moinnadeem (commented):
Added a few comments, correctness looks good to me from what I can tell! Good work Abhi!

composer/datasets/c4.py
abhi-mosaic: Yes, this is very specific to the C4 dataset built/uploaded by AllenAI and hosted on HuggingFace. I can simplify it to a dict.

composer/datasets/c4.py
abhi-mosaic: Sadly, yes... in general it gets pretty tricky to handle non-divisible number of shards, I can think of two options:
* truncate / duplicate some shards to make `new_num_shards` divisible
* figure out some way to iterate through one set of shards normally, and then a second set of shards (the leftovers) via subsampling... but this seems a bit complicated.

For context, the `webdataset` package also just splits the shard list among all CPU workers and assumes that the `num_shards` is divisible.

I think these limitations can be overcome, but I wonder if this is the best place for it, or if we can follow up with some more general adapter that can correct both HF and Webdataset IterableDatasets.

composer/datasets/c4.py
abhi-mosaic: Yup! As long as you think that the EOS -> PAD replacement is generally safe, and not just a GPT2 thing.

composer/datasets/c4.py
abhi-mosaic: Sadly this is what the HF IterableDatset.shuffle function does, and there's no flag to turn off shard shuffling: https://huggingface.co/docs/datasets/package_reference/main_classes.html#datasets.IterableDataset.shuffle. 

However I could potentially move the shuffle op to the very end of the dataset definition, after `.take(samples_per_device)`. This would hypothetically stop the shard shufffling and mean that `max_samples` will always refer to the same subset regardless of seed. I think this is preferable...

... but note that `max_samples` is still affected by the number of devices! And that is impossible to fix... because if using 1 device the first 10 samples are all from shard1, but if using 2 devices the first 10 samples are (5 from shard1, 5 from shard2).

composer/datasets/c4.py
abhi-mosaic: I'm a little wary of exposing this as it's kind of a CPU utilization thing and seems like clutter, I think it's also the HF default (so maybe I can even remove it from my code). What do you think?

composer/datasets/c4.py
hanlint: Ok with not addressing, just wanted clarification.

composer/datasets/c4.py
abhi-mosaic: Closing for now, I think it will be simpler to keep this hparam internal as it's a perf thing.

abhi-mosaic:
Hey @hanlint @moinnadeem , sorry for the delay on this PR. I think I should have addressed all the comments. Hope we can get this merged today! And then I can work with @moinnadeem to update the BERT yaml to use C4, as well as convert some of the large GPT2 yamls (> 125M) to GPT3.

hanlint (approved):
LGTM from code quality stand point, defer to @moinnadeem for NLP correctness

abhi-mosaic:
Also, I'm seeing a lot of new pyright errors.. but I don't think there are any related to `c4.py`. Please let me know if this is expected / what I can do.

abhi-mosaic:
Turns out the self documenting f-string syntax `f"{var=}` did not appear until python3.8, so our python3.7 tests were failing 🙃 . Reverted...

PR #490, opened 2022-02-16
Removing HF XFail on model registry

moinnadeem:
Since our model registry tests don't actually try and call `initialize_object`, we should be fine actually running the tests that pull from the HF hub. They don't download anything from the internet until we actually call `initialize_object` for the hparams.


hanlint (approved):
ah, great!

PR #496, opened 2022-02-16
allow `from composer import ComposerModel`

### LINKED ISSUE #491: Move `ComposerTrainer` to top-level imports
Our most heavily used objects should be easily importable from `composer` via:
```
from composer import Trainer, ComposerModel
```
rather than remember their submodule:

```
from composer.models import ComposerModel
```

Especially the last one, its tricky to remember whether its `models` or `model`

hanlint:
Fixes #491


ravi-mosaicml (approved):
LGTM

PR #502, opened 2022-02-17
Docs cleanup #3

### LINKED ISSUE #430: Refactor `callback._run_event` into `callback.run_event`
I believe the `callback._run_event` private method is a relic from when we had deferred logging and rank zero logging. I think this can be replaced by putting the logic currently there into the public `run_event` method. Thanks @Averylamp for spotting.

ravi-mosaicml:
1. Explicitly setting `autodoc_inherit_docstrings = False` to avoid documenting inherited docstrings, and removed all `:meta private:` that was set as a workaround
2. Adding `composer.functional` to the api reference
3. Fixed many broken links
4. Also closed #430 (not a docs issue; this is going along for the ride. but it's really small!)


docs/source/tables/algorithms_table.md
```diff
@@ -4,18 +4,18 @@ Name|Functional|Attribution|tl;dr
...
 [ProgressiveResizing](https://github.com/mosaicml/composer/tree/dev/composer/algorithms/progressive_resizing)|`cf.resize_batch`|[Fast AI](https://github.com/fastai/fastbook/blob/780b76bef3127ce5b64f8230fce60e915a7e0735/07_sizing_and_tta.ipynb)|Increases the input image size during training
 [RandAugment](https://github.com/mosaicml/composer/tree/dev/composer/algorithms/randaugment)|`cf.randaugment_image`|[(Cubuk et al, 2020)](https://openaccess.thecvf.com/content_CVPRW_2020/html/w40/Cubuk_Randaugment_Practical_Automated_Data_Augmentation_With_a_Reduced_Search_Space_CVPRW_2020_paper.html)|Applies a series of random augmentations
 [SAM](https://github.com/mosaicml/composer/tree/dev/composer/algorithms/sam)|`cf.`|[(Foret et al, 2021)](https://arxiv.org/abs/2010.01412)|SAM optimizer measures sharpness of optimization space
-[ScaleSchedule](https://github.com/mosaicml/composer/tree/dev/composer/algorithms/scale_schedule)|`cf.`|[Many](www.mosaicml.com)|Scale the learning rate schedule by a factor
+[ScaleSchedule](https://github.com/mosaicml/composer/tree/dev/composer/algorithms/scale_schedule)|`cf.`|[Many](https://www.mosaicml.com)|Scale the learning rate schedule by a factor
```
hanlint, line 18: nit: we would want to fix this in the `metadata.json` of the file, as this is an auto-generated file.

hanlint (approved):

docs/source/tables/algorithms_table.md
ravi-mosaicml: Good catch, thanks!

PR #506, opened 2022-02-17
Adding network retries for downloading GLUE

moinnadeem:
If we incur network instabilities, then connection errors in downloading the dataset will cause the job to fail. This adds HTTP retries to downloading GLUE datasets over the network.


composer/datasets/glue.py
```diff
@@ -76,10 +79,12 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
         self.validate()
         self.tokenizer = transformers.AutoTokenizer.from_pretrained(self.tokenizer_name)  #type: ignore (thirdparty)
 
-        log.info(f"Loading {self.task.upper()}...")
-        self.dataset = datasets.load_dataset("glue", self.task, split=self.split)
+        log.info(f"Loading {self.task.upper()} on rank ", dist.get_global_rank())
+        download_config = datasets.utils.DownloadConfig(max_retries=self.max_network_retries)
+        self.dataset = datasets.load_dataset("glue", self.task, split=self.split, download_config=download_config)
+
+        n_cpus = cpu_count() // dist.get_world_size()
```
ravi-mosaicml, line 86: This is outside of this PR, but using `cpu_count()` is dangerous since on k8s, it will show the CPUs of the entire node, not the number of CPUs that are allocated. Instead, I would pass this argument explicitly -- e.g. have a `num_workers` hparam.

ravi-mosaicml (approved):
LGTM

composer/datasets/glue.py
moinnadeem: Good point, refactoring that now.

PR #507, opened 2022-02-17
Anaconda configuration

### LINKED ISSUE #304: Add Composer to conda

ravi-mosaicml:
* Moved all jenkins scripts out of the Jenkinsfile and into the `.ci` folder, since only Jenkins should use these scripts
* Fixed up the Makefile so we can run tests via Makefile commands
* Added Anaconda `meta.yaml`, which as part of the build, runs tests via `Makefile` commands
* Updated the Jenkinsfile to build the anaconda package whenever the requirements or anaconda's meta.yaml changes
* Updated all dependency missing errors to include the corresponding anaconda command to use to install the missing dependency

Closes #304


Averylamp (approved):
lgtm

Makefile
```diff
@@ -1,41 +1,43 @@
 # several pytest settings
-DURATION ?= all  # pytest duration, one of short, long or all
+DURATION ?= short  # pytest duration, one of short, long or all
```
hanlint, line 2: since the `test.sh` uses DURATION=all, this should be default so someone can easily reproduce the failing tests without needing to pass in DURATION=all to `make test`.

composer/datasets/brats.py
```diff
@@ -257,8 +258,9 @@ def get_data_split(path: str):
         from sklearn.model_selection import KFold
     except ImportError as e:
         raise ImportError(
-            "Composer was installed without unet support. To use unet with Composer, run: `pip install mosaicml[unet]`."
-        ) from e
+            textwrap.dedent("""\
```
hanlint, line 261: seems to a Jenkins issue if we need to litter our codebase with `textwrap.dedent`s

Makefile
ravi-mosaicml: Discussed offline; leaving as all here and changing to default to all in conftest.py

composer/datasets/brats.py
ravi-mosaicml: This isn't a Jenkins issue but rather an effort to not rely on implicit string concatenation...see https://www.python.org/dev/peps/pep-3126 and https://lwn.net/Articles/551426/

PR #514, opened 2022-02-17
Fix pil_image_collate

Landanjs:
#439 Removed an essential part of `pil_image_collate` when the target is in the shape of an image. Without this code, training errors out. Adding this code back in, but happy to take in suggestions to make it cleaner


ravi-mosaicml (approved):
Whoops, my bad, but LGTM. Can you add a test case that exercises this code path?

Landanjs:
I added tests for `pil_image_collate`, one test for when the target is scalar and one test for when the target is an image. Let me know if more needs to be added!

I also changed the `pil_image_collate` output type to `Tuple[Tensor, Tensor]` since pyright complained with `BatchPair` and I think the typing is more accurate.

PR #516, opened 2022-02-17
Docstrings improvements for core.algorithm, core.callback, etc.

dskhudia:
Docstrings improvements for core.algorithm, core.callback, core.data_spec, core.engine and core.event


ravi-mosaicml:
Overall, looks good! Some comments:

- [ ] Can you add a module-level docstring to composer.core?
- [ ] In Algorithm.apply, can you remove the :class: notation from the parameters?
- [ ] For `find_unused_parameters`, can you comment on the effect of using this with deepspeed?
- [ ] The 2nd bullet for the callback docstring shows `#.` instead of `2.`  The for example section is on a background. I suspect there is some indentation issue.
- [ ] For the `..note`` in the callback's section's `batch_end` and `epoch_end`, can you mention that these fields refer to the state.timer? Perhaps also remove the `~` so it will print the full path. It's a bit confusing to see these fields here that have nothing to do with the callback itself.
- [ ] For the DataSpec class, can you include an example of when one would use a DataSpec? Perhaps we can show the ImageNet example, with comments for each field.
- [ ] On the engine, "For the events that follow" -> "For the events that follow the"
- [ ] "These are logged with the key ‘<algorithm_name>/<event>’" : Can you make the key preformatted?
- [ ] For the Engine's init, can you add a `..note::` saying that the trainer constructs this automatically -- the user does not need to construct the engine?
- [ ] In run_event, same comment for the key -- can it be preformatted?
- [ ] In composer.core.engine.Traces, same comment for the key -- can it be preformatted?
- [ ] In the Event docs, Trainer.fit() does not auto-link.

dskhudia:
@ravi-mosaicml Added an example too for DataSpec.

composer/core/__init__.py
```diff
@@ -1,5 +1,9 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Central parts of composer such as :class:`.engine.Engine`, base class for critical components such as
+:class:`.core.Algorithm` and :class:`.core.Callback` and other useful functionality such as :class:`.logger.Logger` and
```
ravi-mosaicml, line 4: core.Algorithm and core.Callback are not auto-linking.

composer/core/data_spec.py
```diff
@@ -12,43 +14,63 @@
...
+    An example of constructing a :class:`DataSpec` object with a ``device_transforms`` callable
+    (:class:`~composer.datasets.utils.NormalizationFn`) and then using it with :class:`~.Trainer`:
+
+       >>> # In this case, we apply NormalizationFn 
+       >>> # Construct DataSpec as shown below to apply this transformation
+       >>> from composer.datasets.utils import NormalizationFn
+       >>> CHANNEL_MEAN = (0.485 * 255, 0.456 * 255, 0.406 * 255)
+       >>> CHANNEL_STD = (0.229 * 255, 0.224 * 255, 0.225 * 255)
+       >>> device_transform_fn = NormalizationFn(mean=CHANNEL_MEAN, std=CHANNEL_STD)
+       >>> dspec = DataSpec(train_dataloader, device_transforms=device_transform_fn)
```
ravi-mosaicml, line 32: For this example, to follow what imagenet actually does, could the eval dataloader also be a dataspec and use the same normalization function?

composer/core/engine.py
```diff
@@ -1,36 +1,49 @@
...
+    * The design of :class:`~.engine.Engine` is subject to change in future releases to accommodate more complexity as
+      we investigate composition of algorithms.
 
 
 Currently, the following passes are registered:
 
 * **LIFO order for events**
 
-   For events that follow the ``after_*`` and ``before_*`` pattern, the ordering of algorithms is reversed for the ``after_*`` events. For example, algorithms will run in a ``ABCD -> DCBA`` ordering before and after say, the loss computation.
+   For the events that follow the ``before_*`` (e.g., :attr:`~.Event.BEFORE_LOSS`) and ``after_*`` (e.g.,
```
ravi-mosaicml, line 24: Could these lines (24-31) be un-indented by 1 space, so it isn't highlighted in the docs?

composer/core/engine.py
```diff
@@ -1,36 +1,49 @@
...
 
-   This allows algorithms to "clean up" their changes. e.g. Label smoothing will smooth the labels upon entry to the loss, and then restore the original unsmoothed labels upon exit.
+   This allows algorithms to "clean up" their changes. For example, :class:`~.LabelSmoothing` will smooth the labels
+   upon on :attr:`~.Event.BEFORE_LOSS` event and then restore the original unsmoothed labels on
+   :attr:`~.Event.AFTER_LOSS` event.
 
 * **Run Selective Backprop first**
 
-   Selective backprop runs after the dataloader returns the batch, and executes an extra forward pass to rank and prune the examples in the batch by loss. To ensure a clean estimate of the example, Selective backprop should run before any other data augmentations during ``AFTER_DATALOADER`` (e.g. such as MixUp).
+   :class:`~.SelectiveBackprop` runs after the dataloader returns the batch, and executes an extra forward pass to rank
```
ravi-mosaicml, line 35: Same thing as above -- can you unindent by 1 space?

ravi-mosaicml (approved):
LGTM; see comments.

hanlint (approved):
Excellent, as usual thanks @dskhudia

composer/core/data_spec.py
dskhudia: It can be used for eval_dataloader as well. Added a comment + DataSpec for eval in the example.

composer/core/engine.py
dskhudia: fixed

composer/core/engine.py
dskhudia: fixed

PR #520, opened 2022-02-18
Fix Sphinx Warnings

ravi-mosaicml:
1. Fixes all sphinx warnings. To do so, I had to fix some links and properly define `__all__` in various places.
2. Failing sphinx builds if warnings are present. The linting check builds sphinx, so that will fail if any new PRs introduce sphinx warnings.


docs/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -W
```
bandish-shah, line 6: [nit] add `--keep-going` so the user gets the complete list of warnings rather than failing and stopping on the first warning.

bandish-shah (approved):
LGTM this is awesome!

PR #522, opened 2022-02-18
Docstrings for composer.trainer

ajaysaini725:
Docstrings for composer.trainer (trainer, checkpointing, deepspeed, ddp, devices, scale schedule, grad scaler)

Addresses #396


composer/trainer/__init__.py
```diff
@@ -2,72 +2,66 @@
...
-
-
-.. code-block:: python
-
-    from composer.trainer import TrainerHparamms
-
-    # Create a trainer from hparams and train train the model
-    trainer = hparams.initialize_object()
-    trainer.fit()
+.. doctest::
```
ajaysaini725, line 12: Thoughts on examples to include here? I figured since we have so many tutorials about the trainer it's best to just include the most straightforward use case in the docs here (create a trainer, save a checkpoint, resume training later)

Could maybe add something with speedup methods too if anyone thinks it's relevant? Just trying to not make the example more verbose than it already is

ravi-mosaicml: I would also be for documenting the Trainer on composer.trainer instead of composer.trainer.trainer -- to do this, rename `trainer.py` to `_trainer.py`, and then in `composer/trainer/__init__.py` add a line like this:

```python3

from composer.trainer._trainer import Trainer
Trainer.__module__ = __name__
```

composer/trainer/__init__.py
```diff
@@ -2,72 +2,66 @@
...
+    ...                   validate_every_n_epochs=1,
+    ...                   load_path=checkpoint_path)
+    >>>
+    >>> ### Continue training and running evaluation where the previous trainer left off
+    >>> ### until the new max_duration is reached.
+    >>> ### In this case it will be one additional epoch to reach 2 epochs total.
+    >>> trainer.fit()
 """
 from composer.trainer import devices as devices
 from composer.trainer.trainer import Trainer as Trainer
```
ravi-mosaicml, line 62: Thoughts about documinting the Trainer on the `composer.trainer` page rather than the composer.trainer.trainer` submodule?

To do this, rename `trainer.py` to `_trainer.py`, and then in `composer/trainer/__init__.py`:

```python3

from composer.trainer._trainer import Trainer
Trainer.__module__ = __name__
```

composer/trainer/checkpoint.py
```diff
@@ -70,21 +77,22 @@ class CheckpointLoader:
...
-        object_store (ObjectStoreProvider, optional): If the ``checkpoint`` is in an object store
-            (i.e. AWS S3 or Google Cloud Storage), an instance of :class:`ObjectStoreProvider` which will be used
+        object_store (ObjectStoreProvider, optional): If the ``path`` is in an object store
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~composer.utils.object_store.ObjectStoreProvider` which will be used
             to retreive the checkpoint. Otherwise, if the checkpoint is a local filepath, set to ``None``.
-            (default: ``None``).
-        load_weights_only (bool): Whether to only restore the weights from the checkpoint without
+            (default: ``None``)
+        load_weights_only (bool): Whether or not to only restore the weights from the checkpoint without
```
ravi-mosaicml, line 88:
```suggestion
        load_weights_only (bool, optional): Whether or not to only restore the weights from the checkpoint without
```

composer/trainer/checkpoint.py
```diff
@@ -70,21 +77,22 @@ class CheckpointLoader:
...
             restoring the associated state.
-        strict_model_weights (bool): Whether to force that the checkpointed weights must exactly match the model
-            weights.
-        chunk_size (int, optional): Chunk size (in bytes) to use when downloading checkpoints. Ignored if the checkpoint
-            is a local file path. (default: ``1_048_576`` bytes (1 MB))
-        progress_bar (bool): Whether to show a progress bar when downloading checkpoints. Ignored if the checkpoint
-            is a local file path. (default: ``True``)
+        strict_model_weights (bool): Whether or not to force that the checkpointed weights must exactly
```
ravi-mosaicml, line 90:
```suggestion
        strict_model_weights (bool, optional): Whether or not to force that the checkpointed weights must exactly
```

composer/trainer/checkpoint.py
```diff
@@ -70,21 +77,22 @@ class CheckpointLoader:
...
+            match the model weights.
+        chunk_size (int, optional): Chunk size (in bytes) to use when downloading checkpoints.
+            Ignored if the checkpoint is a local file path. (default: ``1_048_576`` bytes (1 MB))
+        progress_bar (bool): Whether or not to show a progress bar when downloading checkpoints.
```
ravi-mosaicml, line 94:
```suggestion
        progress_bar (bool, optional): Whether or not to show a progress bar when downloading checkpoints.
```

composer/trainer/checkpoint.py
```diff
@@ -1,5 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Load and save checkpoints during training."""
```
ravi-mosaicml, line 3: Thoughts about making this module private? I think I saw this discussed on slack; would be in favor of that. We'll just need to add a helper on the trainer class to get the saved checkpoints.

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
...
+        save_folder (str): The folder to store checkpoints in. The ``save_folder`` will be relative
+            to the folder returned by :meth:`~composer.utils.run_directory.get_run_directory`.
+            If the ``save_folder`` does not exist, it will be created.
         interval (Time or str): The amount of time units to wait between checkpoints.
         compression (str, optional): Compression algorithm to run on checkpoints. Can be ``gzip``, ``bzip2``,
             ``lzma``, or ``None`` for no compression.  (default: ``None`` for no compression).
 
     Attributes:
+        checkpoint_folder (str): The folder in which checkpoints are stored. This folder is relative
+            to the run directory of the training run. If no run directory is proivded, then by default,
```
ravi-mosaicml, line 417:
```suggestion
            to the run directory of the training run (e.g. ``{run_directory}/{save_folder}``). If no run directory is provided, then by default,
```

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
...
+            and ``save_folder`` is the save_folder argument provided upon construction.
+
+            .. seealso:: :mod:`~composer.utils.run_directory` for details on the format of the run directory
+                and how to customize it.
         saved_checkpoints (Dict[Timestamp, List[str]): A dictionary mapping a save timestamp
             to a list of filepaths corresponding to the checkpoints saved at that time.
 
             .. note:: The list of filepaths is for all ranks. The path at index ``i`` is the
-                      path that global rank ``i`` wrote to.
+                path that global rank ``i`` wrote to.
```
ravi-mosaicml, line 428: Can you add to the note that the path is valid only on global rank i's node.

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
...
     Attributes:
+        checkpoint_folder (str): The folder in which checkpoints are stored. This folder is relative
+            to the run directory of the training run. If no run directory is proivded, then by default,
+            it is of the form ``runs/<timestamp>/rank_<GLOBAL_RANK>/<save_folder>`` where ``timestamp``
+            is the the start time of the run in iso-format, ``GLOBAL_RANK`` is the global rank of the process,
+            and ``save_folder`` is the save_folder argument provided upon construction.
+
+            .. seealso:: :mod:`~composer.utils.run_directory` for details on the format of the run directory
+                and how to customize it.
         saved_checkpoints (Dict[Timestamp, List[str]): A dictionary mapping a save timestamp
```
ravi-mosaicml, line 424:
```suggestion
        saved_checkpoints (Dict[Timestamp, List[str]]): A dictionary mapping a save timestamp
```

composer/trainer/__init__.py
hanlint: I'm good with what you have here, but can you break it up into individual Examples? And these should probalby be in the `Trainer` docstrings and not module levle right?

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
...
             to a list of filepaths corresponding to the checkpoints saved at that time.
 
             .. note:: The list of filepaths is for all ranks. The path at index ``i`` is the
```
ravi-mosaicml, line 427:
```suggestion
            .. note:: When using DeepSpeed, the index of a filepath in each list corresponds to the global rank of the process that wrote that file. These filepaths are valid only on the global rank's node. Otherwise, when not using DeepSpeed, this list will contain only one filepath since only rank zero saves checkpoints.
```

composer/trainer/__init__.py
```diff
@@ -2,72 +2,66 @@
 
 """Train models!
 
-The trainer supports models with :class:`ComposerModel` instances.
-The :class:`Trainer` is highly customizable and can support a wide variety of workloads.
+The trainer supports models with :class:`~composer.models.base.ComposerModel` instances.
```
hanlint, line 5: can use :class:`.Trainer` throughout to make them more readable.

composer/trainer/checkpoint.py
```diff
@@ -436,18 +462,26 @@ def should_checkpoint(self, state: State, event: Event) -> bool:
...
+        """Save the current state to a new checkpoint file.
+
+        There are 3 cases for the format in which the checkpoint is saved:
+
+        1. The default is to save checkpoints in a ``.pt`` file if DeepSpeed is not being used to
+        train the model and there is no compression specied.
+
+        2. If DeepSpeed is being used to train the model and there is no compression, then the checkpoint
+        is stored in a ``.tar`` format because DeepSpeed saves model checkpoints
+        as multiple files.
```
ravi-mosaicml, line 481:
```suggestion
        as multiple files (one for model state, and one for optimizer state).
```

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
     """Manager for saving trainer state to checkpoint files.
 
     Args:
-        save_folder (str): The path to store checkpoints in.
+        save_folder (str): The folder to store checkpoints in. The ``save_folder`` will be relative
```
hanlint, line 408: also mention that the `save_folder` can be abosolute path

composer/trainer/checkpoint.py
hanlint, line 417:
```suggestion
            to the run directory of the training run. If no run directory is provided, then by default,
```

composer/trainer/checkpoint.py
```diff
@@ -393,17 +405,27 @@ class CheckpointSaver:
...
+            If the ``save_folder`` does not exist, it will be created.
         interval (Time or str): The amount of time units to wait between checkpoints.
         compression (str, optional): Compression algorithm to run on checkpoints. Can be ``gzip``, ``bzip2``,
             ``lzma``, or ``None`` for no compression.  (default: ``None`` for no compression).
 
     Attributes:
+        checkpoint_folder (str): The folder in which checkpoints are stored. This folder is relative
+            to the run directory of the training run. If no run directory is proivded, then by default,
+            it is of the form ``runs/<timestamp>/rank_<GLOBAL_RANK>/<save_folder>`` where ``timestamp``
+            is the the start time of the run in iso-format, ``GLOBAL_RANK`` is the global rank of the process,
```
hanlint, line 419:
```suggestion
            is the start time of the run in iso-format, ``GLOBAL_RANK`` is the global rank of the process,
```

composer/trainer/ddp.py
```diff
@@ -79,6 +93,16 @@ def ddp_sync_context(state: State, is_final_microbatch: bool, sync_strategy: Uni
 
 
 def prepare_ddp_module(module: Model, find_unused_parameters: bool) -> Model:
```
ravi-mosaicml, line 95: Thoughts about removing this function and just embedding it in the trainer? It's called only in one place, and the trainer already guards calls to this function to ensure that dist is configured and initialized.

composer/trainer/checkpoint.py
hanlint: also, mention absolute?

composer/trainer/checkpoint.py
```diff
@@ -436,18 +462,26 @@ def should_checkpoint(self, state: State, event: Event) -> bool:
...
         return False
 
     def save_checkpoint(self, state: State, seed: int, device: Device) -> None:
-        """Save the current state to a a new checkpoint file.
+        """Save the current state to a new checkpoint file.
+
+        There are 3 cases for the format in which the checkpoint is saved:
+
+        1. The default is to save checkpoints in a ``.pt`` file if DeepSpeed is not being used to
+        train the model and there is no compression specied.
```
hanlint, line 477:
```suggestion
        train the model and there is no compression specified.
```

composer/trainer/ddp.py
```diff
@@ -79,6 +93,16 @@ def ddp_sync_context(state: State, is_final_microbatch: bool, sync_strategy: Uni
...
+    """Wraps the module in a :class:`~torch.nn.parallel.DistributedDataParallel` object if
```
hanlint, line 96:
```suggestion
    """Wraps the module in a :class:`torch.nn.parallel.DistributedDataParallel` object if
```

long paths preferred for torch-based ones for clarity

composer/trainer/ddp.py
```diff
@@ -42,6 +46,16 @@ class DDPSyncStrategy(StringEnum):
 
 @contextmanager
 def ddp_sync_context(state: State, is_final_microbatch: bool, sync_strategy: Union[str, DDPSyncStrategy]):
```
ravi-mosaicml, line 48: Thoughts about making it private? It's used only by the trainer in one place.

composer/trainer/trainer.py
```diff
@@ -50,60 +52,80 @@
...
+__all__ = ["Trainer"]
 
-class Trainer:
-    """Trainer for training a model with algorithms.
 
-    Can be created either with ``__init__`` or by providing a
-    :class:`~composer.trainer.TrainerHparams` object
-    (see :meth:`~composer.trainer.Trainer.create_from_hparams`).
+class Trainer:
+    """Trainer for training a models with Composer algorithms.
```
hanlint, line 59: should put the examples form the module level here, and also say "See our Trainer guide for more information" (we'll add the link later)

composer/trainer/trainer.py
```diff
@@ -50,60 +52,80 @@
...
-            degradations expected. Certain Torch modules may not have deterministic implementations,
-            which will result in a crash. (default: ``False``)
-        loggers (List[LoggerCallback], optional): The destinations to log training information to.
-            (default: ``[TQDMLogger()]``).
-        callbacks (Sequence[Callback], optional): The callbacks to run during training. (default: ``[]``)
-        load_path (str, optional): Path to a specific checkpoint to load. If not set (the default),
-            then no checkpoint will be loaded. (default: ``None``)
+        deterministic_mode (bool, optional): Run the model deterministically. (default: ``False``)
+
+            .. note:: This is an exerpimental feature. Performance degradations expected. Certain Torch modules may
```
hanlint, line 114:
```suggestion
            .. note:: This is an experimental feature. Performance degradations expected. Certain Torch modules may
```

composer/trainer/ddp.py
```diff
@@ -10,6 +12,8 @@
 from composer.utils import dist
 from composer.utils.string_enum import StringEnum
 
+__all__ = ["DDPSyncStrategy", "ddp_sync_context", "prepare_ddp_module"]
+
 
 class DDPSyncStrategy(StringEnum):
```
ravi-mosaicml, line 18: Thoughts about documenting this on the module level page for the trainer? Then, we can make this module private.

composer/trainer/ddp.py
```diff
@@ -1,5 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Helpers for running distributed data parallel training."""
```
ravi-mosaicml, line 3: Thoughts about making this module private? See comments below.

composer/trainer/trainer.py
```diff
@@ -50,60 +52,80 @@
...
-            class contains metrics relevant to the specific dataset. Set to ``None`` for no evaluation.
-        algorithms (List[Algorithm], optional): The algorithms to use during training.
-            (default: ``[]``)
-        optimizers: (Optimizers, optional): The optimizers.
-            (default: ``DecoupledSGDW(model.parameters(), lr=0.1)``)
-        schedulers: (Schedulers, optional): The schedulers.
-            (default: ``[CosineAnnealingLR()]``).
-        device (str or Device, optional): The device to use for training. Either `cpu` or `gpu`.
-            (default `cpu`)
+        eval_dataloader (DataLoader, DataSpec, or Tuple[Evaluator], optional): The :class:`DataLoader`,
```
hanlint, line 67: Since this entire section may often be consulted in code (vs the docs), please us global scope for sphinx as much as possible. e.g. 
:class:`~composer.trainer.ddp.DDPSyncStrategy`
->
:class:`.DDPSyncStrategy`

Or, if not possible due to duplication, use sphinx substitutions:

ravi-mosaicml (commented):
Leaving some comments for now...

composer/trainer/__init__.py
ajaysaini725: Oh do you need the "." in front of the `Trainer`? It was originally written as :class:`Trainer` without the dot and the linking didn't work but I can fix

composer/trainer/__init__.py
ajaysaini725: Lol @hanlint @ravi-mosaicml you have conflicting opinions 
- First is to document the trainer in the top level module by renaming `trainer.py` to `_trainer.py`
- Second is to move the trainer docstraings to `trainer.py` and leave the module-level docs (relatively) empty


Thoughts?

composer/trainer/ddp.py
ajaysaini725: Yup sounds good!

composer/trainer/__init__.py
hanlint: 🤔 let's resolve offline!

composer/trainer/devices/device_cpu.py
```diff
@@ -11,6 +13,8 @@
 
 logger = logging.getLogger(__name__)
 
+__all__ = ["DeviceCPU"]
```
ravi-mosaicml, line 16: Can you denote in the init signature that this class takes no arguments? Maybe give an example of how you would pass this in when constructing the trainer.

ravi-mosaicml (commented):
Finished reviewing! Great work overall; I know the trainer docs are a beast.

composer/trainer/devices/device.py
```diff
@@ -10,15 +12,18 @@
 from composer.core.types import Batch, BatchPair, Optimizer, Precision, Tensor
 from composer.utils.iter_helpers import map_collection
 
+__all__ = ["Device", "T_nnModule"]
+
 T_nnModule = TypeVar("T_nnModule", bound=torch.nn.Module)
```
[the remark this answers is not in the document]
ajaysaini725: I only exported it because it's used as a parameter to some things in device so I figured it's best to cross reference.

composer/trainer/devices/device.py
```diff
@@ -1,5 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""The base :class:`~composer.trainer.devices.device.Device` class."""
```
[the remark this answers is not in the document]
ajaysaini725: Added the missing docstrings - based on offline discussion should I still move all device docs to `composer.trainer.devices`?

composer/trainer/devices/device_gpu.py
```diff
@@ -13,6 +15,8 @@
 from composer.trainer.devices.device import Device, T_nnModule
 from composer.utils import dist
 
+__all__ = ["DeviceGPU"]
```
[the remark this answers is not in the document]
ajaysaini725: Can't use doctest for this because you can't instantiate a `DeviceGPU` in doctest but added the comment about init taking no args

ajaysaini725:
Addressed all the style related comments only things left should be to make `checkpoint.py`, `ddp.py`, and `deepspeed.py` private.

composer/trainer/__init__.py
ajaysaini725: Resolved offline

composer/trainer/ddp.py
ajaysaini725: Resolved offline

composer/trainer/ddp.py
ajaysaini725: Edit: Since we can't make `ddp.py` private I'm going to leave it here since it is "ddp logic" and it's nice to keep it in a separate file from the trainer which is already a huge file - lmk if you think otherwise though and I can move it

composer/trainer/deepspeed.py
```diff
@@ -136,9 +159,17 @@ def _convert_fp32_tensor_to_fp16(tensor: Tensor):
 def fix_batch_precision_for_deepspeed(batch: Batch, precision: Precision) -> Batch:
```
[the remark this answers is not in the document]
ajaysaini725: made the whole module private

composer/trainer/devices/device.py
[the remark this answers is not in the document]
ajaysaini725: Discussed offline - not going to do because we decided to not re-export modules

composer/trainer/__init__.py
ravi-mosaicml: Agree with Hanlin's points here to break up the examples and move it to the init docstring. Because of how long the example is, it may make it slighlty cleaner to use ` ..testcode::` rather than `.. doctest::` (see https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html). But your call :)

composer/trainer/ddp.py
```diff
@@ -1,5 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
```
ravi-mosaicml, line 1: There's a typo at the bottom -- "contextmanager" should be "context manager"

ravi-mosaicml: Also, even if the module remains public, ddp_sync_context and prepare_ddp_module can be private functions. Not a big deal -- could go either way -- but could simplify the docs a bit.

composer/trainer/devices/device.py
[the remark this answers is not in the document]
ravi-mosaicml: That's fine; it would probably be better to re-define this in each file where it is used (I think it's rare for TypeVars to be re-imported across files)

composer/trainer/devices/device_cpu.py
```diff
@@ -11,9 +13,15 @@
...
+
 
 class DeviceCPU(Device):
-    """An extension of :class:`~composer.trainer.devices.device.Device` for CPUs."""
+    """An extension of :class:`~composer.trainer.devices.device.Device` for CPUs.
+
+    This class takes no arguments. Below is an example of how to pass it into the
```
ravi-mosaicml, line 22:
```suggestion
    This class takes no arguments.
```

composer/trainer/devices/device_gpu.py
```diff
@@ -13,9 +15,15 @@
...
+
 
 class DeviceGPU(Device):
-    """An extension of :class:`~composer.trainer.devices.device.Device` for GPUs."""
+    """An extension of :class:`~composer.trainer.devices.device.Device` for GPUs.
+
+    This class takes no arguments. Below is an example of how to pass it into the
```
ravi-mosaicml, line 24:
```suggestion
    This class takes no arguments.
```

composer/trainer/devices/device_hparams.py
```diff
@@ -22,13 +27,15 @@ def initialize_object(self) -> Device:
...
 class GPUDeviceHparams(DeviceHparams):
+    """Used to construct a :class:`~composer.trainer.devices.device_gpu.GPUDevice`"""
 
     def initialize_object(self) -> DeviceGPU:
         return DeviceGPU()
 
 
 @dataclass
 class CPUDeviceHparams(DeviceHparams):
+    """Used to construct a :class:`~composer.trainer.devices.device_cpu.CPUDevice`"""
```
ravi-mosaicml, line 38: not autolinking...

ravi-mosaicml:
I think the `scale_scheduler` can also be private -- it is only used by the trainer (and the deprecated ScaleSchedule algorithm).

ravi-mosaicml (commented):
Will continue reviewing tomorrow!

composer/trainer/trainer.py
```diff
@@ -47,46 +50,74 @@
...
+            :class:`.DataSpec`, or :class:`.Evaluators` for the evaluation data.
+            In order to evaluate one or more specific metrics across one or more datasets, pass in an
+            :class:`.Evaluator`. If a :class:`.DataSpec` or :class:`.DataLoader` is passed in, then all
+            metrics returned by ``model.metrics()`` will be used during evaluation.
+            ``None`` results in no evaluation. (default: ``None``)
+        algorithms (List[Algorithm], optional): The algorithms to use during training. If ``None``, then
+            no algorithms will be used. (default: ``None``)
+
+            .. seealso:: :mod:`composer.algorithms` for the different algorithms built into Composer.
+        optimizers: (Optimizers, optional): The optimizer.
```
ravi-mosaicml, line 86:
```suggestion
        optimizers (Optimizers, optional): The optimizer.
```

composer/trainer/trainer.py
```diff
@@ -47,46 +50,74 @@
...
+            If ``None``, will be set to ``DecoupledSGDW(model.parameters(), lr=0.1)``. (default: ``None``)
+
+            .. seealso:: :mod:`composer.optim` for the different optimizers built into Composer.
+        schedulers: (Schedulers, optional): The learning rate schedulers. If ``[]`` or ``None``, will be set to
```
ravi-mosaicml, line 90:
```suggestion
        schedulers (Schedulers, optional): The learning rate schedulers. If ``[]`` or ``None``, will be set to
```

composer/trainer/trainer.py
```diff
@@ -47,46 +50,74 @@
...
+        compute_training_metrics (bool, optional): ``True`` to compute metrics on training data and ``False`` to not.
             (default: ``False``)
-        precision (str or Precision, optional): Numerical precision to use for training, one of 'fp32', 'fp16'
-            or 'amp' (recommended). (default: ``Precision.FP32``).
+        precision (str or Precision, optional): Numerical precision to use for training. One of ``fp32``, ``fp16``
+            or ``amp`` (recommended). (default: ``Precision.FP32``)
+
+            .. note::
+                ``fp16`` only works if ``deepspeed_config`` is also provided.
         scale_schedule_ratio (float, optional): Ratio by which to scale the training duration and learning rate
```
ravi-mosaicml, line 116: If the scale scheduler function becomes private, can you copy the relevant parts of the docstring from the scale scheduler to here, and remove the seealso?

composer/trainer/trainer.py
```diff
@@ -47,46 +50,74 @@
...
-            schedules. See :func:`scale_schedule` for details. (default: ``1.0``)
+            schedules. (default: ``1.0``)
+
+            .. seealso:: :func:`.scale_scheduler` for details on how scale scheduling
+                works.
         use_stepwise_schedulers (bool, optional): Whether schedulers will update after every optimizer step
```
ravi-mosaicml, line 121: Can we rename this parameter to `step_schedulers_every_batch`? Stepwise is a bit unclear, since all schedulers have a step function.

"By default, stepwise schedulers are used when functional schedulers are provided, but not if only native Pytorch schedulers are provided (i.e. subclasses of torch.optim.lr_scheduler._LRScheduler). (default: ``None)" is formatting incorrectly. I would also rephrase this to say something like "By default, native pytorch schedulers (https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate) are updated every epoch, while :mod:`composer schedulers<composer.optim>` are updated every step. Just not a big fan of mentioning a private class, and we need to explain what a "functional scheduler" is,

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
-            Set to ``None`` to not save checkpoints. (default: ``None``)
-        save_interval (str or int): How often to save checkpoints. For example, set to "1ep" to save checkpoints
+
+            .. note:: In order to get reproducible results, call the
+                :func:`.seed_all` function at the start of your script with the seed
+                passed to the trainer. This will ensure any initialization done before the trainer init
+                (ex. model weight initialization) also uses the provided seed.
+
+            .. seealso:: :mod:`composer.utils.reproducibility` for more details on reproducibility.
+        deterministic_mode (bool, optional): Run the model deterministically. (default: ``False``)
```
ravi-mosaicml, line 142: Can you add a similar note here that you added for the seed?

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
+            For example, suppose that checkpoints are stored in the following structure:
+
+            .. code-block::
+
+                my_model/rank_0/ep1.tar
+                my_model/rank_1/ep1.tar
+                my_model/rank_2/ep1.tar
+                ...
+
+            Then, ``path`` should be set to ``my_model/rank_{RANK}/ep1.tar``, and all ranks will load the correct
```
ravi-mosaicml, line 174:
```suggestion
            Then, ``load_path`` should be set to ``my_model/rank_{RANK}/ep1.tar``, and all ranks will load the correct
```

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
+            data.
+
+            If ``None`` then no checkpoint will be loaded. (default: ``None``)
+        load_object_store (ObjectStoreProvider, optional): If the ``load_path`` is in an object store
```
ravi-mosaicml, line 178: Can you add an example of loading a checkpoint from an object store? I feel that it is advanced feature to have a parameterized path combined with retrieving a checkpoint from a bucket.

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
+        load_chunk_size (int, optional): Chunk size (in bytes) to use when downloading checkpoints.
+            Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
+        load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
+            Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        save_folder (str, optional): The folder to store checkpoints in. If an absolute path is specified, then
+            that path will be used. Otherwise, the ``save_folder`` will be relative
+            to the folder returned by :func:`.get_run_directory`.
+            If the ``save_folder`` does not exist, it will be created. If ``None``, then no checkpoints will
+            be saved. (default: ``None``)
+        save_interval (str or int, optional): How often to save checkpoints. For example, set to "1ep" to save checkpoints
```
ravi-mosaicml, line 195:
```suggestion
        save_interval (str or int, optional): How often to save checkpoints. For example, set to ``1ep`` to save checkpoints
```

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
             every epoch, or "10ba" to save checkpoints every 10 batches. An integer will be assumed to be epochs.
```
ravi-mosaicml, line 196:
```suggestion
            every epoch, or ``10ba`` to save checkpoints every 10 batches. An integer will be assumed to be epochs.
```

composer/trainer/trainer.py
```diff
@@ -97,37 +128,85 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
+        save_compression (str, optional): Compression algorithm to run on checkpoints. Can be ``gzip``, ``bzip2``,
+            ``lzma``, or ``None`` for no compression. Ignored if ``save_folder`` is ``None``. (default: ``None``)
+        train_subset_num_batches (int, optional): If specified, finish every epoch early after training
+            on this many batches. This parameter has no effect if it is greater than ``len(train_dataloader)``.
+            If ``None``, then the entire dataloader will be iterated over. (default: ``None``)
+        eval_subset_num_batches (int, optional): If specified, evaluate on this many batches.
+            This parameter has no effect if it is greater than ``len(eval_dataloader)``.
+            If ``None``, then the entire dataloader will be iterated over. (default: ``None``)
         profiler_trace_file (str, optional): Name of the trace file, relative to the run directory.
             Must be specified to activate the profiler. (default: ``None``).
```
ravi-mosaicml, line 207:
```suggestion
            Setting this parameter activates the profiler. (default: ``None``).
```

composer/trainer/trainer.py
```diff
@@ -163,20 +242,16 @@ class contains metrics relevant to the specific dataset. Set to ``None`` for no
...
-        train_subset_num_batches (int, optional): If specified, finish every epoch early after training
-            on this many batches. This parameter has no effect if it is greater than ``len(train_dataloader)``.
-            If None (the default), then the entire dataloader will be iterated over.
-        eval_subset_num_batches (int, optional): If specified, evaluate on this many batches.
-            This parameter has no effect if it is greater than ``len(eval_dataloader)``.
-            If None (the default), then the entire dataloader will be iterated over.
-        deepspeed_config (Dict[str, Any], optional): Configuration for DeepSpeed, formatted as a dictionary
-            according to the `DeepSpeed documentation <https://www.deepspeed.ai/docs/config-json/>`_. If any
-            non-None value is provided, the trainer will initialize the DeepSpeed engine. (default: ``None``)
+        deepspeed_config (Dict[str, Any], optional): Configuration for DeepSpeed, formatted as a JSON
```
ravi-mosaicml, line 245: Can you move the deep speed config above the profiling parameters?

Also, can we update the type annotation to be `Union[bool, dict]` and change the default value to `False`. It would be nice if you could set `deepspeed_config = True` to be equivalent to `deepspeed_config = {}`....it feels a tad bit weird that setting something to an empty dictionary has a very different effect than setting something to None.

composer/trainer/trainer.py
```diff
@@ -526,15 +601,54 @@ def __init__(
...
+
+        .. seealso:: `DeepSpeed's documentation <https://www.deepspeed.ai/docs/config-json/>`_
+        """
+        return self._deepspeed_config is not None
+
+    @property
+    def saved_checkpoints(self) -> Dict[Timestamp, List[str]]:
+        """The times and paths to checkpoint files saved across all ranks during training.
+
+        .. note:: When using DeepSpeed, the index of a filepath in each list corresponds to the
```
ravi-mosaicml, line 622: maybe move the note into the return docstring? It's a bit confusing without knowning the return format.

```suggestion
        .. note:: When using DeepSpeed, the index of a filepath corresponds to the
```

composer/trainer/trainer.py
```diff
@@ -526,15 +601,54 @@ def __init__(
...
+            global rank of the process that wrote that file. These filepaths are valid only on
+            the global rank's node. Otherwise, when not using DeepSpeed, this list will contain
+            only one filepath since only rank zero saves checkpoints.
+
+        Returns:
+            Dict[Timestamp, List[str]]: A dictionary mapping a save :class:`.Timestamp`. to a list of
+                filepaths corresponding to the checkpoints saved at that time.
```
ravi-mosaicml, line 629:
```suggestion
                filepaths, indexed by global rank, corresponding to the checkpoints saved at that time.
```

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+    .. seealso:: The documentation for the :class:`.Trainer`.
+
+    Args:
+        model (ModelHparams): Hparams for constructing the model to train.
+
+            .. seealso:: :mod:`composer.models` for models built into Composer.
+        train_dataset (DatasetHparams): Hparams used to construct the dataset used for training.
+
+            .. seealso:: :mod:`composer.datasets` for datasets built into Composer.
+        train_batch_size (int): The batch size to use for training.
```
ravi-mosaicml, line 123: Can you elaborate that this should be the optimziation batch size?

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+
+            .. seealso:: :mod:`composer.optim.scheduler` for the different schedulers built into Composer.
+        device (DeviceHparams): Hparams for constructing the device used for training.
+            (default: ``CPUDeviceHparams``)
+        grad_accum (int, optional): See :class:`.Trainer`.
+        grad_clip_norm (float, optional): See :class:`.Trainer`.
+        validate_every_n_batches (int, optional): See :class:`.Trainer`.
+        validate_every_n_epochs (int, optional): See :class:`.Trainer`.
+        compute_training_metrics (bool, optional): See :class:`.Trainer`.
+        precision (str or Precision, optional): See :class:`.Trainer`.
```
ravi-mosaicml, line 163: yahp doesn't support unions like this...

```suggestion
        precision (Precision, optional): See :class:`.Trainer`.
```

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+        scale_schedule_ratio (float, optional): See :class:`.Trainer`.
+        use_stepwise_schedulers (bool, optional): See :class:`.Trainer`.
+        dist_timeout (float, optional): See :class:`.Trainer`.
+        ddp_sync_strategy (str or DDPSyncStrategy, optional): See :class:`.Trainer`.
```
ravi-mosaicml, line 167:
```suggestion
        ddp_sync_strategy (DDPSyncStrategy, optional): See :class:`.Trainer`.
```

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+        seed (int, optional): See :class:`.Trainer`.
+        deterministic_mode (bool, optional): See :class:`.Trainer`.
+        loggers (List[LoggerCallbackHparams], optional): Hparams for constructing the destinations
+            to log to. (default: ``[]``)
+
+            .. seealso:: :mod:`composer.loggers` for the different loggers built into Composer.
+        log_level (str): The Python log level to use. (default: ``INFO``)
```
ravi-mosaicml, line 174:
```suggestion
        log_level (str): The Python log level to use for log statements in the :mod:`composer` module. (default: ``INFO``)
```

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+        callbacks (List[CallbackHparams], optional): Hparams to construct the callbacks to
+            run during training. (default: ``[]``)
+
+            .. seealso:: :mod:`composer.callbacks` for the different callbacks built into Composer.
+        load_path (str, optional): See :class:`.Trainer`.
+        load_object_store (ObjectStoreProvider, optional): See :class:`.Trainer`.
+        load_weights_only (bool, optional): See :class:`.Trainer`.
+        load_chunk_size (int, optional): See :class:`.Trainer`.
+        save_folder (str, optional): See :class:`.Trainer`.
+        save_interval (str or int, optional): See :class:`.Trainer`.
```
ravi-mosaicml, line 186:
```suggestion
        save_interval (str, optional): See :class:`.Trainer`.
```

composer/trainer/trainer_hparams.py
```diff
@@ -107,10 +109,105 @@
...
+        save_compression (str, optional): See :class:`.Trainer`.
+        train_subset_num_batches (int, optional): See :class:`.Trainer`.
+        eval_subset_num_batches (int, optional): See :class:`.Trainer`.
+        profiler_trace_file (str, optional): See :class:`.Trainer`.
+        prof_event_handlers (List[ProfilerEventHandler], optional): See :class:`.Trainer`.
```
ravi-mosaicml, line 191:
```suggestion
        prof_event_handlers (List[ProfilerEventHandlerHparams], optional): See :class:`.Trainer`.
```

ravi-mosaicml (commented):
I think this is the last round of comments I'll have -- otherwise looks great! Hopefully we can merge in tomorrow or Thrusday.

composer/trainer/devices/device_gpu.py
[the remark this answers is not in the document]
ajaysaini725: Discussed offline - not adding an example here because it's a very simple use case + no clean way to run gpu tests in doctest

composer/trainer/devices/device_cpu.py
ajaysaini725: Discussed offline - not including an example due to simplicity of the use case + the fact that DeviceGPU can't have one due to not being able to run doctest with GPUs

composer/loggers/file_logger.py
```diff
@@ -43,6 +43,15 @@ class FileLogger(LoggerCallback):
                 loggers=[logger]
             )
 
+        .. testcleanup::
```
ajaysaini725, line 46: The file logger test was writing a file that never got removed so when you ran `make doctest` multiple times it failed saying the file already exists

ajaysaini725:
Seems like the one failing test is due to a timeout: https://ci.int.mosaicml.com/job/scratch/job/command/7160/ although I may be misinterpreting

composer/loggers/file_logger.py
ravi-mosaicml: damn good catch

composer/trainer/ddp.py
```diff
@@ -10,6 +12,8 @@
 from composer.utils import dist
 from composer.utils.string_enum import StringEnum
 
+__all__ = ["DDPSyncStrategy", "_ddp_sync_context", "_prepare_ddp_module"]
```
ravi-mosaicml, line 15: private methods need not be in `__all__`; you can still import them directly

composer/loggers/file_logger.py
```diff
@@ -43,6 +43,15 @@ class FileLogger(LoggerCallback):
...
+
+            import os
+            from composer.utils.run_directory import get_run_directory
+
+            path = os.path.join(get_run_directory(), "log.txt")
+            if os.path.exists(path):
+                os.remove(path)
```
ravi-mosaicml, line 53: we are very unlikely to encounter this, but would prefer a try / except here to avoid the race condition

ravi-mosaicml (approved):
LGTM! Awesome work here; two minor nits but otherwise good to go

ravi-mosaicml:
Red button merged due to a flaky jenkins gpu test.

PR #524, opened 2022-02-18
Unify augmentation API part 1

dblalock:
This PR partially addresses #343 by allowing the functional API for some of the augmentation functions to handle PIL images, single images as tensors, or batches of images as tensors, as per our design doc. The most concrete usability issue this solves is users having to add some augmentations before the `ToTensor` transform, while others had to be called after `ToTensor`.

To keep the size down, this PR only alters Augmix, Colout, Cutout, and Randaugment, since these are the augmentations that only take in images and not their labels.

The most significant change is that `colout_image` is now gone. No other algorithm had both a `whatever_image` and `whatever_batch` version, so this is now consistent with the rest of the functional API.

I'm especially open to feedback on where to put the shared functions I used to make these changes; made a new `algorithms/utils/augmentation_common.py` and `test/algorithms/test_augmentations_functional.py` since these seemed like the least bad places to put this code.


composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+
+import torch
+from PIL.Image import Image as PillowImage
+from torchvision import transforms
+
+_InputImgT = TypeVar("_InputImgT", torch.Tensor, PillowImage)
+_OutputImgT = TypeVar("_OutputImgT", torch.Tensor, PillowImage)
+
+
+def image_as_type(image: _InputImgT, need_type: Type[_OutputImgT]) -> _OutputImgT:
```
hanlint, line 13: `need_type` = `type`

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+    """Creates a copy of an image-like object with a different type.
+
+    Args:
+        image: a tensor or PIL image. Batches of images in NCHW format are
```
hanlint, line 17: docstrings say batch is supported but then below there is a check?

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+        need_type: type of the copied image
+
+    Returns:
+        A copy of ``image`` with type ``need_type``.
+
+    Raises:
+        RuntimeError: if ``image`` is batch of images, but ``need_type`` is
+            a PIL image. We could convert to a batch of PIL images, but we
+            instead fail fast since this is not expected behavior at any
+            call sites (within data augmenation functions).
```
hanlint, line 28: This fast fail silently enforces a design decision that is elsewhere. If the conversion is possible and correct within the scope of this method (e.g. moving this to as_type), we should allow it.

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+import torch
+from PIL.Image import Image as PillowImage
+from torchvision import transforms
+
+_InputImgT = TypeVar("_InputImgT", torch.Tensor, PillowImage)
+_OutputImgT = TypeVar("_OutputImgT", torch.Tensor, PillowImage)
+
+
+def image_as_type(image: _InputImgT, need_type: Type[_OutputImgT]) -> _OutputImgT:
+    """Creates a copy of an image-like object with a different type.
```
hanlint, line 14: Example calls? Discussion of expected dimensions?

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+        TypeError: if ``need_type`` is not one of :class:`torch.Tensor` or
+            :class:`PIL.Image.Image`.
+    """
+    if isinstance(image, need_type):
+        return image
+
+    # fail fast if attempting to convert a batch of images to a single PIL image
+    is_batch = isinstance(image, torch.Tensor) and (image.ndim == 4) and (image.shape[0] > 1)
+    if is_batch and issubclass(need_type, PillowImage):
+        raise RuntimeError(f"Could not convert batch of images to PIL image")
```
hanlint, line 38: as mentioned above, suggest removing this

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if not issubclass(need_type, (torch.Tensor, PillowImage)):
+        raise TypeError(f"Only need_type={{torch.Tensor, Image}} is supported; got {need_type}")
+
+    if isinstance(image, PillowImage):
+        return transforms.functional.to_tensor(image)  # PIL -> Tensor
+    return transforms.functional.to_pil_image(image)  # Tensor -> PIL
```
hanlint, line 45: I'm a bit confused -- if I provided my Tensor as PIL and want it converted to Tensor, seems like this would always return PIL?

composer/algorithms/utils/augmentation_common.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def image_typed_and_shaped_like(image: _InputImgT,
```
hanlint, line 48: Do we need this? can't the user just do:
```
new_image = image_as_type(old_image, type=type(reference_image))
```

? What is the need for all the typecasting behavior?

hanlint (commented):
Overall looks good, but would benefit from actually writing out what the augmentation API you are enforcing is, and the expected behavior. I tried to divine by looking at the code, but it's a little unclear. 

something like:

```
All data augmentation functional methods should have the following signature:

def my_augmentation(X: ImgT, **kwargs) -> ImgT:

If the augmentation is meant to run on a GPU, and a PIL image is provided, the method should do XYZ.

If the augmentation uses PIL functions , and a GPU-based Tensor is provided (meaning that extra device to <-> from is being incurred, the method should do XYZ. Or should each method have both implementation for performance reasons?

etc.

File: composer/algorithms/utils/augmentation_common.py (hunk 1 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:13]: I'd prefer not redefining builtins. Changed `need_type` -> `typ` for consistency with `typing.cast`.

File: composer/algorithms/utils/augmentation_common.py (hunk 2 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:17]: Clarified the docstring. The check makes it fail if we ask to convert a batch of >1 image to a PIL image. We could have it convert to a collection of PIL images, but randaug and augmix don't support that (seemed like too muchto add in one PR), so we just fail fast instead.

File: composer/algorithms/utils/augmentation_common.py (hunk 4 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:14]: Clarification: this is a package private function that users are never supposed to call. I refrained from prefixing it with an underscore + added a passable docstring for consistency with out other package-private functions. But (evidently!) this is confusing, so I now prefix this function (and the one below) with leading underscores.

Also moved the file to `algorithms/utils` based on discussion with ravi.

I still can definitely make the docstring nicer if you want, but we should probably do this in lots of other places as well in this case.

File: composer/algorithms/utils/augmentation_common.py (hunk 6 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:45]: in that case, the condition for the `if` is true. Added a comment to clarify

File: composer/algorithms/utils/augmentation_common.py (hunk 7 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:48]: The typecasting is purely to appease pyright; this was a much shorter function without it. Calling `image_as_type` is not sufficient because the shape doesn't get changed.

**Comment by dblalock** [comment]
Actually, most of this design ambiguity (see various above comment threads) stems from trying to make this PR short. I'm just gonna do this the "right" way in another ~100loc and hopefully everything should be good.

File: composer/algorithms/utils/augmentation_common.py (hunk 4 above)
Comment (hanlint) [reply composer/algorithms/utils/augmentation_common.py:14]: If an external user is developing their own methods, wouldn't they want to call this helper function?

**Comment by hanlint** [comment]
I'd recommend putting a short paragraph together defining the API (would be good for our docs eventually anyway) before embarking on a major code change.

File: composer/algorithms/utils/augmentation_common.py (hunk 6 above)
Comment (hanlint) [reply composer/algorithms/utils/augmentation_common.py:45]: my bad misread the code here. no need for an additional comment.

also -- does it matter that you check with `issubclass` but then guard here with `isinstance`?

File: composer/algorithms/utils/augmentation_common.py (hunk 7 above)
Comment (hanlint) [reply composer/algorithms/utils/augmentation_common.py:48]: OK I see, that's unfortunate. Is there a reason we only reshape when `reference_image` is of type `Tensor`?

**Comment by dblalock** [comment]
Here it is in one sentence: All input-only data augmentation functions return images of the same type and (if applicable) tensor rank as their inputs.

The change this entails is letting `randaug` and `augmix` take in {single pil images, single-image tensors, batches of image tensors} like all the other input-only data augmentation functions, instead of special-casing them.

And ideally we rename them accordingly (no special `_image` suffix for just them), though I'll probably punt that to the next PR.

File: composer/algorithms/utils/augmentation_common.py (hunk 4 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py]: yes; I was thinking this is analogous to utility functions like sklearn `_check_X_y` that are often used when building your own estimators but not public, but now it looks like that function [is public](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_X_y.html?highlight=check_x_y). ...So I made it not prefixed with an underscore and gave it a nicer docstring.

Still no usage examples since it's small and straightforward (especially now), but I can add them if you want

File: composer/algorithms/utils/augmentation_common.py (hunk 6 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py:45]: in theory shouldn't, but I made this check for type equality in both cases to avoid weird behavior if a subclass is passed in

File: composer/algorithms/utils/augmentation_common.py (hunk 7 above)
Comment (dblalock) [reply composer/algorithms/utils/augmentation_common.py]: managed to rm this whole function

File: composer/algorithms/utils/augmentation_common.py [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+
+    """
+    if isinstance(image, typ):
+        return image
+    if not typ in (torch.Tensor, PillowImage):
+        raise TypeError(f"Only typ={{torch.Tensor, Image}} is supported; got {typ}")
+
+    if typ is torch.Tensor:
+        return transforms.functional.to_tensor(image)  # PIL -> Tensor
+    return transforms.functional.to_pil_image(image)  # Tensor -> PIL
```
growlix, line 44: It's possible I missed something in your code or your tests to handle this, but when I run `transforms.functional.to_pil_image(image)` on a rank 4 tensor, I get `ValueError: pic should be 2/3 dimensional. Got 4 dimensions.`

growlix (commented):
I think `torchvision.transforms.functional.to_pil_image()` does not accept rank 4 tensors (see comment), but otherwise LGTM!

composer/algorithms/utils/augmentation_common.py
dblalock: yes, it's supposed to fail if it gets a rank 4 tensor. I called this out more explicitly in the Raises section of the docstring

dblalock:
So, assuming I can finally appease jenkins, this should be ready to re-review.

The one janky part here is that `image_as_type` does nothing (rather than throwing) if asked to convert a rank 4 tensor to type tensor. This violates the otherwise clean rule that this function only accepts single images.

I could instead add a couple duplicate lines in each caller to check for rank 4 tensors, but it's DRY-er and looks more readable to let them all just say `img_tensor = image_as_type(img, torch.Tensor)`. So my YAGNI senses are telling me to just let this package-private function be slightly inelegant and move on to higher priorities, rather than writing more code.

growlix:
Sorry, I misinterpreted the docstring re: rank 4 tensors.

I'm somewhat inclined towards not allowing rank 4 tensors to be passed to prevent useless calls to this function, but I could go either way.

Otherwise, LGTM. LMK if you want help with a blood offering to appease Jenkins.

growlix (approved):

dblalock:
@hanlint  I'm probably going to merge this right after lunch to quit while I'm ahead wrt Jenkins, unless you object. Happy to add changes to the follow-up PR (already planning to unify + expand relevant functional API docstrings).

hanlint:
Sounds good, apologies a bit distracted by other bits today. 

One thing to think about -- if something like randaugment (which runs Pillow) can now take in both `Tensor, PIL Image`, then that allows users to mistakenly use them after the dataloader has returned, which could have major perf impacts as we would be going from `Tensor (GPU)` -> `PIL (host)` -> `Tensor GPU`. How do we protect from that unexpected behavior?

Similarly, if a certain augmentation runs best on GPU, and users may mistakenly put them in `datasets.transform` and get unexpected performance impacts.

That's the downside of this flexibility that i don't now yet how to reoslve.

PR #532, opened 2022-02-18
Rename `selective_backprop` to `select_using_loss`

ravi-mosaicml:
Sphinx wasn't happy with a re-export named the same thing as a submodule. It's not so much a sphinx issue as it was a bad practice since the following import should be equivalent (but they were not):

```python
from composer.algorithms.selective_backprop import selective_backprop
import composer.algorithms.selective_backprop.selective_backprop
```

`make style` also applied a bunch of formatting changes; hence the large PR. Not sure why `make lint` didn't catch these earlier...

Shoutout for @dskhudia for finding this issue!


dskhudia (approved):
LGTM!

dblalock:
I don't really like the name `apply_selective_backprop` because "apply" is used if and only if we're doing model surgery right now. So ends up introducing inconsistency. But LGTM other than that

ravi-mosaicml:
Discussed offline with @dblalock; renamed to `select_using_loss`.

ravi-mosaicml:
Red button merging as all tests passed before and the latest GPU test failed due to a jenkins error. This test has passed before dev was merged in.

PR #535, opened 2022-02-18
Make batchnorm default for UNet

dskhudia:
Instance norm doesn't preserve memory format, i.e., if input is NHWC the output is NWHC. Therefore, instanceNorm pays the cost of NHWC => NCHW while the preceding conv pays the cost of NCHW => NHWC memory format change. 

5.5% improvement in wall_clock_train.

dice (validation): 69.99 (instance norm) 69.33 (batch norm)

### DIFF
--- a/composer/models/unet/unet.py
+++ b/composer/models/unet/unet.py
@@ -107,7 +107,7 @@ class UNet(ComposerModel):
                           strides=strides,
                           dimension=2,
                           residual=True,
-                          normalization_layer="instance",
+                          normalization_layer="batch",
                           negative_slope=0.01)
 
         return model


ravi-mosaicml (approved):
LGTM; from the code perspective. Though maybe wait for an approval from the reserach team?

hanlint:
cc: @bandish-shah and @florescl

bandish-shah:
Changes look good to me but I would prefer @florescl sign off on this since there is a bit of an accuracy regression.

florescl:
@dskhudia would it be possible to do a few more runs to see if the acc regression is just noise?

dskhudia:
@florescl Few more runs with different random seed or the same random seed?

dskhudia:
@florescl 
With instance norm (i.e., without this PR) I see the following dice values for 4 seeds: 68.7, 69.4, 69.61, 70.07

So it seems like the change in dice is within noise margin.

bandish-shah (approved):
LGTM, thanks for running the multi-seed experiment!

florescl (approved):
awesome, thanks!

PR #540, opened 2022-02-18
Update new README

hanlint:
Updates the README with the new copy. 

A few sections that were not ready yet were excluded, and should be added in a later PR.


moinnadeem (approved):
It's a good start, I think I'll clean up the sections that we aren't happy with next week for v0.4

Issue #521, opened 2022-02-18
Logger docstrings cleanup


ravi-mosaicml:
- [x] For https://mosaicml-composer--499.com.readthedocs.build/en/499/api_reference/composer.loggers.html#module-composer.loggers, could you elaborate more on the logger system architecture? I think a flow chart here could be helpful -- e.g. algorithms and callbacks can make calls to the logger, which then routes the calls to the logging backends. For the example logger, would it make sense to also show how flushing may want to be done on batch_end (after some interval) or epoch_end?
- [x] Could you move the example to https://mosaicml-composer--499.com.readthedocs.build/en/499/api_reference/composer.core.logging.base_backend.html#composer.core.logging.base_backend.LoggerCallback, and then reference this example from the module-level docstrings?
- [x] For the file logger filename, can you comment that this is relative to the run directory (and link to `composer.utils.run_directory`?)
- [x] For the hparam classes, can you enumerate the parameters (similar to what you did for the algorithm hparams)? From the docs page, it isn't clear what the values are for each hparams class. (For the individual parameters, I think it's cool to say something like `see FileLogger` rather than repeating the docstring)
- [x] For the in memory logger, what does "Be sure to confirm logger.backends index." mean?
- [x] For the wandb logger, can you have the `run_directory` auto-link to `composer.utils.run_directory`

_Originally posted by @ravi-mosaicml in https://github.com/mosaicml/composer/issues/499#issuecomment-1043735660_

growlix:
For the example, I'm tempted to keep it simple in the docstring and also link to the [logging tutorial](https://www.notion.so/Logging-locked-74196d271b7c43e494ba6cbd5511f58c) (and possibly the source for a LoggingCallback).

ravi-mosaicml:
Fixed in #544

PR #544, opened 2022-02-19
Additional updates to the loggers docstrings

growlix:
For #521


growlix:
Ravi's requests: https://github.com/mosaicml/composer/issues/521#issue-1142186629

growlix:
@ravi-mosaicml:
> could you elaborate more on the logger system architecture?

The module-level docstring now reads :
```
In Composer, algorithms and callbacks can make calls to the :class:`~.logger.Logger`,
which then routes the calls to the appropriate :class:`~.base_backend.LoggerCallback`. The
:class:`~.base_backend.LoggerCallback` does the actual logging, for example to a
file, or Weights and Biases.

See the :class:`~.base_backend.LoggerCallback` documentation for an example of how to
define a custom logger and use it when training.
```

ravi-mosaicml:
This is gre

> @ravi-mosaicml:
> 
> > could you elaborate more on the logger system architecture?
> 
> The module-level docstring now reads :
> 
> ```
> In Composer, algorithms and callbacks can make calls to the :class:`~.logger.Logger`,
> which then routes the calls to the appropriate :class:`~.base_backend.LoggerCallback`. The
> :class:`~.base_backend.LoggerCallback` does the actual logging, for example to a
> file, or Weights and Biases.
> 
> See the :class:`~.base_backend.LoggerCallback` documentation for an example of how to
> define a custom logger and use it when training.
> ```
> 
> TBH I'm not actually sure I understand the logging system well enough to provide that much more detail. Do users ever actually call `Logger()`, or does that only happen internally inside the trainer?

This is great! The trainer is responsible for constructing the logger, so users do not need to call `Logger()`.

composer/loggers/__init__.py
```diff
@@ -2,31 +2,15 @@
 
 """Logging.
 
-The trainer includes a :class:`~.logger.Logger`, which routes logging
-calls to a logger. Each logger inherits from
-:class:`~.base_backend.LoggerCallback`, which inherits from
-:class:`~.callback.Callback`.
+In Composer, algorithms and callbacks can make calls to the :class:`~.logger.Logger`,
+which then routes the calls to the appropriate :class:`~.base_backend.LoggerCallback`s.
```
ravi-mosaicml, line 6:
```suggestion
which then routes the calls to the appropriate :class:`~.base_backend.LoggerCallback`\\s.
```

composer/loggers/logger_hparams.py
```diff
@@ -77,16 +88,18 @@ class WandBLoggerHparams(LoggerCallbackHparams):
...
-        group (str, optional): Weights and Biases group name.
-        name (str, optional): Weights and Biases run name.
-        entity (str, optional): Weights and Biases entity name.
-        tags (str, optional): Comma-seperated list of tags to add to the run.
-        log_artifacts (bool, optional): Whether to log artifacts. Defaults to False.
-        log_artifacts_every_n_batches (int, optional). How frequently to log artifacts.
-            Default: ``100``. Only applicable if ``log_artifacts`` is True.
-
-        extra_init_params (JSON Dictionary, optional): Extra parameters to pass into :func:`wandb.init`.
+        project (str, optional): See :class:`~composer.loggers.wandb_logger.WandBLogger`.
```
ravi-mosaicml, line 91: The project, group, name, entity, and tags fields do not exist on the WandB class, since the initialize_object method dumps these into the init_kwargs. Can you add an actual docstring here for these fields? (Will probably be a bunch of one-liners).

composer/loggers/logger_hparams.py
```diff
@@ -202,8 +215,9 @@ class TQDMLoggerHparams(LoggerCallbackHparams):
     """:class:`~composer.loggers.tqdm_logger.TQDMLogger`
     hyperparameters.
 
-    See :class:`~composer.loggers.tqdm_logger.TQDMLogger`
-    for documentation.
+    Args:
+        config (dict or None, optional):
```
ravi-mosaicml, line 219: Config is an argument for `initailize_object` -- the TQDM Logger Hparams has no parameters.

ravi-mosaicml (approved):
LGTM; see comments.

PR #547, opened 2022-02-19
Fix the run directory uploader when use_procs=True and not using the …

### LINKED ISSUE #495: `RunDirectoryUploader` not uploading files when `use_procs=True`
It looks like there's a bug in the `RunDirectoryUploader`. The following code has two main issues whenever `use_procs` is set to `True` (the default):
```
credentials = {"provider": "GOOGLE_STORAGE",
               "container": "checkpoints-debugging",
               "key_environ": "GCE_KEY",
               "secret_environ": "GCE_SECRET",
               }
hp = ObjectStoreProviderHparams(**credentials)
prefix = f"foo-{str(uuid.uuid4())[:6]}"
uploader = RunDirectoryUploader(hp, object_name_prefix=prefix, use_procs=True)
trainer = Trainer(model=model, 
                  train_dataloader=dataloader, 
                  eval_dataloader=None, 
                  max_duration="10ep", 
                  save_folder="checkpoints",
                  save_interval="1ep",
                  train_subset_num_batches=100,
                  device="gpu",
                  callbacks=[uploader])
trainer.fit()
```
1. A new run directory is created for each worker, probably from [this line](https://github.com/mosaicml/composer/blob/dev/composer/callbacks/run_directory_uploader.py#L230), though I haven't yet verified. 
2. No files get uploaded because each worker is looking for filepaths relative to its run directory. Not sure if that was an intentional change, but it comes from [this line](https://github.com/mosaicml/composer/blob/dev/composer/callbacks/run_directory_uploader.py#L194) 
 
I believe they both stem from the fact that the `run_directory` gets chosen based on the timestamp when run_directory.py  is first imported. When using multi-processing, this causes each worker to get its own `run_directory`  path.

ravi-mosaicml:
When using `use_procs=True`, each python spawned subprocess got a different _start_time_str in composer.utils.run_directory. So, each run directory worker was looking in a different folder.

This PR fixes this by not having the run directory upload workers look in the run directory.

Closes #495


anisehsani (approved):
lgtm

PR #549, opened 2022-02-19
ssd-resnet34 on COCO fast; updated

florescl:
Adapted from MLcommons.
Current acc run: https://wandb.ai/mosaic-ml/ssd/runs/quf83hlz?workspace=user-laura

to run: `composer -n 8 examples/run_composer_trainer.py -f composer/models/ssd/hparams.yaml loggers.wandb.name ssd --loggers file wandb --loggers.wandb.project ssd --loggers.wandb.entity mosaic-ml`


ajaysaini725 (changes requested):
Can this PR be split up into a few PRs that are approx 200-400 lines long? 

My suggestion would be to at least make this 2 PRs:
- Add the dataset
- Add the model (adding the model by itself can also be split into 2 PRs - one for the base and one for the rest of the model)

hanlint (commented):
A few comments, looking good!

Landanjs (commented):
Thanks for integrating our previous comments! This is still a lot of code, so I was only able to focus on specific sections. Most of the comments are nits that are up to your discretion.

The major points:
1. You might need to use a sampler in the `DataSpec` like the ImageNet dataset. I'm not sure about the impact of the distributed sampler, but the sampler will allow shuffling (should this be true in `hparams.yaml`?)
2. The `split_dict_fn` may not work as expected, but I may have made a mistake.
3. Hanlin's comment on using `from pycocotools import COCO`
4. Instantiating `COCODetection` and `Encoder` every validate pass may be expensive. Could these be instantiated in `__init__`? Maybe only `inv_map` needs to be an attribute for `COCODetection`?
5. I could be mistaken, but a few objects seem to never be used (`L2Norm`, `LightingNoice`, `lr_warmup`, `ToTensor`). Could these be deleted?

Please let me know if any of my interpretations are wrong or if you would like to discuss any design!

PR #550, opened 2022-02-19
v0.4.0 release candidate

hanlint:
Bump version number for release candidate.


ravi-mosaicml:
The version number needs to be changed in other places too, such as in setup.py and meta.yaml. I would grep for 0.3.1 to ensure that we capture all places where it is referenced.

hanlint:
Closing for #609

PR #551, opened 2022-02-20
[v0.4rc] Allow for batch sizes not divisible by grad accum in sequence length warmup

hanlint:
Removed an offending assert that is not needed for sequence length warmup. Also moved the check for `train_dataloader.batch_size is not None` to the `INIT` event to fail faster.


siriuslee (approved):
LGTM as long as `state.train_dataloader` is always set by `Event.INIT`

moinnadeem (approved):
LGTM!

PR #556, opened 2022-02-20
[v0.4rc] remove unused embeddings from optimizer for Alibi

### LINKED ISSUE #553: [v0.4rc] Alibi algorithm removes parameters without updating optimizer
While running Alibi algorithm with `GPT`, the trainer errors out because of a mismatch between the `model.parameters()` and the optimizer's parameters after Alibi surgery is applied.

This has been root-caused to the alibi's `_zero_and_freeze_expand_position_embeddings` method, which removes the embedding layer's weights and replaced with new zeroed, frozen, and expanded embeddings. However, the old layer weights are not removed from the `optimizer`, leading to a mismatch downstream.

hanlint:
Fixes #553

### DIFF
--- a/composer/algorithms/alibi/alibi.py
+++ b/composer/algorithms/alibi/alibi.py
@@ -9,7 +9,7 @@ import logging
 import math
 from operator import attrgetter
 from types import MethodType, ModuleType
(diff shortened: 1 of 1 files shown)


growlix (approved):
LGTM

PR #580, opened 2022-02-22
Decorator microbatch

mvpatel2000:
For adaptively changing grad_accum, it seems we need to wrap multiple additional things:
- training eval metrics
- AFTER_DATALOADER event because of seq-length-warmup
- [in future] eval code
I think it's cleaner to rewrite this as a decorator, so I'm refactoring, but I want input before we commit to this

Note that this is merging into https://github.com/mosaicml/composer/pull/485 and not dev

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -748,17 +748,18 @@ class Trainer:
             if self.checkpoint_saver and self.checkpoint_saver.should_checkpoint(state=state, event=Event.EPOCH_END):
                 self.checkpoint_saver.save_checkpoint(state=state, seed=self.seed, device=self.device)
 
-    def _dynamic_microbatch_wrapper(self, func: Callable) -> Callable:
+    def _dynamic_microbatch_wrapper(func: Callable) -> Callable:
         """Wraps function to catch CUDA Out of Memory Errors and adaptively change microbatch
         size if enabled to miaximize GPU usage.
         """
         # TODO: switch batch_num_samples with new state variable
         def wrapper(*args, **kwargs):
+            self = args[0]
             rerun = True
             while rerun:
                 try:
                     rerun = False
-                    return func()
+                    return func(*args, **kwargs)
                 except RuntimeError as e:
                     if self.adaptive_grad_accum and "CUDA out of memory" in str(e):
                         # Raise runtime error if training 1 sample at a time still resulted in CUDA out of memory
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -862,7 +862,7 @@ class Trainer:
 
         for microbatch_idx, state.batch in enumerate(microbatches):
             state.batch_num_tokens = self._train_data_spec.get_num_tokens_in_batch(state.batch)
-            state.batch_num_samples = self._train_data_spec.get_num_samples_in_batch(state.batch)
+            micro_batch_num_samples = self._train_data_spec.get_num_samples_in_batch(state.batch)
             is_final_microbatch = microbatch_idx + 1 == len(microbatches)
             sync_context = contextlib.nullcontext() if self.deepspeed_enabled else ddp_sync_context(
                 state, is_final_microbatch, self.ddp_sync_strategy)
@@ -890,7 +890,7 @@ class Trainer:
                 # Likely need to look into the performance impact
                 if not self.deepspeed_enabled:
                     for loss in ensure_tuple(state.loss):
-                        loss.mul_(state.batch_num_samples / current_batch_size)
+                        loss.mul_(micro_batch_num_samples / current_batch_size)
                         total_loss += loss.detach().clone()
 
                 assert state.loss is not None
@@ -907,7 +907,7 @@ class Trainer:
 
                     # This is the same loss scaling and reporting we skipped earlier.
                     for loss in ensure_tuple(state.loss):
-                        loss.mul_(state.batch_num_samples / current_batch_size)
+                        loss.mul_(micro_batch_num_samples / current_batch_size)
                         total_loss += loss.detach().clone()
                 else:
                     for loss in ensure_tuple(state.loss):
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -695,7 +695,7 @@ class Trainer:
                         "trainer/batch_idx": self.state.timer.batch_in_epoch.value,
                     })
 
-                    total_loss = self._train_and_compute_loss(use_grad_scaling, state.batch_num_samples)
+                    total_loss = self._train_and_compute_loss(use_grad_scaling)
 
                     if use_grad_scaling:
                         state.scaler.update()
@@ -755,13 +755,16 @@ class Trainer:
         # TODO: switch batch_num_samples with new state variable
         def wrapper(*args, **kwargs):
             self = args[0]
+            # Skip adaption if not enabled
+            if not self.adaptive_grad_accum:
+                return func
             rerun = True
             while rerun:
                 try:
                     rerun = False
                     return func(*args, **kwargs)
                 except RuntimeError as e:
-                    if self.adaptive_grad_accum and "CUDA out of memory" in str(e):
+                    if "CUDA out of memory" in str(e):
                         # Raise runtime error if training 1 sample at a time still resulted in CUDA out of memory
                         if self.state.grad_accum == self.state.batch_num_samples:
                             raise RuntimeError(
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -771,7 +771,6 @@ class Trainer:
                                 "CUDA out of memory. Train loop failed with an internal microbatch of size 1")
                         else:
                             rerun = True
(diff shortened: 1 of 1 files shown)


mvpatel2000:
Yay look it now works with seq_length_warmup
```
Epoch 0:   0%|          | 0/9000 [00:00<?, ?it/s]                                                                                                            
1 -> 2
Epoch 0:   0%|          | 0/9000 [00:05<?, ?it/s]                                                                                                            
2 -> 4
Epoch 0:   0%|          | 0/9000 [00:05<?, ?it/s]                                                                                                            
4 -> 8
Epoch 0:   0%|          | 0/9000 [00:05<?, ?it/s]                                                                                                           
 8 -> 16
Epoch 0:   0%|          | 0/9000 [00:05<?, ?it/s]                                                                                                           
 16 -> 32
Epoch 0:   0%|          | 0/9000 [00:05<?, ?it/s]                                                                                                            
32 -> 64
Epoch 0:   0%|          | 40/9000 [01:26<4:59:54,  2.01s
```

composer/trainer/trainer.py
```diff
@@ -47,6 +47,35 @@
 
 log = logging.getLogger(__name__)
 
+def _dynamic_microbatch_wrapper(func: Callable) -> Callable:
+    """Wraps function to catch CUDA Out of Memory Errors and adaptively change microbatch
```
mvpatel2000, line 51: Is there a better place to put a wrapper? Maybe a util folder / fn? Making this part of the class is messy because the first argument then has to be self, which doesn't play nicely with decorators. I'm not sure of any workarounds where we can include this in the class, but maybe someone with more decorator knowledge knows how to do that

composer/trainer/trainer.py
```diff
@@ -837,8 +863,7 @@ def _train_batch_inner(self, microbatches: Sequence[Batch]):
         current_batch_size = sum([self._train_data_spec.get_num_samples_in_batch(batch) for batch in microbatches])
 
         for microbatch_idx, state.batch in enumerate(microbatches):
-            state.batch_num_tokens = self._train_data_spec.get_num_tokens_in_batch(state.batch)
-            state.batch_num_samples = self._train_data_spec.get_num_samples_in_batch(state.batch)
```
mvpatel2000, line 841: Cleared renaming this to `micro_batch_num_samples` with Ravi. It's not used anywhere and it makes the `state` object confusing, plus it messes w the decorator

mvpatel2000:
Not bothering about lint etc for now, will clean it all up together in https://github.com/mosaicml/composer/pull/485

composer/trainer/trainer.py
jbloxham: It would be reasonable to just create a new file in `composer/trainer/` to hold dynamic grad accum related stuff. Even if it's just this one function.

composer/trainer/trainer.py
```diff
@@ -47,6 +47,35 @@
 
 log = logging.getLogger(__name__)
 
+def _dynamic_microbatch_wrapper(func: Callable) -> Callable:
+    """Wraps function to catch CUDA Out of Memory Errors and adaptively change microbatch
+    size if enabled to miaximize GPU usage.
+    """
+    def wrapper(*args, **kwargs):
+        self = args[0]
```
jbloxham, line 55: Do you know if there's a better way to handle these types? Can `_dynamic_microbatch_wrapper` be defined specifically as a wrapper for methods of a given type? Maybe something in `functools` can help you?

composer/trainer/trainer.py
```diff
@@ -47,6 +47,35 @@
...
+                return func(*args, **kwargs)
+            except RuntimeError as e:
+                if "CUDA out of memory" in str(e):
+                    # Raise runtime error if training 1 sample at a time still resulted in CUDA out of memory
+                    if self.state.grad_accum == self.state.batch_num_samples:
+                        raise RuntimeError(
+                            "CUDA out of memory. Train loop failed with an internal microbatch of size 1")
+                    else:
+                        rerun = True
+                        self.state.grad_accum = min(2 * self.state.grad_accum, self.state.batch_num_samples)
```
jbloxham, line 72: Side thought, this would be a good place to check that `batch_num_samples` is still divisible by `grad_accum`. If not, can raise a `RuntimeError` telling the user this feature is only support for powers-of-two-sized batches.

composer/trainer/trainer.py
```diff
@@ -752,7 +776,27 @@ def _train_loop(self) -> None:
...
+
+        Args:
+            train_metrics (Metric): Training metrics
+        """
+        with torch.no_grad():
+            for eval_microbatch in self._train_data_spec.split_batch(self.state.batch, self.state.grad_accum):
+                # TODO: Detect if self.run_event(Event.AFTER_DATALOADER) changes the training
+                # data and if so print a warning that metrics may return unexpected results
+                outputs, targets = self.original_model.validate(eval_microbatch)
+                train_metrics.update(outputs, targets)
```
jbloxham, line 796: I don't think I'm on board with moving wrapping these methods too. I know that `AFTER_DATALOADER` can OOM if we're using seq_len_warmup, but I don't think fixing that is worth the complexity of added calls to this wrapper. seq_len_warmup can always be fixed by moving the dummy pass into something like `BEFORE_FORWARD`.

jbloxham: But thoughts @ajaysaini725 @hanlint ? Essentially the question here is what scope of OOMs should dynamic grad accum attempt to capture - algorithms can potentially cause large memory consumption in almost arbitrary parts of the training loop.

jbloxham (approved):
On the whole I like the idea of moving this to a decorator! The only thing I'm unsure about is whether we should be using it in so many places. I'm hoping to get feedback from some other team members.

jbloxham:
Also, feel free to merge this whenever - any and all comments can just be addressed on the other PR.

composer/trainer/trainer.py
hanlint: Agreed that we don't want to extend this to theese other functions. I'm OK keeping the scope of dynamic GA tight, and then expanding it later.

composer/trainer/trainer.py
ajaysaini725: Can this just be written as `def wrapper(self, *args, **kwargs)`? Having `self = args[0]` without any safety checks is not great style

composer/trainer/trainer.py
```diff
@@ -47,6 +47,35 @@
...
+        # Skip adaption if not enabled
+        if not self.adaptive_grad_accum:
+            return func
+        rerun = True
+        while rerun:
+            try:
+                rerun = False
+                return func(*args, **kwargs)
+            except RuntimeError as e:
+                if "CUDA out of memory" in str(e):
```
ajaysaini725, line 65: Maybe put this in a function called `def _is_cuda_oom` just to minimize hardcoding in this function

composer/trainer/trainer.py
ajaysaini725: Same - it's not worth adding hacks like this to the trainer code just to enable dynamic grad accum to work in an edge case like this.

My proposal is to do one of:
- Abandon dynamic grad accum for now and just tune grad accum before the start of the training loop (after `Event.INIT` though so it can at least take into account model surgery). In this case, if the user is doing auto grad accum and also has methods that fire on anything other than `Event.INIT`, emit a warning that changing memory usage may break auto grad accum and that they should manually set grad accum in this case,
- Keep dynamic grad accum but emit a warning if any algorithms that fire on something other than `Event.INIT` are being used - in the warning can just say that there are known issues with algorithms that change memory and one should not use auto grad accum in this setting.

For now it's better to have a clean implementation that works in most cases rather than a messy implementation that handles all the known edge cases. Once we have people using this feature, we can come up with better ways of handling the edge cases.

composer/trainer/trainer.py
ajaysaini725: Small nit: let's use `microbatch` instead of `micro_batch` to be consistent with how it's named everywhere else in the codebase

mvpatel2000:
Super helpful feedback! Thanks everyone for chiming in. It seems like there's a few takeaways:
1. Decorators here aren't great b/c of state changes. Agree with this
2. Decompose the trainer loop into more fns
3. I like Ravi's recursive structure -- will switch to that instead of while loop
4. Keep it a little less general: wrap the core training loop but not engine calls
5. I like Ajay's suggestion: lets do dummy forward/backward passes on init to start the process (avoids some edge cases like seq warmup) and spit out a warning 

Sadly it looks like this won't be as general as originally hoped :'( but hopefully still useful

composer/trainer/trainer.py
```diff
@@ -47,6 +47,35 @@
...
+                    # Raise runtime error if training 1 sample at a time still resulted in CUDA out of memory
+                    if self.state.grad_accum == self.state.batch_num_samples:
+                        raise RuntimeError(
```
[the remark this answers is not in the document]
mvpatel2000: Would prefer to specify autograd failed -- I think that's more helpful?

PR #581, opened 2022-02-22
[v0.4.0] Matthew/swa scheduler fix

growlix:
[W&B runs here](https://wandb.ai/mosaic-ml/matthew-swa-test?workspace=user-mleavitt)
Major changes:
- `swa_end` hparam: sets an end `duration` for SWA. If `swa_end` is prior to the final epoch, this allows the SWA/averaged model to have its BN statistics updated without a separate/additional pass through the training data. It also has the side effect that the SWA model is "fine-tuned".
- `schedule_swa_lr` hparam: a boolean that sets whether to use an swa-specific LR. While an swa-specific LR schedule is part of the original paper, my experiments showed that performance improved without it, so the default is not to use it.
- `anneal_strategy` hparam: `schedule_swa_lr` is `True`, sets the type of schedule to use (linear or cosine).
- Added docstrings


composer/algorithms/swa/swa.py
```diff
@@ -30,8 +30,9 @@ class SWA(Algorithm):
     memory required doubles, however, since stored activations and the
     optimizer state are not doubled.
 
-    This algorithm runs on :attr:`~composer.core.event.Event.EPOCH_END` if training
-    duration >= `swa_start`.
+    Uses PyTorch's `torch.optim.swa_util
```
ravi-mosaicml, line 33: Prefer to use intersphinx cross-reference notation if that works.

```suggestion
    Uses PyTorch's :mod:`torch.optim.swa_util`
```

composer/algorithms/swa/swa.py
```diff
@@ -18,7 +18,7 @@
 
 
 class SWA(Algorithm):
-    """Implements Stochastic Weight Averaging (`Izmailov et al, 2018 <https://arxiv.org/abs/1803.05407>`_).
+    """Apply Stochastic Weight Averaging (`Izmailov et al., 2018 <https://arxiv.org/abs/1803.05407>`_)
```
ravi-mosaicml, line 21: Periods in the summary line have been causing issues with the docformatter FYI.

composer/algorithms/swa/swa.py
```diff
@@ -41,43 +42,79 @@ class SWA(Algorithm):
...
-                max_duration="1ep",
+                max_duration="10ep",
                 algorithms=[swa_algorithm],
                 optimizers=[optimizer]
             )
 
     Args:
-        swa_start (float): Fraction of training completed before stochastic weight
-            averaging is applied. Defalt = ``0.8``.
+        swa_start (float, optional): Fraction of training completed before stochastic
```
ravi-mosaicml, line 58: Should these be time strings instead of floats?

composer/algorithms/swa/swa.py
```diff
@@ -41,43 +42,79 @@ class SWA(Algorithm):
...
+        # Check annealing_strategy string
+        if self.anneal_strategy.lower() in ["linear", "lin"]:
+            self.anneal_strategy = "linear"
+        elif self.anneal_strategy.lower() in ["cos", "cosine"]:
+            self.anneal_strategy = "cos"
+        else:
+            raise ValueError("Parameter 'anneal_strategy' must have an argument that is one of {'linear', 'cos'}.")
 
         assert 0 < swa_start < 1, "swa_start must be between 0 and 1."
+        assert swa_end <= 1, "swa_end must be ≤ 1"
```
ravi-mosaicml, line 106: Can you use ValueErrors for validation?

composer/algorithms/swa/swa.py
```diff
@@ -96,22 +133,22 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
-            state.model = self.swa_model
-            log.info('Updated BN and set model to the averaged model')
+        if self.schedule_swa_lr:
+            if self.swa_scheduler is None:
+                raise ValueError('SWA LR scheduler was not set.')
+            self.swa_scheduler.step()
+
+        # End of swa window
+        if float(state.get_elapsed_duration()) >= self.swa_end:
+            assert type(self.swa_model.module) == type(state.model)
```
ravi-mosaicml, line 151: Should this be a ValueError or RuntimeError? (I presume this check is for deepspeed?)

ravi-mosaicml (approved):
LGTM; see comments.

composer/algorithms/swa/swa.py
growlix: I tried that at first but it doesn't render a link, possibly because `torch.optim.swa_util` isn't actually documented, it just exists in tutorial form.

composer/algorithms/swa/swa.py
```diff
@@ -41,43 +42,79 @@ class SWA(Algorithm):
...
+        self.swa_end = swa_end
+        self.schedule_swa_lr = schedule_swa_lr
+        self.anneal_strategy = anneal_strategy
         self.anneal_epochs = anneal_epochs
         self.swa_lr = swa_lr
         self.swa_model: Optional[torch.nn.Module] = None
+        self.swa_completed = False
+
+        # Check annealing_strategy string
+        if self.anneal_strategy.lower() in ["linear", "lin"]:
```
hanlint, line 98: perhaps just support `linear` and `cosine` , and not the shorthand abbreviations?

hanlint (approved):

composer/algorithms/swa/swa.py
growlix: Wooo, go time abstraction!

composer/algorithms/swa/swa.py
growlix: I'm actually not sure why this is here. It was a pre-existing part of the code, and I believe it predates deepspeed

composer/algorithms/swa/swa.py
growlix: `SWALR` (the SWA LR scheduler) accepts `cos` or `linear`. We could have it match PT exactly, but I think it's weird to strictly enforce one being abbreviated and the other not.

hanlint: ah ok sounds good.

composer/algorithms/swa/swa.py
hanlint: yeah that's an odd check.. sound like ti should be part of a `test_swa` instead of in this code?

hanlint (approved):

composer/algorithms/swa/swa.py
florescl: in my original implementation this check was needed to make sure we did not change the type of the Model to AveragedModel (from torch.optim.swa_utils.AveragedModel)

composer/algorithms/swa/swa.py
```diff
@@ -96,22 +161,25 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
 
-        if self.swa_scheduler is None:
-            raise ValueError('SWA LR scheduler was not set.')
-        self.swa_scheduler.step()
-
-        ## end of training
-        if float(state.get_elapsed_duration()) >= 1.0:
-            device = next(self.swa_model.parameters()).device
-            # TODO(laura) this does not apply the batch split fn. This may result in cuda OOM
-            update_bn(state.train_dataloader, model=self.swa_model, device=device)
```
florescl, line 115: I think we still need to have a call to update_bn if we're doing SWA all the way until the end, no? @growlix

growlix: `update_bn()` does a full epoch of forward passes on the data internally by iterating through `train_dataloader`, which—I think—makes it basically incompatible with Composer because it can't account for microbatching and all the other logic the trainer does in between a single iteration through `train_dataloader` and actually doing a forward pass. @jbloxham and @abhi-mosaic would probably have a better idea than me though

growlix: For now maybe we should just spit a warning if they set `swa_end="1dur"` or something similar

Issue #224, opened 2022-01-13
Remove `max_epochs` from progressive resizing


ravi-mosaicml:
Let's update progressive resizing to use the new `state.timer` and the current training duration instead of relying on `state.max_epochs`

PR #583, opened 2022-02-23
per #224, update function to use Timer and Time

jzf2101:
Let me know if this is in line with the changes you're referring to @ravi-mosaicml


composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -170,7 +170,7 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
         # Calculate the current size of the inputs to use
         initial_size = self.initial_scale
         finetune_fraction = self.finetune_fraction
-        scale_frac_elapsed = min([(state.epoch / state.max_epochs) / (1 - finetune_fraction), 1])
+        scale_frac_elapsed = min([(state.timer._epoch / state.max_duration)._value / (1 - finetune_fraction), 1])
```
growlix, line 173: Wooo, first PR and code review! 🚀 🎆 

I think we try to avoid using private attributes outside classes/modules. Accordingly `state.timer._epoch` can be replaced with `state.timer.get("ep")`. And `._value` -> `.value`.

But but we don't need to make either of those changes here, because `(state.timer._epoch / state.max_duration)` is a computation we use enough—and a computation that we don't want bound to a fixed unit of time—that it has its own method: `state.get_elapsed_duration()`.

Cool aside: our time objects support arithmetic and comparison, so

```
scale_frac_elapsed = min([state.get_elapsed_duration() / (1 - finetune_fraction), 1]).value
```
and
```
scale_frac_elapsed = min([state.get_elapsed_duration().value / (1 - finetune_fraction), 1])
```
both work (Although subtraction and addition require identical units (e.g. "ep" (epoch), "ba" (batch), "dur" (duration))).

The safer thing to do is probably call `.value` ASAP, so I think the first example is preferable.

growlix (changes requested):
Yay first review and PR! I left a comment.

@ravi-mosaicml, should the `finetune_fraction` argument be changed from a float to a time string (e.g. 0.2 - "0.2dur")? I feel like no, because it's "duration before end" as opposed to "duration from start).

composer/algorithms/progressive_resizing/progressive_resizing.py
hanlint: Unrelated question.. is it better if we let state have the property for `epoch` that calls into `state.timer`? Users may be unaware of how `timer` works and really just want the epochs without too much spelunking in our docs.

jzf2101: I think this may be the easier way to do it imho 🤷

growlix: Agree, I think we should have both `Timer.get({unit})` and `Timer.{unit}`

growlix (approved):
LGTM!

jzf2101:
closes #224

PR #584, opened 2022-02-23
Update section titles

hanlint:
Putting this PR up for consideration


ravi-mosaicml (approved):
Love it, though could we use the `|:emoij_name:|` notation for all emoijs instead of hardcoding the literal emoij into the code?

hanlint:
Unfortunately, the markdown files do not support |:emoji_name:|, and even `:emoji_name:` doesn't render properly.

PR #587, opened 2022-02-23
Clean up Transformer models __init__ function

moinnadeem:
It doesn't really make sense for a ComposerTransformer to accept a `transformers.PreTrainedModel`, but a `tokenizer_name: str`. This interface was just too tightly coupled to YAHP.

This remedies the problem by the `__init__` function taking in a `transformers.PreTrainedTokenizer` instead of a `str`.

I also welcome any other suggestions for quality-of-life fixes!


hanlint (commented):
LGTM

hanlint (approved):

Issue #226, opened 2022-01-13
Remove old timing calls from Sequential Length Warmup


ravi-mosaicml:
Let's update sequential length warmup to use the timing abstraction (#146).

Specifically, all calls to:
- `state.epoch`
- `state.step`
- `state.batch_idx`
- `state.max_epochs`
- `state.steps_per_epoch`
 
should be replaced with:
- `state.timer` (and it's methods)
- `state.max_duration`
- `state.get_elapsed_duration()`

As an example, please see #221.

ravi-mosaicml:
For the 0.4 release, Sequential Length Warmup will support only epochs and batches.

ravi-mosaicml:
Closing this issue as it will likely not be implemented. Feel free to re-open if there is a use case for it.

PR #594, opened 2022-02-24
Remove all old timing calls

### LINKED ISSUE #146: Time Abstraction

### LINKED ISSUE #229: Remove old timing calls from stochastic depth
Let's update stochastic depth to use the timing abstraction (#146).

Specifically, all calls to:
- `state.epoch`
- `state.step`
- `state.batch_idx`
- `state.max_epochs`
- `state.steps_per_epoch`
 
should be replaced with:
- `state.timer` (and it's methods)
- `state.max_duration`
- `state.get_elapsed_duration()`

As an example, please see #221.

### LINKED ISSUE #512: Missing functionality for specifying time in units other than epochs
`NotImplementedError: Max duration must be specified in epochs. Other units are not yet supported.`

Is this supposed to be happening? It's not currently mentioned in the docs. Perhaps this is part of the updated time abstraction that's coming.

ravi-mosaicml:
1. Remove all uses of `state.max_epochs`; replaced with `state.max_duration` -- specifically in stochastic depth (#229) and sequential length warmup (#226). However, it does not fix the latter, as we still need to update the algorithm to support max_duration in terms of tokens or samples.
2. Replaced all `state.step` with `int(state.timer.batch)`
3. Replaced all `state.epoch` with `int(state.timer.epoch)`
4. Replaced all `state.batch_idx` with `int(state.timer.batch_in_epoch)`
5. Removed the constraints that max_duration be specified in epochs; it can now be in any unit. Added test cases for specify max duration in batches.
6. Added a helper method to the `Time` class to convert it to a timestring (with a test case)

Closes #146
Closes #229
Closes #512


composer/callbacks/run_directory_uploader.py
```diff
@@ -40,15 +40,28 @@ class RunDirectoryUploader(Callback):
     timestamp. Only files that have a newer last modified timestamp since the last upload will be  uploaded.
 
     Example
-        .. testsetup:: *
+        .. testsetup::
+
+           import os
+           import functools
+           from composer.callbacks import RunDirectoryUploader, run_directory_uploader
```
ravi-mosaicml, line 47: This doctest was never working properly, but because it spawned subprocesses, it had a race condition that allowed the doctest build to pass. This fix is unrelated to this PR, but including it here since it was causing the doctest to fail occasionally when testing locally.

composer/__init__.py
```diff
@@ -1,14 +1,5 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-from composer import algorithms as algorithms
```
ravi-mosaicml, line 3: Having these re-exports here broke the `from composer import *` in the doctest fixtures

hanlint (commented):
Hmm would it make sense to attach to state the epoch, etc. properties? That would reduce the code verbosity here and also be easier for developers to access directly.

composer/__init__.py
ajaysaini725: Random thought - should we have a doctest here to test for `from composer import ____` imports that we expect to work (that way if someone changes something within a module like callbacks that somehow breaks importing directly from Composer we catch it here)

composer/algorithms/stochastic_depth/stochastic_depth.py
```diff
@@ -130,9 +131,9 @@ class StochasticDepth(Algorithm):
             ``"uniform"`` assigns the same ``drop_rate`` across all layers.
             ``"linear"`` linearly increases the drop rate across layer depth
             starting with 0 drop rate and ending with ``drop_rate``.
-        drop_warmup: Percentage of training epochs to linearly
-            increase the drop probability to `linear_drop_rate`. Must be between
-            0.0 and 1.0.
+        drop_warmup (str | Time | float): A :class:`Time` object, time-string, or float
```
ajaysaini725, line 134: Since you're editing this can you also add typing to the other args here?

composer/core/time.py
```diff
@@ -200,6 +200,22 @@ def unit(self) -> TimeUnit:
...
 
+    def __str__(self) -> str:
+        return f"{self.value}{self.unit.value}"
+
+    def to_timestring(self):
+        """Get the time-string representation.
+
+        For example:
+
+        >>> Time(5, TimeUnit.EPOCH).to_timestring()
```
ajaysaini725, line 211: Does this need `.. doctest::` syntax?

tests/trainer/test_scale_schedule.py
```diff
@@ -85,7 +86,8 @@ def test_epochs_scaled(self, ssr: float, use_algorithm: bool, composer_trainer_h
             composer_trainer_hparams.scale_schedule_ratio = ssr
         trainer = composer_trainer_hparams.initialize_object()
 
-        assert trainer.state.max_epochs == int(10 * ssr)
+        assert trainer.state.max_duration.unit == TimeUnit.EPOCH
```
ajaysaini725, line 89: Add a test here to test that `1ep` is also scaled correctly by SSR (i.e. your change in the `_convert_time` function?

ajaysaini725 (commented):
Overall looks great just a few comments

ravi-mosaicml:
> Hmm would it make sense to attach to state the epoch, etc. properties? That would reduce the code verbosity here and also be easier for developers to access directly.

Unfortunately, this would lead to naming conflicts, since `batch` is already defined on state. If we bound it directly on the state, then we would need to use a different name on the state object -- e.g. `batch_idx`, `sample_idx`. I think this could lead to confusion if it's called `state.epoch_idx` on the state and `timer.epoch` on the timer (it also wouldn't be that much less verbose -- saves two characters now!)

I could rename it in both places, but am not a huge fan of having everything in the `timer` be `_idx`'ed.

Alternatively, we could rename `batch` to `batch_data`? Also not a huge fan of that.

Agree it would be nice if these properties were more visible, but a good solution doesn't come to mind. Thoughts?

ravi-mosaicml:
> Random thought - should we have a doctest here to test for from composer import ____ imports that we expect to work (that way if someone changes something within a module like callbacks that somehow breaks importing directly from Composer we catch it here)

Any import from a submodule will attempt to load the parent module's `__init__.py`, so our tests already check to ensure that the parent's `__init__.py` is bug-free. As for ensuring that, say, `composer.callbacks` doesn't have a bug in it, we should use the highest-level import whenever possible (e.g. `from composer import Trainer`, not `from composer.trainer.trainer import Trainer`) in the tests. But going to leave that out of this PR as that would be a larger change.

composer/algorithms/stochastic_depth/stochastic_depth.py
ravi-mosaicml: https://github.com/mosaicml/composer/pull/602 added those in.

composer/core/time.py
ravi-mosaicml: If there is no `..testsetup` needed, then no ``..doctest`` is needed -- you can do the `>>>` directly.

hanlint:
Good point on the naming conflicts -- let's keep them in `state.timer` for now then.

hanlint (approved):

hanlint (changes requested):
For the method cards, the actual READMEs should sit in the `composer/algorithms/` folder, with a symlink in the `docs/source`.

ravi-mosaicml:
> For the method cards, the actual READMEs should sit in the `composer/algorithms/` folder, with a symlink in the `docs/source`.

Whoops, good catch. Just fixed this.

hanlint (approved):

composer/__init__.py
ravi-mosaicml: (pasting reply from the conversation)

Any import from a submodule will attempt to load the parent module's __init__.py, so our tests already check to ensure that the parent's __init__.py is bug-free. As for ensuring that, say, composer.callbacks doesn't have a bug in it, we should use the highest-level import whenever possible (e.g. from composer import Trainer, not from composer.trainer.trainer import Trainer) in the tests. But going to leave that out of this PR as that would be a larger change.

composer/trainer/trainer.py
```diff
@@ -974,8 +974,15 @@ def _train_loop(self) -> None:
                     if self._checkpoint_saver and self._checkpoint_saver.should_checkpoint(state=state,
                                                                                            event=Event.BATCH_END):
                         self._checkpoint_saver.save_checkpoint(state=state, seed=self._seed, device=self._device)
+
+                    if state.timer >= state.max_duration:
+                        # If max_duration is specified in batches, samples, or tokens, and the batch
```
ravi-mosaicml, line 979:
```suggestion
                        # If max_duration is specified in batches, samples, or tokens, and
```

PR #595, opened 2022-02-24
Pin pre-1.0 packages

### LINKED ISSUE #593: Pin the pyright version
The environment variable `PYRIGHT_PYTHON_FORCE_VERSION` needs to be set when pyright is installed from pip. Ideally we can set this variable in the setup.py (e.g. `os.environ['PYRIGHT_PYTHON_FORCE_VERSION'] = '...'`); but if not, then it should be in the `.ci/Jenkinsfile`.

bandish-shah:
* Any pre-1.0 Python `extra_requires` packages can silently update and cause issues
* Most recently updated:
   * Pyright 1.1.223 -> 1.1.224: Caused unexpected lint failures in `dev`
   * Sphinx MyST plugin 0.16.1 -> 0.17.0: Changed link resolution syntax
* To avoid surprises, we should start by pinning all pre-1.0 `extra_requires` to current latest versions and manage explicitly 
   * Pin Pyright to version 1.1.224 in Makefile
   * Pin all pre-1.0 `extra_deps` packages in `setup.py` to current latest versions
* No changes to requirements on `install_requires` packages
* Resolves #593


Averylamp:
> [](https://github.com/RobertCraigie/pyright-python#installation)This project is not suitable for production usage.

Note: https://github.com/RobertCraigie/pyright-python
From the source

ravi-mosaicml:
@bandish-shah does pyright read the env variable when it runs or when it installs?

bandish-shah:
> @bandish-shah does pyright read the env variable when it runs or when it installs?

It's read when it runs, doesn't make any difference to have it set during the PyPi package install.

bandish-shah:
> > This project is not suitable for production usage.
> 
> Note: https://github.com/RobertCraigie/pyright-python From the source

Yes I saw this as well, let's patch this issue for now and start a parallel discussion on migrating to a more production ready linter.

ravi-mosaicml:
I tried this locally, but I can't get it to use the version specified in the makefile. @bandish-shah were you able to verify that the version is installing what is specified in the variable?

If not, could we go back to installing pyright via npm (outside of pip)?

bandish-shah:
> I tried this locally, but I can't get it to use the version specified in the makefile. @bandish-shah were you able to verify that the version is installing what is specified in the variable?
> 
> If not, could we go back to installing pyright via npm (outside of pip)?

Hmmm... more going on here than meets the eye.  Thanks for pointing this out, I initially tried this locally in a virtualenv and it was working for me but that doesn't seem to be consistent across different environments.

The testing procedure is as follows:
1. `cd composer; pip install -e .[all]`
2. `for micro in {220..224}; do PYRIGHT_PYTHON_FORCE_VERSION=1.1.${micro} pyright --version; done`

Results testing across environments are as follows:
- Locally in a virtualenv: Version pinning works as expected
- `pytorch_internal` Docker image as `mosaicml` user: Version remains fixed at 1.1.223
- `pytorch_internal` Docker image as `mosaicml` user in a venv: Version remains fixed at 1.1.223
- `pytorch_internal` Docker image as `root` user in a venv: Version remains fixed at 1.1.223

I also repeated the experiments above on our Docker image but uninstalling `nodejs` first and the version is still fixed.  

Only other option now is to pre-install pyright in our Docker images using `npm` and pinning the version that way.

bandish-shah:
Follow up:
Played around with this in both `pytorch` and `pytorch_internal` Docker images.  Tried un-installing and also building a test image without `nodejs` installed and behavior remains the same.  

I also tried pinning the pyright version using `npm` (e.g., `npm install -g pyright@1.1.224`) and this seems to be the only reliable way pin the pyright version in our Docker images.  I'm going to go ahead and update our Docker build to set the pyright version as well using this method as we've already discussed.  In addition to setting it in the Makefile that will cover most of our users, however there's no simple way to guarantee the pyright versions for users running locally.

bandish-shah:
Recent wrapper changes resolved the pyright pinning issue, for now we'll just pin the pyright pip wrapper since new versioning scheme matches pyright version with the pyright-python wrapper version so pinning in setup.py should be sufficient.

This PR now merely pins the other pre-1.0 dependencies.

ravi-mosaicml (approved):
LGTM, thanks!

PR #598, opened 2022-02-24
docstrings improvements for core modules

dskhudia:
For evaluator.py, precision.py, serializable.py, state.py, time.py and types.py in composer.core

### DIFF
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -1,4 +1,7 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
+
+"""A wrapper for a dataloader to include metrics that apply to a specific dataset."""
+
 from __future__ import annotations
 
 import copy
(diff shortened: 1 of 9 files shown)


composer/core/evaluator.py
```diff
@@ -11,15 +14,32 @@
...
+__all__ = ["Evaluator"]
+
 
 class Evaluator:
-    """Wrapper for a dataloader to include metrics that apply to a specific dataset.
+    """A wrapper for a dataloader to include metrics that apply to a specific dataset.
+
+    For example, :class:`~.nlp_metrics.CrossEntropyLoss` metric for NLP models.
+
+    Example,
```
hanlint, line 25:
```suggestion
    Example:
```

composer/core/logging/logger.py
```diff
@@ -1,15 +1,6 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-"""Base classes, functions, and variables for logger.
-
-Attributes:
```
hanlint, line 5: Should we still describe the types upfront? easier than scrolling down into the docs.

composer/core/state.py
```diff
@@ -50,52 +50,43 @@ def null(precision):
...
-
+        * To support multi-GPU training, ``State.model`` may be wrapped in :class:`torch.nn.parallel.DistributedDataParallel`,
+          and the dataloaders may be wrapped in a device-specific dataloader that handles moving tensors to device.
 
+        * :attr:`~.types.Schedulers` are wrapped in :attr:`~.scheduler.ComposerScheduler`, which handles stepping either
+          stepwise or epochwise, and also properly sets up learning rate warmups.
 
     Args:
-        model (types.Model, often ComposerModel): The model, typically as a subclass of :class:`ComposerModel`.
+        model (types.Model): The model, typically as a subclass of :class:`~.base.ComposerModel`.
```
hanlint, line 67: is it less verbose here to do :class:`.ComposerModel` ?

hanlint (approved):
LGTM, one small nit on trying to reduce the verbosity of the comments by doing:
:class:`.ComposerModel` instead of :class:`~.base.ComposerModel`.

composer/core/logging/logger.py
dskhudia: Sure. I can keep them at the top.

composer/core/state.py
dskhudia: Some of these I kept because they weren't linking correctly otherwise.

hanlint:
> How do we usually link non-api references?
You can use `:doc:` directive, e.g. :doc:`Time Guide</trainer/time>`

> Are you suggesting that we get rid of them ?

No, I think describe the plural options. e.g. Schedulers can be pytorch schedulers, our composer schedulers, or a list or tuple of these schedulers.

ravi-mosaicml:
> > Also, I think it would be helpful to link to the [timing guide](https://docs.mosaicml.com/en/v0.4.0/trainer/time.html)
> 
> How do we usually link non-api references? using full path such as https://docs.mosaicml.com/en/v0.4.0/trainer/time.html?

It should be possible to use `:doc:` notation.

> > can you move the serialized_attributes from being a class-level member to something defined on self.?
> 
> We are not mutating `serialized_attributes`

The deepspeed checkpointing logic does modify serialized_attributes so it can skip serialization of the models and optimizers on the state dict.

> > Some types are not autolinking
> 
> I forced link them. They are not getting picked by the Sphinx automatically.

KK, we can address in a separate PR.

> > our custom plural types in the docstring can be confusing. Can we enumerate this out?
> 
> Are you suggesting that we get rid of them ?

Actually, don't worry about this one. It will depend on #640. Generally speaking, we will want to remove the plural types from the docstrings and instead replace it with `x | Sequence[x]` It will probably be better to replace this all at once later.

dskhudia:
@ravi-mosaicml, @hanlint Addressed all comments.

PR #600, opened 2022-02-24
updated docstring examples

growlix:
code-block (not doctest 😢 ) examples for functional and class forms of ALiBi

### DIFF
--- a/composer/algorithms/alibi/alibi.py
+++ b/composer/algorithms/alibi/alibi.py
@@ -42,6 +42,7 @@ def apply_alibi(
     more details.
 
     Example:
+
     .. code-block:: python
 
         import torch.nn.functional as F
@@ -146,6 +147,7 @@ class Alibi(Algorithm):
     See the :doc:`Method Card </method_cards/alibi>` for more details.
 
     Example:
+    
     .. code-block::
 
         from composer.algorithms import Alibi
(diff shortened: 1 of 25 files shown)


dblalock:
can the examples go after the {args, returns, raises}?

growlix:
> can the examples go after the {args, returns, raises}?

I think this is the correct ordering, but given that current convention is the ordering [description, example, args, returns], I'm inclined to abide by current convention here and file a separate PR that changes the convention.

composer/algorithms/alibi/alibi.py
```diff
@@ -34,10 +34,33 @@ def apply_alibi(
     optimizers: Optional[Optimizers] = None,
 ) -> None:
     """Removes position embeddings and replaces the attention function and attention mask
-    according as per :class:`~composer.algorithms.alibi.alibi.Alibi`. See the
-    :doc:`Method Card </method_cards/alibi>` for more details.
+    according as per :class:`~composer.algorithms.alibi.alibi.Alibi`. Note that the
+    majority of the training speed-up from using ALiBi comes from being able to train on
+    shorter sequence lengths; thisfunction does not scale the training sequence length as
```
hanlint, line 39:
```suggestion
    shorter sequence lengths; this function does not scale the training sequence length as
```

hanlint (approved):

PR #601, opened 2022-02-24
Skip ResNet50 + DeepSpeed tests that are timing out

hanlint:
These tests are causing timeout errors on Jenkins, but not reproduced locally. We will need to investigate, but skipping for now to unblock PRs.


dblalock (approved):
passed 2/2 times; LGTM

PR #603, opened 2022-02-25
Export Transforms in `composer.algorithms`

ajaysaini725:
`AugmentAndMixTransform`, `ColOutTransform`, and `RandAugmentTransform` are currently not exported. However, we expect that the usage of these will be more common than the usage of the purely functional form because these transforms are written to be easily composable with other transformations in an image pipeline.

This change exports them in `composer.algorithms` to make them more easily accessible.


hanlint:
cc: @dblalock on this for comment

dblalock (approved):
LGTM

PR #604, opened 2022-02-25
Update progressive resizing to work for semantic segmentation

coryMosaicML:
This PR includes a number of changes to progressive resizing to handle resizing targets and improve logging.

Changes are:
1. Switch from bilinear downsampling to nearest-neighbor downsampling. This allows the tensor of target indices to be resized directly so one-hot/dense labels are not required. 
2. Adds some pre/post processing to dimensions and type of labels when `resize_targets=True` to make shapes and types compatible with the resizing functions. Ideally this should be invisible outsize `resize_batch`.
3. Changes to `resize_batch` to handle different pairs of `mode` and `resize_targets`.
5. Move creation of resizing transforms to separate functions `_make_crop()` and `_make_resize()`.
4. Use `transforms.functional.crop` instead of `transforms.RandomCrop`.
6. Introduces `_make_crop_pair()` to make a pair of crop transforms to crop the same region from  inputs and targets. 
7. Adds logging for the height, width, and scale factor for the input images.

Test runs:

[Before changes ResNet50-i1k mode='crop' Acc 76.05%](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/18iskeys)
[After changes ResNet50-i1k mode='crop' Acc 76.18%](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/3abx8oj8)

[Before changes ResNet50-i1k mode='resize' Acc 76.56%](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/1mvaud97)
[After changes ResNet50-i1k mode='resize' Acc 76.58%](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/2scdmsws)

[After changes DeeplabV3-ADE20k mode='crop'  mIoU 44.55](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/m9s18j34)
[After changes DeeplabV3-ADE20k mode='resize' mIoU 44.96](https://wandb.ai/mosaic-ml/cory-progressive-resizing-fix/runs/3n61j28d)


composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -44,42 +46,61 @@ def resize_batch(X: torch.Tensor,
...
-            resize_transform = partial(F.interpolate, scale_factor=scale_factor, mode='bilinear')
-        else:
-            raise ValueError(f"Progressive mode '{mode}' not supported.")
-        return resize_transform(tensor)
-
-    X_sized = resize_tensor(X)
-    if resize_targets:
-        y_sized = resize_tensor(y)
+    # Prep targets for resizing if necessary
+    if check_for_index_targets(y) and resize_targets is True:
```
siriuslee, line 69: What is the reasoning here? Looks like `check_for_index_targets` just checks the type of the tensor. Why not check the dimensionality?

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -44,42 +46,61 @@ def resize_batch(X: torch.Tensor,
...
+        # Add a dimension to match shape of the input and change type for resizing
+        y_sized = y.float().unsqueeze(1)
     else:
         y_sized = y
+
+    if mode.lower() == "crop" and resize_targets is False:
+        # Make a crop transform for X
+        resize_transform = _make_crop(tensor=X, scale_factor=scale_factor)
+        X_sized, y_sized = resize_transform(X), y
```
siriuslee, line 78:
```suggestion
        X_sized = resize_transform(X)
```
I'm finding the logic behind when you use `y` vs `y_sized` a little complicated and unnecessary.

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -44,42 +46,61 @@ def resize_batch(X: torch.Tensor,
...
+        X_sized, y_sized = resize_transform(X), resize_y(y_sized)
+    elif mode.lower() == "resize":
+        # Make a resize transform (can be used for X or y)
+        resize_transform = _make_resize(scale_factor=scale_factor)
+        X_sized = resize_transform(X)
+        if resize_targets:
+            y_sized = resize_transform(y_sized)
+    else:
+        raise ValueError(f"Progressive mode '{mode}' not supported.")
+        X_sized, y_sized = X, y
```
siriuslee, line 91:
```suggestion
```
This won't be reached since you raised an error

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -181,3 +202,50 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
...
+    topy = int(height_ratio * topX)
+    lefty = int(width_ratio * leftX)
+    # Make the two transforms
+    resize_X = partial(transforms.functional.crop, top=topX, left=leftX, height=HcX, width=WcX)
+    resize_y = partial(transforms.functional.crop, top=topy, left=lefty, height=Hcy, width=Wcy)
+    return resize_X, resize_y
+
+
+def _make_resize(scale_factor: float) -> Callable[[torch.Tensor], torch.Tensor]:
+    resize_transform = partial(F.interpolate, scale_factor=scale_factor, mode='nearest')
```
siriuslee, line 250:
```suggestion
    """Makes a nearest-neighbor interpolation transform at the specified scale factor."""
    resize_transform = partial(F.interpolate, scale_factor=scale_factor, mode='nearest')
```

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -181,3 +202,50 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
...
+
+        if logger is not None:
+            logger.metric_batch({
+                "progressive_resizing/height": new_input.shape[2],
+                "progressive_resizing/width": new_input.shape[3],
+                "progressive_resizing/scale_factor": scale_factor
+            })
+
+
+def _make_crop(tensor: torch.Tensor, scale_factor: float) -> Callable[[torch.Tensor], torch.Tensor]:
```
siriuslee, line 214: Might be nicer to add a
```
T_ResizeTransform = Callable[[torch.Tensor], torch.Tensor]
```
to simplify the return types for these helper functions

siriuslee (changes requested):
This mostly looks good. I added a few suggestions. The only thing I'd like to understand better is the `check_for_index_targets` call. It feels strange to me that not checking dimensions is the right thing to do there. 

Could you devise a target that is `(N, H, W)` but isn't an index? greyscale images?

composer/algorithms/progressive_resizing/progressive_resizing.py
coryMosaicML: The check for dimensionality isn't enough by itself for 3D index targets, as these will have the same dimensionality as 2D non-index targets, but the resizing transforms will fail without the conversion to floats. Progressive resizing should still be able to function in this case, though the third dimension will not be resized. 

I agree this check is not perfect, so I added some explicit checks to exclude some cases where this isn't correct, which coincide with cases where `resize_targets=True` shouldn't be used anyways. In the long run we will need a more robust `check_for_index_targets` but that's not something I wanted to include in this PR.

composer/algorithms/progressive_resizing/progressive_resizing.py
coryMosaicML: This was done so that I could do a single `return X_sized, y_sized` at the end rather than multiple, and I found it easier to read the different parts of the if statement if they could each be read as applying a transform to X, y, which is sometimes the identity transform. Open to other suggestions here though.

composer/algorithms/progressive_resizing/progressive_resizing.py
coryMosaicML: Fixed

composer/algorithms/progressive_resizing/progressive_resizing.py
coryMosaicML: Done

composer/algorithms/progressive_resizing/progressive_resizing.py
coryMosaicML: Fixed

composer/algorithms/progressive_resizing/progressive_resizing.py
siriuslee: That makes sense

siriuslee (approved):
LGTM

PR #611, opened 2022-02-28
Fix Lint

ravi-mosaicml:
1. Fix the run directory uploader doctest. This doctest example was never actually running, but because it was launching subprocesses, it somehow did not trigger a failure exit code. Instead, it would occasionally hang the linting process, triggering a timeout.
2. Fixed pyright issues. Pinned version to 1.1.224.post1.
3. Added doctest groups to all doctests that patch env variables. Groups are needed to ensure that the environments don't conflict.

### DIFF
--- a/composer/datasets/glue.py
+++ b/composer/datasets/glue.py
@@ -16,6 +16,17 @@ from composer.utils import dist
 
 log = logging.getLogger(__name__)
 
+_task_to_keys = {
+    "cola": ("sentence", None),
+    "mnli": ("premise", "hypothesis"),
+    "mrpc": ("sentence1", "sentence2"),
+    "qnli": ("question", "sentence"),
+    "qqp": ("question1", "question2"),
+    "rte": ("sentence1", "sentence2"),
+    "sst2": ("sentence", None),
+    "stsb": ("sentence1", "sentence2"),
+}
+
 
 @dataclass
 class GLUEHparams(DatasetHparams):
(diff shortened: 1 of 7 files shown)


composer/profiler/json_trace.py
```diff
@@ -22,8 +22,8 @@
 
 
 class JSONTraceHandler(ProfilerEventHandler):
-    """Records trace events in `JSON trace format
-    <https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview>`_.
+    """Records trace events in `JSON trace format <https://\\
```
ravi-mosaicml, line 25: needs to be on a seperate line so `make style` doesn't add a period, which would break the link.

hanlint (approved):

composer/datasets/glue.py
```diff
@@ -85,10 +85,10 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
 
         log.info(f"Loading {self.task.upper()} on rank ", dist.get_global_rank())
         download_config = datasets.utils.DownloadConfig(max_retries=self.max_network_retries)
-        self.dataset = datasets.load_dataset("glue", self.task, split=self.split, download_config=download_config)
+        dataset = datasets.load_dataset("glue", self.task, split=self.split, download_config=download_config)
```
ravi-mosaicml, line 88: Pyright complained about setting a self. attribute outside of `__init__`

composer/datasets/glue.py
```diff
@@ -16,6 +16,17 @@
 
 log = logging.getLogger(__name__)
 
+_task_to_keys = {
```
ravi-mosaicml, line 19: Pyright complained about setting a self. attribute outside of __init__

docs/source/doctest_fixtures.py
```diff
@@ -86,9 +86,12 @@
 
 image = Image.fromarray(np.random.randint(0, 256, size=(32, 32, 3), dtype=np.uint8))
 
-X_example = torch.randn(batch_size, num_channels, 32, 32)
-logits = torch.randn(batch_size, num_classes)
-y_example = torch.randint(num_classes, (batch_size,))
+# error: "randn" is not a known member of module (reportGeneralTypeIssues)
```
ravi-mosaicml, line 89: Pyright 1.1.224.post1 complained about these.

dskhudia (approved):

PR #613, opened 2022-02-28
Optim cleanup + docstrings

jbloxham:
- Deleted `composer.optim._time_conversion` and its associated test, since it is unused.
- Moved scheduler hparams into a new file


jbloxham:
Superseded by #653

PR #614, opened 2022-02-28
Fix no_op_model algorithm

dskhudia:
`composer -n 1 examples/run_composer_trainer.py -f composer/yamls/models/resnet9_cifar10.yaml --max_duration 1ep --datadir /localdisk/CIFAR10 --validate_every_n_epochs=0 --algorithms no_op_model`

Above command results in the following error without this fix

```
  File "examples/run_composer_trainer.py", line 60, in <module>
    main()
  File "examples/run_composer_trainer.py", line 56, in main
    trainer.fit()
  File "/mnt/cota/daya/composer/composer/trainer/trainer.py", line 774, in fit
    self._train_loop()
  File "/mnt/cota/daya/composer/composer/trainer/trainer.py", line 936, in _train_loop
    total_loss = self._train_batch(microbatches)
  File "/mnt/cota/daya/composer/composer/trainer/trainer.py", line 1009, in _train_batch
    return self._train_batch_inner(microbatches)
  File "/mnt/cota/daya/composer/composer/trainer/trainer.py", line 1040, in _train_batch_inner
    state.outputs = state.model.forward(state.batch)
  File "/mnt/cota/daya/composer/composer/algorithms/no_op_model/no_op_model.py", line 46, in forward
    return y * self.weights
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```


composer/algorithms/no_op_model/no_op_model.py
```diff
@@ -26,7 +26,7 @@ class NoOpModelClass(ComposerModel):
 
     def __init__(self, original_model: torch.nn.Module):
         super().__init__()
-        self.weights = torch.tensor([1.5], requires_grad=True, dtype=torch.float)
+        self.weights = torch.nn.Parameter(torch.Tensor([1.5]))
```
ravi-mosaicml, line 29: I'm a bit confused on how this would fix a device error...would we need to set the device explicitly, somewhat like this?

```suggestion
        original_device = next(original_model.parameters()).device
        self.weights = torch.nn.Parameter(torch.Tensor([1.5]), device=original_device)
```

dskhudia: no_op_model algorithm runs on the init event (https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L649), i.e., before moving the model parameters to gpu. So `original_device` in your code above is still cpu. Once we make it a parameter, module_to_device (https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L710) takes care of it.

hanlint (approved):
LGTM, thanks for catching this!

composer/algorithms/no_op_model/no_op_model.py
ravi-mosaicml: Ahhh, makes sense. Thanks!

PR #617, opened 2022-02-28
Updated dark mode composer logo, and graph

nqn:
**Before**
Screen Shot 2022-02-28 at 2 22 09 PM

**After**
Screen Shot 2022-02-28 at 2 22 40 PM


moinnadeem (approved):
a much needed upgrade! great work Niklas

Averylamp (approved):
looks awesome!  I think the second gradient behind the graph does make the whole image look very dark, but definitely still an improvement

PR #629, opened 2022-03-01
Remove state shorthand

mvpatel2000:
Based on offline discussion with Ravi.

Using state shorthand, specifically `state = self.state` and then just using `state` everywhere else isn't a great abstraction because it becomes hard for developers / people new with the codebase to understand scoping. This is especially true when you assign something to a field of `state`. IMO shorthands are not worth harder comprehensibility, especially given good autocomplete by most IDEs. 

This PR moves to using `self.state` everywhere, and makes it much easier to understand the scope of assignments.


ravi-mosaicml (approved):
LGTM, thanks

PR #643, opened 2022-03-02
Hide showing `object` as a base class; fix skipping documentation of `forward`; fixed docutils dependency.

### LINKED ISSUE #638: Docs Nit: Hide the base class when the base is `object`
If the base class is `object`, then let's not show that in the docs build. We should be able to adjust a setting or use a hook in `conf.py`.

_Originally posted by @ravi-mosaicml in https://github.com/mosaicml/composer/pull/632#discussion_r817238607_

ravi-mosaicml:
* Fixed how we skip documenting `forward` so it will still show up for overriden subclasses
* Stopped showing base classes when the only base class is `object`. Closes #638
* Added docutils>=0.17 to setup.py. See https://myst-parser.readthedocs.io/en/latest/sphinx/use.html?highlight=parser#include-markdown-files-into-an-rst-file


bandish-shah (approved):

ravi-mosaicml:
The readthedocs.com build is still showing `bases: object` but the local build isn't. @bandish-shah thoughts?
image

bandish-shah:
> The readthedocs.com build is still showing `bases: object` but the local build isn't. @bandish-shah thoughts? image

Interesting, I ran the build locally on this branch and am still seeing `bases: object` show up:
image

ravi-mosaicml:
Fixed it; needed to add a `:py:` in front.

Issue #513, opened 2022-02-17
Clean data access interface for `InMemoryLogger`


growlix:
## 🚀 Feature Request
Currently data is stored in the `InMemoryLogger` in a `dict` in the form

```
{key: [(Timestamp, LogLevel, metric), (Timestamp, LogLevel, metric), ...]
      ...
}
```

This is not a particularly humane format to store the data. Doing much of anything with it requires a user to write their own code to extract and collate the data into a different format. It would be great if the `InMemoryLogger` had a method(s) to return its data in a more easily usable form. Maybe something like 

```
print(in_memory_logger.get_data(data="accuracy/val", interval="epoch"))
>>> {"accuracy/val": [15, 25, 31, 33, ...], "epoch": [0, 1, 2, 3, ...]}
```

ravi-mosaicml:
@growlix assigned to you! Feel free to add helper methods that would make sense.

PR #644, opened 2022-03-02
InMemoryLogger has get_timeseries() method

growlix:
Solution to #513


composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
             self.data[k].append((timestamp, log_level, v))
         self.most_recent_values.update(data)
         self.most_recent_timestamps.update({k: timestamp for k in data})
+
+    def get_timeseries(self, metric: str, time_unit: Union[str, None] = None):
```
hanlint, line 79: Would be good to add a quick test into pytest on this method, e.g.:
```
logger = InMemoryLogger(..)

# patch the data directly
logger.data = { ..}

assert logger.get_timeseries(...) ==
```




```

Comment (ravi-mosaicml) [inline composer/loggers/in_memory_logger.py:79]: For TimeUnit, we generally take the string or enum value:

```suggestion
    def get_timeseries(self, metric: str, time_unit: Optional[Union[str, TimeUnit]] = None):
```

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+        """Returns logged data as dict containing values of a desired metric over time.
+
+        Args:
+            metric (str): Metric of interest. Must be present in self.data.keys().
+            time_unit(str or None, optional) Which unit of time to return. Options =
```
ravi-mosaicml, line 84:
```suggestion
            time_unit (str or None, optional) Which unit of time to return. Options =
```

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+                values of that time unit for the corresponding index of the metric. For
+                example, plotting timeseries['accuracy/val'] vs. timeseries['epoch'] would
+                plot validation accuracy as a function of epoch. And if
+                timeseries['accuracy/val'][5] == 41.2, and timeseries['batch'][5] == 680,
+                that means that 'accuracy/val' was equal to 41.2 on the 680th batch.
+        """
+
+        # Check time_unit string
+        valid_time_strings = {"batch", "epoch", "sample", "token", "duration", None}
+        if time_unit not in valid_time_strings:
```
ravi-mosaicml, line 101: you can try casting via `time_unit = TimeUnit(time_unit)` -- if it's an illegal unit, then a ValueError will be raised. So no need to validate manually,

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+        Args:
+            metric (str): Metric of interest. Must be present in self.data.keys().
+            time_unit(str or None, optional) Which unit of time to return. Options =
+                {"batch", "epoch", "sample", "token", "duration", None}. If ``None``,
+                will return all available units of time. Default=``None``.
+
+        Returns:
+            timeseries (dict): Dictionary in which one key is ``metric``, and the
+                associated value is a list of values of that metric. The remaining key(s)
+                are each a unit of time, and the associated values are each a list of
```
ravi-mosaicml, line 91: From a typing perspective, I'm not a huge fan of mixing the value types of a Dict.The annotation of the returned dictionary would be something like `Dict[Union[str, TimeUnit], Union[List[TLogData], List[Time]]]`.

I'm wondering whether it would make sense to break this into returning a tuple where one element is the data and the other is a dictionary of the timestamps -- e.g. `Tuple[List[TLogData], Dict[TimeUnit, List[Time]]`?

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+                values of that time unit for the corresponding index of the metric. For
+                example, plotting timeseries['accuracy/val'] vs. timeseries['epoch'] would
+                plot validation accuracy as a function of epoch. And if
+                timeseries['accuracy/val'][5] == 41.2, and timeseries['batch'][5] == 680,
+                that means that 'accuracy/val' was equal to 41.2 on the 680th batch.
+        """
+
+        # Check time_unit string
+        valid_time_strings = {"batch", "epoch", "sample", "token", "duration", None}
```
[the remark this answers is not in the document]
ravi-mosaicml: Validation should not need to be done manually -- see comment below.

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+    def get_timeseries(self, metric: str, time_unit: Union[str, None] = None):
+        """Returns logged data as dict containing values of a desired metric over time.
+
+        Args:
+            metric (str): Metric of interest. Must be present in self.data.keys().
+            time_unit(str or None, optional) Which unit of time to return. Options =
+                {"batch", "epoch", "sample", "token", "duration", None}. If ``None``,
+                will return all available units of time. Default=``None``.
+
+        Returns:
```
ravi-mosaicml, line 88: Can you add a doctest? I presume this helper is for matplotlib, in which case it would be really nice to show how the results from here can be passed directly into a pyplot.

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
         self.most_recent_values.update(data)
         self.most_recent_timestamps.update({k: timestamp for k in data})
+
+    def get_timeseries(self, metric: str, time_unit: Union[str, None] = None):
+        """Returns logged data as dict containing values of a desired metric over time.
+
+        Args:
+            metric (str): Metric of interest. Must be present in self.data.keys().
+            time_unit(str or None, optional) Which unit of time to return. Options =
+                {"batch", "epoch", "sample", "token", "duration", None}. If ``None``,
```
ravi-mosaicml, line 85: I don't think it's necessary to enumerate the units here. Also thoughts about removing the unit parameter and instead always returning all units?

composer/loggers/in_memory_logger.py
hanlint: `TimeUnit` allows for `duration`, but `TimeStamp` does not have the `duration` field.

composer/loggers/in_memory_logger.py
hanlint: The return type here should always be `Dict[str, TLogData]`. Even for time units, the code below does:
```
timeseries[time_unit].append(getattr(timestamp, time_unit).value)
```

where `time_unit` is a string, and the latter is compatible with `TLogData`.

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+                will return all available units of time. Default=``None``.
+
+        Returns:
+            timeseries (dict): Dictionary in which one key is ``metric``, and the
+                associated value is a list of values of that metric. The remaining key(s)
+                are each a unit of time, and the associated values are each a list of
+                values of that time unit for the corresponding index of the metric. For
+                example, plotting timeseries['accuracy/val'] vs. timeseries['epoch'] would
```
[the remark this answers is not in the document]
growlix: It does not support returning multiple metrics. Different metrics may have different sampling resolutions and hence different length time series. I think this would require returning nested dicts, which feels like worse UX to me than making multiple calls to `.get_timeseries()`, but others may disagree and also it's possible there's a nice solution I'm not thinking of.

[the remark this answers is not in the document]
hanlint: Agreed!

composer/loggers/in_memory_logger.py
growlix: TimeUnits are a little clunky here because TimeStamp fields don't follow the same convention: they're named `epoch`, `batch`, `batch_in_epoch`, etc.

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +75,50 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+            timeseries[metric].append(metric_value)
+            timestamp = datapoint[0]
+            if time_unit:
+                timeseries[time_unit].append(getattr(timestamp, time_unit).value)
+            else:
+                # Iterate through time units and add them all!
+                for field in timestamp._fields:
+                    time_value = getattr(timestamp, field).value
+                    timeseries[field].append(time_value)
+        return timeseries
```
growlix, line 124: Having type check problems here: `Expression of type "defaultdict[Unknown, list[Unknown]]" cannot be assigned to return type "Dict[str, TLogData]"`. Suggested solutions @ravi-mosaicml, @hanlint?

composer/loggers/in_memory_logger.py
ravi-mosaicml: If it's just one value that's not supported, can we manually check for that and raise an error? Not a fan of manually repeating this here.

composer/loggers/in_memory_logger.py
ravi-mosaicml: Easiest way to fix would be to not use a defaultdict and instead use a dict

growlix:
I removed the `time_unit` argument and return time series for all time units.

Mostly because`TimeStamp` fields can't be accessed using `TimeUnits` because they don't follow the same convention. `TimeStamp` fields are named e.g. `"epoch"` and `"batch"`, and there exist TimeStamp fields that aren't TimeUnits (e.g. `"batch_in_epoch"`). This makes me wonder if we should create a `TimeUnit` -> `TimeStamp` field mapping somewhere for easy conversion.

Anyway, this method is a lot simpler now.

growlix:
Some data in the logger are stored as tensors, such that calling `in_mem_logger.get_timeseries("accuracy/val")` returns `[tensor(0.2287, devic...='cuda:0'), tensor(0.4015, devic...='cuda:0'), ...]`. Should we check for whether the data are tensors, and if so convert them to lists?

hanlint:
I think returning them as numpy arrays would be reasonable.

composer/loggers/in_memory_logger.py
```diff
@@ -74,3 +77,58 @@ def log_metric(self, timestamp: Timestamp, log_level: LogLevel, data: TLogData):
...
+        Example:
+            .. testcode::
+
+                import matplotlib.pyplot as plt
+
+                from composer.core.logging import LogLevel
+                from from composer.loggers import InMemoryLogger
+
+                in_mem_logger = InMemoryLogger(LogLevel.BATCH)
+                in_mem_logger = populate_in_memory_logger(in_mem_logger, datafield="accuracy/val")
```
hanlint, line 106: Might be better to directly populate it in the testcode here with some minimal data, e.g.
```
# for this example, directly populate the data
in_mem_logger.data = {'accuracy/val': [(1, blah, 0.4), (2, blah, 1.0)]}
timeseries = in_mem_logger.get_timeseries("accuracy/val")
timeseries
>>> [0.4, 1.0]
```

growlix: I populate in the example now, but it's pretty verbose (as you can see) because it requires `Timestamp`s. I could go either way!

tests/test_logger.py
```diff
@@ -149,3 +151,28 @@ def test_in_memory_logger(dummy_state: State):
     # the most recent values should have just the last call to epoch
     assert in_memory_logger.most_recent_values["epoch"] == "should_be_recorded_and_override"
     assert in_memory_logger.most_recent_timestamps["epoch"].batch == 1
+
+    # test get_timeseries()
```
hanlint, line 155: this should be a separate test

docs/source/doctest_fixtures.py
```diff
@@ -110,3 +114,27 @@ def Trainer(fake_ellipses='...', *args, **kwargs):
 
 # patch composer so that 'from composer import Trainer' calls do not override change above
 composer.Trainer = Trainer
+
+
+def populate_in_memory_logger(logger: InMemoryLogger,
```
hanlint, line 119: this can be removed

hanlint (approved):
LGTM, two minor comments, and then good to merge.

PR #650, opened 2022-03-03
[Part 1] Adding Synthetic NLP Tokenizers, Models, Datasets w/o Integration

moinnadeem:
This PR adds the synthetic NLP tokenizers, models, and datasets. The next PR [Part 2] will take care of integrating these into our models, and updating the actual tests to pass.


composer/models/bert/bert_hparams.py
```diff
@@ -50,8 +50,10 @@ def initialize_object(self) -> "BERTModel":
             model = transformers.AutoModelForSequenceClassification.from_pretrained(self.pretrained_model_name,
                                                                                     **model_hparams)
         else:
+            for k, v in model_hparams.items():
+                assert getattr(config, k) == v
```
ravi-mosaicml, line 54: is this assert for validation or an invariant? if it's validation, can you make it into a ValueError?

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        json.dump({"pad_token": tokenizer_params.pad_token}, f)
+
+    # instantiate the new tokenizer
+    assert issubclass(tokenizer_params.tokenizer_cls, PreTrainedTokenizer)
+    tokenizer = tokenizer_params.tokenizer_cls.from_pretrained(tmp_tokenizer_dir)
+
+    return tokenizer
+
+
+class SyntheticHFDataset:
```
ravi-mosaicml, line 180: Does this need a base class?

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        self.num_samples = num_samples
+        self.chars_per_sample = chars_per_sample
+        self.column_names = column_names
+
+    def generate_dataset(self):
+        try:
+            import datasets
+        except ImportError as e:
+            raise ImportError(
+                'Composer was installed without NLP support. To use NLP with Composer, run: `pip install mosaicml[nlp]`.'
```
ravi-mosaicml, line 195: The import error should have a corresponding statement for anaconda.

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        decoder=decoders.ByteLevel(),
+        initial_alphabet=pre_tokenizers.ByteLevel.alphabet(),
+        special_tokens=[pad_token, "<endoftext>"],
+        pad_token=pad_token,
+        trainer_cls=tokenizers_trainer.BpeTrainer,
+        tokenizer_cls=GPT2Tokenizer,
+    )
+
+
+def generate_synthetic_tokenizer(tokenizer_family: str, dataset=None, vocab_size=256) -> PreTrainedTokenizer:
```
ravi-mosaicml, line 114: dataset and vocab size are missing type annotations

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        num_samples = 100
+        chars_per_sample = 128
+        column_names = ['sentence']
+        dataset = SyntheticHFDataset(num_samples=num_samples,
+                                     chars_per_sample=chars_per_sample,
+                                     column_names=column_names).generate_dataset()
+
+    # flatten dataset to get a simple stream of text for the tokenizer
+    dataset = [dataset[key] for key in dataset.column_names if key != 'idx']
+    dataset = [i for sublist in dataset for i in sublist]
```
ravi-mosaicml, line 135: can you split this into two statements? double `for in` is hard to follow

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+    decoder: decoders.Decoder
+    initial_alphabet: list
+    special_tokens: list
+    pad_token: str
+    trainer_cls: type
+    tokenizer_cls: type
+
+
+def generate_bert_tokenizer_params(dataset) -> SyntheticTokenizerParams:
+    try:
```
ravi-mosaicml, line 52: docstring or make a private method

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        initial_alphabet=initial_alphabet,
+        special_tokens=[pad_token, unk_token, "[SEP]", "[CLS]", "[MASK]"],
+        pad_token=pad_token,
+        trainer_cls=tokenizers_trainer.WordPieceTrainer,
+        tokenizer_cls=BertTokenizer,
+    )
+
+
+def generate_gpt2_tokenizer_params() -> SyntheticTokenizerParams:
+    try:
```
ravi-mosaicml, line 85: docstring or make a private method

composer/datasets/synthetic.py
ravi-mosaicml: also docstring or make private

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+    else:
+        raise ValueError(f"Synthetic tokenizers for tokenizer family {tokenizer_family} are currently unsupported.")
+
+    tokenizer = Tokenizer(tokenizer_params.tokenizer_model)
+    tokenizer.enable_padding(direction="right",
+                             pad_id=0,
+                             pad_type_id=0,
+                             pad_token=tokenizer_params.pad_token,
+                             pad_to_multiple_of=8)
+    tokenizer.normalizer = tokenizer_params.normalizer  # type: ignore
```
ravi-mosaicml, line 150: why the type ignores?

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+    tokenizer.decoder = tokenizer_params.decoder  # type: ignore
+    tokenizer_trainer = tokenizer_params.trainer_cls(
+        vocab_size=vocab_size,
+        initial_alphabet=tokenizer_params.initial_alphabet,
+        special_tokens=tokenizer_params.special_tokens,
+    )
+    tokenizer.train_from_iterator(dataset, trainer=tokenizer_trainer)
+
+    # save the tokenizer config
+    tmp_tokenizer_dir = mkdtemp()
```
ravi-mosaicml, line 161: use `with tempfile.TemporaryDirectory() as tempdir` instead of `mkdtemp`...tempfile ensures that the directory is removed at the end.

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        from transformers import PreTrainedTokenizer
+    except ImportError as e:
+        raise ImportError(
+            textwrap.dedent("""\
+            Composer was installed without NLP support. To use NLP with Composer, run `pip install mosaicml[nlp]`
+            if using pip or `conda install -c conda-forge transformers tokenizers` if using Anaconda.""")) from e
+
+    # generate a synthetic dataset with reasonable defaults is none is provided
+    if dataset is None:
+        num_samples = 100
```
ravi-mosaicml, line 126: should these be parameters?

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+        dataset = SyntheticHFDataset(num_samples=num_samples,
+                                     chars_per_sample=chars_per_sample,
+                                     column_names=column_names).generate_dataset()
+
+    # flatten dataset to get a simple stream of text for the tokenizer
+    dataset = [dataset[key] for key in dataset.column_names if key != 'idx']
+    dataset = [i for sublist in dataset for i in sublist]
+
+    if "bert" in tokenizer_family:
+        tokenizer_params = generate_bert_tokenizer_params(dataset)
```
ravi-mosaicml, line 138: instead of taking the `tokenizer_family`, could you take the `tokenizer_params`? That way a user could write their own tokenizer params generation function and still use this synthetic dataset. Or, alternatively, they could use the bert / gpt2 functions that you provide. In this case, then `dataset` will need to be a required argument and constructed beforehand.

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+    # save the vocabulary and potential merges file
+    tokenizer_params.tokenizer_model.save(tmp_tokenizer_dir)  # type: ignore
+
+    # the .from_pretrained method doesn't load our padding for some reason, so we save it as a special kwarg
+    tmp_tokenizer_config = join(tmp_tokenizer_dir, "tokenizer_config.json")
+    with open(tmp_tokenizer_config, "w") as f:
+        json.dump({"pad_token": tokenizer_params.pad_token}, f)
+
+    # instantiate the new tokenizer
+    assert issubclass(tokenizer_params.tokenizer_cls, PreTrainedTokenizer)
```
ravi-mosaicml, line 174: is this validation or an invariant?

ravi-mosaicml (commented):
Overall LGTM; left some code style comments. CC @abhi-mosaic @anisehsani @growlix for a more NLP-focused review.

tests/datasets/test_synthetic_data.py
```diff
@@ -5,7 +5,81 @@
...
 from composer.datasets.synthetic import (SyntheticBatchPairDataset, SyntheticDataLabelType, SyntheticDataType,
-                                         SyntheticPILDataset)
+                                         SyntheticHFDataset, SyntheticPILDataset, generate_synthetic_tokenizer)
+
+
+@pytest.mark.parametrize('num_samples', [50])
+@pytest.mark.parametrize('chars_per_sample', [128])
+@pytest.mark.parametrize('column_names', [['sentence'], ['sentence1', 'sentence2']])
+@pytest.mark.parametrize('tokenizer_family', ['bert', 'gpt2'])
+def test_synthetic_hf_dataset_creation(num_samples: int, chars_per_sample: int, column_names: list,
```
hanlint, line 15: Recommend breaking up all the asserts here to individual tests, using fixtures for setup, e.g.
```
@pytest.mark.parametrize('num_samples', [50])
@pytest.mark.parametrize('chars_per_sample', [128])
@pytest.mark.parametrize('column_names', [['sentence'], ['sentence1', 'sentence2']])
@pytest.mark.parametrize('tokenizer_family', ['bert', 'gpt2'])
class TestSyntheticNLP:

   @pytest.fixture
    def dataset_generator(...)
         return SyntheticHFDatset(..)
    
    @pytest.fixture
    def dataset(dataset_generator):
        return dataset_generator.generate_dataset()

    def test_generator_sample(dataset_generator, chars_per_sample):
        sample = dataset_generator.generate_sample()
        assert len(sample) == chars_per_sample

    def test_dataset_shape(dataset, ...):
        assert len(dataset) ==
        assert len(datsaet[..) == 
```

And so forth for the individual tokenizer tests, etc. Essentially everywhere you have a comment below should be converted into an isolated test.



```

**Comment by moinnadeem** [comment]
Thanks for the comments everyone! I'll resolve these shortly

Adding @growlix for an NLP-focused review!

File: tests/datasets/test_synthetic_data.py [hunk 14]
```diff
@@ -5,7 +5,81 @@
...
+    print(tokenizer.convert_ids_to_tokens(x))
+    print(x[-1], tokenizer.pad_token, tokenizer.pad_token_id, tokenizer.convert_ids_to_tokens(x[-1]))
+    assert x[-1] == tokenizer.pad_token_id
+
+
+@pytest.mark.parametrize('tokenizer_family', ["bert", "gpt2"])
+@pytest.mark.parametrize('vocab_size', [512])
+def test_synthetic_tokenizer_creation(tokenizer_family, vocab_size):
+    try:
+        from transformers import BertTokenizer, GPT2Tokenizer
```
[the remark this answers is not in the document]
moinnadeem: Done!

tests/datasets/test_synthetic_data.py
moinnadeem: Yeah, I agree -- thanks for pointing this out, I didn't know about this!

composer/models/bert/bert_hparams.py
moinnadeem: It's an invariant! The purpose is to ensure that internally casting from the model hparams to an HF config doesn't lose keys. I can add a comment describing why

composer/datasets/synthetic.py
moinnadeem: Resolved!

composer/datasets/synthetic.py
moinnadeem: Done on both!

composer/datasets/synthetic.py
moinnadeem: Done!

composer/datasets/synthetic.py
moinnadeem: Done!

composer/datasets/synthetic.py
moinnadeem: Done!

composer/datasets/synthetic.py
moinnadeem: Done, this is now possible after the refactor! Earlier on, I didn't want the temporary directory being immediately deleted.

composer/datasets/synthetic.py
moinnadeem: In my opinion, if the user wants to change these parameters, they should pass in a dataset themselves. We could make these into parameters, and pass that into the dataset (to the same effect), but I thought this would enable a short function signature with the same flexibility. In other words, those are just defaults

composer/datasets/synthetic.py
moinnadeem: Validation! I just changed the assert into a ValueError!

moinnadeem: ...wow, i sound way too excited about that

composer/datasets/synthetic.py
```diff
@@ -10,6 +18,13 @@
 from composer.core.types import MemoryFormat
 from composer.utils.string_enum import StringEnum
 
+if TYPE_CHECKING:
+    import tokenizers.decoders as decoders
```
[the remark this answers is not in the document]
moinnadeem: > is it possible to do from tokenizers import normalizers, pre_tokenizers, ... etc? or just do import tokenizers

I slacked you about this! I originally did `import tokenizers`, but PyRight was complaining.

I can refactor this into multiple synthetic files if we think the file is getting too long!

composer/datasets/synthetic.py
moinnadeem: I was getting errors like: _ error: Cannot assign member "normalizer" for type "Tokenizer"
    Property "normalizer" has no defined setter (reportGeneralTypeIssues)_

This seemed unrelated to our code, but happy to discuss!

composer/datasets/synthetic.py
```diff
@@ -21,6 +36,201 @@ class SyntheticDataLabelType(StringEnum):
...
+            if using pip or `conda install -c conda-forge transformers tokenizers` if using Anaconda.""")) from e
+
+    unk_token = "[UNK]"
+    pad_token = "[PAD]"
+
+    initial_alphabet = "".join([i for i in dataset])
+    initial_alphabet = list(set(initial_alphabet))
+
+    return SyntheticTokenizerParams(
+        tokenizer_model=tokenizers_models.WordPiece(unk_token=unk_token),  # type: ignore
```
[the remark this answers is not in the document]
moinnadeem: Without it, I get the following error about default values that haven't been filled in:  `error: Arguments missing for parameters "vocab", "max_input_chars_per_word" (reportGeneralTypeIssues)` from pyright

composer/datasets/synthetic.py
moinnadeem: Nope, the base class is `object`

moinnadeem:
Cool, I think this is ready for another review!

@ravi-mosaicml , I addressed all of your feedback! Let me know if I missed anything. I left the conversations unresolved so you can take a look at my comments for each item.

@hanlint I moved things into fixtures, and created `synthetic_lm` and `test_synthetic_lm_data` files in order to declutter stuff. Good call on both of those!

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+    initial_alphabet: list
+    special_tokens: list
+    pad_token: str
+    trainer_cls: type
+    tokenizer_cls: type
+
+
+def _generate_bert_tokenizer_params(dataset) -> SyntheticTokenizerParams:
+    try:
+        import tokenizers.decoders as decoders
```
hanlint, line 36: this is pretty verbose and repeated multiple times in this file. i think we need to come up with a general pattern for files that need to be imported, but have these optional dependencies. cc: @ravi-mosaicml or @dblalock for suggestions?

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+
+    tokenizer = Tokenizer(tokenizer_params.tokenizer_model)
+    tokenizer.enable_padding(direction="right",
+                             pad_id=0,
+                             pad_type_id=0,
+                             pad_token=tokenizer_params.pad_token,
+                             pad_to_multiple_of=8)
+    tokenizer.normalizer = tokenizer_params.normalizer  # type: ignore
+    tokenizer.pre_tokenizer = tokenizer_params.pre_tokenizer  # type: ignore
+    tokenizer.decoder = tokenizer_params.decoder  # type: ignore
```
hanlint, line 150: type ignore?

hanlint: Is this because your are assigning attributes to tokenizer that are not in its public API?

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+    # the .from_pretrained method doesn't load our padding for some reason, so we save it as a special kwarg
+    tmp_tokenizer_config = join(tmp_tokenizer_dir, "tokenizer_config.json")
+    with open(tmp_tokenizer_config, "w") as f:
+        json.dump({"pad_token": tokenizer_params.pad_token}, f)
+
+    # instantiate the new tokenizer
+    if not issubclass(tokenizer_params.tokenizer_cls, PreTrainedTokenizer):
+        raise ValueError(f"{tokenizer_params.tokenizer_cls} should sub-class transformers.PreTrainedTokenizer.")
+    tokenizer = tokenizer_params.tokenizer_cls.from_pretrained(tmp_tokenizer_dir)
+
```
hanlint, line 175: should the tempdir be cleaned up here? if so, perhaps instead should use the `TemporaryFolder` context instead of the low level `mkdtemp`

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+        self.num_samples = num_samples
+        self.chars_per_sample = chars_per_sample
+        self.column_names = column_names
+
+    def generate_dataset(self):
+        try:
+            import datasets
+        except ImportError as e:
+            raise ImportError(
+                textwrap.dedent("""\
```
hanlint, line 194: For DRY, we probably want to store the message in a constant up front, e.g.
```
_IMPORT_MESSAGE = textwrap.dedent(...)
```
and then here just do:
```
raise ImportError(_IMPORT_MESSAGE) from e
```
```

**Review by hanlint — COMMENTED** [review]
LGTM, with a few comments, with one major comment on establishing a good pattern for optional imports that we should raise for a braoder discussion.

[nit] Not critical, but could benefit from a bit more tweaking on the pytest structure.

File: composer/datasets/synthetic_lm.py (hunk 17 above)
Comment (moinnadeem) [reply composer/datasets/synthetic_lm.py:36]: Strong agree here, I'm pretty interested in fixing this!

File: composer/datasets/synthetic_lm.py (hunk 18 above)
Comment (moinnadeem) [reply composer/datasets/synthetic_lm.py:150]: Correct! See this: https://github.com/mosaicml/composer/pull/650#discussion_r821361537

Is there a better way to do this? Happy to implement it if so

File: composer/datasets/synthetic_lm.py (hunk 19 above)
Comment (moinnadeem) [reply composer/datasets/synthetic_lm.py:175]: The temporary directory should not be cleaned up, because the tokenizer folder that is created needs to exist for the model to access the tokenizer as well.

This might imply a potential race condition if the temporary directory is cleaned up by the OS before the model gets instantiated? Is that a concern?

File: composer/datasets/synthetic_lm.py (hunk 20 above)
Comment (moinnadeem) [reply composer/datasets/synthetic_lm.py:194]: Agreed, happy to implement this.

**Comment by moinnadeem** [comment]
> [nit] Not critical, but could benefit from a bit more tweaking on the pytest structure.

Can you provide any suggestions? I might be interested in implementing it, but at the very least, I'd be interested in learning what I could have done better.

File: composer/datasets/synthetic_lm.py (hunk 17 above)
Comment (dblalock) [reply composer/datasets/synthetic_lm.py:36]: I don't know what the best way of dealing with this is. I could imagine a few solutions:

1.  We have a `composer.tokenizers` that lazily tries to import hf tokenizers in its getattr.
2.  We have a singleton imported as `composer.tokenizers` that does the same thing.
3.  We have a `composer.tokenizers` that tries to import hf tokenizers, but instead creates a Mock if that fails. The Mock throws the informative ImportError below whenever it's called.

3 seems the least bad, but none of these seem ideal.

File: tests/datasets/test_synthetic_lm_data.py [hunk 21]
```diff
@@ -0,0 +1,117 @@
+from itertools import product
```
growlix, line 1: This isn't standard practice, but I wonder if we should start annotating our tests a bit more. Their intent can often be inferred from the code, but it might make things easier for everyone if their intent was clearly stated by the author.

growlix (approved):
Minor comments, mostly regarding documentation. LGTM.

tests/datasets/test_synthetic_lm_data.py
```diff
@@ -0,0 +1,117 @@
...
+
```
growlix, line 2: Does it make sense to test whether the `[sep]`, `[mask]`, and `[cls]` tokens are present as expected? More generally, are there any assumptions that algorithms make about the structure of the input data, and do those need to be checked here?

growlix (approved):
Added a comment about tests

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+        decoder=decoders.ByteLevel(),
+        initial_alphabet=pre_tokenizers.ByteLevel.alphabet(),
+        special_tokens=[pad_token, "<endoftext>"],
+        pad_token=pad_token,
+        trainer_cls=tokenizers_trainer.BpeTrainer,
+        tokenizer_cls=GPT2Tokenizer,
+    )
+
+
+def generate_synthetic_tokenizer(tokenizer_family: str,
```
ravi-mosaicml, line 97: Add in the tokenizer directory as an argument here, instead of creating the tempdir in the function below

composer/models/bert/bert_hparams.py
ravi-mosaicml: Discussed offline, but thoughts about hardcoding the model_hparams in line 55 and 57, since the dict contains one item with a constant key?

tests/datasets/test_synthetic_lm_data.py
```diff
@@ -0,0 +1,117 @@
...
+import pytest
+
+from composer.datasets.synthetic_lm import SyntheticHFDataset, generate_synthetic_tokenizer
+
+
+def generate_parameter_configs(keys, num_replicas=1):
```
ravi-mosaicml, line 8: Add in a comment on what is happening here...perhaps with an example output and how this is then used by the `dataset_generator` fixture below.

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+        data = {}
+        for column_name in self.column_names:
+            data[column_name] = [self.generate_sample() for _ in range(self.num_samples)]
+        data['idx'] = list(range(self.num_samples))
+
+        hf_synthetic_dataset = datasets.Dataset.from_dict(data)
+        return hf_synthetic_dataset
+
+    def generate_sample(self):
+        MIN_WORD_LENGTH = 3
```
ravi-mosaicml, line 207: can you make these init args with default values?

composer/datasets/synthetic_lm.py
ravi-mosaicml: Use #795

composer/datasets/synthetic_lm.py
ravi-mosaicml: Discussed offline; the folder will be passed in as an argument into the function.

composer/datasets/synthetic_lm.py
ravi-mosaicml: It seems like that the underlying `.pyi` file has incorrect annotations. Let's paste in the error message here.

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+    if not issubclass(tokenizer_params.tokenizer_cls, PreTrainedTokenizer):
+        raise ValueError(f"{tokenizer_params.tokenizer_cls} should sub-class transformers.PreTrainedTokenizer.")
+    tokenizer = tokenizer_params.tokenizer_cls.from_pretrained(tmp_tokenizer_dir)
+
+    return tokenizer
+
+
+class SyntheticHFDataset:
+    """Creates a synthetic HF dataset and passes it to the preprocessing scripts."""
+
```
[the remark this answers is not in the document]
ravi-mosaicml: Discussed offline to refactor this into a function (since it's really a builder, not a dataset), and then make `generate_sample` a private, module-level function.

composer/datasets/synthetic_lm.py
moinnadeem: Done!

composer/datasets/synthetic_lm.py
moinnadeem: Discussed offline: out of scope for this PR

composer/datasets/synthetic_lm.py
moinnadeem: Done!

composer/datasets/synthetic_lm.py
moinnadeem: Done!

composer/datasets/synthetic_lm.py
moinnadeem: Done!

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
+
```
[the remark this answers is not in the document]
moinnadeem: Done! Good catch!

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+from typing import TYPE_CHECKING, NamedTuple, Optional
+
+if TYPE_CHECKING:
+    import tokenizers.decoders as decoders
+    import tokenizers.models as tokenizers_models
+    import tokenizers.normalizers as normalizers
+    import tokenizers.pre_tokenizers as pre_tokenizers
+    from datasets import Dataset
+    from transformers import PreTrainedTokenizer
+
```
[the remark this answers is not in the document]
moinnadeem: Not down to be president, but done regardless!

composer/datasets/synthetic_lm.py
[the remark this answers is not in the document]
moinnadeem: Done!

composer/datasets/synthetic_lm.py
```diff
@@ -0,0 +1,231 @@
...
+    # flatten the list of lists into a single list
+    flattened_dataset = []
+    for sublist in flattened_columns:
+        for item in sublist:
+            flattened_dataset.append(item)
+
+    if "bert" in tokenizer_family:
+        tokenizer_params = _generate_bert_tokenizer_params(flattened_dataset)
+    elif "gpt2" in tokenizer_family:
+        tokenizer_params = _generate_gpt2_tokenizer_params()
```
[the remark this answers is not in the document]
moinnadeem: Ah, good question! This is a difference in WordPiece vs BPE: WordPiece starts with an initial vocabulary (which is drawn from the dataset), and gradually merges characters together into tokens. BPE starts with the text, and gradually breaks text down into tokens. So one requires the initial vocabulary, the other does not. Does this make sense?

composer/datasets/synthetic_lm.py
moinnadeem: Done!

tests/datasets/test_synthetic_data.py
moinnadeem: Resolving this since it is done

tests/datasets/test_synthetic_lm_data.py
moinnadeem: Done!

tests/datasets/test_synthetic_lm_data.py
moinnadeem: Yes! That is what `test_tokenizer_specific_properties` does!

moinnadeem:
@ravi-mosaicml @hanlint Cool, addressed all comments!

@hanlint btw, I resolved the `mkdtemp` thing by using PyTest fixtures to pass in a temporary directory, which can exist for the model tests as well.

hanlint (approved):
LGTM, will merge after tests pass.

PR #651, opened 2022-03-03
Fix bug in trainer test

hanlint:
Fixed a bug using mutable fixtures with how the trainer test generates the reference model. Without this fix, the reference model and any test model actually point to the same object, meaning that all tests would succeed. I also added a few safeguards (negative control test, and asserts in the comparision) to prevent future regressions.


tests/trainer/test_trainer.py
```diff
@@ -110,17 +110,21 @@ class TestTrainerEquivalence():
     reference_folder: pathlib.Path
 
     def assert_models_equal(self, model_1, model_2):
+        assert id(model_1) != id(model_2), "Same model should not be compared."
         for param1, param2 in zip(model_1.parameters(), model_2.parameters()):
             torch.testing.assert_allclose(param1, param2)
 
-    @pytest.fixture
+    @pytest.fixture(scope="function")
```
hanlint, line 117: this should not be neccesary, i can revert and test

ravi-mosaicml: Yes, double-checked the docs and it seems that function scope should be the default.

tests/trainer/test_trainer.py
```diff
@@ -110,17 +110,21 @@ class TestTrainerEquivalence():
     reference_folder: pathlib.Path
 
     def assert_models_equal(self, model_1, model_2):
+        assert id(model_1) != id(model_2), "Same model should not be compared."
```
ravi-mosaicml, line 113: `is` is more standard for identity checks...

```suggestion
        assert model_1 is not model_2, "Same model should not be compared."
```

tests/trainer/test_trainer.py
```diff
@@ -110,17 +110,21 @@ class TestTrainerEquivalence():
...
+        """ Returns the reference config. """
+
+        seed_all(seed=0)
         return {
             'model': SimpleModel(),
             'train_dataloader': DataLoader(
                 dataset=RandomClassificationDataset(),
                 batch_size=4,
-                shuffle=True,
+                shuffle=False,
```
ravi-mosaicml, line 127: not that it makes a big difference, but if we're seeding, then shuffle should still be able to be True?

tests/trainer/test_trainer.py
```diff
@@ -137,6 +141,8 @@ def config(self, device, precision, world_size):
     @pytest.fixture(autouse=True)
     def create_reference_model(self, config, tmpdir_factory, *args):
         """Trains the reference model, and saves checkpoints."""
+        config = deepcopy(config)
```
ravi-mosaicml, line 144: If the config is test-specific (function scoped), can you elaborate why it needs to be deep-copied? I must be missing something here...

tests/trainer/test_trainer.py
hanlint: Actually, having some trouble getting this to pass with `True`. More specifically, if I do:
```
seed_all(0)
dl1 = DataLoader(RandomClassificationDataset(), shuffle=True)
seed_all(0)
dl2 = DataLoader(RandomClassificationDataset(), shuffle=True)

torch.testing.assert_allclose(next(dl1.__iter__()), next(dl2.__iter())
```
fails for `shuffle=True`, but passes for `shuffle=False`. Is there a worker-specific rng that needs to be set?

tests/trainer/test_trainer.py
hanlint: For each test, `config` gets run, and then passed to both this fixture and the test itself. If `config` is not deepcopied, then the reference model and the test point to the same model object, and therefore the test always passes.

tests/trainer/test_trainer.py
ravi-mosaicml: Ahhh...the dataloader is seeded on the first spin, not when it is initialized.

tests/trainer/test_trainer.py
ravi-mosaicml: Ah, I see. I missed this was in a fixture. Maybe could you add this in a comment?

hanlint, line 144:
```suggestion
        config = deepcopy(config)  # ensure the reference model is not passed to tests
```

tests/trainer/test_trainer.py
```diff
@@ -110,17 +111,32 @@ class TestTrainerEquivalence():
...
 
     def assert_models_equal(self, model_1, model_2):
+        assert model_1 is not model_2, "Same model should not be compared."
         for param1, param2 in zip(model_1.parameters(), model_2.parameters()):
-            torch.testing.assert_allclose(param1, param2)
+            torch.testing.assert_allclose(param1, param2, **self.threshold)
+
+    @pytest.fixture(autouse=True)
+    def threshold(self, device, precision, world_size):
+        self.threshold = {'atol': 0, 'rtol': 0}
```
ravi-mosaicml, line 120: if this works, this is awesome!

ravi-mosaicml (approved):
LGTM

tests/trainer/test_trainer.py
hanlint: Yeah, I've identified that it doesn't work in some situations (e.g. mixed precision, and also curiously gradient accumulation on CPU is off by rtol=1e-8), but doing some testing to set strict thresholds where possible.

PR #653, opened 2022-03-03
`composer.optim` docstrings

### LINKED ISSUE #398: Docstrings cleanup! `composer.optim`
The `composer.optim` needs docstrings. As you work through adding docstrings, please:

## Things to pay attention to

## Previewing docs locally

```bash
cd docs
```

jbloxham:
- Moved scheduler hparams into a separate file
- Various scheduler param renames for internal consistency
- Lots and lots of docstrings

Closes #398


ravirahman:
There are also some narrative / tutorial docs that will have to be cleaned up -- see https://github.com/mosaicml/composer/issues/398#issuecomment-1056975826

ravi-mosaicml:
Overall looks great -- love the detail on the schedulers page, especially with the math. I didn't check that page closely but it formats very nicely.

Some comments:
- [ ] Module level docstrings are missing
- [ ] decoupled_weight_decay is missing `__all__`
- [ ] some single ticks should be made double ticks ..e.g in dedcoupled weight decay
- [ ] optimizer_hparams needs `__all__`
- [ ] DecoupledAdamW is not autolinking in optimizer_hparams
- [ ] DecoupledSGDWHparams has nesterov in the signature but amsgrad in the docstring
- [ ] ModelParameters is not auto-linking
- [ ] RMSPropHparams has centered in the signature but centeredx in the docstring
- [ ] SGDHparams has nesterov in the signature but amsgrad in the docstring
- [ ] Can you remove the `get_optimizer` helper method? One should just call hparams.initialize_object(params) -- there's no need for a helper method for it
- [ ] scheduler_hparams is missing `__all__`. In addition, many references are not auto-linking
- [ ] For the scheduler hparams, can you un-document the `scheduler_cls` attribute? Perhaps make that a private variable.
- [ ] Can you simplify the summary line of the module level docstring for schedulers to be something like "Learning Rate Schedulers"?
- [ ] The module level docstring for the schedulers is a bit dense. I love the detail and how it explains how our schedulers are better than the pytorch ones, but it's a bit much for a docstring. Could you move some of the content there into https://mosaicml-composer--653.com.readthedocs.build/en/653/trainer/schedulers.html, and linkt to this page from the docstring?
- [ ] We implemented scale schedule ratio as a multiplication, not division. The math equation shows it's being divided by sigma, not multiplied.
- [ ] Can you provide examples of schedulers in both functional and class form? I'm guessing users may be confused by "an ordinary function can implement this interface as long as it matches the signature of this interface’s [__call__()](https://mosaicml-composer--653.com.readthedocs.build/en/653/api_reference/composer.optim.scheduler.html#composer.optim.scheduler.ComposerScheduler.__call__) method."
- [ ] Why do we pick different defaults for the Linear scheduler? Should we make it the same as the Pytorch scheduler?
- [ ] Can you use ".. seealso::` notation for the "analogous to" lines?
- [ ] "list of str or Time" is confusing -- maybe say "List[str | Time]" (I thought it was a List[str] | Time) when I first read it.
- [ ] for `compile_composer_scheduler`, can you update the return type to be a `LambdaLR`? I don't think we need to explain that we are ignoring the "internal state" -- maybe say something like "Compile a ComposerScheduler into a PyTorch Scheduler" and mention how it's now bound to the trainer state and ssr?

hanlint:
The narrative is cleaned up in #661

jbloxham:
> We implemented scale schedule ratio as a multiplication, not division. The math equation shows it's being divided by sigma, not multiplied.

@ravi-mosaicml can you clarify? Would you prefer something like `alpha_sigma(sigma * t) = alpha(t)`? I decided to avoid that because it's a little more useful to provide a formula for `alpha_sigma(t)` directly.

jbloxham:
> for compile_composer_scheduler, can you update the return type to be a LambdaLR? I don't think we need to explain that we are ignoring the "internal state" -- maybe say something like "Compile a ComposerScheduler into a PyTorch Scheduler" and mention how it's now bound to the trainer state and ssr?

@ravi-mosaicml I agree with your comments for the docstring, and I'll take another pass at it.

However, I disagree with changing the return type from `PyTorchScheduler` to `LambdaLR`. The fact it returns a `LambdaLR` and not some other kind of `PyTorchScheduler` is just an implementation detail. It should be irrelevant to users of this function what kind of `PyTorchScheduler` is being returned, so this function should make no guarantee. I also don't want users to have to wonder what a `LambdaLR` is.

jbloxham:
> Why do we pick different defaults for the Linear scheduler? Should we make it the same as the Pytorch scheduler?

I lean no, because the PyTorch defaults are just dumb, and I can't imagine anyone using them. We're far enough away from PyTorch land at this point, especially after the parameter rename, that I think we have the leeway to do our own thing.

ravi-mosaicml:
> > We implemented scale schedule ratio as a multiplication, not division. The math equation shows it's being divided by sigma, not multiplied.
> 
> @ravi-mosaicml can you clarify? Would you prefer something like `alpha_sigma(sigma * t) = alpha(t)`? I decided to avoid that because it's a little more useful to provide a formula for `alpha_sigma(t)` directly.

Ohhh, I missed that detail. Yeah, that makes sense then. Leave it as is.


> > for compile_composer_scheduler, can you update the return type to be a LambdaLR? I don't think we need to explain that we are ignoring the "internal state" -- maybe say something like "Compile a ComposerScheduler into a PyTorch Scheduler" and mention how it's now bound to the trainer state and ssr?
> 
> @ravi-mosaicml I agree with your comments for the docstring, and I'll take another pass at it.
> 
> However, I disagree with changing the return type from `PyTorchScheduler` to `LambdaLR`. The fact it returns a `LambdaLR` and not some other kind of `PyTorchScheduler` is just an implementation detail. It should be irrelevant to users of this function what kind of `PyTorchScheduler` is being returned, so this function should make no guarantee. I also don't want users to have to wonder what a `LambdaLR` is.

Makes sense, let's leave the annotation as a `PyTorchScheduler` then. Can you see if we can get it to autolink?



> > Why do we pick different defaults for the Linear scheduler? Should we make it the same as the Pytorch scheduler?
> 
> I lean no, because the PyTorch defaults are just dumb, and I can't imagine anyone using them. We're far enough away from PyTorch land at this point, especially after the parameter rename, that I think we have the leeway to do our own thing.

:bomb: :bomb: :bomb:

jbloxham:
This should be ready for another review now.

ravi-mosaicml:
Handing off to @bandish-shah to do the final review

hanlint (commented):
One comment: the docstring code itself is hard to read because the references are very long, e.g. 
```
class:`~composer.optim.decoupled_weight_decay.DecoupledAdamW`
```

Can we convert all the long reference to shorter? e.g. :class:`.DecoupledAdamW`.

composer/optim/__init__.py
```diff
@@ -1,5 +1,17 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
+"""Optimizers and schedulers.
```
bandish-shah, line 3: [nit] While this is a Google compatible style docstring... bc we put this in our docs page it just appears as a sentence fragment.  Maybe we could just delete, reword the next sentence a bit and move up?  '

Would look like:
```
"""Composer is compatible with optimizers based off of PyTorch's native :class:`~torch.optim.Optimizer` API.

Common optimizers...
"""
```

This would still be Google style compliant and remove the fragment.

bandish-shah (approved):
Overall LGTM, few nits inline and here:

- [x] :class:comopser.optim.decoupled_weight_decay.DecoupledSGDW: convert literal `torch.optim.SGD` to cross-reference
- [x] :class:composer.optim.decoupled_weight_decay.DecoupledAdamW: convert literal `torch.optim.AdamW` to cross-reference

composer/optim/__init__.py
jbloxham: Hmm, that's not entirely true. The table for the API reference page shows the docstring summary, so we do need a short phrase.

Screen Shot 2022-03-07 at 18 35 41

bandish-shah: Good point about the docstring summary table.  I would still prefer a complete complete yet concise sentence here.  It would need to serve as a good standalone description for the summary table as well as a good introductory sentence for the doc.

Now that you have me looking at the table I’m actually thinking it could use some cleanup as well.  Some of the descriptions are probably too long (profiler, core?) where as some are not really descriptive (loggers, utils).  Can clean this up in another PR though.

hanlint: Yeah I saw this too and put in a PR to fix: https://github.com/mosaicml/composer/pull/688

PR #654, opened 2022-03-03
Batchwise resolution for SWA

growlix:
This updates SWA to allow batch- and epoch-wise resolution for updating the averaged model. This is exposed to the user via the `update_interval`, a time string denoting how frequently the averaged model is updated (e.g. `'1ep'`, `'5ba'`). The hparam formerly known as `anneal_epochs` has been renamed to `anneal_steps` to reflect the fact that the SWA-specific LR scheduler now operates at the resolution of `update_interval` instead of single epochs.

Excited to get some NLP results, @moinnadeem!

### DIFF
--- a/composer/algorithms/hparams.py
+++ b/composer/algorithms/hparams.py
@@ -370,7 +370,7 @@ class SWAHparams(AlgorithmHparams):
         default="0.97dur")
     update_interval: str = hp.optional(doc='Time string denoting how often the averaged model is updated. For example, '
                                        '"1ep" means the averaged model will be updated once per epoch, and '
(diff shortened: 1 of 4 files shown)


florescl (approved):
awesome! let's see how the wandb runs with lr monitor look like and then we merge!

growlix:
> Have we been able to run this with a Transformer model? Not to convergence, but does it just initialize?

Haven't tried yet! Do you have a config for 1-gpu interactive debugging of LMs?

moinnadeem:
@growlix I think you should just be able to take the 8-GPU copies and set `new_grad_accum = grad_accum * 8`! That should be the only difference required

PR #655, opened 2022-03-03
Let C4 Dataset overwrite `num_workers` if set incorrectly

abhi-mosaic:
This PR is a quality-of-life improvement for training with the C4 dataset, instead of erroring out and killing the training job, it just overwrites `dataloader.num_workers` before initializing the dataloader.

I expect to relax the `num_workers=1` constraint in the future, but for now it just makes sharding simpler.


siriuslee (approved):
LGTM

PR #656, opened 2022-03-03
Lazy imports for `pycocotools`

abhi-mosaic:
1. This PR adds try..catch for the `pycocotools` imports, so that it's possible to use Composer locally without installing all the requirements. Right now with MCLI this was causing some issues.
2. Remove the hack in the readthedocs build
3. Fix the pycocotools in the meta.yaml to not install by default.
5. Remove wget (this library hasn't been updated since 2015) as a dependency; replace with requests

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -15,6 +15,4 @@ python:
    version: "3.8"
    install:
       - method: pip
-        path: .[base]
-      - method: pip
         path: .[all]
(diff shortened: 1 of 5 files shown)


hanlint (approved):

florescl (approved):

ravi-mosaicml:
I'm adding a few other fixes to this PR.

ravi-mosaicml (approved):
LGTM

PR #657, opened 2022-03-03
Composer.core docstring touch-up

### LINKED ISSUE #403: Docstrings cleanup! `composer.core`
The `composer.core` needs docstrings. As you work through adding docstrings, please:

## Things to pay attention to

## Previewing docs locally

```bash
cd docs
```

ravi-mosaicml:
Fixed a few minor nits with composer.core docstrings.

Closes #403.

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -58,14 +58,8 @@ class State(Serializable):
 
     .. note::
 
-        * An instance of this class is automatically constructed by the :class:`~.Trainer` constructor. A user need
-          not instantiate this class.
-
-        * To support multi-GPU training, ``State.model`` may be wrapped in :class:`torch.nn.parallel.DistributedDataParallel`,
-          and the dataloaders may be wrapped in a device-specific dataloader that handles moving tensors to device.
-
-        * :attr:`~.types.PyTorchScheduler` are wrapped in :attr:`~.ComposerScheduler`, which handles stepping either
-          stepwise or epochwise, and also properly sets up learning rate warmups.
+        An instance of this class is automatically constructed by the :class:`~.Trainer` constructor. A user need
+        not instantiate this class.
 
     Args:
         model (:attr:`~.types.Model`): The model, typically as a subclass of :class:`~.ComposerModel`.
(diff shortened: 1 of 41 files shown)


composer/core/state.py
```diff
@@ -58,14 +58,8 @@ class State(Serializable):
 
     .. note::
 
-        * An instance of this class is automatically constructed by the :class:`~.Trainer` constructor. A user need
-          not instantiate this class.
-
-        * To support multi-GPU training, ``State.model`` may be wrapped in :class:`torch.nn.parallel.DistributedDataParallel`,
```
ravi-mosaicml, line 64: These are somewhat out of date and a bit confusing, since it's the trainer that does this, not the state object.

composer/core/time.py
```diff
@@ -554,13 +566,13 @@ class Timestamp(NamedTuple):
...
-        sample_in_epoch (Time[int]): The number of samples seen in the epoch at a snapshot.
-        token (Time[int]): token at a snapshot.
-        token_in_epoch (Time[int]): The number of tokens seen in the epoch at a snapshot.
+        epoch (Time[int]): The epoch.
+        batch (Time[int]): The batch.
+        batch_in_epoch (Time[int]): The number of batch seen in the last epoch.
+        sample (Time[int]): The sample count.
+        sample_in_epoch (Time[int]): The number of samples seen in the last epoch.
+        token (Time[int]): The token count.
+        token_in_epoch (Time[int]): The number of tokens seen in the last epoch.
```
jbloxham, line 575: All of the `x_in_epoch` attributes are incorrect, assuming `Timestamp` could be generated in the middle of the epoch. They are supposed to refer to the number of batches/samples/tokens seen so far in the current epoch, at the time of the snapshot.

Also, could we make `epoch`/`batch`/`sample`/`token`'s definitions consistent with each other?

docs/source/trainer/time.rst
```diff
@@ -43,7 +43,7 @@ Tracking Time
 The trainer has a :class:`.Timer` object stored in :attr:`.State.timer` that
 measures progress in all the time formats above. :attr:`.State.timer` can be
 read by algorithms and callbacks to trigger behavior at different times
-during training. Especially useful is allowing you to use whicheer format
+during training. Especially useful is allowing you to use whichever format
 is most relevant for your algorithm -- e.g. once-every-n batches, or during the
 last 20% of training.
```
jbloxham, line 48: This wording is a bit awkward for me - maybe:
```suggestion
during training. This allows algorithms to specify times in whatever format
is most useful -- e.g. an algorithm could activate once every *n* batches, or
during the last 20% of training.
```

jbloxham (approved):

PR #658, opened 2022-03-03
Unify algorithms part 2: mixup, cutmix, label smoothing

dblalock:
Follow-up to #524 to continue addressing #343. This is mostly cleaning up the docs to be consistent across methods + match the code's behavior, fixing the doctests, and modifying the docstrings in accordance with some slight API changes (mostly param renaming). These changes are:

Mixup:

- `n_classes` -> `num_classes` in `mixup_batch`, for consitency with torch + rest of composer rather than sklearn.
- `interpolation_lambda` -> `mixing` in `mixup_batch`
- `x` -> `input` in `mixup_batch` for consistency with torch
- `y` -> `target` in `mixup_batch` for consistency with torch
- Removed the wrapper properties for attributes in the algorithm. These weren't documented and were seemingly (?) only there so that the tests didn't access private attrs

Cutmix:

- `n_classes` -> `num_classes` again
- `X` -> `input` in `cutmix_batch` for consistency with torch
- `y` -> `target` in `cutmix_batch` for consistency with torch
- `cutmix_lambda` -> `length` in `cutmix_batch`, matching `cutout`'s `length` argument. As part of this change, I also added changed the logic to 1) correctly compute the internal `cutmix_lambda` when `bbox` is provided, and 2) throw if both length and bbox are provided.
- `cutmix_batch` now returns the permutation it used for consistency with `mixup_batch`. @coryMosaicML , do we need to return these? We apparently didn't here, but seemingly did in `mixup_batch`?
- Removed the wrapper properties for attributes for same reasons as in mixup

Label smoothing:
- `alpha` -> `interpolation` in `smooth_labels` and algorithm
- `targets` -> `target` in `smooth_labels` for consistency with torch

Parts I'm not sure about (updated after slack discussion):

- Whether to return the permutation in functional forms (see above)
- How best to handle cutmix allowing 3 different ways to specify the box to use (`alpha`, `length` / `cutmix_lambda`, and `bbox`). Right now `alpha` is just ignored if others are provided, while the other two now raise a `ValueError` if both are not `None`. Seems like the least bad way to do this, but a little inconsistent. Maybe we could remove ``bbox`` as a parameter?
- For removed wrapper properties, we'd ideally just [not have the tests depend on class internals](https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice). I went with having tests access the private attrs rather than cluttering the code + public API, but we could also just make these attrs public and document them. Probably should have put this change in a separate PR, but I was already renaming the attrs and I was on a roll / changing a lot anyway, so I just went with it.

NOTE: need to fix jenkins error + double-check docstring output in the docs, but this is ready for feedback, esp regarding the API questions above.


hanlint:
I like where this is going! Agreed on `interpolation`... how strictly do we want to maintain cross-functional consistency versus what the original paper/authors terminology. e.g. `alpha` for label smoothing?

coryMosaicML:
> cutmix_batch now returns the permutation it used for consistency with mixup_batch. @coryMosaicML , do we need to return these? We apparently didn't here, but seemingly did in mixup_batch?

I think this behavior in mixup_batch is a holdover from the old repo that got ported over. I don't think the original reason for doing that exists in composer, and not returning the permutation should be fine. It makes the functional interface feel a lot cleaner to me. However, two possible reasons to return permutations: 1) If someone wants to do further analysis/debugging of what mixup_batch and cutmix_batch are doing to the inputs, this could be handy, and 2) returning the permutation makes it possible to use the loss interpolation trick people often do to use mixup and cutmix with index labels after using mixup_batch and cutmix_batch if they so choose.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -281,8 +339,8 @@ def _rand_bbox(W: int,
 
     bbx1 = np.clip(cx - cut_w // 2, 0, W)
     bby1 = np.clip(cy - cut_h // 2, 0, H)
-    bbx2 = np.clip(cx + cut_w // 2, 0, W)
-    bby2 = np.clip(cy + cut_h // 2, 0, H)
+    bbx2 = np.clip(cx + (cut_w + 1) // 2, 0, W)
```
coryMosaicML, line 342: This is a change from the way the [original implementation ](https://github.com/clovaai/CutMix-PyTorch/blob/2d8eb68faff7fe4962776ad51d175c3b01a25734/train.py#L292) does things, and it could potentially change how cutmix behaves slightly, I think? Differences would likely be minor though.

composer/algorithms/mixup/mixup.py
```diff
@@ -19,75 +19,77 @@
...
+            num_classes = 10
+            X = torch.randn(N, C, H, W)
+            y = torch.randint(num_classes, size=(N,))
+            X_mixed, y_mixed, perm = mixup_batch(
+                X, y, num_classes=num_classes, alpha=0.2)
     """
-    if interpolation_lambda is None:
-        interpolation_lambda = _gen_interpolation_lambda(alpha)
+    if mixing is None:
+        mixing = _gen_interpolation_lambda(alpha)
```
coryMosaicML, line 73: Should this call something like `_gen_mixing(alpha)` instead?

composer/algorithms/mixup/mixup.py
```diff
@@ -171,30 +177,30 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
+            input,
+            target,
+            mixing=self.mixing,
+            num_classes=self.num_classes,
         )
 
         state.batch = (new_input, new_target)
 
 
 def _gen_interpolation_lambda(alpha: float) -> float:
```
coryMosaicML, line 192: Rename to `_gen_mixing`?

tests/algorithms/test_mixup.py
```diff
@@ -48,10 +48,10 @@ def test_mixup_batch(self, fake_data, alpha):
 
         # Apply mixup
         x_mix, _, _ = mixup_batch(
-            x=x_fake,
-            y=y_fake,
-            interpolation_lambda=interpolation_lambda,
-            n_classes=x_fake.size(1),  # Grab C
+            x_fake,
+            y_fake,
+            mixing=interpolation_lambda,
```
coryMosaicML, line 53: Should swap out `interpolation_lambda`  for `mixing` here too?

coryMosaicML (approved):

composer/algorithms/cutmix/cutmix.py
```diff
@@ -19,90 +19,147 @@
...
+            latter case, rows of ``target`` may be arbitrary vectors of targets,
+            including, e.g., one-hot encoded class labels, smoothed class
+            labels, or multi-output regression targets.
+        num_classes (int): total number of classes or output variables
+        length (float, optional): side length of the masked region. Must be
+            greater than 0. If ``0 < length < 1``, ``length`` is
+            interpreted as a fraction of ``H`` and ``W`` and the resulting box
+            is of size ``(length * H, length * W)``. If ``length >= 1``,
+            ``length`` the resulting box is of size ``(length, length)``.
+            Default: ``None``.
```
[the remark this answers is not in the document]
dblalock: Ya, only allowing fractional values makes sense to me. People can just set the fraction based on the H and W of the inputs if they want specific dimensions.

I'll rip out the `int` length option from this PR, and have a separate PR to rip it out from cutout.

composer/algorithms/cutmix/cutmix.py
dblalock: good point; I added this so that odd `length` values would be respected, but if we remove integer length arguments, all the changes to this function can get reverted I think

composer/algorithms/mixup/mixup.py
dblalock: I'm happy to go with whatever you think makes sense. I'm was hoping to mirror the notation from the paper internally for all the reasons you identified, but if you think having the differing names is net negative that works too.

I'll go with this change unless you tell me otherwise.

hanlint (commented):
I noticed two instances of significant test setup code to create the model and dataloaders. Are those needed, given that some of these are auto-loaded in https://github.com/mosaicml/composer/blob/dev/docs/source/doctest_fixtures.py ?

hanlint:
Also -- update yamls?

composer/algorithms/cutmix/cutmix.py
```diff
@@ -19,90 +19,147 @@
...
+        target_mixed (torch.Tensor): soft labels for mixed input samples.
+            These are a convex combination of the (possibly one-hot-encoded)
+            labels from the original samples and the samples chosen to fill
+            the masked regions, with the relative weighting equal to the
+            fraction of the spatial size that is cut.
+            E.g., if a sample was originally an image with label ``0`` and
+            40% of the image of was replaced with data from an image with label
+            ``2``, the resulting labels, assuming only three classes, would be
+            ``[1, 0, 0] * 0.6 + [0, 0, 1] * 0.4 = [0.6, 0, 0.4]``.
+        perm (torch.Tensor): the permutation used
```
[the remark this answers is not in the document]
dblalock: resolved by removing this return value

dblalock:
> Also -- update yamls?

The algorithm yamls? The label_smoothing one is updated, but cutmix and mixup fortunately don't have to change anything since they only use `alpha` and `num_classes`, and these are unchanged.

hanlint (approved):
LGTM

PR #663, opened 2022-03-03
Update GPT3-yamls for default 8xA100-40GB

abhi-mosaic:
This PR updates our GPT3 YAMLs with sizes >= 125M so that they will run without modification on 8xA100-40GB nodes.


siriuslee (approved):
LGTM

PR #665, opened 2022-03-03
[XS] Fix a TQDM deserialization bug

jbloxham:
No I'm not writing a test for this


hanlint (approved):
Unclear what the bug is, but I trust this is correct.

ravi-mosaicml (approved):
LGTM, makes sense

PR #667, opened 2022-03-04
Fix ZeRO config

jbloxham:
This is essentially a follow-up from https://github.com/mosaicml/composer/pull/322 - where it was missed that the ZeRO stage would have to be configured differently now. Annoyingly, there was nothing on DeepSpeed's end to catch this issue - but I've at least filed a ticket now: https://github.com/microsoft/DeepSpeed/issues/1809

This is probably not related to #610.

### DIFF
--- a/composer/trainer/trainer_hparams.py
+++ b/composer/trainer/trainer_hparams.py
@@ -424,7 +424,7 @@ class TrainerHparams(hp.Hparams):
             self.deepspeed["steps_per_print"] = cast(int, self.deepspeed.get("steps_per_print", 1e20))
 
             if "zero_optimization" in self.deepspeed:
-                zero_stage = cast(dict, self.deepspeed["zero_optimization"]).get("stage")
+                zero_stage = cast(dict, self.deepspeed["zero_optimization"]).get("stage", 0)
             else:
                 zero_stage = 0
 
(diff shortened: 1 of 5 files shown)


composer/trainer/trainer_hparams.py
```diff
@@ -421,10 +421,15 @@ def validate(self):
...
-            self.deepspeed["zero_stage"] = cast(int, self.deepspeed.get("zero_stage", 0))
             self.deepspeed["steps_per_print"] = cast(int, self.deepspeed.get("steps_per_print", 1e20))
 
-            if self.deterministic_mode and self.deepspeed["zero_stage"] > 0:
+            zero_stage = 0
+            zero_config = cast(dict,
+                               self.deepspeed["zero_optimization"]) if "zero_optimization" in self.deepspeed else None
+            if zero_config:
+                zero_stage = cast(int, zero_config.get("stage"))
+
```
hanlint, line 431: The double ifs and use of zero_config is a bit awkward, what about:
```
if "zero_optimization" in self.deepspeed:
    zero_stage = cast(dict, self.deepspeed["zero_optimization"]).get("stage")
else:
    zero_stage = 0
```

hanlint (approved):
One minor suggestion on code style

composer/trainer/trainer_hparams.py
jbloxham: I like that, thanks for the suggestion.

ravi-mosaicml (approved):
LGTM; I didn't look at the deepspeed docs but assume this is correct.

Issue #424, opened 2022-02-08
Remove the `WrappedDataloader` and `DataLoader` protocol; replace DataLoader types with `Iterable[Batch]`


hanlint:
We can remove our own DataLoader type and the WrappedDataLoader, and instead type the dataloader attribute on state with `Iterable[Batch]`. This is a prerequesite for FFCV integration, which uses custom dataloader types.

Steps:

- [ ] Anything that assumes a PyTorch dataloader should use an `if isisntance(dataloader, torch.utils.data.DataLoader)` check (e.g. algorithms that insert augmentations into vision datsets)
- [ ] Add in a `BEFORE_DATALOADER` event. This event should fire immediately before the underlying dataloader iterator is `__next__`'ed. See the `DataLoaderProfiler` for an example on how to wrap the dataloader
- [ ] Re-implement that DataloaderProfiler to use the `BEFORE_DATALOADER` event. It can measure the time between BEFORE_DATALOADER and AFTER_DATALOADER.
- [ ] Remove the `WrappedDataloader` class and `DataLoader` protocol class. On the `state`, update the dataloader type to `Iterable[Batch]`.

PR #672, opened 2022-03-04
ffcv integration for cifar10 dataset

dskhudia:
Plan is just add it for Cifar10 first along with taking care of dependencies and installation issues. 

ImageNet will be added in a later PR

Currently this PR is to get high level comments for integration. 

### Questions

- Should we preprocess data to ffcv format if the datafiles don't exist? 

### Instructions 

`apt update; apt install libturbojpeg-dev libopencv-dev` is needed before pip install -e .[ffcv] can work. 

If ffcv can't find opencv4, use `cp /usr/lib/x86_64-linux-gnu/pkgconfig/opencv.pc /usr/lib/x86_64-linux-gnu/pkgconfig/opencv4.pc`

Once this is installed, the following should work. 

Without ffcv: 

`composer -n 1 examples/run_composer_trainer.py -f composer/yamls/models/resnet9_cifar10.yaml --max_duration 1ep --datadir /localdisk/CIFAR10`

With ffcv:

This processes the data to ffcv format first in the /tmp (Default) dir first. 

`composer -n 1 examples/run_composer_trainer.py -f composer/yamls/models/resnet9_cifar10.yaml --train_dataset.cifar10.use_ffcv true --val_dataset.cifar10.use_ffcv true --train_dataset.cifar10.ffcv_write_dataset true --val_dataset.cifar10.ffcv_write_dataset true --datadir /localdisk/CIFAR10`

### DIFF
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -8,6 +8,7 @@ import textwrap
 from typing import TYPE_CHECKING, Callable, List, Optional, Sequence
 
 import torch
+import torch.utils.data
 
 from composer.utils.iter_helpers import ensure_tuple
 
@@ -95,7 +96,9 @@ class DataSpec:
             self.num_samples = num_samples
 
         else:
-            if isinstance(dataloader.dataset, collections.abc.Sized):
+            # FFCV dataloaders don't have an associated dataset
+            if isinstance(dataloader, torch.utils.data.DataLoader) and isinstance(dataloader.dataset,
+                                                                                  collections.abc.Sized):
                 try:
                     self.num_samples = len(dataloader.dataset)
                 except (TypeError, NotImplementedError):
--- a/composer/datasets/cifar.py
+++ b/composer/datasets/cifar.py
@@ -6,9 +6,13 @@ The CIFAR datasets are a collection of labeled 32x32 colour images. Please refer
 <https://www.cs.toronto.edu/~kriz/cifar.html>`_ for more details.
 """
 
+import logging
+import os
+import textwrap
 from dataclasses import dataclass
 from typing import List
 
+import torch
 import yahp as hp
 from torchvision import transforms
 from torchvision.datasets import CIFAR10
@@ -17,12 +21,15 @@ from composer.core.types import DataLoader
 from composer.datasets.dataloader import DataLoaderHparams
 from composer.datasets.hparams import DatasetHparams, SyntheticHparamsMixin, WebDatasetHparams
 from composer.datasets.synthetic import SyntheticBatchPairDataset
+from composer.datasets.utils import create_ffcv_dataset
 from composer.utils import dist
 
 __all__ = [
     "CIFAR10DatasetHparams", "CIFARWebDatasetHparams", "CIFAR10WebDatasetHparams", "CIFAR20WebDatasetHparams",
     "CIFAR100WebDatasetHparams"
 ]
+
+log = logging.getLogger(__name__)
 
 
 @dataclass
@@ -31,12 +38,25 @@ class CIFAR10DatasetHparams(DatasetHparams, SyntheticHparamsMixin):
 
     Args:
         download (bool): Whether to download the dataset, if needed. Default: ``True``.
+        use_ffcv (bool): Whether to use FFCV dataloaders. Default: ``False``.
+        ffcv_dir (str): A directory containing train/val <file>.ffcv files. If these files don't exist and
+            ``ffcv_write_dataset`` is ``True``, train/val <file>.ffcv files will be created in this dir. Default: ``"/tmp"``.
+        ffcv_dest_train (str): <file>.ffcv file that has training samples. Default: ``"cifar_train.ffcv"``.
+        ffcv_dest_val (str): <file>.ffcv file that has validation samples. Default: ``"cifar_val.ffcv"``.
+        ffcv_write_dataset (std): Whether to create dataset in FFCV format (<file>.ffcv) if it doesn't exist. Default:
+        ``False``.
     """
     download: bool = hp.optional("whether to download the dataset, if needed", default=True)
+    use_ffcv: bool = hp.optional("whether to use ffcv for faster dataloading", default=False)
+    ffcv_dir: str = hp.optional(
+        "A directory containing train/val <file>.ffcv files. If these files don't exist and ffcv_write_dataset is true, train/val <file>.ffcv files will be created in this dir.",
+        default="/tmp")
+    ffcv_dest_train: str = hp.optional("<file>.ffcv file that has training samples", default="cifar_train.ffcv")
+    ffcv_dest_val: str = hp.optional("<file>.ffcv file that has validation samples", default="cifar_val.ffcv")
+    ffcv_write_dataset: bool = hp.optional("Whether to create dataset in FFCV format (<file>.ffcv) if it doesn't exist",
+                                           default=False)
 
     def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHparams) -> DataLoader:
-        cifar10_mean = 0.4914, 0.4822, 0.4465
-        cifar10_std = 0.247, 0.243, 0.261
 
         if self.use_synthetic:
             total_dataset_size = 50_000 if self.is_train else 10_000
@@ -49,9 +69,83 @@ class CIFAR10DatasetHparams(DatasetHparams, SyntheticHparamsMixin):
                 memory_format=self.synthetic_memory_format,
             )
 
+        elif self.use_ffcv:
+            try:
+                import ffcv  # type: ignore
+                from ffcv.fields.decoders import IntDecoder, SimpleRGBImageDecoder  # type: ignore
+                from ffcv.pipeline.operation import Operation  # type: ignore
+            except ImportError:
+                raise ImportError(
+                    textwrap.dedent("""\
+                    Composer was installed without ffcv support.
+                    To use ffcv with Composer, please install ffcv in your environment."""))
+
+            dataset_file = self.ffcv_dest_train if self.is_train else self.ffcv_dest_val
+            dataset_filepath = os.path.join(self.ffcv_dir, dataset_file)
+            # always create if ffcv_write_dataset is true
+            if self.ffcv_write_dataset:
+                if dist.get_local_rank() == 0:
+                    if self.datadir is None:
+                        raise ValueError(
+                            "datadir is required if use_synthetic is False and ffcv_write_dataset is True.")
+                    ds = CIFAR10(
+                        self.datadir,
+                        train=self.is_train,
+                        download=self.download,
+                    )
+
+                    create_ffcv_dataset(dataset=ds, write_path=dataset_filepath)
+
+                # Wait for the local rank 0 to be done creating the dataset in ffcv format.
+                dist.barrier()
+
+            if not os.path.exists(dataset_filepath):
+                raise ValueError(
+                    f"Dataset file containing samples not found at {dataset_filepath}. Use ffcv_dir flag to point to a dir containing {dataset_filepath}."
+                )
+
+            # Please note that this mean/std is different from the mean/std used for regular PyTorch dataloader as
+            # ToTensor does the normalization for PyTorch dataloaders.
+            cifar10_mean_ffcv = [125.307, 122.961, 113.8575]
+            cifar10_std_ffcv = [51.5865, 50.847, 51.255]
+            label_pipeline: List[Operation] = [IntDecoder(), ffcv.transforms.ToTensor(), ffcv.transforms.Squeeze()]
+            image_pipeline: List[Operation] = [SimpleRGBImageDecoder()]
+
+            if self.is_train:
+                image_pipeline.extend([
+                    ffcv.transforms.RandomHorizontalFlip(),
+                    ffcv.transforms.RandomTranslate(padding=2, fill=tuple(map(int, cifar10_mean_ffcv))),
+                    ffcv.transforms.Cutout(4, tuple(map(int, cifar10_mean_ffcv))),
+                ])
+            # Common transforms for train and test
+            image_pipeline.extend([
+                ffcv.transforms.ToTensor(),
+                ffcv.transforms.ToTorchImage(channels_last=False, convert_back_int16=False),
+                ffcv.transforms.Convert(torch.float32),
+                transforms.Normalize(cifar10_mean_ffcv, cifar10_std_ffcv),
+            ])
+
+            ordering = ffcv.loader.OrderOption.RANDOM if self.is_train else ffcv.loader.OrderOption.SEQUENTIAL
+
+            return ffcv.Loader(
+                dataset_filepath,
+                batch_size=batch_size,
+                num_workers=dataloader_hparams.num_workers,
+                order=ordering,
+                distributed=False,
+                pipelines={
+                    'image': image_pipeline,
+                    'label': label_pipeline
+                },
+                batches_ahead=dataloader_hparams.prefetch_factor,
+                drop_last=self.drop_last,
+            )
         else:
             if self.datadir is None:
                 raise ValueError("datadir is required if use_synthetic is False")
+
+            cifar10_mean = 0.4914, 0.4822, 0.4465
+            cifar10_std = 0.247, 0.243, 0.261
 
             if self.is_train:
                 transformation = transforms.Compose([
--- a/composer/datasets/utils.py
+++ b/composer/datasets/utils.py
@@ -4,21 +4,21 @@
 
 import logging
 import textwrap
-from typing import Callable, List, Tuple, Union
+from typing import Callable, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
-import torch.utils.data
 from PIL import Image
 from torchvision import transforms
 from torchvision.datasets import VisionDataset
 
-from composer.core.types import Batch
+from composer.core.types import Batch, Dataset
 
 __all__ = [
     "add_vision_dataset_transform",
     "NormalizationFn",
     "pil_image_collate",
+    "create_ffcv_dataset",
 ]
 
 log = logging.getLogger(__name__)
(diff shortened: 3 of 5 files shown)


composer/datasets/cifar10.py
```diff
@@ -36,6 +40,54 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+                ToTorchImage(channels_last=False, convert_back_int16=False),
+                Convert(torch.float32),
+                transforms.Normalize(cifar10_ffcv_mean, cifar10_ffcv_std),
+            ])
+
+            ordering = OrderOption.RANDOM if self.is_train else OrderOption.SEQUENTIAL
+
+            filepath = "/cifar10_train.beton" if self.is_train else "/cifar10_test.beton"
+
+            return Loader(self.datadir + filepath,
```
hanlint, line 82: this is more future-thinking, but some of our algorithms run via inserting into dataset transforms. Is there an analogus mechnism here where FFCV-compliant transforms can be inserted to the pipelines _after_ the dataloader is created?

composer/core/data_spec.py
```diff
@@ -95,7 +95,8 @@ def __init__(
             self.num_samples = num_samples
 
         else:
-            if isinstance(dataloader.dataset, collections.abc.Sized):
+            # FFCV dataloaders don't have an associated dataset
+            if hasattr(dataloader, "dataset") and isinstance(dataloader.dataset, collections.abc.Sized):
```
hanlint, line 99: may want to check with `isinstance` rather than ducktype for dataset and sampler but defer to @ravi-mosaicml here

composer/datasets/cifar10.py
dskhudia: We can insert any torchvision transforms here with ffcv (or write our own in ffcv format) but we need to add "recompile" flag so that ffcv can recompile the data processing pipeline.

composer/core/data_spec.py
ravi-mosaicml: We should probably remove our own `DataLoader` type and the `WrappedDataLoader` (see #424), and instead type the `dataloader` attribute on state with `Iterable[Batch]`. Anything that needs to assume it is a PyTorch dataloader should use   `if isisntance(dataloader, torch.utils.data.DataLoader)` check.

We'll also need to re-implement that `DataloaderProfiler`. Callbacks sholdn't be modifying state to begin with (which it does), and without the WrappedDataLoader, then `isisntance(dataloader, torch.utils.data.DataLoader)` checks will fail once it has been wrapped. Instead, I'm thinking we can add in the `BEFORE_DATALOADER` event to fire immediately before the underlying dataloader iterator is `__next__`'ed, and then restructure the `DataloaderProfiler` to simply measure the time between the before and after calls.

composer/datasets/cifar10.py
```diff
@@ -36,6 +40,54 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+            except ImportError:
+                raise ImportError(
+                    textwrap.dedent("""\
+                    Composer was installed without ffcv support.
+                    To use ffcv with Composer, run `pip install mosaicml[ffcv]`"""))
+
+            if self.datadir is None:
+                raise ValueError("datadir is required if use_synthetic is False")
+
+            cifar10_ffcv_mean = [125.307, 122.961, 113.8575]
```
ravi-mosaicml, line 59: Could you move this variable up so it is shared across the FFCV and non-ffcv codepaths?

composer/trainer/trainer.py
```diff
@@ -851,14 +851,16 @@ def _spin_dataloaders(self):
         # so it does not affect any other RNG reads
         for evaluator in self.state.evaluators:
             dataloader = evaluator.dataloader.dataloader
-            if isinstance(dataloader.sampler, torch.utils.data.DistributedSampler):
+            # FFCV dataloaders use their own sampling strategy
+            if hasattr(dataloader, "sampler") and isinstance(dataloader.sampler, torch.utils.data.DistributedSampler):
```
ravi-mosaicml, line 855: Same comment re: using `isinstance(dataloader, torch.utils.data.Dataloader)` rather than `hasattr`

composer/datasets/cifar10.py
dskhudia: Since they are compiling the code to machine code, if we modify in-place we still need to recompile it otherwise it uses stale code.

composer/datasets/cifar10.py
```diff
@@ -36,6 +40,54 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+            image_pipeline.extend([
+                ToTensor(),
+                ToTorchImage(channels_last=False, convert_back_int16=False),
+                Convert(torch.float32),
+                transforms.Normalize(cifar10_ffcv_mean, cifar10_ffcv_std),
+            ])
+
+            ordering = OrderOption.RANDOM if self.is_train else OrderOption.SEQUENTIAL
+
+            filepath = "/cifar10_train.beton" if self.is_train else "/cifar10_test.beton"
```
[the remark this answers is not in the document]
dskhudia: Yes. I had a related question in the description. 

> Should we preprocess data to ffcv format if the datafiles don't exist?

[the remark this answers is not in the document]
dskhudia: I was talking to @bandish-shah as well and this seems like the consensus that we have a flag that controls whether we write the dataset in ffcv format if it doesn't exist.

composer/datasets/cifar10.py
```diff
@@ -36,6 +40,54 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataloaderHpara
...
+
+            return Loader(self.datadir + filepath,
+                          batch_size=batch_size,
+                          num_workers=dataloader_hparams.num_workers,
```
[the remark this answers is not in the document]
dskhudia: Yes. I think there is one or more other args also missing and will add that.

composer/datasets/cifar10.py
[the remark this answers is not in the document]
bandish-shah: The reasoning here was that we have a `download` flag which controls whether or not to automatically download the dataset if it is not found in the specified location.  Similarly we we would have a `convert` flag if the beton file doesn't exist in the specified location.

composer/core/data_spec.py
dskhudia: Replaced with isinstance and rest will be taken care of in #424

composer/datasets/cifar10.py
dskhudia: Done

composer/datasets/cifar.py
```diff
@@ -24,19 +28,34 @@
...
+            ``ffcv_write_dataset`` is ``True``, train/val <file>.beton files will be created in this dir. Default: ``"/tmp"``.
+        ffcv_dest_train (str): <file>.beton file that has training samples. Default: ``"cifar_train.beton"``.
+        ffcv_dest_val (str): <file>.beton file that has validation samples. Default: ``"cifar_val.beton"``.
+        ffcv_write_dataset (std): Whether to create dataset in FFCV format (<file>.beton) if it doesn't exist. Default:
+        ``False``.
     """
     download: bool = hp.optional("whether to download the dataset, if needed", default=True)
+    use_ffcv: bool = hp.optional("whether to use ffcv for faster dataloading", default=False)
+    ffcv_dir: str = hp.optional(
+        "A directory containing train/val <file>.beton files. If these files don't exist and ffcv_write_dataset is true, train/val <file>.beton files will be created in this dir.",
```
ravi-mosaicml, line 51: Can we separate out the FFCV dataset writing into a seperate helper function. I'm thinking this function would have a signature like:

```python
def create_ffcv_dataset(dataset: Iterable[Sample], desetination_filepath: str):
    """Iterates through ``dataset`` and converts it into FFCV beton format at filepath ``destination_filepath``."""
```

Then, the user can use a seperate CLI tool to create the dataset, and pass in the ffcv path here.

The reasoning behind separating this out is so 1) ffcv users are not bound to our hparams framework, 2) the dataset hparams is used for loading datasets, and 3) I would like this functionality to be re-usable as we add other datasets in.

composer/datasets/cifar.py
```diff
@@ -49,10 +68,84 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
                 memory_format=self.synthetic_memory_format,
             )
 
+        elif self.use_ffcv:
+            try:
+                import ffcv  # type: ignore
```
ravi-mosaicml, line 73: Curious why the type ignore?

composer/datasets/cifar.py
```diff
@@ -49,10 +68,84 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+            except ImportError:
+                raise ImportError(
+                    textwrap.dedent("""\
+                    Composer was installed without ffcv support.
+                    To use ffcv with Composer, run `pip install mosaicml[ffcv]`"""))
+
+            dataset_file = self.ffcv_dest_train if self.is_train else self.ffcv_dest_val
+            dataset_filepath = os.path.join(self.ffcv_dir, dataset_file)
+            # always create if ffcv_write_dataset is true
+            if self.ffcv_write_dataset and not os.path.exists(dataset_filepath):
```
ravi-mosaicml, line 85: Would this have a race condition with multi-rank training?

composer/datasets/cifar.py
```diff
@@ -49,10 +68,84 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+                    'label': ffcv.fields.IntField()
+                })
+                writer.from_indexed_dataset(ds)
+
+            if not os.path.exists(dataset_filepath):
+                raise ValueError(
+                    f"Dataset file containing samples not found at {dataset_filepath}. Use ffcv_dir flag to point to a dir containing {dataset_filepath}."
+                )
+
+            # Please note that this mean/std is different from the mean/std used for regular PyTorch dataloader as
```
ravi-mosaicml, line 106: oh man.....is there a way around this? ideally these constants would be the same regardless of dataset format.

composer/datasets/cifar.py
```diff
@@ -49,10 +68,84 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+
+            if self.is_train:
+                image_pipeline.extend([
+                    ffcv.transforms.RandomHorizontalFlip(),
+                    ffcv.transforms.RandomTranslate(padding=2, fill=tuple(map(int, cifar10_mean_ffcv))),
+                    ffcv.transforms.Cutout(4, tuple(map(int, cifar10_mean_ffcv))),
+                ])
+            # Common transforms for train and test
+            image_pipeline.extend([
+                ffcv.transforms.ToTensor(),
```
ravi-mosaicml, line 121: This is probably beyond the scope of this PR, but would it be possible to automatically convert a ComposedTransformation pipleine into a corresponding FFCV transformation pipeline? This would presume there is a 1:1 mapping between torchvision transforms and ffcv transforms. It would be nice if this didn't have to be manually defined for each dataset.

composer/datasets/cifar.py
```diff
@@ -49,10 +68,84 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+                ffcv.transforms.ToTorchImage(channels_last=False, convert_back_int16=False),
+                ffcv.transforms.Convert(torch.float32),
+                transforms.Normalize(cifar10_mean_ffcv, cifar10_std_ffcv),
+            ])
+
+            ordering = ffcv.loader.OrderOption.RANDOM if self.is_train else ffcv.loader.OrderOption.SEQUENTIAL
+
+            return ffcv.Loader(
```
ravi-mosaicml, line 129: surprised this doesn't raise a pyright error....

composer/trainer/trainer.py
```diff
@@ -912,14 +912,17 @@ def _spin_dataloaders(self):
...
+            if isinstance(dataloader, torch.utils.data.DataLoader) and isinstance(dataloader.sampler,
+                                                                                  torch.utils.data.DistributedSampler):
                 dataloader.sampler.set_epoch(0)
             for _ in dataloader:
                 break
 
         # spin the train dataloader's sampler to get to the state of the desired epoch
         for epoch in range(int(self.state.timer.epoch)):
-            if isinstance(self.state.train_dataloader.sampler, torch.utils.data.DistributedSampler):
+            if hasattr(self.state.train_dataloader, "sampler") and isinstance(self.state.train_dataloader.sampler,
```
ravi-mosaicml, line 924: Can you add a `# TODO` in the comments, saying that the hasattr check will be removed in #424?

composer/trainer/trainer.py
```diff
@@ -962,7 +965,8 @@ def _train_loop(self) -> None:
                     self.engine.run_event(Event.EPOCH_START)
                     self.logger.data_epoch({"epoch": int(self.state.timer.epoch)})
 
-                if isinstance(self.state.train_dataloader.sampler, torch.utils.data.DistributedSampler):
+                if hasattr(self.state.train_dataloader, "sampler") and isinstance(self.state.train_dataloader.sampler,
```
ravi-mosaicml, line 968: Same as above: Can you add a `# TODO` in the comments, saying that the hasattr check will be removed in #424?

composer/trainer/trainer.py
```diff
@@ -1212,7 +1216,8 @@ def eval(self, is_batch: bool):
             for evaluator in self.state.evaluators:
                 dataloader = evaluator.dataloader.dataloader
                 metrics = self._ensure_metrics_device_and_dtype(evaluator.metrics)
-                if isinstance(dataloader.sampler, torch.utils.data.DistributedSampler):
+                if hasattr(dataloader, "sampler") and isinstance(dataloader.sampler,
```
ravi-mosaicml, line 1219: Same as above: Can you add a `# TODO` in the comments, saying that the hasattr check will be removed in #424?

composer/datasets/cifar.py
dskhudia: Was getting pyright error because these were conditional imports.

composer/datasets/cifar.py
dskhudia: Yes. will fix this.

composer/datasets/cifar.py
dskhudia: This is one of the pain points. There is no 1:1 mapping between torchvision transforms and ffcv transformations. In fact, we will need to write some of ffcv transformations. Some organizations around transformations will come up when we do more experiments.

composer/datasets/cifar.py
dskhudia: :-)

composer/trainer/trainer.py
dskhudia: will do.

composer/trainer/trainer.py
dskhudia: will do.

composer/trainer/trainer.py
dskhudia: Will do.

composer/datasets/cifar.py
dskhudia: For CIFAR, it seems there isn't as the way these are used is different between torch vision and ffcv. However, the good news it's the same for ImageNet dataset.

composer/datasets/cifar.py
dskhudia: Only rank 0 creates it.

composer/datasets/cifar.py
dskhudia: Since ffcv is not installed by default, pyright throws lint error without ignore.

composer/datasets/cifar.py
dskhudia: Added a new create_ffcv_dataset function in composer/datasets/utils.py. The same can be used for imagenet dataset as well.

composer/datasets/utils.py
```diff
@@ -162,3 +163,45 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
         else:
             dataset.transform = transforms.Compose([dataset.transform, transform])
             log.warning(transform_added_logstring)
+
+
+def create_ffcv_dataset(dataset: IterableDataset,
```
ravi-mosaicml, line 168: Does it need to be an IterableDataset?

```suggestion
def create_ffcv_dataset(dataset: Dataset,
```

composer/trainer/trainer.py
```diff
@@ -910,14 +910,18 @@ def _spin_dataloaders(self):
...
+            if isinstance(dataloader, torch.utils.data.DataLoader) and isinstance(dataloader.sampler,
+                                                                                  torch.utils.data.DistributedSampler):
                 dataloader.sampler.set_epoch(0)
             for _ in dataloader:
                 break
 
         # spin the train dataloader's sampler to get to the state of the desired epoch
         for epoch in range(int(self.state.timer.epoch)):
-            if isinstance(self.state.train_dataloader.sampler, torch.utils.data.DistributedSampler):
+            # TODO: isinstance check will be removed while fixing https://github.com/mosaicml/composer/issues/424
```
ravi-mosaicml, line 922:
```suggestion
            # TODO: hasattr check will be removed while fixing https://github.com/mosaicml/composer/issues/424
```

composer/datasets/cifar.py
```diff
@@ -17,26 +21,42 @@
...
+        ``False``.
     """
     download: bool = hp.optional("whether to download the dataset, if needed", default=True)
+    use_ffcv: bool = hp.optional("whether to use ffcv for faster dataloading", default=False)
+    ffcv_dir: str = hp.optional(
+        "A directory containing train/val <file>.ffcv files. If these files don't exist and ffcv_write_dataset is true, train/val <file>.ffcv files will be created in this dir.",
+        default="/tmp")
+    ffcv_dest_train: str = hp.optional("<file>.ffcv file that has training samples", default="cifar_train.ffcv")
+    ffcv_dest_val: str = hp.optional("<file>.ffcv file that has validation samples", default="cifar_val.ffcv")
+    ffcv_write_dataset: bool = hp.optional("Whether to create dataset in FFCV format (<file>.ffcv) if it doesn't exist",
```
ravi-mosaicml, line 56: Discussed offline -- ok with this field now, but can you create a GH issue to track the design for non-yahp conversion to a FFCV dataset?

composer/datasets/cifar.py
```diff
@@ -49,10 +69,79 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+            except ImportError:
+                raise ImportError(
+                    textwrap.dedent("""\
+                    Composer was installed without ffcv support.
+                    To use ffcv with Composer, please install ffcv in your environment."""))
+
+            dataset_file = self.ffcv_dest_train if self.is_train else self.ffcv_dest_val
+            dataset_filepath = os.path.join(self.ffcv_dir, dataset_file)
+            # always create if ffcv_write_dataset is true
+            if dist.get_global_rank() == 0 and self.ffcv_write_dataset:
```
ravi-mosaicml, line 86:
```suggestion
            if dist.get_local_rank() == 0 and self.ffcv_write_dataset:
```

composer/datasets/cifar.py
```diff
@@ -49,10 +69,79 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+                if self.datadir is None:
+                    raise ValueError("datadir is required if use_synthetic is False and ffcv_write_dataset is True.")
+                ds = CIFAR10(
+                    self.datadir,
+                    train=self.is_train,
+                    download=self.download,
+                )
+
+                create_ffcv_dataset(dataset=ds, write_path=dataset_filepath)
+
```
ravi-mosaicml, line 96: A `dist.barrier()` is needed here to ensure that rank zero has finished downloading before rank one attempts to read it.

ravi-mosaicml (approved):
LGTM; nice work! See comments.

composer/datasets/cifar.py
dskhudia: https://github.com/mosaicml/composer/issues/799

PR #674, opened 2022-03-04
Yahp version update to 0.1.0

Averylamp:


ravi-mosaicml (approved):
LGTM, though let's use a type ignore for the type error right now...need to fix the underlying base class.

tests/test_dataset_registry.py
```diff
@@ -46,7 +46,7 @@
             split="train",
         ),
     COCODatasetHparams:
-        lambda: COCODatasetHparams(is_train=False, datadir=["hello"], download=False, drop_last=False, shuffle=False),
+        lambda: COCODatasetHparams(is_train=False, datadir="hello", download=False, drop_last=False, shuffle=False),
```
Averylamp, line 49:
```suggestion
        lambda: COCODatasetHparams(is_train=False, datadir=["hello"], download=False, drop_last=False, shuffle=False), # type: ignore
```

hanlint (approved):
LGTM

Issue #640, opened 2022-03-02
Update `ensure_tuple` to support generic sequences for non-strings.


ravi-mosaicml:
Update `ensure_tuple` to support generic sequences for non-strings.

Then, the annotations can be `x | Sequence[x]`, which is both concise and explicit.

_Originally posted by @ravi-mosaicml in https://github.com/mosaicml/composer/pull/632#discussion_r817276515_

Issue #510, opened 2022-02-17
Confusing `Scheduler` and `Schedulers` types


jfrankle:
It makes sense to have types like Scheduler and Schedulers but when trying to use the library, that s is easy to miss and explicitly laying out the type in things like Trainer initialization would help.

ravi-mosaicml:
Good point, in the docstrings, we should use the `Scheduler | List[Scheduler] | Tuple[Scheduler]` notation rather than `Schedulers`.

PR #677, opened 2022-03-05
Remove plural types and aliases for native pytorch types

Landanjs:
I may have misunderstood the assignment, but attempting to address #640 and #510. 

Essentially:
- Added a check for `range` objects in `ensure_tuple()` since these were the last type of `Sequence` we did not support.
- Changed `T | Tuple[T] | List[T]` in `Many` to `T | Sequence[T]`
- Changed docstrings and instances of `T | Tuple | List`  to `Many`

If this actually looks right, I will add tests for `ensure_tuple` and every sequence type.

Side note: I'm a little confused when to use `PytorchSchedulers` vs `ComposerSchedulers`


ravi-mosaicml:
There's some additional typing stuff will need to be adjust in `iter_helpers.pyi` -- specifically we'll need an additional overload methods to take a generic sequence and for `str`, `bytes`, and `range` objects.

composer/core/state.py
```diff
@@ -73,9 +73,10 @@ class State(Serializable):
         precision (str | Precision): The numerical precision to use for training. See :class:`~.Precision` for
             the supported precisions.
         precision_context (Callable[[Precision], ContextManager]): Function to produce a context manager to mandate precision.
-        optimizers (:attr:`~.types.Optimizers`, optional): The optimizers being used to train the model. Multiple optimizers are not currently supported.
-        schedulers (:attr:`~.types.PyTorchScheduler` | List[:attr:`~.types.PyTorchScheduler`] | Tuple[:attr:`~.types.PyTorchScheduler`, ...], optional):
-            The learning rate scheduler (can also be a list or tuple of schedulers).
+        optimizers (:attr:`~.types.Optimizers`, optional): The optimizers being used to train the model.
+            Multiple optimizers are not currently supported.
+        schedulers (Many[:attr:`~.types.PytorchScheduler`], optional): The learning rate scheduler
```
ravi-mosaicml, line 78: While more concise, I would prefer not to use the `Many` notation in docstrings because it is not super clear on what `Many` means (one would have to look at the underlying type annotation for that).

Instead, could we do something like this?

```suggestion
        schedulers (types.PyTorchScheduler | Sequence[types.PyTorchScheduler], optional): The learning rate scheduler
```

ravi-mosaicml, line 78: While more concise, I would prefer not to use the `Many` notation in docstrings. Instead, could we do something like this?
```suggestion
        schedulers (types.PyTorchScheduler | Sequence[types.PyTorchScheduler], optional): The learning rate scheduler
```
```suggestion
        schedulers (Many[:attr:`~.types.PytorchScheduler`], optional): The learning rate scheduler
```

composer/core/types.py
```diff
@@ -17,7 +17,7 @@
     BatchDict (Dict[str, Tensor]): Commonly used in natural language processing tasks.
     Metrics (Metric | MetricCollection): Union type covering common formats for representing metrics.
     Optimizer (torch.optim.Optimizer): Alias for :class:`torch.optim.Optimizer`
-    Optimizers (Optimizer | List[Optimizer] | Tuple[Optimizer, ...]): Union type for indeterminate amounts of optimizers.
+    Optimizers (Many[Optimizer]): Union type for indeterminate amounts of optimizers.
```
ravi-mosaicml, line 20: same thing

composer/core/state.py
```diff
@@ -73,9 +73,10 @@ class State(Serializable):
         precision (str | Precision): The numerical precision to use for training. See :class:`~.Precision` for
             the supported precisions.
         precision_context (Callable[[Precision], ContextManager]): Function to produce a context manager to mandate precision.
-        optimizers (:attr:`~.types.Optimizers`, optional): The optimizers being used to train the model. Multiple optimizers are not currently supported.
-        schedulers (:attr:`~.types.PyTorchScheduler` | List[:attr:`~.types.PyTorchScheduler`] | Tuple[:attr:`~.types.PyTorchScheduler`, ...], optional):
-            The learning rate scheduler (can also be a list or tuple of schedulers).
+        optimizers (:attr:`~.types.Optimizers`, optional): The optimizers being used to train the model.
```
ravi-mosaicml, line 76: Following on to @jfrankle's feedback from #510, it would make the docstrings clearer if we did not use the plural types, and instead did something like:

```suggestion
        optimizers (types.Optimizer | Sequence[types.Optimizer], optional): The optimizers being used to train the model.
```
```suggestion
        optimizers (:attr:`~.types.Optimizers`, optional): The optimizers being used to train the model.
```

composer/core/state.py
hanlint: Unfortunately the new Union Operator `|` is only in Python 3.10 (PEP 604).

ravi-mosaicml:
> Side note: I'm a little confused when to use `PytorchSchedulers` vs `ComposerSchedulers`

Good point; this will need to be clarified in the docs somewhere. At a high level, `ComposerSchedulers` are acceptable to use when constructing the trainer, but they are converted to `PyTorchSchedulers` which are used internally. (ComposerSchedulers simply return the learning rate factor and don't update the optimizer; instead, the trainer "compiles" them into a PyTorch LambdaLR scheduler)

Landanjs:
I think I addressed most comments, but I'm not very familiar with interfaces so it's likely a set that up wrong 😅. 

I will start making tests now

I have a couple of docstring consistency questions that maybe should be addressed somewhere else:
1. Should I remove the :attr:` from types in the state docstring (specifically in `State`)? They don't seem necessary, but not sure if that is the appropriate format. The dataloaders don't seem to have these.
2. Should we use the aliases `Tensor` and `Optimizer` in docstrings? To me, it adds an extra click to figure out the actual type with little benefit.
3. Should Algorithms and Callbacks have the same type format as optimizers and schedulers i.e. they accept a single object or a sequence of objects? Right now, they are `Sequence[Algorithms]`
4. To align with removing plurals, should `Tensors` -> `Tensor | Sequence[Tensor]` in docstrings?

Landanjs:
Errr, there are a lot of `Sequence` object errors now... For example:

```
    first_optimizer = state.optimizers[0]
    expected_param_groups = deepcopy(first_optimizer.param_groups)
    
```

The second line gets `error: Cannot access member "param_groups" for type "Sequence[Optimizer]"`. Do these need to be manually cast?

composer/core/state.py
ravi-mosaicml: Yup, the bitwise or operator is fine in docstrings (nothing is really checking it anyways) + more concise and clear than `Union[...]`

composer/core/state.py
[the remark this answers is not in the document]
ravi-mosaicml: Fair point, and I don't see us supporting multiple optimizers anytime soon? In that case, yeah, feel free to just make it the singleton.

ravi-mosaicml:
> I think I addressed most comments, but I'm not very familiar with interfaces so it's likely a set that up wrong sweat_smile.
> 
> I will start making tests now
> 
> I have a couple of docstring consistency questions that maybe should be addressed somewhere else:
> 
> 1. Should I remove the :attr:`from types in the state docstring (specifically in`State`)? They don't seem necessary, but not sure if that is the appropriate format. The dataloaders don't seem to have these.

It depends on what is being referenced; sometimes, `:attr:` is needed 

> 2. Should we use the aliases `Tensor` and `Optimizer` in docstrings? To me, it adds an extra click to figure out the actual type with little benefit.

Good point; let's go with the torch types. I'd eventually like to remove types.py or make it significantly leaner.

> 3. Should Algorithms and Callbacks have the same type format as optimizers and schedulers i.e. they accept a single object or a sequence of objects? Right now, they are `Sequence[Algorithms]`

Yes, let's do the "singleton or sequence" approach.

> 4. To align with removing plurals, should `Tensors` -> `Tensor | Sequence[Tensor]` in docstrings?

Yup!

ravi-mosaicml:
> Errr, there are a lot of `Sequence` object errors now... For example:
> 
> ```
>     first_optimizer = state.optimizers[0]
>     expected_param_groups = deepcopy(first_optimizer.param_groups)
>     
> ```
> 
> The second line gets `error: Cannot access member "param_groups" for type "Sequence[Optimizer]"`. Do these need to be manually cast?

It shouldn't need a manual cast; double check the type annotation for `ensure_tuple`.

ravi-mosaicml:
@Landanjs I pushed a commit that (hopefully?) fixes the type errors with the `ensure_tuple` change.
If you could finish up the docs fixes in this PR, that'd be awesome.

ravi-mosaicml (commented):
Looking good; feel free to continue to remove the plural types from the codebase (and from types.py)

Landanjs:
The recent change tries to:
1. Use PyTorch's Tensor, Optimizer, Module classes where possible
2. Remove Tenors
3. Attempts to fix missing docstring links by prepending `types.`
4. Remove the illusion of multiple optimizer from docstrings

I did not convert Callbacks and Loggers to "singleton or sequence" format since these are converted into lists instead of tuples and I'm not confident the functionality is the same? But maybe should be changed in the future

composer/core/types.py
```diff
@@ -7,29 +7,31 @@
...
     Metrics (Metric | MetricCollection): Union type covering common formats for representing metrics.
     Optimizer (torch.optim.Optimizer): Alias for :class:`torch.optim.Optimizer`
-    Optimizers (Optimizer | List[Optimizer] | Tuple[Optimizer, ...]): Union type for indeterminate amounts of optimizers.
+    Optimizers (torch.optim.Optimizer | Sequence[torch.optim.Optimizer]): Union type for indeterminate amounts of
+        optimizers.
     PyTorchScheduler (torch.optim.lr_scheduler._LRScheduler): Alias for base class of learning rate schedulers such
         as :class:`torch.optim.lr_scheduler.ConstantLR`
     Scaler (torch.cuda.amp.grad_scaler.GradScaler): Alias for :class:`torch.cuda.amp.GradScaler`.
     JSON (str | float | int | None | List['JSON'] | Dict[str, 'JSON']): JSON Data
     Evaluators (Many[Evaluator]): Union type for indeterminate amounts of evaluators.
```
ravi-mosaicml, line 27: Same as above -- can you remove this docstring (and the associated type alias below in the code)?

composer/core/types.py
```diff
@@ -61,14 +63,14 @@
 ]
 
 T = TypeVar('T')
-Many = Union[T, Tuple[T, ...], List[T]]
+Many = Union[T, Sequence[T]]
```
ravi-mosaicml, line 66: Let's remove `Many` too

composer/core/types.py
```diff
@@ -61,14 +63,14 @@
...
 
 Tensor = torch.Tensor
-Tensors = Many[Tensor]
+Tensors = Union[Tensor, Sequence[Tensor]]
```
ravi-mosaicml, line 69: Same thing...removing the aliases here will ensure that we removed them throughout the entire codebase, and use the underlying `Union[X, Sequence[X]]` notation in the codebase throughout.

composer/trainer/trainer.py
```diff
@@ -146,12 +146,13 @@ class Trainer:
...
 
             .. seealso:: :mod:`composer.algorithms` for the different algorithms built into Composer.
-        optimizers (Optimizers, optional): The optimizer.
+        optimizers (torch.optim.Optimizer, optional): The optimizer.
             If ``None``, will be set to ``DecoupledSGDW(model.parameters(), lr=0.1)``. (default: ``None``)
 
             .. seealso:: :mod:`composer.optim` for the different optimizers built into Composer.
-        schedulers (Schedulers, optional): The learning rate schedulers. If ``[]`` or ``None``, will be set to
-            ``[constant_scheduler]``. (default: ``None``).
+        schedulers (types.PyTorchScheduler or Sequence[types.PyTorchScheduler], optional):
```
ravi-mosaicml, line 153: The trainer takes both pytorch schedulers or composer schedulers. 

```suggestion
        schedulers (types.ComposerScheduler | types.PyTorchScheduler | Sequence[types.ComposerScheduler | types.PyTorchScheduler], optional):
```

composer/trainer/trainer.py
```diff
@@ -146,12 +146,13 @@ class Trainer:
...
+            The learning rate schedulers. If ``[]`` or ``None``, will be set to ``[constant_scheduler]``.
```
ravi-mosaicml, line 154:
```suggestion
            The learning rate schedulers. If ``[]`` or ``None``, the learning rate will be constant.
```

ravi-mosaicml:
There appear to be a few docstring formatting warnings that need to be fixed (Docstring warnings are treated as errors and cause the build to fail).

```
00:04:06.043  /home/mosaicml/.local/lib/python3.9/site-packages/composer/trainer/trainer.py:docstring of composer.trainer.trainer.Trainer:39: WARNING: Field list ends without a blank line; unexpected unindent.
00:04:06.043  /home/mosaicml/.local/lib/python3.9/site-packages/composer/trainer/trainer.py:docstring of composer.trainer.trainer.Trainer:42: WARNING: Inline substitution_reference start-string without end-string.
00:04:06.043  /home/mosaicml/.local/lib/python3.9/site-packages/composer/trainer/trainer.py:docstring of composer.trainer.trainer.Trainer:44: WARNING: Definition list ends without a blank line; unexpected unindent.
```

ravi-mosaicml (approved):
LGTM!

Issue #497, opened 2022-02-16
Ignore parameters from a checkpoint


moinnadeem:
# Motivation

Some GLUE tasks require mid-training: for tasks with small, noisy datasets:
1) First, pre-train BERT on some generic language modeling corpus. (ie. Wikipedia)
2) Then, fine-tune BERT on a task-specific corpus that helps logical inference. For instance, fine-tune BERT on a 3-way classification task on a large, labeled dataset.
3) Finally, fine-tune BERT on your small, relevant task-specific corpus. For instance, a 2-way classification task.

For #3, we load the checkpoint obtained after completing #2. 

This implies that BERT was trained with a task-specific classification head (in code, there is now a `module.model.classifier.weight` and `module.model.classifier.bias`) that exists that is irrelevant to us. This is because:
a) The classification weights are from a task that doesn't matter to us
b) The classification weights are a 3-way classification task, and we now have a 2-way classification task. 

Because of B), the matrix sizes will not match up, and training will crash. In code, the checkpoint from B is something like `batch_size x 3`, and our model has a matrix of shape `batch_size x 2`.

This motivates ignoring certain parameters from a checkpoint.

# Solutions
I see three proposed solutions here:
1. As a part of our checkpointing code, we can decide to ignore certain user-specfied parameters when saving the checkpoint.
2. As a part of our checkpoint loading code, we can ignore certain user-specfied parameters when loading the checkpoint.
3. As a part of checkpoint loading, for any keys with the same name, we can ignore any checkpoint parameters that have different shapes than the model weights, and print a warning.

Before continuing, please decide which one you like the best so I don't bias you.

To unblock myself previously, I have implemented #2: it will delete user-specified weights from the dictionary. This consists of a parameter in checkpointing called `ignore_model_keys` that can be specified with a list of the weights to delete from the `state_dict`. See the code: 
- [hparams API](https://github.com/mosaicml/composer/blob/moin/streaming_bert/composer/trainer/checkpoint_hparams.py)
- [checkpoint loading implementation](https://github.com/mosaicml/composer/blob/moin/streaming_bert/composer/core/state.py#L295)
- [sample YAML file](https://github.com/mosaicml/research2/blob/moin/runs/data_collection/models/bert_base_rte/bert_base_rte.yaml#L77)

We should decide:
1. Which solution do we prefer?
2. What does the API look like for our solution? If it is # 2, are we happy with the API and implementation that I have come up with? 

Once we have decided, I'm happy to code up the pull request!

[...]

PR #680, opened 2022-03-07
[Checkpointing - PR1] Store the `rank_zero_seed` on state

ravi-mosaicml:
This PR is the first in a series for cleaning up the checkpoint API. One of the prerequisites is storing the seed on the state.
Here, only the rank zero seed is stored on state, since only the rank zero state is persisted in a checkpoint. The trainer uses a distributed reduction to share the seed across states, so the same seed will be restored when resuming from checkpointing, even if a seed was not originally specified.

This PR ignores the `seed` parameter passed into the trainer when resuming from a checkpoint. For the time being, if a new seed is desired, the `seed` attribute must be removed from the checkpoint state dict. #497 will introduce a cleaner API for this (edge) use case.


composer/utils/reproducibility.py
```diff
@@ -86,7 +87,9 @@ def get_random_seed() -> int:
     Returns:
         int: A random seed.
     """
-    seed = int(torch.empty((), dtype=torch.int64).random_(to=2**32).item())
+    rng = random.Random(int(time.time_ns()))  # get a new RNG does not respect the current seed
+    seed = rng.randint(0, 2**32 - 1)
+    assert seed >= 0 and seed < 2**32, "seed should be on this range"
```
ajaysaini725, line 92: Can we make 2**32 a constant called `MAX_ALLOWED_SEED` since it's used in trainer.py as well?

ajaysaini725 (approved):
One small comment otherwise LGTM

composer/trainer/trainer_hparams.py
```diff
@@ -471,6 +471,8 @@ def initialize_object(self) -> Trainer:
         seed = self.seed if self.seed else reproducibility.get_random_seed()
         # need to set seed before model initialization for determinism
         # don't need to set different seeds per process since only the rank 0 initialization is used
+        # algorithms should not use the `seed` on `__init__` but rather on `Event.INIT`, which occurs
```
ravi-mosaicml, line 474:
```suggestion
        # Algorithms should not use the `seed` on `__init__` but rather on `Event.INIT`, which occurs
```

PR #684, opened 2022-03-08
Add mmsegmentation DeepLabv3(+)

Landanjs:
## Motivation
We were using torchvision's DeepLabv3 model for segmentation results, but most papers use DeepLabv3+ as a baseline for semantic segmentation tasks. 

## Implementation
This PR uses the DeepLabv3+ model from [mmsegmentation](https://github.com/open-mmlab/mmsegmentation) which is a commonly used segmentation library (used in at least ConvNext, Swin, and SegFormer). One caveat is that mmsegmentation makes adjustments to the original DeepLabv3(+) model, but maybe this can be justified since ADE20k should be more difficult than the datasets DeepLabv3 was originally proposed for.

Instead of completely removing DeepLabv3, I've added a flag to toggle between the two implementations in mmsegmentation.  

Somewhat unrelated, I added an argument to specify the url to download the backbone weights from to make some future experiments easier.

Side notes: 
- In the future, we should have a more general backbone-head interface when we have more backbones and heads.
- One alternative to mmsegmentation is to use detectron2, but this seems less friendly. Please let me know if this would be a better alternative

## Results

mmsegmentation one-run results: 45 mIoU and 45.47 mIoU for DeepLabv3 and DeepLabv3+, respectively. There are several differences that make it difficult to compare our numbers to theirs:
1. Auxiliary loss
2. Different pre-trained weights
3. Different ResNet stem (v1c)

| Model | mIoU | TTT | 
| ------ | -----  | ------ |
| Old Deeplabv3 Unoptimized | 44.21 +/- 0.30 | 6.8 hours | 
| [DeepLabv3  Unoptimized](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/mmseg_deeplabv3_unoptimized/workspace?workspace=user-landanjs) | 44.82 +/- 0.19 | 7.8 hours | 
| [DeepLabv3+ Unoptimized](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/deeplabv3%2B_unoptimized_w_bn/workspace?workspace=user-landanjs) | 44.69 +/- 0.34 | 7.3 hours | 
| Old DeepLabv3 Optimized | 45.62 +/- 0.16 | 4.9 hours | 
| [DeepLabv3 Optimized](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/mmseg_deeplabv3_optimized/workspace?workspace=user-landanjs) | 45.95 +/- 0.20 | 6 hours | 
| [DeepLabv3+ Optimized](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/deeplabv3%2B_optimized_w_bn/workspace?workspace=user-landanjs) | 45.75 +/- 0.3 | TBD |


composer/models/deeplabv3/deeplabv3.py
```diff
@@ -25,27 +42,53 @@ def deeplabv3_builder(num_classes: int,
...
+                                          c1_in_channels=256,
+                                          c1_channels=48,
+                                          dropout_ratio=0.1,
+                                          num_classes=num_classes,
+                                          norm_cfg=norm_cfg,
+                                          align_corners=False)
+    else:
+        # mmseg config:
+        # https://github.com/open-mmlab/mmsegmentation/blob/master/configs/_base_/models/deeplabv3_r50-d8.py
+        head = ASPPHead(in_channels=2048,
```
A-Jacobson, line 82: is this exactly the same as the torchvision `ASPP` head? or do we need to re-run our baselines.

A-Jacobson:
> Somewhat unrelated, I added an argument to specify the url to download the backbone weights from to make some future experiments easier.

I like the thought but is this necessary right now? does mmseg support something like this or are the urls aliased?
> 
> Side notes:
> 
> * In the future, we should have a more general backbone-head interface when we have more backbones and heads.

Agreed, I'm thinking directories can specify whole model families sort of like our resnets. We should discuss this elsewhere

> * One alternative to mmsegmentation is to use detectron2, but this seems less friendly. Please let me know if this would be a better alternative

all the openmm lab stuff is more actively developed. additionally, it looks like mmseg/mmdet are supersets of detectron with similar api's so if we're going to pick one i think mmseg is a good bet.

Landanjs:
> > Somewhat unrelated, I added an argument to specify the url to download the backbone weights from to make some future experiments easier.
> 
> I like the thought but is this necessary right now? does mmseg support something like this or are the urls aliased?

As of now, we are only using the heads from mmsegmentation, but still using torchvision backbones. The url argument functionality is to specify a specific torchvision url weight i.e. swap between weights from the old and new torchvision training recipes. This isn't necessary, but I oversee running experiments that swap between different pretrained weights.

composer/models/deeplabv3/deeplabv3.py
Landanjs: It isn't exactly the same. I think 2x the channels, dropout, and 1-2 more convolution layers. Rerunning baselines now

Landanjs:
@A-Jacobson Updated numbers in first table based off of recent runs. DeepLabv3 looks really good while DeepLabv3+ may be falling a bit short. One seed reaches the target mIoU the epoch before, not sure if that is sufficient 😬

A-Jacobson:
> @A-Jacobson Updated numbers in first table based off of recent runs. DeepLabv3 looks really good while DeepLabv3+ may be falling a bit short. One seed reaches the target mIoU the epoch before, not sure if that is sufficient 😬

I'm good with this as the baseline wasn't significantly different!

A-Jacobson (approved):

composer/models/deeplabv3/deeplabv3.py
```diff
@@ -29,39 +47,74 @@ def deeplabv3_builder(num_classes: int,
...
+
+    try:
+        from mmseg.models import ASPPHead, DepthwiseSeparableASPPHead  # type: ignore
+    except ImportError as e:
+        raise ImportError(
+            textwrap.dedent("""\
+            Either mmcv or mmsegmentation is not installed. To install mmcv, please run pip install mmcv-full==1.4.4 -f
+             https://download.openmmlab.com/mmcv/dist/{cu_version}/{torch_version}/index.html where {cu_version} and
+             {torch_version} refer to your CUDA and PyTorch versions, respectively. To install mmsegmentation, please
+             run pip install mmsegmentation==0.22.0 on command-line.""")) from e
```
Landanjs, line 90: @ravi-mosaicml just double checking... Does this good to you?

ExtReMLapin:
As you quoted detectron, do you have any quick solution to easily integrate this in detectron and possibly speed up training ?

dblalock:
@Landanjs would be the right person to answer that, but he's on vacation this week. Maybe @florescl can comment?

A-Jacobson:
Hi @ExtReMLapin! Thanks for checking out composer. We don't yet have any convenience wrappers for detectron2 like we do for timm, but you could still train a detectron model using composer methods. I can see two ways to accomplish this right now. 

1. choose methods from our functional interface and drop them directly into your training loop.  You may have to copy and modify the training loop from detectron if that's what you plan to use.

2. create a `ComposerModel` from your detectron model and use our trainer. This may be a bit tricky as detectron models can behave differently during training and inference + they calculated loss internally. Some of the huggingface models and SSD models in `composer.models` may serve as good examples for dealing with this. 

Regarding which methods to use, SAM and SWA are likely to play nicely with detection/instance segmentation. Augmentation based methods and mixup style methods likely wont. Yet.

ExtReMLapin:
Thank you for your answer, i'll give a try as soon as I'm confident enough with all the pytorch/detectron2 ecosystem

PR #686, opened 2022-03-08
Fix DeepSpeed checkpointing

jbloxham:
This turned out to be a nasty race condition where rank 0 was able to release its temp dir, where various rank-agnostic checkpoint files were stored, before other ranks had finished using it. The fix is to add another `dist.barrier()`.

I was able to reproduce this locally, which made testing much easier. I wrapped `test_checkpoint` in a dummy fixture that just ran the whole test hundreds of time (commenting out the resnet test, which was slow), which was able to reliably produce plenty of failures. After this fix, I've seen no failures.


ravi-mosaicml (approved):
:rocket: Sometimes the smallest changes take the longest to write :100: 

Awesome find; after validating locally (i.e. by running the tests 100x), feel free to merge.

jbloxham:
Reconfirmed by running with a `pytest.fixture(foo, range(25))` around `test_checkpoint` (without Resnet) - got no failures.

PR #688, opened 2022-03-08
[S] cleanup tldr; standardize `__all__`

hanlint:
* Remove some unneeded reexports now that `__all__` is defined.
* Standardize on `__all__` including the `hparams` classes
* Shorten some of the module tldr into a consistent style. Leave the naming for `optim` and `models` to the two existing PRs.


composer/functional/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-"""Functional API for applying algorithms.
+"""Functional API for applying algorithims in your own training loop.
```
hanlint, line 3:
```suggestion
"""Functional API for applying algorithms in your own training loop.
```

composer/callbacks/__init__.py
```diff
@@ -1,16 +1,12 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-"""Callbacks provide hooks that can run at each training loop :class:`~composer.core.event.Event`
+"""Callbacks that run at each training loop :class:`~composer.core.event.Event`
```
hanlint, line 3:
```suggestion
"""Callbacks that run at each training loop :class:`~composer.core.event.Event`.
```

composer/callbacks/__init__.py
```diff
@@ -1,16 +1,12 @@
...
 examples for writing your own callbacks at the :class:`~composer.core.callback.Callback` base class.
 """
-from composer.callbacks.callback_hparams import CallbackHparams as CallbackHparams
-from composer.callbacks.callback_hparams import GradMonitorHparams as GradMonitorHparams
-from composer.callbacks.callback_hparams import LRMonitorHparams as LRMonitorHparams
-from composer.callbacks.callback_hparams import MemoryMonitorHparams as MemoryMonitorHparams
-from composer.callbacks.callback_hparams import RunDirectoryUploaderHparams as RunDirectoryUploaderHparams
-from composer.callbacks.callback_hparams import SpeedMonitorHparams as SpeedMonitorHparams
+from composer.callbacks.callback_hparams import (CallbackHparams, GradMonitorHparams, LRMonitorHparams,
+                                                 MemoryMonitorHparams, RunDirectoryUploaderHparams, SpeedMonitorHparams)
```
dskhudia, line 9: Should we use the same pattern at other places ? For example, I was working on composer.datasets documentation and was wondering If I should use this same pattern there.

hanlint: Yes, @ravi-mosaicml said how the `loggers` init is setup is the correct approach.

PR #692, opened 2022-03-08
[Checkpointing - PR3] Clean up RNG and State serialization

ravi-mosaicml:
1. RNG serialization / deserialization is moved from `composer.trainer._checkpoint` to `composer.utils.reproducibility`. This change is needed to refactor the checkpoint saver into a public module. Storing and loading rng states no longer requires a device.
2. Moved helper methods from `composer.trainer._deepspeed` to `composer.core.state` to determine whether the model is deepspeed
3. Added a similar helper for `is_model_ddp` to complement `is_model_deepspeed`.
3. Refactored how the state_dict was serialized and deserialized to support serialization of `@property`s. Stopped storing leading underscores in the checkpoint, as that is a state implementation detail and not something
that should be persisted through the checkpoint.


composer/core/state.py
```diff
@@ -47,6 +49,32 @@ def null(precision):
         return null
 
 
+def _ensure_backwards_compatible_checkpointing(state_dict: types.StateDict):
```
ajaysaini725, line 52: Who is the backwards compatible checkpointing for?

Is there a plan to eventually remove this function?

composer/core/state.py
```diff
@@ -315,36 +334,33 @@ def load_state_dict(self, state: types.StateDict, strict: bool = False):
...
-                        source = serialized_value[target.__class__.__qualname__]
-                        target.load_state_dict(source)
-                else:
-                    # direct serialization
-                    setattr(self, state_field_name, serialized_value)
+                # direct serialization
+                try:
+                    setattr(self, attribute_name, serialized_value)
+                except AttributeError:
+                    # ignore AttributeError for properties that have getters but not setters.
```
ajaysaini725, line 362: Log a warning here if it's not something that should happen often?

ajaysaini725 (approved):
Couple small questions otherwise LGTM

composer/core/state.py
ravi-mosaicml: This is by design and will happen often -- `is_module_ddp` has no setter, and we'll hit this on every checkpoint load. Considering that the setattr is on `self` (and we don't define `__slots__`), this would only happen on properties that don't have a setter. Since the `attribute_name` comes from `self.serialized_attributes` (and is not something that the user sets, or is defined in the checkpoint), I think this should be safe?

composer/core/state.py
ravi-mosaicml: I figured we'll need this as we tweak the checkpoint format. We may remove parts of it as checkpoints become more and more outdated as composer updates.

composer/core/state.py
ajaysaini725: Ah okay sg

PR #693, opened 2022-03-08
[Checkpointing - PR4] Refactored the `CheckpointLoader` into a `load_checkpoint` function

ravi-mosaicml:
Since the checkpoint loading happens in `Trainer.__init__` (except for the restoration of the rng state), there is no need for a checkpoint loader class.

1. The CheckpointLoader class is replaced with a function `load_checkpoint`, and all private members are converted into private, module-level helper functions.
2. The file downloading portions of the checkpoint loader were refactored into their own, standalone utility at `composer.utils.file_retriever`, with their own test cases.


ajaysaini725 (approved):
LGTM

PR #696, opened 2022-03-09
Docs 0.4.1: Profiler section and tutorials

### LINKED ISSUE #682: [Profiler] Remove instant events
## 🚀 Feature Request
Remove instant events from the trace

## Motivation

They do not add much value and are a distraction for viewing the trace.

bandish-shah:
* Adds trainer/performance landing page
* Adds "Performance Profiling" and "Analyzing Traces" tutorials
* Tweak `profiler_demo.py` to be consistent with tutorials
* Resolves #682


docs/source/trainer/performance_tutorials/analyzing_traces.md
```diff
@@ -0,0 +1,56 @@
+# 🥽 Analyzing Traces
```
dskhudia, line 1: Images in this doc look small on the web page, however, they are fine if opened in a new tab/or downloaded. Let us add a note mentioning this.

docs/source/trainer/performance_tutorials/profiling.md
```diff
@@ -0,0 +1,184 @@
...
+
+Upon completion of the profiling run, the `torch_profiler_trace_dir` will contain the unadulterated Torch Profiler traces while the `profiler_trace_file` will merge the relevant Torch Profiling trace data with the default Profiling trace data.
+
+```{note}
+The location of the `profiler_trace_file` and the `torch_profiler_trace_dir` will be in the specified `$COMPOSER_RUN_DIRECTORY` (see {mod}`~composer.utils.run_directory`).
+```
+
+### Specifying a profiling window
+
+When setting up profiling it is important to specify a *profiling window* which specifies the profiler’s recording behavior per batch within a given epoch.  The profiling window can be configured using the `prof_skip_first`, `prof_wait`, `prof_warmup`, `prof_active` and `prof_repeat` arguments which correspond to the `skip_first`, `wait`, `warmup`, `active` and `repeat` states, respectively.  The profiler behaves as follows in each state:
```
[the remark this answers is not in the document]
bandish-shah: The Composer Profiler should support `prof_skip_first` and `prof_wait` so I don't think a link to the Torch profiler doc is necessary here.  The only noop is `prof_warmup` which we note in detail in the warning below.  

If this is not the case we should file an issue and fix it in a future release.

docs/source/trainer/performance_tutorials/analyzing_traces.md
bandish-shah: I'll play with the `scale` attribute in the Sphinx `{image}` directive to see if helps here.  Otherwise we may need to take a new screenshot of these so the text is easier to read.

bandish-shah: Made it so the images can be clicked to enlarge, let me know what you think.

dskhudia: Thanks. Much better now. Could you do the same for the one image in Profiling doc?

dskhudia (approved):

PR #697, opened 2022-03-09
Fix DatasetHparams, WebDatasetHparams docstring

growlix:
`def initialize_object()` has an improperly indented docstring line that causes rendering problems


ravi-mosaicml (approved):
:+1:

hanlint (approved):

PR #699, opened 2022-03-09
[Artifact Logging PR2] Logger Destination Cleanup

ravi-mosaicml:
Cleaned up the `LoggerDestination` base class in preparation for [Artifact Logging](https://www.notion.so/Artifact-Logging-API-d3754be3ec874233910cb0c4f23c9d32)

1. Removed `will_log`. Having this helper added additional complexity, especially since nothing used deferred logging. Instead, individual logger destinations can implement filtering as they see fit. This greatly cleans up the base logger class.
1. Removed deep copying of the logging data in the base logger class. It turned out that only the InMemoryLogger needed this functionality, and it now deepcopies the data itself. While this can introduce a performance regression if you are using multiple loggers that require a deepcopy, this scenario is very unlikely, since it would require the user to write their own logger callback AND use the InMemoryLogger (the WandB logger already does its own copying). Decided that it was worth a cleaner base logger class over this (rare) performance gain.
1. Changed the signature of `logger_destination.data` to include the `state` instead of the timestamp. The state was included in the (now removed) `will_log` signature. The state gives flexibility to the logger to decide whether to log based on other information in the state (say, the last evaluation metrics).


jbloxham (approved):
LGTM 👀

PR #700, opened 2022-03-09
[Artifact Logging PR3] Add the `run_name` as a property of the `Logger`

ravi-mosaicml:
Added a `run_name` as a property of the Logger, and using a coolname to generate it if it is not specified. Moved this logic out of the wandb logger, since it can be useful for every logger. See [Artifact Logging](https://www.notion.so/Artifact-Logging-API-d3754be3ec874233910cb0c4f23c9d32) for the spec.

CC @abhi-mosaic


ajaysaini725 (approved):
LGTM

composer/core/logging/logger.py
```diff
@@ -67,8 +90,18 @@ def __init__(
             self,
             state: State,
             destinations: Sequence[LoggerDestination] = tuple(),
+            run_name: Optional[str] = None,
     ):
         self.destinations = destinations
+        if run_name is None:
+            # prefixing with the time so experiments sorted alphabetically will
+            # have the latest experiment last
+            run_name = str(time.time_ns()) + "-" + coolname.generate_slug(2)
```
jbloxham, line 99: `time_ns` is definitely more resolution than we need. The point of the `coolname`s is to make these names more readable, but that gets drowned out by an excessively long number up front.

So, two suggestions to make this shorter:
1) Use seconds as resolution. Conflicts between runs should be rare, in which case the coolnames can disambiguate.
2) Encode the time in base 64. As you said, the time is only there for lexicographical sorting, anyways.

composer/trainer/trainer.py
```diff
@@ -224,6 +224,7 @@ class Trainer:
                 :func:`.configure_deterministic_mode` function at the start of your script. This will ensure any initialization done before the trainer init also runs deterministically.
 
             .. seealso:: :mod:`composer.utils.reproducibility` for more details on reproducibility.
+        run_name (str, optional): A name for this training run. If not specified, TODO.
```
jbloxham, line 227: Fix this TODO.

jbloxham (approved):
Big fan of this change :)

composer/core/logging/logger.py
ravi-mosaicml: Ah, you caught my inner 6.172, where code runs so fast one must use nanoseconds haha.

I went with the first suggestion; using `int(time.time())`.

PR #702, opened 2022-03-09
Match ADE20k crop to mmsegmentation baseline

Landanjs:
This was one thing I neglected when setting up the ADE20k baseline. This allows the segmentation crop transform to attempt multiple crops if the class with the max area percentage is larger than some threshold. I added some functionality to maintain the crop with the lowest max area percentage if the threshold is not reached within the specified number of retries.

I need to test the "best crop" functionality, but results for the original setup are as follows:

| Model | mIoU | TTT | 
| ------ | -----  | ------ |
| [Old crop Unoptimized Deeplabv3](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/mmseg_deeplabv3_unoptimized/workspace?workspace=user-landanjs) | 44.82 +/- 0.19 | 7.8 hours | 
| [New crop Unoptimized Deeplabv3](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/mmseg_deeplabv3_unoptimized_new_crop/workspace) | 44.73 +/- 0.2 | 6.8 hours | 
| [Old crop Unoptimized DeepLabv3+](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/deeplabv3%2B_unoptimized_w_bn/workspace?workspace=user-landanjs) | 44.69 +/- 0.34 | 7.3 hours |
| [New crop Unoptimized Deeplabv3+](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/groups/mmseg_deeplabv3%2B_unoptimized_new_crop/workspace) | 45.18 +/- 0.18 | 6.8 hours | 

Interpretations:
- Seems to help DeepLabv3+ quiet a bit
- Does not effect throughput, but the maximum dataloader throughput is likely smaller (would need to use no-op model to measure)
- Negatively impacts optimized baselines, but these should be retuned anyways

### DIFF
--- a/composer/datasets/ade20k.py
+++ b/composer/datasets/ade20k.py
@@ -62,24 +62,44 @@ class RandomCropPair(torch.nn.Module):
 
     def forward(self, sample: Tuple[Image.Image, Image.Image]):
         image, target = sample
-        if image.height > self.crop_size[0] or image.width > self.crop_size[1]:
-            i, j, h, w = transforms.RandomCrop.get_params(
-                image, output_size=self.crop_size)  # type: ignore - transform typing excludes PIL.Image
-            target_crop = TF.crop(target, i, j, h, w)  # type: ignore - transform typing excludes PIL.Image
-            if self.class_max_percent < 1.0:
-                for _ in range(self.n_retry):
-                    labels, counts = np.unique(np.array(target_crop), return_counts=True)
-                    counts = counts[labels != 0]
-                    if len(counts) > 1 and (np.max(counts) / np.sum(counts)) < self.class_max_percent:
-                        break
-                    i, j, h, w = transforms.RandomCrop.get_params(
-                        image, output_size=self.crop_size)  # type: ignore - transform type excludes PIL.Image
-                    target_crop = TF.crop(target, i, j, h, w)  # type: ignore - transform typing excludes PIL.Image
 
-            image = TF.crop(image, i, j, h, w)  # type: ignore - transform typing does not include PIL.Image
+        # if image size is smaller than crop size, no cropping necessary
+        if image.height <= self.crop_size[0] and image.width <= self.crop_size[1]:
+            return image, target
+
+        # generate crop
+        crop = transforms.RandomCrop.get_params(
+            image, output_size=self.crop_size)  # type: ignore - transform typing excludes PIL.Image
+
+        if self.class_max_percent < 1.0:
+            best_max_percent = 1.0
+            best_crop = crop
+            for _ in range(self.n_retry):
+                # Crop target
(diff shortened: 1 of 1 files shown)


composer/datasets/ade20k.py
```diff
@@ -54,17 +54,53 @@ class RandomCropPair(torch.nn.Module):
         crop_size (Tuple[int, int]): the size (height x width) of the crop.
     """
 
-    def __init__(self, crop_size: Tuple[int, int]):
+    def __init__(self, crop_size: Tuple[int, int], class_max_percent: float = 1.0, n_retry: int = 0):
```
A-Jacobson, line 57: add these new args to docstring.

A-Jacobson (commented):
- Could use a docstring.
- Are you sure this should be added in spite of the v3 baseline regression? Is v3+ becoming the new baseline?

composer/datasets/ade20k.py
Landanjs: Oops, should have done that before review request :(

Landanjs:
Yes, for a few papers (SegFormer, Swin, MaskFormer), DeepLabv3+ was the only baseline. So I think it would be good to follow that standard unless we have a reason not to. (Even though I don't like DeepLabv3+ as much...)

In any case, I think the 0.09 mIoU difference from new/old cropping on DeepLabv3 may be reasonable / within noise, but happy to discuss!

A-Jacobson (approved):
approving, conditional on the docstring =)

Landanjs:
I added docstrings and removed the "best crop" functionality since it didn't seem to help.

composer/datasets/ade20k.py
```diff
@@ -314,7 +343,7 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
                     RandomResizePair(min_scale=self.min_resize_scale,
                                      max_scale=self.max_resize_scale,
                                      base_size=(self.base_size, self.base_size)),
-                    RandomCropPair(crop_size=(self.final_size, self.final_size)),
+                    RandomCropPair(crop_size=(self.final_size, self.final_size), class_max_percent=0.75, n_retry=10),
```
hanlint, line 346: these should be mult-line. Pro tip: if you add a common after the last argument (e.g. `RandomCropPair(..., n_retry=10,)`, yapf will autoformat that for you with `make style`.

Also, we are moving to `num` instead of `n` prefixes, so can you make this `num_retry`?

hanlint (commented):
one nit to fix, then good to merge.

composer/datasets/ade20k.py
Landanjs: Done and Done!

hanlint (approved):

Issue #347, opened 2022-02-02
A Better Progress Bar


hanlint:
cc: @siriuslee @A-Jacobson @Landanjs

ravi-mosaicml:
Can you elaborate?

A-Jacobson:
I'll do my best.

TLDR: the progress bar should at least tell us total training time and model metrics at a glance.

our current progress bar looks like this:
<img width="658" alt="Screen Shot 2022-02-11 at 1 36 56 AM" src="https://user-images.githubusercontent.com/9411532/153568483-7403c4be-fd78-4b56-bb86-13f4fb10b10f.png">

There are a few things lacking with this setup:
1. It could use an overall progress bar so the user can get an estimate of total training time without doing some mental math.

2. It doesn't log enough metrics, I had a hard time logging anything except validation accuracy. Which isn't really enough to tell if your model is improving.

the fastai progress bar is sort of the extreme example of this, they render tables of training metrics as they train.
<img width="484" alt="Screen Shot 2022-02-11 at 1 51 13 AM" src="https://user-images.githubusercontent.com/9411532/153570538-6998d223-0fc4-4e5a-b09f-d8ad8d6caacf.png">

Their library can even render plots during training though i dont know if we need to go that far.
https://github.com/fastai/fastprogress

Landanjs:
Austin provided great points for improvement! 

For me specifically, displaying the learning rate would be helpful to debug if it is increasing/decreasing at the expected times. Likely, it would be helpful to visualize most callback monitored metrics on the progress bar. Providing visualizations/table are bonuses for improving debugging

mvpatel2000:
Closing. Tracking elsewhere as low pri

PR #708, opened 2022-03-10
[Artifact Logging PR6] Rename the `TQDMLogger` as the `ProgressBarLogger`; remove terminal logging from the file logger

ravi-mosaicml:
* Rename the `TQDMLogger` to the `ProgressBarLogger`: TQDM is an implementation detail, and thus should not be in the class name. Instead, `ProgressBarLogger` describes exactly where data is being logged.
* Removed listing the `ProgressBarLogger` from config files. If no logger is specified, the `ProgressBarLogger` will be used by default.
* Removed logging to the terminal from the file logger. As the file logger will support capturing stdout and stderr calls (in #710), it would be recursive to also allow the file logger to print to stdout. This PR limits the file logger to logging to an actual file. Issue #347 will improve the `ProgressBarLogger` so that metrics can be printed via the terminal.


hanlint:
I would vote for keeping TQDMLogger, as it's a well known library by data scientists. Alternatively, ProgressBarLogger?

composer/trainer/trainer_hparams.py
```diff
@@ -101,7 +101,7 @@
 logger_registry = {
     "file": FileLoggerHparams,
     "wandb": WandBLoggerHparams,
-    "tqdm": TQDMLoggerHparams,
+    "terminal": ProgressBarLoggerHparams,
```
jbloxham, line 104:
```suggestion
    "progress_bar": ProgressBarLoggerHparams,
```

jbloxham (approved):
Looks good to me! `ProgressBarLogger` is immediately clear in what it does, since not _everyone_ knows what TQDM is.

PR #710, opened 2022-03-10
[Artifact Logging PR7] Add `stdout` and `stderr` capture to the FileLogger

ravi-mosaicml:
This PR adds options to capture stdout and stderr writes, and print them in the file saved by the FileLogger.


composer/loggers/file_logger.py
```diff
@@ -61,9 +63,11 @@ class FileLogger(LoggerDestination):
...
 
     Args:
-        filename (str): File to log to.
-            Filepaths should be specified relative to the
-            :mod:`~.composer.utils.run_directory`.
+        filename (str): Filepath to log to, relative to the :mod:`~.composer.utils.run_directory`.
+        capture_stdout (bool, optional): If ``True`` (the default), writes to ``stdout`` will be included in
+            ``filename``.
+        capture_stderr (bool, optional): If ``True`` (the default), writes to ``stderr`` will be included in
+            ``filename``.
```
jbloxham, line 70: Definitely emphasize that the default for these args is `True`, not `False` as is often assumed for optional boolean parameters.

composer/loggers/file_logger.py
```diff
@@ -131,24 +155,25 @@ def _will_log(self, log_level: LogLevel) -> bool:
...
             return
-        data_str = format_log_data_value(data)
-        if self.file is None:
-            raise RuntimeError("Attempted to log before self.init() or after self.close()")
-        print(f"[{log_level.name}][batch={int(state.timer.batch)}]: {data_str}", file=self.file, flush=False)
+        data_str = format_log_data_value(data) + "\n"
+        self.write(
+            f'[{log_level.name}][batch={int(state.timer.batch)}]: ',
+            data_str,
+        )
```
jbloxham, line 162: This might be the nittiest thing I've ever commented on, but to me, it's weird that the `"\n"` is being appended to the `data_str` variable, since a newline has nothing to do with the string representation of some data. Save the append for when it is used.
```suggestion
        data_str = format_log_data_value(data)
        self.write(
            f'[{log_level.name}][batch={int(state.timer.batch)}]: ',
            data_str + '\n',
        )
```
Also feel free to tell me to get a life.

jbloxham (approved):
Looks great! This seems like a pretty useful thing to have.

composer/loggers/file_logger.py
ravi-mosaicml: Added the standard (default: ``True``) to the end.

composer/loggers/file_logger.py
ravi-mosaicml: LMAO sure

PR #720, opened 2022-03-11
Add Algorithm Warning for `NoEffectWarning`

hanlint:
Based on offline discussion, we are introducing a few warnings to help algorithms alert users of departures from nominal behavior:

* `NoEffectWarning`: should be emitted when an algorithm is selected but its application leads to no change in state (e.g. blurpool on a fully MLP model)
* `NotIntendedUseWarning`: generic catch-all for algorithms being applied outside of its intended use. e.g. progressive resizing that reduces image down to 1x1 pixels at the start of training.

As this evolves, we can add additional warnings as needed.

This PR only implements the `NoEffectWarning` for `BlurPool`. We will need to make a separate pass to update each algorithm to emit the proper warnings.


A-Jacobson (approved):
LGTM, just like the proposal. Will be interesting to see this evolve to some of the other algorithms.

coryMosaicML (approved):
LGTM! I really like having these.

PR #721, opened 2022-03-11
Polish intro docs

dblalock:
Made a copy editing pass on the prose in:

- Index (main page)
- Installation
- Welcome Tour
- Using the Trainer

Also removed a type on `trainer/algorithms.rst` and added a few links as appropriate. Most noticeable change is that I rewrote the first sentence at the top of index.rst to more clearly articulate our value prop and what composer does.

Main lingering concern is that our algorithms API page and functional API page are pretty redundant. And I'm still not sure about the heading for these two pages; it was "Algorithms" along with the first page also being called "Algorithms", and I made the heading be "Adding Speedup Methods" instead so that it isn't literally the same text. I think there's a nice solution here if we refer to the two approaches as the as the "[Two-Way] Callbacks API" and the "Functional API" (or maybe even "Imperative API"). But this is all 1am thoughts for a separate PR.


docs/source/index.rst
```diff
@@ -3,36 +3,37 @@
...
+Composer is a library for training neural networks better, faster, and cheaper. It contains many state-of-the-art methods for accelerating neural network training and improving generalization, along with an optional :doc:`Trainer<trainer/using_the_trainer>` API that makes `composing` many different enhancements easy.
+
+We implemented dozens of papers and `trained thousands of models <https://app.mosaicml.com/explorer/imagenet>`_ so you don't have to.
 
 Using Composer, you can:
 
-- Train an ImageNet model to 76.1% accuracy for $37 (`with vanilla PyTorch`:$127)
-- Train a GPT-2 125M to a perplexity of 23.9 for $148 (`with vanilla PyTorch`: $255)
-- Use start-of-the-art implementations of methods to speed up your own training loop.
+- Train an ImageNet model to `76.1% accuracy for $38 <https://app.mosaicml.com/explorer/imagenet>`_ (`with vanilla PyTorch` :$127)
```
hanlint, line 12:
```suggestion
- Train a ResNet-50 ImageNet model to `76.1% accuracy for $38 <https://app.mosaicml.com/explorer/imagenet>`_ (`with vanilla PyTorch` :$127)
```

docs/source/index.rst
```diff
@@ -3,36 +3,37 @@
...
+- Train a GPT-2 125M to a perplexity of `23.9 for $149 <https://app.mosaicml.com/explorer/openwebtext>`_ (`with vanilla PyTorch`: $255)
```
hanlint, line 13:
```suggestion
- Train a GPT-2 125M on openwebtext to a perplexity of `23.9 for $149 <https://app.mosaicml.com/explorer/openwebtext>`_ (`with vanilla PyTorch`: $255)
```

hanlint (approved):
looks good, thanks!

PR #722, opened 2022-03-11
Update `C4Dataset` to repeat, handle `max_samples` safely

abhi-mosaic:
This PR modifies the C4 streaming dataset so it will repeat over its underlying data source, until reaching the epoch boundary of `max_samples` provided by the user.

Previously, if `max_samples` were too big, there was a potential bug where the dataset iterator would stop early and our Trainer's time conversions would be incorrect. Also, DDP might not be synced.

To make the repeating HF dataset, I followed the structure of this code: https://github.com/huggingface/datasets/blob/3dfdb054e3636a5784d365bcd65759e5bd697b03/src/datasets/iterable_dataset.py#L394

## Testing
I could use some guidance here... I have manually tested by creating a `C4Dataset` with a `max_samples` value that I know for sure is larger than the underlying dataset, and have seen that it works. But it takes ~5min to reach that point, and I'm not sure if we want a long test like that in our suite.

## Next steps

Two follow-ups from this PR:

1) Upstream the `RepeatExamplesIterable` to HF datasets, because it would be convenient if these objects all had a `.repeat()` operation.

2) Make Composer compatible with infinite datasets with no `len()`, this would remove the need for users to provide `max_samples` for datasets like `C4Dataset`.

### DIFF
--- a/composer/datasets/c4.py
+++ b/composer/datasets/c4.py
@@ -281,6 +281,8 @@ class C4Dataset(IterableDataset):
     def __len__(self):
         return self.max_samples_per_device
 
+    # Repeat a HF iterable dataset infinitely
+    # TODO: This functionality should eventually be upstreamed to HF as `hf_iterable_dataset.repeat()`
     def _repeat(self, dataset):
         try:
             from datasets.iterable_dataset import _BaseExamplesIterable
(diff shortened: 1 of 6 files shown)


hanlint (approved):
Tests look good, verified offline, thanks!

PR #723, opened 2022-03-11
Create codeql-analysis.yml

Averylamp:


ravi-mosaicml (approved):
Lgtm, though I'm surprised it found only one issue.

PR #724, opened 2022-03-11
Docker file for vision applications with ffcv and deeplabv3 dependencies 

dskhudia:
Build:
`cd docker; TAG='mosaicml/pytorch:1.9.1_cu111-python3.7-ubuntu20.04' BASE_IMAGE='nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04' PYTHON_VERSION='3.7' CUDA_VERSION_TAG='cu111' CUDA_VERSION='11.1.1' LINUX_DISTRO='ubuntu:20.04' PYTORCH_VERSION='1.9.1' TORCHVISION_VERSION='0.10.1' make vision`


docker/vision/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    sudo apt-get autoclean && \
+    sudo apt-get clean && \
+    sudo rm -rf /var/lib/apt/lists/*
+
+RUN sudo pip3 install ffcv opencv-python numba 
+
+RUN sudo pip3 install mmcv-full==1.4.4
+
+RUN sudo pip3 install mmsegmentation
+
```
ravi-mosaicml, line 30: nit: extra newlines

ravi-mosaicml (commented):
Thanks for doing this!

However, instead of creating a new Dockerfile, would it be possible to use a [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) on the existing `docker/pytorch/Dockerfile`? The reason why I ask is it would simplify the CI/CD complexity. With multistage docker builds, Jenkins would not need to track whether the base image has changed when the vision image has also changed; Docker would handle that for us.

We will need a build matrix for the vision image. If going with multistage builds, can you add line(s) to `docker/pytorch/build_matrix.sh` so it prints out a line for each vision image? 
Each line should consist of key-value pairs representing build arguments, with the special arguments of `TAG=...` representing the destination where this image will be pushed and `TARGET=...` representing the target build stage to stop at.

- [ ]  I will need to modify the `getDockerBuildMatrix` Jenkins function to support `TARGET`s. Without this, builds will appear to succeed, but the normal pytorch build will include all the vision dependencies.

In addition, do we want another dockerhub repo for the vision image (e.g. `mosaicml/pytorch_vision`), or should we include `vision` in the image tag name (e.g. `mosaicml/pytorch:1.10.0_cu113-python3.9-ubuntu20.04-vision`)? I would be leaning towards the latter, as it simplifies the administration of the dockerhub, but we can ask Nick to create a new dockerhub repo if we want the former.)

(Side note 1: I would like to build the `composer` image using multi-stage builds as well, but that's an issue for another PR).

(Side note 2: We will need to reconfigure our pytest markers to support something like `@pytest.mark.vision`, and then create a new Jenkinsfile entry to run only the vision tests in the vision image.)

docker/vision/Dockerfile
dskhudia: @ravi-mosaicml : mosaicml/pytorch_vision has already been created.

ravi-mosaicml:
I modified the docker build process to support `TARGET`s. I didn't test it, but let's inspect the logs of this PR to ensure that feature is working properly.

docker/vision/Dockerfile
dskhudia: @ravi-mosaicml : added vision image as stage. Ideally I would have liked to install vision stuff before setting the username in Pytorch image but I don't know a way to share code between stages without duplicating some of the code.

Landanjs:
Alternatively, we can install the precompiled mmcv packages using `pip install mmcv-full==1.4.4 -f https://download.openmmlab.com/mmcv/dist/cu111/torch1.9.0/index.html`, but this would need to be changed as we change the cuda and torch versions. If that sounds reasonable, we can install the precompiled packages

dskhudia:
> Alternatively, we can install the precompiled mmcv packages using `pip install mmcv-full==1.4.4 -f https://download.openmmlab.com/mmcv/dist/cu111/torch1.9.0/index.html`, but this would need to be changed as we change the cuda and torch versions. If that sounds reasonable, we can install the precompiled packages

I made it precompiled packages based. There was a complication though. Their [installation page](https://mmcv.readthedocs.io/en/latest/get_started/installation.html) says  `pip install mmcv-full -f https://download.openmmlab.com/mmcv/dist/{cu_version}/{torch_version}/index.html` should work but it doesn't for torch1.9.1 but does for torch1.9.0 so had to manipulate the minor version.

ravi-mosaicml:
S

> @ravi-mosaicml Could you increase the timeout of 40 mins for the Jenkins build. This is exceeding the timeout. Building mmcv-full from sources is very time consuming. cc: @Landanjs

To increase the timeout, set the `POD_TIMEOUT` and/or `P_TIMEOUT` parameters in the `.ci/Jenkinsfile` under the appropriate job.

P_TIMEOUT: How long the job can run, once the command has been started (excluding scheduling delays)
POD_TIMEOUT: How long the job has to finish once it has been added to the queue (including scheduling delays)

So, `POD_TIMEOUT` should be larger than `P_TIMEOUT`

dskhudia:
> > @ravi-mosaicml Could you increase the timeout of 40 mins for the Jenkins build. This is exceeding the timeout. Building mmcv-full from sources is very time consuming. cc: @Landanjs
> 
> To increase the timeout, set the `POD_TIMEOUT` and/or `P_TIMEOUT` parameters in the `.ci/Jenkinsfile` under the appropriate job.
> 
> P_TIMEOUT: How long the job can run, once the command has been started (excluding scheduling delays) POD_TIMEOUT: How long the job has to finish once it has been added to the queue (including scheduling delays). Default is 3600 (1 hour).
> 
> So, `POD_TIMEOUT` should be larger than `P_TIMEOUT`

Actually increasing the timeout is no longer necessary. We are installing from pre-built packages instead of building again.

docker/pytorch/build_matrix.sh
```diff
@@ -3,11 +3,12 @@ set -euo pipefail
...
+echo "TAG='mosaicml/pytorch:1.10.0_cpu-python3.9-ubuntu20.04' BASE_IMAGE='ubuntu:20.04' PYTHON_VERSION='3.9' CUDA_VERSION_TAG='cpu' CUDA_VERSION='cpu' LINUX_DISTRO='ubuntu:20.04' PYTORCH_VERSION='1.10.0' TORCHVISION_VERSION='0.11.1' TARGET='pytorch_stage'"
+echo "TAG='mosaicml/pytorch:1.9.1_cu111-python3.8-ubuntu18.04' BASE_IMAGE='nvidia/cuda:11.1.1-cudnn8-devel-ubuntu18.04' PYTHON_VERSION='3.8' CUDA_VERSION_TAG='cu111' CUDA_VERSION='11.1.1' LINUX_DISTRO='ubuntu:18.04' PYTORCH_VERSION='1.9.1' TORCHVISION_VERSION='0.10.1' TARGET='pytorch_stage'"
+echo "TAG='mosaicml/pytorch:1.9.1_cpu-python3.8-ubuntu18.04' BASE_IMAGE='ubuntu:18.04' PYTHON_VERSION='3.8' CUDA_VERSION_TAG='cpu' CUDA_VERSION='cpu' LINUX_DISTRO='ubuntu:18.04' PYTORCH_VERSION='1.9.1' TORCHVISION_VERSION='0.10.1' TARGET='pytorch_stage'"
+echo "TAG='mosaicml/pytorch_vision:1.9.1_cu111-python3.7-ubuntu20.04' BASE_IMAGE='nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04' PYTHON_VERSION='3.7' CUDA_VERSION_TAG='cu111' CUDA_VERSION='11.1.1' LINUX_DISTRO='ubuntu:20.04' PYTORCH_VERSION='1.9.1' TORCHVISION_VERSION='0.10.1' TARGET='vision_stage'"
```
ravi-mosaicml, line 14: Can you also tag this image as `mosaicml/pytorch_vision:latest`? Multiple `TAG='...'` are supported (see line 10 as an example)

docker/pytorch/Dockerfile
```diff
@@ -149,3 +149,34 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
 WORKDIR /home/mosaicml
 
 USER mosaicml
+
+
+FROM pytorch_stage AS vision_stage
+ARG DEBIAN_FRONTEND=noninteractive
+
+ARG PYTORCH_VERSION
```
ravi-mosaicml, line 157: Nit: Do args used in a previous stage need to be declared again? (Not sure, but if not, can we remove?)

docker/pytorch/Dockerfile
```diff
@@ -149,3 +149,34 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
...
+    libopencv-dev \
+    # For deeplabv3:
+    ffmpeg \
+    libsm6 \
+    libxext6 && \
+    sudo apt-get autoclean && \
+    sudo apt-get clean && \
+    sudo rm -rf /var/lib/apt/lists/*
+
+RUN sudo pip3 install ffcv opencv-python numba
```
ravi-mosaicml, line 178:
```suggestion
RUN sudo pip3 install --no-cache-dir ffcv opencv-python numba 
```

docker/pytorch/Dockerfile
```diff
@@ -149,3 +149,34 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
...
+    ffmpeg \
+    libsm6 \
+    libxext6 && \
+    sudo apt-get autoclean && \
+    sudo apt-get clean && \
+    sudo rm -rf /var/lib/apt/lists/*
+
+RUN sudo pip3 install ffcv opencv-python numba 
+
+RUN sudo pip3 install mmcv-full==1.4.4 -f https://download.openmmlab.com/mmcv/dist/${CUDA_VERSION_TAG}/torch`echo ${PYTORCH_VERSION} | sed -E "s/([[:digit:]]).([[:digit:]]).([[:digit:]])/\1.\2.0/g"`/index.html
```
ravi-mosaicml, line 180:
```suggestion
RUN sudo pip3 install --no-cache-dir mmcv-full==1.4.4 -f https://download.openmmlab.com/mmcv/dist/${CUDA_VERSION_TAG}/torch`echo ${PYTORCH_VERSION} | sed -E "s/([[:digit:]]).([[:digit:]]).([[:digit:]])/\1.\2.0/g"`/index.html
```

ravi-mosaicml, line 180: Should we define another build arg called `MMCV_PYTORCH_VERSION`? (I appreciate this `sed`ing, but it is not very readable.) A comment here would also be helpful to explain why we need to replace version digits.

ravi-mosaicml (approved):
LGTM! See comments. Glad that the targets are working!

docker/pytorch/Dockerfile
dskhudia: Added MMCV_TORCH_VERSION just for pytorch_vision images.

PR #727, opened 2022-03-12
Remove Batch Types - Part 1: recursive to_device function

A-Jacobson:
As discussed offline, this PR includes:

- a recursive `to_device` function that can work on arbitrary batch types. 
- examples of common batch types for testing.
- a very basic test to make sure `to_device` runs without error on each known batch type

Seeking input on:
- better tests. on cpu, i can check that the output is the same as the input (assuming we started on the cpu). on gpu, i can directly check that the device of each tensor is `.cuda`. Alternatively, I could override `batch_to_tensor` and have it convert all tensors to some arbitrary type such as `bool`, then check that they're converted. Thoughts?


composer/utils/iter_helpers.py
```diff
@@ -8,6 +8,21 @@
 import contextlib
 
 
+def recursive_apply(func: callable, data):
```
hanlint, line 11: is there a way to merge this and/or use `map_collection` below? Syntactically they look similar. If not, at minimum they should match signature -- e.g. data, func to match the collection, map_fn below.

composer/trainer/devices/device.py
```diff
@@ -53,22 +53,21 @@ def tensor_to_device(self, tensor: Tensor) -> Tensor:
         """
         pass
 
-    def batch_to_device(self, batch: Batch) -> Batch:
+    def batch_to_device(self, batch: Any):
```
hanlint, line 56: You'll probably want to use `TypeVar` here -- we can help with that if needed.

tests/test_device.py
```diff
@@ -0,0 +1,57 @@
...
+def dummy_dict_batch() -> Dict[str, torch.Tensor]:
+    image = torch.randn(size=(12, 3, 32, 32))
+    target = torch.randint(size=(12,), high=10)
+    return {'image': image, 'target': target}
+
+
+def dummy_maskrcnn_batch() -> List[Tuple[torch.Tensor, Dict[str, torch.Tensor]]]:
+    def generate_maskrcnn_sample(num_detections, image_height=12, image_width=12, num_classes=80):
+        """Generates a maskrcnn style sample (Tensor, Dict[Tensor])."""
+        image = torch.randn(size=(3, image_height, image_width)).type(torch.float)
```
hanlint, line 27: nit: use multiline here so more readable, e.g. 
```
target = {
    'boxes': torch.randint(
        size=...,
        loaw=...,
        
```

tests/test_device.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@pytest.mark.parametrize("batch", [dummy_tensor_batch(),
+                                   dummy_tuple_batch(),
+                                   dummy_dict_batch(),
+                                   dummy_maskrcnn_batch()])
+def test_to_device_cpu(batch):
+    device = DeviceCPU()
+    new_batch = device.batch_to_device(batch)
+    # test something about new batch?
```
hanlint, line 46: maybe do something like:
```
map_collection(new_batch, lambda t: assert t.device == 'cpu')
```

to verify that all the batches were loaded.

tests/test_device.py
```diff
@@ -0,0 +1,57 @@
...
+        return [generate_maskrcnn_sample(n) for n in torch.randint(size=(batch_size,), low=1, high=max_detections)]
+
+    return generate_maskrcnn_batch(batch_size=5, max_detections=5)
+
+
+@pytest.mark.parametrize("batch", [dummy_tensor_batch(),
+                                   dummy_tuple_batch(),
+                                   dummy_dict_batch(),
+                                   dummy_maskrcnn_batch()])
+def test_to_device_cpu(batch):
```
hanlint, line 43: These can be combined with `@device`:
```

@device('gpu', 'cpu')
@pytest.mark.parametrize("batch", ...)
test_to_device(device, batch):
   new_batch = device.batch_to_device(batch):
   ...

```

hanlint (commented):
Looks good -- some comments and a suggestion on testing

composer/utils/iter_helpers.py
A-Jacobson: Map_collection isn't recursive, it's pretty a switch statement between dict, lists, and single items. 

I could rename recursive apply to map_collection possibly? The functionality should be a superset of the existing function.

tests/test_device.py
A-Jacobson: Tensors are cpu by default so I don't think that will do anything.

hanlint: Ah yes, this was a suggestion more for the GPU tests.

composer/utils/iter_helpers.py
ravi-mosaicml: The function signature will also need type annotations in `iter_helpers.pyi`.

ravi-mosaicml:
Re: `batch_to_device`, see https://github.com/mosaicml/composer/pull/677 for a (partially recursive) refactor that doesn't aggravate pyright.

A-Jacobson:
> Re: `batch_to_device`, see https://github.com/mosaicml/composer/pull/677 for a (partially recursive) refactor that doesn't aggravate pyright.

Are you suggesting we use or modify that function instead? It doesn't support dictionaries with lists inside + still casts to the batch types we're attempting to remove.

.gitignore
```diff
@@ -162,3 +162,4 @@ package-lock.json
 
 # pycharm
 .idea/
+*.pyi
```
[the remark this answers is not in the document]
A-Jacobson: My pycharm didn't like them. Will unignore.

ravi-mosaicml (commented):
Overall looks good! Main comments re: docs and simplifying the `to_device logic.

composer/trainer/devices/device.py
```diff
@@ -123,3 +117,24 @@ def precision_context(self, precision: Union[str, Precision]) -> Generator[None,
             Generator[None, None, None]: A context for the precision.
         """
         pass
+
+
+def _map_batch(batch: Any, map_fn: Callable) -> Any:
```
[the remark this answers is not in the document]
A-Jacobson: I don't think this is too difficult to follow and there are probably some other things we'd like to map to the batch.

composer/trainer/devices/device.py
```diff
@@ -2,22 +2,23 @@
...
-import collections.abc
 from abc import ABC, abstractmethod
+from collections.abc import Mapping, Sequence
 from contextlib import contextmanager
-from typing import Generator, Sequence, TypeVar, Union
+from typing import Any, Callable, Generator, TypeVar, Union
 
 import torch
 import torch.nn
+from torch._six import string_classes
```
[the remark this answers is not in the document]
A-Jacobson: i didn't even look it's just 
`string_classes = (str, bytes)` ... lol done.

composer/trainer/devices/device.py
```diff
@@ -123,3 +117,24 @@ def precision_context(self, precision: Union[str, Precision]) -> Generator[None,
...
+    Args:
+        batch: Nested lists and dictionaries.
+        map_fn: A function to invoke on each element.
+
+    Returns:
+        Collections: The result of applying ``map_fn`` on each element of the ``batch``.
+        The type of ``batch`` is preserved.
+    """
+    if isinstance(batch, Mapping):
+        return {k: _map_batch(v, map_fn) for k, v in batch.items()}
```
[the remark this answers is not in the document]
A-Jacobson: I can't think of many mapping objects that support a `type()` other than converting TO `dict` style conversion but it can't hurt. try: except: is cheap.

ravi-mosaicml (approved):
LGTM

PR #735, opened 2022-03-14
Installing pyright via npm

### LINKED ISSUE #733: Pyright version pinning not working
It appears that pyright versioning is still not working when using `pip install pyright`. We should switch to using the node version, which respects versioning.

ravi-mosaicml:
Installing pyright via npm, as the pip pyright does not appear to pin versions properly.

Closes #733


Averylamp (approved):
lgtm

PR #749, opened 2022-03-14
Sample-wise Stochastic Depth Method Card

Landanjs:
Update sample-wise stochastic depth card to new format.

After looking at other cards, it seems like a lot more could be added in some sections, but wait for the future 🙂


composer/algorithms/blurpool/README.md
```diff
@@ -31,7 +31,8 @@ def training_loop(model, train_loader):
                       optimizers=opt,
                       replace_convs=True,
                       replace_maxpools=True,
-                      blur_first=True)
+                      blur_first=True,
+                      optimizers=opt)
```
hanlint, line 35: duplicate argument here

composer/algorithms/stochastic_depth/README.md
```diff
@@ -0,0 +1,92 @@
+# 🎰 Stochastic Depth (Sample-wise)
```
hanlint, line 1:
```suggestion
# 🎰 Stochastic Depth (Sample)
```

shorten this so no multi-line in the menu -- should do the same for Block

hanlint (approved):

composer/algorithms/blurpool/README.md
Landanjs: Ooops...

composer/algorithms/stochastic_depth/README.md
Landanjs: Done

PR #760, opened 2022-03-15
Chmod 644 all files

Averylamp:
Includes scripts directory because the scripts arent set up to be executed properly anyways


ravi-mosaicml (approved):
LGTM, though until we fix the pytorch internal dockerimage and stop running as root in cota, new files will become 755. Perhaps we should also add a Jenkins check to ensure that all files are 644?

Averylamp:
> LGTM, though until we fix the pytorch internal dockerimage and stop running as root in cota, new files will become 755. Perhaps we should also add a Jenkins check to ensure that all files are 644?

I think this is a really good thing to run as a pre-commit hook or something like that.  Doing a check for it if its < 2s I think is ok though

ravi-mosaicml:
`scripts/clean_notebooks.py` should be 0755.

PR #761, opened 2022-03-15
Fix `setup.py`

Averylamp:
* Include the yamls as part of the composer package data
* Exclude cached files (`__pycache__`, `*.pyc`)
* Removed webdatasets, as PyPI does not permit git dependencies (see #771)
* Properly hide dark mode images on the PyPI readme


ravi-mosaicml (approved):
LGTM

PR #763, opened 2022-03-16
Update dense label conversion and soft cross entropy to handle segmentation style labels

coryMosaicML:
This PR updates two functions to work with segmentation style labels:

`ensure_targets_one_hot` should now be able to convert to one-hot labels when there are spatial dimensions, and handle the case where some label indices are negative (likely because these would be ignored in computing cross entropy).  Negative indices are mapped to a zero-hot label along with a warning.

`soft_cross_entropy` should now be able to handle multiple spatial dimensions. When reduction is `mean` the loss is reweighed to account for zero-hot labels which results in them being ignored in the mean, analogous to setting an 'ignore_index' in hard cross entropy. Note: PyTorch's new soft cross entropy doesn't have this behavior unfortunately, so the composer `soft_cross_entropy` might have to stick around a bit longer.

These changes enable dense labels to work with the DeepLabV3+ benchmark, but should not modify behavior on our ResNet50 benchmark. As is, these changes only effect label smoothing.

Runs (Averaged over 3 seeds):
[ResNet50 + label smoothing before change 77.32% +/- 0.13 accuracy
ResNet50 + label smoothing after change 77.37 +/- 0.13 % accuracy
DeepLabV3+ + label smoothing after change 45.5 +/- 0.4 mIoU (Baseline 45.80 +/- 0.4 mIoU)](https://wandb.ai/mosaic-ml/cory-label-fix?workspace=user-corymosaicml)


hanlint:
@coryMosaicML this looks good! We have some tests for soft cross entropy in the (poorly placed) label smoothing tests:

https://github.com/mosaicml/composer/blob/dev/tests/algorithms/test_label_smoothing.py#L56

Could you help consolidate? The tests include some `xfail` ones marked for segmentation that may now pass with your changes.

coryMosaicML:
> @coryMosaicML this looks good! We have some tests for soft cross entropy in the (poorly placed) label smoothing tests:
> 
> https://github.com/mosaicml/composer/blob/dev/tests/algorithms/test_label_smoothing.py#L56
> 
> Could you help consolidate? The tests include some `xfail` ones marked for segmentation that may now pass with your changes.

Ok, I consolidated the tests and removed some `xfail`s.

composer/models/loss.py
```diff
@@ -199,20 +228,31 @@ def soft_cross_entropy(input: Tensor,
         assert reduction in ['sum', 'mean', 'none'], f"{reduction} reduction not supported."
         assert size_average is None, "size_average is deprecated"
         assert reduce is None, "reduce is deprecated"
-        assert ignore_index == -100, "ignore_index not supported."
-        probs = -1 * (target * F.log_softmax(input, dim=1))
+        if ignore_index != -100:
+            warnings.warn("ignore_index not supported when using dense labels. Ignoring targets with 0 probability.")
+        xentropy = -1 * (target * F.log_softmax(input, dim=1))
```
dblalock, line 233: skip the broadcast-multiply?
```suggestion
        xentropy = -(target * F.log_softmax(input, dim=1))
```

composer/models/loss.py
```diff
@@ -199,20 +228,31 @@ def soft_cross_entropy(input: Tensor,
...
-            probs = probs.sum(dim=0)
+            xentropy = xentropy.sum()
         elif reduction == 'mean':
-            probs = probs.mean(dim=0)
-
-        return probs
+            xentropy = xentropy.mean()
+            # Reweight loss to account for examples with less than 1 total probability (ignored examples)
+            total_prob = target.sum()
+            assert total_prob > 0, "No targets have nonzero probability"
```
dblalock, line 250: This isn't a precondition of the function, so we should probably raise a RunTimeError or ValueError here instead of asserting

tests/models/test_loss.py
```diff
@@ -17,6 +18,39 @@ def block_2D_targets():
...
+
+def generate_tensors():
+    return [
+        # Binary classification
+        fake_input_target_pairs((64, 2)),
+        # Classification
+        fake_input_target_pairs((64, 10)),
+        # Segmentation
+        fake_input_target_pairs((64, 2, 5, 5)),
+        fake_input_target_pairs((64, 10, 5, 5))
```
dblalock, line 45: Can we have a couple 5d inputs while we're at it? Medical imaging people often want to do segmentation on 3d volumes.

dblalock (approved):
LGTM. Just a couple minor suggestions

composer/models/loss.py
coryMosaicML: Done!

tests/models/test_loss.py
coryMosaicML: Done!

PR #766, opened 2022-03-16
Increase the timeout for test trainer equivalence

ravi-mosaicml:
It occassionally fails on Jenkins with a 2 second timeout. Increasing to 5 seconds.


Averylamp (approved):
Lgtm

PR #768, opened 2022-03-16
Added maximum versions to dependencies

ravi-mosaicml:
Added maximum versions to all dependencies to limit breaking changes.


moinnadeem (approved):
wow this is a ton of work, great work ravi!

PR #778, opened 2022-03-16
Change ColOut to work with segmentation tasks

Landanjs:
A quick fix to get colout to work with segmentation, but maybe there should be a more general approach in the future? This would take more time though.

The batch version was pretty quick to setup: add a target argument to colout function that defaults to none, apply colout to target if it is present and matches the spatial dimensions of the input.

To get the sample version of colout, I converted the ADE20k dataset to a VisionDataset, included all transforms into a single object, and changed the colout transform to accept a 2-tuple of inputs. 

Wandb runs with 3 seeds can be found [here](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized?workspace=user-landanjs). Specific seeds: [sample method run](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/runs/1eyzeaqq?workspace=user-landanjs) and [batch method run](https://wandb.ai/mosaic-ml/landan-deeplabv3-optimized/runs/33n14w05?workspace=user-landanjs)


composer/algorithms/colout/colout.py
```diff
@@ -173,16 +201,17 @@ def _apply_sample(self, state: State) -> None:
...
         self._transformed_datasets.add(dataset)
 
     def _apply_batch(self, state: State) -> None:
         """Transform a batch of images using the ColOut augmentation."""
-        inputs, labels = state.batch_pair
-        assert isinstance(inputs, Tensor), "Multiple Tensors not supported yet for ColOut"
-        new_inputs = colout_batch(inputs, p_row=self.p_row, p_col=self.p_col)
+        inputs, targets = state.batch_pair
+        assert isinstance(inputs, Tensor) and isinstance(targets, Tensor), \
+            "Multiple tensors not supported for this method yet."
```
coryMosaicML, line 211: This could be confusing as the method does support two tensors now. Maybe split this into separate checks for inputs and targets, or reword to specify that both inputs and targets should be single tensors?

composer/datasets/utils.py
```diff
@@ -140,20 +143,35 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
...
             if isinstance(t, transforms.ToTensor):
                 insertion_index = i
                 break
         if is_tensor_transform:
             insertion_index += 1
-        dataset.transform.transforms.insert(insertion_index, transform)
+
+        if is_target_transformed:
+            assert dataset.transforms is not None
+            dataset.transforms.transforms.insert(insertion_index, transform)
```
coryMosaicML, line 171: The difference between the two cases here is hard to spot. Maybe a comment could point it out.

coryMosaicML (commented):
Overall looks good! Main suggestions: have the resizing targets behavior be set via an algorithm hyper parameter rather than inferred from the shape of the targets, and possibly separate the dataset transform that operates on a pair vs. the one that operates on input only.

composer/algorithms/colout/colout.py
```diff
@@ -73,6 +77,23 @@ def colout_batch(input: ImgT, p_row: float = 0.15, p_col: float = 0.15) -> ImgT:
...
     X_colout = image_as_type(X_colout, type(input))
+
+    if (target is not None):
+        # If a target is given and has the same spatial dimensions as the input, reshape target like the input
+        Y_tensor = image_as_type(target, torch.Tensor)
+        if (len(Y_tensor.shape) > 2) and (Y_tensor.shape[-2:] == X_tensor.shape[-2:]):
+            Y_colout = Y_tensor[..., kept_row_idx, :]
+            Y_colout = Y_colout[..., :, kept_col_idx]
+            if not isinstance(target, torch.Tensor) or (target.ndim < Y_colout.ndim):
+                Y_colout = Y_colout.reshape(Y_colout.shape[-3:])
```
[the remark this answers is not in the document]
Landanjs: I stole this from the X_colout section. Was the X_colout code addressing a particular edge case? I can't tell. Here is the comment from X_colout:

```
# convert back to same type as input, and strip added batch dim if needed;
# we can't just reshape to input shape because we've reduced the spatial size
```

hanlint:
@Landanjs is this stale?

composer/cli/launcher.py
```diff
@@ -317,7 +317,7 @@ def _print_process_exit_status(global_rank: int, process: subprocess.Popen):
         output=output,
         stderr=stderr,
     )
-    error_msg = [f"Global rank {global_rank} (PID {process.pid}) excited with code {process.returncode}"]
+    error_msg = [f"Global rank {global_rank} (PID {process.pid}) exited with code {process.returncode}"]
```
Landanjs, line 320: This should be exited, right? 😂

composer/datasets/utils.py
Landanjs: Deleted this feature, so marking as resolved

Landanjs:
### Major changes:
1. Removed features to add segmentation ColOut transform to `VisionDataset` since our streaming dataloader won't inherit from `VisionDataset` anyway. We need a higher level discussion on how to add individual transforms to datasets.
2. Attempted to add a `resize_target='auto'` option to capture both usability and flexibility.
3. Changed `colout_batch` to take a single input of either a single object or a 2-tuple as opposed to individual `input` and `target` arguments. This was inspired by the realization that we would like input and output types to match. With the change, tuples should return tuples and single objects should return single objects.
4. Added unit tests for situations where ColOut should also resize the target

I've decided to keep the ColOut transform to operate on both a single image or a 2-tuple of images. Although we can't use this feature right now, might be useful to others and is futureproofing for when we can use this feature. Maybe this is the wrong call? 😅 

### Things missing:
1. Update method card
2. Update docstring tests -> how do these even work?

Someone tell me I need to do this so I do it 😂 

### Results

[DeepLabv3+ADE20k ColOut](https://wandb.ai/mosaic-ml/landan-deeplabv3plus-streaming/groups/test_colout/workspace?workspace=user-landanjs) (in progress)
- **Question**: Throughput is about 1.22x, is this expected for 0.15 drop probability?

[ResNet50+ImageNet1k ColOut](https://wandb.ai/mosaic-ml/landan-debug/groups/test_resnet50_colout/workspace?workspace=user-landanjs) (in progress)
- **Question**: Throughput is about 1.3x, is this expected for 0.15 drop probability?

composer/algorithms/colout/colout.py
```diff
@@ -39,16 +69,25 @@ def colout_batch(input: ImgT, p_row: float = 0.15, p_col: float = 0.15) -> ImgT:
...
-            of images of shape ``NCHW``.
-        p_row: Fraction of rows to drop (drop along H). Default: ``0.15``.
-        p_col: Fraction of columns to drop (drop along W). Default: ``0.15``.
+        sample (torch.Tensor | PIL.Image | Tuple[torch.Tensor, torch.Tensor] | Tuple[PIL.Image, PIL.Image]:):
+            Either a single tensor or image or a 2-tuple of tensors or images. When tensor(s), the tensor must be of shape
+            ``CHW`` for a single image or ``NCHW`` for a batch of images of shape.
+        p_row (float, optional): Fraction of rows to drop (drop along H). Default: ``0.15``.
+        p_col (float, optional): Fraction of columns to drop (drop along W). Default: ``0.15``.
+        resize_target (bool | str, optional): If ``sample`` is a tuple, whether to resize both objects in the tuple.
+            If set to 'auto', resizing both objects will be based on if the objects hav the same spatial dimensions.
```
coryMosaicML, line 78: Nit: "hav" -> "have"

Or maybe, "If set to 'auto', both objects will be resized if they have the same spatial dimensions."

coryMosaicML (approved):
Looks good to me! A few minor comments.

composer/algorithms/colout/colout.py
```diff
@@ -26,9 +27,38 @@
 __all__ = ["ColOut", "ColOutTransform", "colout_batch"]
 
 
-def colout_batch(input: ImgT, p_row: float = 0.15, p_col: float = 0.15) -> ImgT:
-    """Applies ColOut augmentation to a batch of images, dropping the same random rows and columns from all images in a
-    batch.
+def _should_resize_target(sample: Union[ImgT, Tuple[ImgT, ImgT]], resize_target: Union[bool, str]) -> bool:
```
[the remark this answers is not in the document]
Landanjs: Done!

composer/algorithms/colout/colout.py
```diff
@@ -73,12 +112,29 @@ def colout_batch(input: ImgT, p_row: float = 0.15, p_col: float = 0.15) -> ImgT:
...
+
+        return X_colout, Y_colout
+
     return X_colout
 
 
 class ColOutTransform:
     """Torchvision-like transform for performing the ColOut augmentation, where random rows and columns are dropped from
-    a single image.
+        up to two images.
```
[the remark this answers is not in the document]
Landanjs: What about explicitly stating "... up to two torch tensors or two PIL images."?

composer/algorithms/colout/colout.py
Landanjs: Ooops. Done!

PR #786, opened 2022-03-17
Set the random seed before each test.

### LINKED ISSUE #767: Fix `test_masked_accuracy`
#761 marked `test_masked_accuracy` with `@pytest.mark.xfail` as it randomly failed. For example:

Error:
```
assert tensor(0.0219) < 0.02
 +  where tensor(0.0219) = abs((tensor(0.5219) - (1.0 / 2)))
```

Stack trace:
```
ignore_index = -100, num_classes = 2

    @pytest.mark.parametrize("ignore_index", [-100])
    @pytest.mark.parametrize("num_classes", [2, 3, 4, 5])
    def test_masked_accuracy(ignore_index, num_classes):
        """Sanity check to make sure that masked accuracy has reasonable performance.
    
        Generates random targets and labels, and then ensures that the random targets and labels
        must hit at-chance accuracy.
    
        Args:
            batch_size (int): how many samples are in each batch
            ignore_index (Optional[int]): if present, the class index to ignore in accuracy calculations.
            num_classes (int): the number of classes in the classification task
        """
        batch_size = int(1e4)
        torchmetrics_masked_acc = MaskedAccuracy(ignore_index=ignore_index)
        # we're only testing binary accuracy -- expecteed accuracy should be 50%
        generated_preds = torch.rand((batch_size, num_classes))
        true_labels = torch.randint(low=0, high=num_classes - 1, size=(batch_size,))
    
        if ignore_index is not None:
            labels_mask = torch.rand((batch_size,))
            labels_mask[labels_mask > 0.8] = 1
            labels_mask[labels_mask <= 0.8] = 0
            labels_mask = labels_mask.bool()
            true_labels[labels_mask] = ignore_index
    
        true_labels = true_labels.float()
        generated_preds = generated_preds.float()
    
        torchmetrics_masked_acc.update(generated_preds, true_labels)
        final_acc = torchmetrics_masked_acc.compute()
>       assert abs(final_acc - (1.0 / num_classes)) < 0.02
E       assert tensor(0.0219) < 0.02
E        +  where tensor(0.0219) = abs((tensor(0.5219) - (1.0 / 2)))

tests/test_nlp_metrics.py:43: AssertionError
```

For more info: https://ci.int.mosaicml.com/job/scratch/job/command2/6016/testReport/junit/tests/test_nlp_metrics/test_masked_accuracy_2__100_/

FWIW 0.02 seems like an arbitrary cutoff. Could we make this test fully deterministic via seeding?

ravi-mosaicml:
Motivation behind using a deterministic seed for each test is to avoid test flakiness and ensure that PRs do not fail due to unrelated test failures. Instead, we should configure Jenkins to run nightly regressions across multiple seeds to identify tests that depend on the random initialization, and triage nightly regression errors as appropriate.

1. Added a pytest CLI option `--seed`. If not set, defaults to 0.
2. Added an `autouse` fixture to set the seed automatically before each test.
3. For tests that create instances of the trainer, manually specified the `rank_zero_seed`, as otherwise the trainer would re-seed the rng with `os.random`.
4. Removed the xfail on `test_masked_accuracy`, as the seed is now deterministic.
5. Unrelated cleanup: Removed unused fixtures from `conftest`; filtering warnings that should be ignored.

Closes #767

### DIFF
--- a/tests/algorithms/test_colout.py
+++ b/tests/algorithms/test_colout.py
@@ -88,21 +88,18 @@ def W(H) -> int:
 @pytest.fixture
 def fake_image(H: int, W: int, C: int) -> Image.Image:
     """Fake PIL Image."""
-    rng = np.random.RandomState(0)
-    return Image.fromarray((255 * rng.uniform(size=(H, W, C)).squeeze()).astype(np.uint8))
+    return Image.fromarray((255 * np.random.uniform(size=(H, W, C)).squeeze()).astype(np.uint8))
 
 
 @pytest.fixture
 def fake_image_tensor(H: int, W: int, C: int) -> torch.Tensor:
     """Fake image tensor."""
-    torch.manual_seed(0)
     return torch.rand(C, H, W)
 
 
 @pytest.fixture
 def fake_image_batch(H: int, W: int, C: int) -> torch.Tensor:
     """Fake batch of images."""
-    torch.manual_seed(0)
     return torch.rand(16, C, H, W)
 
 
--- a/tests/algorithms/test_cutmix.py
+++ b/tests/algorithms/test_cutmix.py
@@ -17,7 +17,6 @@ from tests.common import SimpleConvModel
 def fake_data(request):
     # Generate some fake data
     N, C, d1, d2, num_classes = request.param
-    torch.manual_seed(0)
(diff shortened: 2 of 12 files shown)


bandish-shah (approved):
LGTM

Issue #536, opened 2022-02-18
composer.models cleanup


A-Jacobson:
Putting this here to open discussion

The models module could use some general cleanup:
- Move `ComposerClassifier` out of `base.py` as it's not a base class.
- Move `ComposerTransformer` into `base.py` as it is a base class (pending work into general transformer factory)
- Unify naming scheme for models - some models are `Composer<Architecture>` some `Architecture`
- Move metrics out of `losses.py` and `nlp_metrics.py` into a separate metrics module.
- Maybe remove soft_cross_entropy as default pytorch CEloss supports dense labels now (depending on our supported pt versions)
- Initializers to their own module or removed (ongoing discussions about these).

ravi-mosaicml:
Closing for #834

PR #788, opened 2022-03-18
models cleanup - part 1: composermodel tasks

A-Jacobson:
part 1 of the models cleanup i've been threatening. see, https://github.com/mosaicml/composer/issues/536.

I've moved `ComposerClassifier` out of `base.py` as it's not a base class and have moved `ComposerTransformer` in. 

Started a `templates` module in `composer.models` to contain stuff like `ComposerClassifier`. In future pr's i'd like to make `ComposerClassifier` more configurable by letting it take a loss function and metrics on init. Future task specific templates such as a `ComposerDetector` could live here as well. 

Thoughts on the organization? @ravi-mosaicml, @Landanjs, @growlix


composer/models/base_hparams.py
```diff
@@ -1,19 +1,20 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-"""General `YAHP <https://docs.mosaicml.com/projects/yahp/en/stable/README.html>`_ interface for ComposerModels."""
+"""General `YAHP <https://docs.mosaicml.com/projects/yahp/en/stable/README.html>`_ interface for Base ComposerModels."""
```
ravi-mosaicml, line 3: Would like to keep this file as `model_hparams.py` to match the style of `callback_hparams.py`, `algorithm_hparams.py`, ...

composer/models/base_hparams.py
```diff
@@ -87,3 +88,40 @@ def initialize_object(self) -> ComposerModel:
             ComposerModel: The constructed :class:`.ComposerModel`
         """
         pass
+
+
+@dataclass
+class TransformerHparams(ModelHparams, ABC):
```
ravi-mosaicml, line 94: Should this be in its own file, and perhaps be called `HuggingfaceTransformerHparams`? From a quick glance, it seems to be pretty specific to the architecture of a huggingface transformer.

composer/models/classify_mnist/model.py
```diff
@@ -8,8 +8,8 @@
 import torch.nn as nn
 from torch.nn import functional as F
 
-from composer.models.base import ComposerClassifier
-from composer.models.model_hparams import Initializer
+from composer.models.base_hparams import Initializer
+from composer.models.templates import ComposerClassifier
```
ravi-mosaicml, line 12: With the benchmarks folder refactor, would there be implemented models inside the `composer.models` folder? If not, do we need to subfolder the base classes?

If we did go with the subfolder approach, not a huge fan of `templates`, as that implies to me that it is something that should be copied, pasted, and modified, rather than inherited. Thoughts about `architectures`? Open to other ideas too!

A-Jacobson: Yeah I'm not in love with the name either. These composermodels are actually everything except an architecture. They're forward pass logic + logging and they can be passed an architecture to train.

growlix: They're task-specific (`ComposerClassifier`, `ComposerDetector`) convenience classes, so maybe that's a good approach to naming the class. Perhaps something like `task_wrapper` or `model_task`.

growlix (commented):
Sounds like we're going with `template`. Not much to add beyond Ravi's comments.

composer/models/base_hparams.py
A-Jacobson: After talking to @moinnadeem I'm going to break all the nlp stuff out into its own pr/design. It's more of a redesign than a refactor to align it with the cv parts right now.

A-Jacobson:
> Sounds like we're going with `template`. Not much to add beyond Ravi's comments. 

I was switching it to task :)

composer/models/base_hparams.py
A-Jacobson: Just checked `algorithm_hparams.py`. seems like algorithms hparams are all defined in their own top level `hparam.py` file and they all gain some loading functionality from the base `AlgorithmHparams`object.

That's not at all how things have been handled in the models module. Each model has its own hparams file. and though those hparams objects do inherit from `ModelHparams`.. they aren't gaining much from it. The only shared interfaces we're presupposing are `initializers` and `num_classes` and those are not universal to all models. 

Collecting all the model hparams in a shared file would make our codebase a lot cleaner but it's a odds with the pending `benchmarks` folder setup.

A-Jacobson:
I took out all the NLP related changes. Will be punting those to a later PR that will set us up for better HF integration. This PR now only moves initializers out of `model_hparams` and moves `ComposerClassifier` to `models.tasks`. 

To address @ravi-mosaicml 's question about models moving out of composer: yes, we hope they will though there are still some implementation difficulties with breaking them out while maintaining full functionality, that's happening elsewhere. 

These tasks would stay in as they are useful general utils that could be used to quickly build external models. They also standardize training and metric logging for internal use cases.

composer/models/transformer_hparams.py
```diff
@@ -49,4 +49,4 @@ def validate(self):
             raise Exception("Only one of pretrained_model_name or model_config can be provided.")
 
         if self.use_pretrained and self.model_config:
-            raise Exception("A model cannot load pretrained weights from configuration.")
+            raise Exception("A model cannot load pretrained weights from configuration.")
```
ravi-mosaicml, line 52: newline

composer/models/initializers.py
```diff
@@ -0,0 +1,50 @@
...
+        initializer_dict = {
+            "kaiming_normal": kaiming_normal,
+            "kaiming_uniform": kaiming_uniform,
+            "bn_uniform": bn_uniform,
+            "bn_ones": bn_ones,
+            "xavier_uniform": xavier_uniform
+        }
+        if self.value not in initializer_dict:
+            raise ValueError(f"Initializer '{self.value}' not found.")
+        return initializer_dict[self.value]
```
ravi-mosaicml, line 50: newline

ravi-mosaicml (approved):
LGTM; I like the name `tasks`. Seems specific.

PR #790, opened 2022-03-18
models cleanup - part 2, metrics and loss

A-Jacobson:
moved metrics and loss functions out of models to their own modules. 

@moinnadeem do you think it makes sense to move the logic from `LanguageCrossEntropyLoss.update` and `HFCrossEntropyLoss.update` into their respective `ComposerModels.validate` and use a standard `CrossEntropyLoss` instead? 

@coryMosaicML do we still need `soft_cross_entropy`? It's merged into pytorch (1.10) + I think we can convert labels when we need now?

### DIFF
--- a/composer/algorithms/cutmix/cutmix.py
+++ b/composer/algorithms/cutmix/cutmix.py
@@ -14,7 +14,7 @@ from torch.nn import functional as F
 
 from composer.core import Algorithm, Event, State
 from composer.loggers import Logger
-from composer.models.loss import _check_for_index_targets
+from composer.loss.loss import _check_for_index_targets
 
 log = logging.getLogger(__name__)
 
--- a/composer/loss/loss.py
+++ b/composer/loss/loss.py
@@ -5,6 +5,8 @@ from typing import Optional
 from torch import Tensor
 from torch.nn import functional as F
 
+__all__ = ["soft_cross_entropy"]
+
 
 def _infer_target_type(input: Tensor, targets: Tensor) -> str:
     """Infers whether the target is in indices format or one_hot format.
--- a/composer/loss/loss.py
+++ b/composer/loss/loss.py
@@ -1,0 +1,103 @@
+from __future__ import annotations
+
+from typing import Optional
+
+from torch import Tensor
+from torch.nn import functional as F
+
+
+def _infer_target_type(input: Tensor, targets: Tensor) -> str:
+    """Infers whether the target is in indices format or one_hot format.
+
+    Example indices format: [1, 4, 7] Example one_hot format [[0, 1, 0], [1, 0, 0], ...]
+    """
+    if input.shape == targets.shape:
+        return 'one_hot'
+    elif input.ndim == targets.ndim + 1:
+        return 'indices'
+    else:
+        raise RuntimeError(f'Unable to infer indices or one_hot. Targets has shape {targets.shape}'
+                           f' and the inputs to cross entropy has shape {input.shape}. For one_hot, '
+                           'expect targets.shape == inputs.shape. For indices, expect '
+                           'inputs.ndim == targets.ndim + 1')
+
+
+def ensure_targets_one_hot(input: Tensor, targets: Tensor) -> Tensor:
+    if _infer_target_type(input, targets) == 'indices':
+        targets = F.one_hot(targets, num_classes=input.shape[1])
+    return targets
+
+
+def _check_for_index_targets(targets: Tensor) -> bool:
+    """Checks if a given set of targets are indices by looking at the type."""
+    index_types = ['torch.LongTensor', 'torch.cuda.LongTensor']
+    return targets.type() in index_types
+
+
+def soft_cross_entropy(input: Tensor,
+                       target: Tensor,
+                       weight: Optional[Tensor] = None,
+                       size_average: Optional[bool] = None,
+                       ignore_index: int = -100,
+                       reduce: Optional[bool] = None,
+                       reduction: str = 'mean'):
+    r"""Drop-in replacement for :class:`~torch.nn.CrossEntropyLoss` that can handle class indices or one-hot labels.
+
+    Args:
+        input (torch.Tensor) : :math:`(N, C)` where `C = number of classes` or :math:`(N, C, H, W)`
+            in case of 2D Loss, or :math:`(N, C, d_1, d_2, ..., d_K)` where :math:`K \geq 1`
+            in the case of K-dimensional loss. `input` is expected to contain unnormalized scores
+            (often referred to as logits).
+        target (torch.Tensor) : If containing class indices, shape :math:`(N)` where each value is
+            :math:`0 \leq \text{targets}[i] \leq C-1`, or :math:`(N, d_1, d_2, ..., d_K)` with
+            :math:`K \geq 1` in the case of K-dimensional loss. If containing class probabilities,
+            same shape as the input.
+        weight (torch.Tensor, optional): a manual rescaling weight given to each
+            class. If given, has to be a Tensor of size `C`. Default: ``None``.
+        size_average (bool, optional): Deprecated (see `reduction`). By default,
+            the losses are averaged over each loss element in the batch. Note that for
+            some losses, there multiple elements per sample. If the field ``size_average``
+            is set to ``False``, the losses are instead summed for each minibatch. Ignored
+            when reduce is ``False``. Default: ``True``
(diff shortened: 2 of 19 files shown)


coryMosaicML:
On `soft_cross_entropy`: If we are ok with requiring PyTorch 1.10, we don't need `soft_cross_entropy` most of the time. However, for models that ignore specific indices and use `reduction='mean'`, the PyTorch cross entropy won't give the correct behavior for non-index labels.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -14,7 +14,7 @@
 
 from composer.core import Algorithm, Event, State
 from composer.loggers import Logger
-from composer.models.loss import _check_for_index_targets
+from composer.loss.loss import _check_for_index_targets
```
ravi-mosaicml, line 17: Curious if the loss functions are specific to the models, and if so, would it make sense for it to be something like `composer.models.loss` instead? Thinking it would be nice to be parallel to how schedulers are part of composer.optim. I'm hesitant to add a high level folder here, but can be convinced if this is what other frameworks do or the loss function is uncoupled from the models.

ravi-mosaicml: In addition, `_check_for_index_targets` should be a public function, if it is being imported by another module. I know it's tangential to this PR, but would you be able to clean that up here?

composer/metrics/metrics.py
```diff
@@ -250,3 +153,30 @@ def compute(self) -> Tensor:
...
+class LossMetric(Metric):
+    """Turns a torch.nn Loss Module into distributed torchmetrics Metric.
+
+    Args:
+        loss_function (callable): loss function to compute and track.
+
+        dist_sync_on_step (bool, optional): sync distributed metrics every step. Default: ``False``.
+    """
+
+    def __init__(self, loss_function: callable, dist_sync_on_step=False):
```
ravi-mosaicml, line 167: missing type annotations

composer/metrics/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""A collection of common torchmetrics."""
```
ravi-mosaicml, line 1: Similar question re: folder organization.

As an alternative, I'm wondering about a structure like this, where the loss functions, metrics, and model architectures are grouped together by model type. For example:

```
composer/models/
composer/models/nlp/huggingface_transformer_model.py
composer/models/nlp/metrics.py
composer/models/nlp/loss.py
```

I don't think I have a strong opinion either way, but am curious if others have thoughts or if it would be worthwhile to discuss during design review.

composer/loss/loss.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import annotations
```
ravi-mosaicml, line 1: did not check this file for correctness. I presume it was only a refactor.

composer/loss/loss.py
```diff
@@ -0,0 +1,105 @@
...
+        probs = probs.sum(dim=1)
+
+        if reduction == 'sum':
+            probs = probs.sum(dim=0)
+        elif reduction == 'mean':
+            probs = probs.mean(dim=0)
+
+        return probs
+    else:
+        raise ValueError(f"Unrecognized target type {target_type}")
```
ravi-mosaicml, line 105: missing newline

composer/loss/loss.py
```diff
@@ -0,0 +1,105 @@
...
+    elif input.ndim == targets.ndim + 1:
+        return 'indices'
+    else:
+        raise RuntimeError(f'Unable to infer indices or one_hot. Targets has shape {targets.shape}'
+                           f' and the inputs to cross entropy has shape {input.shape}. For one_hot, '
+                           'expect targets.shape == inputs.shape. For indices, expect '
+                           'inputs.ndim == targets.ndim + 1')
+
+
+def ensure_targets_one_hot(input: Tensor, targets: Tensor) -> Tensor:
```
ravi-mosaicml, line 27: missing from `__all__`

composer/loss/loss.py
```diff
@@ -0,0 +1,105 @@
...
+    if _infer_target_type(input, targets) == 'indices':
+        targets = F.one_hot(targets, num_classes=input.shape[1])
+    return targets
+
+
+def _check_for_index_targets(targets: Tensor) -> bool:
+    """Checks if a given set of targets are indices by looking at the type."""
+    index_types = ['torch.LongTensor', 'torch.cuda.LongTensor']
```
ravi-mosaicml, line 35: Can we compare against the dtype? Not a huge fan of string comparisons for type checks.

```suggestion
    index_types = [torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64]
```

composer/loss/loss.py
```diff
@@ -0,0 +1,105 @@
...
+    return targets.type() in index_types
```
ravi-mosaicml, line 36:
```suggestion
    return targets.dtype in index_types
```

ravi-mosaicml (commented):
Thanks for working on this, but curious to hear others' thoughts on the folder structure proposed here. (I don't think we discussed this change as part of https://www.notion.so/Benchmarks-folder-5000fe694bf04848b4f67174efbe2508? Apologies if I am mis-remembering the design meeting!)

An alternative structure is similar to how it is done now, where we group together the metrics / loss / model by model type rather than by function. (The ComposerModel base class suggests this design). For example:

```
composer/models/
composer/models/nlp/huggingface_transformer_model.py
composer/models/nlp/metrics.py
composer/models/nlp/loss.py
```

composer/algorithms/cutmix/cutmix.py
A-Jacobson: loss functions are used by models but they aren't specific to individual models. for example: cross_entropy can be used for almost any of our models.

will change `check_for_index_targets` back. I changed it during the doc update because i thought it was specific to soft_cross_entropy only. Though maybe it belongs in some sort of util file.

composer/metrics/__init__.py
A-Jacobson: talk about this with cory and matthew. We consider loss functions an area of research we are/should be exploring. metrics are separate from models. Am working with moin to make these even more general by moving some of the logic composermodels.

composer/metrics/metrics.py
```diff
@@ -0,0 +1,185 @@
...
+    tp = ((preds == class_index) * (targets == class_index)).to(torch.long).sum()
+    fp = ((preds == class_index) * (targets != class_index)).to(torch.long).sum()
+    tn = ((preds != class_index) * (targets != class_index)).to(torch.long).sum()
+    fn = ((preds != class_index) * (targets == class_index)).to(torch.long).sum()
+    sup = (targets == class_index).to(torch.long).sum()
+
+    return tp, fp, tn, fn, sup
+
+
+class CrossEntropyLoss(Metric):
```
coryMosaicML, line 124: Maybe it makes sense to refer to this as just `CrossEntropy` instead of `CrossEntropyLoss` if it is meant to only be used as a metric and not a loss during training. Otherwise I'd expect to find it `loss.py`.

composer/metrics/nlp.py
```diff
@@ -52,7 +52,7 @@ def compute(self):
         return self.correct.float() / self.total
 
 
-class CrossEntropyLoss(Metric):
+class LanguageCrossEntropyLoss(Metric):
```
coryMosaicML, line 55: Similar to the other, if this is only meant to be used as a metric for evaluation, it might make more sense to call it `LanguageCrossEntropy`. If it gets used as a loss during training I'd expect it to be in `loss.py` or the nlp equivalent.

coryMosaicML (approved):
Overall looks good! 

A very minor thing: I'm a little bit confused about what is called a `metric` and what is called a `loss`. To me, if it gets used as the training objective I'd expect to find it in `loss.py`, and if it's used as an evaluation metric I'd expect to find it in `metrics.py` but it doesn't seem to be the case here.

A-Jacobson:
> Overall looks good! 
> 
> 
> 
> A very minor thing: I'm a little bit confused about what is called a `metric` and what is called a `loss`. To me, if it gets used as the training objective I'd expect to find it in `loss.py`, and if it's used as an evaluation metric I'd expect to find it in `metrics.py` but it doesn't seem to be the case here. 

That's the distinction I was trying to make. Though it looks like I could clean up some of the naming to make this more obvious. Especially for the nlp cases. I also added LossContainer on the hope that we can just wrap stuff from loss.py rather than re-implementing everything as a metric in the future.

PR #791, opened 2022-03-18
models cleanup - part 3: one model family per directory (cifar resnets)

A-Jacobson:
Combined all the `cifar_resnets` into one directory as they were sort of taking over the models module. In the future I think we should try to keep model families together if possible. 

yaml config looks like this:

```yaml
 model:
  resnet_cifar10:
    model_name: cifar_resnet_56
```

was debating:
```yaml
 family:
  resnet_cifar10:
    model_name: cifar_resnet_56
```
@growlix I think these are yours?

In the future, I'd like to expand the effecientnet and vit model families with this pattern.


growlix:
Great idea.

> @growlix I think these are yours?

Indeed I added resnet20.

I'd personally go with `model_family` over `family` to provide a bit more context, but otherwise I like this structure.

composer/models/resnet_cifar10/model.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import List, Optional
+
+from composer.models.base import ComposerClassifier
+from composer.models.model_hparams import Initializer
+from composer.models.resnet_cifar10.resnets import CIFARResNet, ResNet9
+
+__all__ = ["CIFAR10ResNet"]
+
+
+class CIFAR10ResNet(ComposerClassifier):
```
growlix, line 14: The nomenclature seems a bit confusing: there exists `CIFAR10ResNet` (a `ComposerClassifier`), `CIFARResNet` (an `nn.Module`), and `resnet_cifar10` (a registry/config key). What do you think about renaming to `CIFAR10ResNet` -> `ComposerCIFARResNet` and `resnet_cifar10` -> `cifar_resnet` (or ResNetCIFAR/resnet_cifar)?

composer/models/resnet_cifar10/model.py
```diff
@@ -0,0 +1,50 @@
...
+    """ResNet models for CIFAR10 extending :class:`.ComposerClassifier`.
+
+    From `Deep Residual Learning for Image Recognition <https://arxiv.org/abs/1512.03385>`_ (He et al, 2015).
+
+    Args:
+        model_name (str): ``"cifar_resnet_9"``, ``"cifar_resnet_20"``, or ``"cifar_resnet_56"``.
```
growlix, line 20: Given that ResNet9 is not a canonical ResNet, I might add a disclaimer to that effect and/or link directly to the ResNet9 documentation.

growlix (approved):
This is much nicer than how things were before. I just left some minor nomenclature and documentation suggestions.

ravi-mosaicml (commented):
Overall looks fine. Should I take another look after parts 1 and 2 are merged?

PR #795, opened 2022-03-21
[XS] Consolidating NLP Import Message

moinnadeem:
Moves the "Composer was installed without NLP utilities" to a shared file instead of repeating the message in multiple places


composer/utils/dynamic_import.py
```diff
@@ -3,9 +3,14 @@
 """Dynamically import a Python object (e.g. module, class, function, ...)."""
 
 import importlib
+import textwrap
 from typing import Any
 
-__all__ = ["import_object"]
+__all__ = ["import_object", "NLP_IMPORT_MESSAGE"]
+
+NLP_IMPORT_MESSAGE = textwrap.dedent("""\
```
growlix, line 11: `NLP_IMPORT_ERROR_MESSAGE` might be more informative than `NLP_IMPORT_MESSAGE`.

growlix (approved):
`NLP_IMPORT_ERROR_MESSAGE` might be more informative than `NLP_IMPORT_MESSAGE`, but your call.

ravi-mosaicml (changes requested):
I'm hesitant to hardcode nlp stuff in composer.utils. Intead, would it instead be possible to write a helper exception class like this, which could then live in `composer.utils`?

```python

class MissingCondinalImportError(ImportError):
   def __init__(extra_deps_group: str, conda_pacakge: str, conda_channel: str = "conda-forge")`
      ...
```
Then, we can do `raise MissingCondinalImportError(...) from e`.

ravi-mosaicml (approved):
LGTM, though:
1. Rename `composer.utils.dynamic_import` to `composer.utils.import_helpers`
2. Re-export `MissingConditionalImportError` on `composer.utils.__init__`

composer/utils/__init__.py
```diff
@@ -2,8 +2,8 @@
 
 """Helper utilities."""
 from composer.utils.checkpoint import load_checkpoint, save_checkpoint
-from composer.utils.dynamic_import import import_object
 from composer.utils.file_retriever import get_file
+from composer.utils.import_helpers import MissingConditionalImportError, import_object
```
ravi-mosaicml, line 6: nit: add this to `__all__` below

moinnadeem: Done!

PR #810, opened 2022-03-22
Remove the LoggerData and LoggerDataDict types

ravi-mosaicml:
This PR switches the types of `LoggerData` to `Any` and `LoggerDataDict` to `Dict[str, Any]`.

While this change makes typing looser, it adds flexibility for what users can log via the logger callbacks. For example, [wandb](https://docs.wandb.ai/ref/python/log) knows how to log images and pyplot objects. With this change, one can now log an image to wandb without raising a type error (before it would work, but pyright would complain).


jbloxham (approved):
LGTM

PR #811, opened 2022-03-22
Remove config from the logger base classes

ravi-mosaicml:
Moved the saving of the config out of the loggers. The concept of a config as a YAML dictionary is a yahp-centric feature. Moreover, only some loggers (mainly, wandb) know how to handle a config. As such, it did not make sense to have it be part of the base class.

Instead, this PR:

1. Adds the `config` as an hparams argument of the wandb logger, and removes it from the file logger and progress bar logger.
2. Modifies the entrypoint script to:
    1. Set the config as on the wandb logger hparams, if the wandb logger is being used.
    2. Print the config to the terminal
    3. Log the config as a file artifact

Logging the config as an artifact will be needed as part of #421.


composer/loggers/logger_hparams.py
```diff
@@ -104,10 +95,12 @@ class WandBLoggerHparams(LoggerDestinationHparams):
             If not specified, the :attr:`~composer.loggers.logger.Logger.run_name` will be used.
         entity (str, optional): WandB entity name.
         tags (str, optional): WandB tags, comma-separated.
-        log_artifacts (bool, optional): See
-            :class:`~composer.loggers.wandb_logger.WandBLogger`.
+        config (Dict[str, Any], optional): WandB run configuration.
+        flatten_config (bool, optional): Whether to flatten the run config. (default: ``False``)
```
jbloxham, line 99: This docstring has `flatten_config`, but the method signature has `flatten_hparams`. I prefer `flatten_config`, personally.

jbloxham (approved):

PR #818, opened 2022-03-24
Updating setup.py with missing dependancies

dlmgary:
#### Description

This pull request updates `setup.py` with missing dependancies so developers have all the tools needed to lint their code (e.g., by running `pylint composer examples`, or `make lint`, or `make style`). Prior to this, running linters would fail and require manual installation of dependancies. It also updates `numpy` to a newer version.

As an extra, it updates the string formatting of the file to use double quotes—which is used throughout the project.


dlmgary:
@hanlint oopsies, my bad! You're totally right. 

Quick one, setting up environment with `pip install .[dev]` works but it seemed like `pyright` is still missing. I'll remove the other dependencies that I added, but should I add `pyright` to the dev part of the `setup.py`?

```
(composer) ➜  composer git:(fresh-environment-installed-dependancies-update) ✗ pip install .[dev]
...

(composer) ➜  composer git:(fresh-environment-installed-dependancies-update) ✗ make lint           
python   -m isort -c --diff composer examples tests
python   -m yapf -dr composer examples tests
python   -m docformatter -r --wrap-summaries 120 --wrap-descriptions 120 composer examples tests
pyright   composer examples tests
make: pyright: No such file or directory
make: *** [lint] Error 1
```

ravi-mosaicml:
> Quick one, setting up environment with `pip install .[dev]` works but it seemed like `pyright` is still missing. I'll remove the other dependencies that I added, but should I add `pyright` to the dev part of the `setup.py`?

We have found https://pypi.org/project/pyright/ to be unreliable with respect to version pinning, so pyright will need to be installed manually.

dlmgary:
Oh, interesting. I wonder why that happens (but that's probably an unrelated issue).

Do you want to keep this PR as one just for updating the string formatting and updating `numpy` versions? Or you prefer to just close it?

hanlint:
Yeah we think it's related to the pyright pypi package, which is not an unofficial package. 

If you would update the `Contributing.md` with the pyright install instructions, and maybe add a comment in our `README`, we can still merge this PR, thanks!

setup.py
```diff
@@ -66,96 +66,96 @@ def package_files(prefix: str, directory: str, extension: str):
     "torch>=1.9,<2",
     "yahp==0.1.0",
     "requests>=2.26.0,<3",
-    "numpy==1.21.5",  # TODO should this be >=1.21.5.10,<2
+    "numpy>=1.21.5.10,<2",
```
dlmgary, line 69: Only update here

CONTRIBUTING.md
```diff
@@ -12,33 +12,45 @@ We welcome contributions for bug fixes, new efficient methods you'd like to cont
...
+# NOTE. You might need sudo permissions.
+npm install -g pyright@1.1.224
+```
 
 ## Submitting a contribution
 
 To submit a contribution:
 
-1. Fork a copy of the [Composer](https://github.com/mosaicml/composer) library to your own account.
+1\. Fork a copy of the [Composer](https://github.com/mosaicml/composer) library to your own account.
```
dlmgary, line 36: Adds manual numbering here so that code block doesn't break the document numbering order.

dlmgary:
Sounds good! I added some comments to the `README.md` and `CONTRIBUTING.md`. The only update to the dependencies was to `numpy` per the TODO comment in `setup.py`.

hanlint (approved):
Looks good, thanks @dlmgary !

setup.py
hanlint:
```suggestion
    "numpy>=1.21.5,<2",
```
Removing the `.10` that is causing some jenkins errors

PR #821, opened 2022-03-24
Adding tokenizers as optional with reasonable defaults

moinnadeem:
Adds reasonable defaults for model input names, with optional tokenizer acceptability if needed

### DIFF
--- a/composer/algorithms/seq_length_warmup/seq_length_warmup.py
+++ b/composer/algorithms/seq_length_warmup/seq_length_warmup.py
@@ -197,7 +197,7 @@ class SeqLengthWarmup(Algorithm):
                 raise RuntimeError("'labels' must be in model inputs")
 
             # create fake inputs
-            vocab_size = len(self._original_model.tokenizer)
+            vocab_size = self._original_model.config.vocab_size
 
             # simplifying assumption: Composer doesn't support model-parallelism,
             # so the first parameter's device is likely the same device for
--- a/composer/models/bert/model.py
+++ b/composer/models/bert/model.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Mapping, Sequence, Union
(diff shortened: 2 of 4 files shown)


composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -197,7 +197,7 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
                 raise RuntimeError("'labels' must be in model inputs")
 
             # create fake inputs
-            vocab_size = len(self._original_model.tokenizer)
+            vocab_size = self._original_model.config.vocab_size
```
abhi-mosaic, line 200: Why is this necessary? We are creating dummy inputs to do 1 fwd/bkw, so can we fill it with all zeros? That way we don't have to rely on the model having `.config` attribute.

composer/models/bert/model.py
```diff
@@ -45,12 +45,20 @@ class BERTModel(ComposerTransformer):
...
-                 tokenizer: transformers.BertTokenizer) -> None:
+    def __init__(self,
+                 module: transformers.BertModel,
+                 config: transformers.BertConfig,
+                 tokenizer: Optional[transformers.BertTokenizer] = None) -> None:
+
+        if tokenizer is None:
+            model_inputs = {"input_ids", "attention_mask", "token_type_ids"}
+        else:
+            model_inputs = set(tokenizer.model_input_names)
```
abhi-mosaic, line 56: Not recommending changing this now, but this seems to suggest that the BERTModel has a set of required input names, would it make sense to validate those like `assert "input_ids" in tokenizer.model_input_names` ? If I passed a tokenizer with the wrong input names, would the rest of the model crash?

abhi-mosaic (approved):
Left 2 brief comments above, but otherwise LGTM! Feel free to address or push to a future PR.

PR #825, opened 2022-03-25
Update MixUp to allow use of index labels

coryMosaicML:
This PR updates the MixUp algorithm to allow the use of index labels, and makes this the new default behavior. This is consistent with other implementations and likely the expectations of users. This PR also removes the need to specify the number of classes as a hyper parameter, and refactors the dense/one-hot mode to instead infer this from the output shape similar to how the current implementation of label smoothing works. These changes also allow mixup to be used for semantic segmentation.

Some downsides:
- When `interpolate_loss=True` the training loss is no longer logged correctly, as only the result of the first call to `loss` is logged.
- The functional form is not as clean as the conversion to one-hot labels no longer happens in `mixup_batch`.
- Composition can become strange if multiple algorithms modify the train loss like this. Right now only mixup does this.
 
Runs (3 seeds, streaming data):
[ResNet50-i1k + mixup before change Acc. 77.05% +/- 0.15
ResNet50-i1k + mixup after change interpolate_loss=True Acc. 77.01% +/- 0.16
ResNet50-i1k + mixup after change interpolate_loss=False Acc. 77.12% +/- 0.25
DeepLabV3+-ADE20k + mixup after change interpolate_loss=True mIoU 45.64 +/- 0.23
DeepLabV3+-ADE20k + mixup after change interpolate_loss=False mIoU 45.57 +/- 0.25](https://wandb.ai/mosaic-ml/cory-mixup-changes)

Note on DeepLabV3+, running with `interpolate_loss=False` (using dense/one-hot labels) reduces the throughput due to the conversion to one-hots.


composer/algorithms/mixup/mixup.py
```diff
@@ -105,18 +95,19 @@ class MixUp(Algorithm):
...
 
     Args:
-        num_classes (int): the number of classes in the task labels.
         alpha (float, optional): the psuedocount for the Beta distribution used to sample
             mixing parameters. As ``alpha`` grows, the two samples
             in each pair tend to be weighted more equally. As ``alpha``
             approaches 0 from above, the combination approaches only using
             one element of the pair. Default: ``0.2``.
+        interpolate_loss (bool, optional): Interpolates the loss rather than the labels.
+            A useful trick when using a cross entropy loss. Default: ``True``
```
dblalock, line 104: Should warn user about this likely yielding incorrect behavior for other losses. Also not sure it should be on by default, but I'll defer to your judgment there.

composer/algorithms/mixup/mixup.py
```diff
@@ -127,46 +118,67 @@ class MixUp(Algorithm):
...
-            target,
-            mixing=self.mixing,
-            num_classes=self.num_classes,
-        )
+        if event == Event.INIT:
+            # Grab the loss function
+            if isinstance(state.model, ComposerModel):
+                self._loss_fn = state.model.loss
+            else:
+                raise RuntimeError("Model must be of type ComposerModel")
```
dblalock, line 141: A little more future-proof to test for behavior rather than identity here; something like:
```suggestion
            try:
                self._loss_fn = state.model.loss
                assert isinstance(self._loss_fn, Callable)
            except AttributeError, AssertionError:
                raise RuntimeError("Model must be of type ComposerModel or implement method loss()")
```

composer/algorithms/mixup/mixup.py
```diff
@@ -127,46 +118,67 @@ class MixUp(Algorithm):
...
+
+        if event == Event.AFTER_DATALOADER:
+            input, target = state.batch_pair
+            assert isinstance(input, torch.Tensor) and isinstance(target, torch.Tensor), \
+                "Multiple tensors for inputs or targets not supported yet."
```
dblalock, line 146: `raise NotImplementedError` instead of `assert`-ing

composer/algorithms/mixup/mixup.py
```diff
@@ -127,46 +118,67 @@ class MixUp(Algorithm):
...
+            )
 
-        state.batch = (new_input, new_target)
+            state.batch = (new_input, target)
+
+        if not self.interpolate_loss and event == Event.BEFORE_LOSS:
+            # Interpolate the targets
+            input, target = state.batch_pair
+            assert isinstance(state.outputs, torch.Tensor), "Multiple output tensors not supported yet"
+            assert isinstance(target, torch.Tensor), "Multiple target tensors not supported yet"
```
dblalock, line 164: also `raise NotImplementedError` instead of `assert`

composer/algorithms/mixup/mixup.py
```diff
@@ -127,46 +118,67 @@ class MixUp(Algorithm):
...
+            # Interpolate to get the new target
+            mixed_up_target = (1 - self.mixing) * target + self.mixing * permuted_target
+            # Create the new batch
+            state.batch = (input, mixed_up_target)
+
+        if self.interpolate_loss and event == Event.AFTER_LOSS:
+            input, target = state.batch_pair
+            # Interpolate the loss
+            assert self._loss_fn is not None, "loss_fn should be set on Event.INIT"
+            new_loss = self._loss_fn(state.outputs, (torch.Tensor(), self.permuted_target))
```
dblalock, line 177: Do we want snapshotted loss from INIT here, or do we want to read whatever the loss function currently is at this point? Thinking about case of someone combining several methods that alter the loss function, with mixup not last in the order. The snapshot-ed loss might just the vanilla xent, but the current loss could be something more complex. Using only the former will make the permuted targets use a different loss than the original targets, which is no longer equivalent to interpolating the targets during AFTER_BATCH.

composer/algorithms/mixup/mixup.py
```diff
@@ -127,46 +118,67 @@ class MixUp(Algorithm):
...
+            assert isinstance(state.loss, torch.Tensor), "Multiple losses not supported yet"
+            state.loss *= (1 - self.mixing)
+            assert isinstance(new_loss, torch.Tensor), "Multiple losses not supported yet"
```
dblalock, line 180: `raise NotImplementedError`

composer/algorithms/mixup/mixup.py
dblalock: also I'm a little confused as to why we construct an empty Tensor here in the call to `self._loss_fn`. Is it just to get it to typecheck? This probably works for most loss functions, but lying to an arbitrary user `ComposerModel` about the batch input scares me--the user's code might assume it's getting the real input, or at least a tensor of the correct shape. Could we just pass in the input tensor?

docs/source/getting_started/welcome_tour.rst
```diff
@@ -112,11 +112,21 @@ simple:
...
+
+            if event == Event.AFTER_DATALOADER:
+                new_input, _, self.permuted_target, self.mixing = mixup_batch(input, target, alpha=0.2)
+                state.batch = (new_input, target)
+
+            if event == Event.AFTER_LOSS:
+                modified_batch = (input, self.permuted_target)
+                new_loss = state.model.loss(state.outputs, modified_batch)
+                state.loss *= (1 - self.mixing)
+                state.loss += self.mixing * new_loss
```
dblalock, line 129: can this be one line? Took me a second to see that this was just a convex combo with it split on two lines with in-place operators.

docs/source/trainer/using_the_trainer.rst
```diff
@@ -172,15 +172,14 @@ argument.
...
-            MixUp(num_classes=10, alpha=0.1),
-    ])
+    trainer = Trainer(model=model,
+                        train_dataloader=train_dataloader,
+                        eval_dataloader=eval_dataloader,
+                        max_duration='2ep',
+                        algorithms=[
+                            LayerFreezing(freeze_start=0.5, freeze_level=0.1),
+                            MixUp(alpha=0.1),
+                        ])
```
dblalock, line 182: over-indentation

tests/trainer/test_trainer.py
```diff
@@ -357,7 +357,7 @@ def algorithm(self, request):
             pytest.importorskip("torch", minversion="1.10", reason="Pytorch 1.10 required.")
 
         # create the algorithms
-        if name in ('cutmix, mixup'):  # these algos have required algorithms
+        if name in ('cutmix'):  # these algos have required algorithms
```
dblalock, line 360:
```suggestion
        if name in ('cutmix'):  # these algos have required hparams
```

composer/algorithms/mixup/mixup.py
```diff
@@ -54,8 +52,9 @@ def mixup_batch(input: torch.Tensor,
 
     Returns:
         input_mixed (torch.Tensor): batch of inputs after mixup has been applied
-        target_mixed (torch.Tensor): labels after mixup has been applied
-        perm (torch.Tensor): the permutation used
+        target_orig (torch.Tensor): The original labels
```
dblalock, line 55: Fairly strong preference for not returning `target_orig`, since the user has to already have it right there as a variable in order to call this function. Just adds code + makes the return value even longer.

composer/algorithms/mixup/README.md
```diff
@@ -14,54 +14,98 @@ Training in this fashion improves generalization.
...
+          loss = (1 - mixing) * loss_fn(y_hat, y_orig) + mixing * loss_fn(y_hat, y_perm)
+          loss.backward()
+          opt.step()
+          opt.zero_grad()
+```
 
+Running `mixup` using dense/one-hot labels and interpolating the labels (general case)
 ```python
+# Run the mixup algorithm directly on the inputs and targets using the Composer functional API
+
```
dblalock, line 45:
```suggestion
Running `mixup` using dense/one-hot labels and interpolating the labels (general case):
```python
```

composer/algorithms/mixup/README.md
```diff
@@ -14,54 +14,98 @@ Training in this fashion improves generalization.
...
-<!--## How to Use
+## How to Use
 
 ### Functional Interface
 
-TODO(CORY): FIX
+Running `mixup` using index labels and interpolating the loss (a trick when using cross entropy)
+```python
+# Run the mixup algorithm directly on the inputs and targets using the Composer functional API
+
```
dblalock, line 24:
```suggestion
Running `mixup` using index labels and interpolating the loss (a trick when using cross entropy):
```python
```

composer/algorithms/mixup/README.md
```diff
@@ -14,54 +14,98 @@ Training in this fashion improves generalization.
...
 
 trainer.fit()
 ```
 
 ### Implementation Details
 
-TODO(CORY): Briefly describe how this is implemented under the hood in Composer.-->
+Composer's MixUp has two modes of use. When using `interpolate_loss=True` MixUp does not directly interpolate the targets, but rather interpolates the loss function by calling it on the original target and the target of the mixed in examples and interpolating the two loss values. For loss functions that are linear in the targets (such as cross entropy) this is mathematically equivalent to interpolating the targets. Many other implementations of mixup work this way.
+
+For loss functions that are not linear in the targets, this trick may still be used but results in behavior that differs from the description in the [original paper](https://arxiv.org/abs/1710.09412). If the loss function in use can accept non-index targets (such as dense or one-hot labels) then Composer's MixUp can be used with `interpolate_loss=False` which interpolates the targets of the two samples. This gives behavior matching the usual description of MixUp.
```
dblalock, line 100: just a comma
```suggestion
For loss functions that are not linear in the targets, this trick may still be used but results in behavior that differs from the description in the [original paper](https://arxiv.org/abs/1710.09412). If the loss function in use can accept non-index targets (such as dense or one-hot labels) then Composer's MixUp can be used with `interpolate_loss=False`, which interpolates the targets of the two samples. This gives behavior matching the usual description of MixUp.
```

composer/algorithms/mixup/README.md
```diff
@@ -80,13 +124,17 @@ Doing so avoids putting additional work on the CPU (since augmentation occurs on
 > MixUp requires a small amount of additional GPU compute and memory to produce the mixed-up batch.
 > In our experiments, we have found these additional resource requirements to be negligible.
 
-> ❗ MixUp Produces a Full Distribution, Not a Target Index
+> ❗ When `interpolate_loss=False` MixUp Produces a Full Distribution, Not a Target Index
 >
 > Many classification tasks represent the target value using the index of the target value rather than the full one-hot encoding of the label value.
-> Since MixUp interpolates between two target values for each example, it must represent the final targets as a dense distribution.
+> With `interpolate_loss=False` MixUp interpolates between two target values for each example, it must represent the final targets as a dense distribution.
```
dblalock, line 130: Also distribution -> vector of probabilities in the sentence right after this, and possibly elsewhere in the method card. Or define what we mean by "dense distribution" somewhere. I can ready this as referring to, e.g., a Gaussian, a Uniform, or basically any other continuous distribution, which makes this text seem confusing.

```suggestion
> With `interpolate_loss=False`, MixUp interpolates between two target values for each example, so it must represent the final targets as a dense vector of probabilities.
```

composer/algorithms/mixup/README.md
```diff
@@ -80,13 +124,17 @@ Doing so avoids putting additional work on the CPU (since augmentation occurs on
...
 > Our implementation of MixUp turns each label into a dense distribution (if it has not already been converted into a distribution).
 > The loss function used for the model must be able to accept this dense distribution as the target.
 
+>❗ When `interpolate_loss=True` MixUp interpolates the loss rather than the targets.
+
+> This is fine for loss functions that are linear in the targets, such as cross entropy, but may produce unexpected results for other loss functions.
```
dblalock, line 136: Currently shows up as two different blocks [in the docs](https://mosaicml-composer--825.com.readthedocs.build/en/825/method_cards/mixup.html).
```suggestion
>❗ When `interpolate_loss=True` MixUp interpolates the loss rather than the targets.
> 
> This is fine for loss functions that are linear in the targets, such as cross entropy, but may produce unexpected results for other loss functions.
```

dblalock: also, at the bottom of the method card, should "mixup" be capitalized in the paper name?

dblalock (commented):
Just a bunch of local nits. Only substantive thing is that I don't think we should return the original targets in `mixup_batch` since the caller already has them (unless I'm missing something).

Also, sorry for the latency; I've been more AFK than I'd hoped the past couple days

composer/algorithms/mixup/mixup.py
coryMosaicML: Added a warning in the documentation of `interpolate_loss` and set the default back to `False`. I agree it makes more sense to not turn this on by default.

composer/algorithms/mixup/mixup.py
coryMosaicML: I'm not quite sure what to do here, as checking that `state.model.loss` exists and is callable isn't enough, since the loss also must operate on `(outputs, batch)` pairs, which is fairly composer specific and atypical of how loss functions are usually implemented. Open to suggestions!

composer/algorithms/mixup/mixup.py
coryMosaicML: Fixed, and broke this up into two so the error message will identify whether `input` or `target` was not as expected.

composer/algorithms/mixup/mixup.py
coryMosaicML: Switched to passing the actual input to the loss instead of an empty tensor. 

For grabbing the loss on INIT, this is mostly to avoid having to do extra checks on `state.model`, since after INIT it will sometimes be wrapped by either torch ddp or deepspeed, in which case `state.model.loss` will no longer exist. In the ddp case, things can still work if there's an explicit check for ddp, and `state.model.module.loss` is used instead. Unsure about the deepspeed case. I agree that this can/will cause problems down the line for composition, so perhaps it's worth handling each case explicitly. This discussion is also relevant for [selective backprop](https://github.com/mosaicml/composer/blob/4bcda6e50474df4073e2e69e3443bb9980482239/composer/algorithms/selective_backprop/selective_backprop.py#L222) which does something similar. What do you think?

docs/source/getting_started/welcome_tour.rst
coryMosaicML: Switched to doing it in one line.

docs/source/trainer/using_the_trainer.rst
coryMosaicML: Fixed, I hope.

composer/algorithms/mixup/mixup.py
coryMosaicML: Curious to heard @A-Jacobson's thoughts on this. I don't have a strong opinion either way, so unless he does, I'll change it.

composer/algorithms/mixup/README.md
coryMosaicML: Fixed.

composer/algorithms/mixup/README.md
coryMosaicML: Fixed.

composer/algorithms/mixup/README.md
coryMosaicML: Cleaned up the language here according to your suggestions.

composer/algorithms/mixup/README.md
coryMosaicML: Fixed the the blocking here. 

For the name, the original paper used "mixup" in all lower case, so for the citation I left it as it appears in the paper. Is that what you meant?

dblalock: yes; sgtm

composer/algorithms/mixup/mixup.py
dblalock: ah; then I guess either way seems fine. Current version will fail fast based on type, while my version will throw lazily when loss() gets called with unexpected inputs. Happy with whatever you prefer.

composer/algorithms/mixup/mixup.py
dblalock: hmm...you're right, that's tricky. @jbloxham, when using deepspeed, how would one access the ComposerModel?

I'm leaning towards just `try...excepting` access to `state.model.loss`, `state.module.module.loss`, and whatever deepspeed does if necessary. IMO not supporting composition with this right now is taking on tech debt that we'll have to refactor or work around at some point.

coryMosaicML: Thinking about this a bit more, I don't think it's currently possible for the training loss function to be different from `state.model.loss` at INIT since the loss that actually gets computed by the trainer comes from `self.original_model` which is identical to `state.model` at INIT, and cannot be changed by algorithms. So, the function mixup grabs will always be the one used to compute `state.loss` unless something changes with the trainer. There can be composition issues if other algorithms try to modify the loss the way mixup does here, but that should be able to be handled by setting the proper algorithm order [here](https://github.com/mosaicml/composer/blob/4bcda6e50474df4073e2e69e3443bb9980482239/composer/core/engine.py#L231) when possible (e.g. adding a label smoothing term after mixup), or raising the appropriate error when composition cannot work as intended (e.g. cutmix + mixup won't compose correctly in any order if cutmix were also implemented this way).

So, as is, things should be ok, but the trainer should probably be changed such that it is possible for algorithms to modify the loss.  I think we should use the more future-proof way of getting the loss function right before the mixup loss is calculated, if it's possible with deepspeed.

jbloxham: Pretty sure `state.model.module.loss` also works for DeepSpeed.

Also note this PRD, which would provide a much more direct way of accessing the loss function if implemented: https://www.notion.so/457ebaced580405c9e87eb0032f034f7?v=ff1ae88e75294fcaa796d004d7c06504&p=c4c247ee219a4e238b50ee4cca8201f4

coryMosaicML: Ok, made this change.

composer/algorithms/mixup/mixup.py
coryMosaicML: This got fixed as part of moving things out of INIT.

dblalock (approved):
LGTM. Except that I still don't think mixup_batch should return the input.

composer/algorithms/mixup/mixup.py
coryMosaicML: Ok, changed it so `mixup_batch` no longer returns the original targets.

PR #827, opened 2022-03-25
Filepath formatting helper utilities

ravi-mosaicml:
Added filepath formatting helper utilities.

These helpers allow for formatting a filepath with distributed variables and timing information.


composer/utils/file_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+    return any(str(name).endswith(x) for x in (".tar", ".tgz", ".tar.gz", ".tar.bz2", ".tar.lzma"))
+
+
+def ensure_folder_is_empty(folder_name: Union[str, pathlib.Path]):
+    """Ensure that ``folder_name`` is empty.
+
+    Hidden files and folders (those beginning with ``.``) and ignored. Sub-folders are checked recursively.
+
+    Args:
+        folder_name (str | pathlib.Path): The folder to create or to ensure that it is empty.
```
jbloxham, line 50: The folder _to create_? This function doesn't seem like it has anything to do with creation.

composer/utils/file_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+        name (str | pathlib.Path): The name to check.
+
+    Returns:
+        bool: Whether ``name`` is a tarball.
+    """
+    return any(str(name).endswith(x) for x in (".tar", ".tgz", ".tar.gz", ".tar.bz2", ".tar.lzma"))
+
+
+def ensure_folder_is_empty(folder_name: Union[str, pathlib.Path]):
+    """Ensure that ``folder_name`` is empty.
```
jbloxham, line 45: This is super nitpicky, but this wording makes me think that the literal variable `folder_name` (i.e. the string) is being checked for empty-ness, not the folder itself. I think it's clear enough just to say something like "Ensure that the given folder is empty."

composer/utils/file_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+
+    Raises:
+        FileExistsError: If ``folder_name`` contains any non-hidden files, recursively.
+    """
+    for root, dirs, files in os.walk(folder_name, topdown=True):
+        # Filter out hidden folders
+        dirs[:] = (x for x in dirs if not x.startswith('.'))
+        for file in files:
+            if not file.startswith("."):
+                raise FileExistsError(f"{folder_name} is not empty; {os.path.join(root, file)} exists.")
```
jbloxham, line 60: This function doesn't look recursive.

composer/utils/file_helpers.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+def format_name_with_dist(format_str: str, run_name: str, **extra_format_kwargs: object):
```
jbloxham, line 63: Is `object` the preferred type annotation for kwargs?

jbloxham (approved):

composer/utils/file_helpers.py
ravi-mosaicml: `os.walk` is recursive

composer/utils/file_helpers.py
ravi-mosaicml: Good catch, refactored this function multiple times and forgot to change this.

composer/utils/file_helpers.py
ravi-mosaicml: See https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

Because the kwargs are passed into .format(), the type can be anything, so typing as object here.

composer/utils/file_helpers.py
jbloxham: Ah, I see. Well in that case I'm pretty sure this function wouldn't behave properly for non-hidden files belonging to hidden directories. The `dirs[:] = ...` line doesn't actually do anything.

ravi-mosaicml: It should -- see https://docs.python.org/3/library/os.html#os.walk.

> When topdown is True, the caller can modify the dirnames list in-place (perhaps using [del](https://docs.python.org/3/reference/simple_stmts.html#del) or slice assignment), and [walk()](https://docs.python.org/3/library/os.html#os.walk) will only recurse into the subdirectories whose names remain in dirnames; this can be used to prune the search, impose a specific order of visiting, or even to inform [walk()](https://docs.python.org/3/library/os.html#os.walk) about directories the caller creates or renames before it resumes [walk()](https://docs.python.org/3/library/os.html#os.walk) again. Modifying dirnames when topdown is False has no effect on the behavior of the walk, because in bottom-up mode the directories in dirnames are generated before dirpath itself is generated.

jbloxham: Oh that's so jank but yeah that works.

PR #828, opened 2022-03-25
Profiler Refactor for Artifact Logging

ravi-mosaicml:
* Refactored the Profiler to use the artifact logging API
* Storing seperate trace files at the end of each cycle, rather than merging all trace files together, for better performance. Left trace merging available as an optional flag.
* Seperated the profiling scheduling out from the profiler, to allow for custom scheduling functions
* Fixed the SystemProfiler so it shuts down at the end of training

The diff appears large because of renaming -- Git did not pick up files that were renamed and subsequently modified.


composer/profiler/torch_profiler.py
```diff
@@ -54,84 +53,242 @@ class TorchProfiler(Callback):
...
+            +------------------------+-------------------------------------------------------+
+            | ``{world_size}``       | The world size, as returned by                        |
+            |                        | :func:`~composer.utils.dist.get_world_size`.          |
+            +------------------------+-------------------------------------------------------+
+            | ``{local_world_size}`` | The local world size, as returned by                  |
+            |                        | :func:`~composer.utils.dist.get_local_world_size`.    |
+            +------------------------+-------------------------------------------------------+
+            | ``{node_rank}``        | The node rank, as returned by                         |
+            |                        | :func:`~composer.utils.dist.get_node_rank`.           |
+            +------------------------+-------------------------------------------------------+
```
bandish-shah, line 81: [nit] Could we use an f-string or substitution here to avoid copying this table from `composer.utils.file_helpers`?

composer/profiler/torch_profiler.py
```diff
@@ -54,84 +53,242 @@ class TorchProfiler(Callback):
...
+            | ``{sample_in_epoch}``  | The sample count in the current epoch, as returned by |
+            |                        | :meth:`~composer.core.time.Timer.sample_in_epoch`.    |
+            +------------------------+-------------------------------------------------------+
+            | ``{token}``            | The total token count, as returned by                 |
+            |                        | :meth:`~composer.core.time.Timer.token`.              |
+            +------------------------+-------------------------------------------------------+
+            | ``{token_in_epoch}``   | The token count in the current epoch, as returned by  |
+            |                        | :meth:`~composer.core.time.Timer.token_in_epoch`.     |
+            +------------------------+-------------------------------------------------------+
```
bandish-shah, line 137: [nit] Could we use an f-string or substitution here to avoid copying this table from `composer.utils.file_helpers`?

bandish-shah (approved):
LGTM, some very minor docstring [nit]'s.  Your call on if you want to address or not.

PR #829, opened 2022-03-25
Serialize model state without `module.` prefix when using DDP

ravi-mosaicml:
This PR strips the `module.` prefix from the checkpoint state when saving DDP checkpoints. Before, the `module.` prefix was consumed when loading checkpoints.

This change should improve usability, as `state_dict['state']['model']` reflects the original model that was passed to the trainer. The user should not have to worry about DDP wrapping.


abhi-mosaic (approved):
LGTM! Do we have some tests for the new behavior, or would it be a good idea for me to check out this branch and test @ravi-mosaicml ?

ravi-mosaicml:
> LGTM! Do we have some tests for the new behavior, or would it be a good idea for me to check out this branch and test @ravi-mosaicml ?

We have checkpointing tests, but we do not have tests to ensure that checkpoints are backwards compatible. We should add some though (would need to generate and save test fixtures).

ravi-mosaicml:
Red-button merging as tests passed and the dev merge is clean.

Issue #830, opened 2022-03-25
Callbacks need access to computed metrics


hanlint:
Callbacks such as `EarlyStopping` would need access to the metric results. Currently, we compute the metrics, log the results, and then reset the metrics for the next epoch. Thus, simply attaching the metrics to state would not address this issue; we need a design where callbacks can access this time series data.

PR #831, opened 2022-03-25
Add MLPerf logging

hanlint:
This PR contributes a callback `MLPerfCallback`, which will create a submission directory and results file that is compliant with the MLPerf rules for Training v1.1 (e.g. it passes the mlperf logging's package checker).

Upon usage, a submission folder structure will be created with the ``root_folder`` as the base and the following directories:
```
    root_folder/
        results/
            [system_name]/
                [benchmark]/
                    results_0.txt
                    results_1.txt
                    ...
        systems/
            [system_name].json
```

For each training run, a results file will be created, e.g.
```
:::MLLOG {"namespace": "", "time_ms": 1648183840286, "event_type": "INTERVAL_START", "key": "cache_clear", "value": null, "metadata": {"file": "/composer/composer/callbacks/mlperf.py", "lineno": 109}}
:::MLLOG {"namespace": "", "time_ms": 1648183840438, "event_type": "INTERVAL_START", "key": "init_start", "value": null, "metadata": {"file": "/composer/composer/callbacks/mlperf.py", "lineno": 110}}
...
```
The entire directory can then be submitted to mlperf, and checked with the `package_checker` in https://github.com/mlcommons/logging.

Currently this callback only supports the OPEN division benchmark.

This PR is gated by:

- [ ]  https://github.com/mlcommons/logging/issues/218 
- [x] https://github.com/mosaicml/composer/issues/830

TODO:
- [ ] generate and validate ResNet-50 run on 8x A100 GPUs.
- [x] create `Hparams` object


composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+        if rank_zero():
+            if state.train_dataloader.batch_size is None:
+                raise ValueError("Batch size is required to be set for dataloader.")
+
+            self._log_dict({
+                constants.SEED: state.seed,
+                constants.GLOBAL_BATCH_SIZE: state.train_dataloader.batch_size * dist.get_world_size(),
+                constants.GRADIENT_ACCUMULATION_STEPS: state.grad_accum,
+                constants.TRAIN_SAMPLES: len(state.train_dataloader.dataset),
+                constants.EVAL_SAMPLES: len(state.evaluators[0].dataloader.dataloader.dataset)
```
ravi-mosaicml, line 165: Can we add a check to ensure that there is just one evaluator?

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+        if rank_zero():
+            self.mllogger.event(key=constants.EPOCH_START, metadata={'epoch_num': state.timer.epoch.value})
+            self.mllogger.event(key=constants.BLOCK_START,
+                                metadata={
+                                    'first_epoch_num': state.timer.epoch.value,
+                                    'epoch_count': 1
+                                })
+
+    def epoch_end(self, state: State, logger: Logger) -> None:
+        if rank_zero():
```
ravi-mosaicml, line 184: is it OK if every rank logs, and then we only submit the rank zero folder?

setup.py
```diff
@@ -151,7 +153,12 @@ def package_files(prefix: str, directory: str, extension: str):
     "wurlitzer>=3.0.2,<4",
 ]
 
-extra_deps["all"] = set(dep for deps in extra_deps.values() for dep in deps)
+extra_deps['mlperf'] = [
+    # TODO: switch to pip package when available: https://github.com/mlcommons/logging/issues/218
+    'git+https://github.com/mlperf/logging.git',
```
ravi-mosaicml, line 158: (Creating a comment to track this TODO -- cannot merge this PR until this is resolved, as this will break pip builds.)

tests/callbacks/test_mlperf_callback.py
```diff
@@ -0,0 +1,65 @@
...
+    check_training_package(
+        folder=tmpdir,
+        usage="training",
+        ruleset="1.1.0",
+        werror=True,
+        quiet=False,
+        rcp_bypass=False,
+        rcp_bert_train_samples=False,
+        log_output="package_checker.log",
+    )
```
ravi-mosaicml, line 65: newline

tests/callbacks/test_mlperf_callback.py
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+
+@pytest.mark.filterwarnings(
+    "ignore::DeprecationWarning",)
+def test_mlperf_callback(config, tmpdir):
+    tmpdir = 'mlperf_results'
+    pytest.importorskip("mlperf_logging")
+
+    for run in range(5):
```
ravi-mosaicml, line 46: curious about why 5 runs?

tests/callbacks/test_mlperf_callback.py
```diff
@@ -0,0 +1,65 @@
...
+        'deterministic_mode': True,  # testing equivalence
+        'loggers': [],  # no progress bar
+        'callbacks': []
+    }
+
+
+@pytest.mark.filterwarnings(
+    "ignore::DeprecationWarning",)
+def test_mlperf_callback(config, tmpdir):
+    tmpdir = 'mlperf_results'
```
ravi-mosaicml, line 43: Can we use the pytest tmpdir?

```suggestion
```

tests/callbacks/test_mlperf_callback.py
```diff
@@ -0,0 +1,65 @@
...
+from composer import Trainer
+from composer.callbacks import MLPerfCallback
+from tests.common import RandomClassificationDataset, SimpleModel
+
+logging.basicConfig(filename="/Users/hanlintang/composer/package_checker.log", level=logging.INFO)
+logging.getLogger().addHandler(logging.StreamHandler())
+formatter = logging.Formatter("%(levelname)s - %(message)s")
+logging.getLogger().handlers[0].setFormatter(formatter)
+logging.getLogger().handlers[1].setFormatter(formatter)
+
```
ravi-mosaicml, line 16: Can we add in a test to ensure the right things are logged? Thinking that we could monkeypatch-and-capture the `_log_dict` method to ensure that the expected calls are made (or do something similar there).

tests/callbacks/test_mlperf_callback.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+import pytest
+from torch.utils.data import DataLoader
+
+from composer import Trainer
+from composer.callbacks import MLPerfCallback
+from tests.common import RandomClassificationDataset, SimpleModel
+
+logging.basicConfig(filename="/Users/hanlintang/composer/package_checker.log", level=logging.INFO)
```
ravi-mosaicml, line 11: Pytest has built-in log capturing so I don't think we should need to configure this explicitly? We might need to adjust what packages / loglevels are being logged tho.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+    Args:
+        submitter (str): Name of the submitter organization
+        division (str): Submission division (open, closed)
+        status (str): system status (cloud, onprem, preview)
+
+    Returns:
+        system description as a dictionary
+    """
+    is_cuda = torch.cuda.is_available()
+    cpu_info = cpuinfo.get_cpu_info()
```
ravi-mosaicml, line 228: Is this correct on K8S? or is it somehow masked?

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+
+def rank_zero() -> bool:
+    return dist.get_global_rank() == 0
+
+
+class MLPerfCallback(Callback):
+    """Creates a compliant results file for MLPerf Training benchmark.
+
+    A submission folder structure will be created with the ``root_folder``
+    as the base and the following directories:
```
ravi-mosaicml, line 39: Preformat the folder structure.

```suggestion
    as the base and the following directories::
```

ravi-mosaicml (commented):
Overall looks good; left some comments for now. Will re-review once the open TODOs are addressed.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+from composer.utils import dist
+
+try:
+    from mlperf_logging import mllog
+    from mlperf_logging.mllog import constants
+    mlperf_available = True
+except ImportError:
+    mlperf_available = False
+
+BENCHMARKS = ("resnet")
```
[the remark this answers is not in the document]
hanlint: These are the set of currently supported configurations by the callback, so they need to be constants.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+DIVISIONS = ("open")
+STATUS = ("onprem", "cloud", "preview")
+
+
+def rank_zero() -> bool:
```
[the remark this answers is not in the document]
hanlint: Prefer to keep it shorter for brevity, the `global` is assumed.

composer/callbacks/mlperf.py
hanlint: The MLperf logger logs from multiple ranks into the same file, specifcally the cache_clear, init_start, and init_end events -- there should be as many of these for each rank. For an example results file, see [NVIDIA submission](https://github.com/mlcommons/training_results_v1.1/blob/main/NVIDIA/results/dgxa100_n128_ngc21.09_pytorch/bert/result_0.txt).

tests/callbacks/test_mlperf_callback.py
hanlint: The MLperf logging package uses `logging`, so these need to be configured. However, we should remove the hardcoded path there.

tests/callbacks/test_mlperf_callback.py
hanlint: Is it sufficient that we used the `package_checker` below from MLPerf, which is what validates that the log is compliant?

tests/callbacks/test_mlperf_callback.py
hanlint: To pass the compliance checker, we need to submit 5 consecutive runs of the system. See MLcommons Benchmark rules [here](https://github.com/mlcommons/training_policies/blob/master/training_rules.adoc#benchmark-results).

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,285 @@
...
+            Default: ``"onprem"``.
+        target (float, optional): The target metric before the mllogger marks the stop
+            of the timing run. Default: ``0.759`` (resnet benchmark).
+    """
+
+    def __init__(
+        self,
+        root_folder: str,
+        index: int,
+        submitter: str = "MosaicML",
```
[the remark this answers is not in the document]
hanlint: Since the callback currently only supports a specific scenario, I think having them default to the resnet50 benchmark makes the most sense?

ravi-mosaicml (commented):
Overall looks good; main comments re: adding in artifact logging and fixing potential multi-rank race conditions.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+    cpu_info = cpuinfo.get_cpu_info()
+
+    system_desc = {
+        "submitter": submitter,
+        "division": division,
+        "status": status,
+        "number_of_nodes": dist.get_world_size() / dist.get_local_world_size(),
+        "host_processors_per_node": "",
+        "host_processor_model_name": str(cpu_info.get('brand_raw', "CPU")),
+        "host_processor_core_count": str(psutil.cpu_count(logical=False)),
```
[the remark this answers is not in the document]
hanlint: Yes this would be incorrect on k8s, but runs will be occupy the entire node. This is meant to be best effort for the general case by users (see disclaimer) so need not be accurate in every circumstance.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+            self.mllogger.event(key=constants.EVAL_STOP, metadata={'epoch_num': state.timer.epoch.value})
+            self.mllogger.event(key=constants.EVAL_ACCURACY,
+                                value=accuracy,
+                                metadata={'epoch_num': state.timer.epoch.value})
+            self.mllogger.event(key=constants.BLOCK_STOP, metadata={'first_epoch_num': state.timer.epoch.value})
+
+            if accuracy > self.target and not self.success:
+                self.mllogger.event(key=constants.RUN_STOP, metadata={"status": "success"})
+                self.mllogger.logger.removeHandler(self._file_handler)
+                self.success = True  # only log once
```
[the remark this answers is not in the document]
hanlint: No, training should continue, users usually want to see whehter accuracy have improved even further post-target to check if there is room for additional hyperparameter tuning.

composer/callbacks/mlperf.py
hanlint: The NV/Google submission code logs directly to the same file, leading me to believe that this is unneeded code complexity, probably because only <8 entries are concurrently written. Let's keep as-is and can fix later if we run into an issue.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+STATUS = ("onprem", "cloud", "preview")
+
+
+def rank_zero() -> bool:
+    return dist.get_global_rank() == 0
+
+
+def require_mlperf_logging():
+    if not mlperf_available:
+        raise ImportError("""Please install with pip install mosaicml[mlperf] and also
```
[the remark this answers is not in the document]
hanlint: That wouldn't work here -- the mlperf logger install is non a pip package nor available on conda.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # save system description file
+        systems_path = os.path.join(root_folder, 'systems', f'{system_name}.json')
+        if os.path.exists(systems_path):
+            with open(systems_path, 'r') as f:
+                existing_systems_desc = json.load(f)
+                if sorted(existing_systems_desc.items()) != sorted(system_desc.items()):
+                    raise ValueError(f'Existing system description in {systems_path} does not match this machine.')
+        else:
+            with open(systems_path, 'w') as f:
```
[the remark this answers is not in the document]
hanlint: made this a local rank 0 only

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+                json.dump(system_desc, f, indent=4)
+
+        filename = os.path.join(root_folder, 'results', system_name, benchmark, f'result_{index}.txt')
+        if os.path.exists(filename):
```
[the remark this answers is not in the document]
hanlint: The log filename and the system description name are different, and the creation is handled by the logger, so this needs to be explicitly checked.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+            raise FileExistsError(f'{filename} already exists.')
+
+        self._file_handler = logging.FileHandler(filename)
+        self._file_handler.setLevel(logging.INFO)
+        self.mllogger.logger.addHandler(self._file_handler)
+
+        # TODO: implement cache clearing
```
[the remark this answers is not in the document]
hanlint: Ah, good catch. this is a TODO for this PR.

composer/callbacks/mlperf.py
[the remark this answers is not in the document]
hanlint: Went back and forth on this a bit, and decided to keep the defaults, for ease of use.

composer/callbacks/callback_hparams.py
```diff
@@ -119,10 +121,36 @@ def initialize_object(self) -> SpeedMonitor:
         return SpeedMonitor(window_size=self.window_size)
 
 
+@dataclass
+class MLPerfCallbackHparams(CallbackHparams):
+    """:class:`~.MLPerfCallback` hyperparameters.
```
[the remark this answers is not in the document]
hanlint: if the arguments are the same, why duplicate?

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,314 @@
...
+        self.mllogger = mllog.get_mllogger()
+        self.target = target
+        self.system_name = system_name
+        self.benchmark = benchmark
+        self.root_folder = root_folder
+
+        system_desc = get_system_description(submitter, division, status, system_name)
+        system_name = system_desc['system_name']
+
+        self._create_submission_folders(root_folder, system_name, benchmark)
```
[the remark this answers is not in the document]
hanlint: Decided not to, as doing so would require persisting extraneous variables in the object (submitter, division, etc.), hurts ability to debug (cannot call without having a state and loggers to provide), and these methods do not require accessing state or loggers.

hanlint:
Thanks for the comments, this is ready for another review! I did not accept all the suggested changes, so please take a look. A few opens:
* Not sure I implemented `file_artifact` properly
* since cache clearing usually done with `sudo /sbin/sysctl vm.drop_caches=3`, is system-specific, I decided to accept a `cache_clear_cmd` that is run with `os.system` at the appropriate time. However, this seems dangerous to accept running arbitrary commands; I am open to other suggestions.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+        if dist.get_local_rank() == 0:
+            self._create_submission_folders(root_folder, system_name, benchmark)
+            systems_path = os.path.join(root_folder, 'systems', f'{system_name}.json')
+            with open(systems_path, 'w') as f:
+                json.dump(system_desc, f, indent=4)
+
+        dist.barrier()
+
+        self.filename = os.path.join(root_folder, 'results', system_name, benchmark, f'result_{index}.txt')
+        self.upload_name = '{run_name}' + f'/results/{system_name}/{benchmark}/result_{index}.txt'
```
ravi-mosaicml, line 158: Can this include the node_rank or global_rank? Otherwise, if one is doing multi-node training, the nodes will override each-other.

ravi-mosaicml (commented):
Overall looks good; see comment re:

1. Logging the system config as an artifact and including the node rank or global rank as part of the artifact name. We allow the artifact name to be specified as part of the `__init__` in other callbacks, but I'll leave that up to you here.
2. Removing the kwargs from the tests.
3. Requiring that the root folder is empty, or emptying it if it exists (and mentioning so in the docstring)

I think this should otherwise be good to merge!

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+
+        self.mllogger = mllog.get_mllogger()
+        self.target = target
+        self.system_name = system_name
+        self.benchmark = benchmark
+        self.root_folder = root_folder
+        self.metric_name = metric_name
+        self.metric_label = metric_label
+
+        system_desc = get_system_description(submitter, division, status, system_name)
```
ravi-mosaicml, line 146: Can this and everything below be moved to run on `Event.INIT` rather than on `__init__`? It is unknown when the callback will be constructed (and could be before distributed is initialized).

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+
+        if os.path.exists(self.filename):
+            raise FileExistsError(f'{self.filename} already exists.')
+
+        self._file_handler = logging.FileHandler(self.filename)
+        self._file_handler.setLevel(logging.INFO)
+        self.mllogger.logger.addHandler(self._file_handler)
+
+        if cache_clear_cmd is not None:
+            os.system(cache_clear_cmd)
```
ravi-mosaicml, line 168: Would it be possible to use `subprocess.run(cache_clear_cmd, check=True, text=True)`? `os.system` does not include the checks that subprocess does.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+            self.mllogger.event(key=constants.EVAL_ACCURACY,
+                                value=accuracy,
+                                metadata={'epoch_num': state.timer.epoch.value})
+            self.mllogger.event(key=constants.BLOCK_STOP, metadata={'first_epoch_num': state.timer.epoch.value})
+
+            if accuracy > self.target and not self.success:
+                self.mllogger.event(key=constants.RUN_STOP, metadata={"status": "success"})
+                self.mllogger.logger.removeHandler(self._file_handler)
+                self.success = True  # only log once
+
```
ravi-mosaicml, line 263: Can the callback implement `def close(self, state, logger)` and ensure that any opened files are closed (such as the `self._file_handler`)? The trainer will invoke this when shutting down to ensure that open file handlers are not being left behind. It is OK to close the file earlier; but we should also close it here (if the file is still open) in case if there was some exception somewhere.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+        division (str, optional): Division of submission. Currently only ``open`` division supported.
+        metric_name (str, optional): name of the metric to compare against the target. Default: ``Accuracy``.
+        metric_label (str, optional): label name. The metric will be accessed via ``state.current_metrics[metric_label][metric_name]``.
+        submitter (str, optional): Submitting organization. Default: MosaicML.
+        system_name (str, optional): Name of the system (e.g. 8xA100_composer). If
+            not provided, system name will default to ``[world_size]x[device_name]_composer``,
+            e.g. ``8xNVIDIA_A100_80GB_composer``.
+        status (str, optional): Submission status. One of (onprem, cloud, or preview).
+            Default: ``"onprem"``.
+        cache_clear_cmd (str, optional): Command to invoke during the cache clear. This callback
```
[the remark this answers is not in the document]
hanlint: The issue is that if you running inside a docker, the command would fail because it requires privileged dockers, making it difficult to test (and run). I'll keep it as a string for now, until we arrive at some solution.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+                constants.SUBMISSION_DIVISION: division,
+                constants.SUBMISSION_ORG: submitter,
+                constants.SUBMISSION_PLATFORM: system_name,
+                constants.SUBMISSION_STATUS: status,
+            })
+
+        self.success = False
+
+    def _create_submission_folders(self, root_folder: str, system_name: str, benchmark: str):
+        os.makedirs(root_folder, exist_ok=True)
```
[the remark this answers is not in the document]
hanlint: MLperf submissions require multiple training runs (and therefore results files), which is why the test runs it for `n=5` times, resulting in a root_folder with `result_0`, `result_1`, etc. So this is intentional by design. We do, however, throw an error if `result_0` already exists and someone is looking to submit with `index=0`.

composer/callbacks/mlperf.py
hanlint: We cannot -- the naming for the mlperf submission package is fairly strict. Upload is only happening on global rank zero so this is not an issue.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+        self.metric_name = metric_name
+        self.metric_label = metric_label
+
+        system_desc = get_system_description(submitter, division, status, system_name)
+        system_name = system_desc['system_name']
+
+        if dist.get_local_rank() == 0:
+            self._create_submission_folders(root_folder, system_name, benchmark)
+            systems_path = os.path.join(root_folder, 'systems', f'{system_name}.json')
+            with open(systems_path, 'w') as f:
```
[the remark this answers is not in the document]
hanlint: Done. Filename restrictions do not allow for rank specifications, but this is guarded by only global rank zero.

composer/callbacks/mlperf.py
hanlint: Thanks for the suggestion -- not handling this for reasons described previously (e.g. debuggability and readability). Intended use should be with `composer` command.

composer/callbacks/mlperf.py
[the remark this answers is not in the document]
ravi-mosaicml: Sorry, I don't see where this is being uploaded. (The check above is for the local rank, not the global rank, so still a bit confused).

composer/callbacks/mlperf.py
ravi-mosaicml: Distributed isn't configured until the trainer is constructed, so I'm pretty sure the `dist.barrier()` below will error out outside of tests. (In tests, we configure distributed _before_ any tests are run, so that is why they would be passing on Jenkins.) Can you please move where this code is run to Event.INIT?

composer/callbacks/mlperf.py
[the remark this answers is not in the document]
ravi-mosaicml: Hmm OK, but I guess I don't see why this precludes us from hard-coding the command? We can default it to be False (to match the current behavior of the default being an empty string).

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+
+    .. note::
+
+        MLPerf submissions require clearing the system cache prior to any training run.
+        By default, this callback does not clear the cache, as that is a system specific
+        operation. To enable cache clearing, and thus pass the mlperf compliance checker,
+        provide a ``cache_clear_cmd`` that will be executed with ``os.system``.
+
+    Args:
+        root_folder (str): The root submission folder
```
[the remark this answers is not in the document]
ravi-mosaicml: (I see; resolved in another comment).

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,345 @@
...
+            self.mllogger.event(key=constants.BLOCK_START,
+                                metadata={
+                                    'first_epoch_num': state.timer.epoch.value,
+                                    'epoch_count': 1
+                                })
+
+    def epoch_end(self, state: State, logger: Logger) -> None:
+        if rank_zero():
+            self.mllogger.event(key=constants.EPOCH_STOP, metadata={'epoch_num': state.timer.epoch.value})
+            logger.file_artifact(LogLevel.FIT, artifact_name=self.upload_name, file_path=self.filename)
```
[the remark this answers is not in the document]
ravi-mosaicml: (Bumping this?)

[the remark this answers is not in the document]
hanlint: Adding to `close`. We don't need to force the flush.

composer/callbacks/mlperf.py
hanlint: done

Makefile
```diff
@@ -11,6 +11,8 @@ EXTRA_LAUNCHER_ARGS ?= # extra arguments for the composer cli launcher
 # Force append the duration flag to extra args
 override EXTRA_ARGS += --duration $(DURATION)
 
+dirs := composer examples tests
```
ravi-mosaicml, line 14: I think #987 resolved these issues? I presume this is leftover from the merge?

composer/core/state.py
```diff
@@ -77,7 +78,13 @@ class State(Serializable):
             ``rank_zero_seed + dist.get_global_rank()``.
         grad_accum (int, optional): The number of gradient accumulation steps to use. With this argument, micro batch size for
             each device becomes ``microbatch_size = train_batch_size / (num_devices * grad_accum)``.
+<<<<<<< HEAD
```
ravi-mosaicml, line 81: merge leftovers?

composer/core/state.py
```diff
@@ -217,6 +231,9 @@ def __init__(
         self.set_dataloader(dataloader, dataloader_label, dataloader_len)
         self.max_duration = max_duration
 
+        self.train_dataloader = train_dataloader
+        self.evaluators = list(ensure_tuple(evaluators))
```
ravi-mosaicml, line 235: Similar to how the optimizers and schedulers are handled, can we use a private getter / setter that does an in-place modification of the list? This is mainly to ensure that any code like `evaluators = state.evaluators` always points to the latest evaluators.

composer/trainer/trainer.py
```diff
@@ -756,6 +774,8 @@ def __init__(
         self.state = State(
             rank_zero_seed=rank_zero_seed,
             algorithms=algorithms,
+            train_dataloader=train_dataloader.dataloader,
```
ravi-mosaicml, line 777: Instead of binding the the `train_dataloader` and `evaluators` to the `State` before `Event.INIT`, would it be possible set these attributes after this event?

https://github.com/mosaicml/composer/blob/23f888e07456d31a2950db9f962f4dbdf41938fb/composer/trainer/trainer.py#L899-L901

The rational is it will be valid to pass it data loaders and evaluators on either `Trainer.__init__` or `Trainer.fit()`, so code should (algorithms and callbacks) should not depend on them being specified on `Event.INIT`.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,377 @@
...
+            if not isinstance(state.evaluators[0].dataloader.dataloader, DataLoader):
+                raise ValueError("eval dataset must be a torch dataloader.")
+            if state.train_dataloader.batch_size is None:
+                raise ValueError("Batch size is required to be set for dataloader.")
+            if len(state.evaluators) > 1:
+                raise ValueError("Only one evaluator is supported for the MLPerfCallback.")
+            if not isinstance(state.train_dataloader.dataset, Sized):
+                raise ValueError("Train dataset must have __len__ property")
+            if not isinstance(state.evaluators[0].dataloader.dataloader.dataset, Sized):
+                raise ValueError("Train dataset must have __len__ property")
```
ravi-mosaicml, line 240:
```suggestion
                raise TypeError("The eval dataset must have __len__ property")
```

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,377 @@
...
+    def _get_accuracy(self, state: State):
+        if self.metric_name not in state.current_metrics[self.metric_label]:
+            raise ValueError('Accuracy must be a validation metric.')
+        return state.current_metrics[self.metric_label][self.metric_name]
+
+    def fit_start(self, state: State, logger: Logger) -> None:
+        if rank_zero():
+
+            if not isinstance(state.train_dataloader, DataLoader):
+                raise ValueError("train dataloader must be a torch dataloader")
```
ravi-mosaicml, line 230:
```suggestion
                raise TypeError("train dataloader must be a torch dataloader")
```

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,377 @@
...
+            if not isinstance(state.evaluators[0].dataloader.dataloader, DataLoader):
+                raise ValueError("eval dataset must be a torch dataloader.")
```
ravi-mosaicml, line 232:
```suggestion
                raise TypeError("eval dataset must be a torch dataloader.")
```

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,377 @@
...
+            if state.train_dataloader.batch_size is None:
+                raise ValueError("Batch size is required to be set for dataloader.")
+            if len(state.evaluators) > 1:
+                raise ValueError("Only one evaluator is supported for the MLPerfCallback.")
+            if not isinstance(state.train_dataloader.dataset, Sized):
+                raise ValueError("Train dataset must have __len__ property")
```
ravi-mosaicml, line 238:
```suggestion
                raise TypeError("Train dataset must have __len__ property")
```

composer/trainer/trainer.py
```diff
@@ -722,6 +722,24 @@ def __init__(
                     To fix, please do not iterate over the dataloader before passing it into
                     the trainer."""))
 
+        # Evaluators
```
ravi-mosaicml, line 725: See comment below, but is it possible to bind the evaluators to the state after Event.INIT (and leave this code where it was?)

ravi-mosaicml (commented):
Really close! Just a few nits, plus one comment re: binding the train dataloader and evaluator to the state after `Event.INIT` runs. Rational is that these parameters will not need be specified in `Trainer.init(...)` after #948 lands (and algorithms / callbacks are already OK with no dataloaders or evaluators during `Event.INIT`).

composer/callbacks/callback_hparams.py
```diff
@@ -119,10 +120,65 @@ def initialize_object(self) -> SpeedMonitor:
...
+            will call ``subprocess(cache_clear_cmd)``. Default is disabled (None)
+
+    """
+
+    root_folder: str = hp.required("The root submission folder.")
+    index: int = hp.required("The repetition index of this run.")
+    benchmark: str = hp.optional("Benchmark name. Default: resnet", default="resnet")
+    target: float = hp.optional("The target metric before mllogger marks run_stop. Default: 0.759 (resnet)",
+                                default=0.759)
+    divison: str = hp.optional("Division of submission. Currently only open division"
```
bandish-shah, line 153: Typo:

```suggestion
    division: str = hp.optional("Division of submission. Currently only open division"
```

composer/callbacks/callback_hparams.py
```diff
@@ -119,10 +120,65 @@ def initialize_object(self) -> SpeedMonitor:
...
+                               default="open")
+    metric_name: str = hp.optional('name of the metric to compare against the target. Default: Accuracy',
+                                   default='Accuracy')
+    metric_label: str = hp.optional(
+        'label name. The metric will be accessed via state.current_metrics[metric_label][metric_name]. Default: eval',
+        default='eval')
+    submitter: str = hp.optional("Submitting organization. Default: MosaicML", default='MosaicML')
+    system_name: str = hp.optional("Name of the system, defaults to [world_size]x[device_name]", default=None)
+    status: str = hp.optional("Submission status. Default: onprem", default="onprem")
+    cache_clear_cmd: str = hp.optional(
```
bandish-shah, line 164:
```suggestion
    cache_clear_cmd: Optional[str] = hp.optional(
```

composer/core/state.py
```diff
@@ -317,6 +331,14 @@ def algorithms(self):
     def algorithms(self, algorithms: Sequence[Algorithm]):
         self._algorithms[:] = algorithms
 
+    @property
+    def evaluators(self):
+        return self._evaluators
```
ravi-mosaicml, line 336: nit: docstring

composer/trainer/trainer.py
ravi-mosaicml: Can these two lines be removed? I see they are being set below.

composer/callbacks/mlperf.py
```diff
@@ -0,0 +1,377 @@
...
+            self._log_dict({
+                constants.SEED: state.seed,
+                constants.GLOBAL_BATCH_SIZE: state.train_dataloader.batch_size * dist.get_world_size(),
+                constants.GRADIENT_ACCUMULATION_STEPS: state.grad_accum,
+                constants.TRAIN_SAMPLES: len(state.train_dataloader.dataset),
+                constants.EVAL_SAMPLES: len(state.evaluators[0].dataloader.dataloader.dataset)
+            })
+
+            # optionally, upload the system description file
+            logger.file_artifact(LogLevel.FIT, self.system_desc_upload_name, self.systems_path)
```
ravi-mosaicml, line 251: nit: this line could be moved up to line 177 in INIT, where the file is written. It will work down here too (just slightly less intuitive to follow)

ravi-mosaicml (approved):
LGTM :100:! See comments

setup.py
```diff
@@ -153,6 +153,12 @@ def package_files(prefix: str, directory: str, extension: str):
     "wurlitzer>=3.0.2,<4",
 ]
 
+extra_deps["mlperf"] = [
+    # TODO: use pip when available: https://github.com/mlcommons/logging/issues/218
+    # "mlperf_logging @ git+https://github.com/mlperf/logging.git",
+    "py-cpuinfo>=8.0.0,<9",
```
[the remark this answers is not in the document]
bandish-shah: You actually don't need `py-cpuinfo` here anymore, added it as a base dependency as it's needed for `composer_collect_env`.

composer/core/state.py
hanlint: no docstrings exist for the other getter/setters, i think State in general needs a cleanup that I'll get to in a separate PR.

PR #832, opened 2022-03-25
[Eval-Only]: Made the `state.dataloader` optional; removed `state.ste…

### LINKED ISSUE #363: Replace `state.train_dataloader` and `state.eval_dataloader` with just `state.dataloader`
#329 will result in `state.train_dataloader` and `state.eval_dataloader` on the State. However, algorithms and callbacks should only be interested in the currently activate dataloader.

For example, callbacks should not be touching the training dataloader, as evaluation could be running mid-epoch. For callbacks that simply want to run on every batch (like TQDM), it makes the logic more complicated since they will have to check which dataloader to use.

Instead, `state.train_dataloader` and `state.eval_dataloader` can be merged into one field -- called `state.dataloader`. The trainer is responsible for setting this to the correct dataloader that is being actively iterated.
* Upon init, `state.dataloader` should be set to the training dataloader, so dataloader augmentations can be applied.
* Before `Event.EVAL_START`, the trainer should set `state.dataloader` to the dataloader of the evaluator being evaluated. After `Event.EVAL_END`, the trainer should revert `state.dataloader` back to the training dataloader (i.e. the value before swapping it at `Event.EVAL_START`).

TODO:
- [ ] @hanlint stamp of approval

ravi-mosaicml:
…ps_per_epoch`.

1. Made the `state.dataloader` optional, since it will not be provided on `__init__` as part of #40.
2. Binding the active dataloader to the state on `Event.FIT_START`, and switching the dataloader to each evaluation dataloader before `Event.EVAL_START`. Restoring the previous (training) dataloader after `Event.EVAL_END`.
3. Moved `Event.EVAL_START` and `Event.EVAL_END` to run for each evaluator, instead of once for all evaluators. With #40, the eval() will take in a dataloader, which would then require `Event.EVAL_START` and `Event.EVAL_END`. This change also permits for algorithms that wish to modify (each) evalution dataloader.
4. Moved scaling of the LR schedulers to `Trainer.fit()` before `Event.FIT_START` fires. Schedulers will be passed in on `Trainer.fit()` as part of #40.
5. Removed `steps_per_epoch` as part of the state. Instead, algorithms and callbacks can read len(state.dataloader) directly. While this change will make schedulers no longer accurate when using `train_subset_num_batches`, that flag should only be used for performance measurements. As such, it is not necessarry that SSR behaves correctly for performance runs. Added a warning for the `train_subset_num_batches` field.

Implements the first part of #40.
Closes #363.

### DIFF
--- a/composer/algorithms/augmix/augmix.py
+++ b/composer/algorithms/augmix/augmix.py
@@ -251,7 +251,10 @@ class AugMix(Algorithm):
         self._transformed_datasets = weakref.WeakSet()
 
     def match(self, event: Event, state: State) -> bool:
-        return event == Event.FIT_START and state.train_dataloader.dataset not in self._transformed_datasets
+        if event != Event.FIT_START:
+            return False
+        assert state.dataloader is not None, "dataloader should be defined on fit start"
+        return state.dataloader.dataset not in self._transformed_datasets
 
     def apply(self, event: Event, state: State, logger: Logger) -> None:
         am = AugmentAndMixTransform(severity=self.severity,
@@ -259,7 +262,8 @@ class AugMix(Algorithm):
                                     width=self.width,
                                     alpha=self.alpha,
                                     augmentation_set=self.augmentation_set)
-        dataset = state.train_dataloader.dataset
+        assert state.dataloader is not None, "dataloader should be defined on fit start"
+        dataset = state.dataloader.dataset
         if not isinstance(dataset, VisionDataset):
             raise TypeError(
                 textwrap.dedent(f"""\
--- a/composer/algorithms/colout/colout.py
+++ b/composer/algorithms/colout/colout.py
@@ -160,11 +160,15 @@ class ColOut(Algorithm):
         if self.batch:
             return event == Event.AFTER_DATALOADER
         else:
-            return event == Event.FIT_START and state.train_dataloader.dataset not in self._transformed_datasets
+            if event != Event.FIT_START:
+                return False
+            assert state.dataloader is not None, "dataloader should be defined on fit start"
+            return state.dataloader.dataset not in self._transformed_datasets
 
     def _apply_sample(self, state: State) -> None:
         """Add the ColOut dataset transform to the dataloader."""
-        dataset = state.train_dataloader.dataset
+        assert state.dataloader is not None, "dataloader should be defined on fit start"
+        dataset = state.dataloader.dataset
 
         transform = ColOutTransform(p_row=self.p_row, p_col=self.p_col)
 
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import contextlib
 import logging
-import textwrap
 import warnings
 from typing import TYPE_CHECKING, Any, Callable, ContextManager, Dict, List, Optional, Sequence, Union, cast
 
@@ -97,8 +96,6 @@ class State(Serializable):
             ``rank_zero_seed + dist.get_global_rank()``.
         grad_accum (int): The number of gradient accumulation steps to use. With this argument, micro batch size for
             each device becomes ``microbatch_size = train_batch_size / (num_devices * grad_accum)``.
-        train_dataloader (types.DataLoader, DataSpec, or dict):
-            The :class:`~.types.DataLoader`, :class:`~.DataSpec`, or dict of :class:`~.DataSpec` kwargs to used for training.
         evaluators (evaluator.Evaluator | Sequence[evaluator.Evaluator]):
             The evaluators contain the evaluation dataset(s) used for evaluation with specific metrics.
         max_duration (str or Time): The maximum duration to train for.
@@ -115,6 +112,7 @@ class State(Serializable):
         profiler (Optional[Profiler]): The Composer profiler.
 
     Attributes:
+        dataloader (types.DataLoader): The active :class:`~.types.DataLoader`.
         batch (types.Batch): The batch. This will be the entire batch during the :attr:`.Event.AFTER_DATALOADER`, or a
             microbatch between :attr:`.Event.BATCH_START` and :attr:`.Event.BATCH_END`.
         batch_num_samples (int): The number of samples in the :attr:`batch`.
@@ -152,7 +150,6 @@ class State(Serializable):
     """
 
     _max_duration: Time[int]
-    _steps_per_epoch: Optional[int]
     batch: types.Batch
     batch_num_samples: int
     batch_num_tokens: int
@@ -170,7 +167,6 @@ class State(Serializable):
         rank_zero_seed: int,
 
         # data configurations
-        train_dataloader: types.DataLoader,
         evaluators: Optional[Union[Evaluator, Sequence[Evaluator]]] = None,
         grad_accum: int = 1,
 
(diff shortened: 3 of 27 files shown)


hanlint:
I'm slightly concerned about 5 here -- `subset_num_batches` is also sometimes used to ensure a model overfits on a small amount of data.

ravi-mosaicml:
> I'm slightly concerned about 5 here -- `subset_num_batches` is also sometimes used to ensure a model overfits on a small amount of data.

Ah ok, makes sense. I will refactor the schedulers to use `train_subset_num_batches if not None else len(state.dataloader)`

composer/algorithms/augmix/augmix.py
```diff
@@ -251,15 +251,19 @@ def __init__(self,
         self._transformed_datasets = weakref.WeakSet()
 
     def match(self, event: Event, state: State) -> bool:
-        return event == Event.FIT_START and state.train_dataloader.dataset not in self._transformed_datasets
+        if event != Event.FIT_START:
```
hanlint, line 254: This is harder to read, can we redo this to:
```
assert state.dataloader is not None, "train dataloader must be defined"
return event == Event.FIT_START and state.dataloader.dataset not in self._transformed_datasets
```

composer/algorithms/colout/colout.py
```diff
@@ -160,11 +160,15 @@ def match(self, event: Event, state: State) -> bool:
         if self.batch:
             return event == Event.AFTER_DATALOADER
         else:
-            return event == Event.FIT_START and state.train_dataloader.dataset not in self._transformed_datasets
+            if event != Event.FIT_START:
```
hanlint, line 163: more pythonic to put the check all in one line, as before.

composer/core/state.py
```diff
@@ -170,7 +167,6 @@ def __init__(
         rank_zero_seed: int,
 
         # data configurations
-        train_dataloader: types.DataLoader,
```
hanlint, line 173: I did not catch the subtlety in the design documentation, but prefer to keep the dataloader in the `__init__`, this is consistent with other trainers.

composer/trainer/trainer.py
```diff
@@ -415,25 +424,26 @@ def __init__(
...
-        algorithms: Optional[Union[Algorithm, Sequence[Algorithm]]] = None,
-        optimizers: Optional[torch.optim.Optimizer] = None,
-        schedulers: Optional[Union[ComposerScheduler, Sequence[ComposerScheduler]]] = None,
+        train_dataloader: Union[DataLoader, DataSpec],  # TODO move param to fit()
+        max_duration: Union[int, str, Time],  # TODO move param to fit()
+        eval_dataloader: Optional[Union[DataLoader, DataSpec, Evaluator,
+                                        Sequence[Evaluator]]] = None,  # TODO move param to fit()
+        algorithms: Optional[Union[Algorithm, Sequence[Algorithm]]] = None,  # TODO move param to fit()
+        optimizers: Optional[torch.optim.Optimizer] = None,  # TODO move param to fit()
+        schedulers: Optional[Union[ComposerScheduler, Sequence[ComposerScheduler]]] = None,  # TODO move param to fit()
```
hanlint, line 433: I am not in favor of moving all these parameters into `fit`, does not seem neccessary for the goal of enabling eval-only mode?

composer/core/state.py
```diff
@@ -187,17 +183,13 @@ def __init__(
...
         callbacks: Optional[Union[Callback, Sequence[Callback]]] = None,
-
-        # steps per epoch
-        steps_per_epoch: Optional[int] = None,
     ):
         self.rank_zero_seed = rank_zero_seed
         self.model = model
         self.grad_accum = grad_accum
-        self.train_dataloader = train_dataloader
+        self.dataloader: Optional[types.DataLoader] = None
```
hanlint, line 190: is there a motivation for this rename?

ravi-mosaicml:
Upon further thought, I think it would make sense to leave `subset_num_batches` as part of state, so an algorithm (e.g. the future benchmarker) can end an epoch early to do dynamic profiling.

composer/algorithms/augmix/augmix.py
```diff
@@ -251,15 +251,19 @@ def __init__(self,
...
+        return state.dataloader.dataset not in self._transformed_datasets
 
     def apply(self, event: Event, state: State, logger: Logger) -> None:
         am = AugmentAndMixTransform(severity=self.severity,
                                     depth=self.depth,
                                     width=self.width,
                                     alpha=self.alpha,
                                     augmentation_set=self.augmentation_set)
-        dataset = state.train_dataloader.dataset
+        assert state.dataloader is not None, "dataloader should be defined on fit start"
```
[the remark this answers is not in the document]
ravi-mosaicml: I added the getter/setter on state so `state.dataloader_len` can be automatically whenever the dataloader is swapped. However, I would still prefer to require the caller to handle a nullable dataloader, rather than raising an exception (such as a  RuntimeError) if it is not set, as that will require the caller code to be more robust.

composer/loggers/progress_bar_logger.py
```diff
@@ -103,21 +102,14 @@ def _start(self, state: State):
         if dist.get_global_rank() != 0:
             return
         assert self.is_train is not None, "self.is_train should be set by the callback"
-        if self.is_train:
-            total_steps = state.steps_per_epoch
-        else:
-            total_steps = 0
-            for evaluator in state.evaluators:
```
ravi-mosaicml, line 110: Event.EVAL_START is now called for each evaluator, so the ProgressBarLogger does not need to sum over all evaluators anymore.

composer/core/state.py
ravi-mosaicml: Leaving the dataloader in init for now.

composer/trainer/trainer.py
ravi-mosaicml: Left arguments as-is for now; can revisit later.

tests/trainer/test_scale_schedule.py
```diff
@@ -91,21 +99,30 @@ def test_epochs_scaled(
             composer_trainer_hparams.algorithms = [ScaleScheduleHparams(ratio=ssr)]
         else:
             composer_trainer_hparams.scale_schedule_ratio = ssr
-        trainer = composer_trainer_hparams.initialize_object()
 
-        assert trainer.state.max_duration.unit == TimeUnit.EPOCH
-        assert trainer.state.max_duration.value == int(10 * ssr)
-        scheduler = trainer.state.schedulers[0]
+        class CheckScaleSchedule(Callback):
```
ajaysaini725, line 103: Should prob pull this out and make it a private class in the file rather than an inner class for this one test

In general best to avoid nested classes

ajaysaini725 (commented):
Just a few comments overall looks good!

composer/core/state.py
```diff
@@ -410,25 +418,74 @@ def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
...
-                    is greater than the number of batches in the training dataloader
-                    ({dataloader_len})"""))
-        self._steps_per_epoch = steps_per_epoch
+        if dataloader_len is not None and num_batches is not None and int(num_batches) > dataloader_len:
+            warnings.warn((f"DataloaderNumBatchesWarning: The dataloader_len ({int(num_batches)}) "
+                           f"is greater than the length (i.e. number of batches) of the dataloader, which is "
+                           f"{dataloader_len}. State.dataloader_len is thus being set to {dataloader_len}."))
+            num_batches = Time(dataloader_len, TimeUnit.BATCH)
+        if num_batches is None and dataloader_len is not None:
+            # len(dataloader) is an approximation -- see https://pytorch.org/docs/stable/data.html.
```
ajaysaini725, line 484: Worth putting this in the docstring?

ajaysaini725 (approved):
Just a small note otherwise LGTM 👍

composer/algorithms/augmix/augmix.py
hanlint: For evaluation only, we would never call this algorithm, though right?

PR #833, opened 2022-03-26
Remove Batch Types - Part 2: unify split batch

A-Jacobson:
The goal of this PR is to cover as many default cases as possible with one `split_batch` function. I was able to cover both of our current use cases:
1. dictionaries/mappings - for huggingface style datasets
2. tuples - for common vision datasets (batch dimension is the 2nd index for these)

I wasn't able to cover sequences with the batch dimension as the first index due to not being able to tell the difference between these and the tuple case because: 
1. can't typecheck - pytorch loaders  unfortunately convert tuples to lists by default so these tuples actually end up as lists.
2. can't check based on batch_size (even if we knew it ahead of time) because batch dimensions and tuple dimensions could end up the same length (batch of size 2 and `image, target` pair)

For now, these types of batches will have to use a separate `split_batch` function such as `_split_list`.

I also opted not to move `split_batch` to `Trainer._init_` for now as we would have to write an enum to set it with yahp.


abhi-mosaic (commented):
Added a few comments, still need to look over the tests later today.

tests/common.py
```diff
@@ -196,3 +196,85 @@ def __getitem__(self, index: int):
             return self.transform(x), y
         else:
             return x, y
+
+
+def dummy_tensor_batch(batch_size=12) -> torch.Tensor:
```
hanlint, line 201: Can these be moved  to the `test_split_batch` directly? They aren't "common" enough to be worth centralizing just yet.

hanlint (approved):
LGTM, can you do some quick canary runs with BERT to make sure the batch is being slip properly?

composer/core/data_spec.py
```diff
@@ -18,6 +18,68 @@
 __all__ = ["DataSpec"]
 
 
+def _split_list(l, num_microbatches):
```
ravi-mosaicml, line 21: Can you add type annotations?

composer/core/data_spec.py
```diff
@@ -18,6 +18,68 @@
...
+    if len(l) < num_microbatches:
+        raise ValueError(
+            textwrap.dedent(f"""\
+        Cannot split list of length {len(l)} into {num_microbatches} batches.
+         make sure `grad_accum` is less than or equal to `train_batch_size // world_size`."""))
+    return [l[i::num_microbatches] for i in range(num_microbatches)]
+
+
+def _split_tensor(t, num_microbatches):
```
ravi-mosaicml, line 30: Can you add type annotations?

composer/core/data_spec.py
```diff
@@ -18,6 +18,68 @@
...
+    if len(t) < num_microbatches:
+        raise ValueError(
+            textwrap.dedent(f"""\
+        Cannot split tensor of length {len(t)} into {num_microbatches} batches.
+         make sure `grad_accum` is less than or equal to `train_batch_size // world_size`."""))
+    return t.chunk(num_microbatches)
+
+
+def _split_mapping(m, num_microbatches):
```
ravi-mosaicml, line 39: Can you add type annotations?

composer/datasets/lm_datasets.py
```diff
@@ -163,10 +153,5 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
         sampler = dist.get_sampler(dataset, drop_last=self.drop_last, shuffle=self.shuffle)
 
         return DataSpec(dataloader=dataloader_hparams.initialize_object(
```
ravi-mosaicml, line 155: No need for a dataspec anymore -- can just return the dataloader

composer/datasets/glue.py
```diff
@@ -135,12 +134,10 @@ def tokenize_function(inp):
...
-        return DataSpec(
-            dataloader=dataloader_hparams.initialize_object(
-                dataset=dataset,  #type: ignore (thirdparty)
-                batch_size=batch_size,
-                sampler=sampler,
-                drop_last=self.drop_last,
-                collate_fn=data_collator,
-            ),
-            split_batch=_split_dict_fn)
+        return DataSpec(dataloader=dataloader_hparams.initialize_object(
```
ravi-mosaicml, line 137: No need for a dataspec anymore -- can just return the dataloader

composer/datasets/c4.py
```diff
@@ -127,13 +108,7 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
                                                                   mlm_probability=self.mlm_probability)
         # Return DataSpec
         return DataSpec(dataloader=dataloader_hparams.initialize_object(
```
ravi-mosaicml, line 110: No need for a dataspec anymore -- can just return the dataloader

composer/core/data_spec.py
```diff
@@ -18,6 +18,68 @@
...
+    chunked = {}
+    for k, v in m.items():
+        if isinstance(v, torch.Tensor):
+            chunked[k] = _split_tensor(v, num_microbatches)
+        if isinstance(v, (List, Tuple)):
+            chunked[k] = _split_list(v, num_microbatches)
```
ravi-mosaicml, line 45: can you add an error here for the else case?

composer/core/data_spec.py
```diff
@@ -18,6 +18,68 @@
...
+    return [{k: v[idx] for k, v in chunked.items()} for idx in range(num_microbatches)]
+
+
+def _default_split_batch(batch: Any, num_microbatches: int) -> Sequence:
```
ravi-mosaicml, line 49: Would more specific typing work (where `TBatch = TypeVar('TBatch')`)?
```suggestion
def _default_split_batch(batch: TBatch, num_microbatches: int) -> Sequence[TBatch]:
```

ravi-mosaicml (approved):
Overall LGTM; please see comments re: typing annotations.

Issue #814, opened 2022-03-23
Update conditional imports to use MissingConditionalImportError


moinnadeem:
Any conditional import (seen in `wandb_logger.py`, `brats.py`, `timm/model.py`, `ssd.py`, etc), should now use the [MissingConditionalImportErrors](https://github.com/mosaicml/composer/blob/dev/composer/utils/import_helpers.py#L12) class.

IanWorley:
I will take a shot at this issue

hanlint:
Great, thank you!

moinnadeem:
Great, let me know if you have any questions @IanWorley!

Here's some more details:
1. Look for any file that contains the text "if using pip" -- this is informing the user that they might need to install extra packages. If you need a hand finding these files, then let us know! I see the following: 
`composer/datasets/brats.py, `composer/loggers/wandb_logger.py`, `composer/models/unet/unet.py`, `composer/trainer/trainer.py`, `composer/models/ssd/ssd.py`, `composer/models/vit_small_patch16/hparams.py`, `composer/models/timm/model.py`, and `composer/models/timm/timm_hparams.py`. 

2. Replace the ImportError with a [MissingConditionalImportError](https://github.com/mosaicml/composer/blob/dev/composer/utils/import_helpers.py#L12).

Happy to quickly review the PR whenever you have it ready!

ravi-mosaicml:
Closed via #835

PR #835, opened 2022-03-26
conditional imports use MissingConditionalImportError #814

IanWorley:
I have finally updated the conditional imports.

### DIFF
--- a/composer/datasets/brats.py
+++ b/composer/datasets/brats.py
@@ -15,13 +15,13 @@ from dataclasses import dataclass
 import numpy as np
 import torch
 import torch.utils.data
-import torchvision
 import yahp as hp
 
 from composer.core.types import DataLoader, Dataset
 from composer.datasets.dataloader import DataLoaderHparams
 from composer.datasets.hparams import DatasetHparams
 from composer.utils import dist
+from composer.utils.import_helpers import MissingConditionalImportError
 
 PATCH_SIZE = [1, 192, 160]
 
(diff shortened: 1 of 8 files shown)


composer/datasets/brats.py
```diff
@@ -264,8 +264,8 @@ def get_split(data, idx):
 def get_data_split(path: str):
     try:
         from sklearn.model_selection import KFold
-    except ImportError as e:
-        raise ImportError(
+    except MissingConditionalImportError as e:
```
ravi-mosaicml, line 267: The `except` should still be an `ImportError`, as the external libraries will not raise `MissingConditionalImportError`.

composer/datasets/brats.py
```diff
@@ -264,8 +264,8 @@ def get_split(data, idx):
...
+        raise MissingConditionalImportError(
             textwrap.dedent("""\
```
ravi-mosaicml, line 269: The `MissingConditionalImportError` signature takes three arguments instead of an error string. Please see https://github.com/mosaicml/composer/blob/d765fa23909dc1f14f27fecde194d3e4842670d3/composer/utils/import_helpers.py#L12.

ravi-mosaicml (commented):
Thanks for working on this issue @IanWorley!

The initialization of `MissingConditionalImportError` has a different signature than a normal `ImportError`. If you could adjust the arguments to match the function signature, that would be great. Thanks!

IanWorley:
@ravi-mosaicml I will get this fix by tonight thank you.

hanlint (approved):
Looks good, thanks @IanWorley !

IanWorley:
@hanlint thank you so much I do not have a proper linux box.

hanlint:
@IanWorley no problem was an easy fix. Thanks for contributing!

PR #840, opened 2022-03-29
Fix the entrypoint and launch script

ravi-mosaicml:
* In the entrypoint, save the hparams to a tempfile rather than to a folder
* In the launch script, properly exit if the stdout or stderr files already exist. In addition, print the exception before attempting to shut down. Also fixed a bug where process groups were not being properly killed.


Landanjs (approved):
I don't understand the code, so maybe should get a review from someone on eng as well?

I believe I was able to run both resnet50 and deeplabv3 with this fix, thanks!! Not sure if there are particular things I should check for the run? I only checked if they did not error out

hanlint (approved):

PR #842, opened 2022-03-29
Add Computed Metrics to State

### LINKED ISSUE #830: Callbacks need access to computed metrics

ravi-mosaicml:
Add the current metrics to the state.
Reconfigured test case tolerances and deterministic mode, as the metrics are stored on the state. While deterministic mode works well for CPU tests, it does not fully work on GPU tests, especially for large models. So, the state test cases use tolerances for all tensor type comparisons.
Closes #830


composer/trainer/trainer.py
```diff
@@ -1107,10 +1089,17 @@ def _train_loop(self) -> None:
                 for scheduler in self.state.schedulers:
                     scheduler.step()
 
+            if self.train_metrics is not None:
```
hanlint, line 1092: this was not in the original code, should it be introduced here?

tests/algorithms/test_stochastic_depth.py
```diff
@@ -223,6 +223,7 @@ def get_drop_rate_list(self, module: torch.nn.Module, drop_rates: Optional[List]
     @pytest.mark.parametrize("drop_distribution", ['uniform', 'linear'])
     @pytest.mark.parametrize("use_same_gpu_seed", [True])
     @pytest.mark.parametrize("drop_warmup", ["0.1dur"])
+    @pytest.mark.timeout(15)
```
hanlint, line 226: is this a regression?

tests/conftest.py
```diff
@@ -19,9 +19,8 @@
 # default timout threshold is 2 seconds for determinign long and short
 DEFAULT_TIMEOUT = 2.0
 
-# Enforce use of deterministic kernels
-# see composer.utils.reproducibility.configure_deterministic_mode
-os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
```
hanlint, line 24: This was introduced in https://github.com/mosaicml/composer/commit/79d0bec0b9e2fdf22837ef1632d81c8b03ee883e

based on the note in: https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility

I don't see it added back in for `configure_deterministic_mode`. Was this supposed to have been removed?

tests/metrics/test_computed_metrics.py
```diff
@@ -0,0 +1,130 @@
...
+def test_computed_metrics(
+    dummy_train_dataloader: DataLoader,
+    dummy_val_dataloader: DataLoader,
+    dummy_num_classes: int,
+    dummy_in_shape: Tuple[int, ...],
+    compute_training_metrics: bool,
+    validate_every_n_batches: int,
+    validate_every_n_epochs: int,
+):
+    num_channels = dummy_in_shape[0]
```
hanlint, line 66: this test could benefit from line breaks and organization to break up into paragraphs, e.g. before `Trainer(`, before `trainer.fit(`.

The code around creaing the metrics callback should also be moved above the `trainer(`, since the model and mock loggers are also created.

hanlint (commented):
LGTM, made a few minor comments and questions.

tests/conftest.py
ravi-mosaicml: `configure_deterministic_mode` sets this variable. See https://github.com/mosaicml/composer/blob/dc920e3c13d03b33a4b7d2443cfc7fb7e31f9d2f/composer/utils/reproducibility.py#L116

tests/trainer/test_trainer.py
```diff
@@ -157,7 +157,6 @@ def config(self, device: Device, precision: Precision, world_size: int, rank_zer
             'seed': rank_zero_seed,
             'device': device,
             'precision': precision,
-            'deterministic_mode': True,  # testing equivalence
```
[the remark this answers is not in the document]
ravi-mosaicml: Since the computed metrics are on state, deterministic mode is needed so the metrics are exactly equal for the checkpoint test. However, since we configure deterministic mode in `conftest.py`, this is no longer needed here.

composer/trainer/trainer.py
ravi-mosaicml: We do something similar for the evaluators, we we compute on each batch, and then afterwards, so I think it would make sense to do the same for the training dataloader. This also ensures that the metrics will get logged at least once per epoch if someone sets the LogLevel to EPOCH instead of batch.

docs/source/trainer/checkpointing.rst
```diff
@@ -83,7 +83,7 @@ The above code, when run, will produce the checkpoints below:
     >>> list(state_dict)
     ['state', 'rng']
     >>> list(state_dict['state'].keys())
-    ['model', 'optimizers', 'schedulers', 'algorithms', 'callbacks', 'scaler', 'timer', 'rank_zero_seed']
+    ['model', 'optimizers', 'schedulers', 'algorithms', 'callbacks', 'scaler', 'timer', 'rank_zero_seed', 'computed_metrics']
```
hanlint, line 86:
```suggestion
    ['model', 'optimizers', 'schedulers', 'algorithms', 'callbacks', 'scaler', 'timer', 'rank_zero_seed', 'current_metrics']
```

docs/source/trainer/checkpointing.rst
```diff
@@ -159,6 +159,8 @@ state from the checkpoint are not compatible with these new objects.
     +-----------------------+-------------------------------------------------------------+
     | rank_zero_seed        | The seed of the rank zero process.                          |
     +-----------------------+-------------------------------------------------------------+
+    | computed_metrics      | The computed metrics.                                       |
```
hanlint, line 162: Is this supposed to also be `current_metrics`?

hanlint (approved):
LGTM

docs/source/trainer/checkpointing.rst
ravi-mosaicml: Good catch, thanks! (also caught via doctests)

```
00:06:11.021  File "trainer/checkpointing.rst", line 83, in default
00:06:11.021  Failed example:
00:06:11.021      list(state_dict['state'].keys())
00:06:11.021  Expected:
00:06:11.021      ['model', 'optimizers', 'schedulers', 'algorithms', 'callbacks', 'scaler', 'timer', 'rank_zero_seed', 'computed_metrics']
00:06:11.021  Got:
00:06:11.021      ['model', 'optimizers', 'schedulers', 'algorithms', 'callbacks', 'scaler', 'timer', 'rank_zero_seed', 'current_metrics']
```

PR #843, opened 2022-03-29
Ade20k streaming dataset yaml

Landanjs:
Adding a yaml for ade20k streaming dataset.

I created datasets with a varying number of shards to try to get the baseline as similar to the localdisk as possible

| # of shards (train/val) | train samples dropped | val samples dropped | # of iterations | mIoU              | TTT (hrs) |
|-------------------------|-----------------------|---------------------|-----------------|-------------------|-----------|
| Localdisk               | 0                     | 0                   | 80768           | 45.556 +/- 0.3025 | 5.34      |
| 512/64                  | 242                   | 16                  | 79872           | 45.298 +/- 0.283  | 5.25      |
| 64/64*                  | 46                    | 16                  | 79872           | 45.718 +/- 0.3073 | 5.25      |
| 48/48*                  | 46                    | 32                  | 80640           | 45.526 +/- 0.154  | 5.31      |
* I found 4 samples to be corrupted in ADE20k, I've removed these 4 samples for these tests

[wandb runs for localdisk](https://wandb.ai/mosaic-ml/cory-deeplabv3-variability?workspace=user-landanjs). Group by `Group`.
[wandb runs for sharding tests](https://wandb.ai/mosaic-ml/landan-deeplabv3plus-streaming?workspace=user-landanjs). Group by `metadata.sweep`

In the case of 48 shards, the number of shards needs to be divisible by the number of workers, so I decreased the number of workers to 48.

Note:
- For the sharding in the last row, if we add 2 duplicate samples than no samples would be dropped.
- I think the number of lost iterations (outside of the samples completely removed from training) can be calculated with: `((int(dataset_size / n_shards) % (batch_size / n_gpus)) * num_workers * n_gpus / batch_size) * n_epochs`, so there is definitely some weird interactions here where increasing or decreasing num_workers effects the number of training iterations.

## Conclusion
I think we should go with the 48 sharding and 48 num workers setup since I felt that this was closest to the original besides the additional validation samples dropped (we should do validation on the full dataset after training anyway).


coryMosaicML (approved):
LGTM, Fantastic eval work!

PR #848, opened 2022-03-30
Remove no longer needed xfails

ravi-mosaicml:
1. The xfail on ghost_batchnorm was incorrect; instead it should use a `with pytest.raises`.
1. Since we seed each test, `train_model` should no longer be flaky.


tests/utils/trainer_fit.py
```diff
@@ -33,7 +32,6 @@ def get_total_loss(model: ComposerModel, dataloader: DataLoader, device: Device)
 
 
 def train_model(composer_trainer_hparams: TrainerHparams, max_epochs: int = 2, run_loss_check: bool = False):
```
hanlint, line 34: Do we still need this test given `test_trainer.py` ? Should we refactor this into that test file and its pattern?

hanlint (approved):

tests/utils/trainer_fit.py
ravi-mosaicml: Nothing is using `train_model`....should I remove it?

hanlint: Yes

PR #849, opened 2022-03-30
[FFCV] webdataset from local + download only once

dskhudia:
1) Webdataset can be specified from a local dir as well. 
2) Use FFCV's internal api to avoid downloading dataset twice. 

Test:

```
python scripts/ffcv/create_ffcv_datasets.py --dataset cifar --split train --remote s3://mosaicml-internal-dataset-cifar10 --num_workers 64

python scripts/ffcv/create_ffcv_datasets.py --dataset cifar --split train --remote /tmp/webdata --num_workers 64
```

### DIFF
--- a/composer/datasets/ffcv_utils.py
+++ b/composer/datasets/ffcv_utils.py
@@ -1,10 +1,12 @@
 import json
 import logging
-import subprocess
 import textwrap
 from typing import Optional
 
+import numpy as np
+
 from composer.core.types import Dataset
+from composer.datasets.webdataset_utils import init_webdataset_meta
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


composer/datasets/ffcv_utils.py
```diff
@@ -4,13 +4,50 @@
...
+    """Read a WebDataset meta file from local filesystem.
+
+    Args:
+        remote (str): Local filesystem directory.
+    """
+    path = f'{remote}/meta.json'
+    return open(path, 'rb').read()
+
+
+def _init_webdataset_meta(remote: str) -> bytes:
```
knighton, line 39: Thoughts on making _init_webdataset_meta in webdataset_utils non-underscored and reusing that?

If someone changes webdataset shard metadata fields in the future (eg, for non-perfectly-even shard sizes with no leftovers), this code could get overlooked and become broken

dskhudia: Sounds good. Only issue I see is use of split in _init_webdataset_meta from webdataset_utils. In my case, split is already part of remote.

dskhudia: I can change _init_webdataset_meta from webdataset_utils to make split optional and then reuse that. Does that work for you?

knighton: I trust your excellent judgment, indeed that API is slightly awkward

Example prototyping some changes at https://github.com/mosaicml/composer/pull/846/files

dskhudia: @knighton Fixed it. Let me know if it looks good now.

knighton (approved):

PR #853, opened 2022-03-30
Preserving module parameter values in surgery.

### LINKED ISSUE #297: `surgery.replace_module_classes` should copy initialization weights where possible
## 🚀 Feature Request

Right now, `surgery.replace_module_classes` does not preserve any initialized model weights for modules that are replaced. However, in cases where there is a 1-1 mapping of weights in the replacement (ex. in StoachasticDepth), we should preserve initialized model weights. 

## Motivation

Surfaced by a user.

ravi-mosaicml:
Copies over parameter tensor values from the old module to the new module when tensor weights and shapes do not change.

Closes #297.


composer/utils/module_surgery.py
```diff
@@ -101,13 +101,11 @@ def replace_module_classes(
         >>> replace_module_classes(module, policies, recurse_on_replacements=True)
         {Conv2d(1, 32, kernel_size=(3, 3), stride=(1, 1)): Linear(in_features=16, out_features=32, bias=True), Linear(in_features=16, out_features=32, bias=True): Linear(in_features=32, out_features=64, bias=True)}
 
-    .. warning::
-
-        When a module is replaced, any tensor values within the module are not copied over
-        to the new module even when the shape is identical. For example, if model weights
-        are initialized prior to calling this function, the initialized weights will not
-        be preserved in any replacements.
+    .. note::
```
hanlint, line 104: Let's add a `copy_weights: bool = False` flag to control this behavior. Some of our algorithms, such as BlurPool, handle the weight copy logic internally (e.g. see `from_Conv2d`) that we should not override automagically here.

tests/test_surgery.py
```diff
@@ -95,6 +95,37 @@ def test_module_replacement(model_cls: Type[SimpleReplacementPolicy], recurse_on
...
+        assert isinstance(module, torch.nn.Linear)
+        new_linear = torch.nn.Linear(
+            in_features=module.in_features,
+            out_features=module.out_features,
+            bias=module.bias is not None,
+            dtype=module.weight.dtype,
+        )
+
+        torch.nn.init.constant_(new_linear.weight, 0.314)
+        torch.nn.init.constant_(new_linear.bias.data, 0.25)
```
hanlint, line 113: This should be `0.271` to match `0.314`

dblalock (commented):
Think there are a couple subtleties here that make this a bit more complex, sadly. tl;dr is that we probably need a separate `copy_params` function that tries to copy over parameter attributes by name, recursing on submodules by name.

composer/utils/module_surgery.py
ravi-mosaicml: Agree that this should be a parameter, though I don't have an opinion on the default value.

ravi-mosaicml:
Thanks for these comments! Upon further thought, I'm wondering if this should be a feature of the surgery function. The alternative would be that the replacement policy function should be responsible for copying (or rather, setting) parameter values. It (slightly) scares me to attempt to use a heuristic to attempt to determine whether parameters can be copied (e.g. same name, shape, and dtype), especially in the recursive case. I do like the idea of refactoring this into a separate function that then the replacement policy could call on specific modules, if desired. This function could be strict (e.g. all parameters in the passed-in module must match exactly).

dblalock:
So we write a separate helper function in `surgery` that replacement functions can choose to call? And then refactor existing replacement logic to use this when applicable? Sounds reasonable to me, although I'm not certain in how many cases it actually would be applicable.

hanlint:
Agree that this may require some additional design discussion. Closing this PR for now, but queued it for an offline discussion.

PR #854, opened 2022-03-31
Remove the Scale Schedule Algorithm

### LINKED ISSUE #434: Fully remove ScaleSchedule algorithm
ScaleSchedule as an algorithm is being deprecated shortly. This issue is to track its full removal.

ravi-mosaicml:
The algorithm form of scale schedule has been deprecated. It is available as an argument to the trainer.

This PR removes the Scale Schedule Algorithm as an algorithm (it must be specified via the trainer init args).
It also restores the scale schedule method card that was in the 0.3.1 release. This method card has been updated to
reflect the non-algorithm-class usage.

Closes #434.


hanlint:
This also removes the `README` about scale schedule, which I think it an integral part. Perhaps we leave it deprecated, but existing in the codebase, for now until we have a better path.

ravi-mosaicml:
> This also removes the README about scale schedule, which I think it an integral part. Perhaps we leave it deprecated, but existing in the codebase, for now until we have a better path.

The README for the scale schedule is a stub. It currently read:

># 🗜️ Scale Schedule
> 
> This method is deprecated. It has been replaced by the `scale_schedule_ratio` param supported directly by the Composer Trainer. For backwards compatibility, the Composer Trainer detects when this algorithm has been initialized and pulls the `ratio` param accordingly.

We seemed to have lost the original method card for scale schedule (https://docs.mosaicml.com/en/v0.3.1/method_cards/scale_schedule.html). Would it make sense to add this documentation back under the "Using The Trainer", since this flag is now under the trainer args?

ravi-mosaicml:
We have docs for the scale schedule ratio here: https://docs.mosaicml.com/en/v0.5.0/trainer/schedulers.html#scale-schedule-ratio

hanlint:
Hmm defer to @jfrankle on visibility concerns of scale schedule here, but would it be possible to keep the Scale Schedule stub in the README, and point it to the appropriate documentation?

It's one of our core methdology, so want to make sure it's not buried somewhere.

jfrankle:
I'd like to keep the stub so that, when people go looking for it, they get redirected to the right place. It should be as visible as possible. Ideally, the stub will produce an error if used so that people can correct their behavior.

ravi-mosaicml:
Finally got around to revising the docs. I took the method card we had in the 0.3.1 release and modified it to reflect the new implementation. Could you please re-review? Thanks!

hanlint (approved):
LGTM

PR #858, opened 2022-04-04
Rewrite streaming datasets

knighton:
- No shuffle buffer, does optimal shuffle (see _iter_ids)
- Split on the basis of samples, not shards
-> No leftover samples, no balanced shards requirement
- Boto3 instead of awscli


composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,227 @@
...
+        """Load a StreamingDatasetIndex from raw bytes.
+
+        Args:
+            data (bytes): The serialized form.
+
+        Returns:
+            cls: The loaded object.
+        """
+        begin = 0
+        end = begin + 2 * 8
```
ravi-mosaicml, line 118: These constants (2, 8) seem a bit magical. Could you define them somewhere or read them from a library

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,227 @@
...
+                 bytes_per_sample: Sequence[int], fields: Sequence[str]) -> None:
+        """Initialize with sample statistics.
+
+        Args:
+            samples_per_shard (Sequence[int]): Number of samples of each shard.
+            bytes_per_shard (Sequence[int]): Size in bytes of each shard.
+            bytes_per_sample (Sequence[int]): Size in bytes of each sample across all shards.
+            fields (Sequence[str]): The names of the samples' fields in order.
+        """
+        self.samples_per_shard = np.asarray(samples_per_shard, np.int32)
```
ravi-mosaicml, line 89: Curious about the choice to encode these values as 32 bit integers rather than 64 bit? This means that the maximum size of a shard (or sample) is only 2.1 GB, which seems small, especially on a 800 Gbps (100 GBps) infiniband backbone. Is 32 bit numbers the convention in the web dataset spec, or do we get to write the spec?

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,227 @@
...
+
+    The shards are just dumb buffers with samples catted together. All the
+    offset info across the whole dataset is contained in the index file. Workers
+    read this file to calculate how much of which shards their slice is.
+
+    Each sample is a dict of str to bytes. All samples must contain the same
+    dict keys (fields). These strings are stored in the index file for efficiency.
+    """
+
+    def __init__(self, samples_per_shard: Sequence[int], bytes_per_shard: Sequence[int],
```
ravi-mosaicml, line 79: Is it possible to use numpy arrays instead of python lists for these variables? Each python integer is 28 bytes minimum...a bit concerned about how much memory these lists will take up, especially for large datasets when replicated multiple times per node.

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+        self.new_samples = []
+        self.new_shard_size = 0
+
+    def _flush_shard(self) -> None:
+        """Flush cached samples to a new dataset shard."""
+        if not self.samples_per_shard:
+            os.makedirs(self.dirname)
+        shard = len(self.samples_per_shard)
+        basename = get_shard_basename(shard)
+        filename = os.path.join(self.dirname, basename)
```
ravi-mosaicml, line 43: Is it possible to write first to a temp file, and them rename upon successful completion, so that the writing can be gracefully resumed in case if it crashes?

ravi-mosaicml (commented):
Wow -- supporting straming datssets seems wuite complex. Thanks for taking this stab at it.

I didn't give this PR a close review, but at a high level, I am not sure of the assumptions w.r.t memory complexity. In addition, have we tested this to see how it would work when using multiple dataloader workers? I think the streaming dataset reader would not function correctly nor be as effecient as possible when num_workers>0. (For example, I don't think multiple workers would need to open the same shard?) Multiple datalosder workers are useful to compute transformations on multiple samples in parallel.

In addition, can you add in test cases? Especially interested in tests that use multiple dataloader workers and also validate the correctness of reading and writing the index file.

Finally, it would be nice if the index file written could be parallielized via subprocesses. OK deferring this feature to later.

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+        self.transform = transform
+        self.target_transform = target_transform
+        self.shuffle = shuffle
+
+        # Load the index file containing the shard metadata, either over the
+        # network or cached locally.
+        # Precomputes the shard and offset in bytes of each sample (for direct
+        # access).
+        local = self._download_if_missing(get_index_basename())
+        self.index = StreamingDatasetIndex.load(open(local, 'rb'))
```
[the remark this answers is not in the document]
knighton: On my machine, __del__ gets called when creating object inside method arguments

composer/datasets/streaming/format.py
knighton: (Firsts epoch only) shard size is directly related to (a) how choppy/laggy training is, (b) how lumpy the number of samples you have to randomly sample from, and (c) how likely/unlikely samples are to be seen together (impacts quality). I've been eyeballing shard sizes to <=32MB or even smaller to keep it smooth/random.

Note that we could allocate contiguous space for the entire epoch of shards up front, and load the shards into offsets into that, if needed.

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+            min_id = max(part_min_id, shard_min_id)
+            max_id = min(part_max_id, shard_max_id)
+            new_ids += list(range(min_id, max_id + 1))
+
+        with self._lock:
+            # Keep open a file handle to each loaded shard for fast access.
+            for shard in shards:
+                basename = get_shard_basename(shard)
+                filename = os.path.join(self.local, self.split, basename)
+                self._files[shard] = open(filename, 'rb')
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,227 @@
...
+def get_shard_basename(shard: int) -> str:
+    """Get the basename for a streaming dataset shard.
+
+    Args:
+        shard (int): Which shard.
+
+    Returns:
+        str: Basename of file.
+    """
+    return '%05d.mds' % shard
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,227 @@
...
+    for begin, end in zip(begins, ends):
+        value = data[begin:end]
+        values.append(value)
+    return dict(zip(keys, values))
+
+
+class StreamingDatasetIndex(object):
+    """Streaming dataset index file, containing all the info about shards.
+
+    The shards are just dumb buffers with samples catted together. All the
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/format.py
knighton: Yeah, I was trying to be careful about that. In the Writer, where ram usage isn’t a concern, these values are accumulated in lists of int. Then in the Index, they are loaded from serialized format in np.ndarray of np.int32.

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,90 @@
+import os
+import shutil
+from time import sleep
+
+import boto3
+
+
+def wait_for_download(local: str) -> None:
```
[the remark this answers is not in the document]
knighton: This is just informational for early usage of the library. Not expected to happen in practice. Neighboring partitions both trying to download the same shard at the same time will be rare. Shard sizes will be small. Added an optional hard timeout.

composer/datasets/streaming/format.py
knighton: Done

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def download_from_s3(remote: str, local: str) -> None:
+    """Download a file from remote to local.
+
+    Args:
+        remote (str): Remote path (S3).
+        local (str): Local path (local filesystem).
+    """
+    assert remote.startswith('s3://')
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+        self.bytes_per_sample = []
+
+        # Data of the shard in progress.
+        self.new_samples = []
+        self.new_shard_size = 0
+
+    def _flush_shard(self) -> None:
+        """Flush cached samples to a new dataset shard."""
+        if not self.samples_per_shard:
+            os.makedirs(self.dirname)
```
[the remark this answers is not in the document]
knighton: If we allow to continue if directory already exists, we could mistakenly overwrite new shards over old shards, in which there are fewer new shards so some of the old gets carried along too (the index would still work fine, but the user might not notice)

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+        shard = len(self.samples_per_shard)
+        basename = get_shard_basename(shard)
+        filename = os.path.join(self.dirname, basename)
+        with open(filename, 'wb') as out:
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/writer.py
knighton: I'd rather have it die very obviously than risk unnoticed subtle differences in a continued run to write shards?

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,90 @@
...
+        remote (str): Remote path (S3 or local filesystem).
+        local (str): Local path (local filesystem).
+    """
+    # If we already have the file cached locally, we're done.
+    if os.path.exists(local):
+        return
+
+    # Else if someone else is currently downloading the shard, wait for that download to complete.
+    local_tmp = local + '.tmp'
+    if os.path.exists(local_tmp):
```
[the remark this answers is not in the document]
knighton: Nice

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # Load the index file containing the shard metadata, either over the
+        # network or cached locally.
+        # Precomputes the shard and offset in bytes of each sample (for direct
+        # access).
+        local = self._download_if_missing(get_index_basename())
+        self.index = StreamingDatasetIndex.load(open(local, 'rb'))
+
+        # Fields, protected by the lock, relating to loading shards in the background.
+        self._lock = Lock()
```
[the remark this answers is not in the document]
knighton: Lock is created in `__init__`, not taken until `__iter__` is called later.  `__init__` is called once, in the original process, not in the workers. Believe it's fine

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+        self._files: List[Optional[BufferedReader]] = [None] * self.index.num_shards
+        self._next_epoch_key = 0
+        self._key2epoch = {}
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+        """Initialize with the given remote path and local cache.
+
+        Loads all the samples that are available in local cache, then starts a background thread to download the rest
+        during training. As samples are added, shuffled sample selection becomes more random.
+
+        Args:
+            remote (str): Download shards from this remote directory.
+            local (str): Download shards to this local filesystem directory for reuse.
+            split (str): Which dataset split.
+            transform (callable, optional): X transformation.
```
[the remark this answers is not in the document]
knighton: Split into StreamingDataset and StreamingVisionDataset

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,314 @@
...
+    world_size = dist.get_local_world_size()
+    part = rank * workers_per_device + worker_id
+    num_parts = workers_per_device * world_size
+    return part, num_parts
+
+
+class StreamingDataset(IterableDataset):
+    """StreamingDataset."""
+
+    def __init__(self, remote: str, local: str, split: str, transform: Callable, target_transform: Callable,
```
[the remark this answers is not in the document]
knighton: Library is currently designed to do random access to shards on disk, doesn't keep them in memory, and will always need all shards

Thoughts on maybe adding a `in_memory: bool = True` parameter for the shard data? Maybe with a size limit, but having it partially hot in ram sounds complicating

composer/datasets/streaming/dataset.py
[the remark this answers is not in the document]
ravi-mosaicml: Sure, though it's generally not best practice to rely on `__del__` (it's a cpython implementation detail that causes this behavior)

composer/datasets/streaming/format.py
ravi-mosaicml: I could see a case where we cache datasets locally via a s3 appliance on the infiniband backbone (even if we don't do this, OSS users definitely could)

knighton: Switching all int32 to int64 after seeing another cumsum overflow

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,101 @@
...
+        timeout (Optional[float]): How long to wait before raising an exception. Default: 10 sec.
+    """
+    # If we already have the file cached locally, we are done.
+    if os.path.exists(local):
+        return
+
+    # No local file, so check to see if someone else is currently downloading
+    # the shard. If they are, wait for that download to complete.
+    local_tmp = local + '.tmp'
+    if os.path.exists(local_tmp):
```
dskhudia, line 83: This might still be a race condition if two processes reach at the same but none have create local_tmp. How about using multiprocessing.Lock?

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,385 @@
...
+        """Get an iterator over all our sample IDs.
+
+        Returns:
+            Iterator[int]: Each sample ID.
+        """
+        with self._lock:
+            have_full_epoch = self._is_loading_complete
+
+        if have_full_epoch:
+            epoch = list(self._loaded_epoch)
```
dskhudia, line 269: Make _loaded_epoch as np.array. Converting to python list and then passing it to np.random.shuffle may be expensive.

knighton: Oof, benchmarked this and it turns out the shuffling and array creation are expensive at >1e6 samples

Proposed solution (shard shuffle then interleave into global) vs baseline (extend then global shuffle):

```
from argparse import ArgumentParser
from matplotlib import pyplot as plt
import numpy as np
from time import time
from tqdm import tqdm


def parse_args():
    args = ArgumentParser()
    args.add_argument('--num_shards', type=int, default=1000)
    args.add_argument('--approx_shard_size', type=int, default=1000)
    args.add_argument('--out', type=str, default='plot.png')
    return args.parse_args()


def extend_and_shuffle(samples_per_shard):
    shard_ends = samples_per_shard.cumsum()
    shard_begins = shard_ends - samples_per_shard

    t0 = time()
    tt = [0]
    nn = [0]
    ids = []
    for shard in tqdm(np.random.permutation(len(samples_per_shard))):
        begin = shard_begins[shard]
        end = shard_ends[shard]
        new_ids = list(range(begin, end))
        ids += new_ids
        np.random.shuffle(ids)
        t = time() - t0
        tt.append(t)
        n = len(ids)
        nn.append(n)
    return tt, nn


def shuffle_and_interleave(samples_per_shard):
    shard_ends = samples_per_shard.cumsum()
    shard_begins = shard_ends - samples_per_shard

    t0 = time()
    tt = [0]
    nn = [0]
    ids = None
    for shard in tqdm(np.random.permutation(len(samples_per_shard))):
        begin = shard_begins[shard]
        end = shard_ends[shard]
        new_ids = begin + np.random.permutation(end - begin)
        if ids is None:
            ids = new_ids
            continue
        num_cycles = len(new_ids)
        cycle_len = len(ids) // num_cycles
        cutoff = num_cycles * cycle_len
        old_block = ids[:cutoff].reshape(num_cycles, cycle_len)
        tail = ids[cutoff:]
        new_block = np.expand_dims(new_ids, 1)
        block = np.concatenate([old_block, new_block], 1)
        ids = np.concatenate([tail, block.flatten()])
        t = time() -t0
        tt.append(t)
        n = len(ids)
        nn.append(n)
    return tt, nn


def main(args):
    samples_per_shard = (args.approx_shard_size * np.exp(np.random.normal(0, 0.25, args.num_shards))).astype(np.int64)
    print(samples_per_shard)
    print()
    print('Shards: %d' % len(samples_per_shard))
    print()
    x = samples_per_shard
    print('Samples per shard:')
    print(f'* Min {x.min()}')
    print(f'* Mean {x.mean():.3f}')
    print(f'* Std {x.std():.3f}')
    print(f'* Max {x.max()}')
    print()

    plt.title('Shard Loading Speed on Simulated ImageNet')
    plt.xlabel('Time (seconds)')
    plt.ylabel('Samples')

    print('Bench extend_and_shuffle...')
    tt, nn = extend_and_shuffle(samples_per_shard)
    plt.plot(tt, nn, c='blue', label='extend + shuffle')

    print('Bench shuffle_and_interleave...')
    tt, nn = shuffle_and_interleave(samples_per_shard)
    plt.plot(tt, nn, c='green', label='shuffle + interleave')

    plt.savefig(args.out)


if __name__ == '__main__':
    main(parse_args())
    ```

Comment (knighton) [reply composer/datasets/streaming/dataset.py]: plot

File: composer/datasets/streaming/download.py (hunk 18 above)
Comment (knighton) [reply composer/datasets/streaming/download.py:83]: How do the (8 gpus x 8 workers per gpu) processes get the same lock? Was thinking of a file lock (fcntl)

File: composer/datasets/streaming/dataset.py (hunk 19 above)
Comment (abhi-mosaic) [reply composer/datasets/streaming/dataset.py]: How important is 40 seconds for epoch 1? From my understanding the processing of 1M samples will in ~64 separate threads, on high-end CPUs, and the download speed from S3 is far slower, O(minutes). So if we get perfect shuffling in 40 seconds, I think that's totally fine.

Comment (knighton) [reply composer/datasets/streaming/dataset.py]: Worth noting this is worst-case, in which no sample IDs are being popped off and used at the same time, so the sample ID list grows maximally. In real world usage, that limits the badness

The extend and shuffle is O(n^2) though right? Not tractable for much larger than ImageNet

Comment (abhi-mosaic) [reply composer/datasets/streaming/dataset.py]: Agreed on this being the worst case, and will be better in practice as ids are being removed as samples are being processed.

Also, I think `np.random.shuffle` is O(n), the blue line is definitely leveling off, I think the underlying algo may be something like Fisher-Yates.

Overall as an ML-user, I would prefer the blue line with perfect shuffling, rather than the green line with imperfect shuffling across shards. I think there is also a 2x2 quadrant analysis here: If our buffers are growing to 100s of millions of samples, then the model must be going really slow relative to the dataloader, so the shuffling time doesn't make a difference. And if the model is going really fast, then IDs are being popped off really fast, so the shuffle list does not grow large.

File: composer/datasets/streaming/download.py (hunk 18 above)
Comment (dskhudia) [reply composer/datasets/streaming/download.py:83]: Usually needs to be passed in as argument while creating processes but here training processes are created beforehand by composer command. Some way to lock this critical section is needed otherwise we will run into some hard-to-debug weird issues.

**Comment by knighton** [comment]
CIFAR10 end-to-end test run:
runs

Baseline: train mean 4.242 batch/s std 0.090 batch/s, val mean 9.348 batch/s std 0.400 batch/s
Streaming: train mean 4.361 batch/s std 0.115 batch/s, val mean 14.190 (!) batch/s std 0.734 batch/s

Something looks off about performance on initial epoch(s), but at parity afterward

File: composer/datasets/streaming/dataset.py (hunk 19 above)
Comment (knighton) [reply composer/datasets/streaming/dataset.py]: Sounds good, will not change to interleave

To be clear, leveling off means infinite time

**Comment by knighton** [comment]
Scaling num_workers, both cached locally on disk, same transforms

mds8-imagenet1k
mds9-cifar10

File: composer/datasets/streaming/dataset.py (hunk 19 above)
Comment (abhi-mosaic) [reply composer/datasets/streaming/dataset.py]: Oh oops I had my axes mixed up... 😂 but I think we're still OK, let's try it without interleave and revisit if we find a workload that is negatively impacted by the shuffling time.

**Comment by knighton** [comment]
Included for completeness, but not apples to apples comparison because streaming shards contain the raw 64x64x3 uint8s not image files

mds7-tinyimagenet200

File: composer/datasets/ade20k.py [hunk 20]
```diff
@@ -388,6 +391,135 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+        else:
+            both_transform = None
+            image_transform = transforms.Resize(size=(self.final_size, self.final_size),
+                                                interpolation=TF.InterpolationMode.BILINEAR)
+            annotation_transform = transforms.Resize(size=(self.final_size, self.final_size),
+                                                     interpolation=TF.InterpolationMode.NEAREST)
+
+        remote = os.path.join(self.remote, self.split)
+        local = os.path.join(self.local, self.split)
+        dataset = StreamingADE20k(remote, local, self.shuffle, both_transform, image_transform, annotation_transform, batch_size)
```
abhi-mosaic, line 510: @knighton just a heads up, here's where I pass the device `batch_size` to the dataset, so it can coordinate it's workers to split their indices in a way such that there is only 1 incomplete batch per epoch.

knighton: Ty ty

composer/datasets/streaming/dataset.py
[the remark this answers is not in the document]
knighton: Changed to contextmanager <sassy comment about cpython being the standard>

ravi-mosaicml (commented):
(Discussed offline that this PR will be split up -- at least one for the base streaming + test cases; one or more for the streaming dataset hparam classes, and potentially a third for a tutorial page) Please re-request a review with that "refresh" button when it is ready!

abhi-mosaic:
Closing this branch as we are breaking it into a couple different PRs, starting with #950 . 

Please do not delete this branch yet as I will be cherry-picking commits into new branches.

PR #859, opened 2022-04-04
Proofreading for docs "Getting Started" section

mcneela:


docs/source/getting_started/welcome_tour.rst
```diff
@@ -122,7 +123,7 @@ Putting all the pieces together, our trainer looks something like this:
 
 .. code-block:: python
 
-    state = State(...)
+    state = State(model, max_duration, rank_zero_seed, dataloader)
```
hanlint, line 126: Can probably keep this ambiguous (`...`) since these variables were not created previously.

hanlint (approved):
LGTM, thanks for your help!

docs/source/getting_started/welcome_tour.rst
mcneela: Okay, I could also create them above? I added them because I saw the dataloader was loading `state.targets` and `state.inputs`, so that led me to wonder how I get the `State` to attach to the dataloader.

hanlint: Ah yes agreed create them above would make sense.

docs/source/getting_started/welcome_tour.rst
```diff
@@ -123,15 +123,14 @@ Putting all the pieces together, our trainer looks something like this:
...
-    model = SomeComposerModel() # a Composer model which subclasses `ComposerModel`
-    max_duration = Time(10, TimeUnit.EPOCH) # max duration to train for
-    rank_zero_seed = 0
-    dataloader = (...) # load some dataset. Initialize the same way you would `torch.utils.DataLoader`.
-
-    state = State(model, max_duration, rank_zero_seed, dataloader)
+    from composer import Time
+
+    state = State(
+        model=your_model,  # ComposerModel
```
mcneela, line 129: Are these valid kwargs? In the docs [here](https://docs.mosaicml.com/en/v0.5.0/api_reference/composer.core.state.html#composer.core.state.State) they are not listed as kwargs.

docs/source/getting_started/welcome_tour.rst
```diff
@@ -51,7 +51,7 @@ training loop:
             optimizer.step()
             optimizer.zero_grad()
 
-This works, and is recommend if you want to quickly modify an existing training loop to use our implementation of
+This works and is recommended if you want to quickly modify an existing training loop to use our implementation of
 MixUp! However, the goal of the Composer library is to be able to rapidly experiment with different combinations of
```
hanlint, line 55:
```suggestion
MixUp! However, the goal of Composer is to be able to rapidly experiment with different combinations of
```

PR #860, opened 2022-04-06
CLI Launcher supports environment variables and tells fewer lies

jbloxham:
1. Multi-node values for the CLI launcher can be specified via environment variables, akin to other common distributed launching tools.
2. A lot more validation to make sure that invalid multi-node configurations cannot be provided.
3. Only one of `node_rank` and `base_rank` needs to be provided when there's a constant amount of processes per node.
4. Another pass on the (internal) documentation for the launcher. In particular, the documentation is simplified for most users by distinguishing the multi-node options from everything else, and not documenting the "actual" placeholder defaults.

`composer -h` now returns:

```
% composer -h
usage: composer [-h] [-n NPROC] [--stdout STDOUT] [--stderr STDERR] [-v] [-m]
                [--world_size WORLD_SIZE] [--base_rank BASE_RANK] [--node_rank NODE_RANK]
                [--master_addr MASTER_ADDR] [--master_port MASTER_PORT]
                training_script ...

Utility for launching distributed machine learning jobs.

optional arguments:
  -h, --help            show this help message and exit
  --stdout STDOUT       Format string for a filename to dump the STDOUT from the non-local-rank-
                        zero processes. The local rank zero process will be piped through to
                        STDOUT. The available format variables are: '{rank}', '{local_rank}',
                        '{world_size}', '{node_rank}', and '{local_world_size}'. If specified, it
                        is recommended to include '{rank}' or '{local_rank}' in the filename so
                        each rank will write to its own file. By default, the STDOUT of the non-
                        local-rank-zero processes is discarded; instead, use the FileLogger within
                        Composer. This logger captures and saves the STDOUT of each process.
  --stderr STDERR       Format string for a filename to dump the STDERR from the non-local-rank-
                        zero processes. The local rank zero process will be piped through to
                        STDERR. The available format variables are: '{rank}', '{local_rank}',
                        '{world_size}', '{node_rank}', and '{local_world_size}'. If specified, it
                        is recommended to include '{rank}' or '{local_rank}' in the filename so
                        each rank will write to its own file. By default, the STDERR of the non-
                        local-rank-zero processes is discarded; instead, use the FileLogger within
                        Composer. This logger captures and saves the STDERR of each process.
  -v, --verbose         If set, print verbose messages
  -m, --module_mode     If set, run the training script as a module instead of as a script.

required arguments:
  -n NPROC, --nproc NPROC
                        The number of processes to launch on this node. Overrides env var
                        LOCAL_WORLD_SIZE.
  training_script       The path to the training script used to initialize a single training
                        process. Should be followed by any command-line arguments the script
                        should be launched with.
  training_script_args  Any arguments for the training script, given in the expected order.

multi-node arguments:
  These arguments generally only need to be set when training in a multi-node environment, i.e.
  when the world_size is bigger than nproc.

  --world_size WORLD_SIZE
                        The total number of processes to launch across all nodes. Setting this to
                        a value greater than nproc indicates a multi-node environment. Overrides
                        env var WORLD_SIZE. Defaults to nproc.
  --base_rank BASE_RANK
                        The rank of the lowest ranked process to launch on this node. Specifying a
                        base_rank B and an nproc N will spawn processes with global ranks [B, B+1,
                        ... B+N-1]. In a multi-node environment, at least one of base_rank and
                        node_rank must be specified. If only one of base_rank and node_rank are
                        provided, it is assumed that all nodes have the same amount of processes,
                        and that the two values are related as node_rank * nproc = base_rank. If
                        this is not the case, both base_rank and node_rank must be provided.
                        Overrides env var BASE_RANK. Defaults to 0 in a single-node environment.
  --node_rank NODE_RANK
                        The rank of this node. See base_rank for information on when this must be
                        provided. Overrides env var NODE_RANK. Defaults to 0 in a single-node
                        environment.
  --master_addr MASTER_ADDR
                        The FQDN of the node hosting the C10d TCP store. For single-node
                        operation, this can generally be left as 127.0.0.1. Overrides env var
                        MASTER_ADDR. Defaults to 127.0.0.1 in a single-node environment.
  --master_port MASTER_PORT
                        The port on the master hosting the C10d TCP store. If you are running
                        multiple trainers on a single node, this generally needs to be unique for
                        each one. Overrides env var MASTER_PORT. Defaults to a random free port in
                        a single-node environment.
```

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -21,25 +21,23 @@ log = logging.getLogger(__name__)
 def get_parser():
     parser = ArgumentParser(description="Utility for launching distributed machine learning jobs.")
 
-    parser.add_argument("-n",
-                        "--nproc",
-                        type=int,
-                        required=True,
-                        help="The number of processes to launch on this node. Required.")
-    parser.add_argument("--world_size",
-                        type=int,
-                        default=-1,
-                        help="The total number of processes to launch on all "
-                        "nodes. Set to -1 to default to nproc (single-node operation). "
-                        "Defaults to -1.")
+    required_args = parser.add_argument_group("required arguments")
+
+    required_args.add_argument("-n",
+                               "--nproc",
+                               type=int,
+                               default=None,
+                               help="The number of processes to launch on this node. Overrides env var "
+                               "LOCAL_WORLD_SIZE.")
+
     parser.add_argument(
         "--stdout",
         type=str,
         default=None,
-        help=("Format string for a filename to dump the STDOUT from the non-local-rank-zero processes."
+        help=("Format string for a filename to dump the STDOUT from the non-local-rank-zero processes. "
               "The local rank zero process will be piped through to STDOUT. The available format variables are: "
               "'{rank}', '{local_rank}', '{world_size}', '{node_rank}', and '{local_world_size}'. If specified, "
-              "it is recommended to include '{rank}' or '{local_rank}' in the filename so each rank will write to its"
+              "it is recommended to include '{rank}' or '{local_rank}' in the filename so each rank will write to its "
               "own file. By default, the STDOUT of the non-local-rank-zero processes is discarded; instead, use the "
               "FileLogger within Composer. This logger captures and saves the STDOUT of each process."),
     )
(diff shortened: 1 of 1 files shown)


composer/cli/launcher.py
```diff
@@ -107,30 +120,87 @@ def parse_args():
     parser = get_parser()
 
     args = parser.parse_args()
-    if args.world_size == -1:
+
+    env_var_mapping = {
```
ravi-mosaicml, line 124: Thoughts about setting the defaults in the argparse above? E.g. something like `default=int(os.environ.get('BASE_RANK', '0')`? For world size, could default it to None, since the default is thr value of nproc. Just not a big fan of `getattr` and `setattr` inside the for loop.

composer/cli/launcher.py
```diff
@@ -107,30 +120,87 @@ def parse_args():
...
-            raise ValueError("node_rank not specified, but unable to infer since nodes appear to "
-                             "have different amounts of processes.")
-        args.node_rank = args.base_rank // args.nproc
+    if args.world_size < args.nproc:
+        raise ValueError(f"world_size({args.world_size}) cannot be less than nproc({args.nproc})")
 
-    return args
+    is_multinode = args.world_size > args.nproc
+
+    if is_multinode:
```
ravi-mosaicml, line 149: Could this validation be moved to the `launch_procs` function below? If someone calls launch_procs functionally, then it would be nice if they could benefit from this validation.

ravi-mosaicml (commented):
Lgtm, thanks! Just some stylistic comments. If you have the bandwidth, it would also be great to add some test cases to the launcher script, but not crucial.

composer/cli/launcher.py
jbloxham: Not a huge fan of putting complicated defaults in, since these will be displayed verbatim in our docs. I'm just going to rewrite this to not use a for loop, and instead be a bit verbose.

Averylamp (approved):
lgtm

PR #862, opened 2022-04-06
LICENSE: copyright and cleanup

kobindra:
Add copyright, remove boilerplate instructions, remove empty line at the top. Additional goal is to make the Apache 2.0 license automatically detected by GitHub (currently says "View license" instead).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,3 @@
-
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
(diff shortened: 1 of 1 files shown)


hanlint (approved):
Good find, thanks!

mbrukman:
FWIW, I think that the reason that you were seeing "View license" on this repo before this PR, and this is still the case after this change was made, is because you have another file that looks like a copyright / license override, namely: https://github.com/mosaicml/composer/blob/dev/COPYRIGHT_HEADER which confuses the license detector script. In other words, it was not due to having the appendix on the Apache 2.0 license, which you should actually keep so that it matches the full Apache 2.0 license text.

Please consider making both of the following changes:

1. removing the `COPYRIGHT_HEADER` file (or if you need it, moving it outside of the root of the repository), and 
2. reverting this commit (https://github.com/mosaicml/composer/commit/72177414af8f284f6cc137cb29978b443bc37756) to make the license match the Apache 2.0 license text verbatim (https://www.apache.org/licenses/LICENSE-2.0.txt)
 
and GitHub should show "Apache 2.0" instead of "View license" for your repository.

PR #863, opened 2022-04-06
rename LICENSE_HEADER to fix GH license detection

kobindra:


ravi-mosaicml:
The license_header file is used by to ensure that each file has a proper license header. Could you please elaborate on what this is being replaced by?

kobindra:
The goal state is for GitHub's automatic license detection to identify this repo as being licensed with the Apache 2.0 license. The code responsible for detection is here: https://github.com/licensee/licensee
At first reading of that repo, it appears that all files that start with `LICENSE` are parsed, and that the presence of `LICENSE_HEADER` is causing the license detection to fail.

My initial attempt at resolving this was to insert the copyright statement into the LICENSE file (https://github.com/mosaicml/composer/blob/7162ddf70cec98d84d530a7e20c4334272991f47/LICENSE#L178)
which would then obviate the need for the LICENSE_HEADER file if it were only needed in that one place. However, I did not account for how the LICENSE_HEADER file was being used for multiple insertions throughout the codebase.

Perhaps we can rename `LICENSE_HEADER` to something like `MOSAICML_LICENSE_HEADER` and update the automation to read that file for insertion -- wdyt?

kobindra:
`COPYRIGHT_HEADER` it is -- thanks, Bandish!

hanlint (approved):

PR #865, opened 2022-04-07
Make wandb tags optional

siriuslee:
WandB tags can be specified using the `WANDB_TAGS` environment variable. This gets ignored by the logger by default because an empty string is passed instead. This PR changes that behavior by passing `None` by default.


composer/loggers/logger_hparams.py
```diff
@@ -117,7 +117,9 @@ class WandBLoggerHparams(LoggerDestinationHparams):
         doc="Whether to flatten the config, which can make nested fields easier to visualize and query.", default=False)
 
     def initialize_object(self) -> WandBLogger:
-        tags = list(set([x.strip() for x in self.tags.split(",") if x.strip() != ""]))
+        tags = None
+        if self.tags:
```
ravi-mosaicml, line 121: To distinguish between the empty string and setting it to None:

```suggestion
        if self.tags is not None:
```

ravi-mosaicml (approved):
Lgtm

composer/loggers/logger_hparams.py
ravi-mosaicml: (Alternatively, would be OK with leaving the yaml / hparams as-is and just having this change in the initialized object function...realize it could be confusing to have this distinction)

siriuslee: This issue only occurs when using hparams. If the user initializes in the object function on their own, they just have to not pass a `tags` field in `init_params`.

Issue #15, opened 2021-10-13
Implement `predict()` in `Trainer`


ajaysaini725:
The `Trainer` class should have a predict() function as a convenience for a user who wants to run inference on a trained model.

PR #866, opened 2022-04-07
`Trainer.predict()` method

QiyaoWei:
Hi! This is Qiyao Wei, an applicant to Mosaic ML. I thought I would start my contribution by tackling one of the "good first issues", specifically #15 . It makes sense that we want a predict() method, but are there specifications for what the input and output for this method is? For example, do we expect the user to input a dataloader or just a batch of data? Are we outputting the softmax logit values or one of the ten classification classes? Please bear with me as I am getting familiar with the Trainer interface! I would much appreciate comments both to my questions and my code!

----

Implementation overview:

* Added `Trainer.predict(dataloader, subset_num_batches)`.
* Added events for prediction. Prediction events match evaluation events.
* Added test cases.
* Fixed a bug in `State` where `_dataloader_len` was not being cleared when updating the dataloader.


hanlint:
Thanks @QiyaoWei, we'll take a look at this soon!

composer/trainer/trainer.py
```diff
@@ -1316,6 +1316,20 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
+        self.state.model.eval()
+        for batch_idx, self.state.batch in enumerate(
+            itertools.islice(loader, batchsize)):
+
+            self.state.batch = self._device.batch_to_device(self.state.batch)
+            if self.deepspeed_enabled:
+                self.state.batch = _fix_batch_precision_for_deepspeed(self.state.batch, self.state.precision)
+
+            with torch.no_grad():
+                outputs, targets = self._original_model.validate(self.state.batch)
```
ravi-mosaicml, line 1331: For prediction, it is typical to call the `forward()` function instead of the validation function.

```suggestion
                outputs, targets = self.state.model(self.state.batch)
```

composer/trainer/trainer.py
```diff
@@ -1316,6 +1316,20 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
+        return outputs, targets
```
ravi-mosaicml, line 1332: Similar to `train` and `eval`, I don't think it is necessary for `predict` to return anything.

ravi-mosaicml, line 1332: Before returning, could you restore whether the model was in training mode? (See `eval()` for how this is captured at the start of the function, and then set again at the end.)

ravi-mosaicml (commented):
Hi there! Thanks for working on this issue, and apologies for the delay on getting back to you!

The overall logic looks great! :rocket: (Please see the comments below). At a higher level:

1. Could you add in [events](https://github.com/mosaicml/composer/blob/e5d22f401ac7160bfca2c0633e94d177912fe0fa/composer/core/event.py) for predictions? I think mirroring the eval events would work here -- e.g. `PREDICT_START`, `PREDICT_BATCH_START`, `PREDICT_BEFORE_FORWARD`, .... These events will need to be defined in the [Event Enum](https://github.com/mosaicml/composer/blob/e5d22f401ac7160bfca2c0633e94d177912fe0fa/composer/core/event.py#L10), hooks will need to be added to the [Callback class](https://github.com/mosaicml/composer/blob/d025644e5786a84ac7edebbc6284813df1503c98/composer/core/callback.py#L18), and then `self.engine.run_event(...)` calls will need to be added to the `predict()` function.
2. Could you add a test case to test the predict function?

ravi-mosaicml:
> I notice that the line in your comment outputs, targets = self.state.model(self.state.batch) should actually be outputs = self.state.model.forward(self.state.batch) because of the way the model is written. 

See https://discuss.pytorch.org/t/any-different-between-model-input-and-model-forward-input/3690; PyTorch recommends calling the model rather than the underlying forward method.

For the test case, there wouldn't be any output to verify (since the input would be random) nor would there be metrics (since the output target class generally isn't know). It would be good to verify that the predict events are called the expected number of times -- see https://github.com/mosaicml/composer/blob/dev/tests/trainer/test_checkpoint.py#L57-73 as an example. Perhaps this test case can go in [tests/trainer/test_trainer.py](https://github.com/mosaicml/composer/blob/56c37b48459d6027ad23a516a3cc45b9aa32f1ed/tests/trainer/test_trainer.py) or a new file, say, `test_predict.py`.

QiyaoWei:
Ah I see. Is the most recent commit something you had in mind? I started a new `test_predict.py` file, but I'm not sure whether it would be beneficial to combine that code into `test_trainer.py`?

ravi-mosaicml:
> Ah I see. Is the most recent commit something you had in mind? I started a new `test_predict.py` file, but I'm not sure whether it would be beneficial to combine that code into `test_trainer.py`?

Yes, that is great. Could you please fix up the test case so it passes? We can help with any pyright or linting checks that are failing.

QiyaoWei:
hmmm the tests are taking very long, because I made my last commit 2 days ago. Is that normal? @ravi-mosaicml

ravi-mosaicml:
> hmmm the tests are taking very long, because I made my last commit 2 days ago. Is that normal? @ravi-mosaicml Or is it because you would have to review my commit first before it can be tested?

I just updated the latest dev branch into your PR. This should hopefully fix the build!
Jenkins should be building changes to your PR, and you can see the test results under the "Tests" check. Please let us know if this is not working!

QiyaoWei:
Thanks a lot! So I'm not sure what this error message means since I did not get it locally (I should also mention that I took inspiration from [this line](https://github.com/mosaicml/composer/blob/bb577ef5dd751eff5219c5fb082495b97560fe00/tests/trainer/test_trainer.py#L36))---
```
self = <tests.trainer.test_predict.TestTrainerPredict object at 0x7f1f73d8e390>

    @pytest.fixture
    def config(self):
        return {
            'model': SimpleModel(),
>           'train_dataloader': DataLoader(dataset=RandomClassificationDataset()),
            'eval_dataloader': DataLoader(dataset=RandomClassificationDataset()),
            'max_duration': '2ep',
        }
E       TypeError: DataLoader() takes no arguments

/home/jenkins/agent/workspace/scratch/command2/tests/trainer/test_predict.py:74: TypeError
```

ravi-mosaicml:
> Thanks a lot! So I'm not sure what this error message means since I did not get it locally (I should also mention that I took inspiration from [this line](https://github.com/mosaicml/composer/blob/bb577ef5dd751eff5219c5fb082495b97560fe00/tests/trainer/test_trainer.py#L36))---
> 
> ```
> self = <tests.trainer.test_predict.TestTrainerPredict object at 0x7f1f73d8e390>
> 
>     @pytest.fixture
>     def config(self):
>         return {
>             'model': SimpleModel(),
> >           'train_dataloader': DataLoader(dataset=RandomClassificationDataset()),
>             'eval_dataloader': DataLoader(dataset=RandomClassificationDataset()),
>             'max_duration': '2ep',
>         }
> E       TypeError: DataLoader() takes no arguments
> 
> /home/jenkins/agent/workspace/scratch/command2/tests/trainer/test_predict.py:74: TypeError
> ```

Could you try `from torch.utils.data import DataLoader` instead of `from composer.datasets import DataLoader`? The `DataLoader` in Composer is a protocol, not an alias for the PyTorch dataloader (This confusion will be addressed via #424).

composer/trainer/trainer.py
```diff
@@ -1316,6 +1316,20 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
 
         if self.deepspeed_enabled:
             self.state.deepspeed_model.step()
+            
+    def predict(self, loader, batchsize):
```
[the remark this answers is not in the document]
QiyaoWei: please point out if I missed something!

composer/trainer/trainer.py
QiyaoWei: please point out if I missed something!

composer/trainer/trainer.py
QiyaoWei: please point out if I missed something!

QiyaoWei: please point out if I missed something!

ravi-mosaicml:
> @ravi-mosaicml can you kindly take a look at this error? I'm not sure if I am misunderstanding something, but it only gives me an error link without allowing me access

The error seems to be flaky. I just merged dev into your PR, which should cause it to re-build. Let's see if it passes.

QiyaoWei:
@ravi-mosaicml after a few more commits all tests passed now :)

ravi-mosaicml (approved):
LGTM

PR #867, opened 2022-04-07
Add missing ffcv dependency in pytorch_vision docker image

dskhudia:
cupy was missing. Adding it in the PyTorch_vision image.

### DIFF
--- a/docker/pytorch/Dockerfile
+++ b/docker/pytorch/Dockerfile
@@ -179,6 +179,7 @@ ARG PYTHON_VERSION
 RUN sudo pip${PYTHON_VERSION} install --no-cache-dir \
     ffcv==0.0.3 \
     'opencv-python>=4.5.5.64,<4.6' \
-    'numba>=0.55.0,<0.56' && \
+    'numba>=0.55.0,<0.56' \
+    cupy-`echo ${CUDA_VERSION_TAG} | sed "s/cu/cuda/g"`>=10.2.0 && \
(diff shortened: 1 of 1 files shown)


docker/pytorch/Dockerfile
```diff
@@ -179,6 +179,7 @@ ARG PYTHON_VERSION
 RUN sudo pip${PYTHON_VERSION} install --no-cache-dir \
     ffcv==0.0.3 \
     'opencv-python>=4.5.5.64,<4.6' \
-    'numba>=0.55.0,<0.56' && \
+    'numba>=0.55.0,<0.56' \
+    'cupy-`echo ${CUDA_VERSION_TAG} | sed "s/cu/cuda/g"`' && \
```
bandish-shah, line 183: Single quotes won't let you interpolate so this will be taken as a literal... think you need to do:

```suggestion
    cupy-`echo ${CUDA_VERSION_TAG} | sed "s/cu/cuda/g"` && \
```

bandish-shah: [nit] Do you need to pin the version?  Looks like we're building the `vision` image with `cuda113` so may want to pin or add a minimum version check to something here:
https://pypi.org/project/cupy-cuda113/#history

This will complicate your version selection logic in the Dockerfile though, mostly just want you to think about whether or not it's ok to let this package silently update to the latest version should the image be rebuilt at some point.

bandish-shah (changes requested):
Fix CI, noted possible root cause.

docker/pytorch/Dockerfile
dskhudia: I think it's ok to let it update to latest. FFCV is not using cupy extensively.

dskhudia: Added a min version check.

bandish-shah (approved):
LGTM

PR #869, opened 2022-04-07
Initialize distributed before dataloaders are created

dskhudia:
Initialize torch distributed before Dataloaders are created. FFCV Dataloaders make some calls to some functions from dist namespace. 

Test:
`composer -n 8 examples/run_composer_trainer.py -f composer/yamls/models/resnet50.yaml --max_duration 2ep --train_batch_size 32 --datadir /localdisk/ImageNet --loggers progress_bar`

### DIFF
--- a/composer/trainer/trainer_hparams.py
+++ b/composer/trainer/trainer_hparams.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import datetime
 import logging
 import os
 import textwrap
@@ -473,6 +474,11 @@ class TrainerHparams(hp.Hparams):
 
         # devices and systems
         device = self.device.initialize_object()
+
+        # initialize distributed early so that it's already initialized when dataloders
+        # are created.
+        if dist.get_world_size() > 1:
+            dist.initialize_dist(device.dist_backend, datetime.timedelta(seconds=self.dist_timeout))
 
         seed = self.seed if self.seed else reproducibility.get_random_seed()
         # need to set seed before model initialization for determinism


composer/trainer/trainer.py
```diff
@@ -574,7 +574,8 @@ def __init__(
         if self.deepspeed_enabled or dist.get_world_size() > 1:
             # deepspeed requires torch.distributed to be initialized, even if the world size is 1
             # distributed is always required with multi-rank training
-            dist.initialize_dist(self._device.dist_backend, datetime.timedelta(seconds=dist_timeout))
+            if not dist.is_initialized():
```
jbloxham, line 577: No need to worry about this check - `dist.initialize_dist` is idempotent.

jbloxham (approved):
This looks fine to me, and it will unblock FFCV.

But, this doesn't do help users who are initializing the trainer not through hparams but through direct initialization. This is totally out of scope for this PR, but @ravi-mosaicml are you still on board with providing a `preinit` function for those users to call?

dskhudia:
> This looks fine to me, and it will unblock FFCV.
> 
> But, this doesn't do help users who are initializing the trainer not through hparams but through direct initialization. This is totally out of scope for this PR, but @ravi-mosaicml are you still on board with providing a `preinit` function for those users to call?

They can call dist.initialize_dist? Is there any other initialization you were thinking of lumping into preinit?

jbloxham:
> > This looks fine to me, and it will unblock FFCV.
> > But, this doesn't do help users who are initializing the trainer not through hparams but through direct initialization. This is totally out of scope for this PR, but @ravi-mosaicml are you still on board with providing a `preinit` function for those users to call?
> 
> They can call dist.initialize_dist? Is there any other initialization you were thinking of lumping into preinit?

I think there's a world in which `preinit` actually takes the form of a context wrapper. We have some ideas for how model initialization could be done more efficiently that would require this sort of approach.

PR #871, opened 2022-04-08
Fixing version number

ajaysaini725:
Version number was never bumped to 0.5.0 upon release. Just updating it now for the next release to avoid that in the future :)


bandish-shah (approved):
LGTM thanks for catching this.

PR #875, opened 2022-04-09
Switch mixup events to avoid grad accum issues

coryMosaicML:
This PR fixes the issue with mixup + grad_accum > 1 by switching it to operate on the batch after it has been split. This should be math equivalent to the previous behavior.

Test runs currently gated due to lack of GPUs
[ResNet50-I1k with grad_accum=2 and interpolate_loss=True Accuracy 77.27%](https://wandb.ai/mosaic-ml/cory-mixup-changes/runs/nanpf6er)


growlix (approved):
LGTM

PR #878, opened 2022-04-09
Proofreading docs through "Callbacks" section

mcneela:


docs/source/trainer/checkpointing.rst
```diff
@@ -129,17 +128,17 @@ will continue training from where the checkpoint left off.
     )
     trainer.fit()
 
-The above code will load the checkpoint from epoch 25, and continue training
+The above code will load the checkpoint from epoch 25 (if it exists, otherwise epoch 1) and continue training
```
hanlint, line 131: this shuold be reverted -- the code logic in the setup is just to mock things so the test passes

hanlint (changes requested):
One suggested change above.

docs/source/trainer/checkpointing.rst
mcneela: got it

hanlint (approved):

PR #879, opened 2022-04-09
Remove the `Timer`; replace with `Timestamp`

### LINKED ISSUE #669: Unify the `Timer` and `TimeStamp`
The `Timer` object is a full-featured Python class with many convenience methods, whereas the `TimeStamp` object is simply a named tuple. Instead, let's convert the `TimeStamp` into a full featured Python class, and add all of the comparison methods and getters currently on the `Timer` to it. Then, internally, the `Timer` will use a `TimeStamp` object as its internal state to represent the current training time.

This will allow for queries like this:

```python
if timestamp < Time(10, "ep"): do_something()  # easy comparisons
timestamp.get(time.unit)  # get the value of a dynamic unit without getattr
```

The Timer should still maintain all of its comparison methods, which should now call the relevant methods on the underlying timestamp object.

pavithranrao:
Based on the recommendation provided in the task #669. This PR has following changes:
1. Convert the `Timestamp` into a full featured Python class.
2. Add all of the comparison methods and getters currently on the `Timer` to `Timestamp`.
3. The Timer will use a `Timestamp` object as its internal state to represent the current training time.
4. Refractor `State` to use the new `Timestamp` and `Timer` objects

Closes #669


composer/core/time.py
```diff
@@ -542,7 +560,25 @@ def __ge__(self, other: object):
         self_counter = self.get(other.unit)
         return self_counter >= other
 
-    def get_timestamp(self):
+
+class Timer(Serializable):
```
ravi-mosaicml, line 564: Could you remove the `Timer` class? I don't think this would be needed anymore.

composer/core/time.py
```diff
@@ -356,7 +356,7 @@ def from_timestring(cls, timestring: str) -> Time:
         return cls(value, unit)
 
 
-class Timer(Serializable):
+class Timestamp(Serializable):
```
ravi-mosaicml, line 359: While we originally envisioned that the `Timer` class would use a `Timestamp` internally, I do appreciate how binding the latest `Timestamp` to the state, rather than binding the `Timer` to the state, simplifies the logic wherever the timer is used (i.e. no need to call `.get_timestamp()` anymore). Thanks for doing it this way!

One request: Since the `Timestamp` is supposed to be treated like an immutable object (i.e. once it is created, it is never modified), would it be possible to remove the methods `on_epoch_complete` and `on_batch_complete` from the timestamp, and instead hard-code this logic directly into the Trainer? The trainer can create a new Timestamp object, and bind the new object to the State, instead of modifying the existing Timestamp object.


For example, when the epoch is complete, the trainer can set `state.timestamp = Timestamp(epoch=state.timestamp.epoch + Time(1, TimeUnit.EPOCH), batch=state.timestamp.batch, sample=state.timestamp.sample, token=state.timestamp.token, batch_in_epoch=Time(0, TimeUnit.BATCH), ...).` The logic would be parallel for where `on_batch_complete` is currently called.

composer/core/time.py
```diff
@@ -367,14 +367,21 @@ class Timer(Serializable):
...
-    def __init__(self):
-        self._epoch = Time(0, TimeUnit.EPOCH)
-        self._batch = Time(0, TimeUnit.BATCH)
-        self._sample = Time(0, TimeUnit.SAMPLE)
-        self._token = Time(0, TimeUnit.TOKEN)
-        self._batch_in_epoch = Time(0, TimeUnit.BATCH)
-        self._sample_in_epoch = Time(0, TimeUnit.SAMPLE)
-        self._token_in_epoch = Time(0, TimeUnit.TOKEN)
+    def __init__(self,
+                 epoch: Time = Time(0, TimeUnit.EPOCH),
```
ravi-mosaicml, line 371: Would it be possible to remove the default values? While 0 is a reasonable default (most of the time), I think it would be preferable for all arguments to be explicitly specified (similar to how the current Timestamp class works). I am mainly concerned about the case where we attempt to create a new timestamp from an existing timestamp, but forget to copy over one of the fields.

tests/test_time.py
```diff
@@ -58,13 +58,22 @@ def test_time_repr():
 
 def test_timer():
```
ravi-mosaicml, line 59: This test case can be removed, as the `Timer` would also be removed.

ravi-mosaicml (commented):
Hi there! Apologies for the delay on the PR review, but thanks so much for working on this issue. First off, awesome work :rocket:

The main piece of feedback we have is to remove the `on_epoch_complete` and `on_batch_complete` methods from the new `Timestamp` object, and instead incorporate that logic directly into the trainer. This will likely require a few of the test cases (that currently call these methods) to be reworked. After that, the PR should be good to merge in!

tests/test_time.py
pavithranrao: Makes sense, will remove `Timer` and its usages.

composer/core/time.py
pavithranrao: Sure, thanks for providing the rationale. I will try and implement your suggestions.

composer/core/time.py
pavithranrao: Thank you so much for the comments and suggestions. Very reasonable and well put. Sorry had a busy week, haven't got time to resume the work, I will try and get address your comments and update the PR soon. Thanks again!

ravi-mosaicml:
Hi @pavithranrao!

Curious about the status of this PR? We're eager to merge it in :)

Let us know if you need any help dealing with the merge conflicts / getting tests to pass / or general help finishing it up. Thanks!

pavithranrao:
Hey @ravi-mosaicml I have been meaning to make the changes. I will try and get the comments addressed this weekend. If not will let you know by end of this weekend. Thanks for your patience!

pavithranrao:
Hey @ravi-mosaicml Sorry last few days have been very busy, I understand this change might be critical for the project. Please feel free to commandeer this if need be. Thanks!

ravi-mosaicml:
> Hey @ravi-mosaicml Sorry last few days have been very busy, I understand this change might be critical for the project. Please feel free to commandeer this if need be. Thanks!

No worries! Thanks for starting it -- this will definitely help clean things up, and we're planning other features on time of this change (wall clock time tracking, evaluation time tracking). No problem; we will take it from here to get it in.

composer/core/time.py
ravi-mosaicml: Since these methods were used throughout tests, and could be useful outside of the Mosaic trainer, I kept the logic of `on_batch_complete` and `on_epoch_complete` as part of the `Timestamp`, but renamed these methods to `to_next_batch` and `to_next_epoch`, respectively, and have them return a copy of the `Timestamp` rather than keep it in place.

composer/core/time.py
ravi-mosaicml: Resolved; removed the `Timer` class.

composer/core/time.py
ravi-mosaicml: After further thought, I left the default values here, but added a `copy` method to get a copy of the timestamp with some values changed. That will handle the case where we create a new timestamp from the existing one, and only want to update some of the fields.

tests/test_time.py
ravi-mosaicml: Because of the `to_next_batch` and `to_next_epoch` methods, I ended up keeping (but refactoring) this test.

ravi-mosaicml (approved):
Thanks for starting this -- let's merge it in!

PR #880, opened 2022-04-11
Symlinks for Graceful Resumption

mvpatel2000:
Adds effective symlinks, which are necessary for object stores. Existing tests (the one edited to include `.symlink`) file extension should provide some coverage for local files, deepspeed tars, etc

What's the best way to test for the object store scenario?

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -242,7 +242,7 @@ def _download_checkpoint(
     rank_n_checkpoint_filepath = os.path.join(node_checkpoint_folder, f"rank{dist.get_global_rank()}_checkpoint")
     extracted_checkpoint_folder = None
     extracted_rank_n = False
-    if is_tar(path):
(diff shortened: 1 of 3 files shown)


composer/utils/checkpoint.py
```diff
@@ -202,7 +264,7 @@ def _download_checkpoint(
             # every NODE needs the GLOBAL rank zero checkpoint
             path = _format_path_with_rank_zero(path)
             get_file(destination=rank_zero_checkpoint_filepath,
-                     path=path,
+                     path=_follow_symlink(_format_path_with_rank_zero(path), node_checkpoint_folder, 0),
```
mvpatel2000, line 267: Previously, this didn't have `_format_path_with_rank_zero`, so it seems like it would have used a path which still had `{rank}` in it. It looked like a bug to me so I've changed it to format -- would be great if someone who is more familiar with this code can confirm

mvpatel2000:
Redoing in https://github.com/mosaicml/composer/pull/887 -- this is very off original design so easier to scrap lol

PR #882, opened 2022-04-11
Graceful resumption params

mvpatel2000:
Note: This will be rebased and merged after https://github.com/mosaicml/composer/pull/880 is merged

Reworks parameters based on graceful resumption plan. Separates out checkpoint and weights path args.

Open question on default values: how do we set path to default value of save folder? If we set load path to a default value when unspecified, it will load from the junk weights I have sitting in the folder if I try doing overwrite save folder

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -260,6 +260,10 @@ class Trainer:
             Then, ``load_path`` should be set to ``my_model/ep1-rank{rank}.tar``, and all ranks will load the
             correct state.
 
+            If using graceful resumption, this path will be used to load weights following preemption. Since missing
+            checkpoints are ignored with a warning, to enable graceful resumption, set ``load_path`` to
+            ``{save_folder}/{save_latest}``.
+
             If ``None`` then no checkpoint will be loaded. (default: ``None``)
         load_object_store (ObjectStore, optional): If the ``load_path`` is in an object store
             (i.e. AWS S3 or Google Cloud Storage), an instance of :class:`.ObjectStore` which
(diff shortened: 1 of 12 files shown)


composer/trainer/trainer.py
```diff
@@ -306,8 +312,41 @@ class Trainer:
 
                 trainer.engine.close()
 
-        load_weights_only (bool, optional): Whether or not to only restore the weights from the checkpoint without
-            restoring the associated state. Ignored if ``load_path`` is ``None``. (default: ``False``)
+        load_only_weights_path (str, optional): The path format to an existing checkpoint file where only weights
```
ravi-mosaicml, line 315: nit: Can we call this `initial_weights_path`? I just updated the design doc to reflect this.

composer/utils/checkpoint.py
```diff
@@ -119,6 +121,8 @@ def load_checkpoint(
             restoring the associated state. (default: ``False``)
         strict_model_weights (bool, optional): Whether or not to force that the checkpointed weights must exactly
             match the model weights. (default: ``False``)
+        load_ignore_missing_checkpoint (bool, optional): Whether or not to ignore missing files and emit a warning or otherwise
```
ravi-mosaicml, line 124: Would it work to catch the `FileNotFoundException` exception directly in the trainer? Would prefer not to emit a warning every time.

composer/trainer/trainer.py
```diff
@@ -260,7 +260,13 @@ class Trainer:
             Then, ``load_path`` should be set to ``my_model/ep1-rank{rank}.tar``, and all ranks will load the
             correct state.
 
+            If using graceful resumption, this path will be used to load weights following preemption. To enable graceful
+            resumption, set ``load_path`` to ``{save_folder}/{save_latest}`` and set ``load_ignore_missing_checkpoint`` to ``True``.
+
             If ``None`` then no checkpoint will be loaded. (default: ``None``)
+        load_ignore_missing_checkpoint (bool, optional): Whether or not to ignore missing checkpoint on ``load_path``. If using
```
ravi-mosaicml, line 267: Thoughts about calling this flag `graceful_resumption: bool = False`?

composer/trainer/trainer.py
```diff
@@ -836,15 +876,30 @@ def __init__(
...
                                               object_store=load_object_store,
-                                              load_weights_only=load_weights_only,
+                                              load_weights_only=False,
                                               strict_model_weights=load_strict,
+                                              load_ignore_missing_checkpoint=load_ignore_missing_checkpoint,
                                               chunk_size=load_chunk_size,
                                               progress_bar=load_progress_bar)
             reproducibility.seed_all(self.state.seed)
+        # If ``load_path`` is not specified or failed, next try ``load_only_weights_path``.
+        if load_only_weights_path is not None and self._rng_state is None:
```
ravi-mosaicml, line 894: Would it work to catch catch the `FileNotFound` exception directly to make this clearer of when we will use the initial weights path. (Conditioning this on self.)rng_state is None is a bit tricky of why that is the case)

composer/trainer/trainer.py
```diff
@@ -836,15 +876,30 @@ def __init__(
         # DDP.
 
         self._rng_state = None
+        # First try to load from ``load_path``.
```
ravi-mosaicml, line 879: If `graceful_resumption` is True and `load_path` is None, can we set `load_path = f'{save_folder}/{save_latest}'`? (Maybe also emit a `log.info` for this case)

ravi-mosaicml (commented):
I will take a closer look after #880 is merged.

mvpatel2000:
Putting a pause on this -- looks like theres some potential redesign, and the base PR is being completely reworked anyways

ravi-mosaicml:
Closing as #1037 added in graceful resumption functionality

PR #884, opened 2022-04-12
Docker image with MOFED drivers

jbloxham:
Adds a new Docker container that extends the base `pytorch` container with some Mellanox OFED drivers for multinode training.

Unfortunately, I've been unable to get this to pass CI, due to apparently sporadic failures. Locally there are no problems. I've heard there's scoping for addressing these failures, so this probably won't merge until then.


.ci/Jenkinsfile
```diff
@@ -2,7 +2,7 @@ pCloud = "colo-research-01"
 gitUrl = null
 gitBranch = null
 gitCommit = null
-pTimeout = '1800' // in seconds
+pTimeout = '3600' // in seconds
```
bandish-shah, line 5: Is this needed because the `pytorch_mofed` image build actually takes longer or did you just do it because you were combating the test timeouts?

ravi-mosaicml: Can the higher timeout be docker-image specific rather than global? Would still like pytests to be 30 minutes max.

docker/pytorch/Dockerfile
```diff
@@ -184,3 +184,19 @@ RUN sudo pip${PYTHON_VERSION} install --no-cache-dir \
     cupy-`echo ${CUDA_VERSION_TAG} | sed "s/cu/cuda/g"`>=10.2.0 && \
     sudo pip${PYTHON_VERSION} install --no-cache-dir mmcv-full==1.4.4 -f https://download.openmmlab.com/mmcv/dist/${CUDA_VERSION_TAG}/${MMCV_TORCH_VERSION}/index.html && \
     sudo pip${PYTHON_VERSION} install --no-cache-dir 'mmsegmentation>=0.22.0,<0.23'
+
+
+FROM pytorch_stage AS mofed_stage
+ARG DEBIAN_FRONTEND=noninteractive
+
+###################################
+# Mellanox OFED driver installation
```
ravi-mosaicml, line 193: Will we need a vision  + mellanox combo image?

ravi-mosaicml (approved):
LGTM

docker/pytorch/Dockerfile
jbloxham: Probably not yet - I'm fine with only adding that when someone requests it. `pytorch_vision` isn't even public in Dockerhub yet, though it probably should be?

.ci/Jenkinsfile
jbloxham: I'm fine with undoing this change. Jenkins was just highly sus yesterday; I don't think this is still needed.

docker/pytorch/Dockerfile
```diff
@@ -185,3 +185,19 @@ RUN sudo pip${PYTHON_VERSION} install --no-cache-dir \
     cupy-`echo ${CUDA_VERSION_TAG} | sed "s/cu/cuda/g"`>=10.2.0 && \
     sudo pip${PYTHON_VERSION} install --no-cache-dir mmcv-full==${MMCV_VERSION} -f https://download.openmmlab.com/mmcv/dist/${CUDA_VERSION_TAG}/${MMCV_TORCH_VERSION}/index.html && \
     sudo pip${PYTHON_VERSION} install --no-cache-dir 'mmsegmentation>=0.22.0,<0.23'
+
+
+FROM pytorch_stage AS mofed_stage
```
bandish-shah, line 190: I think there's a mismatch here between what we want it to do and what it will actually do.

The way staged builds work is the specified target is actually a stop target so it will actually try running all the other stages before this stage.  It's not completely clear to me but it's documented here:
https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage

So I don't think this is going to run just `pytorch_stage` followed by `mofed_stage`, and since you're not specifying any of the arguments required for the `vision_stage` I'm guessing this will error out.

Few options to workaround:
1. Set defaults for `CUDA_VERSION_TAG`, `MMCV_VERSION`, `MMCV_TORCH_VERSION` and `PYTHON_VERSION` in the `vision_stage` in the Dockerfile or in the build matrix so the stage can at least build.  Because you're resuming `mofed_stage` from `pytorch_stage` I don't believe any of the `vision_stage` layers will be in the final image
2. Move the `mofed_stage` _before_  `vision_stage` and set defaults for the arguments
3. Build the MOFED dependencies into the `pytorch_stage` image.
4. Add another Dockerfile for the MOFED image

From a UX perspective I'm not really sure it makes sense to have separate images for single-node vs. multi-node anyways so I wouldn't be opposed to just going with option 3 and keeping the multi-stage stuff simple for now.

Options 1 and 2 are really the same problem just swapped and 4 is probably going to involve some Jenkins wrangling.

bandish-shah (changes requested):

docker/pytorch/Dockerfile
jbloxham: Yeah so I noticed this too.

I think this is actually a bug with `kaniko`, the tool we use for generating Docker images in the Jenkins process. When I just run this build with `docker build` locally, I actually do only see the `pytorch_stage` and `mofed_stage` running.

Option 1 is the workaround I was quietly shooting for; if you look at my recent commits I've been adding those defaults. I don't really care if the `kaniko` build is longer than it's supposed to be - I might even submit an issue to that repo. I think there's a bug in the defaults I'm adding, but once that's resolved the build should succeed.

Option 3 is something I think we'd discussed explicitly wanting to avoid. The Mellanox drivers are substantial in size, and I think @hanlint may have mentioned that we'd prefer to avoid including them by default?

Option 4 would work but I don't feel like Jenkins wrangling right now, but if you put a gun to my head I'll do it.

bandish-shah: FWIW the OFED packages are adding ~1.6GB of bloat to the `mosaicml/pytorch` image so may want to just consider adding these dependencies to the base images unless there's some functional reason not to vs. creating another stage target:
`REPOSITORY                TAG                                  IMAGE ID       CREATED        SIZE
mosaicml/pytorch_vision   1.11.0_cu113-python3.9-ubuntu20.04   9f2b3a75aba9   21 hours ago   14.3GB
mosaicml/pytorch          1.11.0_cu113-python3.9-ubuntu20.04   b52fc4b3663f   21 hours ago   12.5GB
mosaicml/pytorch_mofed    jamie-temp                           5d27b1a2405e   24 hours ago   14.1GB`

jbloxham: Alright well personally, I'm totally fine with adding the OFED packages to `mosaicml/pytorch`. I don't remember who it was with the objection - maybe @ravi-mosaicml ? I'll make the change, and if I get a sign off I'll merge it.

bandish-shah: > I think this is actually a bug with `kaniko`, the tool we use for generating Docker images in the Jenkins process. When I just run this build with `docker build` locally, I actually do only see the `pytorch_stage` and `mofed_stage` running.

Gotcha that would make more sense at least intuitively, the docs aren't really clear on this.

PR #887, opened 2022-04-13
Object store symlinks for graceful resumption

mvpatel2000:
Adds effective symlinks, which are necessary for object stores. Adds corresponding tests


composer/utils/object_store.py
```diff
@@ -286,7 +288,27 @@ def upload_object_via_stream(self,
...
+            with tempfile.TemporaryDirectory() as tmpdir:
+                tmppath = os.path.join(tmpdir, str(uuid.uuid4()))
+                self._provider.download_object(obj=obj,
+                                               destination_path=tmppath,
+                                               overwrite_existing=True,
+                                               delete_on_failure=True)
+                # Read object name in symlink and recurse
+                with open(tmppath) as f:
+                    symlinked_object_name = f.read()
+                    self._get_object(symlinked_object_name)
```
ravi-mosaicml, line 309:
```suggestion
                    return self._get_object(symlinked_object_name)
```

composer/utils/object_store.py
```diff
@@ -286,7 +288,27 @@ def upload_object_via_stream(self,
...
+                pass
```
ravi-mosaicml, line 310:
```suggestion
```

composer/utils/object_store.py
```diff
@@ -286,7 +288,27 @@ def upload_object_via_stream(self,
                                                 headers=headers)
 
     def _get_object(self, object_name: str):
-        return self._provider.get_object(self._container.name, object_name)
+        """Get object from object store. Recursively follow any symlinks.
+
+        Args:
+            object_name (str): The name of the object.
+        """
+        obj = self._provider.get_object(self._container.name, object_name)
```
ravi-mosaicml, line 296: if `object_name` does not exist and does not end with `'.symlink'`, thoughts about attempting to get `object_name + ".symlink"`? This would allow the caller to omit the `'.symlink'` extension when getting objects, which would clean up the user API a bit when retrieving checkpoints from object stores (e.g. the user could specify `latest` instead of `latest.symlink`)

composer/loggers/object_store_logger.py
```diff
@@ -294,6 +294,20 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
         object_name = self._object_name(artifact_name)
         self._file_upload_queue.put_nowait((copied_path, object_name, overwrite))
 
+    def log_symlink_artifact(self, state: State, log_level: LogLevel, existing_artifact_name: str,
+                             symlink_artifact_name: str, overwrite: bool):
+        # Only symlink if we're logging artifact to begin with
```
ravi-mosaicml, line 299: Could you add a bit of a description here that object stores do not natively support symlinks, so we emulate symlinks with a `.symlink` file, where this file contains the object name that it points to?

ravi-mosaicml (commented):
Overall LGTM -- main comment: re adding a new parameter for the `latest_artifact_name`, and test cases for this.

composer/callbacks/checkpoint_saver.py
```diff
@@ -347,9 +347,11 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
                                      overwrite=self.overwrite)
 
             if self.latest_filename is not None:
+                symlink_artifact_name = format_name_with_dist_and_time(self.latest_filename, logger.run_name,
```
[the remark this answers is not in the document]
mvpatel2000: 😭  oke

composer/utils/checkpoint.py
```diff
@@ -202,7 +202,7 @@ def _download_checkpoint(
             # every NODE needs the GLOBAL rank zero checkpoint
             path = _format_path_with_rank_zero(path)
             get_file(destination=rank_zero_checkpoint_filepath,
-                     path=path,
+                     path=_format_path_with_rank_zero(path),
```
mvpatel2000, line 205: I assume this was bug? Seems like path should be formatted -- would like confirmation

mvpatel2000:
Looks like param description for `save_artifact_name` was missing in trainer -- would like some proof reading on that since I'm not as comfortable with the the description I wrote of it

composer/utils/checkpoint.py
ravi-mosaicml: It is formatted above on line 203

composer/trainer/trainer.py
```diff
@@ -331,10 +331,22 @@ class Trainer:
 
             .. seealso:: :class:`~.CheckpointSaver`
 
+        save_artifact_name (str, optional): A format string describing how to name checkpoints in loggers.
+            This parameter has no effect if ``save_folder`` is ``None``.
+            (default: ``"{run_name}/checkpoints/ep{epoch}-ba{batch}-rank{rank}"``)
```
ravi-mosaicml, line 336: Same comments re: the docstring in checkpoint_saver

composer/utils/object_store.py
```diff
@@ -286,7 +288,36 @@ def upload_object_via_stream(self,
                                                 headers=headers)
 
     def _get_object(self, object_name: str):
-        return self._provider.get_object(self._container.name, object_name)
+        """Get object from object store. Recursively follow any symlinks. If an object does not exist, automatically
+        check if it is a symlink by appending ``.symlink``.
```
ravi-mosaicml, line 292: nit

```suggestion
        checks if it is a symlink by appending ``.symlink``.
```

composer/utils/object_store.py
```diff
@@ -286,7 +288,36 @@ def upload_object_via_stream(self,
...
+
+        Args:
+            object_name (str): The name of the object.
+        """
+        obj = None
+        try:
+            obj = self._provider.get_object(self._container.name, object_name)
+        except Exception as e:
```
ravi-mosaicml, line 300: At one point in time, `libcloud` was an optional dependency, which is why we did the conditional comparison. But now it is included in core.

```suggestion
        except ObjectDoesNotExistError:
```

composer/utils/checkpoint.py
mvpatel2000: lol I'm blind, good catch

Issue #817, opened 2022-03-24
SAM (sometimes) works on NLP too + SAM can be combined with SWA


JeanKaddour:
## Two minor suggestions for your model cards on SAM/SWA: 
* We found SAM to be effective on many NLP tasks too; sometimes also on GNN tasks, see Tables 2 and 3, respectively, in [this paper](https://arxiv.org/pdf/2202.00661.pdf) for more information.
* In the same paper, we also combined SAM and SWA (which seems straightforward to implement with your excellent Composer trainer) and empirically demonstrated that it improves over either approach in 39 out of 42 cases.

Thanks a lot for creating this exciting library!

mvpatel2000:
Closing. Tracking elsewhere as low pri. We're open to community suggestions!

PR #888, opened 2022-04-13
Support for SAM variants

Pikus16:
This branch is intended to provide support for several SAM variants, in particular [ESAM](https://arxiv.org/abs/2110.03141), [SS-SAM](https://arxiv.org/pdf/2110.03141.pdf), [LookSAM](https://arxiv.org/abs/2203.02714) and [ASAM](https://arxiv.org/abs/2102.11600). These variants can be run together or separately as part of a larger SAM algorithm (see 'SAM Algorithm' below).

#### ASAM (Adaptive Sharpness-aware Minimization)
Proposes a scale-invariant calculation of sharpness that is used when finding the local maximum.
#### LookSAM
Proposes to only periodically compute the SAM gradient, and for all other iterations use the project of the SAM gradient instead of recalculating it.
Note that the paper also includes a layer-wise LookSAM variant for larger batches that was not included here.
#### SS-SAM (Stochastic Scheduled Sharpness-aware Minimization)
Proposes different stochastic functions to randomly determine if a SAM update should be done at each step.
#### ESAM (Efficient Sharpness-aware Minimization)
Adds Stochastic Weight Perturbation (SWP), where "the sharpness measure is approximated by perturbing a stochastically chosen set of weights" and Sharpness-Sensitive Data Selection (SDS), where the "SAM loss is optimized using only a judiciously selected subset of data that is sensitive to the sharpness."
Note that this implementation only adds support for SWP. SDS requires that the loss not be reduced (or to recompute the loss again) which could be a larger change for integration.

### SAM Algorithm PseudoCode
```
// T = total number of iterations, theta = model parameters
for t in 1...T:
  Sample batch B
  // ESAM SWP
  for p in theta:
    if random() > ESAM_SWP_beta: 
      p.requires_grad = False  // don't use for SAM calculation
  // Check if should do SAM-update or not. This supports SS-SAM as well as the
  // current interval system in the codebase.
  if should_do_sam_this_iteration(t):
    g_orig = loss(theta, B)
    if ASAM:
      t_w = abs(theta)
      grad_norm = norm(g_orig * t_w)
    else:
      grad_norm = norm(g_orig)
    scale = rho / grad_norm
    if ESAM_SWP:
      scale /= (1 - ESAM_SWP_beta)
    eps_w = g_orig * scale
    if ASAM:
      eps_w *= t_w ** 2
    g_new = loss(theta + eps_w, B)
    if LookSAM:
      cos_proj = (g_orig * g_new) / (norm(g_orig) * norm(g_new))
      g_v = g_new - abs(g_new) * cos_proj * g_orig / norm(g_orig)
  else:
    g_new = loss(theta, B)
    if LookSAM:
      g_new += alpha * norm(g_new) / norm(g_v) * g_v
  theta -= lr * g_new // SGD Update
```

### Running
Specifying the SAM variants to choose is passed as arguments to the SAM class.
```
trainer = Trainer(
    ...
    algorithms=[SAM(
         interval=SAM_FixedInterval, // type of interval scheduler to use (for backwards compatibility, 
            // can also just accept an int which will then default to a SAM_FixedInterval which is the current behavior
         interval_num = 3, // if specifying a SAM_FixedInterval, what interval to use (in future may make sense to remove
           // and just specify interval as an int when wanting to use SAM_FixedInterval
         use_LookSAM=True, // whether to use LookSAM 
         use_ESAM_SWP=True, // whether to use ESAM SWP
         use_ASAM=False) // whether to use ASAM 
])
```
Different SAM variants can be combined as desired. 

As of now each SAM variant is specified with its own boolean variable if it should be run, and any arguments specific to that variant. This can become unwieldy, and in the future may require some redesign (ex: have subclasses of SAM for each variant, and a master parent class that can still run each at each part so the variants can be combined if desired).

The current SAM implementation in the codebase (and this modified one) can also support [WASAM](https://arxiv.org/pdf/2202.00661.pdf) (which combines SAM and SWA) as referenced in #817 by appending a SWA step after the SAM updates. (Usually SWA is done at the epoch level, so this would be after N SAM updates, where N is the number of steps per epoch). Something like this may work:
```
swa_algorithm = SWA(swa_start="6ep", swa_end="8ep")
sam_algorithm = SAM()
trainer = Trainer(
    ...
    algorithms=[sam_algorithm swa_algorithm[
)
```

Note: this requires more testing to replicate some of the experiments in the original papers. I'm currently running some of those experiments and will add results here when I have them.


hanlint:
Thanks @rahulvigneswaran and @Pikus16  for the contribution! We've created another repository where third-party contributions (https://github.com/mosaicml/experimental) can go first for validation, testing, and eventual upstreaming into Composer. Could you perhaps make a PR into that repository?

PR #889, opened 2022-04-13
Modify ResNet9 benchmark to enable channels_last and progressive_resizing

coryMosaicML:
This PR switches ResNet9 to use average pooling rather than max pooling, which has two effects:
1) The model can handle different input sizes, which enables `progressive_resizing`.
2) The reshape before the fully connected layer now plays nicely with `channels_last`, which didn't run before.

These changes also seem to slightly improve accuracy with current hparams.

Runs (5 seeds):
[Current dev: Accuracy 92.93% +/- 0.30 
After change: Accuracy 93.68% +/- 0.22](https://wandb.ai/mosaic-ml/cory-resnet9-speedtest?workspace=user-corymosaicml)


dblalock (approved):
LGTM

PR #891, opened 2022-04-13
Change NLP yamls to use evaluators

anisehsani:
Without using the evaluators, the Bert model uses all the metrics in `self.val_metrics` as defaults. This causes it to crash. This changes all the nlp tasks to use evaluators with the right metrics.


composer/yamls/models/bert-base.yaml
```diff
@@ -12,20 +12,27 @@ train_dataset:
...
+          shuffle: true
+          drop_last: false
+          subsample_ratio: 0.02
+          use_masked_lm: true
+          mlm_probability: 0.15
+          max_seq_length: 128
+      metric_names:
+        - Accuracy
+        - BinaryF1Score
+        - MatthewsCorrCoef
```
moinnadeem, line 35: BERT-base requires `LanguageCrossEntropy` and `MaskedAccuracy`, not any of these metrics

moinnadeem (approved):
Just one change in the BERT-base YAML, otherwise LGTM! Good work!

composer/yamls/models/bert-base.yaml
hanlint, line 35:
```suggestion
        - LanguageCrossEntropy
        - MaskedAccuracy
```

PR #893, opened 2022-04-14
Print the stdout/stderr of the crashing process

ravi-mosaicml:
Fix the launcher script to always print the stdout/stderr of the crashing process. To do so, the stdout and stderr must always be saved to a file. Then, the launcher script can read from the stdout and stderr files to print these to the terminal.

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -327,11 +327,9 @@ def _print_process_exit_status(global_rank: int, process: subprocess.Popen):
 
 
 def _cleanup_processes(processes: Dict[int, subprocess.Popen]):
-    living_processes_at_end = set()
     for global_rank, process in processes.items():
         process.poll()
         if process.returncode is None:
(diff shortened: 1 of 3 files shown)


hanlint:
@Landanjs can you verify that this fixes the issue?

Landanjs:
In my example, the stdout from each process is now printed to terminal which is helpful! But it is difficult to parse the stdout though since each process prints the training configuration. On VScode, the full logs does not fit on my terminal, so I can only see processes 4-6

Is there anyway to print the stdout from the crashing process or to only print the traceback?

Landanjs:
We decided to only print the configuration on rank 0 which should reduce the length of the stdout

Landanjs (approved):
Can't comment on the code, but I think the functionality is a good first step! It can still be difficult to parse the logs, but at least the necessary info is provided.

 In the future, it would be good to explore better ways of conveying traceback info.

PR #896, opened 2022-04-14
Changing defaults in selective backprop to not downsample to allow for non-visual input

jzf2101:
Per discussion with @growlix, downsampling by default in `selective_backprop` with `scale_factor=0.5` means we are assuming the data is image data.  This PR turns off downsampling by default.


hanlint (approved):

hanlint:
Should the corresponding `yaml` (https://github.com/mosaicml/composer/blob/dev/composer/yamls/algorithms/selective_backprop.yaml) also be changed?

growlix (approved):
echo what Hanlin said. Otherwise lgtm

jzf2101:
@growlix updated per @hanlint

jzf2101:
Followup question: Do we want to update the benchmarks under the assumption that the images are *not* downsampled, or put a caveat that downsampling is used on the benchmarks for image data?

abhi-mosaic:
Just took a quick look , I think this is fine but the default throughput increase for vision tasks will be smaller now that `scale_factor=1`. I wonder if there is some way to detect if the inputs are images (like checking for 4-d inputs) and only apply the scaling in that situation? I don't have a strong opinion here.

> Followup question: Do we want to update the benchmarks under the assumption that the images are not downsampled, or put a caveat that downsampling is used on the benchmarks for image data?

I think the old vision benchmarks should still use downsampling as that gave us a substantial improvement in our tradeoff curves, in old experiments.

jzf2101:
I would be fine with making a change so that images are downsampled by default.  I think we could make assumptions on datatype using the model perhaps?

hanlint:
I'm a bit wary of magical defaults. I would suggest we pick one, and seems like the default that doesn't crash for different data types would be the most appropriate?

coryMosaicML:
We should be aware that downsampling the input is more restrictive than restricting to image data. Downsampling the input can fail for image models in a few notable cases:
- The network can't accept inputs with a reduced size. For example, if one used MaxPool2D at the end rather than AvgPool2D this would break.
- The task isn't classification. For example, semantic segmentation tasks will require the targets to match the shape of the output, which won't be true when doing downsampling, for object detection bounding boxes will need to be resized as well, etc.

I don't think it's quite right to think of downsampling as an image specific default, as it's sort of specific to the case of classification with nicely written models. It probably won't work on the majority vision tasks. I'm in favor of the default behavior not doing downsampling even for images such that things work out of the box in more cases.

jfrankle:
This sounds good to me so long as we have recipes that demonstrate the right default hparams (e.g., the final FKS for ImageNet). This is also something we can document in an updated method card.

jfrankle:
Decision:

- We are going to follow @coryMosaicML's proposal and turn off the downsampling by default.
- We are going to update the method card to explain the inefficiencies of the extra forward pass and explain the workaround of using downsampling with warnings on the limited scenarios in which this is possible.
- We will eventually (end of this sprint) provide recipes that invoke the method with downsampling both as an example and demonstrating the value it provides. When we do that, we will link to the recipes from the method card.

jzf2101:
As a follow-up to this discussion, additional work will need to be done to handle the datatypes that are used by language models.  @hanlint @moinnadeem @growlix have provided some additional context as to how this may be handled:

For `selective_backprop` we assume the data is currently in a [`(input, target)` tuple structure.](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L224)

However, the text datasets we work with are typically in a dictionary structure similar to [`{"input_ids", "attention_mask", "token_type_ids"}` as seen in this BERT example](https://github.com/mosaicml/composer/blob/204129b810ce71d4089781b6ee2bef9f0b2703f1/composer/models/bert/model.py#L57).

This makes applying `selective_backprop` using `select_using_loss` a bit difficult to generalize, since we currently assume the [`input` and `target` are tensors](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L55).  

My current proposal is to keep [`select_using_loss`]((https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L55)) more or less as is, but [extract out the component that performs the batch index selection](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L130-L142) so that we can use it both in `select_using_loss` and an alternate `select_using_loss_text`-like function.  I'd like to open this for discussion before I make this change.

growlix:
> As a follow-up to this discussion, additional work will need to be done to handle the datatypes that are used by language models. @hanlint @moinnadeem @growlix have provided some additional context as to how this may be handled:
> 
> For `selective_backprop` we assume the data is currently in a [`(input, target)` tuple structure.](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L224)
> 
> However, the text datasets we work with are typically in a dictionary structure similar to [`{"input_ids", "attention_mask", "token_type_ids"}` as seen in this BERT example](https://github.com/mosaicml/composer/blob/204129b810ce71d4089781b6ee2bef9f0b2703f1/composer/models/bert/model.py#L57).
> 
> This makes applying `selective_backprop` using `select_using_loss` a bit difficult to generalize, since we currently assume the [`input` and `target` are tensors](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L55).
> 
> My current proposal is to keep `select_using_loss` more or less as is, but [extract out the component that performs the batch index selection](https://github.com/mosaicml/composer/blob/009822e57573113db4342f04e88edd3284a3b3f1/composer/algorithms/selective_backprop/selective_backprop.py#L130-L142) so that we can use it both in `select_using_loss` and an alternate `select_using_loss_text`-like function. I'd like to open this for discussion before I make this change.

Tagged @ravi-mosaicml for his thoughts.

jzf2101:
@jfrankle there's an update to the method card that has been reviewed by @growlix, if you want to do an additional look over as well.

ravi-mosaicml (commented):
Changes look fine, but I do not have the research background to make an informed decision. Removing myself from review...

composer/algorithms/selective_backprop/selective_backprop.py
```diff
@@ -213,15 +213,23 @@ def match(self, event: Event, state: State) -> bool:
         )
         return is_chosen
 
-    def apply(self, event: Event, state: State, logger: Optional[Logger] = None) -> None:
+     def apply(self, event: Event, state: State, logger: Optional[Logger] = None, input_pos: Union[int, str, list, None] = None, target_pos: Union[int, str, list, None] = None) -> None:
```
hanlint, line 216: Suggest moving these to the algorithm's `__init__` signature, and then accessing here via attributes. Can also reduce the verbosity by defining a type alias.

```
def __init__(..., input_pos: Position = None, target_pos: Position = None):
    self.input_pos = input_pos
    self.target_pos = target_pos
```
cc: @eracah as FYI as well.

hanlint (commented):
Addressing the merge conflict and updating the branch should resolve your jenkins test issues.

composer/algorithms/selective_backprop/selective_backprop.py
hanlint: Actually, this change seems out of scope with the PR's intent to simply change the defaults? Consider staging this into two PRs, one to simply change the defaults, and another to generalize to multiple batch types.

eracah: yeah I think we should avoid adding code that tries to handle different types of batches on a per-algorithm basis until we resolve [this issue](https://mosaicml.atlassian.net/browse/CO-2).

eracah:
@jfrankle, @growlix, @jzf2101, why do we do two forward passes? It seems like the original paper doesn't use two forward passes right? What's the reason that we don't just do two loss computations instead of two full forward passes like so:
```python
outputs = model.forward(batch)
with torch.no_grad():
   loss = model.loss(outputs, labels)
selected_outputs, selected_labels = select_examples_based_on_loss(outputs, labels, loss)
loss = model.loss(selected_outputs, selected_labels)
loss.backward()
```

eracah:
oh ok @abhi-mosaic mentioned to me that the matrix multiplications are the same either way except for the examples not selected are masked out by multiplying by zero, but the flops are the same.

jzf2101:
@growlix and i discussed and it seems as if taking out the additional forward can't be done?

eracah:
@jzf2101, yeah I talked with @abhi-mosaic and @growlix and they said it CAN be done, but it's just as slow as normal backprop, so no point in doing it.

composer/algorithms/selective_backprop/selective_backprop.py
eracah: we will add some helper getters and setters and custom types that will make the specifying of which portion of the batch to grab easier for algorithm-writers.

hanlint (approved):

PR #897, opened 2022-04-14
Fix loss logging with DeepSpeed

abhi-mosaic:
Discovered that the loss being logged was `grad_accum` times larger than it should be when using DeepSpeed (any ZERO stage). I think this was due to a typo. Future `trainer.py` refactor should probably have some more explicit names like `state.microbatch`, rather than just `state.batch` which is getting set to minibatches/microbatches in different places.


jbloxham (approved):

abhi-mosaic:
Screenshot in WandB before/after this fix:

Screen Shot 2022-04-14 at 4 44 07 PM

PR #898, opened 2022-04-15
Update GPT2 YAMLs

abhi-mosaic:
This PR is for replacing our GPT2 yamls to better match the 4 model sizes from the original paper.

I am saving this as a draft because removing the old GPT2 yamls will break Composer regression testing on the GPT2 benchmark (which needs to be replaced, but not in scope for this release cycle).

In the future, I think I would ideally like to provide GPT2 and GPT3 yamls that are good starting points for users who are interested in replicating the original papers, but our NLP generation benchmark (what we develop algorithms for and report in Explorer) should go through its own design process.


hanlint:
Closing this as stale -- please reopen when ready.

PR #900, opened 2022-04-15
Console log level

ravi-mosaicml:
Added back console logging. Before, this functionality lived in the FileLogger, which could previously write to `stdout`. That functionality was removed in #710, which introduced stdout and stderr capture to the file logger. This PR brings it back and merges it with the progress bar logger. Doing so in the same class allows prints to go through `tqdm.write`, so log lines do not interfere with the progress bar.

In addition, the progress bar logger parameters are flattened into `Trainer.__init__`. This change makes it is possible to always append the progress bar logger (unless if it is explicitly set to `False` via the `progress_bar: bool` argument).


composer/trainer/trainer_hparams.py
```diff
@@ -469,9 +475,6 @@ def validate(self):
 
     def initialize_object(self) -> Trainer:
         self.validate()
-        import composer
-        logging.getLogger(composer.__name__).setLevel(self.log_level)
```
ravi-mosaicml, line 473: Removed this parameter -- was thinking it would be too confusing with the console log level parameter below

composer/trainer/trainer_hparams.py
```diff
@@ -568,10 +571,16 @@ def initialize_object(self) -> Trainer:
...
+            # Callbacks
+            callbacks=callbacks,
+
+            # Logging
             run_name=self.run_name,
             loggers=loggers,
-            callbacks=callbacks,
+            progress_bar=self.progress_bar,
+            log_to_console=self.log_to_console,
+            log_level=self.log_level,
```
ravi-mosaicml, line 582: For brevity, I called this `log_level`, though I realize that may lead to confusion. Should I call it `console_log_level`?

ravi-mosaicml: (Going to call this console_log_level; I don't like log_level after looking at it haha)

composer/trainer/trainer_hparams.py
ravi-mosaicml: (Adding this parameter back in, but going to call it python_log_level)

moinnadeem (approved):
There are two things I would like fixed, but we can do it in a future PR if it'll block a release:

1) Since `print(...)`, `log.warning(...)` or `log.info(...)` statements aren't captured in this flow, they actually get broken up in a really annoying way, see this: 
image. I would prefer to also pass these through `tqdm.write(...)` so this behavior doesn't occur.

2) Can we enable `dynamic_ncols = True` in TQDM so we are able to resize the window without the progress bar breaking?

In either case, great work Ravi!

composer/loggers/progress_bar_logger.py
```diff
@@ -1,14 +1,15 @@
...
 from __future__ import annotations
 
 import collections.abc
+import sys
 from dataclasses import asdict, dataclass
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, TextIO, Union
 
-from tqdm import auto
+import tqdm.auto
```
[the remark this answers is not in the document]
ravi-mosaicml: Calling a function `auto` later in the code seemed a bit too undescriptive

ravi-mosaicml:
Added in dynamic ncols. Attempting to redirect all calls to `sys.stderr.write` to print via `tqdm.write` is going to be non-trivial, since tqdm likely also calls `sys.stderr.write` internally. That said, it should be possible to redirect all logging statements without too much work. Print statements go to sys.stdout by default, so that won't be possible.

Landanjs:
Welp, this was merged... I was no longer getting the traceback after the error when I logged to console:

```
[FIT][batch=0]: { "trainer/algorithms": [], }
[EPOCH][batch=0]: { "epoch": 0, }                                                                                                                                                                
[BATCH][batch=0]: { "trainer/global_step": 0, "trainer/batch_idx": 0, }                                                                                                                          
Epoch 0:   0%|          | 0/630 [00:02<?, ?it/s]                                                                                                                                                 ERROR:composer.cli.launcher:Rank 1 crashed.
Terminated
```

Is this expected behavior?

composer/loggers/progress_bar_logger.py
```diff
@@ -57,50 +62,97 @@ def state_dict(self) -> Dict[str, Any]:
...
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        progress_bar: bool = True,
+        log_to_console: Optional[bool] = None,
+        console_log_level: Union[LogLevel, str, Callable[[State, LogLevel], bool]] = LogLevel.EPOCH,
+        stream: Union[str, TextIO] = sys.stderr,
+    ) -> None:
+        self.show_pbar = progress_bar
+        log_to_console = not progress_bar if log_to_console is None else log_to_console
```
abhi-mosaic, line 109: Bit confused about this line, shouldn't the default be to show both the progress bar AND console logging? Wonder if this is related to Landan's comment above.

moinnadeem:
Update: Agreed with @Landanjs, I was able to see `log.warnings(...)` that I was previously seeing, but it seems as if `stderr` isn't getting logged, or something similar?

Without any modifications, and with the latest dev merged into my fork, I see this:
```
/mnt/cota/moin/env/lib/python3.8/site-packages/apex/pyprof/__init__.py:5: FutureWarning: pyprof will be removed by the end of June, 2022
  warnings.warn("pyprof will be removed by the end of June, 2022", FutureWarning)
/mnt/cota/moin/moin_composer/composer/cli/launcher.py:193: UserWarning: AutoSelectPortWarning: The distributed key-value port was auto-selected. This may lead to race conditions when launching multiple training processes simultaneously. To eliminate this
 race condition, explicitly specify a port with --master_port PORT_NUMBER
  warnings.warn("AutoSelectPortWarning: The distributed key-value port was auto-selected. "
/mnt/cota/moin/env/lib/python3.8/site-packages/apex/pyprof/__init__.py:5: FutureWarning: pyprof will be removed by the end of June, 2022
  warnings.warn("pyprof will be removed by the end of June, 2022", FutureWarning)
UserWarning: MalformedYAMLWarning: schedulers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
UserWarning: MalformedYAMLWarning: loggers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
ERROR:composer.cli.launcher:Rank 1 crashed.
[1]    5879 terminated  composer -n 8 examples/run_composer_trainer.py -f
```

If I comment out the `stdout` and `stderr` redirection in `composer/cli/launcher.py`, then I see this traceback, which I should be seeing:
```
UserWarning: MalformedYAMLWarning: schedulers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
UserWarning: MalformedYAMLWarning: schedulers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
UserWarning: MalformedYAMLWarning: loggers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
UserWarning: MalformedYAMLWarning: loggers should be a dict. (source: /mnt/cota/moin/env/lib/python3.8/site-packages/yahp/create/create.py:44)
  File "examples/run_composer_trainer.py", line 65, in <module>
  File "examples/run_composer_trainer.py", line 65, in <module>
    main()
  File "examples/run_composer_trainer.py", line 44, in main
    main()
  File "examples/run_composer_trainer.py", line 44, in main
    trainer = hparams.initialize_object()
  File "/mnt/cota/moin/moin_composer/composer/trainer/trainer_hparams.py", line 476, in initialize_object
    trainer = hparams.initialize_object()
  File "/mnt/cota/moin/moin_composer/composer/trainer/trainer_hparams.py", line 476, in initialize_object
    device = self.device.initialize_object()
  File "/mnt/cota/moin/moin_composer/composer/trainer/devices/device_hparams.py", line 33, in initialize_object
    device = self.device.initialize_object()
  File "/mnt/cota/moin/moin_composer/composer/trainer/devices/device_hparams.py", line 33, in initialize_object
    return DeviceGPU()
  File "/mnt/cota/moin/moin_composer/composer/trainer/devices/device_gpu.py", line 34, in __init__
    return DeviceGPU()
  File "/mnt/cota/moin/moin_composer/composer/trainer/devices/device_gpu.py", line 34, in __init__
        torch.cuda.set_device(self._device)torch.cuda.set_device(self._device)

  File "/mnt/cota/moin/env/lib/python3.8/site-packages/torch/cuda/__init__.py", line 311, in set_device
  File "/mnt/cota/moin/env/lib/python3.8/site-packages/torch/cuda/__init__.py", line 311, in set_device
        torch._C._cuda_setDevice(device)torch._C._cuda_setDevice(device)

RuntimeErrorRuntimeError: : CUDA error: invalid device ordinal
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.CUDA error: invalid device ordinal
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

ERROR:composer.cli.launcher:Rank 1 crashed.
[1]    6574 terminated  composer -n 8 examples/run_composer_trainer.py -f
```

My logger looks like this:
```
loggers:
    - progress_bar: {}
```

Something is up -- why are we no longer catching stderr?

PR #906, opened 2022-04-18
Fix unformatted artifact names

moinnadeem:


moinnadeem:
Yup can do!

examples/run_composer_trainer.py
```diff
@@ -47,7 +47,7 @@ def main() -> None:
     with tempfile.NamedTemporaryFile(mode="x+") as f:
         f.write(hparams.to_yaml())
         trainer.logger.file_artifact(LogLevel.FIT,
-                                     artifact_name="{run_name}/hparams.yaml",
+                                     artifact_name=f"{run_name}/hparams.yaml",
```
hanlint, line 50: This looks to be an incorrect fix -- the `artifact_name` is a format string that gets formatted internally. 

For clarity, perhaps the argument should be named `name_format` or something else instead?

hanlint (changes requested):

examples/run_composer_trainer.py
moinnadeem: Hm, so for reference, I see this on my cloud bucket
image
image

so it seems to me that we're missing a format function on the string somewhere -- let me start running jobs and see if I can track down where this is happening

hanlint: Ah the confusion arises because `artifact_name` in `FileLogger` is formattable (https://github.com/mosaicml/composer/blob/d025644e5786a84ac7edebbc6284813df1503c98/composer/loggers/file_logger.py#L82), but in the `logger.file_artifact` method, `artifact_name` is a string.

moinnadeem:
Ah, @ravi-mosaicml, that isn't the correct solution either. See the attached screenshot.

image

As far as I can tell, I think the crux of the issue comes from `log_file_artifact` failing to format the artifact names at all. Double check me on this though!

See:
1. The `file_artifact` [function implementation](https://github.com/mosaicml/composer/blob/dev/composer/loggers/logger.py#L132)
2. The `log_file_artifact` function for [ObjectStoreLogger](https://github.com/mosaicml/composer/blob/dev/composer/loggers/object_store_logger.py#L287) and [WandB logger](https://github.com/mosaicml/composer/blob/dev/composer/loggers/wandb_logger.py#L99)

If I am seeing this correctly, it looks like neither of these actually format the filename? If you guys agree, I see two options:
1. In `composer/loggers/logger.py`, add a `format_name_with_dist` call before passing the `artifact_name` to destination loggers.
2. In each logger (aka `composer/loggers/object_store_logger.py` and `composer/loggers/wandb_logger.py`), add a call to `format_dist_with_name`. 

I am not opinionated as to which one is correct, any thoughts?

moinnadeem:
@hanlint How sure are you that the `artifact_name` is the format string? I just printed a few values of the artifact name from this call:
https://github.com/mosaicml/composer/blob/dev/composer/loggers/logger.py#L150-159

and they're all formatted. For example, the `artifact_name` parameter will look something like: 
`Artifact name in the file_artifact call: 1650318262-radiant-anaconda/checkpoints/ep0-ba100-rank0`

PR #908, opened 2022-04-18
WandB multi-rank logging

abhi-mosaic:
This PR changes a few things to improve WandB arifact logging and make multinode logging more user-friendly. Note that the multi-rank features are only really necesary when using DeepSpeed ZERO.

Changes:
* Move the `rank_zero_only` naming behavior to `WandBLogger`, rather than `WandBLoggerHparams`
  * previously, `traner.run_name` was not making its way into the logger
* Stick with WandB Artifacts rather than WandB `.save()`, because the latter doesn't have safe tmpdir+upload+delete_on_finish, which is required for our `LoggerDestination`
* WandB artifact extension is detected based on destination artifact name, not source filename (caused an issue with our temporary YAML file)

Right now, the way I would recommend a user to use the WandB logger with multi-rank is something like this:
```bash
composer -n 8 examples/run_composer_trainer.py -f composer/yamls/models/gpt3_125m.yaml \
  --run_name myrun \
  --save_folder "/tmp/{run_name}" \
  --save_artifact_name "{run_name}/checkpoints/rank{rank}" \
  --save_interval 5ba \
  --save_num_checkpoints_to_keep 1 \
  ...
  --loggers wandb \
  --loggers.wandb.project abhi-multinode-checkpointing \
  --loggers.wandb.rank_zero_only false \
  --loggers.wandb.log_artifacts true
```

This creates `WORLD_SIZE` artifacts that are each linked to one of the `WORLD_SIZE` runs, and within each artifact, there are user-readable tagged versions of the checkpoint, like `entity/runname/runname.checkpoints.rank0.tar:ep0-ba10`. When you download this artifact, you get a `.tar` file with the original `save_filename`.

I tried consolidating the multiple ranks into a single artifact, but had a lot of difficulty doing so. I found a [potentially useful api call here](https://docs.wandb.ai/guides/artifacts/artifact-creation-modes#collaborative-mode), but would like to get this PR merged first as it is useable enough.

TODO @abhi-mosaic : launch an E2E multi-node test and sanity check the WandB UI. Also run with frequent checkpointing to make sure we are never running out of local disk space.

### DIFF
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -131,7 +131,7 @@ class WandBLogger(LoggerDestination):
 
             if self._enabled and self._log_artifacts:
                 import wandb
-                extension = file_path.name.split(".")[-1]
+                extension = new_artifact_name.split(".")[-1]
                 artifact = wandb.Artifact(name=new_artifact_name, type=extension)
                 artifact.add_file(os.path.abspath(file_path))
                 wandb.log_artifact(artifact, aliases=aliases)
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -7,7 +7,9 @@ from __future__ import annotations
 import os
 import pathlib
 import re
+import shutil
 import sys
+import tempfile
 import textwrap
 import warnings
 from typing import Any, Dict, Optional
@@ -64,6 +66,10 @@ class WandBLogger(LoggerDestination):
             init_params = {}
         self._init_params = init_params
 
+        # Temporary directory for staging artifact upload
+        tempdir = tempfile.TemporaryDirectory()
+        self._upload_staging_folder = tempdir.name
+
     def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
         import wandb
         del log_level  # unused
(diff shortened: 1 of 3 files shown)


composer/loggers/wandb_logger.py
```diff
@@ -89,30 +90,45 @@ def state_dict(self) -> Dict[str, Any]:
...
+            if new_artifact_name != artifact_name:
+                warnings.warn(("WandB permits only alpha-numeric, periods, hyphens, and underscores in artifact names. "
+                               f"The artifact with name '{artifact_name}' will be stored as '{new_artifact_name}'."))
+
+            if self._enabled and self._log_artifacts:
+                import wandb
+                extension = new_artifact_name.split(".")[-1]
+                artifact = wandb.Artifact(name=new_artifact_name, type=extension)
+                artifact.add_file(os.path.abspath(file_path))
+                wandb.log_artifact(artifact, aliases=aliases)
```
ravi-mosaicml, line 131: ooooh, I didn't realize that wandb supports `alias`'s. That's awesome, so we can implement the `symlink_artifact` API for wandb. (Not in this PR; that's for another one).

examples/run_composer_trainer.py
```diff
@@ -43,15 +43,16 @@ def main() -> None:
...
-    # Log the config to an artifact store
-    with tempfile.NamedTemporaryFile(mode="x+") as f:
-        f.write(hparams.to_yaml())
-        trainer.logger.file_artifact(LogLevel.FIT,
-                                     artifact_name=f"{trainer.logger.run_name}/hparams.yaml",
-                                     file_path=f.name,
-                                     overwrite=True)
-
-    # Print the config to the terminal on each local rank 0
+    # Log the config to an artifact store if global rank 0
```
ravi-mosaicml, line 46: I'm being a real nit lol

```suggestion
    # Only log the config once, since it should be the same on all ranks.
```

ravi-mosaicml (approved):
LGTM; thanks!

composer/loggers/wandb_logger.py
abhi-mosaic: WandB also supports references to external object stores like S3, GCS, which could be a more "enterprise" solution than storing everything on WandB: https://docs.wandb.ai/guides/artifacts/references#s3-gcs-references

This would also simplify multi-gpu and multi-node WandB logging in 99% of cases, because rank0 could build a single artifact that just stores pointers to S3 URIs of each ranks' checkpoints.

Fair warning, this might make your head spin... do `LoggerDestinations` have to know about eachother? Does `ObjectStoreLogger` call `wandb_logger.log_file_artifact(s3_uri)`? Do we implement a 2-stage system like:

```python
# self.storage_logger is a single choice, like ObjectStoreLogger
# self.non_storage_loggers is a list of loggers like WandB, MLFLow, TensorBoard, etc.

device_reference_url = self.storage_logger.log_file_artifact(..., ...)
all_reference_urls = dist.all_gather(device_reference_url)
for logger in self.non_storage_loggers:
  for url in all_reference_urls:
    logger.log_reference(url)
```

So much fun to be had! 🙃

Issue #907, opened 2022-04-18
Time initialization behaves inconsistently


coryMosaicML:
## To reproduce

Creating an instance of time this way fails correctly because noninteger values are not allowed when the unit is epochs:
```
from composer.core import Time, TimeUnit

time = Time(0.5, TimeUnit.EPOCH)
```

However, creating it this way succeeds, but not with the value the user intended:
```
from composer.core import Time, TimeUnit

time = Time.from_timestring("0.5ep")
print(time)
```
```Time(0, TimeUnit.EPOCH)```

## Expected behavior

Both of these should fail because integer values are required when using epoch units. The same behavior also exists for units of batches, samples, and tokens. Right now, converting from float to integer using `from_timestring` can potentially introduce unexpected and difficult to find bugs.

## Additional information
A better solution could be to allow non-integer values for these units. A time of `0.5 epochs` is a valid time, and would be useful in cases where a quantity has units of time but should not be restricted to integer values.

ravi-mosaicml:
The time tracking design does not incorporate any time conversion between units, so supporting fractional epochs will not be possible. But yes, we should error if `int(time) != float(time)` when the unit is epochs / batches / tokens / samples.

PR #909, opened 2022-04-18
Add Exponential Moving Average (EMA) algorithm

coryMosaicML:
This PR adds exponential moving average (ema) as an algorithm. 

The functional implementation should be similar to other EMA implementations as far as hparams go.

The composer trainer implementation's hparams are a little different, since the time abstraction makes it possible to convert the hyperparams into a more interpretable form. There is a little ugliness here though because
1. It is not possible to do unit conversion with the time abstraction currently, so `half_life` and `update_interval` must have the same units. It would be really nice if one could write `half_life="2ep"` and `update_interval="10ba"` but unit conversion is needed for this. I don't believe we intend to change this.
2. The time abstraction doesn't allow non-integer values in units of epochs or batches, so one can't set `half_life="1.5ep"`. Because of [#907](https://github.com/mosaicml/composer/issues/907#issue-1207423730) this will still work but silently give behavior equivalent to `half_life="1ep"`. I plan on doing a PR to fix this later.

Another bit of ugliness is that currently composer does not support before/after checkpointing events, or a separate evaluation model, so `state.model` in the checkpoint will be the training model rather than the ema model, while the ema model should be used for inference. To remedy this, I added a `ema_algorithm.get_ema_model(model)` method to the algorithm to extract the ema model, but this places the burden on the user of extracting the model themselves after training.  Otherwise, graceful resumption becomes difficult. 

Default hyperparameters for the composer algorithm corresponds to a smoothing value of `smoothing=0.993` (`half_life="100ba"`) which is a little lower (less smoothing) than TensorFlow's recommended values of `decay` between `0.999` and `0.9999`. This default is a little more conservative, so open to suggestions on if it should be changed to match the TF recommendations.  

Runs:
[ResNet50-i1k-wds default hparams
](https://wandb.ai/mosaic-ml/cory-ema/runs/26nggw97?workspace=user-corymosaicml)[ResNet50-i1k-wds train with ema weights](https://wandb.ai/mosaic-ml/cory-ema/runs/3ks1l8f3?workspace=user-corymosaicml)
[DeepLabV3+-wds default hparams
](https://wandb.ai/mosaic-ml/cory-ema/runs/1eo29qlb?workspace=user-corymosaicml)[GPT3-125M default hparams](https://wandb.ai/mosaic-ml/cory-ema/runs/etnwtgiu?workspace=user-corymosaicml)


composer/algorithms/ema/ema.py
```diff
@@ -0,0 +1,260 @@
...
+        shadow_model.param_list = parameters
+        shadow_model.buffer_list = buffers
+        setattr(self, name, shadow_model)
+
+    def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
+        for attribute_name, serialized_value in state.items():
+            self.load_shadow_model(attribute_name, serialized_value["parameters"], serialized_value["buffers"])
+
+
+class ShadowModel:
```
growlix, line 228: Spooky 👻

composer/functional/__init__.py
```diff
@@ -22,6 +22,7 @@
 from composer.algorithms.colout import colout_batch
 from composer.algorithms.cutmix import cutmix_batch
 from composer.algorithms.cutout import cutout_batch
+from composer.algorithms.ema import ema
```
growlix, line 25: might have to rename cuz of docs issue

composer/functional/__init__.py
```diff
@@ -44,6 +45,7 @@
     "colout_batch",
     "cutmix_batch",
     "cutout_batch",
+    "ema",
```
growlix, line 48: see above

growlix (approved):
Minor docs rendering issue, but otherwise implementation looks great!

composer/algorithms/ema/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Suggested Hyperparameters
+
+The Composer Trainer implementation of EMA has two hyperparameters:
+
+- `half_life` - The half life for terms in the average. A longer half life means old information is remembered longer, a shorter half life means old information is discared sooner.
+- `update_interval` - The period at which updates to the moving average are done. A longer update interval means that updates are computed less frequently.
+
+A good typical starting value for `half_life` is `half_life="100ba"`, for a half life of 100 batches. At the same time, `update_interval` can be left unspecified which will default to `update_interval="1ba"`, or set to a larger value such as `update_interval="10ba"` to improve runtime. Shorter update intervals typically result in better generalization performance at the cost of somewhat reduced runtime.
```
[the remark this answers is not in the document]
coryMosaicML: Oops, should be longer runtime.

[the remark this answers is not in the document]
coryMosaicML: I renamed the `ema` function to `compute_ema` and this seems to have fixed the docs issue. It's a bit of a shame that this was necessary, but it should be fine.

PR #913, opened 2022-04-19
Fix Docker condition for MOFED drivers

jbloxham:


.ci/Jenkinsfile
```diff
@@ -2,7 +2,7 @@ pCloud = "colo-research-01"
 gitUrl = null
 gitBranch = null
 gitCommit = null
-pTimeout = '1800' // in seconds
+pTimeout = '3600' // in seconds
```
ravi-mosaicml, line 5: Can you make this timeout just for the docker build, rather than globally? Will need a new variable.

PR #914, opened 2022-04-19
Log the rank zero seed via the logger

ravi-mosaicml:
You know something's wrong if it's not 42.


A-Jacobson (approved):
Lgtm: wouldn't want to forget the most important hyperparameter.

hanlint (changes requested):

PR #916, opened 2022-04-19
Fix failing CPU memory monitor test

hanlint:
`make test` fails when run on a GPU machine because `test_memory_monitor_cpu` still picks up on active GPU memory from other tests / concurrent jobs and debugging sessions.

This PR skips `test_memory_monitor_cpu` if GPU devices are available. This should not affect our CI/CD since we run CPU tests in a CPU machine.


ravi-mosaicml (approved):
LGTM

PR #917, opened 2022-04-19
Fix the launcher script

ravi-mosaicml:
Killing by process group ID was accidentally killing the launcher script before it could terminate child processes. Instead, the launcher script will attempt the SIGTERM kill with just the PID. If that fails, then it will attempt a SIGKILL kill with the PID and all child process PIDs. This is necessary since, during a SIGKILL, the killed process cannot intercept the signal to manually terminate its children.

This approach also ensures that when the launcher script is killed via SIGKILL by the process group ID, or via SIGTERM by the process ID, all training processes also are killed. If we were to instead give each training process a new process group ID, then killing the launcher script via process group ID would result in zombie training processes, as the SIGKILL would not propagate to different process groups.

Manually tested this approach in the terminal.

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -1,5 +1,4 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
-
 import datetime
 import logging
 import os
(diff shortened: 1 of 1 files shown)


Landanjs (approved):
The functionality LGTM! Resolves the issue for my specific example.

There is some extra printouts from the dataloader workers after rank 0 process exits, but @ravi-mosaicml thinks this is an issue on PyTorch's end.

Again, can't say much about the code 😅

jbloxham (approved):
LGTM - looks like these changes will improve UX a lot

PR #918, opened 2022-04-19
Attempt to destroy process groups after each test runs

hanlint:
This could potentially fix the issues seen in #831 , but local testing reveals significant errors. Opening this for feedback.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ from typing import List, Optional
 
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


ravi-mosaicml:
Until pytorch fixes the issues with properly destroying process groups, so a new one can be created at the same port, this won't work.

ravi-mosaicml:
Closing for https://github.com/mosaicml/composer/pull/919

PR #919, opened 2022-04-19
Fix Distributed Timeouts

ravi-mosaicml:
A logger test that was manually initializing dist configured the timeout to be too low, resulting in timeouts in other tests.


jbloxham (approved):
LGTM, as long as it passes.

PR #923, opened 2022-04-19
Publish the junit results via the GH checks api

ravi-mosaicml:
1. Publish junit results under the `Tests` status check. This will allow external contributors to see where tests are failing, since they do not have Jenkins access. Note that only failing tests, not any passing tests, will show up.
2. Run lint checks and doctests via python and junit so any errors will be captured and reported via JUnit, which will make any errors visible through the GitHub UI.

Example: See https://github.com/mosaicml/composer/pull/923/checks?check_run_id=6099500888

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -2,6 +2,7 @@ pCloud = "colo-research-01"
 gitUrl = null
 gitBranch = null
 gitCommit = null
+pTimeoutDocker = '3600'
 pTimeout = '1800' // in seconds
 dependenciesChanged = null
 pytorchDockerBuildMatrix = null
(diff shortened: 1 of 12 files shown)


.ci/Jenkinsfile
```diff
@@ -350,7 +353,7 @@ stage('Build') {
                 sh "mkdir -p $buildOutputFolder"
 
                 archiveArtifacts(artifacts: artifactsGlob, fingerprint: true, allowEmptyArchive: true)
-                junit(allowEmptyResults: true, testResults: junitGlob)
+                junit(allowEmptyResults: true, testResults: junitGlob, checksName: 'Tests')
```
bandish-shah, line 356: How will this work if there are multiple failures across different containers?  Not sure if this is really an issue given we do early termination?

bandish-shah (approved):
LGTM, this is great.

.ci/Jenkinsfile
ravi-mosaicml: The Junit results are merged and collected from all spawned pods. So, multiple failures will show up here.

PR #924, opened 2022-04-20
Add Adaptive Gradient Clipping to Algorithms

eracah:
This PR adds [Adaptive Gradient Clipping](https://arxiv.org/abs/2102.06171) to `composer/algorithms`.

This includes:

* a functional interface: `apply_agc`
* a trainer interface: `AGC`

Two hyperparameters:
* clipping_threshold: float
* eps: float

Mainly uses `state.model` as an entry point to modify the gradients.

Potential issues could be:
* if grad_clipping is enabled at the same time.


composer/algorithms/agc/agc.py
```diff
@@ -0,0 +1,161 @@
...
+        eps (float, optional): Minimum value that weight norms are clamped to.
+    """
+
+    def __init__(self, clipping_threshold: float = 0.01, eps: float = 1e-3):
+        self.clipping_threshold = clipping_threshold
+        self.eps = eps
+
+    def match(self, event: Event, state: State) -> bool:
+        """Run on ``Event.AFTER_BACKWARD``."""
+        return event == Event.AFTER_BACKWARD
```
ravi-mosaicml, line 96: In the trainer, we apply `torch.nn.utils.clip_grad_norm_` immediately before the AFTER_TRAIN_BATCH event. Is AFTER_BACKWARD the correct event for adaptive gradient clipping?

eracah: Looking in the code, it looks like it gets called right before `AFTER_TRAIN_BATCH` and right after `self._train_microbatch` which runs `AFTER_BACKWARD`, so I think `AFTER_BACKWARD` might be the appropriate event?

composer/algorithms/agc/README.md
```diff
@@ -0,0 +1,97 @@
+# 📎 AGC
```
hanlint, line 1: @jfrankle

hanlint (commented):
Code quality looks good to me -- defer to @abhi-mosaic for reviewing the implementation.

composer/algorithms/agc/agc.py
abhi-mosaic: I think AGC will have to occur on `AFTER_TRAIN_BATCH`, basically in the same location as our current gradient clipping, because it has to be after microbatch grad accumulation, and after gradient unscaling.

composer/algorithms/agc/agc.py
```diff
@@ -0,0 +1,161 @@
...
+        # computational graph.
+        weights = param.detach()
+        grad = param.grad.detach()
+
+        # Get clipped version of gradients.
+        clipped_grad = _get_clipped_gradients(weights, grad)
+
+        # Copy clipped gradients into param.grad attribute, so they can be accessed by
+        # optimizer.
+        grad.copy_(clipped_grad)
```
abhi-mosaic, line 54: This will increase the gradient memory by 2x, I think we need to do gradient scaling in-place, simlar to how `torch.nn.utils.clip_grad` does it: https://pytorch.org/docs/stable/_modules/torch/nn/utils/clip_grad.html#clip_grad_norm_

I also double-checked whether we can in-place multiply a tensor (grads) by a vector (unit-wise scales), and I think it works:
```
>>> a
tensor([[1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., 1., 1., 1.]])
>>> b
tensor([[3.],
        [4.],
        [5.]])
>>> a.mul_(b)
tensor([[3., 3., 3., 3.],
        [4., 4., 4., 4.],
        [5., 5., 5., 5.]])
```

abhi-mosaic (changes requested):
This looks really good so far! Awesome docs and tests. I think there are 2 changes that need to be addressed:

* algorithm should run on `AFTER_TRAIN_BATCH` rather than `AFTER_BACKWARD` due to microbatching and gradient scaling (AMP)
* gradient clipping should be in-place to avoid 2x memory overhead for grads

I think the tests should also be updated to catch the first issue, maybe we can re-use an existing convergence with both AMP and microbatching? That's really where the difference between `AFTER_TRAIN_BATCH` and `AFTER_BACKWARD` shows up.

Happy to help on either point and review again after changes!

abhi-mosaic:
Meta-comment -- if gradient clipping is a first-class Composer Trainer flag, should gradient clipping with AGC also be first class? It's a little weird having one as an algorithm and the other as not. I could see something like the below, just a thought. Maybe out of scope for this PR:

```yaml
trainer:
  gradient_clipping: 
    agc:
      threshold: 0.01
      eps: 1e-3
```
or
```yaml
trainer:
  gradient_clipping: 
    clip_by_global_norm:
      norm_type: 2
      threshold: 1.0
```
or
```yaml
trainer:
  gradient_clipping: 
    clip_by_value:
      threshold: 0.01
```

composer/algorithms/agc/agc.py
eracah: Oh good call, @abhi-mosaic! Also, sorry, @ravi-mosaicml, you said basically the same thing, but I misread it. Should be fixed now in 28dcddae23dc3f8ba5fccca3381acb32c3f3a692 . Thanks, folks!

composer/algorithms/agc/agc.py
eracah: Thanks, @abhi-mosaic! Good call! Should be fixed in 16148f1. PTAL. Thanks!

hanlint:
Long term, there will probably be many gradient clipping heuristics -- by putting this as an algorithm, it's probably more flexible and sets the right pattern.

eracah:
> * algorithm should run on `AFTER_TRAIN_BATCH` rather than `AFTER_BACKWARD` due to microbatching and gradient scaling (AMP)

Done. 28dcdda

> * gradient clipping should be in-place to avoid 2x memory overhead for grads

Done. 16148f1

Thanks, @abhi-mosaic !

eracah:
> Long term, there will probably be many gradient clipping heuristics -- by putting this as an algorithm, it's probably more flexible and sets the right pattern.

I agree! Maybe, I could make a design doc to discuss. I guess there are 3 options:
1. Leave as is
2. Make clip_by_global_norm an algorithm also.
3. Make AGC a first-class composer trainer flag

I actually like 2 the best.

abhi-mosaic:
> > Long term, there will probably be many gradient clipping heuristics -- by putting this as an algorithm, it's probably more flexible and sets the right pattern.
> 
> I agree! Maybe, I could make a design doc to discuss. I guess there are 3 options:
> 
> 1. Leave as is
> 2. Make clip_by_global_norm an algorithm also.
> 3. Make AGC a first-class composer trainer flag
> 
> I actually like 2 the best.

I think 2) makes sense to me as a longer term solution. Maybe a single algorithm called `gradient_clipping(method: str)` so we can group them all together?

composer/algorithms/agc/README.md
```diff
@@ -0,0 +1,97 @@
...
+    model.train()
+  
+    for epoch in range(num_epochs):
+        for X, y in train_loader:
+            y_hat = model(X)
+            loss = loss_fn(y_hat, y)
+            loss.backward()
+            cf.apply_agc(model)
+            opt.step()
+            opt.zero_grad()
```
abhi-mosaic, line 36: minor, could you move the `opt.zero_grad()` to the top of the inner loop? I think that's how PyTorch recommends doing it these days, and also what we do in Composer

composer/algorithms/agc/agc.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+
+    # Compute and clamp grad and weight norms.
+    w_norm = _unitwise_norm(weights).clamp_(min=eps)
+    grad_norm = _unitwise_norm(grad).clamp_(min=1e-6)
+
+    # Gradients whose norms are greater than weight_norm * clipping_threhsold are
+    # scaled down by (weight_norm * clipping_threhsold) / grad_norm.
+    max_norm = w_norm * clipping_threshold
+    clipped_grad_coeff = torch.where(grad_norm > max_norm, max_norm / grad_norm, torch.ones_like(grad_norm))
```
abhi-mosaic, line 133: Bit paranoid here, but could we do this in-place?
```
max_norm = w_norm.mul_(clipping_threshold)
clipped_grad_coeff = max_norm.div_(grad_norm).clamp_(max=1.0)

File: composer/algorithms/agc/agc.py [hunk 6]
```diff
@@ -0,0 +1,161 @@
...
+
+    return clipped_grad_coeff
+
+
+def _unitwise_norm(tensor: torch.Tensor):
+    """Implements unitwise norm as described in Brock et al, 2021.
+
+    For bias vectors (1D Tensors): normalize across entire vector.
+    For MLP Weight matrix (2D tensors): we normalize across rows (dim = 1)
+    For CNNs (4D Tensors): we normalzie across the entire kernel (channel, height,
```
abhi-mosaic, line 143:
```suggestion
    For CNNs (4D Tensors): we normalize across the entire kernel (channel, height,
```

abhi-mosaic, line 143: Could we be a bit more specific in the comment here that the normalization is intended to be per _output feature_? I think there is a little bit of danger depending on if someone's custom module stores their weight matrix in shape [out, in] or [in, out]. Just a quick sanity check for common PyTorch layers:

```
torch.nn.Linear weight shape: [out_features, in_features]
torch.nn.Conv2d weight shape: [out_channels, in_channels, kernel_height, kernel_width]
```

So when we do our unitwise norms, we are producing a norm per output feature for the above^

composer/algorithms/algorithm_registry.py
```diff
@@ -35,6 +35,7 @@
     'sam': SAMHparams,
     'alibi': AlibiHparams,
     'selective_backprop': SelectiveBackpropHparams,
+    'agc': AGCHparams
```
abhi-mosaic, line 38:
```suggestion
    'agc': AGCHparams,
```

composer/algorithms/agc/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+### Implementation Details
+
+AGC is implemented as follows. On `Event.AFTER_TRAIN_BATCH`:
+1. For every parameter in the model that has gradients:
+    a. The L2 norm of the parameter is computed (normalized across rows for MLP's, across entire filters for CNN's, and across the entire vector for biases).
+    b. The L2 norm of each parameters corresponding gradients are computed in a similar fashion.
+    c. If a norm of some gradients is greater than the norm of the correpsonding weights multiplied by the `clipping_threshold`:
+            * Scale all the gradients that contributed to that norm by the clipping threshold multiplied bu the ratio of weight norm to the gradient norm.
```
abhi-mosaic, line 67:
```suggestion
            * Scale all the gradients that contributed to that norm by the clipping threshold multiplied by the ratio of weight norm to the gradient norm.
```

composer/algorithms/agc/agc.py
```diff
@@ -0,0 +1,161 @@
...
+
+log = logging.getLogger(__name__)
+
+__all__ = ["AGC", "apply_agc"]
+
+
+def apply_agc(
+    model: torch.nn.Module,
+    clipping_threshold: float = 0.01,
+    eps: float = 1e-3,
```
abhi-mosaic, line 23: I'm not a fan of the `eps` argument, as we don't know ahead of time how the user is initializing their model weights. `1e-3` might be really big or really small.

Instead, can we jsut get rid of the `eps` entirely? We can remove all the clamps and epsilons except the very last one and I think it will be safe because `torch.clamp` replaces infinities:

```
>>> a = torch.ones(3) / torch.zeros(3)
>>> a
tensor([inf, inf, inf])
>>> torch.clamp(a, max=1)
tensor([1., 1., 1.])
```

abhi-mosaic (approved):
Awesome! Thank you for addressing the previous issues. I added a final round of suggestions... I think the only major one is the `eps` parameter, which might not actually be needed as `torch.clamp` can safely remove `infs` produced by divide-by-zero.

I glanced over the tests too and I think this is good to merge! Are we planning to run convergence tests on ImageNet after this PR is merged?

composer/algorithms/agc/README.md
eracah: done in 7b7bd14

composer/algorithms/agc/agc.py
eracah: oooh clever way of doing it fully in place. nice! ok fixed in 
2065360

composer/algorithms/agc/README.md
eracah: fixed in 
4a111e1

composer/algorithms/algorithm_registry.py
eracah: fixed in 
4a111e1

composer/algorithms/agc/agc.py
eracah: fixed in 
4a111e1

composer/algorithms/agc/agc.py
eracah: When I run that code I actually get nans which don't get clipped by the clamp call. Seems like we have to specify a min value somewhere? What do you think of keeping `grad_norm = _unitwise_norm(grad).clamp_(min=1e-6)`, but removing eps: `w_norm = _unitwise_norm(weights)`~~.clamp_(eps)~~, like I did here 
8ef0af4

composer/algorithms/agc/agc.py
eracah: ok fixed here: 51af085

eracah:
>I think the only major one is the `eps` parameter, which might not actually be needed as `torch.clamp` can safely remove `infs` produced by divide-by-zero.

Yeah I removed eps, but I kept a hard-coded clamp for the grad norm (as specified in the paper) because I kept getting nans when my tests tried dividing by zero and the `torch.clamp(max=1.0)` does not seem to resolve that situation. Here's the commit: 
8ef0af4

> I glanced over the tests too and I think this is good to merge! Are we planning to run convergence tests on ImageNet after this PR is merged?

I was planning on doing it after, but do you think we should do it before? That might delay things because I don't quite have a reliable setup for running convergence tests yet.

abhi-mosaic:
> Yeah I removed eps, but I kept a hard-coded clamp value for the grad norm (as specified in the paper) because I kept getting nans when my tests tried dividing by zero and the torch.clamp(max=1.0) does not seem to resolve that situation. Here's the commit:
> https://github.com/mosaicml/composer/commit/8ef0af4e104333f41ef182aeab313fa767a13a0c

Got it! I think that's the 0 / 0 case, but I think we can still resolve with this, should be same runtime as before since we got rid of a clamp:
```python
max_norm.div_(grad_norm).nan_to_num_(nan=1.0).clamp_(max=1.0)
```

Sorry to be so pedantic, I just have PTSD from epsilon bugs involving [FP16+Adam](https://github.com/pytorch/pytorch/issues/26218), and I try to avoid them wherever I can.

Also, it looks like PyTorch themselves [still uses an epsilon](https://github.com/pytorch/pytorch/blob/6e292f1a2185eee28a8645edadca09d32ef9600f/torch/nn/utils/clip_grad.py#L49), so maybe we should open a PR there too ;)

> I was planning on doing it after, but do you think we should do it before? That might delay things because I don't quite have a reliable setup for running convergence tests yet.

Convergence tests after this PR sounds fine to me, don't want to hold this up.

composer/algorithms/agc/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 MosaicML. All Rights Reserved.
+
+"""`Automatic Gradient Clipping <https://arxiv.org/abs/2102.06171>`_ Clips all gradients in model based on ratio of
```
hanlint, line 3: I think its Adaptive gradient clipping, not Automatic

eracah: Fixed in 9992338

eracah: Good catch, @hanlint ! I found another place where I did that too.

eracah:
oh ok make sense, @abhi-mosaic! good idea future proofing from precision errors. Your change passes all the tests. Fixed here: [5abb38b](https://github.com/mosaicml/composer/pull/924/commits/5abb38b4628ba141014cb9e3b3817b37ae606cc6)

PR #926, opened 2022-04-20
Make surgery algorithms torchscriptable; add initial tests

hanlint:
To help prepare our models for export post-surgery, we need to ensure they are TorchScript compatible. This PR has two components:
* Add tests for torchscript and model surgery methods
* Fixes torchscript issues with several algorithms

After this PR, the status of torchscriptable algorithms is as follows:

|                        | eval | train |
|------------------------|------|-------|
| apply_blurpool         | X    | X     |
| apply_factorization    |      |       |
| apply_ghost_batchnorm  | X    |       |
| apply_squeeze_excite   | X    | X     |
| apply_stochastic_depth | X    |       |
| apply_channels_last    | X    | X     |

Notes:
* `apply_ghost_batchnorm` (train mode): issue with accessing the `momentum` attribute of `BatchNorm` layers, filed an pytorch issue here: https://github.com/pytorch/pytorch/issues/76086
* `apply_stochastic_depth` (train mode): `torch.Generator` arguments are not supported, using during training
* `apply_factorization`: type-related issues that are yet to be investigated
 
A few open questions:
- [x] is it sufficient to convert to torchscript without failures, or should we also test the functionality of torchscript'ed models?
- [x] besides model surgery methods, are there any other methods that should be added to this test?


tests/algorithms/test_torchscript.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.parametrize("surgery_method", [
+    pytest.param(apply_blurpool),
+    pytest.param(apply_factorization, marks=pytest.mark.xfail),
+    pytest.param(apply_ghost_batchnorm, marks=pytest.mark.xfail),
+    pytest.param(apply_squeeze_excite),
+    pytest.param(apply_stochastic_depth),
+])
+def test_surgery_torchscript_eval(surgery_method, input):
```
hanlint, line 42: some duplicated code here for eval so reader can easily tell which algorithms are TS-supportable in what modes from the `parameterize` list.

tests/algorithms/test_torchscript.py
```diff
@@ -0,0 +1,51 @@
...
+def test_surgery_torchscript_train(surgery_method, input):
+    """Tests torchscript model in train mode."""
+    model = resnet50()
+    kwargs = algo_kwargs.get(surgery_method, {})
+
+    model = surgery_method(model, **kwargs)
+    model.train()
+
+    scripted_func = torch.jit.script(model)
+    scripted_func(input)  # type: ignore (third-party)
```
ajaysaini725, line 32: A simple check would be to verify the output is roughly the same for a scripted and non-scripted model (same for the eval check below)

tests/algorithms/test_torchscript.py
```diff
@@ -0,0 +1,51 @@
...
+def input():
+    return torch.Tensor(1, 3, 224, 224)
+
+
+@pytest.mark.parametrize("surgery_method", [
+    pytest.param(apply_blurpool),
+    pytest.param(apply_factorization, marks=pytest.mark.xfail),
+    pytest.param(apply_ghost_batchnorm, marks=pytest.mark.xfail),
+    pytest.param(apply_squeeze_excite),
+    pytest.param(apply_stochastic_depth, marks=pytest.mark.xfail),
```
ajaysaini725, line 21: Why does this fail for train and not for eval? Is the sample function that gets called during training not torch scriptable?

ajaysaini725 (commented):
Looks great!

Might be worth adding channels last to this (even though it's not really an architecture change it's work throwing it in just for completeness)

tests/algorithms/test_torchscript.py
hanlint: Yes it seems difficult to torch script, in particular the `torch.Generator` type. I believe if that were removed, it may work.

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -11,13 +11,13 @@
 
 
 def _default_2d_filter():
-    default_filter = np.array([[[
+    default_filter = torch.tensor([[[
```
hanlint, line 14: torchscript does not support `numpy`

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -155,36 +153,49 @@ def __init__(self, base_batchnorm: _TORCH_BATCHNORM_BASE_CLASS, ghost_batch_size
...
     def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
         batch_size = input.shape[0]
+
         if batch_size < self.ghost_batch_size:
             raise ValueError(f"Worker batch size {batch_size} < ghost_batch_size {self.ghost_batch_size}")
 
-        nchunks = int(np.ceil(batch_size / self.ghost_batch_size))
-        has_momentum = self._has_momentum()
-        if has_momentum:
+        nchunks: int = int(math.ceil(batch_size / self.ghost_batch_size))
```
hanlint, line 167: torchscript does not support numpy, using math instead

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -155,36 +153,49 @@ def __init__(self, base_batchnorm: _TORCH_BATCHNORM_BASE_CLASS, ghost_batch_size
...
         return torch.cat(normalized_chunks, dim=0)
 
     @staticmethod
-    def from_batchnorm(module: torch.nn.Module, ghost_batch_size: int) -> _GhostBatchNorm:
+    def from_batchnorm(module: torch.nn.Module, ghost_batch_size: int) -> "_GhostBatchNorm":
         assert isinstance(module, _TORCH_BATCHNORM_BASE_CLASS), "Module is not a BatchNorm subclass!"
         bn_type = _corresponding_ghost_batchnorm_type(module)
         return bn_type(ghost_batch_size=ghost_batch_size, base_batchnorm=module)
 
+    @torch.jit.unused
```
hanlint, line 191: To support `ghost_batchnorm` in `eval` mode, we refactor the offending lines into functions. Torchscript does not support accessing the `momentum` attribute, a pytorch issue has been filed.

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -1,11 +1,9 @@
 # Copyright 2021 MosaicML. All Rights Reserved.
 
-from __future__ import annotations
```
hanlint, line 3: torchscript runs into type inference issues when importing this feature

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -155,36 +153,49 @@ def __init__(self, base_batchnorm: _TORCH_BATCHNORM_BASE_CLASS, ghost_batch_size
         super().__init__()
         self.ghost_batch_size = ghost_batch_size
         self.batchnorm = base_batchnorm
+        self.current_nchunks: Optional[int] = None
```
hanlint, line 156:
```suggestion
```

composer/algorithms/stochastic_depth/stochastic_layers.py
```diff
@@ -66,18 +66,25 @@ def __init__(self, drop_rate: float, module_id: int, module_count: int, use_same
         self.use_same_depth_across_gpus = use_same_depth_across_gpus
         self.rand_generator = rand_generator
 
+    @torch.jit.unused
+    def _get_sample(self, device: int) -> torch.Tensor:
```
hanlint, line 70: We refactor the incompatible code into its own function that is only called during training.

`_sample_bernoulli` is not compatible because the function has an argument of type `torch.Generator`, and torchscript only supports built-ins and `Tensor` types.

hanlint:
@Landanjs @dblalock please check my code refactors carefully for correctness, I am not completely familiar with these algorithms.

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -161,30 +159,42 @@ def _has_momentum(self) -> bool:
...
         batch_size = input.shape[0]
+
         if batch_size < self.ghost_batch_size:
             raise ValueError(f"Worker batch size {batch_size} < ghost_batch_size {self.ghost_batch_size}")
 
-        nchunks = int(np.ceil(batch_size / self.ghost_batch_size))
-        has_momentum = self._has_momentum()
-        if has_momentum:
+        nchunks: int = int(math.ceil(batch_size / self.ghost_batch_size))
+        scale = self.batchnorm.momentum is not None
```
Landanjs, line 167: Nit: any reason to avoid calling this `has_momentum`? `scale` is ambiguous to me

Landanjs (approved):
LGTM besides a few nits on GhostBN.

Random thought: maybe there is a way to organize the tests as a class, but might not be possible and is not necessary

Landanjs:
As for other algorithms to test for torchscript, I've seen references to torchscripting data augmentations, but not sure how common this is

tests/algorithms/test_torchscript.py
Landanjs: If this becomes high priority, happy to refactor since I've wanted to refactor this for so long

dblalock (approved):
LGTM aside from agreeing with landan's ghostbn comments + one nitpick below

composer/algorithms/ghost_batchnorm/ghost_batchnorm.py
```diff
@@ -161,30 +159,42 @@ def _has_momentum(self) -> bool:
...
+        original_momentum: float = self.batchnorm.momentum
```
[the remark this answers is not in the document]
hanlint: Ah no - otherwise `original_momentum` is not bound during the second call (pyright complains).

tests/algorithms/test_torchscript.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.timeout(5)
+def test_surgery_torchscript_eval(surgery_method, input):
+    """Tests torchscript model in eval mode."""
+    model = resnet50()
+    kwargs = algo_kwargs.get(surgery_method, {})
+
+    surgery_method(model, **kwargs)
+    model.eval()
+
+    scripted_func = torch.jit.script(model)
```
schmidt-jake, line 62: Do you need to call `scripted_func.eval()` or will the scripted model already be in `eval` mode b/c `model.eval()` is called prior?

schmidt-jake (approved):
lgtm, I left one clarifying question! Thank you.

tests/algorithms/test_torchscript.py
hanlint: It would already be in `eval` mode, however I made this more clear by now doing:
```
scripted_func = torch.jit.script(model)
scripted_func.eval()
model.eval()

torch.testing.assert...
```
 in the test. Thank you for the feedback!

ajaysaini725 (approved):
LGTM 👍

PR #927, opened 2022-04-20
Daniel/proofread notebooks

mcneela:


mcneela:
@hanlint @dblalock Sorry, I accidentally deleted/closed the old PR. I made your requested changes, Davis, and they should be reflected in this new PR.

hanlint:
@mcneela for this, can you run `make clean-notebooks` to clean up the output?

mcneela:
> @mcneela for this, can you run `make clean-notebooks` to clean up the output?

@hanlint Done!

notebooks/Composer_Functional.ipynb
```diff
@@ -209,7 +213,7 @@
    "source": [
     "As we might expect, adding data augmentation doesn't help us when we aren't training long enough to start overfitting.\n",
     "\n",
-    "Now let's try using some algorithms that modify the model. We're going to keep things simple and just add a [Squeeze-and-Excitation](https://docs.mosaicml.com/en/latest/method_cards/squeeze_excite.html) module after the larger of the two conv2d operations in our model. We show examples of how to use other algorithms as well as a reference."
+    "Instead, let's try using some algorithms that modify the model. We're going to keep things simple and just add a [Squeeze-and-Excitation](https://docs.mosaicml.com/en/latest/method_cards/squeeze_excite.html) module after the larger of the two conv2d operations in our model."
```
dblalock, line 216: This isn't really "instead", since the data augmentation algorithms are still applied within the training dataloader.

notebooks/Composer_Functional.ipynb
```diff
@@ -200,7 +203,8 @@
     "model = Net()\n",
     "# only use one data augmentation since our small model runs quickly\n",
     "# and allows the dataloader little time to do anything fancy\n",
-    "train_and_eval(model, trainloader, testloader)"
+    "device = 'cuda' if torch.cuda.is_available() else 'cpu'\n",
```
dblalock, line 206: Can we rm the "device" arg in `train_and_eval` and set it using this line (within the function), instead of duplicating this line each time we call `train_and_eval`?

dblalock (commented):
Looks like the [docs](https://colab.research.google.com/github/mosaicml/composer/blob/main/notebooks/Composer_Functional.ipynb) aren't updated to point to an updated colab notebook? Same (I think) for the custom method tutorial notebook.

Is there a way to split up the changes (or reduce the number of lines changed) for the custom method tutorial? It's so many changes right now that github won't show the diff, which makes it difficult to comment on.

Also found a wording nitpick I should have caught the first time, sorry.

notebooks/Composer_Functional.ipynb
mcneela: Done

notebooks/Composer_Functional.ipynb
mcneela: Done

mcneela: I moved the custom_method_tutorial changes to a separate branch, but I'm not sure how to adjust the rewrite that git did. I just made the changes in the notebook locally (and ran `make clean-notebooks`, so should be no outputs) and git created the diff the way it did. They weren't really huge changes.

I think for Colab, I will have to create a new notebook on Colab and then point the docs to them? I suggest we merge these changes to the local notebooks first since this PR has been pending for a while, and then I will update the colab versions.

dblalock (approved):
LGTM. Like @mcneela said, we can handle the colab version + custom method notebook in a follow-up PR.

hanlint:
The colab version shouldn't be any different from this version though? our colab just points directly to the notebook on `main`, which would be updated after the next release.

mcneela:
> Looks like the [docs](https://colab.research.google.com/github/mosaicml/composer/blob/main/notebooks/Composer_Functional.ipynb) aren't updated to point to an updated colab notebooks?
> 
> Related, could you split up the changes (or reduce the number of lines changed) for the custom method tutorial? It's so many changes right now that github won't show the diff, which makes it difficult to comment on. Even when I inspect the diff manually, it's gargantuan and just shows one giant change with the whole file getting deleted and replaced (also, maybe notebook outputs getting checked in? there's a huge binary blob in the middle of the diff). tl;dr I have no way of reviewing these changes right now.
> 
> Also found a wording nitpick I should have caught the first time, sorry.



> The colab version shouldn't be any different from this version though? our colab just points directly to the notebook on `main`, which would be updated after the next release.

@hanlint I'm not sure how the colab linking works in the docs -- I was confused myself. That sounds correct though.

dblalock:
I don't know either way. I assumed it was manual since the built docs for this PR point to the old version, but having them point to `main` would also explain that. Sounds like Hanlin knows the actual behavior and that this behavior addresses the outdated colab concern.

composer/algorithms/blurpool/README.md
```diff
@@ -72,6 +72,10 @@ def train_model(model, train_dataloader):
 The Composer implementation of BlurPool uses model surgery to replace instances of pooling and downsampling operations with the BlurPool equivalents.
 For max pooling, it replaces `torch.nn.MaxPool2d` instances with instances of a custom `nn.Module` subclass that decouples the computation of the max within a given spatial window from the pooling and adds a spatial low-pass filter in between. This change roughly doubles the data movement required for the op, although it shouldn’t add significant overhead unless there are many maxpools in the network. For convolutions, it replaces strided `torch.nn.Conv2d` instances (i.e., those where the stride is larger than 1) with a custom module class that 1) applies a low-pass filter to the input, and then 2) applies a copy of the original convolution operation.
 
+🚧 Implementation Note
+>
+> Blurpool does not replace strided convolutions with fewer than 16 input channels. This is a heuristic used to avoid blurpooling the network's input. Doing so is undesirable since it amounts to downsampling the input by more than the amount specified in the preprocessing pipeline.
```
hanlint, line 77:
```suggestion
> Blurpool does not replace strided convolutions with fewer than `min_channels` input channels, which by default is set to `16`. This is a heuristic used to avoid blurpooling the network's input. Doing so is undesirable since it amounts to downsampling the input by more than the amount specified in the preprocessing pipeline. 
```

PR #936, opened 2022-04-21
Generalize progressive resizing

dskhudia:
Add delay_fraction and pin_sizes hyperparameters to progressive resizing algorithm. Addition of these two new hyperparameters doesn't change the default behavior of progressive resizing. 

delay_fraction allows us  to skip a fraction of training in the beginning and progressive resizing keeps input initial_scale-sized during this period. 

pin_sizes allows us to pin sizes of images to a multiple of pin_sizes, i.e., if pin_sizes is 32 image width will always be a multple of 32 after progressive resizing.

### DIFF
--- a/composer/algorithms/hparams.py
+++ b/composer/algorithms/hparams.py
@@ -237,6 +237,8 @@ class ProgressiveResizingHparams(AlgorithmHparams):
     initial_scale: float = hp.optional(doc="Initial scale factor", default=0.5)
     finetune_fraction: float = hp.optional(doc="Fraction of training to reserve for finetuning on full-sized inputs",
                                            default=0.2)
+    delay_fraction: float = hp.optional(doc="Fraction of training before resizing ramp begins", default=0.0)
+    size_increment: int = hp.optional(doc="Align sizes to a multiple of this number.", default=1)
     resize_targets: bool = hp.optional(doc="Also resize targets", default=False)
 
     def initialize_object(self) -> ProgressiveResizing:
--- a/composer/algorithms/progressive_resizing/progressive_resizing.py
+++ b/composer/algorithms/progressive_resizing/progressive_resizing.py
@@ -129,6 +129,8 @@ class ProgressiveResizing(Algorithm):
                                                 mode='resize',
                                                 initial_scale=1.0,
                                                 finetune_fraction=0.2,
+                                                delay_fraction=0.2,
+                                                size_increment=32,
                                                 resize_targets=False
                                             )
             trainer = Trainer(
@@ -148,6 +150,9 @@ class ProgressiveResizing(Algorithm):
             value in between 0 and 1. Default: ``0.5``.
         finetune_fraction (float, optional): Fraction of training to reserve for finetuning on the
             full-sized inputs. Must be a value in between 0 and 1. Default: ``0.2``.
+        delay_fraction (float, optional): Fraction of training before resizing ramp begins.
+            Must be a value in between 0 and 1. Default: ``0.0``.
+        size_increment (int, optional): Align sizes to a multiple of this number. Default: ``1``.
         resize_targets (bool, optional): If True, resize targets also. Default: ``False``.
     """
 
@@ -155,6 +160,8 @@ class ProgressiveResizing(Algorithm):
                  mode: str = 'resize',
                  initial_scale: float = .5,
                  finetune_fraction: float = .2,
+                 delay_fraction: float = .0,
+                 size_increment: int = 1,
                  resize_targets: bool = False):
 
         if mode not in _VALID_MODES:
@@ -166,9 +173,15 @@ class ProgressiveResizing(Algorithm):
         if not (0 <= finetune_fraction <= 1):
             raise ValueError(f"finetune_fraction must be between 0 and 1: {finetune_fraction}")
 
+        if not (delay_fraction + finetune_fraction <= 1):
+            raise ValueError(
+                f"delay_fraction + finetune_fraction must be less than 1: {delay_fraction + finetune_fraction}")
+
         self.mode = mode
         self.initial_scale = initial_scale
         self.finetune_fraction = finetune_fraction
+        self.delay_fraction = delay_fraction
+        self.size_increment = size_increment
         self.resize_targets = resize_targets
 
     def match(self, event: Event, state: State) -> bool:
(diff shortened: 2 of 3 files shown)


composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -197,14 +210,23 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
...
+            scale_frac_elapsed = 0.0
 
         # Linearly increase to full size at the start of the fine tuning period
-        scale_factor = initial_size + (1 - initial_size) * scale_frac_elapsed
+        scale_factor = self.initial_scale + (1 - self.initial_scale) * scale_frac_elapsed
+
+        # adjust scale factor so that we make width a multiple of pin_sizes
+        width = input.shape[3]
+        scaled_width_pinned = round(width * scale_factor / self.pin_sizes) * self.pin_sizes
+        scale_factor_pinned = scaled_width_pinned / width
```
dblalock, line 225: should this use `max(width, height)` (or max of `input.shape[2:]`) instead of `width`? Or maybe min so that the worst-case dim is a mulitple of pin_sizes? I'm actually not sure what the right semantics are but I would kind of expect the resizing to not designate a particular dimension as special.

dblalock (approved):
Just little nitpicks, except for a single part that (might?) have a typo or something in a formula

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -197,14 +210,23 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
         # Calculate the current size of the inputs to use
         initial_size = self.initial_scale
         finetune_fraction = self.finetune_fraction
-        scale_frac_elapsed = min([state.get_elapsed_duration().value / (1 - finetune_fraction), 1])
+        dur_elapsed = state.get_elapsed_duration().value
+        if dur_elapsed >= self.delay_fraction:
+            scale_frac_elapsed = min([dur_elapsed / (1 - self.finetune_fraction - self.delay_fraction), 1])
```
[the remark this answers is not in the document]
coryMosaicML: Agree with Davis here, this looks a bit funny. For example, if `dur_elapsed = self.delay_fraction` one gets `scale_frac_elapsed = self.delay_fraction/(1 - self.finetune_fraction - self.delay_fraction)` which means `scale_frac_elapsed` will change discontinuously when `dur_elapsed` becomes larger than `self.delay_fraction` if `self.delay_fraction` is not zero. I don't think this is the intended behavior.

coryMosaicML (changes requested):
Some general comments:

Before this is ready, we should:
- Verify that the image sizes are behaving as expected for different values of `delay_fraction`
- Update `README.md` to include discussion of the new hparams and some recommended settings
- Add tests for the new behavior if possible
- Check experimentally what the effect of these new hparams are on accuracy and runtime.

composer/algorithms/progressive_resizing/progressive_resizing.py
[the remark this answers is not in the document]
dskhudia: @dblalock @coryMosaicML 
Thanks for catching this. It should be `scale_frac_elapsed = min([(dur_elapsed - self.delay_fraction) / (1 - self.finetune_fraction - self.delay_fraction), 1])` and we will get a linear ramp.

composer/algorithms/progressive_resizing/progressive_resizing.py
dskhudia: Width is an inner dimension (NHWC). I made width to be pinned to a multiple of pin_sizes (say 32) and if cuda kernels rely on inner most after reduction dim(C) to be a nice multiple of a power of 2 we get performance benefits. If H == W, H will also be a multiple of pin_sizes otherwise it will be whatever it becomes after multiplying with scale_factor_pinned.

composer/algorithms/hparams.py
```diff
@@ -232,6 +232,9 @@ class ProgressiveResizingHparams(AlgorithmHparams):
     initial_scale: float = hp.optional(doc="Initial scale factor", default=0.5)
     finetune_fraction: float = hp.optional(doc="Fraction of training to reserve for finetuning on full-sized inputs",
                                            default=0.2)
+    delay_fraction: float = hp.optional(
+        doc="Fraction of training in the beginning to keep at initial-scale-sized inputs", default=0.0)
```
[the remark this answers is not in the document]
dskhudia: I can change it suggested but does it make it clear that resizing is still happening but we do not start ramping up the sizes until `delay_fraction` training has passed.

dskhudia:
@coryMosaicML See embedded comments
> Verify that the image sizes are behaving as expected for different values of delay_fraction
Do you mean verifying by running an example run? 

> Update README.md to include discussion of the new hparams and some recommended settings
Will do. We don't have recommended settings yet though. 

> Add tests for the new behavior if possible
Are there existing unit tests for algorithms? 

> Check experimentally what the effect of these new hparams are on accuracy and runtime.
These experiments are planned but not done yet. However, with default values they don't change existing behavior.

coryMosaicML:
> Do you mean verifying by running an example run?

Mostly this was aimed at the above issue with `scale_frac_elapsed` and it looks like that will get resolved. But this can also be an example run (combined with the experiments suggested below)
 
> Will do. We don't have recommended settings yet though.

I think some experimental runs will help with this. In particular if we observe a benefit with `pin_sizes` that would be really good to include as a recommended setting, or even perhaps make as a default.

> Are there existing unit tests for algorithms?

Tests for Progressive Resizing can be found [here](https://github.com/mosaicml/composer/blob/dev/tests/algorithms/test_progressive_resizing.py)

> These experiments are planned but not done yet. However, with default values they don't change existing behavior.

For changes that introduce extra hparams we should ideally be able to show some kind of accuracy/speed benefit from using them since they increase complexity and can make later hparams sweeps more expensive. I'm fairly certain we'll see a benefit from `pin_sizes` and from a properly tuned `delay_fraction` but it would be good to verify. In the case of `pin_sizes` if the experimental results are strong enough we might want to consider making it a default as well.

dskhudia:
Cool. Sounds good. Let me do an experimental run to just verify it works as expected and add a test. 

Recommended settings and Readme update: I can do this in a followup PR once we have experimental data to back it up.

dskhudia:
@coryMosaicML This is how the dimensions are changing in a 100 epoch experimental run with delay_fraction of .4 and pin_sizes = 8. 

For first 40% of the training stays at 112 (50% reduction) and ramps up in increment of 8 till 80% of the training and stays at 224 (100% size) afterwards. 
[change.log](https://github.com/mosaicml/composer/files/8535763/change.log)

dskhudia:
@coryMosaicML Modified test to cover new hparams.

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -129,6 +129,8 @@ class ProgressiveResizing(Algorithm):
                                                 mode='resize',
                                                 initial_scale=1.0,
                                                 finetune_fraction=0.2,
+                                                delay_fraction=0.2,
```
dblalock, line 132: default delay fraction should be consistent everywhere

dblalock (approved):
LGTM aside from slight nitpicks. Also, we do need to update the method card like Cory said, but I'll leave it to your judgment whether to put that in this PR or a separate one.

coryMosaicML (approved):
LGTM!

Only thing might be to update the readme with some discussion of hparams to avoid documentation debt, but it can also be done later.

composer/algorithms/progressive_resizing/progressive_resizing.py
dskhudia: Default is 0.0 everywhere for delay_fraction. It's 0.2 in this example.

composer/algorithms/progressive_resizing/progressive_resizing.py
```diff
@@ -129,6 +129,8 @@ class ProgressiveResizing(Algorithm):
...
+                                                pin_sizes=32,
```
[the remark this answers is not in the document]
dskhudia: Changed it to ```size_alignment```

composer/algorithms/progressive_resizing/progressive_resizing.py
dblalock: ah, good point; I failed at reading the surrounding context, sorry

PR #943, opened 2022-04-21
Add `min_channels` to blurpool

hanlint:
BlurPool currently uses the hardcoded heuristic of `in_channels > 16` to prevent bluring the first input layer (essentially blurring the input image). 

This PR exposes that as an argument to surface the behavior and allow it to be configured. This helps with testing blurpool as we can then use existing `SimpleConvModel` (in_channels=8).

NB: unclear to me if we actually want to continue using this heuristic, or if a `ignore_first_layer: bool = True` is more appropriate, as using `min_channels` seems a bit of a hack.

### DIFF
--- a/composer/algorithms/blurpool/blurpool.py
+++ b/composer/algorithms/blurpool/blurpool.py
@@ -24,6 +24,7 @@ def apply_blurpool(model: torch.nn.Module,
                    replace_convs: bool = True,
                    replace_maxpools: bool = True,
                    blur_first: bool = True,
+                   min_channels: int = 16,
                    optimizers: Optional[Union[Optimizer, Sequence[Optimizer]]] = None) -> torch.nn.Module:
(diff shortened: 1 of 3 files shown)


composer/algorithms/blurpool/blurpool.py
```diff
@@ -93,12 +97,21 @@ class BlurPool(Algorithm):
             applied with a stride of 1 before the blurring, resulting in
             significant overhead (though more closely matching the paper).
             See :class:`.BlurConv2d` for further discussion. Default: ``True``.
+        min_channels (int, optional): Skip replacing layers with less than ``min_channels``.
+            Commonly used to prevent the blurring of the first layer. Default: 16.
```
dblalock, line 101: I think the above description ^ is a little clearer.
```suggestion
        min_channels (int, optional): Skip replacing layers with in_channels < min_channels.
            Commonly used to prevent the blurring of the first layer. Default: 16.
```

dblalock (approved):
LGTM

PR #944, opened 2022-04-22
Bind the `Logger` to `ComposerModel`

ravi-mosaicml:
- Bind the `Logger` to `ComposerModel`
- Added test case


A-Jacobson (approved):
Lgtm, pretty simple change that we discussed in the design meeting.

PR #945, opened 2022-04-22
Default to the `root` user in the Dockerfile

Averylamp:
This PR switches the default user in the Dockerfile to `root`. This is necessary to match MCLI default behavior and because RDMA requires root access. The `mosaicml` user is still exists and can be used to run as non-root.


bandish-shah (approved):
LGTM

docker/pytorch/Dockerfile
```diff
@@ -162,9 +162,6 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
 
 WORKDIR /home/mosaicml
 
-USER mosaicml
```
bandish-shah, line 165: [nit] Can we also just get rid of the `useradd` for `mosaicml` user above?

docker/pytorch/Dockerfile
```diff
@@ -162,9 +162,6 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
 
 WORKDIR /home/mosaicml
```
dskhudia, line 164: Need to remove the lines above? 

```
RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml && \
    usermod -a -G sudo mosaicml && \
    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
    ```

**Comment by ravi-mosaicml** [comment]
I would like to keep the mosaicml user, as editable installs do not work on root. Can MCLI just set the user argument for the pod if it requires root?

**Comment by bandish-shah** [comment]
> I would like to keep the mosaicml user, as editable installs do not work on root. Can MCLI just set the user argument for the pod if it requires root?

The `Dockerfile` doesn't do any editable installs.  Am I correct in saying we only need this for interactive development and for `composer`?  In which case I'm not sure how this applies bc we switch to using the `root` user in the `pytorch_internal` images anyways and it's largely expected internal interactive dev is done one these images anyways.  Iirc editable installs work in a virtualenv for the `root` user so there are several mitigations to this vs. having `mcli` perform image user management.

**Comment by dskhudia** [comment]
@Averylamp Is there plan to merge this?

File: docker/pytorch/Dockerfile (hunk 1 above)
Comment (ravi-mosaicml) [reply docker/pytorch/Dockerfile:165]: I'd like to keep this user around for now, should someone want to run the docker image not as root (e.g. so permissions don't get screwed up if using a volume mount). But based off the PR discussion, agree it does make sense to default to root.

File: docker/pytorch/Dockerfile [hunk 3]
```diff
@@ -162,9 +162,6 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
...
```
ravi-mosaicml: This should be removed if not defaulting to the `mosaicml` user.

ravi-mosaicml (approved):
LGTM

PR #946, opened 2022-04-22
Added `Timer.reset`

ravi-mosaicml:
Added a `Timer.reset` method and test cases, which will be needed for multiple calls to fit()


tests/test_time.py
```diff
@@ -68,3 +68,11 @@ def test_timer():
     assert timer.sample_in_epoch == 5
     assert timer.token == 20
     assert timer.token_in_epoch == 0
+    timer.reset()
```
hanlint, line 71: nit: this should be a separate timer test:
```
def test_timer_reset():
```

hanlint (approved):

PR #948, opened 2022-04-22
Multiple calls to `Trainer.fit()`

### LINKED ISSUE #138: Multiple calls to .fit

ravi-mosaicml:
This PR allows for arguments to be specified when calling `Trainer.fit()` instead of when constructing the trainer.

1. Refactored `Trainer.init` to move out shared helper code, so it can be re-used when passing in parameters on `fit`.
2. Modified the signature of `fit` to take training parameters. If not specified on `Trainer.__init__`, parameters must be specified on `fit`.
3. Rearranged the position of arguments in the `Trainer` and `TrainerHparams` classes to group by functionality. Re-ordered the docstrings to be in the same order as the arguments.

Closes #138.


composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+                if `persistent_workers=True` and the dataloader has already been iterated,
+                or if the dataloader is mid-epoch. It is required that the training dataloader
+                does not have an active iterator, so CPU dataset augmentations can be
+                correctly inserted.
+
+                To fix, please do not iterate over the dataloader before passing it into
+                the trainer."""))
+    return data_spec
+
+
```
hanlint, line 216: maybe this:
```
  if evaluator.should_eval is None:
                if callable(eval_interval):
                    evaluator.should_eval = eval_interval
                else:
                    evaluator.should_eval = evaluate_periodically(eval_interval)
            if evaluator.subset_num_batches is None:
                evaluator.subset_num_batches = subset_num_batches
```

should actually be a function within the evaluator class? e.g. something like:
```
    def update_eval_interval:
```

```

File: composer/trainer/trainer.py [hunk 2]
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+    if adaptive_grad_accum:
+        grad_accum = 1
+        warnings.warn(textwrap.dedent("""Setting `grad_accum='auto'` is an experimental feature
+            which may cause uncaught Cuda Out of Memory errors. In this case, please manually
+            set grad_accum explicitly to an integer instead.
+            """),
+                      category=UserWarning)
+    # Cannot use adaptive grad accum on CPU
+    if isinstance(device, DeviceCPU) and adaptive_grad_accum:
+        raise ValueError("Cannot use adaptive grad_accum on CPU. Please set grad_accum >= 1")
```
hanlint, line 261: why not? Understand that there would never be a CUDA OOM, but allowthing this make its easier for folks to swtich between testing on CPU vs GPU. Maybe we allow this ,and this should be a no-op? cc: @mvpatel2000

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+    # grad_accum should be int as we've already handled "auto" case
+    if isinstance(grad_accum, str):
+        raise ValueError("grad_accum must be an int or ``auto``")
```
hanlint, line 264: This doesn't actually enforce that grad_accum is an `int`. e.g. I could have passed in `grad_accum=1.5` and that would have passed this check. Instead, we want the error message to match the check. Suggest moving this guard earlier to something like..
```
if grad_accum == "auto":
    return (1, True)  # grad_accum, adaptive_grad_accum
if not isinstance(grad_accum, int):
    raise ValueError(...)
return (grad_accum, False)
```

thoughts?

cc: @mvpatel2000

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+    return grad_accum, adaptive_grad_accum
+
+
+def _is_cuda_oom(e: RuntimeError):
+    """Determines if error is CUDA Out of Memory and if adaptive_grad_accum is enabled."""
+    return "CUDA out of memory" in str(e)
+
+
+def _unpack_device(device: Optional[Union[str, Device]], deepspeed_enabled: bool):
```
hanlint, line 273: `deepspeed_enabled` is not used here?

rename to `_get_device()`

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+        if device.lower() == 'cpu':
+            device = DeviceCPU()
+        elif device.lower() == 'gpu':
+            device = DeviceGPU()
+        else:
+            raise ValueError(f'device ({device}) must be one of (cpu, gpu).')
+    return device
+
+
+def _unpack_seed(seed: Optional[int], device: Device):
```
hanlint, line 286: This function does many things (1) uses a default seed, (2) distributes the rank zero seed to all ranks, and (3) modifies the seed by adding the rank. Any ideas to simplify?

rename: `_get_random_seed`

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+    if seed < 0 or seed > reproducibility.MAX_SEED:
+        raise ValueError(f"Invalid seed: {seed}. It must be on [0; 2**32 - 1)")
+
+    # using int64 to prevent overflow
+    rank_zero_seed = device.tensor_to_device(torch.tensor([seed], dtype=torch.int64))
+    dist.broadcast(rank_zero_seed, src=0)
+    rank_zero_seed = rank_zero_seed.item()
+    assert isinstance(rank_zero_seed, int)
+    seed = rank_zero_seed + dist.get_global_rank()
+    log.info(f"Setting seed to {seed}")
```
hanlint, line 303: seed is actually not being set here

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+    return rank_zero_seed, seed
+
+
+def _unpack_max_duration(max_duration: Union[int, str, Time]):
+    if isinstance(max_duration, str):
+        max_duration = Time.from_timestring(max_duration)
+    elif isinstance(max_duration, int):
+        max_duration = Time.from_epoch(max_duration)
```
hanlint, line 311: rename to `parse_max_duration`.

question: should this functionality be in `Time`?

composer/trainer/trainer.py
```diff
@@ -140,6 +246,140 @@ def _unpack_evaluators(
...
+        if find_unused_parameters:
+            ddp_sync_strategy = DDPSyncStrategy.FORCED_SYNC
+        else:
+            ddp_sync_strategy = DDPSyncStrategy.SINGLE_AUTO_SYNC
+    else:
+        ddp_sync_strategy = DDPSyncStrategy(ddp_sync_strategy)
+    return ddp_sync_strategy
+
+
+def _unpack_profiler(
```
hanlint, line 326: This maybe outside the scope, but i would be in favor of setting a:
```
trainer = Trainer(
    enable_profiler: bool = True
)
```

and then any other fancy configuration someone should just create and pass in the object (e.g. unflatten)?

composer/trainer/trainer.py
```diff
@@ -1018,9 +1146,150 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
 
-    def fit(self):
-        """Train and evaluate the model on the provided data."""
-        # Print any exception, so it can be caputred by any callbacks or loggers (e.g. WandB, FileLogger)
+    def fit(
+        self,
+        *,
+        # Train Dataloader
+        train_dataloader: Optional[Union[Iterable, DataSpec, Dict[str, Any]]] = None,
+        train_dataloader_label: str = "train",
```
hanlint, line 1154: is this new? what is the use case for this label?

composer/trainer/trainer.py
```diff
@@ -1018,9 +1146,150 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+        train_subset_num_batches: Optional[int] = None,
+        compute_training_metrics: Optional[bool] = None,
```
hanlint, line 1156: this is barely used, I think we can remove this option?

hanlint (commented):
Thanks for doing this, this is a significant endeavour! I made a first pass on some helper functions and tests.

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
 __all__ = ["Trainer"]
 
 
+class MissingArgumentException(ValueError):
```
ajaysaini725, line 111: Should either move this out to a util file or make it a private class prefixed with `_`

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+    log.info(f'max_duration changed from {orig_max_duration} to {max_duration}')
+    if max_duration.value == 0:
+        raise ValueError('Scale schedule has reduced the max_duration to 0. Set a higher ratio or use more epochs.')
+    return max_duration
+
+
+def _unpack_step_schedulers_every_batch(
+    schedulers: Union[None, ComposerScheduler, PyTorchScheduler, Sequence[Union[ComposerScheduler, PyTorchScheduler]]],
+    step_schedulers_every_batch: Optional[bool],
+):
```
ajaysaini725, line 134: Seems like this function is missing a return type?

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+                               orig_max_duration: Optional[Time[int]]) -> Optional[Time[int]]:
+    if orig_max_duration is None:
+        return None
+    max_duration = cast(Time[int], orig_max_duration * scale_schedule_ratio)
+    log.info(f'max_duration changed from {orig_max_duration} to {max_duration}')
+    if max_duration.value == 0:
+        raise ValueError('Scale schedule has reduced the max_duration to 0. Set a higher ratio or use more epochs.')
+    return max_duration
+
+
```
ajaysaini725, line 130: Thoughts on moving all of these `unpack` functions to a separate helper file and importing them here for readability purposes? This `trainer.py` file is almost 2k lines long - would be good to shorten it somehow.

composer/trainer/trainer.py
ajaysaini725: Yeah for example we could have a `ProfilerConfig` class that has all these other params (I know this is basically hparams but we do want to explicitly avoid hparams use here)

composer/trainer/trainer.py
ajaysaini725: This was originally added because researchers requested it - we should prob confirm with people first

ajaysaini725 (commented):
Nice job! Left some preliminary comments

composer/trainer/trainer.py
ravi-mosaicml: I'm not too sure about the consequences of an OOM on CPU. (The oom would be determined by the amount of RAM or the k8s limit). I think K8s automatically kills processes if they try to use too much ram, so there would be nothing to recover, but am not entirely sure. Anyways, tracking this feature request [here](https://mosaicml.atlassian.net/browse/CO-295).

composer/trainer/trainer.py
ravi-mosaicml: Good point. I will make this a private helper function rather than a subclass.

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+def _unpack_step_schedulers_every_batch(
```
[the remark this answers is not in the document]
ravi-mosaicml: I like 2) where possible. Re naming, makes sense -- will split up further and rename as appropriate.

composer/trainer/trainer.py
ravi-mosaicml: A good number of those lines are docstrings :)
But yes, I can move it to a private helper file.

composer/trainer/trainer.py
ravi-mosaicml: PyRight can automatically infer return types, so they generally are not needed (except for abstract methods).

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+                textwrap.dedent("""\
+                The `train_dataloader` has an active iterator. This could occur
+                if `persistent_workers=True` and the dataloader has already been iterated,
+                or if the dataloader is mid-epoch. It is required that the training dataloader
+                does not have an active iterator, so CPU dataset augmentations can be
+                correctly inserted.
+
+                To fix, please do not iterate over the dataloader before passing it into
+                the trainer."""))
+    return data_spec
```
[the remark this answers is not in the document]
ravi-mosaicml: Yup! Will do.

composer/trainer/trainer.py
ravi-mosaicml: Passing in anything other than a float or a string would not pass type checking. PyRight infers this automatically and types the return value as an (int, bool) implicitely. I can add an explicit type check, though, in case if the user passed in something else.

composer/trainer/trainer.py
ravi-mosaicml: Makes sense. I can fix by either allowing a timestring as a single-argument in the `Time` constructor, or move this functionality to an `ensure_time` helper in the Time class.

composer/trainer/trainer.py
hanlint: That is important, we cannot assume the user is using pyright.

composer/trainer/trainer.py
ravi-mosaicml: Agree it would be nice to unflatten it. Tracking [here](https://mosaicml.atlassian.net/browse/CO-296). There are a few ways to address this -- one is to pass in the profiler object directly (would require a bit of re-writing, as the `Profiler` constructor currently requires the state, or as @ajaysaini725 mentioned, pass in a config/hparams class.

composer/trainer/trainer.py
```diff
@@ -625,106 +875,37 @@ def __init__(
...
-            else:
-                raise ValueError(f'device ({device}) must be one of (cpu, gpu).')
-        else:
-            if not isinstance(device, Device):
-                raise ValueError('device must be of class Device')
-            self._device = device
-
-        if self.deepspeed_enabled or dist.get_world_size() > 1:
+        # Determine whether DeepSpeed is enabled
+        if deepspeed_config is True:
```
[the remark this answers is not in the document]
ravi-mosaicml: This won't work, as `deepspeed_config` could also be a dictionary with parameters already specified. The one-liner would have to be something like `deepspeed_config = {} if deepspeed_config is True else None if deepspeed_config is False else deepseed_config` which I would have trouble understanding :D (the explicit True/False checks are required since, if the user passes in an empty dictionary, that should enable deepspeed, but `bool({}) is False`)

composer/trainer/trainer.py
ravi-mosaicml: This is if you want to train on seperate datasets via multiple calls to fit, and track metrics separately (since metrics are indexed by the dataloader label)

composer/trainer/trainer.py
ravi-mosaicml: Going to leave this here for now, but this will be addressed in metrics API refactor.

I suspect this parameter would be renamed to something like `train_metrics: Optional[List[str]] = None` where the user specifies the metric names to compute on the train dataset.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Sounds good; we can simplify like that!

composer/trainer/trainer.py
ravi-mosaicml: OK, I will add the check here. But FYI -- I don't think we do explicit type checks anywhere else in the codebase to validate that the user passed in a type that matches the typing annotation. (This is generally undefined behavior -- it may work, it could fail with an exception later, or it could silently work but do something different than expected).

tests/trainer/test_trainer.py
```diff
@@ -1,82 +1,428 @@
...
+    ):
+        train_dataloader = DataLoader(RandomClassificationDataset())
+        trainer = Trainer(
+            model=model,
+            train_dataloader=train_dataloader,
+            save_interval=10,
+            max_duration=1,
+            save_folder=str(tmpdir),
+        )
+        assert trainer.state.max_duration == "1ep"
```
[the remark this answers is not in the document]
ravi-mosaicml: Max duration is checked in `TestTrainerInitOrFit.test_max_duration`; will remove the assert here.

tests/trainer/test_trainer.py
```diff
@@ -1,82 +1,428 @@
...
+        checkpoint_saver = None
+        for callback in trainer.state.callbacks:
+            if isinstance(callback, CheckpointSaver):
+                checkpoint_saver = callback
+        assert checkpoint_saver is not None
+        trainer.state.timer.epoch._value = 10
+        assert checkpoint_saver.save_interval(trainer.state, Event.EPOCH_CHECKPOINT)
```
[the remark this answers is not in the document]
ravi-mosaicml: Yeah, I'm not sure why I added this test. Going to remove, since we already have this covered in `test_checkpoint.py`

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Added an `ensure_data_spec` helper, and move the iterator check into the DataSpec class.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Done

tests/trainer/test_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Ended up removing the `test_checkpoint` check.

tests/trainer/test_trainer.py
```diff
@@ -1,82 +1,428 @@
...
+            save_interval=10,
+            max_duration=1,
+            save_folder=str(tmpdir),
+        )
+        assert trainer.state.max_duration == "1ep"
+        checkpoint_saver = None
+        for callback in trainer.state.callbacks:
+            if isinstance(callback, CheckpointSaver):
+                checkpoint_saver = callback
+        assert checkpoint_saver is not None
```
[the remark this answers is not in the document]
ravi-mosaicml: Ended up removing this test

tests/trainer/test_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Removed the test

tests/trainer/test_trainer.py
```diff
@@ -1,82 +1,428 @@
...
+        # Train a model
+        train_dataloader = DataLoader(RandomClassificationDataset())
+        optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+        max_duration = 1
+        trainer = Trainer(
+            model=model,
+            max_duration=max_duration,
+            train_dataloader=train_dataloader,
+            optimizers=optimizer,
+        )
```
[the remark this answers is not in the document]
ravi-mosaicml: Done

composer/trainer/trainer.py
```diff
@@ -108,6 +108,112 @@
...
+             "`step_schedulers_every_batch` to True, or alternatively use a ComposerScheduler. For more information, "
+             f"see https://docs.mosaicml.com/en/v{composer.__version__}/trainer/schedulers.html."))
+
+    else:
+        step_schedulers_every_batch = True
+
+    return step_schedulers_every_batch
+
+
+def _unpack_compute_training_metrics(compute_training_metrics: bool, model: ComposerModel):
```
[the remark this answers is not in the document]
ravi-mosaicml: Done

composer/trainer/trainer.py
ravi-mosaicml: Done

composer/trainer/trainer.py
ravi-mosaicml: Done

composer/trainer/trainer.py
ravi-mosaicml: I renamed to `_get_random_seed`.

composer/trainer/trainer.py
ravi-mosaicml: Added an `ensure_time` helper function; removed `_unpack_max_duration`.

composer/trainer/trainer.py
ravi-mosaicml: Converted to a private helper function.

composer/trainer/trainer.py
ravi-mosaicml: After further thought, I'd prefer to leave it in `trainer.py`. While it would help with limiting git diffs if they were in a separate file, I do think it's helpful to keep them here, as they do belong as part of the trainer (perhaps they should be private, static methods?)

I know the trainer is long, but not _unreasonably_ long. Both the PTL and HF training loops are on a similar order of magnitude.

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/trainer.py
https://github.com/huggingface/transformers/blob/v4.18.0/src/transformers/trainer.py

FastAI is way shorter, but they "outsource" a good amount of logic to callbacks.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: I took a pass through refactoring these helpers. Could you take another look?

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+                    warnings.warn("Specifying `eval_subset_num_batches` without an `eval_dataloader` has no effect.")
+                if eval_interval != 1:
+                    warnings.warn("Specifying `eval_interval` without an `eval_dataloader` has no effect.")
+
+        if len(self.evaluators) == 0:
+            warnings.warn(("No `eval_dataloader` was specified. Please specify `eval_dataloader` to periodically "
+                           "evaluate your model while training."))
+
+        # Grad Accum
+        if grad_accum is not None:
```
hanlint, line 1230: for readability, and to downscope the helper functions, I wonder if this should be:
```
if grad_accum == 'auto':
     self.state.grad_accum, self.adaptive_gradient_accumulation = _configure_adaptive_ga(grad_accum, self._device)
elif grad_accum is not None:
    self.state.grad_accum = grad_accum
```

```

File: composer/trainer/trainer.py [hunk 23]
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+            self.state.grad_accum, self.adaptive_gradient_accumulation = _get_grad_accum(grad_accum, self._device)
+
+        # Grad Clip Norm
+        if grad_clip_norm is not None:
+            if self.state.is_model_deepspeed:
+                raise ValueError(
+                    ("When using DeepSpeed, the `grad_clip_norm` must be set when constructing the . "
+                     f"{type(self).__name__} -- e.g. via {type(self).__name__}(grad_clip_norm=...). It cannot be "
+                     f"specified on {type(self).__name__}.{type(self).fit.__name__}()."))
```
hanlint, line 1239:
```suggestion
                raise ValueError("Changing the grad_clip_norm when using DeepSpeed is not supported.")
```

edited to be shorter and mirror the message below re: precision.

hanlint: Don't think we need the complexity of the handle the `type` and names etc?

composer/trainer/trainer.py
```diff
@@ -1348,7 +1579,7 @@ def _train_microbatches(self, microbatches: Sequence[Batch], ddp_sync: bool = Tr
 
             use_grad_scaling = self._use_grad_scaling(self.state.precision, self.state.scaler)
 
-            if not self.deepspeed_enabled:
+            if not self.state.is_model_deepspeed:
```
hanlint, line 1582: having a trainer level argument on `deepspeed_enabled` was very helpful, can we keep that?

hanlint (commented):
Took another pass at the review, I think this is close!

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+
+            All arguments to :meth:`.fit` are optional. Any values specified here will override
+            what was provided when constructing the :class:`.Trainer`.
+
+        Args:
+            train_dataloader (Iterable | DataSpec | Dict[str, Any], optional): See :class:`.Trainer`.
+            train_dataloader_label (str, optional): See :class:`.Trainer`.
+            train_subset_num_batches (int, optional): See :class:`.Trainer`.
+            compute_training_metrics (bool, optional): See :class:`.Trainer`.
+            max_duration (Time[int] | str | int, optional): See :class:`.Trainer`.
```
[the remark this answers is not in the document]
ravi-mosaicml: Good point; I ran into this when debugging the tests. I think it could make sense to raise an exception if `state.timer == max_duration`? Given that max duration could be specified in `__init__`, it should not need to be specified here for the first call to `fit()`.

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+        # Train Dataloader
+        if train_dataloader is not None:
+            self._train_data_spec = ensure_data_spec(train_dataloader)
+            self.state.set_dataloader(self._train_data_spec.dataloader, train_dataloader_label)
+        if self._train_data_spec is None:
+            _raise_missing_argument_exception("train_dataloader")
+        if train_subset_num_batches is not None:
+            self.state.dataloader_len = train_subset_num_batches
+        if compute_training_metrics is not None:
+            self.train_metrics = _get_training_metrics(self._original_model) if compute_training_metrics else None
```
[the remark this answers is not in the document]
ravi-mosaicml: `None` implies a different meaning here than `False`

`None`: Do not change whatever was passed in on `__init__`
`False`: Do not compute training metrics

Both would evaluate to False with the implicit `if`

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+            if step_schedulers_every_batch is not None:
+                warnings.warn("Specifying `step_schedulers_every_batch` without `schedulers` has no effect.")
+
+        if len(ensure_tuple(schedulers)) == 0:
+            warnings.warn(f"NoSchedulerWarning: No schedulers were specified. The learning rate will be constant.")
+
+        # Evaluators
+        if eval_dataloader is not None:
+            self.evaluators = [
+                ensure_evaluator(evaluator, self._original_model.metrics(train=False))
```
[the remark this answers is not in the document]
ravi-mosaicml: I'll add a comment re: original model

I think the model argument will disappear with the metrics API refactor, so would prefer to leave this for now.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: See https://github.com/mosaicml/composer/pull/948#discussion_r864189610

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: The new max_duration may be in different units than the existing max_duration, so it might not necessarily work to add them together.

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+            _set_evaluator_interval_and_subset_num_batches(
+                evaluators=self.evaluators,
+                eval_interval=eval_interval,
+                subset_num_batches=eval_subset_num_batches,
+            )
+            if len(self.evaluators) == 0:
+                if eval_subset_num_batches != -1:
+                    warnings.warn("Specifying `eval_subset_num_batches` without an `eval_dataloader` has no effect.")
+                if eval_interval != 1:
+                    warnings.warn("Specifying `eval_interval` without an `eval_dataloader` has no effect.")
```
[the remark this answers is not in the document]
ravi-mosaicml: Converted to `ValueError`s

composer/trainer/trainer.py
```diff
@@ -1022,9 +1100,157 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+                                                                                    step_schedulers_every_batch)
+        else:
+            if scale_schedule_ratio != 1.0:
+                warnings.warn("Specifying `scale_schedule_ratio` without `schedulers` has no effect.")
+
+            if step_schedulers_every_batch is not None:
+                warnings.warn("Specifying `step_schedulers_every_batch` without `schedulers` has no effect.")
+
+            if step_schedulers_every_batch is not None:
+                warnings.warn("Specifying `step_schedulers_every_batch` without `schedulers` has no effect.")
```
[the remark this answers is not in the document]
ravi-mosaicml: Converted to `ValueError`s

composer/trainer/trainer.py
ravi-mosaicml: Added this property back.

composer/trainer/trainer.py
ravi-mosaicml: Fixed both!

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Refactored as described above. Added in additional tests.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: (Resolving -- more comments were added :) )

composer/trainer/trainer.py
ravi-mosaicml: Refactored into two helper functions -- `_is_adaptive_grad_accum` and `_get_initial_grad_accum`

composer/trainer/trainer.py
```diff
@@ -108,40 +47,293 @@
...
+        return None
+    max_duration = cast(Time[int], orig_max_duration * scale_schedule_ratio)
+    log.info(f'max_duration changed from {orig_max_duration} to {max_duration}')
+    if max_duration.value == 0:
+        raise ValueError('Scale schedule has reduced the max_duration to 0. Set a higher ratio or use more epochs.')
+    return max_duration
+
+
+def _should_step_schedulers_every_batch(
+    schedulers: Union[None, ComposerScheduler, PyTorchScheduler, Sequence[Union[ComposerScheduler, PyTorchScheduler]]],
```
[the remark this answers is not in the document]
ravi-mosaicml: We'll lose some of the log statements, but agree that is worth it for the readability.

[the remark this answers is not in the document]
ravi-mosaicml: Refactored this function.

composer/trainer/trainer.py
ravi-mosaicml: Added a setter for the `eval_interval` to the evaluator classes.

composer/trainer/trainer.py
```diff
@@ -106,41 +44,281 @@
...
+    if orig_max_duration is None:
+        return None
+    max_duration = cast(Time[int], orig_max_duration * scale_schedule_ratio)
+    log.info(f'max_duration changed from {orig_max_duration} to {max_duration}')
+    if max_duration.value == 0:
+        raise ValueError('Scale schedule has reduced the max_duration to 0. Set a higher ratio or use more epochs.')
+    return max_duration
+
+
+def _should_step_schedulers_every_batch(schedulers: Optional[Union[Scheduler, Sequence[Scheduler]]]):
```
hanlint, line 71:
```suggestion
def _get_default_scheduler_frequency(schedulers: Optional[Union[Scheduler, Sequence[Scheduler]]]):
```

composer/trainer/trainer.py
hanlint: Agreed in keep this together in one file.

composer/trainer/trainer.py
```diff
@@ -1018,9 +1092,182 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+
+        # Grad Clipping
+        grad_clip_norm: Optional[float] = None,
+    ):
+        """Train the model.
+
+        .. note::
+
+            All arguments to :meth:`.fit` are optional. Any values specified here will override
+            what was provided when constructing the :class:`.Trainer`.
```
hanlint, line 1131: This should not be in a note, but probably in the main body of the docstring.

Should also mention that we support the multiple calls to fit, perhaps with an example. this should also demonstrate how this interacts with the `max_duration` argument, as that can be confusing to folks.

composer/trainer/trainer.py
hanlint: should we be using it here instead of checking the model, like before, to demonstrate the recommended path to checking if deepspeed is being used?

composer/trainer/trainer.py
ravi-mosaicml: (This is really minor as it is a private helper) but since this function returns a boolean, I feel like `_get_default_scheduler_frequency` is ambiguous -- e.g. what does it mean when frequency is "True"? Having trouble thinking of a name that is shorter while also conveying that the returned boolean value means to step every batch.

hanlint: That's a good point. I don't have a good alternative either, maybe make a comment in the code return statement
```
# True means to step schedulers every batch, otherwise every epoch
return not pytorch_schedulers
```

or be explicit:
```
return TimeUnit.EPOCH if pytorch_schedulers else TimeUnit.BATCH
```

and then in the code do `step_schedulers_every_batch = _get... == TimeUnit.BATCH`

The latter seems a bit verbose, but it does leave flexibility down the road if we want to make this bool into `Time` interval.

ravi-mosaicml: I like returning a TimeUnit -- makes it much cleaner.

ravi-mosaicml: Resolved

composer/trainer/trainer.py
ravi-mosaicml: Done

hanlint (approved):
LGTM! As a big PR, might be good to also have @eracah or @A-Jacobson take a quick pass on the significant portions (e.g. `trainer.py`).

eracah:
> LGTM! As a big PR, might be good to also have @eracah or @A-Jacobson take a quick pass on the significant portions (e.g. `trainer.py`).

Ok I'll take a look today.

eracah:
Sorry, this PR is a bit more work than I realized to really look at thoroughly (especially for someone who is still learning the innards of Trainer). I probably won't finish giving it a look through until sometime Monday. Sorry about that. No worries if you want to merge it in before then.

ravi-mosaicml:
> Sorry, this PR is a bit more work than I realized to really look at thoroughly (especially for someone who is still learning the innards of Trainer). I probably won't finish giving it a look through until sometime Monday. Sorry about that. No worries if you want to merge it in before then.

No rush on merging it in; can wait till Monday!

composer/trainer/trainer.py
```diff
@@ -1020,9 +1099,248 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+            _raise_missing_argument_exception("max_duration")
+
+        if self.state.max_duration <= self.state.timestamp.get(self.state.max_duration.unit) and not reset_time:
+            raise ValueError(
+                (f"The max_duration ({self.state.max_duration}) is less than or equal to the elapsed training duration "
+                 f"({self.state.timestamp.get(self.state.max_duration.unit)}). No training would occur. "
+                 "Please provide the `duration` or specify `reset_time=True` in Trainer.fit()."))
+
+        # Scale Schedule Ratio and Schedulers
+        if scale_schedule_ratio != 1.0:
```
eracah, line 1278: It might be simpler to call this function even if ssr is 1.0 because it just multiplies the value by 1.0.

composer/trainer/trainer.py
```diff
@@ -1020,9 +1099,248 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+            schedulers (PyTorchScheduler | ComposerScheduler | Sequence[PyTorchScheduler | ComposerScheduler], optional):
+                See :class:`.Trainer`.
+            scale_schedule_ratio (float, optional): See :class:`.Trainer`.
+            step_schedulers_every_batch (bool, optional): See :class:`.Trainer`.
+            eval_dataloader (Iterable | DataSpec | Evaluator | Sequence[Evaluator], optional): See :class:`.Trainer`.
+            eval_subset_num_batches (int, optional): See :class:`.Trainer`.
+            eval_interval (int | str | Time | (State, Event) -> bool, optional): See :class:`.Trainer`.
+            grad_accum (int | str, optional): See :class:`.Trainer`.
+            precision (Precision | str, optional): See :class:`.Trainer`.
+            grad_clip_norm (float, optional): See :class:`.Trainer`.
```
eracah, line 1242: Maybe add a comment here that setting this value in fit if you are using deepspeed is not allowed?

composer/trainer/trainer.py
```diff
@@ -1020,9 +1099,248 @@ def saved_checkpoints(self) -> List[Tuple[Timestamp, List[pathlib.Path]]]:
...
+
+        # Grad Accum
+        if grad_accum is not None:
+            self.adaptive_gradient_accumulation = _is_adaptive_grad_accum(grad_accum, device=self._device)
+            self.state.grad_accum = _get_initial_grad_accum(grad_accum)
+
+        # Grad Clip Norm
+        if grad_clip_norm is not None:
+            if self.deepspeed_enabled:
+                raise ValueError("Changing the grad_clip_norm when using DeepSpeed is not supported.")
```
eracah, line 1333: Maybe this should just be a warning?

eracah (approved):
Looks pretty good, but I wasn't really able to look at everything closely since this PR is so massive. Maybe would have been better to split it up into some smaller PR's, especially the refactoring into functions parts. I mostly looked at `trainer.py`. Might be worth getting one more set of eyes on this one because I was a bit overwhelmed with this one.

composer/trainer/trainer.py
ravi-mosaicml: We don't attempt to scale the scheduler if the ratio is 1.0, in case if the scheduler cannot be scaled (e.g. the user passes in a custom PyTorch LambdaLR). I'll add a comment here to explain.

composer/trainer/trainer.py
ravi-mosaicml: Good point; added a comment.

composer/trainer/trainer.py
ravi-mosaicml: If the user specifies this value (i.e. not None), and we aren't able to apply it, then I would prefer to keep this an error, since it is easy to miss over a warning.

composer/trainer/trainer.py
ravi-mosaicml: Added the comment.

ravi-mosaicml:
> Looks pretty good, but I wasn't really able to look at everything closely since this PR is so massive. Maybe would have been better to split it up into some smaller PR's, especially the refactoring into functions parts. I mostly looked at `trainer.py`. Might be worth getting one more set of eyes on this one because I was a bit overwhelmed with this one.

Thanks for taking a look -- I know it's quite large. Since @hanlint already looked and other PRs are starting to stack up on this one (#966, #1020, additional logging refactoring), going to merge this in. Will fix any issues as they come up during regression testing.

composer/trainer/trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Will address in the metrics refactor.

PR #949, opened 2022-04-22
Update method cards

nqn:
Before:

Screen Shot 2022-04-22 at 12 51 17 PM

After:

Screen Shot 2022-04-22 at 12 51 12 PM


docs/source/method_cards/methods_overview.rst
```diff
@@ -18,61 +17,54 @@
    .. link-button:: augmix.html
       :text: AugMix
       :classes: card-title
-
    :badge:`CV,badge-primary`
    ++++++++++++++
-   Image-preserving data augmentations
+   Image-perserving data augmentations
```
hanlint, line 22: should be preserving?

hanlint (approved):

hanlint:
Although looks like `sphinx` doesn't like the removal of blank lines?

docs/source/method_cards/methods_overview.rst
nqn: Ah - it's grabbed from the docstring, I think, so will update that shortly

PR #950, opened 2022-04-22
MosaicML Streaming Dataset (Part 1)

abhi-mosaic:
# Overview
This is a new dataset class: `StreamingDataset(torch.utils.data.IterableDatset)` and associated dataset format: `shard-[00x].mds` that has much better performance, shuffling, and usability than existing solutions (e.g. `Webdataset`)

A brief list of improvements:

  - No requirement of `n_samples % n_shards == 0`: Sharded datasets are **complete** with no dropped samples
  - No requirement of `n_shards % n_cpu_workers == 0`: Supports reading from any # of devices, with any # of CPU workers
  - Dataset is downloaded only ~once, regardless of # nodes and # devices and #  CPU workers, no duplicate downloads and egress fees
  - Dataset is cached on local storage after epoch 1
  - When used with a `DataLoader`, the epoch boundaries are consistent (# samples, # batches) regardless of `num_workers`, producing same behavior as a map-style `torch.utils.data.Dataset`
  - Shuffling is best-effort in epoch 1, and samples are made available for random acess as they are being downloaded
  - Shuffling is perfect, i.e. random access (per-worker), in all subsequent epochs
     - will be improved to random access (per-node) in follow up PR

### Dataset format
* Samples are provided as an iterable of dictionaries `Dict[str, bytes]`, such as `[{"img": img_bytes, "class": class_bytes}, ...]`  
* `StreamingDatasetWriter` takes the list of samples and writes out binary `.mds` shards along with an index file `index.mds` that provides shard and byte offsets to every single sample.
* When read using `StreamingDataset`, samples are read in as the raw bytes dictionaries `Dict[str, bytes]`, and users can provide custom decoders to convert bytes to objects. Users can also override `__getitem__` to do more complicated data processing.

### Dataset behavior
* In the context of multi-gpu, multi-worker data loading, a `DataLoader` is created on each device, and each `DataLoader` creates `num_workers` processes, each with an independent copy of `StreamingDataset`.
* On `__iter__`, each worker's `StreamingDataset` queries `torch.dist` info and computes its sample indices, which will be a contiguous range of indices that spans 1 or more shards. Each worker is responsible for a disjoint subset of samples.
  * When the `batch_size` hint is provided, the worker indices are constructed carefully such that there will only be 1 incomplete batch at the end of the epoch _per dataloader_, rather than _per worker_, which is what usually happens with `IterableDatasets`. 
* A background thread is created per-worker to start downloading that worker's shards.
* As shards are downloaded, the worker's available samples grow, and as samples are requested from that worker, they are provided in-order (shuffle=False) or the worker selects randomly from its available samples (shuffle=True)
* Once all shards are downloaded (as in epoch 2+), shuffling is random-access within each worker's samples. 

## Limitations
* Currently we only have download handlers for local dataset paths and S3 URLs, will add more soon
* No cross-worker communication is happening yet, required for node-level shuffling, will be implemented in a follow-up PR


hanlint (commented):
Generally for docstrings we will need to move all the docstrings from `__init__` to the main class.

I did request more details on how to use, supported formats, etc. Essentially putting in some information here from the excellent design documentation. I recommend we follow the torch `DataLoader` example here and put them as a module level in `composer.datasets.streaming`.

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+            sleep(0.25)
+
+    def _iter_ids(self) -> Iterator[int]:
+        """Get an iterator over all our sample IDs.
+
+        Returns:
+            Iterator[int]: Each sample ID.
+        """
+        if self._lock is None:
+            self._lock = Lock()
```
ravi-mosaicml, line 283: is a `threading.Lock` sufficient when using multiple dataloader workers (which would be subprocesses?)

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+        basename = get_shard_basename(shard)
+        shard_filename = os.path.join(self.local, basename)
+        fp = open(shard_filename, 'rb', 0)
+        fp.seek(offset)
+        data = fp.read(size)
+        fp.close()
+
+        return self._unpack_sample(data)
+
+    def _new_growing_epoch(self) -> int:
```
ravi-mosaicml, line 240: Nit: Can the function be named to be more descriptive? (E.g. `_make_new_growing_epoch`?)

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+            epoch (int): The epoch, an identifier for this sequence of samples.
+
+        Returns:
+            int: ID of next sample.
+        """
+        while True:
+            with self._lock:
+                todo_ids = self._epoch_to_todo_ids[epoch]
+                if todo_ids:
+                    return todo_ids.pop(0)
```
ravi-mosaicml, line 268: Can we pop a random element if shuffling (rather than always the 0th)? (If popping a random element, can we do it in `O(1)` by moving the last element into the position of the popped element?)

In addition, if `shuffle=False`, could we use a deque instead of a list (to avoid the `O(n^2)`)?

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+                elif self._are_all_shards_downloaded:
+                    del self._epoch_to_todo_ids[epoch]
+                    return None
+                else:
+                    pass
+            sleep(0.25)
```
ravi-mosaicml, line 274: nit: Can this sleep be replaced by a `threading.Event` that the downloader thread sets when new data is available (and this getter thread can reset it if it exhausts the available data)?

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+    def _iter_ids(self) -> Iterator[int]:
+        """Get an iterator over all our sample IDs.
+
+        Returns:
+            Iterator[int]: Each sample ID.
+        """
+        if self._lock is None:
+            self._lock = Lock()
+
+        self._load()
```
ravi-mosaicml, line 285: Is this safe if `__iter__` is called twice before the dataset finishes downloading (e.g. really fast epoch time or in tests?)

ravi-mosaicml: In addition, can this line (and the lock creation above it) be moved into `__iter__`? Seems like a significant side effect to be placed in the generator.

composer/datasets/streaming/dataset.py
ravi-mosaicml, line 283: Can you add a comment explaining why the lock is created here rather than on `__init__` (I presume this is to be safe with multiple dataloader workers?)

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+    def _new_growing_epoch(self) -> int:
+        """Start a new growing epoch, in which we own the sample sequence because it grows.
+
+        Returns:
+            int: The epoch ID, an identifier which is given back to the caller.
+        """
+        with self._lock:
+            epoch = self._next_epoch
+            self._next_epoch += 1
+            self._epoch_to_todo_ids[epoch] = list(self._downloaded_ids)
```
ravi-mosaicml, line 249: Do we need to track epoch numbers, or can we reset the current buffer of TODO ids? E.g. at the beginning of each epoch / iteration, we set `self._todo_ids = list(self._download_ids)`. I think it is OK to assume that there is only one active iterator per dataset at any given time (we don't iterate over a dataset while already iterating over it).

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+        index_filename = self._download_if_missing(get_index_basename())
+        with open(index_filename, 'rb') as fp:
+            self.index = StreamingDatasetIndex.load(fp)
+
+        # Fields, protected by the lock, relating to loading shards in the background.
+        self._lock = None
+        self._next_epoch = 0
+        self._epoch_to_todo_ids = {}
+        self._downloaded_ids = []
+        self._are_all_shards_downloaded = False
```
ravi-mosaicml, line 63: Do we need to track this? To simplify the codepaths, can we assume that we are never finished downloading? (In reality, once we have all samples downloaded, then the functionality should be identical?)

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+
+        Returns:
+            str: Local cache filename.
+        """
+        remote = os.path.join(self.remote, basename)
+        local = os.path.join(self.local, basename)
+        safe_download(remote, local)
+        return local
+
+    def _load_shards(self, shards: Sequence[int], part_min_id: int, part_max_id: int) -> None:
```
ravi-mosaicml, line 79: Does this function need to track shard IDs, or can it track just the sample IDs (e.g. the signature would be something like `mark_samples_as_available(self, sample_ids: np.ndarray)`?) This would be called once per shard as it is downloaded (or if it is already downloaded)?

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+        # From the partition, derive our shard overlap range and exact sample range.
+        world = World.from_env()
+        todo_shards, part_min_id, part_max_id = self.index.get_partition(world, self.batch_size)
+
+        # Preload all of our shards that are already available in the cache.
+        todo_shards = self._load_shards_if_downloaded(todo_shards, part_min_id, part_max_id)
+
+        # Start downloading our missing shards in a background thread, if there are any.
+        if todo_shards:
+            thread = Thread(target=self._download_thread, args=(todo_shards, part_min_id, part_max_id), daemon=True)
```
ravi-mosaicml, line 182: Would it make sense to support multiple downloading threads? (If the shard size is small, I presume we would not be bandwidth bottlenecked with just one thread?)

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+                for todo_ids in self._epoch_to_todo_ids.values():
+                    todo_ids.extend(new_ids)
+                    np.random.shuffle(todo_ids)
+            else:
+                if not self._are_all_shards_downloaded:
+                    self._downloaded_ids.extend(new_ids)
+                for todo_ids in self._epoch_to_todo_ids.values():
+                    todo_ids.extend(new_ids)
+
+    def _load_shards_if_downloaded(self, shards: Sequence[int], part_min_id: int, part_max_id: int) -> List[int]:
```
ravi-mosaicml, line 123: For code readability, would it make sense to rename this function `determine_downloaded_and_missing_shards(self) -> Tuple[downloaded_shards, missing_shards]`, and then have `_load` process over the downloaded shards and missing shards as appropriate?

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        shards = list(shards)
+        if self.shuffle:
+            np.random.shuffle(shards)
+        for shard in shards:
+            basename = get_shard_basename(shard)
+            self._download_if_missing(basename)
+            shards = shard,
+            self._load_shards(shards, part_min_id, part_max_id)
+        self._done_loading()
```
ravi-mosaicml, line 168: Will this thread gracefully handle intermittent network errors? In addition, if it does crash, will it also crash the training process?

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, dirname: str, fields: List[str], shard_size_limit: int = 1 << 24) -> None:
+        """Initialize with the given output dirname.
+
+        Args:
+            dirname (str): Directory to write shards to.
+            fields: (list of str): The fields to save for each sample.
+            shard_size_limit (int, optional): Maximum shard size in bytes. Default: 1 << 24.
+        """
+        assert len(fields) == len(set(fields))
```
ravi-mosaicml, line 21: Can these asserts be ValueErrors?

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+        self.bytes_per_sample = []
+
+        # Data of the shard in progress.
+        self.new_samples = []
+        self.new_shard_size = 0
+
+    def _flush_shard(self) -> None:
+        """Flush cached samples to a new dataset shard."""
+        if not self.samples_per_shard:
+            os.makedirs(self.dirname)
```
ravi-mosaicml, line 40: `exist_ok`?

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,101 @@
...
+            for data in self.new_samples:
+                out.write(data)
+        self.samples_per_shard.append(len(self.new_samples))
+        self.bytes_per_shard.append(self.new_shard_size)
+        self.new_samples = []
+        self.new_shard_size = 0
+
+    def _write_index(self) -> None:
+        """Save dataset index file."""
+        assert not self.new_samples
```
ravi-mosaicml, line 54: Can this be a RuntimeError? (in case if one does not use the context manager and forgets to call finish)?

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,141 @@
...
+    # Build StreamingDataset
+    dataset = StreamingDataset(remote=remote, local=local, shuffle=shuffle, decoders=decoders, batch_size=batch_size)
+
+    # Build DataLoader
+    dataloader = DataLoader(dataset=dataset, batch_size=batch_size, num_workers=num_workers, drop_last=drop_last)
+
+    # Test dataloader length
+    expected_num_batches = (num_samples // batch_size) if drop_last else math.ceil(num_samples / batch_size)
+    num_batches = len(dataloader)
+    assert num_batches == expected_num_batches, f"Got dataloader length={num_batches} batches, expected {expected_num_batches}"
```
ravi-mosaicml, line 141: Can we iterate over the data loader? That's where all the magic happens :)
Would be nice to track which samples appear and assert that each sample appears once and only once per epoch.

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,141 @@
...
+        sample = dataset[17]
+    except Exception as e:
+        assert False, f"Unable to get random sample, got exception: {e}"
+
+
+@pytest.mark.parametrize("batch_size", [1, 2, 5, 128])
+@pytest.mark.parametrize("drop_last", [False, True])
+@pytest.mark.parametrize("num_workers", [0, 1, 8])
+@pytest.mark.parametrize("share_remote_local", [False, True])
+@pytest.mark.parametrize("shuffle", [False, True])
```
ravi-mosaicml, line 120: Can we also parameterize with `peristent_workers`?

tests/test_dataset_registry.py
```diff
@@ -6,8 +6,8 @@
 
 from composer.datasets import (ADE20kDatasetHparams, BratsDatasetHparams, C4DatasetHparams, CIFAR10DatasetHparams,
                                COCODatasetHparams, DataLoaderHparams, DatasetHparams, GLUEHparams,
-                               ImagenetDatasetHparams, LMDatasetHparams, MNISTDatasetHparams, SyntheticHparamsMixin,
-                               WebDatasetHparams)
+                               ImagenetDatasetHparams, LMDatasetHparams, MNISTDatasetHparams, StreamingDatasetHparams,
```
ravi-mosaicml, line 9: Where is `StreamingDatasetHparams` defined?

Should it be a mixin (similar to SyntheticHparamsMixin) rather than a dataset type?

tests/test_dataset_registry.py
```diff
@@ -72,6 +72,8 @@ def test_dataset(dataset_name: str, dummy_dataloader_hparams: DataLoaderHparams)
     hparams_cls = dataset_registry[dataset_name]
     if issubclass(hparams_cls, WebDatasetHparams):
         return pytest.skip("Skipping test for webdatasets")
+    elif issubclass(hparams_cls, StreamingDatasetHparams):
```
ravi-mosaicml, line 75: Curious why is this being skipped?

ravi-mosaicml (commented):
Wow -- there's a lot here! Overall good work.

In addition to @hanlint's comments to increase the documentation, my high level comments are:
1. Can we store sample sizes in each dataset shard file rather than in the index file? Sorry I did not bring this up in the design review.
2. The StreamingDatasetReader class could probably use another pass for a refactor readability. Is there logic that can be simplified? Can functions be broken up or renamed as appropriate? I tried to comment on what popped out, but did have some trouble following all the logic as I jumped around.
3. Can we unify the file downloading logic into `composer.utils.file_helpers.get_file`? I am not a huge fan of having two different file download helpers in the codebase -- one for streaming datasets, and a different one for retrieving checkpoints.
4. The test cases should attempt to spin the dataloader.

I will take another (docs) pass later, once the docstring comments are addressed. I did not look at the docs build.

knighton:
> Can we store sample sizes in each dataset shard file rather than in the index file? Sorry I did not bring this up in the design review.

Nope, we need all the sample offsets up front (in the index), from which we calculate where to find them/which shards to pull.

ravi-mosaicml:
> > Can we store sample sizes in each dataset shard file rather than in the index file? Sorry I did not bring this up in the design review.
> 
> Nope, we need all the sample offsets up front (in the index), from which we calculate where to find them/which shards to pull.

I understand that the index needs to know the start/stop sample IDs for each shard, so it knows which shard any given index belongs to. But I'm still confused on why the index needs to know the size of each sample?

knighton:
> > > Can we store sample sizes in each dataset shard file rather than in the index file? Sorry I did not bring this up in the design review.
> >
> > Nope, we need all the sample offsets up front (in the index), from which we calculate where to find them/which shards to pull.
>
> I understand that the index needs to know the start/stop sample IDs for each shard, so it knows which shard any given index belongs to. But I'm still confused on why the index needs to know the size of each sample?

Philosophically shards are units of download of the dataset, but they should ideally live reuinted as one file on localdisk with one always-open file handle for performance reasons (punted to future PR, but benchmarked speedup on cifar10, runs out of file handles on imagenet).

True, we could save 8MB (for Imagenet) and lazily compute entire-dataset sample offsets by storing sample offsets at the head of each shard file, it would just be more complicated and you'd lose "just works" concatenation.

tests/test_dataset_registry.py
abhi-mosaic: I think I was planning on removing `StreamingDatasetHparams`, I may have just forgotten to update this test.

tests/datasets/test_streaming.py
abhi-mosaic: Yup, the issue is torch on my Mac OSX does not support pickling of certain objects (but it works on remote Linux machines), so I got a bit stuck in the test-writing, I am planning have to work on a remote machine to finish up the DataLoader tests.

ravi-mosaicml:
> > > > Can we store sample sizes in each dataset shard file rather than in the index file? Sorry I did not bring this up in the design review.
> > > 
> > > 
> > > Nope, we need all the sample offsets up front (in the index), from which we calculate where to find them/which shards to pull.
> > 
> > 
> > I understand that the index needs to know the start/stop sample IDs for each shard, so it knows which shard any given index belongs to. But I'm still confused on why the index needs to know the size of each sample?
> 
> Philosophically shards are units of download of the dataset, but they should ideally live reuinted as one file on localdisk with one always-open file handle for performance reasons (punted to future PR, but benchmarked speedup on cifar10, runs out of file handles on imagenet).
> 
> True, we could save 8MB (for Imagenet) and lazily compute entire-dataset sample offsets by storing sample offsets at the head of each shard file, it would just be more complicated and you'd lose "just works" concatenation.

Ah OK, this makes more sense. Though in practicality, with this optimization, each worker would have its own file handle, and this handle would point to approximately a (contiguous) `1 / (dist.get_world_size() * dataloader.num_workers)` portion of the data. To preserve the concatenation ability (so each worker will have just one file handle down the line), could each shard have its own index file?

Back-of-the-envelope calculation is 2^30 (bytes per GB) / 2^3 (bytes per int64) = 134,217,728 samples per 1 GB of index file. I think this may be a fine assumption for computer vision (given how "large" each sample is), but am skeptical whether this assumption holds for NLP (e.g. for a 100B token NLP model, that is around 781 M samples, which is ~6GB of index file each node would have to download and process before training can begin).

If these numbers are OK until we get version 2 of streaming datasets out, then I'll reluctantly go along with keeping everything in one massive index file. But if it's not too much additional work to store the sizing information so it can be shard specific, then I would prefer that, it is inherently much more scalable for whatever dataset sizes users might have.

abhi-mosaic:
> If these numbers are OK until we get version 2 of streaming datasets out, then I'll reluctantly go along with keeping everything in one massive index file. But if it's not too much additional work to store the sizing information so it can be shard specific, then I would prefer that, it is inherently much more scalable for whatever dataset sizes users might have.

I think I'd like to postpone until version 2 or 3. I think even with 100B tokens and a 6GB index file, we are looking at ~10s of download time for the index file, per-node. And 100B tokens would be a multi-hour or multi-day training run :) Can we revisit this in a future PR where we add NLP examples? I think at that time we'll get a better sense of how bad the issue is.

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,98 @@
...
+    """Download a file from remote to local.
+
+    Args:
+        remote (str): Remote path (local filesystem).
+        local (str): Local path (local filesystem).
+    """
+    shutil.copy(remote, local)
+
+
+def download(remote: str, local: str) -> None:
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+
+    def __init__(self, samples_per_shard: Union[NDArray, Sequence[int]], bytes_per_shard: Union[NDArray, Sequence[int]],
+                 bytes_per_sample: Union[NDArray, Sequence[int]], fields: Sequence[str]) -> None:
+        """Initialize with sample statistics.
+
+        Args:
+            samples_per_shard (NDArray): Number of samples of each shard.
+            bytes_per_shard (NDArray): Size in bytes of each shard.
+            bytes_per_sample (NDArray): Size in bytes of each sample across all shards.
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR.

setup.py
```diff
@@ -153,6 +153,10 @@ def package_files(prefix: str, directory: str, extension: str):
     "wurlitzer>=3.0.2,<4",
 ]
 
+extra_deps["streaming"] = [
+    "boto3>=1.21.45",
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR.

I agree that if we get everything to work through libcloud then we can probably remove this dependency.

composer/datasets/streaming/dataset.py
abhi-mosaic: I think the current implementation has no cross-worker communication, but you're right that when we do implement that in V2, we will need a multiprocessing context or something like that.

composer/datasets/streaming/dataset.py
abhi-mosaic: I'll make a quick attempt at improving the runtime while still using a fully random permutation, but if it ends up being hard to read I'm going to defer to a future PR.

Luckily there's something of a 2-case analysis here:
* if the data is being consumed from the iterator quickly, then `todo_ids` is short, and shuffling is fast
* if the data is being consumed from the iterator slowly, then `todo_ids` is long, and shuffling the ids is slow, but probably not an issue since the data is being consumed slowly anyways

composer/datasets/streaming/dataset.py
abhi-mosaic: Will add a test for `__iter__` being called twice when all samples are not downloaded, and will also try to move `Lock()`.

composer/datasets/streaming/dataset.py
abhi-mosaic: Will attempt to refactor, would like to add a note though that in the future we want to support fast resumption, which may/may not require an `epoch` field, depending on how we implement it.

composer/datasets/streaming/dataset.py
abhi-mosaic: Tracking this but not addressing in this PR.

In practice we have many workers per device, and many devices per node, so maybe 32 or 64 download threads per node, which is pretty good about maxing out our bandwidth. Shard size is also defaulted to 16MB, which seems OK, maybe could be bigger? It's a tradeoff of network download efficiency vs. how long you are willing to wait for sample 1.

tests/test_dataset_registry.py
abhi-mosaic: My bad, didn't mean to commit this file. Hparams tests will come with the new examples of `StreamingDataset` in Part 2...

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
...
+
+      - global_device / global_num_devices
+      - device_of_node / devices_per_node
+
+      - global_worker / global_num_workers
+      - worker_of_node / workers_per_node
+      - worker_of_device / workers_per_device
+    """
+
+    def __init__(self, global_worker: int, num_nodes: int, devices_per_node: int, workers_per_device: int) -> None:
```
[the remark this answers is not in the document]
abhi-mosaic: I think there's a tradeoff of consistency vs. clarity here, it's a bit more readable to use fields like `global_device` and `num_global_devices` rather than `rank` and `world_size` which it's the user has to already know refer to devices rather than workers? Could we re-open this when we potentially refactor `world.py` into `dist.py`?

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
...
+        """Initialize from environment and get_worker_info()."""
+        info = get_worker_info()
+        if info:
+            worker_of_device = info.id
+            workers_per_device = info.num_workers
+        else:
+            worker_of_device = 0
+            workers_per_device = 1
+
+        devices_per_node = dist.get_local_world_size()
```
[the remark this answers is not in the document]
abhi-mosaic: Relaxed as much as I can, see 8af8954.

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
+from torch.utils.data import get_worker_info
+
+from composer.utils import dist
+
+
+class World(object):
```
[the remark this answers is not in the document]
abhi-mosaic: I changed `World` to a dataclass as I would be afraid of index-based access to the attributes in a `NamedTuple`, let me know if this is acceptable.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+def get_shard_basename(shard: int) -> str:
+    """Get the basename for a streaming dataset shard.
+
+    Args:
+        shard (int): Which shard.
+
+    Returns:
+        str: Basename of file.
+    """
+    return f'{shard:05}.mds'
```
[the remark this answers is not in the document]
abhi-mosaic: I think this is an issue with shard writing, we don't know the number of shards ahead of time as the samples are arbitrary size and we write out shards with an approx `shard_size_limit` in _byes,_ not samples.

Luckily this will not produce a bug, it will just be slightly inconsistent filenames if the `num_shards` exceeds 10^5 (or whatever value we choose):

```
>>> print (f"{123:05}")
00123
>>> print (f"{123456:05}")
123456
```

I will increase the zero buffer to 6 so that it's up to 100,000 shards... I think that should keep us safe for now. I will track this issue as well so we can implement a better fix.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+        self.total_bytes = sum(bytes_per_shard)
+        self.total_samples = len(bytes_per_sample)
+        self.num_fields = len(fields)
+
+        # Shard -> sample range.
+        self.shard_ends = self.samples_per_shard.cumsum()
+        self.shard_begins = self.shard_ends - self.samples_per_shard  # type: ignore
+
+        # Sample -> shard, byte offset within shard.
+        self.sample_shards, self.sample_shard_offsets = self.locate_samples()
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+        data = fp.read()
+        return cls.loads(data)
+
+    def dumps(self) -> bytes:
+        """Dump a StreamingDatasetIndex to raw bytes.
+
+        Returns:
+            bytes: The serialized form.
+        """
+        header = np.array([self.total_samples, self.total_bytes, self.num_shards, self.num_fields], np.int64)
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR, as we would need to regenerate our existing datasets, if we change the shard format. I think we will follow up on this pretty quickly though.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    def get_partition(self, world: World, batch_size: Optional[int] = None) -> Tuple[Sequence[int], int, int]:
+        """Get the shards and sample range of a given partition of the dataset.
+
+        Args:
+            world (World): Context about workers, devices, and nodes.
+            batch_size (int): Hint the batch size that will be used on each device's DataLoader.
+                              Worker indices will be constructed so that there is at most 1 incomplete batch at the end of each epoch.
+
+        Returns:
+            shards (Sequence[int]): The shards that this partition overlaps.
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this as part of the "index file per shard" refactor, but not addressing in this PR.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    def dump(self, fp: BufferedWriter) -> None:
+        """Dump a StreamingDatasetIndex to the file.
+
+        Args:
+            fp (file): The file to write.
+        """
+        data = self.dumps()
+        fp.write(data)
+
+    def locate_samples(self) -> Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
abhi-mosaic: Making the function private.
Tracking the comment above as part of the "index file per shard" refactor, but not addressing in this PR.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    Args:
+        shard (int): Which shard.
+
+    Returns:
+        str: Basename of file.
+    """
+    return f'{shard:05}.mds'
+
+
+def sample_dict_to_bytes(obj: Dict[str, bytes], keys: Sequence[str]) -> bytes:
```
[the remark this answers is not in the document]
abhi-mosaic: I think this function is used by `writer.py` so not making private.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    values = []
+    for key in keys:
+        value = obj[key]
+        values.append(value)
+    sizes = list(map(len, values))
+    sizes = np.array(sizes, np.int64)
+    return sizes.tobytes() + b''.join(values)
+
+
+def bytes_to_sample_dict(data: bytes, keys: Sequence[str]) -> Dict[str, bytes]:
```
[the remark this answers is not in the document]
abhi-mosaic: I think this function is used by `reader.py` so not making private.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    ends = num_values * np.int64().nbytes + sizes.cumsum()
+    begins = ends - sizes
+    values = []
+    for begin, end in zip(begins, ends):
+        value = data[begin:end]
+        values.append(value)
+    return dict(zip(keys, values))
+
+
+def read_array(fp: BufferedIOBase, count: int, dtype: DTypeLike) -> NDArray:
```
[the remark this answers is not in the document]
abhi-mosaic: Going to defer to @knighton, leaving as-is for now.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Totals.
+        self.num_shards = len(samples_per_shard)
+        self.total_bytes = sum(bytes_per_shard)
+        self.total_samples = len(bytes_per_sample)
+        self.num_fields = len(fields)
+
+        # Shard -> sample range.
+        self.shard_ends = self.samples_per_shard.cumsum()
+        self.shard_begins = self.shard_ends - self.samples_per_shard  # type: ignore
```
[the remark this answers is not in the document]
abhi-mosaic: Fixed

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Args:
+        fp (BufferedIOBase): File handle.
+        count (int): Number of items to read.
+        dtype (np.dtype): Dtype of the items.
+
+    Returns:
+        np.ndarray: The read array.
+    """
+    num_bytes = count * dtype().nbytes  # type: ignore
```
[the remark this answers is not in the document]
abhi-mosaic: Fixed.

composer/datasets/streaming/world.py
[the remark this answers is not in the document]
abhi-mosaic: Changed to NamedTuple after discussion.

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,237 @@
...
+    def loads(cls, data: bytes):
+        """Load a StreamingDatasetIndex from raw bytes.
+
+        Args:
+            data (bytes): The serialized form.
+
+        Returns:
+            cls: The loaded object.
+        """
+        fp = BytesIO(data)
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking this but not addressing in this PR.

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,132 @@
...
+    # There is no tmp file, so attempt to make it.
+    # If this fails, another download thread beat us to it, so wait.
+    # If we run out of time here, we know a download thread was active, so we should error out.
+    local_dir = os.path.dirname(local)
+    os.makedirs(local_dir, exist_ok=True)
+    try:
+        with open(local_tmp, 'xb') as out:
+            out.write(b'')
+    except FileExistsError:
+        wait_for_download(local, timeout)
```
ravi-mosaicml, line 127: If this wait fails, should we automatically retry?

composer/datasets/streaming/format.py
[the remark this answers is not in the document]
ravi-mosaicml: I think this would be pretty simple to fix? Just move the logic here to `load` below, and have `loads` call `load` (rather than the other way around).

composer/datasets/streaming/format.py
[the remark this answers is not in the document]
ravi-mosaicml: Sorry, I meant to say to include the version number on the index file, not the shard file. I forgot to include versioning in the checkpointing, and it's becoming quite messy to handle backwards-compatible checkpointing, so would rather include this in the original design (even if it means converting any existing datasets -- I presume that would be pretty easy to do?)

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,163 @@
...
+    samples = [{"uid": f"{ix:06}".encode("utf-8"), "data": (3 * ix).to_bytes(4, "big")} for ix in range(num_samples)]
+    decoders = {
+        "uid": lambda uid_bytes: uid_bytes.decode("utf-8"),
+        "data": lambda data_bytes: int.from_bytes(data_bytes, "big")
+    }
+    return samples, decoders
+
+
+def write_synthetic_streaming_dataset(samples: List[Dict[str, bytes]], shard_size_limit: int) -> str:
+    tmpdir = tempfile.TemporaryDirectory().name
```
ravi-mosaicml, line 26: For tests, can you use the `tmp_path` fixture that pytest provides? See https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,163 @@
...
+    first_sample_fields = list(samples[0].keys())
+    with StreamingDatasetWriter(dirname=tmpdir, fields=first_sample_fields,
+                                shard_size_limit=shard_size_limit) as writer:
+        writer.write_samples(samples=samples)
+    return tmpdir
+
+
+@pytest.mark.parametrize("num_samples", [100, 10000])
+@pytest.mark.parametrize("shard_size_limit", [1 << 8, 1 << 16, 1 << 24])
```
ravi-mosaicml, line 35: Curious what value these parameterization combinations add? Would it make sense to specify the (num_samples, shard_size_limit) combinations together to ensure that all the edge cases are covered? Not a huge fan of increasing the test size unless if there is specific value that comes out of it...

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,163 @@
...
+        if shuffle:
+            shuffle_matches += (expected_uid == uid)
+        else:
+            assert uid == expected_uid == uid, f"sample ix={ix} has uid={uid}, expected {expected_uid}"
+            assert data == expected_data, f"sample ix={ix} has data={data}, expected {expected_data}"
+
+    # If shuffling, there should be few matches
+    # The probability of k matches in a random permutation is ~1/(e*(k!))
+    if shuffle:
+        assert shuffle_matches < 10
```
ravi-mosaicml, line 87: I presume if shuffling then the sample numbers should not be monotonically increasing? Would it work to validate that invariant directly? This is clever, but a bit hard to follow...

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,163 @@
...
+    shard_size_limit = 1 << 8
+    samples, decoders = get_fake_samples_decoders(num_samples)
+    remote = write_synthetic_streaming_dataset(samples=samples, shard_size_limit=shard_size_limit)
+    local = tempfile.TemporaryDirectory().name
+
+    os.makedirs(local, exist_ok=True)
+    shutil.copy(os.path.join(remote, "index.mds"), os.path.join(local, "index.mds.tmp"))
+    shutil.copy(os.path.join(remote, "000003.mds"), os.path.join(local, "000003.mds.tmp"))
+    time.sleep(created_ago)
+    dataset = StreamingDataset(remote=remote, local=local, shuffle=False, decoders=decoders, timeout=timeout)
```
ravi-mosaicml, line 104: Is it possible to test multiple crashes? (See comment in the function about calling the wait recursively...)

ravi-mosaicml (commented):
Took another pass through everything except `dataset.py`. See comments re docstrings. Excited to see the dataloader tests as well.

composer/datasets/streaming/dataset.py
abhi-mosaic: Going to postpone this as V2 of StreamingDataset with global shuffle will refactor this.

composer/datasets/streaming/dataset.py
abhi-mosaic: Tracking this but not addressing in this PR.

composer/datasets/streaming/dataset.py
abhi-mosaic: Tracking this but not addressing in this PR.

composer/datasets/streaming/dataset.py
abhi-mosaic: Tracking but not addressing in this PR.

composer/datasets/streaming/dataset.py
abhi-mosaic: Agreed we should probably accept `sample_ids` rather than shards. Will refactor in follow up PR soon.

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def download_from_local(remote: str, local: str) -> None:
+    """Download a file from remote to local.
+
+    Args:
+        remote (str): Remote path (local filesystem).
+        local (str): Local path (local filesystem).
+    """
+    shutil.copy(remote, local)
```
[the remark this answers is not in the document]
abhi-mosaic: I think we actually might want to copy... the situation I'm thinking of is remote is an NFS mount, and local is localdisk, in which case you do want to copy so that on epoch 2, you are reading from localdisk. Does that make sense?

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,77 @@
...
+
+    node_worker: int
+    node_num_workers: int
+
+    device_worker: int
+    device_num_workers: int
+
+
+def get_world() -> World:
+    """Returns a World object, initialized using composer.dist and torch.utils.data.get_worker_info()."""
```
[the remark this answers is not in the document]
abhi-mosaic: Yup, will do a big docs readthrough at the end, thanks for the example of sphinx/rst, I'm new to it :)

tests/datasets/test_streaming.py
abhi-mosaic: I think the non-monotonicity is a weaker check than this one (a reverse order or other poor shuffles would pass it. The `shuffle_matches` is based on a bit of fancy math: the expected number of permutation matches is independent of N (for large enough N) and `1/(e * 10!)` is about 1 in 10million I think it will be pretty safe?

@anisehsani has been working on improved shuffle tests that we may add in afterwards.

tests/datasets/test_streaming.py
abhi-mosaic: These were just a few random values, I think in practice this test takes <1s? Unfortunately I don't really know how to construct good edge cases without perfetctly controlling the bytes per shard, which is pretty hard to do given headers and all... can we leave as is?

composer/datasets/streaming/format.py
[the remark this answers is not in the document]
abhi-mosaic: Done.

composer/datasets/streaming/dataset.py
abhi-mosaic: Moved the lock creation to `__iter__`

composer/datasets/streaming/format.py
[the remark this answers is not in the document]
abhi-mosaic: @knighton ^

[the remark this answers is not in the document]
knighton: In this PR or next? Not that much work to update and regenerate.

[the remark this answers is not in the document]
abhi-mosaic: @knighton if it's possible, could you add the version number to the header in this PR?

abhi-mosaic:
Something fun I discovered... `IterableDataset` will not be deterministic when `persistent_workers: True` when `num_workers > 1` due to PyTorch's DataLoader not resetting the worker cycle.

Luckily, someone else ran into this same bug last month, and it got fixed! Probably will land in PyTorch 1.12? See https://github.com/pytorch/pytorch/pull/73675

For now I set our dataloader tests to `xfail` on those scenarios. We can technically still use `persistent_workers: True` in our research, all the samples will still be received, but the order will not necessarily be deterministic even when `shuffle=True`. The most important thing is: don't use `persistent_workers: True` in any tests that try to repeat an epoch and expect the same order of samples.

tests/datasets/test_streaming.py
abhi-mosaic: Done!

composer/datasets/streaming/format.py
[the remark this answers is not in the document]
abhi-mosaic: After discussion we are postponing this to the next PR, acknowledge the risks of chanigng the shard format later, but we have no plans to support old versions of MDS at this time.

abhi-mosaic:
@hanlint @ravi-mosaicml I think docs should be in an OK place now... would prefer if one of you could take a look through and approve this (if only minor edits left) and I can fix them async and merge tonight. I will be opening Part 2 (ImageNet, ADE20k, COCO) later today branching off from here.

composer/datasets/streaming/dataset.py
```diff
@@ -0,0 +1,422 @@
...
+        Returns:
+            Iterator[Any]: Each sample.
+        """
+        # Lock is created here because DataLoader calls __iter__ in each worker process
+        # and the lock is worker-specific
+        if self._lock is None:
+            self._lock = Lock()
+
+        # Load samples
+        self._load()
```
knighton, line 347: Are we sure load and lock check should go in iter? Was thinking of calling iter_ids directly for testing purposes

abhi-mosaic: This change was motivated from @ravi-mosaicml,  for all our tests we're doing end-to-end dataset build followed by `iter(dataset)` so it should be safer, if anything.

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,315 @@
...
+    dataset = StreamingDataset(remote=remote, local=local, shuffle=False, decoders=decoders)
+
+    # Test retrieving random sample
+    try:
+        _ = dataset[17]
+    except Exception as e:
+        assert False, f"Unable to get random sample, got exception: {e}"
+
+
+# @pytest.mark.daily()
```
abhi-mosaic, line 144: @ravi-mosaicml please let me know if I should uncomment this line, for the single- and multi- gpu DataLoader tests.

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import NamedTuple
+
+from torch.utils.data import get_worker_info
+
+from composer.utils import dist
+
+__all__ = ["World", "get_world"]
+
+
+class World(NamedTuple):
```
ravi-mosaicml, line 15: For readability, can you use the same termonology as `composer.utils.dist` where the names do align?

`global_node` -> `node_rank`
`global_num_nodes` -> `node_world_size`
`global_device` -> `global_rank` (or `rank`)
`global_num_devices` -> `world_size`
`node_device` -> `local_rank`
`node_num_devices` -> `local_world_size`

For the naming of workers, maybe match `torch.utils.data.get_worker_info` (and use `rank` instead of `device`?) Agree that "local" doesn't make sense, as that is ambiguous for node vs rank.

`global_worker` -> `global_worker_id`
`node_worker` -> `node_worker_id`
`device_worker` -> `rank_worker_id`
`device_num_workers` -> `rank_num_workers`

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
...
+    """A :class:`NamedTuple` that provides context about workers, devices, and nodes."""
```
ravi-mosaicml, line 16: Can you add a docstring line under the "Attributes:" section for each field in the namedtuple? See other namedtuples in Composer as examples.

composer/datasets/streaming/world.py
```diff
@@ -0,0 +1,82 @@
...
+
+    node_worker: int
+    node_num_workers: int
+
+    device_worker: int
+    device_num_workers: int
+
+
+def get_world() -> World:
+    """Returns a :class:`World` object, initialized using :mod:`composer.dist` and :meth:`torch.utils.data.get_worker_info`"""
```
ravi-mosaicml, line 37:
```suggestion
    """Returns a :class:`World` object, initialized using :mod:`composer.dist` and :func:`torch.utils.data.get_worker_info`"""
```

ravi-mosaicml, line 37:
```suggestion
    """Returns a :class:`World` object, initialized using :mod:`composer.utils.dist` and :meth:`torch.utils.data.get_worker_info`"""
```

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,286 @@
...
+    Each sample is a dict of str to bytes. All samples must contain the same
+    dict keys (fields). These strings are stored in the index file for
+    efficiency.
+    """
+
+    def __init__(self, samples_per_shard: NDArray[np.int64], bytes_per_shard: NDArray[np.int64],
+                 bytes_per_sample: NDArray[np.int64], fields: List[str]) -> None:
+        """Initialize with sample statistics.
+
+        Args:
```
ravi-mosaicml, line 117: Can you document these ars in the class-level docstring above? (See other classes in Composer as examples)

composer/datasets/streaming/format.py
```diff
@@ -0,0 +1,286 @@
...
+        sample_begins = sample_ends - self.bytes_per_sample
+        sample_shard_offsets = sample_begins - sample_shard_begins
+        return sample_shards, sample_shard_offsets
+
+    def get_partition(self, world: World, batch_size: Optional[int] = None) -> Tuple[List[int], int, int]:
+        """Get the shards and sample range of a given partition of the dataset.
+
+        Args:
+            world (World): Context about workers, devices, and nodes.
+            batch_size (int): Hint the batch_size that will be used on each device's DataLoader.
```
ravi-mosaicml, line 225: I'm confused by what is meant by hint here. My intuition is that providing a hint could make a function run faster, but it should return the same result even if I don't provide the hint. I don't think this is the case?

Perhaps could you rename the variable and provide additional description / an example of what this does?

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,135 @@
...
+
+"""Download handling for :class:`StreamingDataset`.
+"""
+
+import os
+import shutil
+from time import sleep, time
+from urllib.parse import urlparse
+
+
```
ravi-mosaicml, line 11: missing `__all__`

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,150 @@
...
+        """Complete writing the dataset by flushing last samples to a last shard, then write an index file."""
+        if self.new_samples:
+            self._flush_shard()
+        self._write_index()
+
+    def __enter__(self):
+        """Enter as contextmanager."""
+        return self
+
+    def __exit__(self, *args: List, **kwargs: Dict) -> None:
```
ravi-mosaicml, line 148: `from types import TracebackType`
`from typing import Type, Optional`

```suggestion
    def __exit__(self, exc_type: Optional[Type[BaseException]], exc: Optional[BaseException], traceback: Optional[TracebackType]) -> None:
```

composer/datasets/streaming/writer.py
```diff
@@ -0,0 +1,150 @@
...
+            self.write_sample(s)
+
+    def finish(self) -> None:
+        """Complete writing the dataset by flushing last samples to a last shard, then write an index file."""
+        if self.new_samples:
+            self._flush_shard()
+        self._write_index()
+
+    def __enter__(self):
+        """Enter as contextmanager."""
```
ravi-mosaicml, line 145: nit: no docstring needed

setup.py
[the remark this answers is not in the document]
ravi-mosaicml: The dependency will also need to be in meta.yaml -- see the style guide and other conditional imports as examples: https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#61-external-dependencies

tests/datasets/test_streaming.py
ravi-mosaicml: I see. So if the test is to ensure that shards are flushed properly, do we need the parameterization matrix? Or can we accomplish this through one test?

tests/datasets/test_streaming.py
ravi-mosaicml: I elaborated more on this in the download function, but I interpret timeout as a "per attempt" timeout, which I think should be separate from the number of retries.

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,317 @@
...
+            "PyTorch DataLoader has non-deterministic worker cycle iterator when `persistent_workers=True`. Fixed in Mar 2022, likely landing PyTorch 1.12: https://github.com/pytorch/pytorch/pull/73675"
+        )),
+])
+@pytest.mark.parametrize("shuffle", [False, True])
+def test_dataloader_single_device(tmpdir: py.path.local, batch_size: int, drop_last: bool, num_workers: int,
+                                  persistent_workers: bool, shuffle: bool):
+    num_samples = 31
+    shard_size_limit = 1 << 6
+    samples, decoders = get_fake_samples_decoders(num_samples)
+    remote = str(tmpdir.mkdir("remote"))
```
ravi-mosaicml, line 164: Can some of this code be refactored into fixtures? E.g. a dataset fixture? Seems to be similar to the test below.

BTW, this test is awesome -- gives me confidence in what is happening.

tests/datasets/test_streaming.py
ravi-mosaicml: Sure, not disagreeing that the math might be fancy and correct, but it's not intuitive to follow (where does that equation come from?). Mainly don't want flaky tests from this.

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,317 @@
...
+            shuffle_matches += (expected_uid == uid)
+        else:
+            assert uid == expected_uid == uid, f"sample ix={ix} has uid={uid}, expected {expected_uid}"
+            assert data == expected_data, f"sample ix={ix} has data={data}, expected {expected_data}"
+
+    # If shuffling, there should be few matches
+    # The probability of k matches in a random permutation is ~1/(e*(k!))
+    if shuffle:
+        assert shuffle_matches < 10
+
```
ravi-mosaicml, line 88: If shuffle is False, can we assert that the dataset is indeed not shuffled?

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,317 @@
...
+    else:
+        local = str(tmpdir.mkdir("local"))
+
+    # Build StreamingDataset
+    dataset = StreamingDataset(remote=remote, local=local, shuffle=False, decoders=decoders)
+
+    # Test retrieving random sample
+    try:
+        _ = dataset[17]
+    except Exception as e:
```
ravi-mosaicml, line 140: No need to catch the exception -- it's fine to have it propagate (in fact, catching it like this will hide the stack trace).

tests/datasets/test_streaming.py
```diff
@@ -0,0 +1,317 @@
...
+
+@pytest.mark.daily()
+@pytest.mark.timeout(10)
+@pytest.mark.world_size(2)
+@pytest.mark.parametrize("batch_size", [4])
+@pytest.mark.parametrize("drop_last", [False, True])
+@pytest.mark.parametrize("num_samples", [30, 31])
+@pytest.mark.parametrize("num_workers", [1, 3])
+@pytest.mark.parametrize("shuffle", [False, True])
+def test_dataloader_multi_device(tmpdir: py.path.local, batch_size: int, drop_last: bool, num_samples: int,
```
ravi-mosaicml, line 231: Does this test need to be separate from the one above? Or can it be the same (with potentially some checks of `dist.get_global_rank() == 0` for the checks that should be happening only on the rank zero node)?

ravi-mosaicml (commented):
Left some comments -- overall looks pretty close.

For the comments from the first PR that are getting deferred, is there a JIRA where all of this is being tracked? Want to make sure that we won't be losing track of what needs to be done to clean this up for a non-experimental release.

composer/datasets/streaming/world.py
abhi-mosaic: I'm gonna push back pretty hard on this... I agree that consistency is good, but if anything the left side is what we should use everywhere, not the right. The `world_size` and `rank` terminology is a lot less clear than the `global/node/device/worker` hierarchy.

I'll make a JIRA issue to reolve `World` and `composer.utils.dist1, but I would like to keep it as is here.

composer/datasets/streaming/format.py
abhi-mosaic: So providing `batch_size` will only change the behavior if:
 * you pass the object to a DataLoader
 * num_workers > 1
 * 
and I think @hanlint suggested using the same arg name as the DataLoader, so that users can do:

```
batch_size = 32
dataset = StreamingDataset(..., batch_size=batch_size)
loader = DataLoader(..., batch_size=batch_size)
```

The more descriptive variable name would be `device_batch_size_hint`, but IDK how you feel about doing that in the name or just elaborating in the description.

tests/datasets/test_streaming.py
abhi-mosaic: I think the matrix is to make sure that samples are flushed and the correct number of shard are written, which means we want to check a few cases of samples % shard_bytes.

tests/datasets/test_streaming.py
abhi-mosaic: I think the separation is useful because the test matrix is reduced by a lot if I know that single-device dataloader is working... same situation if/when we add smarter multi-node behavior, I'd like to add a `test_dataloader_multi_node`.

I could maybe see a refactor combining the two but it would run a lot more tests than necessary.

tests/datasets/test_streaming.py
abhi-mosaic: This is accomplished within the loop, where every sample is assert'd to be the correct one (in-order)

composer/datasets/streaming/world.py
ravi-mosaicml: PyTorch uses this terminology -- see https://pytorch.org/docs/stable/elastic/run.html#launcher-api
I think we should follow that for anything node/rank related, and then follow the `torch.utils.data.get_worker_info` for the naming of the worker variables.

composer/datasets/streaming/format.py
ravi-mosaicml: Agree it makes sense to keep the variable the same. I guess could you elaborate in the docstring on why this is needed? An example showing the difference of providing it vs not providing it would be very helpful.

tests/datasets/test_streaming.py
ravi-mosaicml: OK, if it would be helpful to keep them seperate, that's fine -- but can the common code be moved to shared fixtures / helper functions?

abhi-mosaic:
> Left some comments -- overall looks pretty close.
> 
> For the comments from the first PR that are getting deferred, is there a JIRA where all of this is being tracked? Want to make sure that we won't be losing track of what needs to be done to clean this up for a non-experimental release.

I have all the comments in Notion but will port to JIRA under the same StreamingDataset epic.

composer/datasets/streaming/format.py
abhi-mosaic: Oh yeah there should be a really big explanation at the `StreamingDataset` docstring but I can copy it here as well.

composer/datasets/streaming/world.py
abhi-mosaic: Not sure if we should follow PyTorch convention for its own sake, if you take a look at `get_partition` the readability is greatly _improved_ with the new hierarchy, and I think we ought to go for readability > consistency.

I think this is a situation similar to batch/minibatch/microbatch, where the current default naming that PyTorch uses is legacy from a time when people were not doing comlex distributed training.

E.g, how many ML users are aware that the DataLoader `batch_size` is really the `device_batch_size`? Is the legacy naming more or less clear than what it could be?

composer/datasets/streaming/download.py
```diff
@@ -0,0 +1,135 @@
...
+            try:
+                wait_for_download(local, timeout)
+                return
+            except TimeoutError:
+                pass
+
+        # The tmp file is old, it is either (1) from a crashed run or (2) another thread is downloading it but is taking too long, and will timeout.
+        # Let's delete the tmp file. If situation (1), this is safe. If situation (2), the other thread is expected to crash with a TimeoutError anyways, so this is fine.
+        try:
+            os.remove(local_tmp)
```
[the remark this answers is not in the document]
abhi-mosaic: Tracking in JIRA.

tests/datasets/test_streaming.py
abhi-mosaic: Tracking this as part of a broader timeout refactor in JIRA.

composer/datasets/streaming/download.py
abhi-mosaic: Totally makes sense. Gonna add this to JIRA as it will require a refactor with multiprocessing locks, that are not currently implemented.

meta.yaml
```diff
@@ -44,6 +44,7 @@ requirements:
     - transformers >=4.11,<5
     - datasets >=1.14,<2
     - pycocotools >=2.0.4,<3
+    - boto3>=1.21.45
```
ravi-mosaicml, line 47: Nit: Sorry, this will re-trigger the 30 min conda build, but max versions are needed (thanks HF for the lesson learned!)

```suggestion
    - boto3 >=1.21.45,<2
```

meta.yaml
```diff
@@ -64,6 +65,7 @@ test:
...
```
ravi-mosaicml, line 68:
```suggestion
    - boto3 >=1.21.45,<2
```

ravi-mosaicml (approved):
LGTM, see comment re: boto3 versioning! :rocket:

meta.yaml
abhi-mosaic: Ah yep! No worries, I remember the hassle this caused during release a few months back. Will fix.

PR #951, opened 2022-04-25
Fix `ComposerModel` copying and pickling

ravi-mosaicml:
-  #944 broke `copy.deepcopy(state.model)` and `pickle.dumps(state.model`), since it bound the Logger to the model, and the Logger cannot be serialized.
- Fixed by manually defining `__copy__`, `__deepcopy__`, and `__getstate__` on the `ComposerModel` base class.
- Added a test case

### DIFF
--- a/composer/models/base.py
+++ b/composer/models/base.py
@@ -60,7 +60,7 @@ class ComposerModel(torch.nn.Module, abc.ABC):
         # directly. Hence, this helper getter, which the @property(logger) uses below, is monkey-patched.
         # This trick preserves the ability to `deepcopy` the model (which is required by some algorithms, such as SWA),
         # as the Logger cannot be deepcopied. (The Logger is attached to the state, and the state has file objects).
-        # This trick works because methods are not copied but are instead re-bound by reference).
+        # This trick works because methods are not copied but are instead re-bound by reference.
         return None
 
     @property
--- a/composer/models/base.py
+++ b/composer/models/base.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import abc
+import copy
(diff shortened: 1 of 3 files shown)


composer/models/base.py
```diff
@@ -49,15 +49,30 @@ def loss(self, outputs, batch):
...
+        # The trainer monkey-patches this getter to set the logger.
+        # Since properties are bound to the class (and not the instance), the `logger` @property cannot be monkey-patched
+        # directly. Hence, this helper getter, which the @property(logger) uses below, is monkey-patched.
+        # This trick preserves the ability to `deepcopy` the model (which is required by some algorithms, such as SWA),
+        # as the Logger cannot be deepcopied. (The Logger is attached to the state, and the state has file objects).
+        # This trick works because methods are not copied but are instead re-bound by reference.
+        return None
+
+    @property
+    def logger(self) -> Optional[Logger]:
```
hanlint, line 67: well retrieving the logger still work after SWA deep copies the model? (which happens after the patch happens)

hanlint: Also, would it be better for the model to implement a `__deepcopy__()` that skips the logger attribute?

e.g. see https://docs.python.org/3/library/copy.html

hanlint (commented):
Suggest implementing `__deepcopy__` and `__copy__` instead of monkeypatching to solve this issue.

composer/models/base.py
ravi-mosaicml: > well retrieving the logger still work after SWA deep copies the model? (which happens after the patch happens)

Yes, added a test case for this.

hanlint (approved):

PR #952, opened 2022-04-25
Fix the lint/doctest tests to capture stdout/stderr in the JUnit results

ravi-mosaicml:
- The Junit results were not capturing the lint / doctest failures, which makes it difficult for external contributors (who do not have Jenkins access) to debug failing lint/doctest builds.
See https://github.com/mosaicml/composer/runs/6133133356 as an example.
- Fixed by adding `capture_output=True` and `text=True` to all `subprocess.run` calls, and wrote a custom `check_output` helper.

### DIFF
--- a/.ci/test_lint_doctests.py
+++ b/.ci/test_lint_doctests.py
@@ -8,10 +8,28 @@ import os
 import pytest
 import shutil
 import pathlib
+import textwrap
+
+def check_output(proc: subprocess.CompletedProcess):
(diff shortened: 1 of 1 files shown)


ravi-mosaicml:
Example junit output is attached

[junit.txt](https://github.com/mosaicml/composer/files/8556645/junit.txt)

hanlint (approved):

PR #953, opened 2022-04-25
Test Cleanup

ravi-mosaicml:
1. Refactored `test/common` into seperate files
2. Moved the `EventCounterCallback` into `tests/common`, since it will be used by non-checkpointing tests
3. Replaced the `SimpleBatchPairModel` with the `SimpleModel` throughout tests
4. Removed duplicated and now-unused test fixtures from `tests/fixtures/dummy_fixtures.py`


tests/test_state.py
```diff
@@ -39,49 +38,12 @@ def get_dummy_state(model: ComposerModel, train_dataloader: DataLoader):
...
-        val2 = dict2[k]
-        _check_item(val1, val2, f"{path}/{k}")
-
-
 def assert_state_equivalent(state1: State, state2: State):
     assert state1.serialized_attributes == state2.serialized_attributes
     assert state1.is_model_deepspeed == state2.is_model_deepspeed
 
-    _check_dict_recursively(state1.state_dict(), state2.state_dict(), "")
+    # Using a loose tolerance as GPU determinism does not work properly
```
hanlint, line 45: Surprised a bit about the need here, where is that coming from?

hanlint (approved):
Great cleanup, thank you!

tests/test_state.py
ravi-mosaicml: On the resent and gpt2 checkpointing tests, I was unable to get an exact comparison to pass.

PR #954, opened 2022-04-25
Remove BatchPair and BatchDict typing annotations and helper functions.

eracah:
[Jira issue](https://mosaicml.atlassian.net/browse/CO-23)


composer/core/types.py
```diff
@@ -33,7 +33,7 @@
     from typing import Protocol
 
 __all__ = [
-    "Batch", "BatchPair", "BatchDict", "PyTorchScheduler", "JSON", "MemoryFormat", "as_batch_dict", "as_batch_pair",
+    "Batch", "BatchPair", "BatchDict", "PyTorchScheduler", "JSON", "MemoryFormat",
```
ravi-mosaicml, line 36: Can you remove "Batch", "BatchPair", and "BatchDict" from `__all__`, and also remove these variables from lines 42-44 below?

ravi-mosaicml (approved):
LGTM; see comment!

composer/core/types.py
eracah: Ok I removed `BatchPair` and `BatchDict` in 
0545b64, but I guess removing `Batch` will be a bit tougher. I guess find-replace `Batch` -> `torch.Tensor`?

hanlint: I think `Batch` would be `Any`? Although this might unleash pyright hell...

ravi-mosaicml: In the ideal world, we could use a `TypeVar` for the batch type. But I think this will introduce more pyright hell than it is worth -- especially for algorithms -- as they would likely be `cast`ing the batch anyways to silence pyright. So agreed; `Any` is the way to go.

A-Jacobson: +1 for Any. I tried using typevar for this and was immediately buried in pyright fire and brimstone.

composer/core/types.py
```diff
@@ -33,15 +33,11 @@
...
+    "Batch", "PyTorchScheduler", "JSON", "MemoryFormat",
     "DataLoader", "BreakEpochException"
 ]
 
-# For BatchPar, if it is a list, then it should always be of length 2.
-# Pytorch's default collate_fn returns a list even when the dataset returns a tuple.
-BatchPair = Sequence[Union[torch.Tensor, Sequence[torch.Tensor]]]
-BatchDict = Dict[str, torch.Tensor]
-Batch = Union[BatchPair, BatchDict, torch.Tensor]
+Batch = torch.Tensor
```
A-Jacobson, line 40:
```suggestion
Batch = Any
```

composer/core/types.py
```diff
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Dict, Iterator, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Dict, Iterator, List, Optional, Union
```
A-Jacobson, line 20:
```suggestion
from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Union
```

A-Jacobson (approved):
batch to `Any` as discussed. other than that LGTM.

composer/core/types.py
eracah: ok fixed in [acf116c](https://github.com/mosaicml/composer/pull/954/commits/acf116c968f0458ddffd28ab9a4ad25b98c7b182)

composer/core/types.py
eracah: done in [acf116c](https://github.com/mosaicml/composer/pull/954/commits/acf116c968f0458ddffd28ab9a4ad25b98c7b182)

composer/core/types.py
eracah: done in [acf116c](https://github.com/mosaicml/composer/pull/954/commits/acf116c968f0458ddffd28ab9a4ad25b98c7b182)

eracah:
Removed a few more mentions of BatchDict and BatchPair in comments and docs here: [28847bf](https://github.com/mosaicml/composer/pull/954/commits/28847bfa1fd10fd56e71e9738e499e610f9815b4)

jzf2101:
@eracah @hanlint I think this reverses the changes from #896 Is this intentional?

PR #955, opened 2022-04-26
Updated ALiBi link in README

ofirpress:
We just uploaded a v2 of the paper and the README linked to v1.


hanlint (approved):
Thank you for the update, and for the great work with Alibi!

PR #957, opened 2022-04-26
Streamline Composer GitHub Issue submission for users

### LINKED ISSUE #673: Streamline Composer GitHub Issue submission for users
## 🚀 Feature Request
Whenever Composer raises an exception let's automatically collect environment information and streamline issue submission by adding a link that the user can click to open a new GitHub issue. 

## Implementation
We can collect environment information using:
`# For security purposes, please check the contents of collect_env.py before running it.`
`wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
python collect_env.py`

Once pertinent information has been collected, we can provide a mechanism to streamline issue submission (customized link, attach error log, etc.).  The user should able to edit the issue before final submission to provide any additional context or remove confidential information.

## Related information
* This should be doable by hooking into sys.excepthook -- see this [StackOverflow post](https://stackoverflow.com/questions/6234405/logging-uncaught-exceptions-in-python) for an example. 
* Jupyter will require special handling -- see this [StackOverflow post](https://stackoverflow.com/questions/1261668/cannot-override-sys-excepthook)

bandish-shah:
Overview:
* Adds the `composer.utils.collect_env` utility module to facilitate gathering of system information for debug purposes.
* Updates the Trainer initialization to register the customized `excepthook`. 
* Resolves #673 

The `collect_env` module provides the following functionality:
* Use Torch's version of `collect_env` to gather system information
* Extends functionality to gather Composer pertinent system information
* Provides a wrapper for `sys.excepthook` to automatically gather system information when an exception is caught
* Provides `configure_excepthook()` helper function to override original `sys.excepthook`
* Provides `enable_env_report()` and `disable_env_report()` helper function to enable and disable the verbose environment report generation, respectively
* Handles IPython exceptions
* Documentation

Sample environment report:

```
        ---------------------------------
        System Environment Report
        Created: 2022-04-27 00:25:33 UTC
        ---------------------------------

        PyTorch information
        -------------------
        PyTorch version: 1-91+cu111
        Is debug build: False
        CUDA used to build PyTorch: 111
        ROCM used to build PyTorch: N/A

        OS: Ubuntu 18.04.6 LTS (x86_64)
        GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
        Clang version: Could not collect
        CMake version: version 3.10.2
        Libc version: glibc-2.27

        Python version: 3.8 (64-bit runtime)
        Python platform: Linux-5.8.0-63-generic-x86_64-with-glibc2.27
        Is CUDA available: True
        CUDA runtime version: 11.1.105
        GPU models and configuration:
        GPU 0: NVIDIA GeForce RTX 3080
        GPU 1: NVIDIA GeForce RTX 3080
        GPU 2: NVIDIA GeForce RTX 3080
        GPU 3: NVIDIA GeForce RTX 3080

        Nvidia driver version: 470.57.02
        cuDNN version: Probably one of the following:
        /usr/lib/x86_64-linux-gnu/libcudnn.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.0.5
        /usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.0.5
        HIP runtime version: N/A
        MIOpen runtime version: N/A

        Versions of relevant libraries:
        [pip3] numpy==1.22.3
        [pip3] pytorch-ranger==0.1.1
        [pip3] torch==1.9.1+cu111
        [pip3] torch-optimizer==0.1.0
        [pip3] torchmetrics==0.7.3
        [pip3] torchvision==0.10.1+cu111
        [pip3] vit-pytorch==0.27.0
        [conda] Could not collect


        Composer information
        --------------------
        Composer version: 0.6.0
        Host processor model name: AMD EPYC 7502 32-Core Processor
        Host processor core count: 64
        Number of nodes: 1
        Accelerator model name: NVIDIA GeForce RTX 3080
        Accelerators per node: 1
        CUDA Device Count: 4
```


composer/trainer/trainer.py
```diff
@@ -77,6 +77,7 @@
 import torch
 import torch.distributed
 import torch.utils.data
+from debugpy import configure
```
ravi-mosaicml, line 80: Is this import used?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+* Accelerator model name
+
+This package can be invoked as a standalone console script or can be invoked from within
+an application to gather and generate a system environment report.
+
+The module can be invoked by using the entrypoint alias:
+
+.. code-block::
+
+    composer_collect_env
```
ravi-mosaicml, line 22: nit: to denote it's a shell script

```suggestion
    $ composer_collect_env
```

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+
+__all__ = ["configure_excepthook", "print_env", "restore_excepthook"]
+
+# Check if PyTorch is installed
+try:
+    import torch.utils.collect_env as torchenv
+    from torch.cuda import get_device_name as accel_device_name
+    from torch.cuda import is_available as cuda_available
+    TORCH_AVAILABLE = True
+except (ImportError, NameError, AttributeError, OSError):
```
ravi-mosaicml, line 55: Would it work to just catch the `ImportError`? Or do we need to catch these other errors?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+# Check if Composer is installed
+try:
+    import composer
+    from composer.utils import dist
+    COMPOSER_AVAILABLE = True
+except (ImportError, NameError, AttributeError, OSError):
+    COMPOSER_AVAILABLE = False
+
+COMPOSER_OPEN_ISSUE_URL = "https://github.com/mosaicml/composer/issues/new/choose"
+EXCEPTION_MSG = f"If you believe this exception was raised due to a Composer bug, " + \
```
ravi-mosaicml, line 67: Nit: "If you believe" is a bit ambiguous.

Generally speaking:

- If it's a `AssertionError, then it's definitely a Composer Bug. A bug report should always be filed.
- If it's a `RuntimeError, then it's likely a Composer bug, or someone using something very incorrectly.
- If it's a `ValueError`, `TypeError`, then it's likely a user bug, but could be a composer bug.
- If it's another exception type, then not sure -- could be anything?

Could we customize the message based off of this?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+                f"file a bug report at: {COMPOSER_OPEN_ISSUE_URL}\n" + \
+                f"Please include the following environment report:\n"
+
+
+# Same convention as Torch collect_env, create a namedtuple to track collected fields
+class ComposerEnv(NamedTuple):
```
ravi-mosaicml, line 73: Would it be possible to use the same names in `composer.utils.dist`? (E.g. `node_world_size` instead of `number_of_nodes`, `local_world_size` instead of `accelerators_per_node`?)

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+    composer_version: str
```
ravi-mosaicml, line 74: Can we also collect the `torch.cuda.device_count()` separately? Someone might have a misconfiguration where the device count does not match the local world size.

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+
+def get_accel_model_name() -> str:
+    """Query the accelerator name."""
+    return accel_device_name(None) if cuda_available() else "N/A"
+
+
+# Excepthook wrapper, wraps default excepthook and prints env info
+def _excepthook_wrapper(type, value, tb) -> None:
+    sys.__excepthook__(type, value, tb)
+    print("\n-------------------\n" + EXCEPTION_MSG)
```
ravi-mosaicml, line 117: Should this go to the `stderr` stream?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def get_accel_model_name() -> str:
+    """Query the accelerator name."""
+    return accel_device_name(None) if cuda_available() else "N/A"
+
+
+# Excepthook wrapper, wraps default excepthook and prints env info
+def _excepthook_wrapper(type, value, tb) -> None:
+    sys.__excepthook__(type, value, tb)
```
ravi-mosaicml, line 116: Instead of calling the fallback `__excepthook__`, could we capture the previous excepthook and call that? If the user or another package monkey-patched the excepthook, then we would not be calling that, but rather the default excepthook).

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+        >>> configure_excepthook()
+        >>> sys.excepthook 
+        <function _excepthook_wrapper at ...>
+    """
+    sys.excepthook = _excepthook_wrapper
+
+
+# Public function to restore original excepthook
+def restore_excepthook() -> None:
+    """Restore default :func:`sys.excepthook` behavior.
```
ravi-mosaicml, line 143: Instead of restoring the default excepthook, could we restore the excepthook that we patched over?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+def get_accel_model_name() -> str:
+    """Query the accelerator name."""
+    return accel_device_name(None) if cuda_available() else "N/A"
+
+
+# Excepthook wrapper, wraps default excepthook and prints env info
+def _excepthook_wrapper(type, value, tb) -> None:
+    sys.__excepthook__(type, value, tb)
+    print("\n-------------------\n" + EXCEPTION_MSG)
+    print_env()
```
ravi-mosaicml, line 118: Same thing here; should it go to `stderr`?

composer/utils/collect_env.py
```diff
@@ -0,0 +1,245 @@
...
+        # Only print Composer system info if installed
+        print(get_composer_env() + "\n\n")
+    else:
+        print("Composer not installed")
+
+
+# Invoke as standalone CLI script
+def main() -> None:
+    print("Collecting system information...")
+    print_env()
```
ravi-mosaicml, line 241: (If someone is calling this as a standalone script, agree that it should go to stdout)

ravi-mosaicml (commented):
Overall looks good! See comments.

In addition, can you post an example output? (Perhaps also include in the doctests?)

composer/utils/collect_env.py
bandish-shah: `ImportError` should work, was just sort of blindly doing this to match the convention used in the Torch `collect_env` script but it might be overly conservative:
https://github.com/pytorch/pytorch/blob/2f2158ae4502667951a4d6f5f95d751c7d0927d8/torch/utils/collect_env.py#L18

composer/utils/collect_env.py
bandish-shah: Sure, changing the messaging based on the exception.

composer/utils/collect_env.py
bandish-shah: Was originally trying to match the convention being used in the MLPerf logger since I was hoping to consolidate this eventually but makes more sense to have our naming stick to what our code uses vs. what MLPerf expects for this report.

composer/utils/collect_env.py
bandish-shah: Was originally trying to match the convention being used in the MLPerf logger since I was hoping to consolidate this eventually but makes more sense to have our naming stick to what our code uses vs. what MLPerf expects for this report.

composer/utils/collect_env.py
```diff
@@ -0,0 +1,391 @@
...
+    Args:
+        exc_type (Exception): Type of exception.
+    """
+    EXCEPTION_MSG = f"Bugs can be reported at: {COMPOSER_OPEN_ISSUE_URL}\n" + \
+                    f"Please include details on how to reproduce the issue and attach the following environment report:\n"
+
+    # Don't print exception report for KeyboardInterrupt
+    if not issubclass(exc_type, KeyboardInterrupt):
+        if issubclass(exc_type, AssertionError):
+            EXCEPTION_SEV_MSG = f"Smells like a Composer bug.\n"
```
ravi-mosaicml, line 148: Asserts are for invariants, so they are always bugs if they surface (even if it means it should be a ValueError instead of an assert)

```suggestion
            EXCEPTION_SEV_MSG = f"This is a Composer bug. Please submit a bug report.\n"
```

composer/utils/collect_env.py
```diff
@@ -0,0 +1,391 @@
...
+        nb.set_custom_exc((Exception,), _nb_custom_exception_handler)
+    else:
+        # Save original excepthook and override
+        global _orig_excepthook
+        _orig_excepthook = sys.excepthook
+        sys.excepthook = _custom_exception_handler
+
+
+# Public function to restore original excepthook
+def restore_excepthook() -> None:
```
ravi-mosaicml, line 210: If another function modified `sys.excepthook` after calling `configure_excepthook`, then when it is restored, the other excepthook would also be removed, which is probably undesired behavior. For example:

```python
configure_excepthook()
sys.excepthook = thidparty_except_hook_handler(sys.excepthook)
restore_excepthook()  # would also remove thirdparty_except_hook_handler, since we are restoring the original excepthook
```

Instead, could the `restore_excepthook` function make the `_custom_exception_handler` or `_nb_custom_exception_handler` no-ops (but still leave it in the stack?) That would preserve the ability of `thirdparty_except_hook_handler` to still fire, even though our excepthook would now be a noop.

composer/utils/collect_env.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+# Get Torch environment info
+def get_torch_env() -> str:
+    """Query Torch system environment via :mod:`torch.utils.collect_env`."""
+    return torchenv.get_pretty_env_info()
+
+
+# Composer environment information string output format
+composer_env_info_fmt = """
```
ravi-mosaicml, line 249:
```suggestion
_COMPOSER_ENV_INFO_FORMAT = """
```

composer/utils/collect_env.py
```diff
@@ -0,0 +1,391 @@
...
+if IPYTHON_AVAILABLE:
+    get_ipython = import_object('IPython:get_ipython')
+    nb = get_ipython()
+
+# Place to keep track of the original excepthook
+_orig_excepthook = None
+
+
+# Same convention as Torch collect_env, create a namedtuple to track collected fields
+class ComposerEnv(NamedTuple):
```
ravi-mosaicml, line 85: nit: Is this named tuple needed? It seems to be used only for making the format string...instead could the format string args be passed in directly?

ravi-mosaicml (approved):
LGTM, though see comments. Feel free to merge after addressing!

ravi-mosaicml:
Can you also post a sample output in the PR description? Curious to see what this looks like!

composer/utils/collect_env.py
bandish-shah: Not really, was originally sticking to the PyTorch `collect_env` convention.  In their case they set the fields to the raw return values of all the `get_*` functions and then use a set of common formatting functions to format the strings it in a function that pretties the string so this bifurcation is more convenient.  In our case for Composer the `get_*` functions output the desired format so it's not really helping.  

We may have a similar pretty function in the future to post process the strings in the future depending on what additional env information we collect and how we collect it (e.g., output of `pip freeze`, exec other sys utilities).  Could just make the change then but given this isn't really a functional or performance issue and more work to change at this point I'm going keep it as is.

hanlint:
Can we ask for feedback from users (e.g. @dblalock) on the following before merging:
```
Provides a wrapper for sys.excepthook to automatically gather system information when an exception is caught
Provides configure_excepthook() helper function to override original sys.excepthook
Provides restore_excepthook() helper function to restore original sys.__excepthook__
```

For example, what if the bug is due to some misconfiguration that happens outside of composer, but then causes exceptions within? We would be assuming a bug in Composer. For rapid prototyping and debugging, do users really _want_ a very verbose env collection every time? My current stance is that this is undesirable feature for UX.

PR #958, opened 2022-04-26
Regex Save Overwrite

mvpatel2000:
Currently, `save_overwrite=False` requires the checkpoint folder to be empty. Now, we only require that there are no files in the folder which will have the same name as a future checkpoint of the current run. We also add tests for this behavior.


composer/utils/file_helpers.py
```diff
@@ -60,6 +62,38 @@ def ensure_folder_is_empty(folder_name: Union[str, pathlib.Path]):
...
+    Args:
+        folder_name (str | pathlib.Path): The folder to inspect.
+        filename (str): The pattern string for potential files.
+        timestamp (Timestamp): Ignore any files that occur before the provided timestamp.
+
+    Raises:
+        FileExistsError: If ``folder_name`` contains any files matching the ``filename`` template before ``timestamp``.
+    """
+    # Prepare regex pattern by replacing f-string formatting with regex. Currently, we only capture and compare time values for
+    # epoch and batch.
```
ravi-mosaicml, line 77: Will the other distributed variables (local_rank, local_world_size, ...) need to be matched?

ravi-mosaicml, line 77: Is it possible to compare all timing values? Epoch, batch, sample, and token should be monotonically increasing. batch_in_epoch, sample_in_epoch, and token_in_epoch should be increasing if the epoch is the same.

ravi-mosaicml (commented):
See comments. I'll take another read after tests are added for `ensure_folder_has_no_conflicting_files`!

composer/utils/file_helpers.py
```diff
@@ -60,6 +62,62 @@ def ensure_folder_is_empty(folder_name: Union[str, pathlib.Path]):
...
+        filename (str): The pattern string for potential files.
+        timestamp (Timestamp): Ignore any files that occur before the provided timestamp.
+
+    Raises:
+        FileExistsError: If ``folder_name`` contains any files matching the ``filename`` template before ``timestamp``.
+    """
+    # Prepare regex pattern by replacing f-string formatting with regex.
+    pattern = f"^{filename}$"
+    rank_names = ["{rank}", "{local_rank}", "{world_size}", "{local_world}", "{node_rank}"]
+    for rank_name in rank_names:
```
ravi-mosaicml, line 80: The rank arguments can be formatted directly, since those are already known.

tests/utils/test_file_helpers.py
```diff
@@ -183,3 +184,86 @@ def test_format_name_with_dist_and_time():
 
 def test_ensure_folder_is_empty(tmpdir: pathlib.Path):
     ensure_folder_is_empty(tmpdir)
+
+
+def test_ensure_folder_has_no_conflicting_files(tmpdir: pathlib.Path):
+    timestamp = Timestamp(2, 7, 1, 15, 3, 31, 7)
```
ravi-mosaicml, line 190: nit: can you use kwargs for this? looks like binary haha

tests/utils/test_file_helpers.py
```diff
@@ -183,3 +184,86 @@ def test_format_name_with_dist_and_time():
...
+    run_name = "blazing-unicorn"
+    filename = f"{run_name}-ep{{epoch}}-batch{{batch}}-tie{{token_in_epoch}}-rank{{rank}}.pt"
+
+    # Ignore empty folder
+    ensure_folder_has_no_conflicting_files(tmpdir, filename, timestamp)
+
+    # Ignore timestamps in past
+    with open(os.path.join(tmpdir, f'{run_name}-ep1-batch3-tie6-rank0.pt'), 'w') as f:
```
ravi-mosaicml, line 198: nit: can each of these tests be seperate? Feel free to use pytest parameterization or a shared helper function...

tests/trainer/test_checkpoint.py
```diff
@@ -213,6 +213,88 @@ def test_load_weights(
...
+    pytest.param(GPUDeviceHparams(), id="gpu", marks=pytest.mark.gpu),
+])
+@pytest.mark.parametrize("save_overwrite", [
+    True,
+    False,
+])
+def test_save_overwrite(
+    device_hparams: DeviceHparams,
+    composer_trainer_hparams: TrainerHparams,
+    save_overwrite: bool,
```
ravi-mosaicml, line 228: Mega nit: Is it possible to wrap the `save_overide` parameterization into the existing checkpoint tests? This code looks to be slightly duplicated from what's already there?

I know this is a bit out of the scope of the PR, but the checkpointing tests are getting quite messy and hard to maintain given all of the different parameterizations. As part of the graceful resumption, I think it would be good to clean it up with better shared helpers and to de-yahpify the tests to reduce code duplication. OK if that happens in another PR or this one, but just mentioning this now as this function (as written) will add to the future cleanup.

composer/utils/file_helpers.py
```diff
@@ -18,6 +19,7 @@
     'GetFileNotFoundException',
     'get_file',
     'ensure_folder_is_empty',
+    'ensure_folder_has_no_conflicting_files',
```
ravi-mosaicml, line 22: can this function be re-exported on `composer.utils.__init__`?

composer/utils/file_helpers.py
```diff
@@ -60,6 +62,62 @@ def ensure_folder_is_empty(folder_name: Union[str, pathlib.Path]):
                 raise FileExistsError(f"{folder_name} is not empty; {os.path.join(root, file)} exists.")
 
 
+def ensure_folder_has_no_conflicting_files(folder_name: Union[str, pathlib.Path], filename: str, timestamp: Timestamp):
+    """Ensure that the given folder does not have any files conflicting with filename. If any file has same format as
```
ravi-mosaicml, line 66:
```suggestion
    """Ensure that the given folder does not have any files conflicting with the ``filename`` format string. If any filename is formatted with a timestamp where the epoch, batch, sample, or token counts are after ``timestamp``, a ``FileExistsError`` will be raised.
```

ravi-mosaicml (approved):
Overall looks good; see comments: mainly re tests.

tests/trainer/test_checkpoint.py
mvpatel2000: Ack, but will do after this https://github.com/mosaicml/composer/pull/962 in a separate PR

PR #960, opened 2022-04-27
Fix when `Engine.close()` is invoked.

ravi-mosaicml:
This PR changes how callbacks are shutdown. The overall goals are:

1. Ensure that any exit traceback is captured via the loggers (e.g. in `WandB` and the `FileLogger`)
2. Not to shutdown the callbacks at the end of `Trainer.fit()`, which is a requirement for #138

To accomplish these goals, this PR changes where `Engine.close` (which shuts down the callbacks) is run. Specifically, this method now runs at the any of these events:
1. Via `atexit`: If the exception causes the python process to crash, `atexit` callbacks will run after the traceback has been printed (test this for yourself! hard to write a test case for this)
2. on `__del__`: If the `trainer` instance is being garbage collected, it should be shut down.
3. If manually invoked via `trainer.close`: If the user is creating multiple trainers, they may want to shut one down at a specific point.

If `engine.close()` is invoked multiple times, it should be a no-op.

Implementation overview:
- Moved `close` to run on `__del__` / `atexit` as described above. The actual shutdown logic needs to be in a static method, so it does not hold a reference count in the `atexit` registry that would prevent garbage collection. Added test cases.
- Added `Event.FIT_END` since closing is now decoupled from the end of training. This allows for callbacks to log stuff at the end of training and flush (but not close) any open file handles.
- Fixed a bug in the `FileLogger` where the prefix was being printed multiple times if `write` was called multiple times per line
- Fixed the documentation for the Event class to show all events in the mock training loop
- Added tests and fixed callbacks and loggers to ensure that `.close()` and `.post_close()` implementations are idempotent.
- Added test cases to ensure that unhandled stack traces are captured in the logfile.


ravi-mosaicml:
Going to re-do this PR to be compatible with #138 and not call `sys.execpthook` on non-exit exceptions. Specifically, we should not be printing the traceback if the user has code like:

```python3
trainer = Trainer(...)

try:
  trainer.fit()
except:
  # gracefully handle exception
  pass
```

To fix, I will move the call to `trainer.engine.close`. Currently, it runs whenever `fit` finishes via a `finally`, but instead, it should be called whenever any of the following events happen:
1. Via `atexit`: If the exception causes the python process to crash, `atexit` callbacks will run after the traceback has been printed.
2. on `__del__`: If the `trainer` instance is being garbage collected, it should be shut down.
3. If manually invoked via `trainer.close`: If the user is creating multiple trainers, they may want to shut one down at a specific point.

If `close()` is invoked multiple times, it should be a no-op.

bandish-shah (approved):
LGTM

PR #961, opened 2022-04-27
Fix symlink overwrite

mvpatel2000:
Currently, we pass in the overwrite param when logging the symlink. However, it should always overwrite since latest is overwritten all the time in a training run


ravi-mosaicml (approved):
Good catch....thanks!

PR #964, opened 2022-04-27
Replaced DataLoader typing with `Iterable`; Removed `WrappedDataLoader` and the `DataLoader` Protocol

### LINKED ISSUE #424: Remove the `WrappedDataloader` and `DataLoader` protocol; replace DataLoader types with `Iterable[Batch]`

ravi-mosaicml:
- Replaced DataLoader typing with `Iterable`. This is necessary for FFCV integration. Added `isinstance` checks throughout the codebase where a `torch.utils.data.DataLoader` is required
- Removed the `WrappedDataLoader` class and `DataLoader` protocol classes. These classes are opinionated and presume a PyTorch dataloader, not a generic iterable.
- Integrated the `DataLoaderProfiler` directly into the trainer. This was much simpler than adding in the `BEFORE_DATALOADER` event (as outlined in #424), which would only be used by the profiler (algorithms would never need to run on this event, since there isn't any data yet loaded! Algorithms already have the `BATCH_START` event.) and would not guarantee accurate timing. This change was necessary as the `DataLoaderProfiler` used the now-removed `WrappedDataLoader` classes.

Closes #424


bandish-shah (approved):
LGTM, I focused on reviewing the type hints and the re-factoring of the Dataloader profiler.  It would be good to have @dskhudia's stamp on this as well to confirm requirements are met for the FFCV dataloader.

I'm not a huge fan of implementing the Dataloader profiler directly in the trainer as longer term this means users that implementing their own trainers will need to also implement this themselves.  But in all fairness this isn't something we support so it's not an issue short-term and I imagine the profiler will go through more changes anyways so this won't be the last time we move things around.

composer/algorithms/augmix/augmix.py
```diff
@@ -262,7 +265,7 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
                                     width=self.width,
                                     alpha=self.alpha,
                                     augmentation_set=self.augmentation_set)
-        assert state.dataloader is not None, "dataloader should be defined on fit start"
+        assert isinstance(state.dataloader, torch.utils.data.DataLoader), "dataloader type checked on match()"
```
dskhudia, line 268: These (augmix/colout/randaugment) algorithms as written won't work with ffcv dataloader as it has no dataset. Checking with isinstance is fine here.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -184,6 +185,8 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
             return
 
         assert state.dataloader is not None, "dataloader should be set on AFTER_DATALOADER"
+        if not isinstance(state.dataloader, torch.utils.data.DataLoader):
```
dskhudia, line 188: This algorithm can work with FFCV dataloader, however, adding this check will make it fail as ffcv.Loader is not an instance of torch.utils.data.DataLoader.

ravi-mosaicml: Good point; Replaced with a manual `getattr` check.

ravi-mosaicml:
> LGTM, I focused on reviewing the type hints and the re-factoring of the Dataloader profiler. It would be good to have @dskhudia's stamp on this as well to confirm requirements are met for the FFCV dataloader.
> 
> I'm not a huge fan of implementing the Dataloader profiler directly in the trainer as longer term this means users that implementing their own trainers will need to also implement this themselves. But in all fairness this isn't something we support so it's not an issue short-term and I imagine the profiler will go through more changes anyways so this won't be the last time we move things around.

Makes sense. We can refactor it out later when there is a need to use it outside of our trainer.

PR #965, opened 2022-04-27
Load from ObjectStoreLogger

mvpatel2000:
Adds support for restoring checkpoints from a LoggerDestination. Adds support for `get_file_artifact` from ObjectStoreLogger. Adds tests for both `get_file_artifact` in ObjectStoreLogger as well as end-to-end saving and loading checkpoints tests from ObjectStoreLogger

Next PR: 
- WandBLogger
- test_checkpoint.py cleanup (Punting this to a second PR -- I want to make the code review easier for this and give a little more thought to how to clean up the mess there)


mvpatel2000:
I'm punting WandBLogger to a different PR. I'm running into weird issues testing it... it doesn't behave nicely in the testing suite. Basically, when it tries to fetch the `project` name and other fields, it returns something strange that causes issues when loading the pickle. I can't really figure out what it's returning... and it will take me a bit of time to get to the bottom of this so I want to put up the rest for review rn

image

composer/loggers/logger_destination.py
```diff
@@ -137,3 +137,22 @@ def log_symlink_artifact(
         """
         del state, log_level, existing_artifact_name, symlink_artifact_name, overwrite  # unused
         pass
+
+    def get_file_artifact(
+        self,
+        artifact_name: str,
```
ravi-mosaicml, line 143: Would a logger destination need access to the `state` to get the file? I presume not, in which case this signature is fine, but just wanted to raise the question.

composer/loggers/wandb_logger.py
```diff
@@ -123,12 +122,10 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
                 warnings.warn(("WandB permits only alpha-numeric, periods, hyphens, and underscores in artifact names. "
                                f"The artifact with name '{artifact_name}' will be stored as '{new_artifact_name}'."))
 
-            if self._enabled and self._log_artifacts:
-                import wandb
-                extension = new_artifact_name.split(".")[-1]
-                artifact = wandb.Artifact(name=new_artifact_name, type=extension)
-                artifact.add_file(os.path.abspath(file_path))
-                wandb.log_artifact(artifact, aliases=aliases)
+            extension = new_artifact_name.split(".")[-1]
```
ravi-mosaicml, line 125: It looks like the if check and the condition import were removed? Can it be added back in?

composer/trainer/trainer_hparams.py
```diff
@@ -185,6 +185,8 @@ class TrainerHparams(hp.Hparams):
             .. seealso:: :mod:`composer.callbacks` for the different callbacks built into Composer.
         load_path (str, optional): See :class:`.Trainer`.
         load_object_store (ObjectStore, optional): See :class:`.Trainer`.
+        load_logger_destination (LoggerDestination, optional): Used to specify a ``LoggerDestination`` for
```
ravi-mosaicml, line 188: Can you mention that it should not be specified with` load_object_store`, and similar thing on the `load_object_store` docstring line?

composer/trainer/trainer_hparams.py
```diff
@@ -205,7 +207,9 @@ class TrainerHparams(hp.Hparams):
             to the trainer for the ``deepspeed_config`` parameter signaling that DeepSpeed will not be used
             for training.
         prof_schedule (ProfileScheduleHparams, optional): Profile schedule hparams. Must be specified to enable the profiler.
-        prof_trace_handlers (List[TraceHandlerHparams], optional): See :class:`.Trainer`. Must be specified to enable the profiler.        prof_skip_first (int, optional): See :class:`.Trainer`.        prof_wait (int, optional): See :class:`.Trainer`.
+        prof_trace_handlers (List[TraceHandlerHparams], optional): See :class:`.Trainer`. Must be specified to enable the profiler.
```
ravi-mosaicml, line 210: Whoops; thanks!

composer/trainer/trainer_hparams.py
```diff
@@ -340,6 +344,12 @@ class TrainerHparams(hp.Hparams):
         connecting to the cloud provider object store. Otherwise, if the checkpoint is a local filepath,
         leave blank. This parameter has no effect if `load_path` is not specified."""),
                                                                   default=None)
+    load_logger_destination: Optional[LoggerDestinationHparams] = hp.optional(doc=textwrap.dedent("""\
```
ravi-mosaicml, line 347: I think this will need an entry in the `hparams_registry`

composer/trainer/trainer_hparams.py
```diff
@@ -553,6 +563,12 @@ def initialize_object(self) -> Trainer:
 
         deepspeed_config = self.deepspeed if self.deepspeed is not None else False
 
+        load_object_store = None
+        if self.load_object_store is not None:
+            load_object_store = self.load_object_store.initialize_object()
+        elif self.load_logger_destination is not None:
+            load_object_store = self.load_logger_destination.initialize_object()
+
```
ravi-mosaicml, line 571: can you rasse a value error if both are specified?

tests/trainer/test_checkpoint.py
```diff
@@ -295,6 +298,86 @@ def test_save_overwrite(
     trainer.fit()
 
 
+pytest.mark.timeout(90)
```
ravi-mosaicml, line 301:
```suggestion
@pytest.mark.timeout(90)
```

tests/trainer/test_checkpoint.py
```diff
@@ -295,6 +298,86 @@ def test_save_overwrite(
...
+
+
+def test_checkpoint_with_object_store_logger(
```
ravi-mosaicml, line 304: Can you parameterize the `load_object_store` to be a object store or a logger destination, so we can have test coverage for this change?

ravi-mosaicml (commented):
Overall looks great! See comments.

composer/loggers/logger_destination.py
```diff
@@ -137,3 +137,22 @@ def log_symlink_artifact(
...
+
+        Args:
+            artifact_name (str): The name of the artifact.
+            destination (str): The destination filepath.
+            chunk_size (int, optional): Chunk size (in bytes). Ignored if ``path`` is a local file. (default: 1MB)
+            progress_bar (bool, optional): Whether to show a progress bar. Ignored if ``path`` is a local file.
+                (default: ``True``)
+        """
+        del artifact_name, destination, chunk_size, progress_bar  # unused
+        pass
```
ravi-mosaicml, line 158: Can you raise a `NotImplementedError` here? Unlike the other APIs (which give the logger the choice on whether to log something), I think it is important that the caller know if the file isn't downloaded.

composer/loggers/logger_destination.py
mvpatel2000: Hm... I don't think so. State should be divorced of the files accessible. I modeled signature after `get_file`

tests/trainer/test_checkpoint.py
mvpatel2000: I agree with this suggestion, but I want to punt it to next PR because I'll be refactoring the checkpoint tests anyways (ideally the parameterization just goes into the big existing test, and most of this gets deleted). Originally wrote this before the refactor decision, and imo it's cleaner to do it all in a separate PR so code review is easier

ravi-mosaicml (approved):
LGTM!

composer/loggers/wandb_logger.py
ravi-mosaicml: LMAO; I promise I can read python. Thanks!

mvpatel2000:
Hm... it's failing this test but locally passing. the error is this where the doctest gets angry for calling isinstance with the ObjectStore class. Note that this code is actually tested in test_checkpoint.py and works fine....
image

PR #967, opened 2022-04-28
Implement LAMB and LARS optimizers

growlix:
Some SOTA resnet recipes use LAMB, and mlperf uses LARS, so I thought they'd be good to try. The LAMB implementation is actually [NVLAMB](https://developer.nvidia.com/blog/pretraining-bert-with-layer-wise-adaptive-learning-rates/) with decoupled weight decay. The LARS implementation will also feature decoupled weight decay.


composer/optim/decoupled_weight_decay.py
```diff
@@ -302,3 +302,222 @@ def step(self, closure=None):
...
+
+                # Compute trust ratio
+                trust_ratio = torch.where(
+                    w_norm > 0,
+                    torch.where(g_norm > 0, w_norm / g_norm, one_tensor),
+                    one_tensor,
+                )
+                if trust_clip:
+                    # LAMBC trust clipping fixes upper bound to one
+                    trust_ratio = torch.minimum(trust_ratio, one_tensor)
```
abhi-mosaic, line 414: Could we do something like this:

```python
w_norm = torch.linalg.vector_norm(param, ord=2, dim=None)
u_norm = torch.linalg.vector_norm(update, ord=2, dim=None)

# Compute trust ratio
# If weight norm or update norm are 0, leave trust_ratio as 1.0
trust_ratio = w_norm.div_(u_norm)._nan_to_num(nan=1.0, inf=1.0)

if trust_clip:
  # LAMBC trust clipping fixes upper bound to one
  trust_ratio = trust_ratio._clamp(max=1.0)

```

growlix: I love it

hanlint:
@growlix closing this for now.. please re-open if needed.

PR #969, opened 2022-04-28
Adds Htop (and forces rebuild)

Averylamp:


docker/Makefile
```diff
@@ -48,6 +48,7 @@ build: composer
 pytorch:
 	cd pytorch && docker build -t $(PYTORCH_TAG) \
 		--target $(TARGET) \
+		--no-cache \
```
ravi-mosaicml, line 51: No cache will not use the cache at all, vs `pull` will attempt to use a later version of the image, if available.

```suggestion
		--pull \
```

ravi-mosaicml (approved):
LGTM; See comments

Averylamp:
Turns out this didnt solve the issue I hoped it would, so closing

PR #970, opened 2022-04-28
Improve formatting of Implementation details for AGC

eracah:
Markdown squashed all the lines into one, so this commit wraps``` around the kinda pseudocode.


ravi-mosaicml (commented):
Can we attempt to fix the markdown rather than using pseudocode? Markdown may require extra newlines.

hanlint (approved):

PR #971, opened 2022-04-28
Typo fixes in AGC method card

eracah:
Markdown squashed all the lines into one, so this commit wraps``` around the kinda pseudocode.


ravi-mosaicml (approved):
LGTM

PR #972, opened 2022-04-28
[nit] Default to `python3` in the `Makefile`

ravi-mosaicml:
When not in a virtualenv, `python` could point to `python2`. Setting the `PYTHON ?= python3` is a better default.


eracah:
LGTM! Thanks, @ravi-mosaicml!

eracah (approved):

PR #974, opened 2022-04-29
Fix docker /tmp folder

ravi-mosaicml:
Somehow the `/tmp` folder was getting removed in the Dockerimage. This PR fixes that by not `rm -rf /tmp/*` but rather by removing the specific files written to `/tmp` in the proper build step.

### DIFF
--- a/docker/pytorch/Dockerfile
+++ b/docker/pytorch/Dockerfile
@@ -104,7 +104,8 @@ ARG PILLOW_SIMD_VERSION=7.0.0.post3
 COPY pillow_stub /tmp/pillow_stub
 
 RUN pip${PYTHON_VERSION} install --no-cache-dir --upgrade /tmp/pillow_stub && \
-    pip${PYTHON_VERSION} install --no-cache-dir --upgrade pillow_simd==${PILLOW_SIMD_VERSION}
+    pip${PYTHON_VERSION} install --no-cache-dir --upgrade pillow_simd==${PILLOW_SIMD_VERSION} && \
+    rm -rf /tmp/pillow_stub
 
 #################
 # Install Pytorch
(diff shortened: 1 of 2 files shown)


docker/pytorch/Dockerfile
```diff
@@ -125,17 +126,13 @@ ARG MOFED_VERSION
...
 
 RUN if [ -n "$MOFED_VERSION" ] ; then \
-        wget -nv -P /tmp http://content.mellanox.com/ofed/MLNX_OFED-${MOFED_VERSION}/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}.tgz && \
-        tar -zxvf /tmp/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}.tgz -C /tmp && \
-        /tmp/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}/mlnxofedinstall --user-space-only --without-fw-update --force ; \
+        mkdir -p /tmp/mofed && \
+        wget -nv -P /tmp/mofed http://content.mellanox.com/ofed/MLNX_OFED-${MOFED_VERSION}/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}.tgz && \
+        tar -zxvf /tmp/mofed/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}.tgz -C /tmp/mofed && \
+        /tmp/mofed/MLNX_OFED_LINUX-${MOFED_VERSION}-${MOFED_OS_VERSION}/mlnxofedinstall --user-space-only --without-fw-update --force && \
+        rm -rf /tmp/mofed \
```
dskhudia, line 133: missing semicolon.

dskhudia:
Not sure if this is the root cause. `rm -rf /tmp/*` shouldn't delete `/tmp`.

dskhudia (approved):

ravi-mosaicml:
> Not sure if this is the root cause. `rm -rf /tmp/*` shouldn't delete `/tmp`.

Agreed that it shouldn't, but these fixes do leave the `/tmp` folder behind. Not sure why this works, though.

dskhudia:
> > Not sure if this is the root cause. `rm -rf /tmp/*` shouldn't delete `/tmp`.
> 
> Agreed that it shouldn't, but these fixes do leave the `/tmp` folder behind. Not sure why this works, though.

May be some docker optimization that removes empty folders and with these fixes /tmp is no longer empty.

PR #976, opened 2022-04-29
Only push docker images if the branch is `dev`

ravi-mosaicml:
Docker changes should only be pushed if the branch is `dev`, not `main`, as not all dockerfile changes may be cherry-picked in a maintenance release. Since all commits already go on the dev branch, the docker images are already up-to-date, so they should not be updated again when the `main` branch is updated for releases.


bandish-shah (approved):
LGTM

PR #978, opened 2022-04-29
Create `batch_get` and `batch_set` helpers with test cases

eracah:
CO-25: Add setters and getters for arbitrary batch types and arbitrary batch keys.

This commit does the following:
* Creates `composer/utils/batch_helpers.py` module
* Adds `batch_get` and `batch_set` functions to  `composer/utils/batch_helpers.py`
* Adds tests for `batch_get` and `batch_set` in `tests/utils/test_batch_helpers.py`
* Adds State.batch_set_item and State.batch_get_item to  `composer/core/state.py`
* Adds tests for State.batch_set_item and State.batch_get_item to `tests/test_state.py`


eracah:
Help! I have all these old commits in this PR!

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
+from typing import Any
+
+from composer.core.types import BatchKeys
+
+
+def batch_get(batch: Any, key: BatchKeys) -> Any:
```
ravi-mosaicml, line 6: I would be OK with typing the `key` as `Any` -- main reason being that if if the batch is a dict, we do not know what type the key would be -- it's likely to be a string, but could really be anything.

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    """Indexes into the batch given the key.
+
+    Tries all the common combination of batch and key pairs:
```
ravi-mosaicml, line 9: I get where we're going with this list, but I find it a bit confusing on what is getting done here. Perhaps could you add (doctested) examples here?

It might also be worth mentioning that the first operation that succeeds is returned.

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+        return [batch[k] for k in key]  # Sequence of ints, slices, or strings.
+    except:
+        pass
+    try:
+        return getattr(batch, key)  # Attribute for object or named tuple.
+    except:
+        pass
+    try:
+        return [getattr(batch, k) for k in key]  # Sequence of attributes for object or named tuple.
+    except:
```
ravi-mosaicml, line 48: Can you also handle the case where the `key` is a "getter' callable or a (getter, setter) tuple? I think this will be helpful when updating the algorithms to use this function, so each algorithm individually does not have to condition what the key is.

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+        raise RuntimeError(f"Batch object can't be indexed by nor has attribute {key}")
```
ravi-mosaicml, line 49: Wondering if this should be a `NotImplementedError`? In addition, could you mention in the error message that one should pass in a getter lambda as the key for custom batch types (see above comment).

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
+from typing import Any
+
+from composer.core.types import BatchKeys
+
```
ravi-mosaicml, line 4: This file will need an `__all__` and class-level docstring -- see https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#62-use-of-__all__ and https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#4-public-apis?

ravi-mosaicml (commented):
Overall looks good -- see comments on `batch_get`. Will re-review once the base branch is updated (that should minimize the diff).

composer/utils/batch_helpers.py
hanlint: yes, agreed!

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    Returns:
+        The part of the batch specified by the key. This could be any type depending on
+            what the batch is composed of.
+
+    Raises:
+        RuntimeError if the batch can't be indexed by the key.
+    """
+    try:
+        return batch[key]  # int or slice for sequence or string for dict.
+    except:
```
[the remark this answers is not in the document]
hanlint: would you consider using explicit guards here instead? e.g.
```
if isinstance(batch, dict):
    return batch[key]
```
etc?

hanlint (commented):
Looks good! Most major comment is whether it would be more readable to use explicit guards rather than a series of try.. except. The try with blanket except can be brittle, e.g. a `torch.Tensor` also supports `[ ]` indexing, but that shouldn't be used to get.

composer/utils/batch_helpers.py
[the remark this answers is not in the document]
A-Jacobson: I agree with hanlin on this. though we'd probably want to use `Mapping` rather than dict to cover other dictlike stuff.

composer/utils/batch_helpers.py
eracah: ok fixed in  [b64b46c](https://github.com/mosaicml/composer/pull/978/commits/b64b46ce1559530b6ccb8dd3747d5799f583d727) and 
d9c6694

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,136 @@
...
+            else:
+                batch_list = list(batch)
+                batch_list[k] = v
+                batch = batch._make(batch_list)
+        return batch
+    
+    else:     # Normal tuple.
+        if isinstance(key, (List, Tuple)):
+            batch = list(batch)
+            for k,v in zip(key, value):
```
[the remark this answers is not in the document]
eracah: Ok makes sense! fixed here: [66465f1](https://github.com/mosaicml/composer/pull/978/commits/66465f175a24a7575115e6fa5678b7c2d091ef3a)

composer/utils/batch_helpers.py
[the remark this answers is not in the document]
eracah: yeah good point. explicit guards makes sense. I wanted to do EAFP instead of LBYL, but I guess in this case LBYL makes more sense. Fixed here:  05fd3f1

tests/utils/test_batch_helpers.py
```diff
@@ -0,0 +1,164 @@
...
+import torch
+
+from composer.utils.batch_helpers import batch_get, batch_set
+
+my_list = [3, 4, 5, 6, 7, 8, 9, 10]
+
+keys = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
+
+
+class myClass(object):
```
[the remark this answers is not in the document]
eracah: fixed here: 
d0b9377

composer/utils/batch_helpers.py
eracah: Done.

eracah: Removed comment and changed from try except to if else

composer/utils/batch_helpers.py
eracah: Any chance I could do that in a new PR? This PR already has a lot of stuff in it. I feel like we should just get it in?

composer/utils/batch_helpers.py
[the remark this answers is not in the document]
eracah: @ravi-mosaicml and I think that maybe EAFP does make more sense because there might be some custom type that has a __getitem__ function that does not fall under Mapping, Sequence, torch.Tensor, np.ndarray, and it may not be scalable to keep adding more `elif` clauses every time the user breaks the library with a custom type.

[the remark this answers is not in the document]
eracah: wdyt, @hanlint, @A-Jacobson ?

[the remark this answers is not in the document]
hanlint: EAFP could be reasonable here too, but just be sure to catch the right errors (e.g. `batch[key]` could either be a `KeyError` or a `TypeError`, but we should only catch the latter).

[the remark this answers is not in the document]
eracah: Ok switched back to EAFP. The code is a little less readable now, but maybe more general to new types. Lemme know what you think, @ravi-mosaicml, @hanlint

composer/utils/batch_helpers.py
eracah: done

tests/utils/test_batch_helpers.py
```diff
@@ -0,0 +1,164 @@
+from collections import OrderedDict
+from typing import NamedTuple
+
+import numpy as np
+import pytest
+import torch
+
+from composer.utils.batch_helpers import batch_get, batch_set
```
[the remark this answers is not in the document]
eracah: thanks!

composer/core/types.py
```diff
@@ -30,6 +30,9 @@
 
 JSON = Union[str, float, int, None, List['JSON'], Dict[str, 'JSON']]
 
+BatchKey = Union[int, str, slice]
```
hanlint, line 33: probably don't need those anymore?

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+    if isinstance(key, Sequence) and not isinstance(key, str):
+        return _batch_get_multiple(batch, key)
+    else:
+        return _batch_get(batch, key)
+
+
+def _batch_get(batch: Any, key: Any) -> Any:
+    if isinstance(batch, tuple):
+        return _batch_get_tuple(batch, key)
+    else:
```
hanlint, line 29: don't need the `else` indent here

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            return value
+
+
+def _batch_get_multiple(batch: Any, key: Any):
+    return [_batch_get(batch, k) for k in key]
+
+
+def batch_set(batch: Any, key: Any, value: Any) -> Any:
+    """Indexes into the batch given the key and sets the element at that index to value.
```
hanlint, line 61: should clarify that this is not garaunteed to be an in-place operation (eg. for tuples).

hanlint (commented):
Looks good. P.S. to ask for a re-review, click the re-request button on the upper right.

tests/test_state.py
```diff
@@ -92,3 +97,66 @@ def test_state_serialize(
 
     # both states should have equivalent state, model parameters, loss, and outputs
     assert_state_equivalent(state1, state2)
+
+
+my_list = [3, 4, 5, 6, 7, 8, 9, 10]
```
ravi-mosaicml, line 102: Since the state helpers use the `batch_get` / `batch_set` helpers, no need for an extensive matrix here. It's best that each test does only one thing -- here that would be to ensure that these methods exist on the state and call the helpers. Perhaps here we could monkeypatch `batch_get` / `batch_set` to use dummy values, and then assert that `state.batch_get_item(...)` returns the dummy value, and after `state.batch_set_item(...)`, `state.batch` is then the dummy batch?

tests/utils/test_batch_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+                    my_named_tuple(*my_list),
+                    defaultdict(list, **dict(zip(keys, my_list))),
+                    ChainMap(dict(zip(keys, my_list)), dict(a=7, j=3)),
+                    Counter(counter_list),
+                    OrderedDict(**dict(zip(keys, my_list)))
+                ])
+def example_map(request):
+    return request.param
+
+
```
ravi-mosaicml, line 66: Could you also add tests for where the key is a torch tensor or numpy array? Mainly want to ensure that this will trigger `batch_get_item` rather than `batch_get_sequence`.

ravi-mosaicml (commented):
Took another pass -- overall looks good. Main comments are re: doc / doctested examples, plus handling the getter / setter / (getter, setter) tuples.

composer/core/types.py
eracah: ok fixed here: 
bfb82cb

composer/utils/batch_helpers.py
eracah: ok fixed here: 
4735295

composer/utils/batch_helpers.py
eracah: ok fixed here: 
2136dcb

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def batch_get(batch: Any, key: Any) -> Any:
+    """Indexes into the batch given the key.
+    Args:
+        batch (Any): An object that contains the input and label of the items in the batch.
+            Can be any abritrary type that user creates, but we assume some sort of
+            sequence (list, tuple, tensor, array), mapping (dictionary),
+            or attribute store (object with data members, namedtuple).
+        key (BatchKeys): A key to index into the batch. BatchKeys is of type Union[BatchKey Sequence[BatchKey]]
```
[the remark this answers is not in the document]
eracah: good catch! fixed here: 
5b714c3

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+            else:
+                raise e
+        else:
+            return value
+
+        try:
+            value = getattr(batch, key)
+
+        # If both getattr and __getitem__ result in exceptions then raise both of them.
+        except Exception as e:  # AttributeError, TypeError.
```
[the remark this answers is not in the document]
eracah: Lol. Fixed here: 
8f7c5e1

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+        return _batch_get(batch, key)
+
+
+def _batch_get(batch: Any, key: Any) -> Any:
+    if isinstance(batch, tuple):
+        return _batch_get_tuple(batch, key)
+    else:
+        exceptions = []
+        try:
+            value = batch[key]
```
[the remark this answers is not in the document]
eracah: fixed here 66ed8da

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+def _batch_set_multiple(batch: Any, key: Any, value: Any) -> Any:
+    """Sets multiple key value pairs in a non-tuple batch."""
+    if len(key) != len(value):
+        raise ValueError(f'value must be the same length as key ({len(key)}), but it is f{len(value)} instead')
+    for single_key, single_value in zip(key, value):
+        batch = _batch_set(batch, single_key, single_value)
+    return batch
+
+
+def _batch_set_tuple(batch: Any, key: Any, value: Any) -> Any:
```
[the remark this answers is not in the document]
eracah: fixed here: 
c9a6a42

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+    """"Sets key value pairs in tuples and NamedTuples."""
+    if hasattr(batch, '_fields'):  # NamedTuple
```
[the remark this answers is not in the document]
eracah: Gonna keep this one as is. Trying to get too general with it sets off pyright issues and unit test failures for keys that are slices. Seems reasonable enough to have explicit cases for NamedTuple[str], tuple, and NamedTuples with int or slice keys.

tests/test_state.py
eracah: Good call! I added a pair of mock.patch tests here using `return_value` and `assert_called_once` here: [b74ebd1](https://github.com/mosaicml/composer/pull/978/commits/b74ebd19e5d87682b6bdb41d31e91ff97ea15716)

tests/utils/test_batch_helpers.py
eracah: I'm not sure what you mean this?

composer/utils/batch_helpers.py
ravi-mosaicml: (Sorry, didn't see your reply here when I made the comment again below!) Sure, if you want, but I also don't think it'd be too hard to do it here....e.g.

```python3
if callable(key):  # key is a getter
  return key(batch)
if isinstance(key, tuple) and len(key) == 2 and callable(key[0]):  # (getter, setter) tuple
  return key[0](batch)
```

tests/utils/test_batch_helpers.py
ravi-mosaicml: Something like

```python
x = torch.tensor([[1, 2], [3, 4]])
y = torch.tensor([[True, False], [True, False]])
z = torch.tensor([1, 3])
assert z == batch_get(batch=x, key=y)
```

eracah:
@ravi-mosaicml, I added support for indexing multi-dimensional arrays and tensors here: 
4a1373f. check it out when you get a chance

tests/utils/test_batch_helpers.py
eracah: ok added the functionality for keys of type lists, tuples, arrays, tensors, where the batch is of type array or tensor here (with tests for tuple and list keys): [4a1373f](https://github.com/mosaicml/composer/pull/978/commits/4a1373f07a36bc1f23ceb7b28093b119c45c1da2) and then added tests for keys of type tensor or array here: [937ea33](https://github.com/mosaicml/composer/pull/978/commits/937ea33f4313d717cf1fcaac8879f32b089bb6e7)

composer/utils/batch_helpers.py
eracah: ok done here: 
418c669

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+            return value
+
+
+def _batch_get_multiple(batch: Any, key: Any):
+    return [_batch_get(batch, k) for k in key]
+
+
+def batch_set(batch: Any, key: Any, value: Any) -> Any:
+    """Indexes into the batch given the key and sets the element at that index to value.
+    Args:
```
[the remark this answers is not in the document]
eracah: ok added doctests here: 
bc6edc3

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+def batch_get(batch: Any, key: Any) -> Any:
+    """Indexes into the batch given the key.
+    Args:
+        batch (Any): An object that contains the input and label of the items in the batch.
+            Can be any abritrary type that user creates, but we assume some sort of
+            sequence (list, tuple, tensor, array), mapping (dictionary),
+            or attribute store (object with data members, namedtuple).
+        key (BatchKeys): A key to index into the batch. BatchKeys is of type Union[BatchKey Sequence[BatchKey]]
+            and BatchKey is of type Union[int, str, slice].
+
```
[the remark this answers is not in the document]
eracah: done! here: 
bc6edc3

eracah:
Ok I tried to address all the comments! Let me know what you think.

tests/test_state.py
```diff
@@ -93,3 +94,43 @@ def test_state_serialize(
 
     # both states should have equivalent state, model parameters, loss, and outputs
     assert_state_equivalent(state1, state2)
+
+
+def test_state_batch_get_item():
+    state = get_dummy_state()
+    state.batch = [1, 2]
+    with patch('composer.core.state.batch_get') as mock_batch_get:
```
ravi-mosaicml, line 102: nit: instead of using `unittest.mock.patch`, could you use the pytest `monkeypatch` fixture? This works; it's just that the rest of the codebase uses the monkeypatch fixture. See: https://docs.pytest.org/en/6.2.x/monkeypatch.html

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,211 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Callable, Optional, Sequence, Union
+
+__all__ = ['batch_get', 'batch_set']
+
+
+def batch_get(batch: Any, key: Optional[Any] = None, get_fn: Optional[Callable] = None) -> Any:
```
ravi-mosaicml, line 9: I like how this API style is explicit, but just want to confirm how an algorithm would call this?

```python

class MyAlg:
  def __init__(self, key: Any = None, get_fn: Optional[Callable[[], Any]] = None, set_fn: Optional[Callable[[str, Any], None]] = None):
    self.key, self.get_fn, self.set_fn = key, get_fn, set_fn

  def apply(self, state, event):
    batch_element = state.get_item_in_batch(self.key, self.get_fn)
    batch_element = apply_alg_magic(batch_element)
    state.set_item_in_batch(key=self.key, value=batch_element, set_fn=self.set_fn)
```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+        try:
+            value = getattr(batch, key)
+
+        # If both getattr and __getitem__ result in exceptions then raise both of them.
+        except Exception as e:  # AttributeError, TypeError.
+            exceptions.append(e)
+            raise Exception(exceptions)
+        else:
+            return value
+
```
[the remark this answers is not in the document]
ravi-mosaicml: (OK with not handling this here -- see above algorithm example.)

ravi-mosaicml (commented):
Took another pass -- getting close!

tests/test_state.py
eracah: Ok, but I still have to use Mock to get the assert_called_with behavior.

A-Jacobson (approved):

composer/core/state.py
```diff
@@ -336,6 +336,12 @@ def schedulers(self):
     def schedulers(self, schedulers: Union[types.PyTorchScheduler, Sequence[types.PyTorchScheduler]]):
         self._schedulers[:] = ensure_tuple(schedulers)
 
+    def batch_set_item(self, key: Optional[Any] = None, value: Optional[Any] = None, set_fn: Optional[Callable] = None):
```
A-Jacobson, line 339: can we move up or duplicate the batch_set/batch_get docstrings to here? Most people implementing algorithms should use this functionality directly from the state object so we should make it easy for them to find out how it works.

tests/test_state.py
eracah: ok done here: 0164bf0

composer/utils/batch_helpers.py
eracah: Yeah that looks right

eracah: I guess that means the set_fn would have to take a value as input. Let me change that!

eracah: Good call!

eracah: fixed here: 
55ace59

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,211 @@
...
+    >>> batch_get([{'a':1, 'b':7},{'c':5}], get_fn=lambda x: x[1]['c'])
+    5
+
+    Args:
+        batch (Any): An object that contains the input and label of the items in the batch.
+            Can be any abritrary type that user creates, but we assume some sort of
+            sequence (list, tuple, tensor, array), mapping (dictionary),
+            or attribute store (object with data members, namedtuple).
+        key (Any): A key to index into the batch. Key is optional if get_fn is supplied.
+        get_fn (Callable): A user-specified function to do the extracting. Note, the key
```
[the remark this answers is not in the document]
eracah: done here: [5933f6e](https://github.com/mosaicml/composer/pull/978/commits/5933f6e1af503ab4a76357e2460eb0b3ff376983)

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,211 @@
...
+    # Indexing a list with a sequence results in TypeError
+    # Indexing an array/tensor with a sequence that is longer than the rank of the array
+    # results in an IndexError.
+    # Indexing a dict with a tuple results in a key error.
+    except (IndexError, TypeError, KeyError):
+        pass
+    return [_batch_get(batch, k) for k in key]
+
+
+def batch_set(batch: Any,
```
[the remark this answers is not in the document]
eracah: everything after batch specified via keyword done here: 
18000e6

composer/utils/batch_helpers.py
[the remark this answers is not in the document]
eracah: fixed here: 
d860421

composer/utils/batch_helpers.py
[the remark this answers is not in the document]
eracah: callable type specificity fixed here: 
d860421

composer/core/state.py
eracah: good idea. done here: [247f2bf](https://github.com/mosaicml/composer/pull/978/commits/247f2bf5f8665e93f0d8c29cd669f11dddb627b9)

eracah:
Ok, @ravi-mosaicml, how we looking? Ready for merge?

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,145 @@
...
+            if 'object does not support item assignment' in str(e):
+                exceptions.append(e)
+                pass
+            else:
+                raise e
+        else:
+            return batch
+
+        try:
+            setattr(batch, key, value)
```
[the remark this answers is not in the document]
eracah: fixed here: 
de417d0

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+    5
+
+    Args:
+        batch (Any): An object that contains the input and label of the items in the batch.
+            Can be any abritrary type that user creates, but we assume some sort of
+            sequence (list, tuple, tensor, array), mapping (dictionary),
+            or attribute store (object with data members, namedtuple).
+        key (Any): A key to index into the batch. Key is optional if get_fn is supplied.
+        get_fn (Callable): A user-specified function to do the extracting. 
+        get_fn is optional if key is supplied.
```
ravi-mosaicml, line 27:
```suggestion
            get_fn is optional if key is supplied.
```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+    # Numpy arrays and Torch tensors can take tuples and lists as keys.
+    try:
+        return batch[key]
+    # Indexing a list with a sequence results in TypeError
+    # Indexing an array/tensor with a sequence that is longer than the rank of the array
+    # results in an IndexError.
+    # Indexing a dict with a tuple results in a key error.
+    except (IndexError, TypeError, KeyError):
+        pass
+    return [_batch_get(batch, k) for k in key]
```
ravi-mosaicml, line 84: Should this recurse with `batch_get` instead of `_batch_get`? (So it would be recursive, if necessary?)

I don't see a use case for this, so feel free to ignore -- if someone is doing something that complex, then they should probably pass in a custom getter.

```suggestion
    return [batch_get(batch, k) for k in key]
```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+
+def _batch_get_tuple(batch: Any, key: Union[int, str]) -> Any:
+    """"Gets keys in tuples and NamedTuples."""
+    is_named_tuple = hasattr(batch, '_fields')
+    if is_named_tuple and isinstance(key, str):  # NamedTuple w/ named key
+        value = getattr(batch, key)
+    else:  # Normal tuple or namedtuple with int key.
+        value = batch[key]
+
+    return value
```
ravi-mosaicml, line 226: newline
```suggestion
    return value

```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+    # Indexing an array/tensor with a sequence that is longer than the rank of the array
+    # results in an IndexError.
+    except (IndexError, TypeError, KeyError):
+        pass
+    if not hasattr(value, '__len__') or isinstance(value, str):
+        raise ValueError(f'value must be a sequence or array or tensor! and not {type(value)}')
+    if len(key) != len(value):
+        raise ValueError(f'value must be the same length as key ({len(key)}), but it is {len(value)} instead')
+    for single_key, single_value in zip(key, value):
+        batch = _batch_set(batch, single_key, single_value)
```
ravi-mosaicml, line 196: (Same question regarding recursion as the `batch_get`)

```suggestion
        batch = batch_set(batch, single_key, single_value)
```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+
+    try:
+        # Make sure batch has key before setting it.
+        getattr(batch, key)
+        setattr(batch, key, value)
+
+    # If both (setattr or getattr) and __setitem__ raise exceptions then raise both of them.
+    except (AttributeError, TypeError) as e:
+        exceptions.append(e)
+        raise Exception(exceptions)
```
ravi-mosaicml, line 171: Is it possible to raise a more specific exception? Perhaps a `TypeError`, `NotImplementedError`, `ValueError`, or a `RuntimeError`? I think any of these would be appropriate; don't have a strong opinion on which one. A more specific exception would make it possible for the caller to catch it.

In addition, is it necessary to re-raise the underlying exceptions? A bit torn on whether that would help the caller fix the issue. Thoughts about something like?

```suggestion
        raise RuntimeError(f"Unable to set key {key} to value {value} on batch {batch}. Please specify a custom set_fn, if necessary.")
```

composer/utils/batch_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+        else:
+            raise e
+
+    try:
+        return getattr(batch, key)
+
+    # If both getattr and __getitem__ result in exceptions then raise both of them.
+    except (AttributeError, TypeError) as e:
+        exceptions.append(e)
+        raise Exception(exceptions)
```
ravi-mosaicml, line 71: See comment below on the set_fn exception.

ravi-mosaicml (approved):
LGTM! Great work; see comments. Main concern is about raising a generic `Exception`, which makes it impossible for the caller to catch specifically, but otherwise :+1: .

composer/utils/batch_helpers.py
eracah: done here: 
bd3dce1

composer/utils/batch_helpers.py
eracah: done here: 
bd3dce1

composer/utils/batch_helpers.py
eracah: hmmm yeah that could be interesting.

composer/utils/batch_helpers.py
eracah: recursion makes sense for a batch that is a complicated data structure, but in this case it would be the key that would be really complicated because as you can see the key is recursively unpacked here, but the batch is unchanged.

eracah: I think for a really complex key the user can just write a get_fn or set_fn

composer/utils/batch_helpers.py
eracah: but only works for recursively unpacking a key not a batch, so probably too niche of a use case to support.

PR #980, opened 2022-05-02
Adding Threshold Callback + Tests

anisehsani:
Threshold callback halts training when the monitored metric reaches a threshold value.


ravi-mosaicml (commented):
This PR is pretty similar to the early stopping callback, so same comments from there apply here to :) Let me know when I should take another look :)

composer/callbacks/callback_hparams.py
```diff
@@ -168,6 +169,53 @@ def initialize_object(self) -> EarlyStopper:
...
+                                      default=False)
+
+    def initialize_object(self) -> ThresholdStopper:
+        """Initialize the ThresholdStopper callback.
+
+        Returns:
+            ThresholdStopper: An instance of :class:`~.ThresholdStopper`.
+        """
+        comp_function = None
+        if self.comp in ("greater", "gt"):
```
ravi-mosaicml, line 206: Can this logic be inside of the ThresholdStopper (so one could pass a string in there)?

ravi-mosaicml (approved):
LGTM! See comments

PR #981, opened 2022-05-02
Early Stopping Callback

anisehsani:
Halts training when a monitored metric doesn't improve for a certain number of epochs.


ravi-mosaicml (commented):
Thanks for working on this one! See comments; also CC @A-Jacobson reviewing the implementation and API interface.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+        comp: Callable = None,
+        min_delta: float = 0.0,
+        patience: int = 1,
+    ):
+        self.monitor = monitor
+        self.eval_label = eval_label
+        self.comp = comp
+        self.min_delta = abs(min_delta)
+        if self.comp is None:
+            self.comp = np.less if 'loss' in monitor.lower() or 'error' in monitor.lower() else np.greater
```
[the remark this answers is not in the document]
hanlint: we probably also want to cover perplexity here, so maybe something like:
```
self.comp = np.less if monitor.lower() in ['loss', 'error', 'perplexity'] else np.greater
```

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+            if self.comp == np.less:
+                self.min_delta *= -1
+
+        if self.comp(1, 2):
```
hanlint, line 42: a custom `comp` might not actually work with integers, so this could be dangerous. I think we can skip setting these and let the user take control here.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                logger.warning(f"Couldn't find the metric {self.monitor} in the current_metrics/{self.eval_label}")
+                return
+        else:
+            logger.warning(
+                f"The label {self.eval_label} isn't in the state's current_metrics. Use the values train, eval, or the name of the Evaluator if using Evaluators."
+            )
+            return
+
+        if self.comp(monitored_metric - self.min_delta, self.best):
```
hanlint, line 70: i think if we do:
```
if self.comp(monitored_metric, self.best) and abs(monitored_metric - self.best) > min_delta
```

we could probably avoid having to sorry about setting the sign of min_delta, correct?

hanlint (commented):
Looking good, a few comments!

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+__all__ = ["EarlyStopper"]
+
+
+class EarlyStopper(Callback):
+    __doc__ = f"""Callback to halt training early.
+    """
+
+    def __init__(
+        self,
+        monitor: str,
```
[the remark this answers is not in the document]
anisehsani: Monitor is the name of the metric that is being monitored. I copied the naming from the fast ai callbacks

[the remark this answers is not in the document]
hanlint: Yeah, it's also similarly used in Keras: https://keras.io/api/callbacks/early_stopping/

tests/callbacks/test_early_stopper.py
```diff
@@ -0,0 +1,80 @@
...
+from composer.loggers import LogLevel
+from tests.common import RandomClassificationDataset, SimpleModel
+
+
+@pytest.mark.parametrize('metric_sequence', [[0.1, 0.2, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9], [0.1, 0.2]])
+def test_threshold_stopper(metric_sequence: List[int]):
+
+    early_stopper = EarlyStopper("Accuracy", "eval", patience=3)
+
+    def test_compute_and_log_metrics(self, dataloader_label: str, log_level: LogLevel, metrics: MetricCollection):
```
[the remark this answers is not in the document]
anisehsani: Ah - that makes sense - will do

hanlint:
I'm sure @moinnadeem will ask this.. but how hard would it be to generalize `patience` to `Time` (e.g. `55ba`, etc.), of course up to the nearest resolution of `eval_interval`.

tests/callbacks/test_early_stopper.py
```diff
@@ -0,0 +1,80 @@
...
+        ),
+        eval_dataloader=DataLoader(
+            RandomClassificationDataset(shape=(5, 1, 1)),
+            batch_size=4,
+        ),
+        max_duration="30ep",
+        callbacks=[early_stopper],
+    )
+
+    trainer._compute_and_log_metrics = MethodType(test_compute_and_log_metrics, trainer)
```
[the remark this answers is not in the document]
anisehsani: thanks

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+from composer.core.callback import Callback
+from composer.loggers import Logger
+
+log = logging.getLogger(__name__)
+
+__all__ = ["EarlyStopper"]
+
+
+class EarlyStopper(Callback):
+    __doc__ = f"""Callback to halt training early.
```
A-Jacobson, line 22: lets add a full docstring explaining usage and args.

also, is there a reason we're manually setting `__doc__` here?

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+            # if the monitored metric is an eval metric or in an evaluator
+            return
+
+        monitored_metric = None
+        current_metrics = state.current_metrics
+        if self.eval_label in current_metrics:
+            if self.monitor in current_metrics[self.eval_label]:
+                monitored_metric = current_metrics[self.eval_label][self.monitor]
+            else:
+                logger.warning(f"Couldn't find the metric {self.monitor} in the current_metrics/{self.eval_label}")
```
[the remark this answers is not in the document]
A-Jacobson: Yes, i think we should error out because early stopping wont work without a metric to monitor. I also think we should do this check way earlier (at training start). It's no fun to train for a full epoch (can be hours with some datasets), eval then because of a typo.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,107 @@
...
+
+class EarlyStopper(Callback):
+    __doc__ = f"""Callback to halt training early.
+    """
+
+    def __init__(
+        self,
+        monitor: str,
+        eval_label: str,
+        comp: Callable = None,
```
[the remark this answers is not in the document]
anisehsani: Need more clarification on this - since I don't know what type of metrics are tracked are, I'm not sure if I'd know what A and B are. Also not sure about combining them.

maybe new docs might help clarify?

anisehsani:
made the changes for the generalized interval types @hanlint

composer/callbacks/early_stopper.py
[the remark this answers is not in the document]
hanlint: In that case, I think you can do `Callable[[Any, Any], bool]` -- it should take any two inputs and return a boolean.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,130 @@
...
+            self.best_occurred = state.timer.get_timestamp()
+
+        if self.patience.unit == TimeUnit.EPOCH:
+            if int(state.timer.epoch) - int(self.best_occurred.epoch) > int(self.patience):
+                state.max_duration = state.timer.batch
+        elif self.patience.unit == TimeUnit.BATCH:
+            if int(state.timer.batch_in_epoch) - int(self.best_occurred.batch_in_epoch) > int(self.patience):
+                state.max_duration = state.timer.batch
+        else:
+            raise ValueError(f"The units of `patience` should be EPOCH or BATCH.")
```
hanlint, line 120: i _think_ `Timer` supports subtraction  and comparisons that would simplify this code (but @ravi-mosaicml is the expert here).

Also, you don't want to use `batch_in_epoch` -- what if patience is `100ba`, but that can span an epoch boundary?

composer/callbacks/early_stopper.py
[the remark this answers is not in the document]
ravi-mosaicml: Since metrics are lazily evaluated, I don't think it will be possible to do this check earlier right now. But with the metrics refactor and/or the restoration of the evaluators on the state, this should be possible.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+        ... )
+
+    Args:
+        monitor (str): The name of the metric to monitor.
+        dataloader_label (str): The label of the dataloader or evaluator associated with the tracked metric. If 
+            monitor is in an Evaluator, the dataloader_label field should be set to the Evaluator's label. If 
+            monitor is a training metric or an ordinary evaluation metric not in an Evaluator, dataloader_label
+            should be set to 'train' or 'eval' respectively.
+        comp (Callable[[Any, Any], bool], optional): A comparison operator to measure change of the monitored metric. The comparison
+            operator will be called `comp(current_value, prev_best)`. For metrics where the optimal value is low
```
ravi-mosaicml, line 55:
```suggestion
            operator will be called ``comp(current_value, prev_best)``. For metrics where the optimal value is low
```

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+            (error, loss, perplexity), use a less than operator and for metrics like accuracy where the optimal value
+            is higher, use a greater than operator. Defaults to numpy.less if loss, error, or perplexity are substrings
```
ravi-mosaicml, line 57:
```suggestion
            is higher, use a greater than operator. Defaults to :func:`numpy.less` if loss, error, or perplexity are substrings
```

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+            of the monitored metric, otherwise defaults to numpy.greater
```
ravi-mosaicml, line 58:
```suggestion
            of the monitored metric, otherwise defaults to :func:`numpy.greater`.
```

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+        ...     callbacks=[early_stopper],
+        ... )
+
+    Args:
+        monitor (str): The name of the metric to monitor.
+        dataloader_label (str): The label of the dataloader or evaluator associated with the tracked metric. If 
+            monitor is in an Evaluator, the dataloader_label field should be set to the Evaluator's label. If 
+            monitor is a training metric or an ordinary evaluation metric not in an Evaluator, dataloader_label
+            should be set to 'train' or 'eval' respectively.
+        comp (Callable[[Any, Any], bool], optional): A comparison operator to measure change of the monitored metric. The comparison
```
ravi-mosaicml, line 54: Is it possible to use torch.less / torch.greater be used instead of numpy (in case if metric tensors are on GPU?)

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+            operator will be called `comp(current_value, prev_best)`. For metrics where the optimal value is low
+            (error, loss, perplexity), use a less than operator and for metrics like accuracy where the optimal value
+            is higher, use a greater than operator. Defaults to numpy.less if loss, error, or perplexity are substrings
+            of the monitored metric, otherwise defaults to numpy.greater
+        min_delta (float, optional): An optional float that requires a new value to exceed the best value by at
+            least that amount. Defaults to 0.
+        patience (int | str | Time, optional): The interval of time the monitored metric can not improve without stopping
```
ravi-mosaicml, line 61: Can you mention that if patience is an integer, then it will be interpreted in units of epochs?

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+        self.monitor = monitor
+        self.dataloader_label = dataloader_label
+        self.comp = comp
+        self.min_delta = abs(min_delta)
+        if self.comp is None:
+            if any(substr in monitor.lower() for substr in ["loss", "error", "perplexity"]):
+                self.comp = np.less
+            else:
+                self.comp = np.greater
+
```
ravi-mosaicml, line 86: Nit: if `self.comp` is set here, then pyright will realize that is is never None, so the assert check on line 109 can be removed.

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+        if self.dataloader_label in state.current_metrics:
+            if self.monitor in state.current_metrics[self.dataloader_label]:
+                return state.current_metrics[self.dataloader_label][self.monitor]
+        raise ValueError(f"Couldn't find the metric {self.monitor} with the dataloader label {self.dataloader_label}."
+                         "Check that the dataloader_label is set to 'eval', 'train' or the evaluator name.")
+
+    def _update_stopper_state(self, state: State):
+        metric_val = self._get_monitored_metric(state)
+
+        assert self.comp is not None
```
ravi-mosaicml, line 109: Nit: this check can be removed -- see above comment

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+        if self.best is None:
+            self.best = metric_val
+            self.best_occurred = state.timer.get_timestamp()
+        elif self.comp(metric_val, self.best) and abs(metric_val - self.best) > self.min_delta:
```
ravi-mosaicml, line 113: Does this need to be torch.abs? See comment re: gpu

```suggestion
        elif self.comp(metric_val, self.best) and torch.abs(metric_val - self.best) > self.min_delta:
```

composer/callbacks/early_stopper.py
```diff
@@ -0,0 +1,139 @@
...
+            if state.timer.epoch - self.best_occurred.epoch > self.patience:
+                state.max_duration = state.timer.batch
+        elif self.patience.unit == TimeUnit.BATCH:
+            if state.timer.batch - self.best_occurred.batch > self.patience:
+                state.max_duration = state.timer.batch
+        else:
+            raise ValueError(f"The units of `patience` should be EPOCH or BATCH.")
+
+    def eval_end(self, state: State, logger: Logger) -> None:
+        if self.dataloader_label != "train":
```
ravi-mosaicml, line 128: For these checks (and the ones below), is it possible to compare `self.dataloader_label` against `state.dataloader_label`? The dataloader label on the state will be set to the currently active dataloader.

(I don't think this check would work if using multiple evaluators, and the train dataloader is not guaranteed to be called 'train')

tests/callbacks/test_early_stopper.py
```diff
@@ -0,0 +1,72 @@
...
+            self._update_metrics(state)
+
+    def batch_end(self, state: State, logger: Logger) -> None:
+        if self.unit == TimeUnit.BATCH:
+            self._update_metrics(state)
+
+
+@pytest.mark.parametrize('metric_sequence', [[0.1, 0.2, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1, 1.2, 1.3], [0.1, 0.2]])
+@pytest.mark.parametrize('unit', [TimeUnit.EPOCH, TimeUnit.BATCH])
+def test_threshold_stopper(metric_sequence: List[float], unit: TimeUnit):
```
ravi-mosaicml, line 45: Is it also possible to parameterize this test for cpu and gpu?

tests/callbacks/test_early_stopper.py
```diff
@@ -0,0 +1,72 @@
...
+        self.monitor = monitor
+        self.dataloader_label = dataloader_label
+        self.metric_sequence = metric_sequence
+        self.unit = unit
+
+    def _update_metrics(self, state: State):
+        idx = min(len(self.metric_sequence) - 1, state.timer.get(self.unit).value)
+        metric_val = self.metric_sequence[idx]
+        state.current_metrics[self.dataloader_label] = state.current_metrics.get(self.dataloader_label, dict())
+        state.current_metrics[self.dataloader_label][self.monitor] = tensor(metric_val)
```
ravi-mosaicml, line 28: (See below comment on gpu tests), but can you create this tensor on the same device as `next(torch.model.parameters()).device`? I believe that a gpu tensor will be stored in state, if using gpu. Would like to test this codepath.

ravi-mosaicml (commented):
Overall looks good! Main comments re: dataloader labels and gpu test coverage. After that should be good to merge.

composer/callbacks/early_stopper.py
anisehsani: I think the checks will work with multiple evaluators, unless they label one of their evaluators as "train". That test doesn't select for which dataloader metrics to use, it just tests whether the user is trying to get a training metric or an eval metric to decide which event to run on. Could you correct if I'm wrong?

In multiple places, docs/code say that the training dataloader will be labelled 'train', could you elaborate?

ravi-mosaicml (commented):
LGTM!

PR #986, opened 2022-05-03
Remove `pyright` from precommit hoook

hanlint:
`pyright` currently emits a verbose logs on failure, making the dev workflow difficult to use. Instead, this PR:
1. Removes `pyright` from the precommit hook
2. Restores `make style` and `make lint` so users can manually invoke
3. edits the CONTRIBUTING file to make the use of precommits optional.

The precommit hook will still handle auto-formatting for the user, which is a significant dev workflow improvement.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 default_language_version:
-    python: python3
+  python: python3
 repos:
--   repo: https://github.com/google/yapf
+  - repo: https://github.com/google/yapf
     rev: v0.32.0
     hooks:
-    -   id: yapf
+      - id: yapf
         name: yapf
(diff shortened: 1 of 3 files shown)


ravi-mosaicml (changes requested):
Jenkins uses pre-commit to validate files, so would like to leave pyright in there. I'm working on a separate PR to reduce the verbosity of pyright warnings, so the errors will be clearly visible. It'll still be a bad ux to push a PR, only to have it rejected for Pyright. I think this should fix the usability issues with the pre-commit hooks?

pre-commit handles the versioning and cli args for yapf, pyright, and isort, so I would rather not have these versions duplicated via a make command and the pre commit config file.

hanlint:
Agreed, closing this.

PR #987, opened 2022-05-03
Reduce PyRight Verbosity; Fix some errors

ravi-mosaicml:
* Disabled warnings for `reportIncompatibleMethodOverride` since that happens intentionally with `forward()` and `initialize_object()`
* Marked `reportUnusedClass` and `reportUnusedFunction` as errors, as they find dead code. Our codebase was (mostly) complaint with these checks; removed the dead code that these checks found.
* Disabled `reportPrivateImportUsage` since PyRight incorrectly marks some public imports as private (like with torchmetrics)
* Disabled `reportMissingParameterType` until we have proper type annotations in tests
* Upgrading `reportCallInDefaultInitializer` to `error` as it is indeed a bug. Fixed the few instances where we did this.
* Disabled `reportImplicitStringConcatenation` until we fix later.
* Disabled `reportUnnecessaryIsInstance` as we sometimes want extra validation. 
* Fixed up other codes to match these pyright config changes
* Not using the pyright warning setting anymore; this will prevent the warnings from blowing up.

Any pyright output should now be limited to new errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,7 @@ repos:
         language: node
         types: [python]
         pass_filenames: false
+        args: [--warnings]
         additional_dependencies: ["pyright@1.1.224"]
 # -   repo: https://github.com/PyCQA/pydocstyle
 #     hooks:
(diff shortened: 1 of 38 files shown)


hanlint (approved):
As usual, the pyright master strikes again.

PR #992, opened 2022-05-04
Synthetic HF State Draft

jzf2101:
[`minimal_state`](https://github.com/mosaicml/composer/blob/dev/tests/fixtures/new_fixtures.py#L17-L28) uses a MLP.  It would be useful to also include a minimal state for transformers and transformer data, dataloaders, and tokenizers.  This is a draft in this direction.  The configurations to test are based on the configurations listed in [`test_synthetic_lm_data.generate_parameter_configs`](https://github.com/mosaicml/composer/blob/d100053198524672f628c3959a8c4e51a9302e2d/tests/datasets/test_synthetic_lm_data.py#L8-L24).

The tests are used to demonstrate that the code works for all the configs in `test_synthetic_lm_data.generate_parameter_configs`.  Presumably, this would be turned into an eventual fixture used for testing other methods on LMs.


tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,78 @@
...
+    return tokenizer
+
+def make_dummy_lm(model_name: str, max_position_embeddings: int, tokenizer):
+    pytest.importorskip("transformers")
+    if model_name == 'gpt2':
+        class_name = GPT2Hparams
+    elif model_name == 'bert':
+        class_name = BERTHparams
+    model_config = generate_dummy_model_config(class_name, tokenizer)
+    if model_name == 'bert':
```
moinnadeem, line 33: @jzf2101 Should we just fold these into the `generate_dummy_model_config` function?

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,78 @@
...
+        model_config['num_labels'] = model_config['vocab_size']
+        model_config['max_position_embeddings'] = max_position_embeddings
+    model = class_name(model_config=model_config).initialize_object()
+    model.eval()
+    return model
+
+def synthetic_to_dataloader(dataset_config: dict):
+    """
+    """
+    dataloader = LMDatasetHparams(use_synthetic=True, tokenizer_name=dataset_config['tokenizer_family'], use_masked_lm=dataset_config['use_masked_lm'], split='train', max_seq_length=dataset_config["chars_per_sample"])
```
moinnadeem, line 43: Can you format this file? I think the indentation is off.

moinnadeem (approved):
Overall, lgtm! Might defer to someone with more testing experience such as @ravi-mosaicml

tests/fixtures/synthetic_hf_state.py
jzf2101: It looks like I was being overly detailed. [`num_labels` is set when calling `BERTHparams` to `vocab_size`](https://github.com/mosaicml/composer/blob/23f888e07456d31a2950db9f962f4dbdf41938fb/composer/models/bert/bert_hparams.py#L114)

jzf2101:
Pyright errors I can't figure out how to fix:


`next(iter(state.dataloader)).data`
```
  /Users/jforde/composer/tests/fixtures/synthetic_hf_state.py:69:29 - error: Argument of type "Iterable[Unknown] | None" cannot be assigned to parameter "__iterable" of type "SupportsIter[_SupportsNextT@iter]" in function "iter"
    Type "Iterable[Unknown] | None" cannot be assigned to type "SupportsIter[_SupportsNextT@iter]"
      "__iter__" is not present (reportGeneralTypeIssues)
```

`state_output['logits'].size() == logits.size()`

```
  /Users/jforde/composer/tests/fixtures/synthetic_hf_state.py:87:56 - error: Cannot access member "size" for type "Mapping[Unknown, Unknown]"
```

tests/fixtures/synthetic_hf_state.py
jzf2101: This issue is handled in #1023

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
+from collections.abc import Iterable
+
+import pytest
+import torch
+from transformers import PreTrainedTokenizer
+
+from composer.core.state import State
+from composer.datasets.dataloader import DataLoaderHparams
+from composer.datasets.lm_datasets import LMDatasetHparams
```
hanlint, line 9: nit: I think these can be combined into one line

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+    class_name = TransformerHparams
+    if model_name == 'gpt2':
+        class_name = GPT2Hparams
+    elif model_name == 'bert':
+        class_name = BERTHparams
+    else:
+        raise ValueError("Model name must be one of 'gpt2' or 'bert'")
+    model_config = generate_dummy_model_config(class_name, tokenizer)
+    model_config['max_position_embeddings'] = max_position_embeddings
+    model = class_name(model_config=model_config).initialize_object()
```
hanlint, line 47: would it be possible to not use `Hparams` objects here, and instantiate the models directly?

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+    model.eval()
```
hanlint, line 48: this seems extraneous and undocumented part of `make_dummy_lm`

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+def make_lm_tokenizer(config: dict):
+    dataset = synthetic_hf_dataset_builder(num_samples=config['num_samples'],
+                                           chars_per_sample=config['chars_per_sample'],
+                                           column_names=config['column_names'])
+    tokenizer = generate_synthetic_tokenizer(config['tokenizer_family'], dataset)
+    return tokenizer
+
+
+def make_dummy_lm(model_name: str, max_position_embeddings: int, tokenizer: PreTrainedTokenizer):
+    pytest.importorskip("transformers")
```
hanlint, line 37: perhaps leave pytest skipping control to the invoking test? probably don't want a helper function silently skipping a test.

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        raise ValueError("Model name must be one of 'gpt2' or 'bert'")
+    model_config = generate_dummy_model_config(class_name, tokenizer)
+    model_config['max_position_embeddings'] = max_position_embeddings
+    model = class_name(model_config=model_config).initialize_object()
+    model.eval()
+    return model
+
+
+def synthetic_to_dataloader(dataset_config: dict):
```
hanlint, line 52: nit: perhaps `make_synthetic_dataloader` would be more consistent with above?

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+                                  use_masked_lm=dataset_config['use_masked_lm'],
+                                  max_seq_length=dataset_config["chars_per_sample"],
+                                  split='train')
+    dataloader = dataloader.initialize_object(batch_size=dataset_config['num_samples'],
+                                              dataloader_hparams=DataLoaderHparams(num_workers=0,
+                                                                                   persistent_workers=False))
+    return dataloader
+
+
+def synthetic_hf_state(model: ComposerTransformer, dataloader: Iterable, rank_zero_seed: int = 0):
```
hanlint, line 66: should this have a pytest fixture decorator?

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+        dataloader=dataloader,
+        dataloader_label="train",
+        max_duration='1ep',
+    )
+    assert isinstance(state.dataloader, Iterable)
+    state.batch = next(iter(state.dataloader)).data
+    return state
+
+
+@pytest.mark.parametrize("config", make_dataset_configs())
```
hanlint, line 82: AFAIK, `pytest` collection would not pick this up, as the file name does not start with `test_*`. Perhaps move this to a different file.

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,100 @@
...
+def test_synthetic_hf_state(config: dict):
```
hanlint, line 83: for readability, consider (1) adding blank lines to separate out concepts, (2) moving `sample` to after line 88 since it's used for the assert, not for the state creation on line 88.

hanlint (commented):
LGTM, a few comments

tests/fixtures/synthetic_hf_state.py
jzf2101: I've been able to collect it so far?

tests/fixtures/synthetic_hf_state.py
jzf2101: so you want to combine all the calls to `composer.datasets` instead of splitting them out into sub-sub folder import calls?

tests/fixtures/synthetic_hf_state.py
jzf2101: I originally had tried to do that but was unable to get over the quirks of each configuration and setup with HF.  @moinnadeem do you have any ideas on how to do this?

tests/fixtures/synthetic_hf_state.py
jzf2101: Moved this into the testing part

tests/fixtures/synthetic_hf_state.py
hanlint: ok, this is more of a nit so not necessary for merging this PR if it's too complicated

ravi-mosaicml (commented):
I agree with the existing comments here ... the real test will be when other tests use this :)

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+#from collections.abc import Iterable
+
+import pytest
```
ravi-mosaicml, line 6: Can the fixtures in this file be moved into the test_sysnthetic_state.py? Would prefer not to add global fixtures that are used by just one test module.

jzf2101: we could - though I believe that reverses the feedback from @hanlint ?

hanlint: Ah to clarify, my feedback was to move the test to a `test_...` file so its picked up by pytest. I'm also in favor of moving the fixtures into the `test_..` file as well.

tests/fixtures/synthetic_hf_state.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+#from collections.abc import Iterable
```
hanlint, line 4:
```suggestion
```

tests/test_synthetic_hf_state.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+#import pdb
+
```
hanlint, line 5:
```suggestion
```

tests/fixtures/test_synthetic_hf_state.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Tuple
```
ravi-mosaicml, line 4: Sorry, I think I misunderstood earlier about what this PR is accomplishing. Looking at it more closely, I do think the original split makes sense -- to put the `synthetic_hf_state` fixture in its own file (inside `tests/fixtures`), and the `test_synthetic_hf_state()` test in a separate file (outside `tests/fixtures`). I don't have a preference on whether `synthetic_hf_state` lives in `tests/fixtures/new_fixtures.py` or its own file.

tests/fixtures/test_synthetic_hf_state.py
```diff
@@ -0,0 +1,118 @@
...
+    else:
+        raise ValueError("Model name must be one of 'gpt2' or 'bert'")
+    model_config = generate_dummy_model_config(class_name, tokenizer)
+    model_config['max_position_embeddings'] = max_position_embeddings
+    model = class_name(model_config=model_config).initialize_object()
+    return model
+
+
+def make_synthetic_dataloader(dataset_config: dict):
+    """
```
ravi-mosaicml, line 57: nit: remove the empty docstring?

tests/fixtures/test_synthetic_hf_state.py
```diff
@@ -0,0 +1,118 @@
...
+    model = make_dummy_lm(config['tokenizer_family'], config['chars_per_sample'], tokenizer)
+    dataloader = make_synthetic_dataloader(config)
+    return model, dataloader
+
+
+def synthetic_hf_state_maker(config) -> Tuple:
+    """
+    An example state using synthetic HF transformer function which could used for testing purposes
+    """
+    #config = request.params
```
ravi-mosaicml, line 81: nit: remove line

tests/fixtures/test_synthetic_hf_state.py
```diff
@@ -0,0 +1,118 @@
...
+        dataloader=dataloader,
+        dataloader_label="train",
+        max_duration='1ep',
+    )
+
+    return state, model, dataloader
+
+
+@pytest.fixture(params=make_dataset_configs())
+def synthetic_hf_state_fixture(request):
```
ravi-mosaicml, line 95: No need to include `fixture` in the fixture name

```suggestion
def synthetic_hf_state(request):
```

ravi-mosaicml (approved):
Overall LGTM; see comments!

ravi-mosaicml (approved):
LGTM! :rocket:

PR #995, opened 2022-05-04
Allow partially populated single rank distributed variables by default

ravi-mosaicml:
* If any distributed variables are set, and they match the single rank defaults, then populate the remaining single rank defaults. Before, if any variables were set, then the reset would not be populated, even if they matched the single rank defaults. An MCLI change that set `LOCAL_WORLD_SIZE` without the other variables broke running single-rank gpu tests.
* Also removed a warning about configuring distributed with just one rank, as this is expected behavior when using DeepSpeed.

Tested this change manually by running:

```python3
import os
from datetime import timedelta
import composer.utils.dist

os.environ['LOCAL_RANK'] = '0'

composer.utils.dist.initialize_dist('gloo', timedelta(seconds=5))
```

(Cannot unit test this, as distributed is configured globally for all unit tests, due to a PyTorch bug with destroying distributed process groups.)

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -33,8 +33,6 @@ from __future__ import annotations
 
 import datetime
 import os
-import textwrap
-import warnings
 from typing import Any, List, Optional, Sequence, TypeVar, cast
 
 import torch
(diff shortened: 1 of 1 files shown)


jbloxham (approved):
I'd prefer if `LOCAL_WORLD_SIZE` were singled out as the only env var that can appear without all the others being provided. But I'm fine with this, too.

ravi-mosaicml:
> I'd prefer if `LOCAL_WORLD_SIZE` were singled out as the only env var that can appear without all the others being provided. But I'm fine with this, too.

Discussed offline; will leave as-is for now to keep the code path simpler.

PR #997, opened 2022-05-05
MosaicML StreamingDataset (Part 2, vision datasets)

abhi-mosaic:
This PR adds implementations of Vision datasets using the new `StreamingDataset` class.

Specifically we will add streaming versions of ADE20k, ImageNet1k, and COCO, with more to come later.

Dataset writing scripts are kept out of the Composer core, as they are meant to be used when you have a local copy of the dataset available and are converting it to the `StreamingDataset` format.

### DIFF
--- a/composer/datasets/ade20k.py
+++ b/composer/datasets/ade20k.py
@@ -412,10 +412,10 @@ class StreamingADE20k(StreamingDataset):
         return data.decode('utf-8')
 
     def decode_image(self, data: bytes) -> Image.Image:
-        return Image.open(BytesIO(data)).convert('RGB')
+        return Image.open(BytesIO(data))
 
     def decode_annotation(self, data: bytes) -> Image.Image:
-        return Image.open(BytesIO(data)).convert('RGB')
+        return Image.open(BytesIO(data))
 
     def __init__(self,
                  remote: str,
--- a/composer/datasets/ade20k.py
+++ b/composer/datasets/ade20k.py
@@ -399,21 +399,22 @@ class StreamingADE20k(StreamingDataset):
     Args:
         remote (str): Remote directory (S3 or local filesystem) where dataset is stored.
         local (str): Local filesystem directory where dataset is cached during operation.
-        split (str): The dataset split to use, either 'train' or 'val'. Default: ``'train```.
+        split (str): The dataset split to use, either 'train' or 'val'.
         shuffle (bool): Whether to shuffle the samples in this dataset.
         base_size (int): initial size of the image and target before other augmentations. Default: ``512``.
         min_resize_scale (float): the minimum value the samples can be rescaled. Default: ``0.5``.
         max_resize_scale (float): the maximum value the samples can be rescaled. Default: ``2.0``.
         final_size (int): the final size of the image and target. Default: ``512``.
+        batch_size (Optional[int]): Hint the batch_size that will be used on each device's DataLoader. Default: ``None``.
     """
 
-    def decode_uid(self, data: bytes) -> Any:
+    def decode_uid(self, data: bytes) -> str:
         return data.decode('utf-8')
 
-    def decode_image(self, data: bytes) -> Any:
+    def decode_image(self, data: bytes) -> Image.Image:
         return Image.open(BytesIO(data))
 
-    def decode_annotation(self, data: bytes) -> Any:
+    def decode_annotation(self, data: bytes) -> Image.Image:
         return Image.open(BytesIO(data))
 
     def __init__(self,
@@ -481,7 +482,7 @@ class StreamingADE20k(StreamingDataset):
             self.annotation_transform = transforms.Resize(size=(final_size, final_size),
                                                           interpolation=TF.InterpolationMode.NEAREST)
 
-    def __getitem__(self, idx: int) -> Any:
+    def __getitem__(self, idx: int) -> Tuple[Any, Any]:
         obj = super().__getitem__(idx)
         x = obj['image']
         y = obj['annotation']
@@ -496,12 +497,14 @@ class StreamingADE20k(StreamingDataset):
 
 @dataclass
 class StreamingADE20kHparams(DatasetHparams):
-    """Streaming ADE20k hyperparameters.
+    """DatasetHparams for creating an instance of StreamingADE20k.
 
     Args:
         remote (str): Remote directory (S3 or local filesystem) where dataset is stored.
+            Default: ``'s3://mosaicml-internal-dataset-ade20k/md/```
         local (str): Local filesystem directory where dataset is cached during operation.
-        split (str): the dataset split to use, either 'train' or 'val'. Default: ``'train```.
+            Default: ``'/tmp/mds-cache/mds-ade20k/```
+        split (str): The dataset split to use, either 'train' or 'val'. Default: ``'train```.
         base_size (int): initial size of the image and target before other augmentations. Default: ``512``.
         min_resize_scale (float): the minimum value the samples can be rescaled. Default: ``0.5``.
         max_resize_scale (float): the maximum value the samples can be rescaled. Default: ``2.0``.
@@ -510,12 +513,11 @@ class StreamingADE20kHparams(DatasetHparams):
             Default: ``true``.
     """
 
-    remote: str = hp.optional('Remote directory (S3 or local filesystem) where dataset is stored',
-                              default='s3://mosaicml-internal-dataset-ade20k/mds/')
-    local: str = hp.optional('Local filesystem directory where dataset is cached during operation',
-                             default='/tmp/mds-cache/mds-ade20k/')
+    remote: str = hp.optional("Remote directory (S3 or local filesystem) where dataset is stored",
+                              default="s3://mosaicml-internal-dataset-ade20k/mds/")
+    local: str = hp.optional("Local filesystem directory where dataset is cached during operation",
+                             default="/tmp/mds-cache/mds-ade20k/")
     split: str = hp.optional("Which split of the dataset to use. Either ['train', 'val']", default='train')
-
     base_size: int = hp.optional("Initial size of the image and target before other augmentations", default=512)
     min_resize_scale: float = hp.optional("Minimum value that the image and target can be scaled", default=0.5)
     max_resize_scale: float = hp.optional("Maximum value that the image and target can be scaled", default=2.0)
@@ -532,7 +534,6 @@ class StreamingADE20kHparams(DatasetHparams):
                                   max_resize_scale=self.max_resize_scale,
                                   final_size=self.final_size,
                                   batch_size=batch_size)
-
         collate_fn = pil_image_collate
         device_transform_fn = NormalizationFn(mean=IMAGENET_CHANNEL_MEAN,
                                               std=IMAGENET_CHANNEL_STD,
--- a/composer/datasets/ade20k.py
+++ b/composer/datasets/ade20k.py
@@ -11,7 +11,7 @@ import os
 from dataclasses import dataclass
 from io import BytesIO
 from math import ceil
-from typing import Any, Callable, Optional, Tuple, Union
+from typing import Any, Optional, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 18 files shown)


composer/datasets/coco.py
```diff
@@ -177,3 +181,102 @@ def split_dict_fn(batch: Batch, num_microbatches: int) -> Sequence[Batch]:  #typ
...
+        local (str): Local filesystem directory where dataset is cached during operation.
+    """
+
+    remote: str = hp.optional('Remote directory (S3 or local filesystem) where dataset is stored',
+                              default='s3://mosaicml-internal-dataset-coco/mds/')
+    local: str = hp.optional('Local filesystem directory where dataset is cached during operation',
+                             default='/tmp/mds-cache/mds-coco/')
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHparams):
+        dboxes = dboxes300_coco()
```
hanlint, line 267: same comment -> move this to StreamingCOCO

composer/datasets/ade20k.py
```diff
@@ -388,6 +391,136 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+            raise ValueError("base_size cannot be zero or negative.")
+
+        if self.min_resize_scale <= 0:
+            raise ValueError("min_resize_scale cannot be zero or negative")
+
+        if self.max_resize_scale < self.min_resize_scale:
+            raise ValueError("max_resize_scale cannot be less than min_resize_scale")
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHparams) -> DataSpec:
+        self.validate()
```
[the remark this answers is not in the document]
abhi-mosaic: Yes can do... I have a question though, do we want this preprocessing functionality like `split`, `base_size`, `min_resize_scale`, ... to all be part of our `StreamingADE20k` dataset? I'm in favor of this, just want to emphasize that now it won't just be an ADE20k dataset that returns PIL images, it's a dataset with some built in user-chosen preprocessing steps. So less reusable code, but more clearly an _example_ of how a user constructs their custom dataset.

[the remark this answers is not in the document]
abhi-mosaic: Also, dataloader construction will still live in `initialize_object` by necessity, because `dataloader_hparams` and `batch_size` are passed in via the Trainer

composer/datasets/ade20k.py
```diff
@@ -388,6 +391,155 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+    Args:
+        remote (str): Remote directory (S3 or local filesystem) where dataset is stored.
+        local (str): Local filesystem directory where dataset is cached during operation.
+        split (str): The dataset split to use, either 'train', or 'val'. Default: ``'train```.
+        shuffle (bool): Whether to shuffle the samples in this dataset.
+        base_size (int): initial size of the image and target before other augmentations. Default: ``512``.
+        min_resize_scale (float): the minimum value the samples can be rescaled. Default: ``0.5``.
+        max_resize_scale (float): the maximum value the samples can be rescaled. Default: ``2.0``.
+        final_size (int): the final size of the image and target. Default: ``512``.
+    """
```
abhi-mosaic, line 407: Refactored so that the `StreamingADE20k` dataset does most of the work, making it easier to use in a standalone way.

Collating and device normalization are still done in the hparams `initialize_object` when we create the dataloader.

@hanlint Can I avoid refactoring the older `ImageFolder`-based ADE20k dataset in this PR? I an open a JIRA issue to track. Would prefer to make this PR simple and additive, and do edits/deletions in a separate one.

scripts/mds/ade20k.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        args (Namespace): Commandline arguments.
+    """
+    fields = ['uid', 'image', 'annotation']
+
+    for (split, expected_num_samples, shuffle) in [
+        ("train", 20206, True),
+        ("val", 2000, False),
+    ]:
```
abhi-mosaic, line 83: Only uploaded the `train` and `val` splits at the moment, as the `test` split doesn't have labels and I'm not sure how we would use it... if @Landanjs has a use case I can add support, there will be some extra logic on the Dataset init to make sure we don't try to decode keys that don't exist.

tests/datasets/test_streaming_remote.py
```diff
@@ -0,0 +1,107 @@
...
+
+import pytest
+from PIL.Image import Image
+from torch.utils.data import DataLoader
+
+from composer.datasets.ade20k import StreamingADE20k
+from composer.datasets.utils import pil_image_collate
+
+
+# @pytest.mark.skip()
```
abhi-mosaic, line 12: @ravi-mosaicml could you provide me a quick example of how to add that @pytest.mark.integration thing? From what I understood I should add **both** the integration flag and the skip(), with the latter being removed once we have later updated our Jenkins workflow to recognize the integration.

tests/datasets/test_streaming_remote.py
```diff
@@ -0,0 +1,107 @@
...
+    for epoch in range(3):
+        rcvd_samples = 0
+        epoch_start = time.time()
+        for _, (images, _) in enumerate(loader):
+            n_samples = images.shape[0]
+            rcvd_samples += n_samples
+        epoch_end = time.time()
+        epoch_dur = epoch_end - epoch_start
+        samples_per_sec = rcvd_samples / epoch_dur
+        print(f"Epoch {epoch}: epoch_dur={epoch_dur:.2f}s, samples_per_sec={samples_per_sec:.2f}")
```
abhi-mosaic, line 104: Running this test with `pytest -s ...` on my Macbook Pro, with a 300Mbps connection, I see speeds like:
```
tests/datasets/test_streaming_remote.py ds_build_dur=1.45s, loader_build_dur=0.00s
Epoch 0: epoch_dur=24.28s, samples_per_sec=82.38
Epoch 1: epoch_dur=5.54s, samples_per_sec=361.02
Epoch 2: epoch_dur=5.83s, samples_per_sec=343.22
```

Our DeepLabV3 benchmark only consumes data at a rate of ~20 img/sec per A100, so if my laptop can support 80+ img/sec on epoch0 and 300+ img/sec on epoch1, I think we are in a good spot :)

abhi-mosaic (commented):
@ravi-mosaicml @hanlint Went through a refactor of the ADE20k dataset and would love to hear your thoughts! After feedback I will basically do the same for ImageNet and COCO (so don't read those yet).

The 3 files to look at are:
* `composer/datasets/ade20k.py`
* `scripts/mds/ade20k.py`
* `tests/datasets/test_streaming_remote.py`

Going to start doing E2E convergence test in parallel as well!

composer/datasets/ade20k.py
```diff
@@ -388,6 +391,155 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+    def decode_annotation(self, data: bytes) -> Any:
+        return Image.open(BytesIO(data))
+
+    def __init__(self,
+                 remote: str,
+                 local: str,
+                 split: str,
+                 shuffle: bool,
+                 base_size: int = 512,
+                 min_resize_scale: float = 0.5,
```
ravi-mosaicml, line 424: Should the dataset class be opinionated about the transformations? (Realize you likely made this change given the previous feedback on the PR!) If the answer is no, then OK to leave this as is and clean up later -- potentially once torch data pipes come out.

tests/datasets/test_streaming_remote.py
ravi-mosaicml: Can you add in the integration marker in this PR? Here's how:

1. Modify `pyproject.toml` -- it'll need to be specified on line 58 (e.g. `and not integration`) and in the list on line 63

  https://github.com/mosaicml/composer/blob/23f888e07456d31a2950db9f962f4dbdf41938fb/pyproject.toml#L58-L64

2. In `meta.yaml`, modify the existing markers there to include `and not integration`

  https://github.com/mosaicml/composer/blob/97fbab157ebd8c2488f5a00a806e64639df85d19/meta.yaml#L82-L83

3. In `.ci/Jenkinsfile`, modify the markers on line 194 to also include `and not integration`

https://github.com/mosaicml/composer/blob/97fbab157ebd8c2488f5a00a806e64639df85d19/.ci/Jenkinsfile#L194

Then, the `@pytest.mark.skip` will not be needed.

ravi-mosaicml (approved):
LGTM -- didn't give this PR a close review since it is implementation specific, but see comments.

scripts/mds/ade20k.py
Landanjs: Results on the test set require submitting to a leaderboard. This is something we should do at some point, but let me check if this should be a requirement for numbers on composer

composer/datasets/ade20k.py
abhi-mosaic: I think we want to stay opinionated (and yes that was based on comments earlier in the PR :), as I think our goal is not to build an ADE20k for someone to re-use (other than basic functionality), rather, ADE20k is an example that a user will look at in order to build their own custom datasets, with custom processing.

composer/datasets/ade20k.py
```diff
@@ -388,6 +391,136 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+        min_resize_scale (float): the minimum value the samples can be rescaled. Default: ``0.5``.
+        max_resize_scale (float): the maximum value the samples can be rescaled. Default: ``2.0``.
+        final_size (int): the final size of the image and target. Default: ``512``.
+        ignore_background (bool): if true, ignore the background class when calculating the training loss.
+            Default: ``true``.
+    """
+
+    remote: str = hp.optional('Remote directory (S3 or local filesystem) where dataset is stored',
+                              default='s3://mosaicml-internal-dataset-ade20k/mds/')
+    local: str = hp.optional('Local filesystem directory where dataset is cached during operation',
```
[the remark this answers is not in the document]
abhi-mosaic: Hmm would you like the defaulting to occur in the Dataset object or the DatasetHparams ?

composer/datasets/ade20k.py
```diff
@@ -388,6 +391,136 @@ def initialize_object(self, batch_size, dataloader_hparams) -> DataSpec:
...
+        split (str): the dataset split to use either 'train', 'val', or 'test'. Default: ``'train```.
+        base_size (int): initial size of the image and target before other augmentations. Default: ``512``.
+        min_resize_scale (float): the minimum value the samples can be rescaled. Default: ``0.5``.
+        max_resize_scale (float): the maximum value the samples can be rescaled. Default: ``2.0``.
+        final_size (int): the final size of the image and target. Default: ``512``.
+        ignore_background (bool): if true, ignore the background class when calculating the training loss.
+            Default: ``true``.
+    """
+
+    remote: str = hp.optional('Remote directory (S3 or local filesystem) where dataset is stored',
```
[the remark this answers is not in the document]
abhi-mosaic: Right... would you prefer that I use a different default? Sadly I cannot remove the default as we need to use `hp.optionals` unless we clean up `DatasetHparams`. Removing these four args (which I honestly think we should do as it's a leaky abstraction) would fix it:


```
class DatasetHparams(hp.Hparams, abc.ABC, metaclass=metaclass):
    is_train: bool = hp.optional("Whether to load the training data (the default) or validation data.", default=True)
    drop_last: bool = hp.optional(textwrap.dedent("""\
        If the number of samples is not divisible by the batch size,
        whether to drop the last batch (the default) or pad the last batch with zeros."""),
                                  default=True)
    shuffle: bool = hp.optional("Whether to shuffle the dataset for each epoch. Defaults to True.", default=True)

    datadir: Optional[str] = hp.optional("The path to the data directory", default=None)
```

Alternatively I can make a whole new `StreamingDatasetHparams` that doesn't subclass `DatasetHparams` but it seems like a hack around the issue.

hanlint (approved):
LGTM, much cleaner, thank you!

composer/datasets/ade20k.py
hanlint: Yeah OK to limit your refactor to streaming datasets

scripts/mds/ade20k.py
abhi-mosaic: Discussed offline, will leave test set out for now.

tests/datasets/test_streaming_remote.py
abhi-mosaic: Done!

composer/datasets/ade20k.py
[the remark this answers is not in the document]
abhi-mosaic: Going to punt this to a `DatasetHparams` refactor, for now I'm just following what we did for the webdatasets.

abhi-mosaic:
Got tests working for both `ADE20k` and `ImageNet`, I had some issues with `COCO` though so I'm going to make a JIRA issue for that and `pytest.skip()` for now. 

Adding streaming YAMLs, rebasing, and will merge once Jenkins is happy :)

abhi-mosaic:
Some notes on ImageNet dataset performance:

On a 2xA100 node with high aggregate network bandwidth (~10Gbps):
- streaming dataloader with 8 workers/GPU can support ~3k img/sec/GPU, equivalent to 24k img/sec on an 8xA100 system, **even on epoch0**, and it gets a bit faster on epoch1+
- as expected, R50+ImageNet+channels_last on A100 **is no longer bottlenecked**, and we should have about 50% headroom for algorithms that speed up the training step

PR #1000, opened 2022-05-05
Fix `atexit()` shutdown procedure

ravi-mosaicml:
- Track whether `atexit` triggered `Engine._close()`, which indicates whether the Python process is shutting down. If so, do not run `Engine._close()` again via `Engine.__del__()`, as Python machinery (e.g. the ability to do conditional imports, which is done with the WandB callback) is destroyed between when `atexit` and `__del__()` fire when Python exits with an exception.
- Added a test case to verify this fix


siriuslee (approved):
LGTM, though I'm not sure I fully understand how this interacts with the WandB logger which seemed to trigger my error.

PR #1005, opened 2022-05-06
Allow the package name to be set via an env variable

ravi-mosaicml:
Allow the pip package name in `setup.py` to be specified by an environment variable, `COMPOSER_PACKAGE_NAME`. This change will allow for automatic pip publishing, as we use a different package name when publishing to https://test.pypi.org.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,7 +167,12 @@ composer_data_files = ["py.typed"]
 composer_data_files += package_files("composer", "yamls", ".yaml")
 composer_data_files += package_files("composer", "algorithms", ".json")
 
-setup(name=os.environ.get('COMPOSER_PACKAGE_NAME', "mosaicml"),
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -167,7 +167,7 @@ def package_files(prefix: str, directory: str, extension: str):
 composer_data_files += package_files("composer", "yamls", ".yaml")
 composer_data_files += package_files("composer", "algorithms", ".json")
 
-setup(name="mosaicml",
+setup(name=os.environ.get('COMPOSER_PACKAGE_NAME', "mosaicml"),
```
hanlint, line 170: would you want to print/log the package name so that automated builds/logs are easier to debug?

hanlint (approved):
LGTM, one comment

PR #1010, opened 2022-05-06
Fix serialization for stochastic depth

hanlint:
Follows the `torch.nn.Conv2d` pattern and uses nn.Module private methods to save and load each layer's individual generator state.


composer/algorithms/stochastic_depth/stochastic_layers.py
```diff
@@ -76,6 +76,19 @@ def _get_sample(self, device: int) -> torch.Tensor:
                                  use_same_gpu_seed=self.use_same_gpu_seed,
                                  use_same_depth_across_gpus=self.use_same_depth_across_gpus)
 
+    """ Special consideration to serialize the layer's rng state.
+    """
+
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
+        super()._save_to_state_dict(destination, prefix, keep_vars)
+        destination[prefix + 'rng_state'] = self.rand_generator.get_state()
```
Landanjs, line 84: I'm guessing the order doesn't matter here, but setting `destination` before calling `_save_to_state_dict` makes more sense to me

Landanjs (approved):
I think this looks good!

PR #1011, opened 2022-05-06
Jenkinsfile Fixes + Updated Docker Build Matrix

ravi-mosaicml:
* Updated the `Jenkinsfile` to run tests on merge commits. This change will allow us to not require each PR to be up to date with `dev` before merging, but rather, note which PRs may have broken dev so they can be reverted.
* Refactored the `Jenkinsfile` variables for readability
* Switched to using the new getDockerBuildMatrix
* Modified the conda file to trigger a conda build
* Added a helper script to autogenerate the docker build matrix
* Use the build matrix in the Jenkinsfile for the non-docker-build tests (before, complicated logic was used to determine the docker image names)
* Remove ubuntu 18.04 support. Nobody uses that anymore, nor does it really change anything. Keeping everything on ubuntu 20.04.
* Fixed bugs with tests that spawned more multiprocesses than cores.
* Hardcoding `getDockerBuildMatrix` and `isPathModified` into the Jenkinsfile -- makes tests faster by skipping an extra git clone, and makes updates easier (so there is no need to coordinate with the testing repo)


composer/datasets/glue.py
```diff
@@ -71,8 +69,6 @@ class GLUEHparams(DatasetHparams, SyntheticHparamsMixin):
     split: str = hp.optional("Whether to use 'train', 'validation' or 'test' split.", default=None)
     max_seq_length: int = hp.optional(
         default=256, doc='Optionally, the ability to set a custom sequence length for the training dataset.')
-    num_workers: int = hp.optional(default=8,
```
ravi-mosaicml, line 74: This was causing issues with test timeouts, as we only allocate 4 CPU cores. Instead, `dataloader_hparams` should be used, which contains this information.

docs/source/getting_started/installation.rst
```diff
@@ -49,24 +49,17 @@ For a developer install, clone directly:
...
-
-Our ``latest`` image has Ubuntu 20.04, Python 3.9, PyTorch 1.10, and CUDA 11.3 and has been tested to work with
-GPU-based instances on AWS, GCP, and Azure. ``Pillow-SIMD`` is installed by default in all images.
+To simplify the environment setup for Composer, we provide a set of convenient Docker Images.
+The ``Base`` flavor of Docker Images contains PyTorch pre-installed; the ``Vision`` flavor also includes OpenCV,
+MM Segmentation, and FFCV dependencies.
+
+
+.. include:: ../../../docker/README.md
+   :parser: myst_parser.sphinx_
```
bandish-shah, line 58: I don't recall the details but there was a dependency (maybe for webdatasets) that was preventing us from using the version of MyST that supported this.  Just want to call it out, I'm assuming if the build passes that hopefully means this is no longer an issue.

bandish-shah (approved):
LGTM, nice job with the refactor and cleanup it's much easier to see what's going on now.

docs/source/getting_started/installation.rst
ravi-mosaicml: Yup! That was resolved -- the issue there was webdatasets used `awscli` (which had its on docutils dependency) instead of `boto3` (which doesn't have a docutils dependency).

PR #1013, opened 2022-05-06
Load from wandb logger

mvpatel2000:
Adds support to load checkpoints from wandb logger. 

**Testing:**
Adding unit tests for this is a pain because you need to start saving and loading checkpoints from wandb, so you need to throw in wandb credentials and every time tests are run, we read and write from there. I spent a bit of time playing around with this, but I'm strongly in favor of not adding a unit test for this behavior. It's not ideal, but I think it's far better than constantly writing to wandb with each github action

To verify it works, I ran two runs which save and load from there. Run 1: 
```
composer -n 1 examples/run_composer_trainer.py -f composer/yamls/models/resnet56_cifar10.yaml 
        --train_dataset.cifar10.datadir /localdisk/CIFAR10/ 
        --val_dataset.cifar10.datadir /localdisk/CIFAR10/ 
        --loggers progress_bar wandb 
        --save_folder experiments  
        --save_overwrite True 
        --loggers.wandb.log_artifacts True 
        --loggers.wandb.entity mosaic-ml 
        --loggers.wandb.name ckpt-test 
        --loggers.wandb.project ckpt
```
which saves to wandb. Run 2: 
```
composer -n 1 examples/run_composer_trainer.py -f composer/yamls/models/resnet56_cifar10.yaml 
        --train_dataset.cifar10.datadir /localdisk/CIFAR10/ 
        --val_dataset.cifar10.datadir /localdisk/CIFAR10/ 
        --loggers progress_bar wandb 
        --save_folder experiments  
        --save_overwrite True 
        --loggers.wandb.log_artifacts True 
        --loggers.wandb.entity mosaic-ml 
        --loggers.wandb.name ckpt-test 
        --loggers.wandb.project ckpt
```
which loads from wandb and continues training. Run 2 uses a slightly different yaml which includes
```
load_logger_destination: 
  wandb:
    entity: mosaic-ml
    name: ckpt-test
    project: ckpt
```
The resulting output is that it continues training from the checkpoint! Woo 
image

### DIFF
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -144,10 +144,13 @@ class WandBLogger(LoggerDestination):
         with tempfile.TemporaryDirectory() as tmpdir:
             artifact_folder = os.path.join(tmpdir, "artifact_folder")
             artifact.download(root=artifact_folder)
+            artifact_names = os.listdir(artifact_folder)
             # We only log one file per artifact
-            checkpoint_name = os.listdir(artifact_folder)[0]
-            checkpoint_path = os.path.join(artifact_folder, checkpoint_name)
-            shutil.move(checkpoint_path, destination)
+            if len(artifact_names) > 1:
+                raise RuntimeError("Found more than one file in artifact. We assume the checkpoint is the only file in the artifact.")
+            artifact_name = artifact_names[0]
+            artifact_path = os.path.join(artifact_folder, artifact_name)
+            shutil.move(artifact_path, destination)
 
     def post_close(self) -> None:
         import wandb
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -7,7 +7,9 @@ from __future__ import annotations
 import os
 import pathlib
 import re
+import shutil
 import sys
+import tempfile
 import textwrap
 import warnings
 from typing import Any, Dict, Optional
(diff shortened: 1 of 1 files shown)


mvpatel2000:
Also, I'm not really sure what's going on, but when you checkpoint with wandb in the unit tests, the checkpoints are corrupted. My hunch is if you don't provide proper credentials to wandb and start using it, it doesn't nicely catch this scenario and give you an error, and instead continues onward stumbling along in a broken fashion. 

For example, if you try to save a checkpoint after adding the Wandb logger in the unit tests, it becomes corrupted since `wandb.run.project` returns something weird which torch doesn't like when loading. I spent some time trying to figure out what it's returning, but I have no idea what's gong on...
image

composer/loggers/wandb_logger.py
```diff
@@ -127,6 +129,26 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
...
+    ):
+        # Note: Wandb doesn't support progress bars for downloading
+        del chunk_size, progress_bar
+        import wandb
+
+        artifact = wandb.use_artifact(artifact_name)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            artifact_folder = os.path.join(tmpdir, "artifact_folder")
+            artifact.download(root=artifact_folder)
+            # We only log one file per artifact
```
ravi-mosaicml, line 147: Can you raise a `RuntimeError` here if this assumption does not hold? (In case if the user uploaded an artifact from a non-composer run?)

composer/loggers/wandb_logger.py
```diff
@@ -127,6 +129,26 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
...
+            checkpoint_name = os.listdir(artifact_folder)[0]
```
ravi-mosaicml, line 148: nit:
```suggestion
            artifact_name = os.listdir(artifact_folder)[0]
```

composer/loggers/wandb_logger.py
```diff
@@ -127,6 +129,26 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
...
+            checkpoint_path = os.path.join(artifact_folder, checkpoint_name)
```
ravi-mosaicml, line 149:
```suggestion
            artifact_path = os.path.join(artifact_folder, checkpoint_name)
```

ravi-mosaicml (approved):
LGTM! See comments.

ravi-mosaicml:
> Also, I'm not really sure what's going on, but when you checkpoint with wandb in the unit tests, the checkpoints are corrupted. My hunch is if you don't provide proper credentials to wandb and start using it, it doesn't nicely catch this scenario and give you an error, and instead continues onward stumbling along in a broken fashion.
> 
> For example, if you try to save a checkpoint after adding the Wandb logger in the unit tests, it becomes corrupted since `wandb.run.project` returns something weird which torch doesn't like when loading. I spent some time trying to figure out what it's returning, but I have no idea what's gong on... image

In the unit tests, we set the variable `WANDB_DISABLED`, so it may be incompatible with that setting? So long as you validated this works manually, then we're all good! It would probably be worthwhile adding in a wandb key into jenkins so we can run integration tests, but that's an issue for another day :)

PR #1015, opened 2022-05-07
Fix SWA Serialization

hanlint:
This PR makes the SWA algorithm serializable. A few changes have been made to realize this:
* The `swa_scheduler` and `swa_model` are now created upon `INIT`, so that the weights can be optionally loaded. 
* Some of the core logic has been refactored for readability
* Added the comparison of `Time` into `deep_compare`. I also added a comparison of callable objects, but the current equality check is a bit hacky, open for ideas here. 

There are a few limitations that still need to be addressed:
* `swa_start` and `swa_end` now must have the same units. This is not strictly necessary, and will be fixed in a future commit. 
* the `AverageModel` is now stored on `cpu`. Need some guidance on whether we want to default to this (less memory, but then more PCIe bandwidth), enable moving this to device, or allowing the user to choose. If we allow non-cpu AverageModel, then the we need to supply the device info during deserialization. 

Because this did refactor the `SWA`, we should run some comparison w&b runs.


composer/core/precision.py
```diff
@@ -65,7 +65,7 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
     elif precision == Precision.BF16:
         if version.parse(torch.__version__) < version.parse("1.10"):
             raise ValueError(f"BF16 precision requires torch > 1.10, got version {torch.__version__}")
-        with torch.cuda.amp.autocast(enabled=True, dtype=torch.bfloat16):
+        with torch.cuda.amp.autocast(enabled=True, dtype=torch.bfloat16):  # type: ignore (torch < 1.10 guard)
```
hanlint, line 68: added this, as pyright fails to detect the version guard when running with `torch==1.09`

tests/common/compare.py
```diff
@@ -42,6 +44,15 @@ def _check_item(item1: Any, item2: Any, path: str, rtol: float = 0.0, atol: floa
         assert isinstance(item2, type(item1)), f"{path} differs: {item1} != {item2}"
         _check_list_recursively(item1, item2, path, atol=atol, rtol=rtol)
         return
+    if isinstance(item1, Time):
+        assert type(item1) == type(item2)
+        assert item1 == item2, f"{path} differs: {item1} != {item2}"
+        return
+    if callable(item1):
+        assert callable(item2)
+        assert item1.__name__ == item2.__name__, f"{path} differs: {item1} != {item2}"
```
hanlint, line 53: This is an extremely loose equality check, open to suggestions.

composer/algorithms/swa/swa.py
```diff
@@ -157,51 +133,81 @@ def __init__(self,
...
+        if self.swa_end.unit == TimeUnit.DURATION and self.swa_end == 1:
+            log.warning("'swa_end' = '1dur'. Batch norm statistics of averaged model "
+                        "will not be updated. This will negatively impact accuracy. "
+                        "See the documentation for the `swa_end` parameter for details.")
+
+        assert_valid_duration(self.swa_start)
+        assert_valid_duration(self.swa_end)
+
+    def _get_time(self, state: State):
+        """helper function to retrieve wiether the epoch or the duration depending on the units"""
```
growlix, line 171:
```suggestion
        """helper function to retrieve either the epoch or the duration depending on the units"""
```

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -44,7 +44,7 @@ def test_algorithm_resumption(
         # see: https://github.com/mosaicml/composer/issues/362
         pytest.importorskip("torch", minversion="1.10", reason="Pytorch 1.10 required.")
 
-    if algorithm in ('layer_freezing', 'swa'):
+    if algorithm in ('layer_freezing'):
```
growlix, line 47: 🙌

tests/common/compare.py
```diff
@@ -5,6 +5,8 @@
 import numpy as np
 import torch
 
+from composer import Time
+
 
 def deep_compare(item1: Any, item2: Any, atol: float = 0.0, rtol: float = 0.0):
     """Compare two items recursively. Supports dicts, lists, tuples, tensors, and numpy arrays.
```
growlix, line 12:
```suggestion
    """Compare two items recursively. Supports dicts, lists, tuples, tensors, numpy arrays, Composer Time objects, and callables.
```

tests/common/compare.py
growlix: I don't have any great suggestions for the equality check, but do we want to
```
assert hasattr(item1, __name__)
assert hasattr(item2, __name__)
```
before comparing the items' `__name__`?

growlix (approved):
Generally LGTM. Small minor comments. Regarding the `AveragedModel` device, I wonder whether our current workloads may be PCIe bottlenecked, but this will increasingly be unlikely as models get larger, so as long as it eventually becomes a hparam we are indeed good.

composer/algorithms/swa/swa.py
```diff
@@ -5,19 +5,25 @@
...
 from composer.core import Algorithm, Event, State, Time, TimeUnit
+from composer.core.types import PyTorchScheduler
 from composer.loggers import Logger
 
 log = logging.getLogger(__name__)
 
 __all__ = ['SWA']
 
 
+def assert_valid_duration(time: Time):
```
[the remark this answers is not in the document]
hanlint: done!

tests/common/compare.py
hanlint: Decided to remove this check, and not save the callable from `SWALR`

PR #1017, opened 2022-05-09
Unflatten Profiler Args

ravi-mosaicml:
- To reduce the complexity of the `Trainer` constructor, the profiler arguments are unflattened. While this violates the style guide, flattening the arguments made the Trainer constructor overly complex for the more common uses cases.
-  Modified the `Profiler` constructor to not take the `state`, and instead added a `bind_to_state` method.
- Added in a `ProfilerHparams` class
- Removed the `run_composer_profiler.py` entrypoint, since we have the `profiler_demo.py`, don't want to emphasize the yahp path, and the entrypoint is quite opinionated (e.g. max duration, no validation)
- Updated docs and tests
- Added a warning when using the profiler, and also when not specifying `train_subset_num_batches`.


composer/trainer/trainer.py
```diff
@@ -792,39 +712,16 @@ def __init__(
...
-            self.state.callbacks.extend(self.state.profiler.trace_handlers)
+        if profiler is not None:
+            warnings.warn("The profiler is enabled. Using the profiler adds additional overhead when training.")
+            self.state.profiler = profiler
+            self.state.profiler.bind_to_state(self.state)
+
+            # Warn the user if there are likely misconfigurations when using the Profiler
+            if train_subset_num_batches != -1:
+                warnings.warn("When using the profiler, it is recommended to set `train_subset_num_batches` "
+                              "to the length of the profile schedule. This setting will end each epoch early to avoid "
```
eracah, line 723: Can you print what the length of the profile schedule is in this warning? Or explicitly tell the user to set `train_subset_num_batches` to -1?

eracah (commented):
Thanks, @ravi-mosaicml! This removes a ton of bloat from the trainer constructor, which is huge! Great work. Any chance you could add a test to `test_trainer` or to `test_profiler` just to check that the profiler is instantiated correctly?

composer/trainer/trainer.py
ravi-mosaicml: Since the `prof_schedule` is a callable, the trainer does not have any insight as to when it will fire / what underlying logic it has. So, unfortunately, we cannot recommend a value, other than suggesting that this parameter should be set.

bandish-shah, line 723: I'm actually not sure we should print this warning.  A lot of times you want to run more iterations than what the profile window is set to capture to make sure execution is in steady state.  Other times you want to capture ramp up and ramp down so I think this message is a bit too opinionated.  Maybe you can recommend setting `train_subset_num_batches` in documentation but I don't think this level of hand holding is necessary.

bandish-shah (approved):
Overall LGTM, see my [nit] about hand holding

composer/trainer/trainer.py
ravi-mosaicml: Good points; removed the warning.

PR #1020, opened 2022-05-09
Add `State.eval_timestamp` and `State.predict_timestamp`

ravi-mosaicml:
- Added `eval_timestamp` and `predict_timestamp` as attributes on the State for tracking evaluation and prediction progress. This is useful for callbacks and loggers to track where we are in the current evaluation or prediction dataloader (for example, logging metrics where the X axis is the evaluation batch number, and the Y axis is some metric, like accuracy, for that batch).
- Added new attributes to the state, instead of hot-swapping `state.timestamp`, since it is still useful to know the training batch number during evaluation (e.g. track how evaluation improves as training progresses)
- Added tests to ensure that the timestamp is properly set.


A-Jacobson (approved):

composer/trainer/trainer.py
```diff
@@ -1367,9 +1366,14 @@ def _train_loop(self) -> None:
                         full_loss = total_loss.cpu().item()
                         self.logger.data_batch({'loss/train': full_loss / dist.get_world_size()})
 
+                    total_num_samples, total_num_tokens = self._accumulate_samples_and_tokens_across_ranks(
+                        num_samples=self.state.batch_num_samples,
+                        num_tokens=self.state.batch_num_tokens,
+                    )
+
```
A-Jacobson, line 1373: we're moving this `all_reduce` from `Event.AFTER_DATALOADER` to `Event.BATCH_START`. is this intentional?

ravi-mosaicml: Good catch; moved back to where it was. However, that's currently a bug -- either it should be before AFTER_DATALOADER or after BATCH_START, depending on whether the timestamp should count the number of samples yielded from the dataloader or the number of samples actually trained. If the latter, then the `batch_num_samples` and `batch_num_token` counts should be updated again after AFTER_DATALOADER and BATCH_START to reflect how algorithms may change the sample count (e.g. selective backprop)

PR #1021, opened 2022-05-09
Configuration Logging

### LINKED ISSUE #826: Improved `wandb` logging experience
Based on feedback from @tcapelle, our `wandb` logging experience can be improved with better naming, and more default things logged, see comment here:

> We could improve the naming of the wandb logging variables (we get a bunch of panes with weird names).
I will wait your next release to take a look. Also you are not logging anything to the config, you can pull a bunch of data from the Trainer and the ComposerModel. Look at all that fastai get's you for free: batch_size, epochs, loss_func, metrics_names, callbacks, loggers, flags, etc...

From: https://github.com/mosaicml/composer/issues/809#issuecomment-1079026212

### LINKED ISSUE #931: Refactor logging of `rank_zero_seed`
This line is logging the `rank_zero_seed` as if it is  time series data, but it's a static value that is only configured once. It looks very out of place in logger backends like WandB. I think `rank_zero_seed` should be stored in config somewhere but not logged as time series.

This can be done by either promoting it to the `TrainerHparams`, or creating a new `update_config()` function that `LoggerDestinations` have to implement.

Also a UX bug -- if each rank is logging in distributed training, then the `seed` stored in each rank's config is now incorrect because `Trainer` has overridden it to be `rank_zero_seed + rank`. Are there any other codepaths we are overriding values in the config?

https://github.com/mosaicml/composer/blob/204129b810ce71d4089781b6ee2bef9f0b2703f1/composer/trainer/trainer.py#L873

ravi-mosaicml:
- Add a `LoggerDestination.log_config` method. It is necessary to handle config logging different than data logging, since each backend may want to handle it differently (e.g. WandB has a config option; the FileLogger can store config as a yaml). This was proposed in #826.

  Logger destinations should override any existing config with the config dictionary as specified in this method. The trainer will invoke `log_config` after `Event.INIT`, and during each call to `Trainer.fit()`.

- Add a `get_config()` helper to the callbacks, algorithms, loggers, and evaluator classes. These instances can optionally implement this method to return a configuration dictionary that should be logged. The trainer then queries this method to capture any config and include it in the call to `log_config`. For example, the `CheckpointSaver` implements this method, to log the folder where checkpoints are saved. While this method could be an `@property`, doing it as a method makes it easier for callback implementations, since they won't need to remember to include the `@property` decorator.

  This style, rather than requiring callbacks to call `Logger.log_config` directly, prevents conflicts where multiple callbacks / algorithms may want to log to the same key. This could happen when using multiple `CheckpointSaver`s, for example. Instead, the `Trainer` prefixes each config dictionary with an index (when there are multiple config dictionaries of the same type). See an example below.
- Added a `_log_config` helper to the training loop. This method computes a configuration dictionary from training loop variables and the results from `get_config()` above. This dictionary is then passed to `LoggerDestination.log_config`. The trainer logs the configuration after `Event.INIT`, and during each call to `Trainer.fit()`.
- Implemented `LoggerDesination.log_config` for the `FileLogger`, `WandBLogger`, `InMemoryLogger`, and `ProgressBarLogger`
- Implemented `get_config()` for the `CheckpointSaver`, `GradMonitor`, `SpeedMonitor`, and `Evaluator`

Closes #826 
Closes #931

---

### Classify MNist from Entrypoint

Below is an example of how the config would appear when using the progress bar logger.

Note that our entry points will result in the config being printed twice -- once as YAML, and again via the config dictionary. However, for one using the functional (non-yahp) API, they would only get the example config below.

Command: `python3 examples/run_composer_trainer.py -f composer/yamls/models/classify_mnist_cpu.yaml --datadir ~/datasets`

Output:

```
###Begin Configuration###
trainer/algorithms: []
trainer/backwards_create_graph: false
trainer/callbacks: []
trainer/ddp_enabled: false
trainer/ddp_sync_strategy: null
trainer/deepspeed_config: null
trainer/deterministic_mode: false
trainer/device: DeviceCPU
trainer/evaluator/batch_size: 1000
trainer/evaluator/label: eval
trainer/evaluator/metrics:
- Accuracy
- CrossEntropy
trainer/evaluator/subset_num_batches: -1
trainer/find_unused_parameters: false
trainer/grad_accum: 1
trainer/grad_clip_norm: -1.0
trainer/loggers:
- ProgressBarLogger
trainer/max_duration: 10ep
trainer/model_class: MNIST_Classifier
trainer/optimizers:
- SGD
trainer/precision: fp32
trainer/profiler_enabled: false
trainer/rank_zero_seed: 42
trainer/run_name: 1652210295-accurate-dog
trainer/start_time: 0ep
trainer/train_batch_size: 2048
trainer/train_dataloader_label: train
trainer/train_dataloader_len: 29
trainer/train_metrics: null
###End Configuration#####
```

### Example from a test

Below is an example output from a test case. Here, the "CallbackWithConfig" is repeated multiple times when passed into the trainer -- e.g.

```python
trainer = Trainer(
    callbacks=[CallbackWithConfig(), CallbackWithConfig()],
)
```

And here is the output:
```
callback/CallbackWithConfig/1/hello: world
callback/CallbackWithConfig/2/hello: world
trainer/algorithms:
- ChannelsLast
trainer/backwards_create_graph: false
trainer/callbacks:
- CallbackWithConfig
- CallbackWithConfig
trainer/ddp_enabled: false
trainer/ddp_sync_strategy: null
trainer/deepspeed_config: null
trainer/deterministic_mode: true
trainer/device: DeviceCPU
trainer/find_unused_parameters: false
trainer/grad_accum: 1
trainer/grad_clip_norm: -1.0
trainer/loggers:
- InMemoryLogger
- ProgressBarLogger
trainer/max_duration: 2ba
trainer/model_class: SimpleModel
trainer/optimizers:
- SGD
trainer/precision: fp32
trainer/profiler_enabled: false
trainer/rank_zero_seed: 0
trainer/run_name: 1652211402-electric-zebra
trainer/scale_schedule_ratio: 1.0
trainer/scheduler_step_frequency: batch
trainer/schedulers:
- StepScheduler
trainer/start_time: 0ba
trainer/train_batch_size: 1
trainer/train_dataloader_label: train
trainer/train_dataloader_len: 100
trainer/train_metrics: null
```


abhi-mosaic:
Read through the motivation and took a brief look at the code. Somehing I'm trying to think about... other than the basic functionality of "every Composer run now saves a YAML that has some info about the run", do we have any stronger guarantees? Is it always true that we can reproduce a Composer run given this config? I don't think so.... Because not every algorithm/callback/etc. is saving all the init information to recreate it exactly. What is the user intended go get from this config if they launched the run using / not using YAHP? 

Going backwards, if the goal was to save a reproducable config, is there any way to turn a Trainer instance into a TrainerHparams, and just save that?

ravi-mosaicml:
(Going to close this PR for the time being until the yaml path is fleshed out more)

PR #1027, opened 2022-05-10
Manually make the /tmp folder in the image

ravi-mosaicml:
#974 didn't fix the missing tmp folder. Instead, let's manually create it.


Averylamp (approved):
lgtm

PR #1029, opened 2022-05-11
Release Testing Script + Fix Test Output

ravi-mosaicml:
- Added a release testing script that the PyPI release job (See https://github.com/mosaicml/testing/blob/main/pipelines/pypiPublish.groovy) will invoke before publishing Composer. These tests run on a clean install of Composer in a new virtual environment, without any of the dev dependencies (hence, cannot use pytest). These tests serve a final sanity check (before doing the irreversible pypi publish) that composer is pip-installable and the most basic examples work. Tests ensure that the examples here stay up to date with the README.md
- Instruct Jenkins to push the full test output to GitHub, not just the summary


bandish-shah (approved):
LGTM

PR #1031, opened 2022-05-11
Test for other export options (torch.fx, ONNX)

hanlint:
This PR adds simple tests for whether our surgery algorithms are exportable for inference via `torch.fx` or `ONNX`. This is an extension of #926 to other export options. 

`test_torchscript.py` is now renamed to `test_torch_export.py` to reflect the additional options.

Currently, the `torch.fx` option fails for several algorithms because their layers have unsupported control flow. 

With these updates, the current status of export support for **inference** is as follows (X=supported):

|                        | torchscript | torch.fx | ONNX |
|------------------------|-------------|----------|------|
| apply_blurpool         | X           |          | X    |
| apply_factorization    |             | X        | X    |
| apply_ghost_batchnorm  | X           |          | X    |
| apply_squeeze_excite   | X           | X        | X    |
| apply_stochastic_depth | X           | X        | X    |
| apply_channels_last    | X           | X        | X    |

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -163,6 +163,11 @@ extra_deps["streaming"] = [
     "boto3>=1.21.45,<2",
 ]
 
+extra_deps["onnx"] = [
+    "onnx>=1.11.0,<2",
+    "onnxruntime>=1.11.0,<2",
+]
+
 extra_deps["all"] = set(dep for deps in extra_deps.values() for dep in deps)
 
 composer_data_files = ["py.typed"]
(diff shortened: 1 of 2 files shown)


dskhudia (approved):

dskhudia:
I can take a look at the failing algorithms with fx later to see if we can avoid the control flow.

hanlint:
Thanks @dskhudia not a priority, I just wanted to at least get the testing framework expanded.

PR #1033, opened 2022-05-11
Proofread Early Stopper, Grad Monitor, LR Monitor, Memory Monitor, MLPerf Callback, Speed Monitor, and Dataset Utils Docstrings

mcneela:
Proofread Early Stopper, Grad Monitor, LR Monitor, Memory Monitor, MLPerf Callback, Speed Monitor, and Dataset Utils Docstrings.


ravi-mosaicml (approved):
LGTM

PR #1035, opened 2022-05-12
Update serialized format, adding magic and version

knighton:
Also switch the serialization of bytes_per_sample from i64 to u32, which cuts the index size in half while restricting samples to <4GB

~TODO~ DONE:
* re-generate all StreamingDataset implmentations (ADE20k, ImageNet, COCO), and upload to S3
* change defaults in YAMLs to point to new versions (e.g. `.../mds/1/`)

### DIFF
--- a/composer/datasets/streaming/format.py
+++ b/composer/datasets/streaming/format.py
@@ -119,7 +119,6 @@ class StreamingDatasetIndex(object):
 
     def __init__(self, samples_per_shard: NDArray[np.int64], bytes_per_shard: NDArray[np.int64],
                  bytes_per_sample: NDArray[np.int64], fields: List[str]) -> None:
-
         self.samples_per_shard = samples_per_shard
         self.bytes_per_shard = bytes_per_shard
         self.bytes_per_sample = bytes_per_sample
@@ -161,14 +160,15 @@ class StreamingDatasetIndex(object):
         Returns:
             cls: The loaded object.
         """
-
-        dtype = np.int64
-        total_samples, total_bytes, num_shards, num_fields = read_array(fp, 4, dtype)
+        magic, version = read_array(fp, 2, np.uint32)
+        assert magic == 0xDA7AD06E
+        assert version == 1
+        total_bytes, total_samples, num_shards, num_fields = read_array(fp, 4, np.int64)
         del total_bytes
-        samples_per_shard = read_array(fp, num_shards, dtype)
-        bytes_per_shard = read_array(fp, num_shards, dtype)
-        bytes_per_sample = read_array(fp, total_samples, dtype)
-        bytes_per_field = read_array(fp, num_fields, dtype)
+        samples_per_shard = read_array(fp, num_shards, np.int64)
+        bytes_per_shard = read_array(fp, num_shards, np.int64)
+        bytes_per_sample = read_array(fp, total_samples, np.uint32).astype(np.int64)
+        bytes_per_field = read_array(fp, num_fields, np.int64)
         fields = [fp.read(size).decode('utf-8') for size in bytes_per_field]
         return cls(samples_per_shard, bytes_per_shard, bytes_per_sample, fields)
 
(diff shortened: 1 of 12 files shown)


composer/datasets/streaming/format.py
```diff
@@ -178,12 +178,15 @@ def dumps(self) -> bytes:
         Returns:
             bytes: The serialized form.
         """
-        header = np.array([self.total_samples, self.total_bytes, self.num_shards, self.num_fields], np.int64)
-        bytes_per_field = np.array(list(map(len, self.fields)), np.int64)
-        arrays = header, self.samples_per_shard, self.bytes_per_shard, self.bytes_per_sample, bytes_per_field
-        arrays = np.concatenate(arrays, dtype=np.int64).tobytes()
-        fields = b''.join(map(lambda s: s.encode('utf-8'), self.fields))
-        return arrays + fields
+        magic = 0xDA7AD06E
```
ravi-mosaicml, line 181: What is this magic for?

moinnadeem:
@ravi-mosaicml To be clear, is this size per individual training example, or something like a shard? If it's a single training example, then 4GB should be a pretty acceptable memory limit. I could see extremely large images (ie. cancer x-rays) being a problem though.

knighton:
size per individual training example

composer/datasets/streaming/format.py
knighton: Spend the first 4 bytes on a file signature "doge data", a common pattern (https://en.wikipedia.org/wiki/List_of_file_signatures) used to identify the file type outside of filename

ravi-mosaicml:
> @ravi-mosaicml To be clear, is this size per individual training example, or something like a shard? If it's a single training example, then 4GB should be a pretty acceptable memory limit. I could see extremely large images (ie. cancer x-rays) being a problem though.

Hmm, if there is a use case that would exceed it, then I would like to support larger sample sizes. Perhaps could we have a flag in the index file on whether the sample sizes are 32 bit or 64 bit integers?

knighton:
Or a bytes per size field (1 2 3 4 8 etc)

Or we do the shard writing in two stages: collect sample sizes in first stage, then optimize a fixed vocabulary of say 65536 sample sizes for which each sample is padded up to nearest that minimizes overall serialized dataset size, write padded samples in second stage, and then we need 2 bytes per sample size when the dataset is in use (would scale up better for some minor % wasted space)

Or we just wait until a customer actually exceeds 4GB/sample which seems crazy to me

hanlint:
Yeah I think we can defer this until we have an actual use case.

ravi-mosaicml:
For now, can you record in the index file what dtype the samples are? I'm OK deferring the implementation of 64 bit samples sizes in the writer and reader till there is more demand for that, but also don't want ot invalidate any existing index files because we made a breaking change to support that. Since this is a standard, I think it's more important to be future thinking, than say any other feature in composer which we can easily change when there is a need.

(As for a possible implementation, we could also add in a sample_size_dtype flag that defaults to int32; then when writing the samples, if a sample size exceeds this flag then it errors. But I think the two pass approach is more elegant)

composer/datasets/streaming/format.py
ravi-mosaicml: Ah, neat!

knighton:
Updated PR. We accumulate sample sizes across shard writing and write the entire index after all shards, so we now rightsize the dtype at that point.

abhi-mosaic:
> Updated PR. We accumulate sample sizes across shard writing and write the entire index after all shards, so we now rightsize the dtype at that point.

I think this should hopefully address the above comments and future-proof the Index format for a while?
I am currently re-generating the datasets and running our remote tests and it looks good so far on ADE20k. I will also be fixing our COCO datset (xfail) so that we can respond to our external user who is requesting StreamingCOCO.

abhi-mosaic:
Summary of changes:
* All MDS datasets upgraded to v1 and regenerated 
* DeeplabV3 and R50 streaming YAMLs updated, along with default paths, to v1
* COCO dataset bugfixed, passing remote tests
* change default timeout to 60s and default shard size to 32MB to address some timeouts observed in pratice (this is a temporary measure until download is refactored)

knighton:
Should we specify remote/local in the yamls if they are already in the hparam defaults (or vice versa)?

abhi-mosaic:
> Should we specify remote/local in the yamls if they are already in the hparam defaults (or vice versa)?

Yes, for safety I always specify every param, this helps catch breaking changes if someone changes the defaults, and makes it easier for a user to override `remote` with their own path, rather than trek back to the source code to figure out what the key name is ("was it 'remote' or 'remote_dir'...")

abhi-mosaic (approved):

abhi-mosaic:
> Should we specify remote/local in the yamls if they are already in the hparam defaults (or vice versa)?

Also, re. the vice-versa, I am very in favor of removing the defaults in the YAHP hparams :) , but I can't until we remove the fields in parent class `DatasetHparams`, which are optional and force us to use optional fields in all subclasses, and the Optionals have to have defaults

knighton:
Roger that

PR #1037, opened 2022-05-12
Graceful Resumption

mvpatel2000:
- adds autoresume param for graceful resumption
- supports loading from latest, then checking from loggers for latest, then checking from load_path with autoresume
- fixes relative path bug in latest symlink. Previously, latest symlink used folder/filename which only worked if save_folder was an absolute path. We instead change symlink to be local


mvpatel2000:
@ravi-mosaicml feel free to add other reviewers as appropriate (Hanlin? Austin? looking at who commented on GR doc)

composer/trainer/trainer.py
```diff
@@ -898,7 +906,56 @@ def __init__(
 
         # Load Checkpoint
         self._rng_state = None
-        if load_path is not None:
+        # If autoresume is enabled, first check for existing checkpoints to load
```
hanlint, line 909: can this be moved into a separate helper method so that the trainer code is more readable? e.g. ideally something like:

```
if autoresume:
    latest_checkpoint = retrieve_latest_checkpoint(...)
    if latest_checkpoint:
        log.info('Found XYZ, ...')
        load_path = latest_checkpoint

load_checkpoint(...)
```
or maybe just putting `autoresume` into the `load_checkpoint` function?
```
load_checkpoint(..., autoresume=autoresume)
```

Prefer not to have two places where we call load_checkpoint.

composer/trainer/trainer.py
```diff
@@ -543,6 +545,9 @@ class Trainer:
 
             This parameter only controls how many checkpoints are kept locally; checkpoints are not deleted from
             artifact stores.
+        autoresume (bool, optional): Whether or not to enable autoresume, which allows jobs to be killed and restarted
+            to continue training. This parameter requires ``save_folder`` and ``run_name`` to be specified and 
+            ``save_overwrite`` to be ``False``. (default: ``False``)
```
ravi-mosaicml, line 550: For the docstring here, can you elaborate on the logic that is in the notion (e.g. how load_path and load_weights only are internalized), how it finds the checkpoints to restore, and maybe run through a few hypothetical examples (no code required)

composer/trainer/trainer.py
mvpatel2000: Picked the first option since we need to ignore certain args (object_store, weights_only) with autoresume

composer/trainer/trainer.py
mvpatel2000: Updated! Feel free to suggest more changes

composer/trainer/trainer.py
hanlint: Ah to clarify, could you move all the auto-resume specific logic (the try except around the logger, etc.) out into a separate helper method (and out of the trainer logic)

composer/callbacks/checkpoint_saver.py
```diff
@@ -392,7 +393,7 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
                     os.remove(symlink_name)
                 except FileNotFoundError:
                     pass
-                os.symlink(checkpoint_filepath, symlink_name)
+                os.symlink(Path(checkpoint_filepath).name, symlink_name)
```
ravi-mosaicml, line 396: Hmm, would this work if the `checkpoint_filepath` contained a subfolder within the checkpoint folder?

E.g. checkpoint folder = `foo`
checkpoint_filename = `rank{rank}/epoch{epoch}_batch{batch}`
latest = `rank_{rank}`

Then would this write `foo/rank_0` to point to `foo/rank0/epoch0_batch1`, for example?

(Sorry, I don't really get how symlinks work, but want to make sure this covers the edge cases)

composer/trainer/trainer.py
```diff
@@ -543,6 +545,19 @@ class Trainer:
...
+            ``save_folder`` and ``run_name`` to be specified and ``save_overwrite`` to be ``False``. (default: ``False``)
+
+            When enabled, the save_folder is checked for checkpoints of the format ``"{save_folder}/{save_latest_filename}"``,
+            which are loaded to continue training. If no local checkpoints are found, each logger is checked for potential
+            checkpoints named ``save_latest_artifact_name``. Finally, if no logged checkpoints are found, ``load_path`` is
+            used to load a checkpoint if specified. This should only occur at the start of a run using autoresume.
+
+            For example, to run a fine-tuning run on a spot instance, ``load_path`` would be set to the original weights and
+            an object store logger would be added. In the original run, ``load_path`` would be used to get the starting
+            checkpoint. For any future restarts, such as due to the spot instance being killed, the latest checkpoint from
```
ravi-mosaicml, line 559: This description is super helpful, thanks!

nit: The wandb logger also saves checkpoints; so would prefer for the docs to be more general.

```suggestion
            checkpoint. For any future restarts, such as due to the spot instance being killed, the loggers would be queried for the latest checkpoint
```

composer/trainer/trainer.py
```diff
@@ -666,6 +681,9 @@ def __init__(
         save_weights_only: bool = False,
         save_num_checkpoints_to_keep: int = -1,
 
+        # Graceful Resumption
+        autoresume: Optional[bool] = False,
```
ravi-mosaicml, line 685: nit: `Optional` means that it can be None, not that it is not required

```suggestion
        autoresume: bool = False,
```

composer/trainer/trainer_hparams.py
```diff
@@ -454,6 +455,13 @@ class TrainerHparams(hp.Hparams):
         default=-1,
     )
 
+    # Graceful Resumption
+    autoresume: Optional[bool] = hp.optional(
```
ravi-mosaicml, line 459:
```suggestion
    autoresume: bool = hp.optional(
```

composer/trainer/trainer.py
```diff
@@ -898,6 +916,48 @@ def __init__(
...
+            if save_overwrite:
+                raise ValueError(
+                    "save_overwrite must be False when autoresume is enabled as autoresume always loads the latest existing checkpoint in save_folder."
+                )
+            if save_latest_filename is None:
+                raise ValueError(
+                    "save_latest_filename must be specified so autoresume knows where to load checkpoints from.")
+            if run_name is None:
+                raise ValueError("run_name must be specified so autoresume knows which run to load from.")
+            latest_filename_formatted = format_name_with_dist(save_latest_filename, run_name)
```
ravi-mosaicml, line 932: Following @hanlint's comment above, would it make sense to move this logic (and below) be moved into the `load_checkpoint` function?

We could change the `object_store` parameter of the `load_checkpoint` function to take `Logger | ObjectStore | Sequence[Logger | ObjectStore]`

For the validation (before this line), agree that could be moved into a helper -- e.g. `_validate_autoresume()`

composer/trainer/trainer_hparams.py
```diff
@@ -538,7 +546,20 @@ def validate(self):
 
         if (isinstance(self.grad_accum, str) and self.grad_accum != "auto") or (isinstance(self.grad_accum, int) and
                                                                                 self.grad_accum < 1):
-            raise ValueError('grad_accum must be "auto" or an int greater than or equal to 1')
+            raise ValueError('grad_accum must be "auto" or an int greater than or equal to 1.')
+
+        if self.autoresume:
```
ravi-mosaicml, line 551: Don't worry about adding the validation in here -- since the hparams constructs the trainer, the same errors would be raised anyways. Unrelated, but I kinda want to remove all the hparams validate functions -- https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

ravi-mosaicml (approved):
Overall LGTM! Great work. Main comment same as @hanlint's with code organization. Would be awesome to get this in for the v0.7.0 release (branch will be cut early next week)

composer/trainer/trainer.py
mvpatel2000: Oh yes, sorry I know understand the code snippet. Updating!

composer/trainer/trainer.py
mvpatel2000: Went with the first suggestion, because I think it's better to keep load_checkpoint as a function that just... loads checkpoint lol. More intuitive for a user if they're building something themselves

composer/callbacks/checkpoint_saver.py
mvpatel2000: Good catch -- fixed it to work in case `checkpoint_filename` is a path instead of just a filename

PR #1041, opened 2022-05-12
Add self.batch_set_item and self.batch_get_item for cutmix

eracah:
Applies the new setters and getters from [PR 943](https://github.com/mosaicml/composer/pull/978) to batches for cutmix


composer/algorithms/cutmix/cutmix.py
```diff
@@ -184,6 +184,8 @@ class CutMix(Algorithm):
             box such that each pixel has an equal probability of being mixed.
             If ``False``, defaults to the sampling used in the original
             paper implementation. Default: ``False``.
+        input_key (Any, optional): A key that indexes to the input from the batch.
+        target_key (Any, optional): A key that indexes to the target from the batch.
```
ravi-mosaicml, line 188: Can you add support for the `get_fn` and `set_fn` as well?

ravi-mosaicml (commented):
Love how simple this is given #978 :)

Main feedback is to add in support for custom getter and setter functions. I think this could be done through one of two ways -- either through additional `__init__` arguments, or via subclassing.

I think I'd prefer the subclassing approach as it will leave the init signature simpler for the 99% use case. For subclassing, I think we'd need four helper functions that a user could optionally override -- e.g. `get_input(self, state)`, `set_input(self, state, new_input)`, `get_target(self, state)`, `set_target(self, state, new_target)` -- where the default implementation is what you have done. However, don't have strong opinions here -- either works for me.

composer/algorithms/cutmix/cutmix.py
A-Jacobson: +1 on this, otherwise Lgtm. Nice two line change.

ravi-mosaicml:
@eracah could you wait until Wednesday to merge this one? I'm going to cut the v0.7.0 branch most likely on Tuesday, and I'm thinking we release the batch type removal for all algorithms together....probably in v0.7.1?

eracah:
> @eracah could you wait until Wednesday to merge this one? I'm going to cut the v0.7.0 branch most likely on Tuesday, and I'm thinking we release the batch type removal for all algorithms together....probably in v0.7.1? (Alternatively, if you can finish all algorithms by Tuesday, then let's get all of them in 🔥 )

Haha that's some good motivation

composer/algorithms/cutmix/cutmix.py
```diff
@@ -200,13 +202,19 @@ class CutMix(Algorithm):
             )
     """
 
-    def __init__(self, num_classes: int, alpha: float = 1., uniform_sampling: bool = False):
+    def __init__(self,
+                 num_classes: int,
+                 alpha: float = 1.,
+                 uniform_sampling: bool = False,
+                 input_key: Any = 0,
```
hanlint, line 209: Can this type be more specific? as a clue to the user?

eracah: I think in the past we decided on `Any` as mentioned [here](https://github.com/mosaicml/composer/pull/978#discussion_r862994938)

A-Jacobson:
Subclassing algorithms wasn't part of the agreed upon design. It's certainly flexible but you'd need a lot more boiler plate and knowledge of the algorithm.

eracah:
> Subclassing algorithms wasn't part of the agreed upon design. It's certainly flexible but you'd need a lot more boiler plate and knowledge of the algorithm.
> 
> Had envisioned callables passed to init for the messy edge cases not covered by indexing.

We do have to add four new arguments: [get, set] x [input, target], but still easier than subclassing

composer/algorithms/cutmix/cutmix.py
eracah: ok added support for `get_fn` and `set_fn`. To avoid extra arguments to the user and extra parsing boilerplate for the algorithm writer, `input_key` and `target_key` optionally accept a sequence of getter, setter. All the parsing logic is pushed down to `batch_set` and `batch_get`. Lemme know what you think, @A-Jacobson

composer/algorithms/cutmix/cutmix.py
ravi-mosaicml: Hmm, we _could_ do a `Union[str, int, callable, Any]`, but that's just more verbose than Any for how pyright will interpret it :)
Perhaps could we clarify in the docstring what this can be, and recommendations for common batch formats?

composer/algorithms/cutmix/cutmix.py
```diff
@@ -184,6 +184,12 @@ class CutMix(Algorithm):
             box such that each pixel has an equal probability of being mixed.
             If ``False``, defaults to the sampling used in the original
             paper implementation. Default: ``False``.
+        input_key (Any, Callable, Sequence[Callable]): A key that indexes to the input
```
ravi-mosaicml, line 187: Could you elaborate how these arguments are used, and perhaps give recommendations for how to set these parameters? Perhaps we can also explain that the default values work for where the batch is an (input, target) pair.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -200,13 +206,19 @@ class CutMix(Algorithm):
             )
     """
 
-    def __init__(self, num_classes: int, alpha: float = 1., uniform_sampling: bool = False):
+    def __init__(self,
+                 num_classes: int,
+                 alpha: float = 1.,
+                 uniform_sampling: bool = False,
+                 input_key: Union[Any, Callable, Sequence[Callable[[], Any]]] = 0,
```
ravi-mosaicml, line 213: Hmm, I think this Union annotation makes it more confusing than just typing it as `Any`. Mentioning only the callable case might confuse one that they have to pass in a callable (which hides all the magic of your batch_get / batch_set functions!)

If we did want to give a hint, would `Union[int, str, Callable, Any]` work? But I think I'd still prefer just `Any` with an explanation in the docstring...

composer/utils/batch_helpers.py
```diff
@@ -1,92 +1,63 @@
...
-    except TypeError as e:
-        if 'object is not subscriptable' in str(e):
-            pass
-        else:
-            raise e
+    # Case 2: key is a getter Callable.
+    if isinstance(key, Callable):
+        return key(batch)
 
+    # Case 3: key some sort of index or key to use to directly extract from the batch.
```
ravi-mosaicml, line 47: Wow this is concise!

composer/utils/batch_helpers.py
```diff
@@ -211,12 +189,11 @@ def _batch_set_tuple(batch: Any, key: Union[int, str], value: Any) -> Any:
...
-    is_named_tuple = hasattr(batch, '_fields')
-    if is_named_tuple and isinstance(key, str):  # NamedTuple w/ named key
-        value = getattr(batch, key)
-    else:  # Normal tuple or namedtuple with int key.
-        value = batch[key]
-
-    return value
+def _is_key_get_and_set_fn_pair(key):
+    if isinstance(key, Sequence) and not isinstance(key, str):
+        if all([isinstance(key_element, Callable) for key_element in key]):
```
ravi-mosaicml, line 194: Nit: prefer to use the `callable()` check instead of `isinstance` -- while it may work here, usually `isinstance` checks on imports from `typing` do not work.

```suggestion
        if all([callable(key_element) for key_element in key]):
```

tests/utils/test_batch_helpers.py
```diff
@@ -26,35 +26,24 @@ def __init__(self, **kwargs):
...
-@pytest.fixture(scope="module",
-                params=[
-                    my_list,
-                    tuple(my_list),
-                    torch.tensor(my_list),
-                    np.asarray(my_list),
-                    deque(my_list),
-                    my_named_tuple(*my_list)
-                ])
+@pytest.fixture(scope="module", params=[
```
ravi-mosaicml, line 29: Wow -- seems like a big refactor for the test cases. It's a bit hard to read through. Is it possible to minimize the diff here to the new test cases to handle the callables, or are some additional edge cases being added?

composer/algorithms/cutmix/cutmix.py
hanlint: I'm fine either way, just that the `Any` typing hint well.. doesn't provide much of a hint to users.

composer/algorithms/cutmix/cutmix.py
hanlint: I find the `Callable` typing confusing. That is an edge case, and makes it hard to read. We probably want a central doc page somewhere that these docstrings can refer to for more detail, as this will be shared across lots of algorithm classes.

composer/core/state.py
```diff
@@ -343,57 +343,44 @@ def schedulers(self):
     def schedulers(self, schedulers: Union[types.PyTorchScheduler, Sequence[types.PyTorchScheduler]]):
         self._schedulers[:] = ensure_tuple(schedulers)
 
-    def batch_get_item(self, key: Optional[Any] = None, get_fn: Optional[Callable[[Any], Any]] = None) -> Any:
+    def batch_get_item(self, key: Union[Any, Callable, Sequence[Callable[[], Any]]]) -> Any:
```
hanlint, line 346: just speculating: should we employ some python magic to make `state.batch` support the get_item and set_item methods? so one could do `state.batch[key]` and `state.batch[key] = value`? Or would that be too magical?

composer/algorithms/cutmix/cutmix.py
eracah: ok changed to `Union[str, int, callable, Any]` here: [a95b28c](https://github.com/mosaicml/composer/pull/1041/commits/a95b28c0b12f769c315eec6a6c4ae5b62369f076)

tests/utils/test_batch_helpers.py
eracah: Yeah, sorry, I did a lot of clean up and reorganization of the tests.

composer/core/state.py
eracah: That is perhaps a bit too magic. It would be cool, but might not be worth implementing. Unfortunately, properties in python only allow getters, setters, and deleters, so we can't do `@batch.setitem` or `@batch.getitem`, even though if we made batch a property, we could do `@batch.setter`

 We would have to either:
a. wrap batch in a class, where we implement our own `__getitem__` and `__setitem__`
b. monkeypatch in our own `__getitem__` and `__setitem__`

Maybe in a future PR? 
@ravi-mosaicml, you have any ideas on how to do this?

composer/algorithms/cutmix/cutmix.py
eracah: ok I can add something to `docs/source/trainer/algorithms.rst` once we change all the other Algorithms

composer/core/state.py
ravi-mosaicml: While wrapping would work, I'd prefer not to, since that would likely break things if a user did `state.model(state.batch)` without first unwrapping it. Regardless if we wrapped it or monkeypatched the item directly, most batch objects already have a `__getitem__` implementation, for which I think it'd feel weird to override the default.

(In most cases, anyways, the batch_get_item heleper just does `state.batch[key]`...)

composer/utils/batch_helpers.py
```diff
@@ -1,92 +1,63 @@
...
-from typing import Any, Callable, Optional, Sequence, Union
+import copy
+from operator import attrgetter, itemgetter
+from typing import Any, Callable, Sequence, Union
 
 __all__ = ['batch_get', 'batch_set']
 
 
-def batch_get(batch: Any, key: Optional[Any] = None, get_fn: Optional[Callable[[Any], Any]] = None) -> Any:
+def batch_get(batch: Any, key: Union[str, int, Callable, Any]):
```
ravi-mosaicml, line 11: Much more readable :)

composer/utils/batch_helpers.py
```diff
@@ -100,37 +71,44 @@ def batch_set(batch: Any,
...
-            * key and set_fn are both unset
-            * key and set_fn are both set
     """
-    if key is None and set_fn is None:
-        raise ValueError("key or set_fn must be specified and neither were!")
-    if key is not None and set_fn is not None:
-        raise ValueError("key and set_fn were both set. Only one can be set!")
-    if set_fn:
-        return set_fn(batch, value)
+    batch_copy = copy.deepcopy(batch)
```
ravi-mosaicml, line 95: Curious why do we copy the batch? I think I would expect that if the batch is a mutable object, then I would want it to be modified in place. (Also not sure whether this would work for GPU tensors -- could cause a memory leak maybe?)

ravi-mosaicml (approved):
Overall LGTM; though see comment re: deepcopy -- curious when we would want to copy the batch rather than modify it in place, if possible (which is what we already do?) (If the batch is an immutable object, like a tuple, then I think a shallow copy makes sense)

I didn't review the test cases closely because of the large diff, but trust that you aren't missing anything :)

composer/utils/batch_helpers.py
eracah: It shouldn't cause memory leak because I think the reference count would go to zero in this case.

eracah: But you're right I don't think we need the deepcopy. Fixed here: [a45b966](https://github.com/mosaicml/composer/pull/1041/commits/a45b9669192efabd3f2607520cb5b0ab3d346a71)

PR #1042, opened 2022-05-13
Change export test to use algorithm settings

hanlint:
Uses the algorithm_settings feature to retrieve the right-sized models to test each surgery algorithm.

Reduces the runtime of the tests from 60+ seconds to 22 seconds :)


ravi-mosaicml (approved):
LGTM, thanks!

PR #1043, opened 2022-05-13
Remove alg hparams

ravi-mosaicml:


composer/algorithms/algorithm_hparams.py
```diff
@@ -1,64 +1,90 @@
...
+from composer.algorithms.randaugment import RandAugment
+from composer.algorithms.sam import SAM
+from composer.algorithms.selective_backprop import SelectiveBackprop
+from composer.algorithms.seq_length_warmup import SeqLengthWarmup
+from composer.algorithms.squeeze_excite import SqueezeExcite
+from composer.algorithms.stochastic_depth import StochasticDepth
+from composer.algorithms.swa import SWA
 from composer.core.algorithm import Algorithm
 
+algorithm_registry: Dict[str, Union[Type[Algorithm], Type[hp.AutoInitializedHparams]]] = {
```
hanlint, line 35: I don't have a strong opinion here, but might be worth a quick design discussion on whether to keep the registry centralized (like we always have) vs. having a a decorator for each class, e.g. something like:

```
@hp.auto_register("blurpool", "algorithm")
class BlurPool:
```

If we keep this "menu" of options, its easy reference, but users would (1) need to know where the codepath of this registry lives in composer to find out this menu of options, and (2) to extend it, know where the specific registry lives and import to modify. Both of these are not very discoverable pieces of information at the moment. 

The decorator approach (1) keeps the key information with the class itself, which is more intuitive than some arbitrary registry file, and (2) keeps the code independant (e.g. a new contributed algorithm would register itself, without having to import/modify the registry code), and (3) is how external users would also register their own stuff. However, you do lose the centralized menu, and also litters our code with decorators. 

cc: @dblalock @eracah @moinnadeem @A-Jacobson who probably have some opinions here.

A-Jacobson: I would prefer a registry. Though hanlin said all that needs to be said about it. The strongest point (which he already made) to me is that it aligns internal and external usage patterns. 

Losing the dictionary doesn't seem  like a large downside since external users would have to know it's there to refer to it. To learn about our algorithms they're likely looking at docs first then code.

ravi-mosaicml: The decorator approach won't work because of how python imports work. Using some sort of autoregistration mechanism would result in phantom imports or offload more work on the entrypoint script. 

Consider the following:

```python
# File `wandb_logger.py`
@hp.register("logger")
class WandBLogger:
    ....
```

```python
# File `my_entrypoint.py`
import yahp as hp

from composer.trainer.trainer_hparams import TrainerHparams  # does NOT import wandb_logger.py

TrainerHparams.create({"loggers": {"wandb": ...}})
```

The above code would not work, as `TrainerHparams` would not know about the `WandBLogger` class.

Now, consider:
```python
# File `my_entrypoint.py`
import yahp as hp
import composer.loggers.wandb_logger  # phantom import

from composer.trainer.trainer_hparams import TrainerHparams

TrainerHparams.create({"loggers": {"wandb": ...}})
```
This would work, because `wandb_logger.py` was imported, calling the call to `hp.register`.

Now, consider this:
```python
# File `my_entrypoint.py`
import yahp as hp

from composer.trainer.trainer_hparams import TrainerHparams

hp.register_subclasses("composer.loggers.wandb_logger")  # explicitly tell yahp to go import this module to discover implementations

TrainerHparams.create({"loggers": {"wandb": ...}})
```
This would also work -- it's basically the same as the above, but offloads the phantom import into yahp. However, I'm also not a fan of this approach, as now every entrypoint would need to add a call to `hp.register_subclass`, and having the decorator + `hp.register_subclass` call is strictly more code than inserting something into a dictionary.

Going to resolve this conversation since I think this needs more design -- but open to alternatives that do not result in phantom imports and that do not involve more code than the current dictionary approach.

composer/callbacks/memory_monitor.py
```diff
@@ -76,12 +76,11 @@ class MemoryMonitor(Callback):
...
-        super().__init__()
-        log.info(
-            "Memory monitor just profiles the current GPU assuming that the memory footprint across GPUs is balanced.")
-        if torch.cuda.device_count() == 0:
-            log.warning("Memory monitor only works on GPU devices.")
+    def init(self, state: State, logger: Logger) -> None:
+        model_device = next(state.model.parameters()).device
+
+        if model_device.type != 'cuda':
+            raise RuntimeError("The memory monitor only works on CUDA devices, but the model is on CPU.")
```
A-Jacobson, line 83: I don't like throwing an error here. Now a user will specifically have to delete this callback or their run still stop if they use the same yaml or script on cpu. I'd prefer to disable it for them in that case rather than forcing code/config chances.

composer/algorithms/algorithm_hparams.py
```diff
@@ -1,64 +1,89 @@
...
+}
 
-@dataclass
-class AlgorithmHparams(hp.Hparams, ABC):
-    """Hyperparameters for algorithms."""
 
-    @abstractmethod
-    def initialize_object(self) -> Algorithm:
-        """Invoked by the :meth:`TrainerHparams.initialize_object` to create an instance of the :class:`Algorithm`.
+def load(algorithm_cls: Union[Type[Algorithm], Type[hp.Hparams]], alg_params: Optional[str]) -> Algorithm:
```
A-Jacobson, line 62: lets remove these! as discussed.

A-Jacobson (approved):
left a few comments mostly on naming and my opinion on the memory logger changes.

Regarding the decorator discussion, I think having decorators is actually complementary to what you've started here and I agree that it needs more design. I ran into the same problems you mentioned when trying to make them work for the benchmarks folder. It would need a function call in the entrypoint similar to how hanlin setup https://github.com/mosaicml/experimental. Either way, we can call them out of scope for this PR.

eracah (approved):

composer/algorithms/algorithm_hparams.py
eracah: What if TrainerHparams does a couple phantom imports: 
```from composer import algorithms
     from composer import loggers
     from composer import callbacks```, which because of the `__init__.py` file in each module will import every Algorithm, Logger, and Callback, which will in turn trigger all the import-time decoration.

Comment (eracah) [reply composer/algorithms/algorithm_hparams.py]: Actually, all it needs to do is: ```import composer```, which will trigger everything. Wdyt @A-Jacobson, @ravi-mosaicml?

File: composer/callbacks/memory_monitor.py [hunk 4]
```diff
@@ -108,9 +107,7 @@ def after_train_batch(self, state: State, logger: Logger):
 
 
 def _get_memory_report() -> Dict[str, Union[int, float]]:
-    if not torch.cuda.is_available():
-        log.debug("Cuda is not available. The memory report will be empty.")
-        return {}
+    assert torch.cuda.is_available(), "the memory report requires cuda"
```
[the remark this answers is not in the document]
ravi-mosaicml: Left as a warning in #1071

composer/algorithms/algorithm_hparams.py
```diff
@@ -1,64 +1,89 @@
...
+from composer.algorithms.randaugment import RandAugment
+from composer.algorithms.sam import SAM
+from composer.algorithms.selective_backprop import SelectiveBackprop
+from composer.algorithms.seq_length_warmup import SeqLengthWarmup
+from composer.algorithms.squeeze_excite import SqueezeExcite
+from composer.algorithms.stochastic_depth import StochasticDepth
+from composer.algorithms.swa import SWA
 from composer.core.algorithm import Algorithm
 
+algorithm_registry: Dict[str, Union[Type[Algorithm], Type[hp.Hparams]]] = {
```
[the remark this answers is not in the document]
ravi-mosaicml: The registry is only for yahp, so I would like to leave hparams or the like in there (e.g. the logger_registry contains the `ObjectStoreLoggerHparams` since that logger has a different hparams path than the functional path). Thoughts about `x_hparams_registry.py` ...would that be too long?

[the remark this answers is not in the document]
ravi-mosaicml: Renamed to `x_hparams_registry.py`

PR #1044, opened 2022-05-13
Ignore params in checkpoint loading

mvpatel2000:
- deletes ignored params from keys in state dict
- for DeepSpeed, passes custom fn for loading which removes keys and then follows same codepath as normal loading

TODO:
- test implementation

Issues:
- current version of deepspeed we use doesn't have the callback_fn


mvpatel2000:
@moinnadeem do you have existing code snippets I can use for testing this?

mvpatel2000:
Looks like the deepspeed callback was added in this PR https://github.com/microsoft/DeepSpeed/commit/50893458d66f27289934dfcc8fd0d1e02a8dcbd7 on May 11th, so its not in any deepspeed releases. Putting a pin in this now + we need to resync on design stuff now that we know what's possible (vs. what we should do)

PR #1045, opened 2022-05-14
Run evaluation at the end of training

moinnadeem:
#947 introduces the concept of `eval_interval` which runs evaluation using our Time abstraction instead of the previous `validate_every_n_batches` and `validate_every_n_epochs`. This works pretty well, but would miss running evaluation at the very end of training, if the training length is not evenly divisible by `eval_interval`. This PR resolves that.

Questions / Discussion Points:
1. I have modified `tests/trainer/test_trainer_eval.py::test_eval_hparams` to enable `assert "eval1" in trainer.state.current_metrics` (before it was `assert "eval1" not in trainer.state.current_metrics`). Is this something we are okay with?

### DIFF
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -37,7 +37,7 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
     if eval_interval.unit not in (TimeUnit.EPOCH, TimeUnit.BATCH):
         raise ValueError("The `eval_interval` must have units of EPOCH or BATCH, or be a function.")
 
-    last_batch_seen = None
+    last_batch_seen = -1
 
     def should_eval(state: State, event: Event):
         if int(eval_interval) <= 0:
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -18,12 +18,14 @@ from composer.core.time import Time, TimeUnit
 __all__ = ["Evaluator", "evaluate_periodically", "ensure_evaluator"]
 
 
-def evaluate_periodically(eval_interval: Union[str, Time, int]):
+def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end: bool = True):
     """Helper function to generate an evaluation interval callable.
 
     Args:
         eval_interval (str | Time | int): A :class:`.Time` instance or time string, or integer in epochs,
             representing how often to evaluate. Set to ``0`` to disable evaluation.
+        eval_at_fit_end (bool): Whether to evaluate at the end of training, regardless of `eval_interval`. 
+            Default: True
     Returns:
         (State, Event) -> bool: A callable for the ``eval_interval`` argument of an :class:`.Evaluator`.
     """
@@ -39,6 +41,9 @@ def evaluate_periodically(eval_interval: Union[str, Time, int]):
         if int(eval_interval) <= 0:
             return False
 
+        # if requested, evaluate at the end of training, as long as the length of training is specified.
+        if eval_at_fit_end and state.max_duration is not None and state.get_elapsed_duration() == 1.0:
+            return True
         if eval_interval.unit == TimeUnit.EPOCH:
             return int(state.timestamp.epoch) % int(eval_interval) == 0 and event == Event.EPOCH_END
         if eval_interval.unit == TimeUnit.BATCH:
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -42,24 +42,25 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
             return False
 
         # if requested, evaluate at the end of training, as long as the length of training is specified.
-        # if eval_at_fit_end and state.max_duration is not None and state.get_elapsed_duration() == 1.0:
-        if eval_at_fit_end and state.max_duration is not None: 
-            # handle for when max_duration is specified in either batches or epochs
-            # handle for when eval_interval is batches
-            # handle for when eval_interval is epochs
-            # print("Evaluating at fit end! Values:", state.timestamp)
-            if eval_interval.unit == TimeUnit.EPOCH:
-                print("Epoch values:", state.timestamp)
-                return int(state.timestamp.epoch) % int(eval_interval) == 0 and event == Event.EPOCH_END
-            if eval_interval.unit == TimeUnit.BATCH:
-                print("Batch values:", state.timestamp)
-                return int(state.timestamp.batch) % int(eval_interval) == 0 and event == Event.BATCH_END
-           # return True
+        if eval_at_fit_end and state.max_duration is not None:
+            # check if we are at the end of training
+            # handle different eval_intervals in order to not duplicate validation steps
+            if eval_interval.unit == TimeUnit.EPOCH and event == Event.EPOCH_END:
+                if state.max_duration.unit == TimeUnit.EPOCH and state.max_duration.value == int(state.timestamp.epoch):
+                    return True
+                if state.max_duration.unit == TimeUnit.BATCH and state.max_duration.value == int(state.timestamp.batch):
+                    return True
+            if eval_interval.unit == TimeUnit.BATCH and event == Event.BATCH_END:
+                if state.max_duration.unit == TimeUnit.EPOCH:
+                    num_epochs = state.max_duration.value
+                    num_total_steps = num_epochs * int(state.dataloader_len)
+                elif state.max_duration.unit == TimeUnit.BATCH:
+                    num_total_steps = state.max_duration.value
+                if num_total_steps == int(state.timestamp.batch):
+                    return True
         if eval_interval.unit == TimeUnit.EPOCH:
-            print("Epoch values:", state.timestamp)
             return int(state.timestamp.epoch) % int(eval_interval) == 0 and event == Event.EPOCH_END
         if eval_interval.unit == TimeUnit.BATCH:
-            print("Batch values:", state.timestamp)
             return int(state.timestamp.batch) % int(eval_interval) == 0 and event == Event.BATCH_END
 
         return False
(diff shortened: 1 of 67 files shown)


composer/core/evaluator.py
```diff
@@ -39,6 +41,23 @@ def should_eval(state: State, event: Event):
         if int(eval_interval) <= 0:
             return False
 
+        # if requested, evaluate at the end of training, as long as the length of training is specified.
+        if eval_at_fit_end and state.max_duration is not None:
```
ravi-mosaicml, line 45: Is it possible to use `Event.FIT_END` to simplify the logic?

composer/core/evaluator.py
```diff
@@ -39,6 +41,23 @@ def should_eval(state: State, event: Event):
...
+            # check if we are at the end of training
+            # handle different eval_intervals in order to not duplicate validation steps
+            if eval_interval.unit == TimeUnit.EPOCH and event == Event.EPOCH_END:
```
ravi-mosaicml, line 48: Could this logic be refactored to work if `max_duration` is specified in tokens or samples? (In addition, I'd prefer to avoid relying on `dataloader_len`, as it can be an approximation and is not always defined.)

I think when using `Event.FIT_END` the logic could be something like:
```python

last_eval_batch = None

def should_eval(state, event):
  nonlocal last_eval_batch  # Not sure if this line is needed? It might be
  
  ... # existing eval logic. Make sure to update last_eval_batch appropriately.
  if event == Event.FIT_END and state.timestamp.batch != last_eval_batch and eval_at_fit_end:
    last_eval_batch = state.timestamp.batch
    return True
```

tests/trainer/test_trainer_eval.py
```diff
@@ -140,7 +140,7 @@ def test_eval_hparams(composer_trainer_hparams: TrainerHparams):
     composer_trainer_hparams.evaluators = [
         EvaluatorHparams(
             label="eval1",
-            eval_interval='3ep',  # should NEVER run, since we train just 2 epochs
+            eval_interval='3ep',  # will run, since eval_at_fit_end = True
```
ravi-mosaicml, line 143: Can you add a seperate test case for this, ideally using the non-hparams codepath? Makes sense that this change is needed, since this functionality changes the default behavior that hparams uses, but would be nice to have a seperate test for that flag specifically (perhaps parameterize the option)

composer/core/evaluator.py
```diff
@@ -18,12 +18,14 @@
...
 
 
-def evaluate_periodically(eval_interval: Union[str, Time, int]):
+def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end: bool = True):
     """Helper function to generate an evaluation interval callable.
 
     Args:
         eval_interval (str | Time | int): A :class:`.Time` instance or time string, or integer in epochs,
             representing how often to evaluate. Set to ``0`` to disable evaluation.
+        eval_at_fit_end (bool): Whether to evaluate at the end of training, regardless of `eval_interval`.
```
ravi-mosaicml, line 27: Can you also update the trainer and evaluator docstrings to denote that evaluation always happens at the end of training when specifying a time string? 

https://github.com/mosaicml/composer/blob/e85302b292bf4857e956e2844713afd61558c302/composer/trainer/trainer.py#L355

https://github.com/mosaicml/composer/blob/0f1624ed3d36aa7c032209f115b3af9ec9e0402d/composer/core/evaluator.py#L85

ravi-mosaicml (commented):
Thanks for adding this in! See comments

composer/core/evaluator.py
moinnadeem: > Is it possible to use Event.FIT_END to simplify the logic?

Let me try, I initially had an implementation that checked `if state.get_elapsed_duration==1.0` , but the following scenario happens. Imagine a case where `eval_interval=1ba` and `max_duration=Xep`, where X is an integer. Let's say that there is 5 batches in an epoch.

```
*model performs gradient updates*
should_eval runs: Time: 2ep, 11ba
*model performs gradient updates*
should_eval runs: Time: 2ep, 12ba
*model performs gradient updates*
should_eval runs: Time: 2ep, 13ba
*model performs gradient updates*
should_eval runs: Time: 2ep, 14ba
*model performs gradient updates*
should_eval runs: Time: 2ep, 15ba <--- A
should_eval runs: Time: 3ep, 15ba <--- B
```

Since there are no gradient updates between event A and B, we shouldn't duplicate evaluation, and should only run it once. I assume `Event.FIT_END` is called after event B, and in this scenario, my current impression is that would lead to running evaluation twice -- this is why I added logic to identify the last _batch_ trained if the `eval_interval` is batches (this won't be a problem if `eval_interval` is epochs). 

Does that make sense? Do you see a way to sidestep this problem and still use `Event.FIT_END`?

tests/trainer/test_trainer_eval.py
moinnadeem: Yup can do!

moinnadeem:
Thanks for the quick review @ravi-mosaicml! I'll get back to you tomorrow on these, going to update the `speed_monitor` to stop tracking validation time today.

tests/trainer/test_trainer_eval.py
moinnadeem: Done!

composer/core/evaluator.py
moinnadeem: This is done!

composer/core/evaluator.py
moinnadeem: Done!

moinnadeem:
@ravi-mosaicml I think this should be ready for another review!

A few thoughts / comments:
I don't _love_ the structure of `evaluate_periodically`, especially how it has `return {bool}` instead of conditional statements. 
 
However, we need to update the `last_batch_seen` at a place where we know we are evaluating. One possible way to mitigate this is by inverting the logic: create conditional statements where we would `return False`, and then at the end of the code, have:
```
return True
```

Do you think this is a problem?

Secondly, 
> Could this logic be refactored to work if max_duration is specified in tokens or samples? 
I assume this is no longer a problem since we have removed `max_duration` from the `should_eval` function?

composer/trainer/trainer.py
```diff
@@ -1515,21 +1507,25 @@ def _train_loop(self) -> None:
...
+        for evaluator in self.state.evaluators:
+            assert evaluator.eval_interval is not None, "eval_interval should have been set on __init__() or fit()"
+            assert evaluator.subset_num_batches is not None, "subset_num_batches should have been set on __init__() or fit()"
+            if evaluator.eval_interval(self.state, event):
+                self.eval(
+                    dataloader=evaluator.dataloader,
+                    dataloader_label=evaluator.label,
+                    subset_num_batches=evaluator.subset_num_batches,
+                    metrics=evaluator.metrics,
+                    log_level=LogLevel.EPOCH,
```
ravi-mosaicml, line 1527: Can you update the `log_level` here to match the event? E.g. epoch_end -> LogLevel.EPOCH; batch_end -> LogLevel.BATCH; fit_end -> LogLevel.FIT

composer/trainer/trainer.py
```diff
@@ -371,6 +371,10 @@ class Trainer:
             This ``eval_interval`` will apply to any :class:`.Evaluator` in ``eval_dataloader`` that does not specify
             an ``eval_interval`` or if a dataloader is passed in directly. This parameter has no effect if
             ``eval_dataloader`` is not specified.
+
+            When specifying ``eval_interval``, the evaluator(s) are also run at the ``Event.FIT_END`` if it doesn't
```
ravi-mosaicml, line 375:
```suggestion
            When specifying time string or integer for the ``eval_interval``, the evaluator(s) are also run at the ``Event.FIT_END`` if it doesn't
```

composer/core/evaluator.py
```diff
@@ -96,6 +110,13 @@ class Evaluator:
             If a callable, it should take two arguments (:class:`.State`, :class:`.Event`) and return a bool
             representing whether the evaluator should be invoked. The event will be either :attr:`.Event.BATCH_END`
             or :attr:`.Event.EPOCH_END`.
+
+            When specifying ``eval_interval``, the evaluator(s) are also run at the ``Event.FIT_END`` if it doesn't
+            evenly divide the training duration.
+
+        eval_at_fit_end (bool, optional):
```
ravi-mosaicml, line 117: To match the trainer signature, can we require that the `eval_interval`  incorporate the `eval_at_fit_end` flag? (E.g. if you want to set it to False, pass in `eval_interval=evaluate_periodically(..., eval_at_fit_end=False)`?

composer/core/evaluator.py
```diff
@@ -108,6 +129,7 @@ def __init__(
         metrics: Union[Metric, MetricCollection],
         subset_num_batches: Optional[int] = None,
         eval_interval: Optional[Union[int, str, Time, Callable[[State, Event], bool]]] = None,
+        eval_at_fit_end: bool = True,
```
ravi-mosaicml, line 132: (See above comment)
```suggestion
```

composer/core/evaluator.py
```diff
@@ -131,7 +154,7 @@ def eval_interval(self, eval_interval: Optional[Union[int, str, Time, Callable[[
         if eval_interval is None:
             self._eval_interval = None
         elif not callable(eval_interval):
-            self._eval_interval = evaluate_periodically(eval_interval)
+            self._eval_interval = evaluate_periodically(eval_interval, eval_at_fit_end=self.eval_at_fit_end)
```
ravi-mosaicml, line 157: (see above comment)

```suggestion
            self._eval_interval = evaluate_periodically(eval_interval)
```

ravi-mosaicml (approved):
LGTM; see comments

composer/trainer/trainer.py
moinnadeem: Done!

composer/core/evaluator.py
moinnadeem: Done!

composer/core/evaluator.py
moinnadeem: Done!

composer/core/evaluator.py
moinnadeem: Done!

composer/core/evaluator.py
ravi-mosaicml: (Implemented)

PR #1046, opened 2022-05-15
Change YAHP to YAPF in STYLE_GUIDE

moinnadeem:
The docs mention YAHP, but I think we mean YAPF for code formatting.


Averylamp (approved):
lgtm

PR #1048, opened 2022-05-16
Remove no schedulers warning

A-Jacobson:
`.fit` warns that no schedulers are set if they are set on `init`. switched warning to check for schedulers from `state.schedulers`.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -846,7 +846,7 @@ class Trainer:
         else:
             self._scheduler_step_frequency = TimeUnit.BATCH if step_schedulers_every_batch else TimeUnit.EPOCH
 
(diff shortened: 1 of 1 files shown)


ravi-mosaicml (commented):
Good catch, thanks! Could you remove the check from `__init__` entirely and just check on `fit()`? (It is not required that schedulers are specified on `__init__`)

I'd also be down to remove this warning entirely, too, as there are valid use cases for training without a scheduler.

A-Jacobson:
> I'd also be down to remove this warning entirely, too, as there are valid use cases for training without a scheduler.


True, feels weird to warn about default behavior. Removing these with your blessing.

ravi-mosaicml (approved):
LGTM

PR #1049, opened 2022-05-16
Notebook for using ffcv dataloaders with composer

dskhudia:
Same as title


hanlint (approved):
This is really well done! Hopefully ffcv does a release soon and we can remove that monkeypatch...

dskhudia:
> This is really well done! Hopefully ffcv does a release soon and we can remove that monkeypatch...

We can remove this with their next release. 

> 
> Can you also manually add this to the https://docs.mosaicml.com/en/stable/getting_started/notebooks.html page?

Done.

PR #1050, opened 2022-05-16
Test and Warning Cleanup

ravi-mosaicml:
- Treat all python warnings as errors in tests. Existing tests that were throwing warnings were either fixed (yay!) or modified with `@pytest.mark.filterwarnings`
- In BlurPool, throwing an error if both `replace_maxpools` and `replace_convs` are False, as that results in the algorithm being a no-op.
- In progressive resizing, explicitly set `recompute_scale_factor=False` to match pytorch default behavior from 1.6+ (and silence a warning)
-  Made the default optimizer warning in the trainer less verbose.
- Converted the bert yaml to have duration specified in terms of samples, to fix an issue where the warmup period, combine with max duration, was a noop.
- Moved `TestMetricSetter` to `tests/metrics` and renamed as `MetricSetterCallback`. Tests should not be importing from other test files.
- Removed TestWandBLogger, since the trainer tests do the same thing
- Explicitly set `recompute_scale_factor=False` when using progressive resizing to silence a PyTorch warning about the default behavior changing in torch 1.6
- Disabled tqdm progress bars in tests
- Use the `tmp_path` fixture in tests instead of `tmpdir`, as `tmpdir` is deprecated: https://docs.pytest.org/en/6.2.x/tmpdir.html
- Fix all pyright `reportUninitailizedInstanceVariable` warnings where an attribute was declared as a class attribute but neither set on the class nor in `__init__`
- Remove `webdatasets` from setup.py as the upstream version isn't being used except for type checking. Intead, using `# type: ignores` for the webdataset imports.
- Set `reportMissingImports=None` so `# type: ignore` comments are not required for optional dependencies. Any typos with missing imports will be caught when attempting to run tests when doing `pip install mosaicml[all]`


composer/datasets/webdataset_utils.py
```diff
@@ -15,11 +15,11 @@
 from tqdm import tqdm
 
 if TYPE_CHECKING:
-    from webdataset import WebDataset
+    from webdataset import WebDataset  # type: ignore
```
hanlint, line 18: this shouldn't throw a pyright error, curious the reason for this? Prefer not to litter the codebase with type ignores..

composer/yamls/models/bert-base.yaml
```diff
@@ -48,7 +48,7 @@ optimizer:
 schedulers:
   - linear_decay_with_warmup:
       t_warmup: 0.06dur
-max_duration: 7ep # Baseline is 256M samples, 7 epochs is ~280M samples
+max_duration: 256_000_000
```
hanlint, line 51: cc: @moinnadeem

tests/algorithms/test_torch_export.py
```diff
@@ -120,18 +120,20 @@ def test_surgery_torchfx_eval(name, surgery_method, input):
...
 @pytest.mark.timeout(10)
-def test_surgery_onnx(name, surgery_method, input, tmpdir):
+@pytest.mark.filterwarnings(
+    r"ignore:Converting a tensor to a Python .* might cause the trace to be incorrect:torch.jit._trace.TracerWarning")
+def test_surgery_onnx(name, surgery_method, input, tmp_path):
     """Tests onnx export and runtime"""
     pytest.importorskip("onnx")
-    import onnx
-    import onnxruntime as ort
+    import onnx  # type: ignore
```
hanlint, line 128: ditto comment from before

tests/callbacks/test_early_stopper.py
```diff
@@ -1,55 +1,14 @@
...
-from composer.core.callback import Callback
 from composer.core.time import Time, TimeUnit
-from composer.loggers import Logger
-from composer.trainer.devices.device import Device
 from composer.trainer.devices.device_cpu import DeviceCPU
 from composer.trainer.devices.device_gpu import DeviceGPU
-from tests.common import RandomClassificationDataset, SimpleModel, device
-
-
-class TestMetricSetter(Callback):
```
hanlint, line 19: I dont think this belongs in `common`, only really used for two specific tests. The threshold for putting in `common` should be high.

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -27,7 +27,7 @@ def _padding_for_filt_2d_same(filt: torch.Tensor):
         raise IndexError(f"Filter must have odd height; got {h}")
     if w % 2 == 0:
         raise IndexError(f"Filter must have odd width; got {w}")
-    return (h // 2, w // 2)
+    return int(torch.div(h, 2)), int(torch.div(w, 2))
```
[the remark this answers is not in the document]
ravi-mosaicml: > UserWarning: __floordiv__ is deprecated, and its behavior will change in a future version of pytorch. It currently rounds toward 0 (like the 'trunc' function NOT 'floor'). This results in incorrect rounding for negative values. To keep the current behavior, use torch.div(a, b, rounding_mode='trunc'), or for actual floor division, use torch.div(a, b, rounding_mode='floor').

https://ci.int.mosaicml.com/blue/organizations/jenkins/internal%2Fmosaicml%2Fcomposer/detail/PR-1050/10/tests/

composer/datasets/webdataset_utils.py
ravi-mosaicml: Same thing as here: https://github.com/mosaicml/composer/pull/1001/files#r866082405

tests/algorithms/test_torch_export.py
ravi-mosaicml: Same as https://github.com/mosaicml/composer/pull/1001/files#r866082405

tests/callbacks/test_early_stopper.py
ravi-mosaicml: OK; will move to `tests/metrics`

composer/algorithms/blurpool/blurpool_layers.py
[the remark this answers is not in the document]
hanlint: Looks like PyTorch plans to keep `floordiv`, and then raise an error for negative values before eventually correcting the `floordiv` behavior (https://github.com/pytorch/pytorch/issues/43874). Since the tensor shapes here can never be negative, suggest we keep the original syntax.

tests/callbacks/test_early_stopper.py
ravi-mosaicml: Moved to `tests/metrics`

composer/datasets/webdataset_utils.py
ravi-mosaicml: Removed the `# type: ignore`; switched the pyproject.toml to ignoring the error

tests/algorithms/test_torch_export.py
ravi-mosaicml: Removed the `# type: ignore`; switched the pyproject.toml to ignoring the error

composer/core/state.py
```diff
@@ -196,18 +196,6 @@ class State(Serializable):
         train_dataloader (Iterable): The training dataloader. (May be ``None`` if not training.)
     """
 
-    _dataloader: Optional[Iterable]
```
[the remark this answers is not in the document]
ravi-mosaicml: Pyright is complaining that `_dataloader` was not set in `__init__` or as a class attribute. They fixed this bug in a https://github.com/microsoft/pyright/releases/tag/1.1.245.

composer/trainer/trainer.py
```diff
@@ -738,7 +738,8 @@ def __init__(
         # optimizers and schedulers
         if not optimizers:
             optimizers = DecoupledSGDW(list(model.parameters()), lr=0.1)
-            warnings.warn(f"No optimizer was specified. Defaulting to {repr(optimizers)}")
+            # hard-coding the optimizer in the warning, as repr(optimizers) would print an annoying, multi-line warning
+            warnings.warn(f"No optimizer was specified. Defaulting to DecoupledSGDW(lr=0.1)")
```
eracah, line 742:
```suggestion
            warnings.warn(f"No optimizer was specified. Defaulting to {optimizers.__class__}(lr={optimizers.defaults['lr']})")
```

composer/algorithms/blurpool/blurpool_layers.py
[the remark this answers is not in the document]
ravi-mosaicml: Sorry, I'm confused -- it seems like the comment here suggests that users should use `torch.div` explicitly? https://github.com/pytorch/pytorch/issues/43874#issuecomment-965222876

> For now we encourage users to follow the deprecation note and replace uses of floor division with a call to div() and the appropriate rounding mode.

composer/trainer/trainer.py
ravi-mosaicml: Updated; thanks!

composer/yamls/models/bert-base.yaml
moinnadeem: I can review this in a bit -- this should be fine since it helps avoid a bug

moinnadeem: LGTM!

moinnadeem (approved):
LGTM on the BERT change only -- the BERT YAML overall is a bit outdated, so we will need to update to C4 soon as well.

hanlint (approved):
LGTM!

PR #1053, opened 2022-05-17
Speed Monitor Part 1: ignoring validation time from training time

### LINKED ISSUE #984: `resize_size` on `val_dataset` changes `wall_clock_train`
** Environment **

- image: `pytorch_internal`
- branch: `dev`
- Hardware (GPU, or instance type): r1z1 8x A100

** To reproduce

1. Run `composer -n 8 examples/run_composer_trainer.py -f composer/yamls/models/resnet50.yaml --val_dataset.imagenet.resize_size 232` (will need to provide your own dataset path/hparams)
2. Run `composer -n 8 examples/run_composer_trainer.py -f composer/yamls/models/resnet50.yaml`
3. Compare `wall_clock_time` for each run

## Expected behavior

`wall_clock_train` should be unaffected by changes to validation. Instead, setting `resize_size` (in this case to `232`) increases `wall_clock_train` by about 8%. [See W&B runs](https://wandb.ai/mosaic-ml/resnet-high-score/reports/Composer-ResNet50--VmlldzoxODc4NTMx#-validation-resizing-slows-wall_clock_train).

moinnadeem:
This refactors the `speed_monitor` to ignore validation time, and plots three quantities instead: (a) wall-clock time spent in training, (b) wall-clock time spent in validation, and (c) wall-clock time spent overall. 

Let's limit the scope of this PR to just fixing measuring validation on the current implementation. I agree that the speed monitor needs to be refactored, and can even get some of that in a separate PR today. However, I want to unblock others from starting runs tomorrow.

Correctness:
[Here's a WandB report](https://wandb.ai/mosaic-ml/speed_monitor_testing/reports/Ensuring-SpeedMonitor-Correctness--VmlldzoyMDIwMjE0) where I ablate the size of the validation set with three runs:
1) A baseline that trains for 1000 batches and evaluates every 1000 batches. Therefore, it runs evaluation once. 
2) A version that trains for 1000 batches and evaluates every 500 batches. Therefore, it runs evaluation twice. You should see that the time spent in `wall_clock/val` is roughly double the time that the baseline spends.
3) A version that trains for 750 batches and evaluates every 500 batches. Therefore, it runs evaluation once. You should see that the time spent in `wall_clock/train` is less than the baseline, but the time spent in `wall_clock/val` is the same as the baseline, and half the version that does it twice.

Finally, if one examines the dip in `samples/step`, then they can sanity check `wall_clock/val` against the seconds that are spent in the dip. The code should be correct.

There are quite a few improvements to this PR, such as:
1. Augment `samples/step` into `samples/train/step` and `samples/val/step`.
2. Augment `samples` into anything tracked by the Time object, such as batches, samples, or tokens.
3. Clean up the implementation
4. Rename `speed_monitor` to `throughput_monitor` or `timing_monitor`.

We should brainstorm general improvements to the `speed_monitor` during a meeting, and I'm happy to take those on in a future PR!

This should resolve #984

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -59,7 +59,7 @@ class SpeedMonitor(Callback):
     +-----------------------+-------------------------------------------------------------+
     |``wall_clock/val``     | Total elapsed validation time                               |
     +-----------------------+-------------------------------------------------------------+
-    |``wall_clock/total``   | Total elapsed time                                          |
+    |``wall_clock/total``   | Total elapsed time (wall_clock/train + wall_clock/val)      |
     +-----------------------+-------------------------------------------------------------+
 
     Args:
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -66,8 +66,19 @@ class SpeedMonitor(Callback):
     def __init__(self, window_size: int = 100):
         super().__init__()
         self.train_examples_per_epoch = 0
+
+        # log the total wall clock time that this program has been running for
+        self.wall_clock_total = 0.0
+
+        # log the total wall clock_time that has been spent in training
         self.wall_clock_train = 0.0
+
+        # log the total wall clock_time that has been spent in validation
+        self.wall_clock_val = 0.0
+
         self.epoch_start_time = 0.0
+        self.validation_start_time = 0.0
+        self.epoch_time_in_validation = 0.0
         self.batch_start_num_samples = None
         self.batch_end_times: Deque[float] = deque(maxlen=window_size + 1)  # rolling list of batch end times
         self.batch_num_samples: Deque[int] = deque(maxlen=window_size)  # rolling list of num samples in batch.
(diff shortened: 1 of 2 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -55,7 +55,11 @@ class SpeedMonitor(Callback):
     |                       | Number of samples processed per second (averaged over       |
     | ``throughput/epoch``  | an entire epoch)                                            |
     +-----------------------+-------------------------------------------------------------+
-    |``wall_clock_train``   | Total elapsed training time                                 |
+    |``wall_clock/train``   | Total elapsed training time                                 |
+    +-----------------------+-------------------------------------------------------------+
+    |``wall_clock/val``     | Total elapsed validation time                               |
+    +-----------------------+-------------------------------------------------------------+
+    |``wall_clock/total``   | Total elapsed time                                          |
```
growlix, line 62: Maybe
`Total elapsed time (wall_clock/val + wall_clock/train)`
?

growlix (approved):
Small doc comment. LGTM, thanks moin!

composer/callbacks/speed_monitor.py
moinnadeem: Done!

ravi-mosaicml (approved):
LGTM. This is outside the scope of this PR, but would it work if someone is training for less than a full epoch?

moinnadeem:
> This is outside the scope of this PR, but would it work if someone is training for less than a full epoch?

The `samples/step` would get recorded, but the training and validation total time would not. I agree that we should record this as a running counter, and should include this in a future PR!

PR #1054, opened 2022-05-17
Add Event.FIT_END to the training loop

moinnadeem:
I might be absolutely blind, but I see that we [run `Event.FIT_START`](https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L1365) in `_train_loop(...)` but never see `Event.FIT_END` actually run. If correct, then this PR aims to resolve that.


ravi-mosaicml (approved):
Good catch, thanks.

PR #1056, opened 2022-05-17
AutoYAHP Part 1: Cleanup the Algorithms for AutoYAHP

ravi-mosaicml:
This PR refactors the algorithms and tests as will be required by AutoYAHP. It does not depend on AutoYAHP itself (a future PR will remove the underlying hparam classes).

- Refactored algorithm tests to not depend on hparams
- Reformatted the factorize and selective backprop docstrings so they would be correctly parsed by auto-yahp
- Refactor algorithm_settings.py to not depend on hparams and to return a list of pytest.param objects for a pytest.mark.parametrize. This change makes it more re-usable since it now includes information about markers required for each algorithm.
- Moved the TestTrainerAlgorithms into tests/algorithms/test_algorithms_train.py, since it tests the individual algorithms, not the trainer, and thus should live in tests/algorithms.
- Add helper methods for scanning a module to discover subclass implementations, check that the registry contains an entry, and test that a class is constructible from yaml


tests/algorithms/test_algorithm_resumption.py
```diff
@@ -1,105 +1,106 @@
...
-        [None, "1ep", "ep{epoch}-rank{rank}", "ep2-rank{rank}", "latest-rank{rank}"
-        ],  # test randomized seed saving and symlinking
-        [42, "1ep", "ep{epoch}-rank{rank}", "ep3-rank{rank}", "ep5-rank{rank}"],  # test save at epoch end
+        ["1ep", "ep{epoch}-rank{rank}", "ep2-rank{rank}", "latest-rank{rank}"],  # symlinking
+        ["1ep", "ep{epoch}-rank{rank}", "ep3-rank{rank}", "ep5-rank{rank}"],  # test save at epoch end
     ],
 )
-@pytest.mark.filterwarnings(r"ignore:.*Detected call of \`lr_schedule.*:UserWarning")  # Needed for SWA
-@pytest.mark.parametrize("algorithm", ALGORITHMS)
+@pytest.mark.parametrize("alg_cls,alg_kwargs,model,dataset", get_algorithm_parametrization(), indirect=['model'])
```
hanlint, line 40: Could we retain the option to select certain algorithms? This is useful for testing the resumption of individual algos during test time

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -1,105 +1,106 @@
...
 def test_algorithm_resumption(
-    algorithm: str,
     device,
-    seed: Optional[int],
     save_interval: int,
     save_filename: str,
     resume_file: str,
     final_checkpoint: str,
     tmp_path: pathlib.Path,
+    alg_cls: Callable[..., Algorithm],
```
hanlint, line 48: `Type[Algorithm]`?

tests/algorithms/test_algorithm_resumption.py
hanlint: I would prefer the previous pattern of retrieval rather than have four bundled parameterizations in the unit tests. More modularized that way.

tests/common/__init__.py
```diff
@@ -1,12 +1,22 @@
...
                                    configure_dataset_hparams_for_synthetic)
 from tests.common.events import EventCounterCallback, EventCounterCallbackHparams
+from tests.common.hparams import assert_is_constructable_from_yaml, assert_registry_contains_entry
 from tests.common.markers import device, world_size
 from tests.common.models import (SimpleConvModel, SimpleConvModelHparams, SimpleModel, SimpleModelHparams,
                                  configure_model_hparams_for_synthetic)
 from tests.common.state import assert_state_equivalent
 
+
+def get_all_subclasses_in_module(module: types.ModuleType, cls: Type) -> List[Type]:
```
hanlint, line 15: nit: can omit the `all` for brevity, or even better `get_module_subclasses`

tests/common/hparams.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def assert_registry_contains_entry(subclass: Type, registry: Dict[str, Callable], ignore_list: Sequence[Callable] = ()):
+    """Assert that the ``registry`` contains """
+    registry_entries = set(registry.values())
+    registry_entries.union(ignore_list)
+    assert subclass in registry_entries, f"Class {type(subclass).__name__} is missing from the registry."
+
+
+def assert_is_constructable_from_yaml(
```
hanlint, line 14: `assert_yaml_builds`?

tests/common/hparams.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Callable, Dict, Optional, Sequence, Type
+
+import yahp as hp
+import yaml
+
+
+def assert_registry_contains_entry(subclass: Type, registry: Dict[str, Callable], ignore_list: Sequence[Callable] = ()):
```
hanlint, line 7: `assert_in_registry`?

hanlint (approved):
LGTM, one comment on the algorithm_settings pattern, but that is not blocking and could be cleaned up later.

tests/common/hparams.py
ravi-mosaicml: Makes sense; decided to go with `assert_yaml_loads` instead (`builds` seems a _bit_ too much for what we're doing with the yaml, since we're not necessarily calling `.initialize_object()` here.)

tests/algorithms/test_algorithm_resumption.py
ravi-mosaicml: If typed as `Type[Algorithm]`, then pyright complains that the base `Algorithm` class takes no arguments. I think I'd need to use a `TypeVar(bound=Algorithm)` to get around that, so would prefer to leave it as a callable.

tests/common/hparams.py
ravi-mosaicml: Done

tests/common/__init__.py
ravi-mosaicml: Done

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -1,105 +1,106 @@
...
+from composer.algorithms.stochastic_depth.stochastic_depth import StochasticDepth
+from composer.core.algorithm import Algorithm
+from composer.models.base import ComposerModel
+from tests.algorithms.algorithm_settings import get_algorithm_parametrization
 from tests.common import deep_compare, device
 
-ALGORITHMS = get_algorithm_registry().keys()
+
+@pytest.fixture
+def model(request):
```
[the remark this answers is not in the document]
ravi-mosaicml: I refactored this test to remove the extra parameterization -- testing that symlink is working doesn't belong here but rather is in the checkpoint saver tests. So, this extra deepcopy is no longer needed.

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -1,105 +1,106 @@
...
+from torch.utils.data import DataLoader, Dataset
 
 from composer import Trainer
-from composer.algorithms import get_algorithm_registry
-from tests.algorithms.algorithm_settings import get_settings
+from composer.algorithms.factorize.factorize import Factorize
+from composer.algorithms.layer_freezing.layer_freezing import LayerFreezing
+from composer.algorithms.sam.sam import SAM
+from composer.algorithms.squeeze_excite.squeeze_excite import SqueezeExcite
+from composer.algorithms.stochastic_depth.stochastic_depth import StochasticDepth
```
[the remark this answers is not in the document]
ravi-mosaicml: (I wish the VSCode auto-import helpers picked the high-level import, not the low-level one :facepalm: ... maybe there's some sort of precommit hook that can automatically clean this up? That would be nice haha)

tests/algorithms/algorithm_settings.py
```diff
@@ -5,7 +5,35 @@
 Each algorithm is keyed based on its name in the algorithm registry.
 """
 
-from composer.algorithms import algorithm_registry
+from typing import Any, Dict, Optional, Type
+
+import pytest
+
+import composer
+from composer.algorithms.agc import AGC
```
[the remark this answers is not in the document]
ravi-mosaicml: Fixed, thanks!

tests/algorithms/test_algorithm_resumption.py
ravi-mosaicml: Makes sense; I like this approach. I will rename `get_algorithm_parametrization` to `get_algs_and_marks`, then add back the retrieval functions for `get_alg_kwargs`, `get_alg_model`, and `get_alg_dataset` that individual tests can call.

tests/algorithms/test_algorithm_resumption.py
ravi-mosaicml: I added in a stub init signature into the parent class for `Algorithm`, so then `Type[Algorithm]` would work.

tests/algorithms/test_algorithm_resumption.py
ravi-mosaicml: (Individual test files can filter the list returned by `get_algs_and_marks` before it is passed to `pytest.mark.parametrize` -- see the torchscript tests as an example)

ravi-mosaicml: Refactored

PR #1058, opened 2022-05-18
Improve download logic.

knighton:
- Move retries up to the main entrypoint, so they happen on all handlers
- Workers are all downloading at once, so no way to set timeout accurately
- boto already uses a randomly-named tempfile then swaps in, so rely on that instead of rolling our own
- If file already exists when swapping in, raises FileNotFoundError, so catch that

### DIFF
--- a/composer/datasets/streaming/download.py
+++ b/composer/datasets/streaming/download.py
@@ -6,7 +6,6 @@
 
 import os
 import shutil
-import tempfile
 import textwrap
 from time import sleep, time
 from urllib.parse import urlparse
@@ -31,13 +30,12 @@ def wait_for_download(local: str, timeout: float = 60) -> None:
         sleep(0.1)
 
 
-def download_from_s3(remote: str, local: str, timeout: float) -> None:
+def download_from_s3(remote: str, local: str) -> None:
     """Download a file from remote to local.
 
     Args:
         remote (str): Remote path (S3).
         local (str): Local path (local filesystem).
-        timeout (float): How long to wait for shard to download before raising an exception.
     """
     try:
         import boto3  # type: ignore (third-party)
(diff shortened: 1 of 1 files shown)


composer/datasets/streaming/download.py
```diff
@@ -80,72 +61,32 @@ def download(remote: str, local: str, timeout: float) -> None:
...
-        # The tmp file is old, it is either (1) from a crashed run or (2) another thread is downloading it but is taking too long, and will timeout.
-        # Let's delete the tmp file. If situation (1), this is safe. If situation (2), the other thread is expected to crash with a TimeoutError anyways, so this is fine.
+    ok = False
+    for _ in range(3):
         try:
-            os.remove(local_tmp)
-        except OSError:
-            # This occurs if another download thread got to the delete first.
+            download(remote, local, timeout)
+            ok = True
```
abhi-mosaic, line 88: Do we need this `ok` flag?

composer/datasets/streaming/download.py
```diff
@@ -80,72 +61,32 @@ def download(remote: str, local: str, timeout: float) -> None:
...
+            break
+        except:
```
abhi-mosaic, line 90: Could we specific about the type of Error that might be thrown here? Also, if we fail at this point are we supposed to continue or abort? Would appreciate a comment here.

composer/datasets/streaming/download.py
```diff
@@ -80,72 +61,32 @@ def download(remote: str, local: str, timeout: float) -> None:
...
-            try:
-                wait_for_download(local, timeout)
-                return
-            except TimeoutError:
-                pass
-
-        # The tmp file is old, it is either (1) from a crashed run or (2) another thread is downloading it but is taking too long, and will timeout.
-        # Let's delete the tmp file. If situation (1), this is safe. If situation (2), the other thread is expected to crash with a TimeoutError anyways, so this is fine.
+    ok = False
+    for _ in range(3):
```
abhi-mosaic, line 85: Could we make this `3` a `max_retries` parameter and potentially lift it up to the `StreamingDataset`?

abhi-mosaic (commented):
Nice! This looks like much simpler logic. @knighton could you confirm (as a comment in the code) why we are safe when two workers try to `download` to the same filename? Also, does this mean that we may potentially download some shards multiple times? I understand that this would probably be a small # of shards but just want to confirm.

The other comments are mostly cosmetic stuff.

Once fixed could you try running this test (it is normally gated by the `remote` mark and does not run in CI) and make sure that all cases still pass? Happy to approve after.

```
pytest -m remote -s "tests/datasets/test_streaming_remote.py::test_streaming_remote_dataloader"
```

composer/datasets/streaming/download.py
knighton: Yep, is download success flag

composer/datasets/streaming/download.py
knighton: Added

composer/datasets/streaming/download.py
knighton: Done

abhi-mosaic:
Closing this and going with #1074 to remove download conflicts.

PR #1069, opened 2022-05-19
W & B + Autoresumption Checkpointing Examples

ravi-mosaicml:
When writing the release notes, I wrote an example for how to save and restore checkpoints from the W & B Logger and how to use autoresumption. These examples don't have any test coverage (need to figure that out later -- we don't have any test coverage for the `examples/` folder), but I did run it locally, and it works.


eracah (commented):
It doesn't seem to work for me. I get `AttributeError: 'WandBLogger' object has no attribute 'get_object_size'` error

ravi-mosaicml:
> It doesn't seem to work for me. I get `AttributeError: 'WandBLogger' object has no attribute 'get_object_size'` error

Huh, strange. Can you share the command you ran and the stack trace? Very confused about this error, as that would imply this check is returning True (which it is not):

https://github.com/mosaicml/composer/blob/f5f02ed86acfbdeb8a09955122d40d0f261832f8/composer/utils/file_helpers.py#L342

```python
>>> isinstance(WandBLogger(), ObjectStore)
False
```

Perhaps there is some monkey-patching going on?

eracah:
> > It doesn't seem to work for me. I get `AttributeError: 'WandBLogger' object has no attribute 'get_object_size'` error
> 
> Huh, strange. Can you share the command you ran and the stack trace? Very confused about this error, as that would imply this check is returning True (which it is not):
> 
> https://github.com/mosaicml/composer/blob/f5f02ed86acfbdeb8a09955122d40d0f261832f8/composer/utils/file_helpers.py#L342
> 
> ```python
> >>> isinstance(WandBLogger(), ObjectStore)
> False
> ```
> 
> Perhaps there is some monkey-patching going on?

Ok! Fixed by uninstalling then re-installing composer. Sorry for the inconvenience.

eracah (approved):
Tested both scripts and they work.

examples/checkpoint_autoresume.py
```diff
@@ -0,0 +1,76 @@
...
+    train_dataloader=train_dataloader,
+    train_subset_num_batches=5,  # For this example, limit each epoch to 5 batches
+
+    # Evaluation Configuration
+    eval_dataloader=eval_dataloader,
+    eval_subset_num_batches=5,  # For this example, limit evaluation to 5 batches
+
+    # Checkpoint Configuration
+    run_name=run_name,
+    save_folder="./my_autoresume_training_run",
```
A-Jacobson, line 63: do the `run_name` and the `save_folder` name have to be the same?

PR #1070, opened 2022-05-19
Add PolynomialWithWarmupScheduler

dskhudia:
Adds PolynomialWithWarmupScheduler and PolynomialWithWarmupSchedulerHparams

This decays learning rate multiplier according to a polynomial of given power. A warmup period can be specified. 

Corresponding docs are updated and added a test.


composer/optim/scheduler.py
```diff
@@ -22,6 +22,7 @@
 
 from composer.core import State, Time, TimeUnit
 from composer.core.types import PyTorchScheduler
+import pdb
```
hanlint, line 25:
```suggestion
```

composer/optim/scheduler.py
```diff
@@ -336,6 +337,7 @@ def __init__(self, alpha_i: float = 1.0, alpha_f: float = 0.0, t_max: Union[str,
         self.t_max = Time.from_timestring(t_max) if isinstance(t_max, str) else t_max
 
     def __call__(self, state: State, ssr: float = 1.0):
+        #pdb.set_trace()
```
hanlint, line 340:
```suggestion
```

hanlint (approved):
LGTM

PR #1071, opened 2022-05-20
AutoYAHP Part 2: Cleanup the Callbacks and Loggers for AutoYAHP

ravi-mosaicml:
Similar to #1056, this PR cleans up the callbacks and loggers for AutoYAHP. It does not depend on AutoYAHP itself (a future PR will remove the underlying hparam classes).

- Refactored callback and logger tests to not depend on hparams
- Reformatted the docstrings so they would be correctly parsed by auto-yahp
- Added `callback_settings.py`, similar to `algorithm_settings.py`, to return a list of pytest param objects for parameterization across callback tests. These param objects include appropriate markers (e.g. conditional skipping for wandb and mlperf; requiring that the memory monitor runs on GPU, ...)
- Moved the `TestTrainerAssets` into `tests/callbacks/test_callbacks.py`, since it tests the individual callbacks and loggers, not the trainer, and thus should live in `tests/callbacks`.
- Cleaned up the `MemoryMonitor` warnings when cuda is not available. Now, it warns when the model is not on cuda, to catch the edge case where one does CPU training when GPUs are available.


composer/callbacks/memory_monitor.py
```diff
@@ -76,12 +76,11 @@ class MemoryMonitor(Callback):
...
-        super().__init__()
-        log.info(
-            "Memory monitor just profiles the current GPU assuming that the memory footprint across GPUs is balanced.")
-        if torch.cuda.device_count() == 0:
-            log.warning("Memory monitor only works on GPU devices.")
+    def init(self, state: State, logger: Logger) -> None:
+        model_device = next(state.model.parameters()).device
+
+        if model_device.type != 'cuda':
+            raise RuntimeError("The memory monitor only works on CUDA devices, but the model is on CPU.")
```
hanlint, line 83: I think the motivation to keeping this as a warning is that often users will switch between CPU and GPU for dev work, and don't want to be editing their code to add/remove this as needed. what do you think?

composer/loggers/wandb_logger.py
```diff
@@ -36,15 +42,22 @@ class WandBLogger(LoggerDestination):
...
-        init_params (Dict[str, Any], optional): Parameters to pass into
+        init_kwargs (Dict[str, Any], optional): Any additional init kwargs
             ``wandb.init`` (see
             `WandB documentation <https://docs.wandb.ai/ref/python/init>`_).
     """
 
-    def __init__(self,
-                 log_artifacts: bool = False,
-                 rank_zero_only: bool = True,
-                 init_params: Optional[Dict[str, Any]] = None) -> None:
```
hanlint, line 47: isn't it better to bundle all `wandb` init params together, rather than have some that are in the WandBLogger constructor, but others that must be appsed via init_params?

tests/callbacks/callback_settings.py
```diff
@@ -0,0 +1,177 @@
+from typing import Any, Callable, Dict, NamedTuple, Optional, Sequence
+
+import pytest
+
+import composer.callbacks
+import composer.loggers
+import composer.profiler
+from composer import Callback
+from composer.callbacks import EarlyStopper, MemoryMonitor, ThresholdStopper
```
hanlint, line 9: clean up these imports by concatentaing all the callback imports together?

tests/callbacks/test_callbacks.py
```diff
@@ -1,24 +1,30 @@
...
-from composer.callbacks.early_stopper import EarlyStopper
-from composer.callbacks.mlperf import MLPerfCallback
-from composer.callbacks.threshold_stopper import ThresholdStopper
+from composer.callbacks.callback_hparams import CallbackHparams, callback_registry
 from composer.core import Event
 from composer.core.callback import Callback
 from composer.core.engine import Engine
 from composer.core.state import State
-from composer.loggers import Logger, ObjectStoreLogger, WandBLogger
+from composer.loggers import Logger
```
hanlint, line 14: consolidate imports? with there was a package to do this automatically.. find the min hop import path..

hanlint (commented):
LGTM, a few comments

composer/loggers/wandb_logger.py
ravi-mosaicml: The original reasoning for listing out some of the wandb imports was so they could be set via CLI overrides with yahp. YAHP does not support setting dictionaries via CLI args. I could leave it as it was before, or replace these parameters with the dict, or leave it like it is now -- don't have strong preferences here. Thoughts?

tests/callbacks/callback_settings.py
ravi-mosaicml: Done

tests/callbacks/test_callbacks.py
ravi-mosaicml: Interview question? :D

Done

composer/callbacks/memory_monitor.py
ravi-mosaicml: True, though part of me wants invalid configurations to be an error. By dev work, would this be external users using composer or devs contributing to composer? If the former, then agree it should be a warning; if the latter, then I think it should be an error, since external ux >> developer ux.

tests/loggers/test_object_store_logger.py
```diff
@@ -9,7 +9,8 @@
 
 from composer.core.event import Event
 from composer.core.state import State
-from composer.loggers import Logger, LogLevel, ObjectStoreLoggerHparams
+from composer.loggers import Logger, LogLevel
+from composer.loggers.logger_hparams import ObjectStoreLoggerHparams
```
eracah, line 13: Still depends on Hparams tho? I thought this PR was supposed to remove dependencies to hparams?

eracah (commented):
Looks pretty good! Just had a few questions I wanted answered.
Thanks!

composer/callbacks/memory_monitor.py
```diff
@@ -76,24 +76,25 @@ class MemoryMonitor(Callback):
...
-            "Memory monitor just profiles the current GPU assuming that the memory footprint across GPUs is balanced.")
-        if torch.cuda.device_count() == 0:
-            log.warning("Memory monitor only works on GPU devices.")
+    def fit_start(self, state: State, logger: Logger) -> None:
+        # TODO(ravi) move this check would be on Event.INIT after #1084 is merged
+        # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
+        model_device = next(state.model.parameters()).device
+
+        if model_device.type != 'cuda':
+            raise RuntimeError(
```
eracah, line 85: From @A-Jacobson: 

> I don't like throwing an error here. Now a user will specifically have to delete this callback or their run still stop if they use the same yaml or script on cpu. I'd prefer to disable it for them in that case rather than forcing code/config chances.

composer/callbacks/memory_monitor.py
```diff
@@ -108,15 +109,8 @@ def after_train_batch(self, state: State, logger: Logger):
 
 
 def _get_memory_report() -> Dict[str, Union[int, float]]:
-    if not torch.cuda.is_available():
-        log.debug("Cuda is not available. The memory report will be empty.")
```
eracah, line 112: from @A-Jacobson: 

> same as my previous comment, I consider this a breaking change.

composer/callbacks/memory_monitor.py
```diff
@@ -108,15 +109,8 @@ def after_train_batch(self, state: State, logger: Logger):
...
-        return {}
     memory_stats = torch.cuda.memory_stats()
 
-    if len(memory_stats) == 0:
-        log.debug("No GPU memory was used, returning empty.")
```
eracah, line 117: Here too I guess?

composer/loggers/wandb_logger.py
ravi-mosaicml: OK, I will leave this as is!

composer/callbacks/memory_monitor.py
ravi-mosaicml: Sounds good; I will revert to a warning.

examples/checkpoint_with_wandb.py
```diff
@@ -16,7 +16,7 @@
 # The project name must be deterministic, so we can restore from it
 wandb_logger = WandBLogger(
     log_artifacts=True,
-    init_params={
+    init_kwargs={
         'project': 'my-wandb-project-name',
```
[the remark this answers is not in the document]
ravi-mosaicml: Good point, thanks!

tests/loggers/test_object_store_logger.py
ravi-mosaicml: I decided to leave the `ObjectStoreLoggerHparams`, as it does some special parsing to load secrets from env variables (to discourage storing of secrets in config).

tests/trainer/test_trainer.py
```diff
@@ -904,132 +901,6 @@ def test_data_not_augmented(self, config):
...
-        config['loggers'] = [logger]
-        trainer = Trainer(**config)
-        self._test_multiple_fits(trainer)
-
-    def _test_multiple_fits(self, trainer):
-        trainer.fit()
-        trainer.state.max_duration *= 2
-        trainer.fit()
-
-
```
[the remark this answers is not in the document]
ravi-mosaicml: Yes; see:

https://github.com/mosaicml/composer/blob/45ad601c4526a84e3a4337cdaea41a6e63c6bd5f/tests/callbacks/test_callbacks.py#L127

This new test doesn't distinguish between callbacks and loggers, so it appears shorter, but it does the same thing.

composer/callbacks/memory_monitor.py
ravi-mosaicml: Made into a warning.

composer/callbacks/memory_monitor.py
ravi-mosaicml: I updated the caller function emit a warning, so I don't think another statement here is necessary.

composer/callbacks/memory_monitor.py
ravi-mosaicml: The new code path doesn't log the memory report unless if the model is on gpu, so it's more or less equivalent.

composer/loggers/wandb_logger.py
```diff
@@ -98,17 +127,17 @@ def init(self, state: State, logger: Logger) -> None:
...
         # Adjust name and group based on `rank_zero_only`.
         if not self._rank_zero_only:
-            name = self._init_params["name"]
-            group = self._init_params.get("group", None)
-            self._init_params["name"] = f"{name} [RANK_{dist.get_global_rank()}]"
-            self._init_params["group"] = group if group else name
+            name = self._init_kwargs["name"]
+            group = self._init_kwargs.get("group", None)
+            self._init_kwargs["name"] = f"{name} [RANK_{dist.get_global_rank()}]"
+            self._init_kwargs["group"] = group if group else name
```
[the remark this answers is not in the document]
ravi-mosaicml: Done, thanks! (Still needed to capture the `name`, since it is used to set the `group`)

composer/callbacks/memory_monitor.py
ravi-mosaicml: Seems like the popular opinion is to leave it as a warning :)

composer/loggers/wandb_logger.py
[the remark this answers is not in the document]
eracah: Oh good call. As long as we don't want group to include the new "name" with the global rank init then this looks good.

eracah:
Ok. LGTM

eracah (approved):
LGTM

PR #1072, opened 2022-05-20
AutoYAHP Part 3: Refactor the models, datasets, and trainer_hparams

ravi-mosaicml:
This PR cleans up the model hparams, dataset hparams, and trainer hparams for AutoYAHP. It does not depend on AutoYAHP itself (a future PR will remove the underlying hparam classes).

- Fix invalid type annotations in hparam classes (Yahp 1.1 has stricter type settings)
- Refactor the trainer hparams and trainer to have the same argument names. Specifically, this meant renaming "optimizer" to "optimizers" and "deepspeed" to "deepspeed_config" in the trainer hparams, and "load_strict" to "load_strict_model_weights" in the Trainer. Updated the YAMLs as appropriate.
- Removed the device and precision settings from the YAMLs, as the trainer now uses smart defaults (AMP + GPU if cuda is available; otherwise FP32 + CPU). This is necessarry as AutoYAHP will parse the yamls into the `DeviceGPU()` / `DeviceCPU()` instances directly, and `DeviceGPU()` throws an error if cuda is not available (before, yahp would parse into `DeviceGPUHparams()`, which could then be replaced with `DeviceCPUHparams()`). (While it would have also worked to patch the yaml directly, I thought it'd be generally better to make the yamls more portable).


hanlint:
I'll defer reviewing this until the others are merged for easier diff-ing. 

> Removed the device and precision settings from the YAMLs, as the trainer now uses smart defaults (AMP + GPU if cuda is available; otherwise FP32 + CPU). 

I think we want to avoid smart defaults here, e.g. the default should always be FP32 + CPU. There are for example, use cases for FP32 + GPU (or bf16). If cuda is available, users may still only want to occupy the CPU (e.g. something else might be running on that GPU while they do debug), so prefer we leave the device setting very explicit.

ravi-mosaicml:
> I'll defer reviewing this until the others are merged for easier diff-ing.
> 
> > Removed the device and precision settings from the YAMLs, as the trainer now uses smart defaults (AMP + GPU if cuda is available; otherwise FP32 + CPU).
> 
> I think we want to avoid smart defaults here, e.g. the default should always be FP32 + CPU. There are for example, use cases for FP32 + GPU (or bf16). If cuda is available, users may still only want to occupy the CPU (e.g. something else might be running on that GPU while they do debug), so prefer we leave the device setting very explicit.

Hmm, I think always using CPU and FP32 would make a bad UX -- if someone is using the trainer on a machine that has a GPU, then by default I presume it should use a GPU device without having to make it explicit. FWIW, this is how PTL does it:

https://github.com/PyTorchLightning/pytorch-lightning/blob/29fe1dad155ee80e8934cf91004b4ea9cdbc455d/pytorch_lightning/trainer/connectors/accelerator_connector.py#L198-L199

For the cases where one wants to do cpu training on a gpu machine (the 1% case), would it be OK to require an additional parameter?

Before #948, the trainer hparams defaulted to gpu vs the functional API of the traiger used CPU. I think that is even worse; they should at least be consistent :)

For precision, PTL always defaults to FP32. I agree that using AMP precision is being opinionated, but also feel that the trainer should use smart defaults for best performance whenever possible.

Thoughts on how to go forward?
1. Undo the change in #948 and always use CPU + FP32, in both the functional api and hparams path. Require our yamls to make these explicit. (Not a fan of this option, as it would require an extra parameter for the 99% case where if one has a GPU, then it should be used).
2. Keep the devices and precisions specified in the yamls, but leave the smart defaults in the trainer. (Basically, revert these changes in this PR). This would make the YAMLs less portable, but can see that making sense if we want to view them more as "benchmarks" rather than "examples".
3. Auto-default the device, but always use FP32 precision unless otherwise specified (Match what PTL does)
4. Leave this PR as-is

composer/datasets/cifar.py
```diff
@@ -195,8 +195,8 @@ class CIFARWebDatasetHparams(WebDatasetHparams):
     height: int = hp.optional('Image height', default=32)
     width: int = hp.optional('Image width', default=32)
     n_classes: int = hp.optional('Number of output classes', default=0)
-    channel_means: List[float] = hp.optional('Mean per image channel', default=(0, 0, 0))
-    channel_stds: List[float] = hp.optional('Std per image channel', default=(0, 0, 0))
+    channel_means: List[float] = hp.optional('Mean per image channel', default_factory=lambda: [0, 0, 0])
+    channel_stds: List[float] = hp.optional('Std per image channel', default_factory=lambda: [0, 0, 0])
```
A-Jacobson, line 199: this seems odd to me, the type is specified as a list but default is a lamba function. is there a yahp specific reason why this needs to be this way?

composer/datasets/glue.py
```diff
@@ -95,6 +96,9 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
             raise MissingConditionalImportError(extra_deps_group="nlp", conda_package="transformers") from e
 
         self.validate()
+        assert self.task is not None, "checked in validate()"
+        assert self.tokenizer_name is not None, "checked in validate()"
```
A-Jacobson, line 100: these checks are already happening in `validate`. If this is necessary for autoyahp + temporary, I'm good with it. Otherwise can we choose one or the other?

A-Jacobson (approved):
Left a few comments asking about why the hparams classes are organized the way the are + my opinion on the device defaults. 

Overall LGTM, especially since many of these hparams classes will be leaving is soon (rest in peace you will not be missed).

composer/models/resnet/resnet_hparams.py
```diff
@@ -32,7 +32,6 @@ class ResNetHparams(ModelHparams):
     model_name: str = hp.optional(
         f"ResNet architecture to instantiate, must be one of {ComposerResNet.valid_model_names}. (default: '')",
         default='')
-    num_classes: int = hp.optional("Number of classes for the classification taks. (default: ``None``)", default=None)
```
eracah, line 35: Wait why is this removed?

composer/models/resnet/resnet_hparams.py
```diff
@@ -32,7 +32,6 @@ class ResNetHparams(ModelHparams):
...
     pretrained: bool = hp.optional("If true, use ImageNet pretrained weights. (default: ``False``)", default=False)
```
eracah, line 35:
```suggestion
    num_classes: Optional[str] = hp.optional("Number of classes for the classification taks. (default: ``None``)", default=None)
    pretrained: bool = hp.optional("If true, use ImageNet pretrained weights. (default: ``False``)", default=False)
```

composer/yamls/models/bert-base.yaml
```diff
@@ -52,14 +52,11 @@ max_duration: 256_000_000
 train_batch_size: 4000
 eval_batch_size: 2000
 seed: 19
-device:
```
eracah, line 55: why are `device` and `precision` removed?

eracah (approved):

composer/models/resnet/resnet_hparams.py
ravi-mosaicml: This field is also on the base `ModelHparams` class, so it didn't need to be duplicated (eventually, we should remove it from the base class and only have it here.)

composer/models/resnet/resnet_hparams.py
ravi-mosaicml: (See above comment)

composer/datasets/glue.py
ravi-mosaicml: > Definitely not `2`, a few of us have already run afoul on this while trying to debug runs on cpu. I don't think we can go wrong with 1, 3 or 4 as long as it's clear what is happening. My personal order or preference is 3. -> 4.  -> 1.  

Discussed offline; decided to go with option 4 (leave as is), as option 3 would require the yamls to specify `precision: amp`, which would defeat the portability.

ravi-mosaicml:
> Definitely not `2`, a few of us have already run afoul on this while trying to debug runs on cpu. I don't think we can go wrong with 1, 3 or 4 as long as it's clear what is happening. My personal order or preference is 3. -> 4.  -> 1.  

Discussed offline; decided to go with option 4 (leave as is), as option 3 would require the yamls to specify `precision: amp`, which would defeat the portability.

composer/datasets/cifar.py
ravi-mosaicml: Yeah, this is a bit unintuitive, but here's what's happening:

1. YAHP does not support tuples or generic sequences; only lists (at least not yet), so the type has to be `List[float]`
2. `hp.optional` now needs to return a list type instead of a tuple
3. Because lists are mutable, it is necessary to provide a factory, so each instance of the dataclass will get a different instance of the list. See https://docs.python.org/3/library/dataclasses.html#dataclasses.field

composer/datasets/glue.py
ravi-mosaicml: These additional checks are required since pyright cannot assume anything what `validate()` does.

But YAHP 0.1.1 will deprecate `validate()`. Instead, all parsing should be done in `initialize_object`. (This is why: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

composer/yamls/models/bert-base.yaml
ravi-mosaicml: > Removed the device and precision settings from the YAMLs, as the trainer now uses smart defaults (AMP + GPU if cuda is available; otherwise FP32 + CPU). This is necessarry as AutoYAHP will parse the yamls into the DeviceGPU() / DeviceCPU() instances directly, and DeviceGPU() throws an error if cuda is not available (before, yahp would parse into DeviceGPUHparams(), which could then be replaced with DeviceCPUHparams()). (While it would have also worked to patch the yaml directly, I thought it'd be generally better to make the yamls more portable).

PR #1074, opened 2022-05-20
Remove download conflicts from `StreamingDataset`

abhi-mosaic:
This PR removes all download conflicts between `StreamingDataset` workers:

* The `StreamingDatasetIndex`, which is built on `__init__` by each device's main thread, is only downloaded by the local rank0 device on each node. All other devices wait for the shard to arrive.
* We explicitly assign `part_shards_to_download` to each DataLoader worker, so that no two workers ever attempt to download the same shard.
  * For shards that overlap multiple workers, the worker at the end, the one responsible for the last sample, is the worker that downloads the shard. This is also the worker who needs the shard the "earliest", so it helps with performance.
  * The other worker(s) wait for the shard to arrive.
  * For performance reasons, if `shuffle=True`, the first shard is still always processed first, so that we don't accidentally make a worker wait a long time for its neighboring worker to download the shard that it needs.
* Since all downloads are conflict free, the `download.py` file is simplified considerably.

We also add some related improvements:
* `max_retries: int = 2`, which is lifted all the way up to `StreamingDataset`
* Download timeouts and retries are all done explicitly above any download handlers
* `tests/datasets/test_streaming.py::test_dataloader_multi_device` now has `multinode` tests


composer/datasets/streaming/dataset.py
```diff
@@ -180,34 +182,45 @@ def _done_loading(self) -> None:
...
             part_max_id (int): Maximum sample ID of this partition.
         """
-        shards = list(shards)
         if self.shuffle:
-            np.random.shuffle(shards)
-        for shard in shards:
+            # Always process first shard first because other workers may be waiting on it
+            first = part_shards[0:1]
+            rest = part_shards[1:]
+            np.random.shuffle(rest)
```
knighton, line 199: np.random.shuffle(part_shards[1:])

abhi-mosaic: Ah interesting...  I think this onlyworks if I convert to numpy array first, but will do that. Looks cleaner.

tests/datasets/test_streaming.py
```diff
@@ -224,33 +224,43 @@ def test_dataloader_single_device(remote_local: Tuple[str, str], batch_size: int
...
 @pytest.mark.parametrize("num_workers", [1, 3])
 @pytest.mark.parametrize("shuffle", [False, True])
-def test_dataloader_multi_device(remote_local: Tuple[str, str], batch_size: int, drop_last: bool, num_samples: int,
-                                 num_workers: int, shuffle: bool):
+def test_dataloader_multi_device(remote_local: Tuple[str, str], batch_size: int, drop_last: bool, multinode: bool,
+                                 num_samples: int, num_workers: int, shuffle: bool):
+
+    if multinode:
+        # Force different nodes
+        os.environ["LOCAL_RANK"] = str(0)
```
ravi-mosaicml, line 236: Curious if this test works? The distributed is configured before any individual test runs, and the dist helpers should complain if there is an environment variable mismatch:

https://github.com/mosaicml/composer/blob/0f1624ed3d36aa7c032209f115b3af9ec9e0402d/composer/utils/dist.py#L77-L80

I see it's marked as `@daily`, but curious if you tried to run it locally with any luck?

ravi-mosaicml: In addition, is it possible to use the `monkeypatch` fixture instead of setting these variables directly? `monkeypatch` will reset the variables back to their previous values after the test runs, so any other tests in the same process will have the original values.

ravi-mosaicml (commented):
I didn't give this too close a read, but overall logic seems fine. Main concern regarding the test case. Approving for now that this a a temporary patch until streaming datasets 2.0, where we will decide whether node local rank 0 will perform all downloads.

composer/datasets/streaming/format.py
```diff
@@ -331,4 +334,14 @@ def _get_min_max_size(start: int, total: int, part: int, num_parts: int):
...
         max_shard = self.sample_shards[worker_max_id]
         shards = list(range(min_shard, max_shard + 1))
-        return shards, worker_min_id, worker_max_id
+
+        # Ensure that each shard only gets downloaded by 1 worker, so there are no race conditions.
+        # To do this, we skip downloading the last shard (likely overlapped with next worker) unless:
+        # - you are the last worker on your node (no files shared across nodes so you have to download it again!)
+        # - you are downloading the last sample of the shard (no overlap with next worker)
+        if ((node_worker + 1 == node_num_workers) or
+            (worker_max_id + 1 < self.total_samples and self.sample_shards[worker_max_id + 1] != max_shard)):
```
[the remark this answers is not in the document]
abhi-mosaic: Yes, the `worker_max_id` should always be between [0, self.total_samples - 1], and then the check is just to make sure we don't have an indexing error. Technically, the check is unnecessary as the very last worker should early exit based on `(node_worker + 1 == node_num_workers)`. I wonder if I should remove the check which would be a better test, but a little less readable?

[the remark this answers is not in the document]
abhi-mosaic: Actually scratch that, there are some really obscure edge cases like `num_samples` = 3, `num_devices=4` where the last sample will be repeated and more than one worker will be reading the end of the range, and so the check is still needed.

tests/datasets/test_streaming.py
abhi-mosaic: Changing the `os.environ` does seem to work.. and I definitely saw different behavior with `multinode` on vs. off as I was debugging. I think maybe becasue I'm only changing the local rank and node rank, `dist` isn't complaining (world size is still 2). I will try to use the `monkeypatch` fixture as suggested.

composer/datasets/streaming/download.py
```diff
@@ -52,100 +33,68 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
...
+                start = time.time()
+                while not os.path.exists(local):
+                    if time.time() - start > timeout:
+                        raise TimeoutError(f"Waited longer than {timeout}s for other worker to download {local}.")
+                    time.sleep(0.25)
+            else:
+                dispatch_download(remote, local, timeout=timeout)
+            ok = True
+            break
+        except Exception as e:  # Retry for all causes of failure.
```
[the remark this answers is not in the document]
abhi-mosaic: It seems like this catch is overly broad because we don't know what errors the underlying download handler will be throwing. Like it may be more than just `TimeoutErrors`.

We should have a refactor coming in soon so that `dispatch_download` gets refactored to `composer.utils.file_helpers`. At that time I think we will have a bit more control over the errors and can make this exception catching stricter. Let me know if that's OK.

tests/datasets/test_streaming.py
abhi-mosaic: Yeah so to close the loop here on why that `RuntimeError` does not occur, it's because the vars I'm changing don't have an associated `fetch_fn_name` which is what `composer.utils.dist` requires to run that code block you posted .

composer/datasets/streaming/download.py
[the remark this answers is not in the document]
ravi-mosaicml: OK, fair point -- especially if we switch to libcloud, then the error handling might be different. But it does appear boto3 has some nice handling: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html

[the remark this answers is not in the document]
abhi-mosaic: Punting this as `libcloud` refactor should be imminent.

Issue #713, opened 2022-03-10
Reorganize algorithm docs


dblalock:
Our docs currently have 1) a lot of redundancy, 2) a difficult interface for finding much or most of our API surface area. This creates both confusion for readers and engineering burden for maintainers. 

Specific problems/challenges:

- **We write independent overviews of each algorithm in three different places.** This increases the barrier to contributing new algorithms and makes maintenance more difficult. E.g., for `BlurPool`:
    - Place 1: `composer.algorithms.blurpool` docstring
    - Place 2: `BlurPool` method card
    - Place 3: `composer.algorithms.blurpool.BlurPool` algorithm class (for most algos)
- Several algorithms have different interfaces. These interfaces show up in different places in the docs, and **only some interfaces are available without clicking through several levels of API reference.** As an example, blurpool has:
    - `BlurPool` the `Algorithm`, documented in:
       - `composer.algorithms` API reference
       - `composer.algorithms.blurpool` API reference
       - `composer.algorithms.blurpool.blurpool` API reference
    - `apply_blurpool` the all-in-one model surgery function
        - `composer.algorithms.blurpool` API reference
        - `composer.algorithms.blurpool.blurpool` API reference
    - `BlurConv2d` and company, documented in
        - `composer.algorithms.blurpool` API reference
        - `composer.algorithms.blurpool.blurpool_layers` API reference
    - `blur_pool2d` and other standalone functions for the core logic, documented in
         - `composer.algorithms.blurpool` API reference
         - `composer.algorithms.blurpool.blurpool` API reference
- Because we take up so much sidebar real estate with method cards and model cards, **any API not documented in one of these pages is really hard to find**; I have to click on "API reference" at the _very bottom_ (not even visible unless I scroll down a lot), and then repeatedly expand submodules and scroll down to find stuff.
    - E.g., it's super annoying to find our functional API right now, even if you know where to look. 
- Searching, especially for algorithms, is hard because we have so many pages with similar names and content. Which makes the recursive "API reference" clicking more mandatory.

### Algorithms
- **Prefix every algorithm directory with an underscore.** E.g., `composer.algorithms._blurpool`
    - This will eliminate redundant documentation of algorithm contents. Everything that should be visible under `composer.algorithms` will now *only* be visible under `composer.algorithms`. Same for `composer.functional`.
- **Make a top level `composer.layers` in which nn.modules like `BlurConv2d` can be discovered** without digging several layers deep in the API reference.
[...]

PR #1076, opened 2022-05-21
Cleaner api reference

dblalock:
### Problem

Right now, if a user looks at the docs for an algorithm like adaptive gradient clipping, they're taken to [an auto-generated page](https://docs.mosaicml.com/en/stable/api_reference/composer.algorithms.agc.agc.html?highlight=agc) with title:

# "composer.algorithms.agc.agc"

and shown a class by the name of `composer.algorithms.agc.agc.AGC`, as well as the function `composer.algorithms.agc.agc.apply_agc`.

This is not only ugly, but also shows users discouraged, quasi-private import paths. The docs should instead show the preferred import paths `composer.functional.apply_agc`and `composer.algorithms.AGC`, ideally in a less ugly way.

I used an algorithm as an example here, but this problem applies to essentially all public names in composer.

### What this PR does

- Makes it so that all public, documented names are shown with the shortest allowed import path, rather than the longest.
- Makes the "API reference" toctree 1) not collapsed by default and 2) only one level deep. 
    - This avoids the user having to repeatedly click to expand it (which is especially annoying since each click force-scrolls to the top of the page for extra inconvenience).
    - This is possible since AFAIK all names we intend for users to use are importable as `composer.name` or `composer.module.name`.
    - This means that all the contents of a module are on a single page with a sensible title and module docstring, with the autosummary tables at the top providing quick links to each member.

The semantic changes are just
1. altering `docs/source/conf.py`'s `determine_sphinx_path()` + some config vars, and 
2. adding a `docs/source/modules` directory with an rst file for each composer module we want to generate docs for. This replaces `api_reference.rst`.
3. reimporting some functions and `nn.Module`s from factorization, blurpool, and squeeze excite that were only available through the algorithm submodules in `composer.functional` and a new `composer.layers` (which we've kinda needed to do for a while anyway).

### Out of scope

 - Changing method cards that have their own autosummaries that generate duplicate docs (possibly with longer import paths). E.g., the SWA method card generates `composer.algorithms.swa.SWA`. Right now method cards aren't even consistent in whether they include the code at all, so I'm considering their content out of scope for this PR.
 - Changing code, including example code, to always use the shortest import paths.

### Potentially controversial behavior

- I included the names accessible through composer directly (e.g., `composer.Algorithm`) on the `composer.core` page, since most of them are re-exported from core and it seemed like the least bad solution.
- Some classes, like `Trainer`, `Callback`, and `ComposerModel` could be placed on this `composer.core` page or {`composer.trainer`, `composer.callbacks`, `composer.models`}, respectively. I put `Trainer` and `ComposerModel` on the latter since their docs are long and they're not from core.
- Probably broke some references. I know I broke the Event refs for most algorithms' match() (though we can regex that away in a subsequent PR).
- The composer.layers module is new, though admittedly only contains a `layers/__init__.py` file.
- Added some stuff like blur_2d to composer.functional that doesn't quite feel like rest of the functional API.
- Special cased composer.utils.{dist, reproducibility} to still show up in the top-level toctree, which is slightly ugly.
- A few submodules like `composer.core.engine` have useful module docstrings that aren't replicated in the imported classes and are now not shown in the docs.
- There are a bunch of sort-of-public names in composer.utils files that aren't re-exported in composer.utils (e.g., contents of `batch_helpers.py`, `collect_env.py`, etc.). **These are now omitted from the docs.** This is the part I'm least sure about.

Partially addresses #713. Immediate cause for looking at this was my experience trying to use our NLP stuff in an external repo and having to guess at what path I should actually be importing any given name from. (And this seemed like it would be a much more straightforward change initially...)


hanlint:
Thanks @dblalock this is indeed a very non-straightforward change, that pytorch, for example, sidesteps by just manually creating and maintaining the pages, e.g. since [nn.rst](https://raw.githubusercontent.com/pytorch/pytorch/master/docs/source/nn.rst),  since their API is more stable. 

I haven't fully reviewed yet, but I'd prefer not to make a `composer.layers` module. Each algorithm folder should be as portable as possible. We are also amidst moving models and benchmarks out of `composer` entirely.

dblalock:
ah, clarification: no code is actually moved to `composer.functional` or `composer.layers`. They're both just empty directories with an `__init__.py` that reimports some stuff. So algorithm modularity should be unaffected.

(This doesn't mean that this change is necessarily the right thing to do, but hopefully it's not as obviously the wrong thing.)

Sounds like maybe it would be good to leave this PR until the models and benchmarks have been removed? (I'll do this unless someone tells me not to)

hanlint:
Agreed @dblalock -- closing this for now, but let's re-open after we complete the schism movement.

PR #1078, opened 2022-05-23
Install Nvidia APEX in the CUDA docker images; remove manual creation of `/tmp`

ravi-mosaicml:
- Install [Nvidia APEX](https://github.com/NVIDIA/apex) in the CUDA docker images, per request from @moinnadeem 
- Remove the manual creation of the `/tmp` folder, as the underlying issue was fixed in https://github.com/mosaicml/testing/commit/e6f64064845ae48f12c3dc38d1e958864c53695a, and manually creating the `/tmp` folder has no effect


docker/pytorch/Dockerfile
```diff
@@ -175,11 +188,6 @@ RUN useradd -rm -d /home/mosaicml -s /bin/bash -u 1000 -U -s /bin/bash mosaicml
     usermod -a -G sudo mosaicml && \
     echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
 
-###################
```
ravi-mosaicml, line 178: This is coming along for the ride -- I think I figured out the underlying issue of why `/tmp` was not in the resulting Dockerimage (see https://github.com/mosaicml/testing/commit/2ac0bc4cf2314ab894b52eac974f9c15a3954723).

moinnadeem: That could be part of it, but also, should we make the permissions for the `/tmp` directory `1777` instead of `777`? See the sticky bit needing to be set here: https://unix.stackexchange.com/questions/71622/what-are-correct-permissions-for-tmp-i-unintentionally-set-it-all-public-recu

ravi-mosaicml: The permissions in the dockerfile by default should be fine. This fix seems to be working in the latest nightly build version:

```
(composer) ravi@dolphin:~/ravi-composer$ docker run --rm -it mosaicml/pytorch:latest_cpu
root@95a41c937ec2:/# ls -al /tmp
total 272
drwxrwxrwt 1 root root   4096 May 24 07:08 .
drwxr-xr-x 1 root root   4096 May 24 16:24 ..
drwx------ 2 root root   4096 May 24 07:07 tmp2x9d4u9k
-rw------- 1 root root 265969 May 24 07:07 tmp3s3g48fbcacert.pem
```

PR #1080, opened 2022-05-24
Enabled most pre-commit hooks

ravi-mosaicml:
- Enabled most pre-commit hooks, except the quote fixer, as that will result in a messy diff. Will wait till PRs are at a minimum before adding that one
- Reformatting existing files to be compliant with the pre-commit config. Mainly involved fixing links, adding newlines, and removing executable bits.


Averylamp (approved):
lgtm

PR #1081, opened 2022-05-24
Fix `grad_accum='auto'` microbatch override

abhi-mosaic:
This PR addresses an issue with `grad_accum: auto` due to Composer's internal `state.batch` overriding behavior.

Assume we are in a situation like `train_batch_size: 128`, on 8 GPUs, and `grad_accum: auto`, but in truth we can only fit device microbatches of size `1` (not known ahead of time). So we should expect `Trainer` keep catching OOMs and increasing `grad_accum` until it hits `grad_accum: 16`.

Here's what currently happens:

> We start with `grad_accum = 1` by default.
> On step 0 we have `self.state.batch` == "the device batch", of size 128 // 8 = 16.
> We enter the microbatching loop
> Then `self.state.batch` is overwritten to be "the device microbatch", which is size 16 // 1 = 16
> We OOM, catch it, double the `grad_accum=2`, and start again

> **Here is the bug**: On this attempt, we have `self.state.batch` == "the device microbatch", which is still 16
> We enter the microbatching loop
> Then `self.state.batch` is overwritten to be "the device microbatch", which is size 16 // 2 = 8
> We OOM, catch it, double the `grad_accum=4`, and start again

> **Here is the bug**: On this attempt, we have `self.state.batch` == "the device microbatch", **which is now 8**
> We enter the microbatching loop
> Then `self.state.batch` is overwritten to be "the device microbatch", which is size 8 // 4 = 2
> We OOM, catch it, double the `grad_accum=8`, and start again

> **Here is the bug**: On this attempt, we have `self.state.batch` == "the device microbatch", **which is now 2**
> We enter the microbatching loop
> Then `self.state.batch` is overwritten to be "the device microbatch", which is size 2 // 8 = ???
> We get an exception like "cannot split a batch of size 2 into 8 chunks"

I ran into this bug when attempting to use `grad_accum: auto` with some NLP workloads that require small device microbatch sizes. I've tested on this branch since and everything seems to work :)

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1587,9 +1587,9 @@ class Trainer:
         """
         assert self._train_data_spec is not None, "The train data spec should be set on __init__ or fit()"
 
-        # Cache the device minibatch, because `self.state.batch` gets overridden in microbatching loop
-        # TODO: fix this name collision!!
-        minibatch = self.state.batch
+        # Cache the device batch, because `self.state.batch` gets overridden in microbatching loop
+        # TODO: fix this name collision!
+        device_batch = self.state.batch
 
         # Retry until we successfully complete training and return loss
         while True:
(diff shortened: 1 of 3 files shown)


ravi-mosaicml (approved):
Lgtm, wow good catch

PR #1085, opened 2022-05-25
Stochastic depth that preserves weights

Landanjs:
This is an almost complete refactor of stochastic depth. I might have gone overboard with this... The main goal was to get stochastic depth to work on DeepLabv3, but I wound up cleaning up a lot of the code that I thought did not provide much benefit given the additional complexity.

### Main feature: stochastic depth preserves weight values

When applying stochastic depth to a module, the original weights should be identical. Stochastic depth changes the computation graph i.e. the forward function. The original stochastic depth implementation reinitialized the modules weights which was unnecessary and added a little bit of complexity.

This implementation only replaces the forward function of an already instantiated model. I thought the best way to do this is to bind the new forward function to the module using `__get__`.

Alternatively, we could reinitialize the model and transfer the weights, but this would not cleanup the implementation.

### Side question: cleaning stochastic depth tech debt

In addition to no longer instantiating a module, I removed a lot of the `use_same_gpu_seed` functionality. This feature is not usable in composer since each device now has a separate seed. This could still be used in a functional form, but given that we don't have a good way to test this feature's usefulness and that it didn't seem useful in the first place, I thought it would be best to remove it. This can be added as a part 2 PR if desired.

### Future

I think this implementation will be an intermediate step as we (Daya) figures out how to use torch.fx to apply the necessary stochastic depth surgery. Hopefully, this will be a cleaner implementation that would make the torch.fx development easier


hanlint:
@Landanjs is this ready for review?

Landanjs:
Not yet, I decided to add one more feature and need to do testing. My goal is Thursday/Friday, depending on schedule

Landanjs:
Unfortunately, the new stochastic depth is 0.1% - 0.2% worse on ImageNet. Comparing the codes, I don't notice any obvious differences. Unless someone would like to look it over with me, I will continue to look over it. If I can't find anything, I might need to backtrack and make smaller changes at a time

ravi-mosaicml:
OK, thanks for the update. Going to close this PR for now -- please re-open when it is ready!

Landanjs:
Okay, I think this is ready for review now. I will update the description with new info

dskhudia (approved):
LGTM!

PR #1089, opened 2022-05-25
Configured `pydocstyle` for the trainer, callbacks, loggers, optimizers, and utils.

ravi-mosaicml:
This PR enables the `pydocstyle` pre-commit hook for most of the codebase. Docstrings were brought up to compliance so they would pass the pydocstyle check.

* The main changes involved ensuring summary docstring lines were indeed just one line, that they ended with periods, and that there were no blank lines between the end of the docstring and the start of the function.
* Added docstrings for missing arguments that were identified.
* (Coming along for the ride): Remove calls to `trainer.engine.close`, as now `close()` is invoked automatically in `__del__` as part of #948 

Model, dataset, and algorithm docstrings will be addressed in future PRs.


eracah:
This PR is obv very large and mostly contains routine docstring edits. I won't look at each one; I will trust that pydoctest does what it claims to do. Any diffs that do other things that we should take special look at in this PR? Like the pre-commit config or anything else?

ravi-mosaicml:
> This PR is obv very large and mostly contains routine docstring edits. I won't look at each one; I will trust that pydoctest does what it claims to do. Any diffs that do other things that we should take special look at in this PR? Like the pre-commit config or anything else?

I know it's a bunch of files, but it would be super helpful if you could double check that there aren't any unintentional changes from a sloppy merge that I might have done (e.g. remove new functionality). Let me update it against the latest dev.

composer/callbacks/speed_monitor.py
```diff
@@ -90,13 +85,6 @@ def __init__(self, window_size: int = 100):
         self.loaded_state: Optional[Dict[str, Any]] = None
 
     def state_dict(self) -> Dict[str, Any]:
-        """Returns a dictionary representing the internal state of the SpeedMonitor object.
-
-        The returned dictionary is pickle-able via :func:`torch.save`.
-
-        Returns:
-            Dict[str, Any]: The state of the SpeedMonitor object
-        """
```
A-Jacobson, line 99: why are these docstrings being deleted?

composer/cli/__init__.py
```diff
@@ -1,2 +1,13 @@
...
+"""Composer CLI."""
+
+import sys
+
+from composer.cli.launcher import main
+
+__all__ = ["main"]
+
+if __name__ == "__main__":
+    sys.exit(main())
```
A-Jacobson, line 13: are these changes meant to be part of this docs pr?

composer/core/algorithm.py
```diff
@@ -42,8 +42,7 @@ def __init__(self, *args, **kwargs):  # Stub signature for PyRight
 
     @property
     def find_unused_parameters(self) -> bool:
-        """Return True to indicate that the effect of this algorithm may cause some model parameters to be unused.
-        Defaults to False.
+        """Whether this algorithm may cause some model parameters to be unused. Defaults to False.
```
A-Jacobson, line 45:
```suggestion
        """Indicates whether this algorithm may cause some model parameters to be unused. Defaults to False.
```

composer/core/algorithm.py
```diff
@@ -66,11 +65,9 @@ def backwards_create_graph(self) -> bool:
 
     @abstractmethod
     def match(self, event: Event, state: State) -> bool:
-        """Determines whether this algorithm should run given the current :class:`~.event.Event` and
-        :class:`~.state.State`.
+        """Whether this algorithm should run given the current :class:`~.event.Event` and :class:`~.state.State`.
```
A-Jacobson, line 68:
```suggestion
        """Determines whether this algorithm should run given the current :class:`~.event.Event` and :class:`~.state.State`.
```

I think removing the verb from these sentences makes them somewhat less clear.

composer/cli/__init__.py
ravi-mosaicml: Yes, this module needed an `__init__.py` file. I was thinking it makes sense to re-rexport the cli so one could do `python -m composer.cli` instead of `python -m composer.cli.launcher` (Ideally, they could just do `python -m composer`, but wil do that later as a bigger refactor for the launcher.)

composer/core/algorithm.py
```diff
@@ -92,14 +89,15 @@ def match(self, event: Event, state: State) -> bool:
...
+
         Returns:
             bool: True if this algorithm should run now.
         """
         raise NotImplementedError(f'implement match() required for {self.__class__.__name__}')
 
     @abstractmethod
     def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
-        """Applies the algorithm to make an in-place change to the :class:`~.state.State`.
+        """Apply the algorithm to make an in-place change to the :class:`~.state.State`.
```
A-Jacobson, line 100: I'm seeing a lot of `Applies` -> `Apply`  style changes here. is it necessary to change all of these? I don't think either is necessarily wrong.

pytorch uses the more narrative `Applies` style
https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/batchnorm.py#L231

numpy and pandas are more declarative.

composer/trainer/trainer.py
```diff
@@ -920,24 +916,25 @@ def __init__(
...
-        # Found latest checkpoint path, load that instead
-        if latest_checkpoint_path:
-            load_path = latest_checkpoint_path
-            # Disable object_store and load_weights_only since we're autoresuming locally
-            load_object_store = None
-            load_weights_only = False
-            log.info(
-                f"Found latest checkpoint at {latest_checkpoint_path}, loading instead of load_path {load_path} as autoresume enabled."
-            )
+        if autoresume:
```
A-Jacobson, line 919: related to autoresume not docs

docs/source/conf.py
```diff
@@ -283,7 +282,7 @@ def add_module_summary_tables(
     options: sphinx.ext.autodoc.Options,
     lines: List[str],
 ):
-    """This hook adds in summary tables for each module, documenting all functions, exceptions, classes, and attributes.
+    """Add in summary tables for each module, documenting all functions, exceptions, classes, and attributes.
```
A-Jacobson, line 285:
```suggestion
    """Add summary tables for each module, documenting all functions, exceptions, classes, and attributes.
```

docs/source/doctest_fixtures.py
```diff
@@ -134,6 +133,7 @@
 
 
 def loss_fun(output, target, reduction="none"):
+    """Dummy los function."""
```
A-Jacobson, line 136:
```suggestion
    """Dummy loss function."""
```

composer/callbacks/speed_monitor.py
ravi-mosaicml: Base class implementations of abstract methods generally do not need docstrings, as the abstract class should document the expected functionality.

A-Jacobson (approved):
there are a few changes in here that are unrelated to docs. I've flagged them so you can confirm you intend them to be in here.

Also fixed a few typos and made some grammar suggestions. Most are a matter of opinion, you don't have to accept them.

Other than that LGTM, thanks for taking the time to clean this stuff up!

composer/core/algorithm.py
ravi-mosaicml: Hmm, good catch -- the google style guide prefers descriptive over imperative, so will adjust the rule for that. Thanks!

https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

ravi-mosaicml: OK, it's going to be too much of a pain to convert all of them -- I did a find and replace for "apply" and "determine", but let's incrementally convert the rest as we refactor the codebase. I disabled the "imperative" setting in the config, so it won't warn on that anymore.

composer/trainer/trainer.py
ravi-mosaicml: You're right, I shouldn't have done this in this PR.

I made this change as we've been going with the convention that the caller should do the `if` check (i.e. `if x: do_y()`) rather than having the callee handle potentially-falsy values (i.e. `def y(x): return if not x else ...`), and figured it was easier to refactor it and then write a simpler docstring.

composer/loggers/object_store_logger.py
```diff
@@ -301,8 +297,8 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
 
     def log_symlink_artifact(self, state: State, log_level: LogLevel, existing_artifact_name: str,
                              symlink_artifact_name: str, overwrite: bool):
-        """Object stores do not natively support symlinks, so we emulate symlinks by adding a .symlink file to the
-        object store, which is a text file containing the name of the object it is pointing to."""
+        # Object stores do not natively support symlinks, so we emulate symlinks by adding a .symlink file to the
+        # object store, which is a text file containing the name of the object it is pointing to.
         # Only symlink if we're logging artifact to begin with
```
Averylamp, line 302: Is this meant to be commented out?

composer/trainer/devices/device.py
```diff
@@ -56,7 +56,9 @@ def tensor_to_device(self, tensor: torch.Tensor) -> torch.Tensor:
         pass
 
     def batch_to_device(self, batch: T_Batch) -> T_Batch:
-        """Invoked by the :class:`.Trainer` move all tensors items in a batch to device. Supports nested sequences and
+        """Invoked by the :class:`.Trainer` move all tensors items in a batch to device.
+
+        Supports nested sequences and
```
Averylamp, line 61: A little weird of a line break here

composer/utils/dist.py
```diff
@@ -392,7 +394,9 @@ def initialize_dist(backend: str, timeout: datetime.timedelta):
 
 
 def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool, shuffle: bool):
-    """Constructs a :class:`~torch.utils.data.distributed.DistributedSampler` for a dataset. The
+    """Constructs a :class:`~torch.utils.data.distributed.DistributedSampler` for a dataset.
+
+    The
```
Averylamp, line 399: nit: Line break here can probably be fixed as well

Averylamp (approved):
lgtm, small nits

composer/loggers/object_store_logger.py
ravi-mosaicml: Yes, since it's an implementation of an abstract method, it should not have a docstring.

PR #1093, opened 2022-05-26
Fix Composer Functional Notebook with CUDA

### LINKED ISSUE #1092: RuntimeError- Expected all tensors to be on the same device, but found at least two devices
I'm using **colab**.
I just tried to execute [Composer_Functional.ipynb](https://github.com/mosaicml/composer/blob/dev/notebooks/Composer_Functional.ipynb) cell by cell. 
After executing the following lines 

```
model = Net()
train_and_eval(model, trainloader, testloader)
```

I got the following error:

`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument target in method wrapper_nll_loss_forward)`

Here is a snippet of that:

Screenshot 2022-05-26 130854

How can I resolve this issue?

ravi-mosaicml:
The composer functional notebook example was missing calls for `y = y.to(device)`, causing errors when running on a colab runtime with GPU.

We do not have any CI coverage for notebooks, because of how we would need to monkey-patch the trainer to not do full training runs, but I did test that this fix works manually in colab.

Closes #1092


eracah (approved):
Works locally and on a colab with a GPU!

A-Jacobson (approved):
Lgtm

PR #1095, opened 2022-05-26
Write imagenet in ffcv format using streaming

dskhudia:
Add support for writing imagenet dataset in ffcv format using streaming dataset. 

Test:

Imagenet
```
python scripts/ffcv/create_ffcv_datasets.py
python scripts/ffcv/create_ffcv_datasets.py --split val
```

### DIFF
--- a/composer/datasets/ffcv_utils.py
+++ b/composer/datasets/ffcv_utils.py
@@ -43,7 +43,6 @@ def ffcv_monkey_patches():
 
 
 def write_ffcv_dataset(dataset: Optional[Dataset] = None,
-                       remote: Optional[str] = None,
                        write_path: str = "/tmp/dataset.ffcv",
                        max_resolution: Optional[int] = None,
                        num_workers: int = 16,
@@ -51,11 +50,10 @@ def write_ffcv_dataset(dataset: Optional[Dataset] = None,
                        compress_probability: float = 0.50,
                        jpeg_quality: float = 90,
                        chunk_size: int = 100):
-    """Converts PyTorch ``dataset`` or streaming dataset at ``remote`` into FFCV format at filepath ``write_path``.
+    """Converts PyTorch compatible ``dataset`` into FFCV format at filepath ``write_path``.
 
     Args:
         dataset (Iterable[Sample]): A PyTorch dataset. Default: ``None``.
-        remote (str): A remote path for a streaming dataset. Default: ``None``.
         write_path (str): Write results to this file. Default: ``"/tmp/dataset.ffcv"``.
         max_resolution (int): Limit resolution if provided. Default: ``None``.
         num_workers (int): Numbers of workers to use. Default: ``16``.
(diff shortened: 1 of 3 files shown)


composer/datasets/ffcv_utils.py
```diff
@@ -80,7 +78,4 @@ def write_ffcv_dataset(dataset: Optional[Dataset] = None,
             ffcv.fields.IntField()
     },
                                        num_workers=num_workers)
-    if dataset:
-        writer.from_indexed_dataset(dataset, chunksize=chunk_size)
-    elif remote is not None:
-        raise NotImplementedError("FFCV with remote datasets is not implemented")
+    writer.from_indexed_dataset(dataset, chunksize=chunk_size)
```
abhi-mosaic, line 81: So.. `StreamingDataset` technically has a `__getitem__` function but it's not actually download-safe currently... it's sort of called internally only on "safe" downloaded indices. I see a couple options

* fix this @knighton 
* pre-download the `StreamingDataset`, I think we could spin a multi-worker DataLoader and it should only take a few min, and then you can run `writer.from_indexed_dataset` as-is. @dskhudia
* implement a `.from_iterable_datset` function to the FFCV writer.. but this seems like a pain

dskhudia: Implemented #2 here.

scripts/ffcv/create_ffcv_datasets.py
```diff
@@ -117,10 +149,12 @@ def main():
         if args.subset > 0:
             ds = Subset(ds, range(args.subset))
     else:
-        remote_location = os.path.join(args.remote, args.split)
+        remote = os.path.join(args.remote, args.split)
+        local = os.path.join(args.local, args.split)
+        cacheStreaming(remote=remote, local=local, num_workers=args.num_workers)
```
knighton, line 154: cache_streaming?

scripts/ffcv/create_ffcv_datasets.py
```diff
@@ -100,12 +111,33 @@ def parse_args():
...
+
+    dl = torch.utils.data.DataLoader(ds,
+                                     batch_size=512,
+                                     num_workers=num_workers,
+                                     pin_memory=False,
+                                     drop_last=False,
+                                     collate_fn=pil_image_collate)
+
+    # empty loop to download and cache the dataset
+    for _ in dl:
```
knighton, line 132: I think this is begging for a blocking StreamingDataset.download() method (future PR)?

knighton (approved):
Side note, do we want to reinclude StreamingCIFAR10? Would that be helpful for create_ffcv_datasets?

scripts/ffcv/create_ffcv_datasets.py
dskhudia: Yes. This helper function would be useful and it can take arguments like how many workers to use for download.

hanlint:
@dskhudia do we still plan on merging this?

dskhudia:
@hanlint: Yes. Clusters were unavailable last week so couldn't do benchmarking with the new way @knighton added to download dataset.

dskhudia:
> Side note, do we want to reinclude StreamingCIFAR10? Would that be helpful for create_ffcv_datasets?

Yes. This will be useful for quick tests.

abhi-mosaic:
If this PR lands before #1128 then we will add a minor refactor of `StreamingImageClassDataset` -> `StreamingDataset` in that PR in accordance with this comment (https://github.com/mosaicml/composer/pull/1128#issuecomment-1148028582)

dskhudia:
@abhi-mosaic : Landed this for now. It should be simple change from StreamingImageClassDataset => StreamingDataset in this file.

PR #1096, opened 2022-05-26
Add StreamingC4 Dataset

abhi-mosaic:
This PR adds:
* `scripts/mds/c4.py`, a utility for converting the publicly available C4 dataset to `StreamingDataset` format
* a dataset object `StreamingC4` that currently supports sample-preserving processing like for BERT MLM pretraining
  * future TODO: build support for sample-mixing ops like `group_method: 'concat'` that are common for language modeling
* an hparams object `StreamingC4Hparams`

TODO:
* sync with @moinnadeem to add an example `bert_base_streaming_c4.yaml` 
* currently, the generation script downloads C4 locally to get faster read speed, but it still take ~5hr to download. The actual conversion writing process only takes 1.5hr afterwards. Future work should look into remote multi-threaded streaming from remote data locations.


scripts/mds/c4.py
```diff
@@ -0,0 +1,76 @@
...
+def main(args: Namespace) -> None:
+    """Main: create C4 streaming dataset.
+
+    Args:
+        args (Namespace): Commandline arguments.
+    """
+    fields = ["text", "timestamp", "url"]
+
+    for (split, split_new_name, expected_num_samples) in [
+        ("train", "train", 364868892),
```
knighton, line 62: Can't call __len__ on object returned by datasets.load_dataset? Dont trust it?

tests/datasets/test_streaming_remote.py
```diff
@@ -121,6 +138,14 @@ def test_streaming_remote_dataloader(tmp_path: pathlib.Path, name: str, split: s
     ds_build_dur = ds_build_end - ds_build_start
     print("Built dataset")
 
+    # Get collate_fn if needed
+    collate_fn = None
+    if name in ["ade20k", "imagenet1k"]:
+        collate_fn = pil_image_collate
+    elif name in ["c4"]:
+        collate_fn = DataCollatorForLanguageModeling(tokenizer=dataset.tokenizer,
+                                                                  mlm=True,
```
knighton, line 147: indent

scripts/mds/c4.py
```diff
@@ -0,0 +1,76 @@
...
+from datasets import IterableDataset
+
+from composer.datasets.streaming import StreamingDatasetWriter
+
+
+def parse_args() -> Namespace:
+    """Parse commandline arguments."""
+    args = ArgumentParser()
+    args.add_argument('--out_root', type=str, required=True)
+    args.add_argument('--shard_size_limit', type=int, default=1 << 27)
```
knighton, line 21: How many shards does this work out to, out of curiosity?

scripts/mds/c4.py
abhi-mosaic: The HF dataset in iterable mode does not have a len, sadly... larger reason is there's no guaranteed format for HF datasets, they're just objects that download a loading script from the web and then run it.

abhi-mosaic:
For posterity, I think this PR serves as a good example of what it takes to add a new streaming dataset to Composer (circa June 2022). There are still a minimum of 7 files to change...  reducing some steps here could improve ease-of-contribution.

1) a writer script `scripts/mds/c4.py` to demonstrate converting native data to `mds` format
2) a dataset reader object `StreamingC4(StreamingDataset)`
3) an hparams object that constructs a dataloader for training `StreamingC4Hparams`
4 and 5) some boilerplate registration under `datasets/__init__.py` and `datasets/dataset_hparams_registry.py` 
6 and 7) an E2E remote streaming test, and a boilerplate hparams registry test

abhi-mosaic:
Some technical details:

C4 train split has 364,868,892 samples, and takes up 777GB in `mds` format.

Iterating over `StreamingC4(..., tokenizer="bert-base-uncased", max_seq_len=512)` from remote S3 bucket with a single dataloader with 8 CPU workers achieves ~4k seqs/sec. This does not change even after cacheing, which implies the throughput is bottlenecked on CPU tokenization, which is happening on the fly and _independently per sample_, so it is embarassingly slow compared to batch processing.

In practice, I believe this is still OK because for NLP training, even our small 100M param models consume data at  ~400 seqs/sec/A100, so we have ~10x headroom before data throughput becomes an issue. The benefit of this approach is that we have high user flexibility, and can do any transforms we want on a per-sample basis.

knighton (approved):
Awesome, have some minor pedantic comments

scripts/mds/c4.py
knighton: Update: testing on 4398 imagenet train shards, I got about a 6% slowdown in shard open/close time by breaking shards into directories of 100 or 1000 and about a 6% speedup by moving the entire flat directory to root level, which I guess is what we deserve for having shards stored as separate files

composer/datasets/c4.py
```diff
@@ -1,29 +1,111 @@
...
+        else:
+            truncation = False
+            padding = False
+            max_length = None
+        return self.tokenizer(text_sample["text"], truncation=truncation, padding=padding, max_length=max_length)
+
+    def __init__(self,
+                 remote: str,
+                 local: str,
+                 split: str,
```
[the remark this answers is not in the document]
abhi-mosaic: Yeah so I want to avoid any notion of filepath structure in the base class, but then `StreamingC4` is meant to be a user-defined class that has user-specific flags, like split. I can also imagine someone building a dataset object that has no splits, or has `partition: int` or something like that, so I'd like to leave that flexibility to them.

composer/datasets/c4.py
```diff
@@ -1,29 +1,111 @@
...
+        max_seq_len (int): The max sequence length of each token sample.
+        group_method (str): How to group text samples into token samples. Currently only supporting ``'truncate'``.
+        batch_size (Optional[int]): Hint batch_size that will be used on each device's DataLoader. Default: ``None``.
+    """
+
+    def _decode(self, data: bytes) -> str:
+        return data.decode('utf-8')
+
+    def _tokenize(self, text_sample):
+        if self.group_method == "truncate":
```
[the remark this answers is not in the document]
abhi-mosaic: gonna punt this for now

composer/datasets/c4.py
```diff
@@ -1,29 +1,111 @@
...
+                 batch_size: Optional[int] = None):
+
+        # HF Transformers is needed to build the tokenizer
+        try:
+            import transformers
+        except ImportError as e:
+            raise MissingConditionalImportError(extra_deps_group="nlp", conda_package="transformers") from e
+
+        # Validation
+        if split not in ['train', 'val']:
```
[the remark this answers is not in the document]
abhi-mosaic: gonna keep this just for consistency with the other datasets.. but I agree with you it probably makes more sense as a set.

PR #1097, opened 2022-05-26
Revert back to DeepSpeed FP16 defaults to prevent underflow

moinnadeem:
Our custom defaults are causing convergence problems with DS+FP16 -- removing these yields proper convergence. I spoke to @abhi-mosaic about it and we should be able to remove it entirely.

### DIFF
--- a/composer/trainer/_deepspeed.py
+++ b/composer/trainer/_deepspeed.py
@@ -97,10 +97,6 @@ def _add_precision_config(config: Dict[str, Any], state: State):
         fp16_config = config["fp16"]
         assert isinstance(fp16_config, dict)
 
-        # For equivalence with the non-DeepSpeed defaults of the Mosaic trainer.
-        fp16_config.setdefault("initial_scale_power", 16)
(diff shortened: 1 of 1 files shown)


moinnadeem:
Another comment: we might need to think about the external users on our cluster who are saving checkpoint using the old loss values. Does changing this default affect them?

jbloxham:
> Another comment: we might need to think about the external users on our cluster who are saving checkpoint using the old loss values. Does changing this default affect them?

Good thinking, but I'd argue that compatibility of checkpoints across Composer versions isn't something we need to worry about until we hit 1.0.

jbloxham (approved):

abhi-mosaic:
I think checkpoints don't currently cache the DeepSpeed loss scaler right? I guess that's its own issue...

moinnadeem:
Wait, do we not cache the entire DeepSpeed config in order to reinitialize it from that config later on? I haven't dug into the DS code too much.

PR #1098, opened 2022-05-26
Prototype streaming with distributed download.

knighton:


ravi-mosaicml:
This PR was to crib from in building a gloo-backended PR for streaming v2. Not going this route.

PR #1099, opened 2022-05-26
Make Gradient Clipping an Algorithm

eracah:
CO-188

Moves all gradient clipping functionality to an algorithm and merges with AGC.

* changes `algorithms/agc/*` to `algorithms/gradient_clipping/*`
* `tests/algorithms/test_agc` to `tests/algorithms/test_gradient_clipping`
* removes grad_clipping logic from `Trainer`


composer/algorithms/gradient_clipping/README.md
```diff
@@ -0,0 +1,128 @@
+# 📎 Gradient Clipping
+
+[\[How to Use\]](#how-to-use) - [\[Suggested Hyperparameters\]](#suggested-hyperparameters) - [\[Technical Details\]](#technical-details) - [\[Attribution\]](#attribution)
+
+ `Computer Vision`, `Natural Language Processing`
+
+Gradient Clipping
```
hanlint, line 7: Need some intro text here (or somehwere) on what gradient clipping is, and the supported clipping types. And perhaps something about how to extend to your own clipping type?

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,230 @@
...
+            else:
+                raise NotImplementedError(f"Deepspeed only supports gradient clipping of type 'norm' not of type '{self.clipping_type}'")
+
+        # If Deepspeed is not enabled apply gradient clipping as normal.
+        apply_gradient_clipping(parameters=state.model.parameters(),
+                                clipping_type=self.clipping_type,
+                                clipping_threshold=self.clipping_threshold)
+
+
+# Factored this to a function to enable easier testing.
```
hanlint, line 174:
```suggestion
```

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,230 @@
...
+        # If Deepspeed is enabled and clipping_type is norm, set the clipping through
+        # deepspeed.
+        if state.is_model_deepspeed:
+            if self.clipping_type == 'norm':
+                state.model._config.gradient_clipping = self.clipping_threshold
+                return
+            else:
+                raise NotImplementedError(f"Deepspeed only supports gradient clipping of type 'norm' not of type '{self.clipping_type}'")
+
+        # If Deepspeed is not enabled apply gradient clipping as normal.
```
hanlint, line 168:
```suggestion
```

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,230 @@
...
+        except ImportError:
+            self.deep_speed_imported = False
+
+    def match(self, event: Event, state: State) -> bool:
+        """Run on ``Event.AFTER_TRAIN_BATCH``."""
+        return event == Event.AFTER_TRAIN_BATCH
+
+    def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
+        # If Deepspeed is enabled and clipping_type is norm, set the clipping through
+        # deepspeed.
```
hanlint, line 160:
```suggestion
        # deepspeed only supports norm clipping
```

composer/trainer/trainer.py
```diff
@@ -868,6 +868,12 @@ def __init__(
         self.state.evaluators = evaluators
 
         # Grad Clip Norm
+        if grad_clip_norm > 0:
+            warnings.warn(
+                DeprecationWarning(
+                    (f"Using the 'grad_clip_norm' field in Trainer is deprecated. PLease use"
+                     "the GradientClipping Algorithm in composer/algorithms/gradient_clipping")))
```
hanlint, line 875:
```suggestion
                    (f"Using the 'grad_clip_norm' field in Trainer is deprecated. Please use"
                     "the GradientClipping Algorithm in composer.algorithms.gradient_clipping.")))
```

hanlint: Similiar to how we handled ssr, should we do something here where we add the gradient clipping algorithm if grad_clip_norm is not None?

composer/algorithms/gradient_clipping/README.md
```diff
@@ -0,0 +1,128 @@
...
+<!--pytest-codeblocks:skip-->
+```python
+# Instantiate the algorithm and pass it into the Trainer
+# The trainer will automatically run it at the appropriate points in the training loop
+
+from composer.algorithms import GradientClipping
+from composer.trainer import Trainer
+
+clipping_type = 'norm' # can also be 'adaptive' or 'value'
+gc = GradientClipping(clipping_type=clipping_type, clipping_threshold = 0.1)
```
ravi-mosaicml, line 53:
```suggestion
gc = GradientClipping(clipping_type=clipping_type, clipping_threshold=0.1)
```

composer/algorithms/gradient_clipping/README.md
```diff
@@ -0,0 +1,128 @@
...
+    max_duration='1ep',
+    algorithms=[gc]
+)
+
+trainer.fit()
+```
+
+### Implementation Details
+
+#### clipping_type='norm'
```
ravi-mosaicml, line 67:
```suggestion
#### `clipping_type='norm'`
```

composer/algorithms/gradient_clipping/README.md
```diff
@@ -0,0 +1,128 @@
...
+Norm-based gradient clipping is implemented as follows:
+On `Event.AFTER_TRAIN_BATCH`, for every parameter in the model that has gradients:
+1. Compute the parameter's gradients and concatenate all parameters' gradients into one big vector
+2. Compute the norm of all the gradients (single scalar), `total_norm`
+3. Compute the clipping coefficient: `clip_coeff` = `clipping_threshold` / `total_norm`
+4. Clamp the `clip_coeff` to be less than or equal to 1.0
+5. Multiply all the gradients by the `clip_coeff`.
+
+#### clipping_type='value'
```
ravi-mosaicml, line 76:
```suggestion
#### `clipping_type='value'`
```

composer/yamls/algorithms/gradient_clipping.yaml
```diff
@@ -0,0 +1,2 @@
+clipping_type: "norm"
+clipping_threshold: 0.1
```
ravi-mosaicml, line 2: newline
```suggestion
clipping_threshold: 0.1

```

tests/trainer/test_trainer.py
```diff
@@ -432,24 +431,26 @@ def test_precision(
         if not should_error:
             assert_state_equivalent(init_trainer.state, fit_trainer.state)
 
-    @pytest.mark.parametrize("grad_clip_norm", [-1.0, 1.0])
+    @pytest.mark.parametrize("grad_clip_norm,context_manager", [(-1.0, contextlib.nullcontext),
+                                                                (1.0, pytest.deprecated_call)])
```
ravi-mosaicml, line 435: Why do we need the deprecated call marker?

composer/trainer/trainer.py
```diff
@@ -868,6 +868,12 @@ def __init__(
...
+        
         self._grad_clip_norm = grad_clip_norm
```
ravi-mosaicml, line 877: Since it's an alg now, can we remove `self._grad_clip_norm` from the trainer?

composer/trainer/trainer.py
```diff
@@ -1278,6 +1284,10 @@ def fit(
 
         # Grad Clip Norm
         if grad_clip_norm is not None:
+            warnings.warn(
+                DeprecationWarning(
+                    (f"Using the 'grad_clip_norm' field in fit() is deprecated. PLease use"
+                     "the GradientClipping Algorithm in composer/algorithms/gradient_clipping")))
             if self.deepspeed_enabled:
                 raise ValueError("Changing the grad_clip_norm when using DeepSpeed is not supported.")
             self._grad_clip_norm = grad_clip_norm
```
ravi-mosaicml, line 1293: Same thing here

composer/trainer/trainer.py
ravi-mosaicml: Then, would it be possible to rip out the following code from the trainer?

https://github.com/mosaicml/composer/blob/baeb171e796737aaea8dcb324113e7963a07f511/composer/trainer/trainer.py#L1717-L1721

That way, when we remove the deprecation, it'll just be in init and wherever we raise the warning -- would prefer not to have two implementations of grad clipping living around.

ravi-mosaicml (commented):
Thanks for doing this cleanup!

Some feedback:

1. Can we fully remove the grad clipping implementation in the trainer? See comments
2. Is it possible to refactor the grad clipping with deepspeed to not use private variables? See comment for a suggested alternative.
3. Can you take another pass through the docs and docstrings? The formatting on https://mosaicml-composer--1099.com.readthedocs.build/en/1099/api_reference/composer.algorithms.gradient_clipping.gradient_clipping.html#module-composer.algorithms.gradient_clipping.gradient_clipping and https://mosaicml-composer--1099.com.readthedocs.build/en/1099/method_cards/gradient_clipping.html looks a bit off in places.

I didn't review the grad clipping implementations; going to defer on that to @hanlint  and #research.

docs/source/index.rst
```diff
@@ -96,7 +96,7 @@ Composer is part of the broader Machine Learning community, and we welcome any c
    :caption: Methods Library
 
    method_cards/methods_overview.rst
-   method_cards/agc.md
+   method_cards/gradient_clipping.md
```
ravi-mosaicml, line 99: Can you move it down to the correct position in alphabetical order?

composer/algorithms/gradient_clipping/README.md
eracah: added some more text here: 7852f42

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,230 @@
...
+        if state.is_model_deepspeed:
+            if self.clipping_type == 'norm':
+                state.model._config.gradient_clipping = self.clipping_threshold
```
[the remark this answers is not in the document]
jbloxham: I don't have the final say here, but I'm actually somewhat in favor of leaving it as currently written. Having special code for an algorithm in the trainer feels really wrong, and I don't want there to be a precedent for that.

On the other hand, DeepSpeed is such a mess that I doubt this will be the last time we access its private variables in this way (getting it to work with adaptive grad accum is a place where we might do something similar).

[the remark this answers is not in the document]
hanlint: fwiw, agreed with @jbloxham 's take here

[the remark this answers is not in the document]
eracah: yeah seems like a lot just to avoid a private var

composer/trainer/trainer.py
```diff
@@ -1186,7 +1192,7 @@ def fit(
             eval_interval (int | str | Time | (State, Event) -> bool, optional): See :class:`.Trainer`.
             grad_accum (int | str, optional): See :class:`.Trainer`.
             precision (Precision | str, optional): See :class:`.Trainer`.
-            grad_clip_norm (float, optional): See :class:`.Trainer`.
+            grad_clip_norm (float, optional): See :class:`.Trainer`. Deprecated. Please use composer.algorithms.GradientClipping
```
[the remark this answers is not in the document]
eracah: done here [81841f9](https://github.com/mosaicml/composer/pull/1099/commits/81841f91bc9bf7a57469769cc566cf7b08815b78)

docs/source/index.rst
eracah: done here 
47c25c1

composer/trainer/trainer.py
eracah: ok done. check out [0e50ebb](https://github.com/mosaicml/composer/pull/1099/commits/0e50ebba2fe60a20cc09f4787bc98a472d0b7742)

composer/trainer/trainer.py
```diff
@@ -1278,6 +1284,10 @@ def fit(
 
         # Grad Clip Norm
         if grad_clip_norm is not None:
```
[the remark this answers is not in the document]
eracah: removed from fit

composer/trainer/trainer.py
eracah: removed everywhere except for used once in `__init__`

composer/algorithms/gradient_clipping/gradient_clipping.py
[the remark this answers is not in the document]
eracah: going to keep it as is

eracah:
> 1. Can we fully remove the grad clipping implementation in the trainer? See comments
I removed it everywhere except for in `__init__`, I just append the GradientClipping algorithm to algorithms if it is specified.
> 2. Is it possible to refactor the grad clipping with deepspeed to not use private variables? See comment for a suggested alternative.
We decided to keep as is since deepspeed is such a mess/pain to work with anyway.
> 3. Can you take another pass through the docs and docstrings? The formatting on https://mosaicml-composer--1099.com.readthedocs.build/en/1099/api_reference/composer.algorithms.gradient_clipping.gradient_clipping.html#module-composer.algorithms.gradient_clipping.gradient_clipping and https://mosaicml-composer--1099.com.readthedocs.build/en/1099/method_cards/gradient_clipping.html looks a bit off in places.
Will take a look at this tmrw

hanlint (approved):
LGTM

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,227 @@
...
+                algorithms=[gc],
+                optimizers=[optimizer]
+            )
+    Args:
+        parameters (torch.Tensor or Iterable[torch.Tensor]): The parameters to of the
+            model for whose gradients we will clip
+        clipping_type ('adaptive', 'norm', 'value'): String denoting which type of
+            gradient clipping to do. The options are:
+
+            =============   ===========                             ========
```
ravi-mosaicml, line 119: Same thing here -- https://mosaicml-composer--1099.com.readthedocs.build/en/1099/api_reference/composer.algorithms.gradient_clipping.gradient_clipping.html#composer.algorithms.gradient_clipping.gradient_clipping.GradientClipping

composer/trainer/trainer.py
```diff
@@ -1186,7 +1197,7 @@ def fit(
             eval_interval (int | str | Time | (State, Event) -> bool, optional): See :class:`.Trainer`.
             grad_accum (int | str, optional): See :class:`.Trainer`.
             precision (Precision | str, optional): See :class:`.Trainer`.
-            grad_clip_norm (float, optional): See :class:`.Trainer`.
+            grad_clip_norm (float, optional): See :class:`.Trainer`. Deprecated. Please use composer.algorithms.GradientClipping
```
ravi-mosaicml, line 1200: Can you remove this docstring line? The arg was removed in the singature above.

composer/algorithms/gradient_clipping/gradient_clipping.py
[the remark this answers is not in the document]
ravi-mosaicml: I'm really not a fan of this, as it could cause a breaking change without any errors when we upgrade the deepspeed version. In addition, how do we know this works? Deepspeed could be reading this variable during `deepspeed.initialize`, in which case this monekypatch would have no effect.

One alternative would be to put the `deepspeed_config` on the state during `Event.INIT`, so algorithms can modify it directly.

If the decision has already been made to modify this private variable, please add a test case to ensure that the gradients are indeed being clipped (not just a functional test, but a test to ensure that doing it this way actually clips the gradients).

ravi-mosaicml (commented):
It looks like the docs still need some work. I also disagree with how deepspeed is being handled. Otherwise looks fine to me.

> I'm really not a fan of this, as it could cause a breaking change without any errors when we upgrade the deepspeed version. In addition, how do we know this works? Deepspeed could be reading this variable during deepspeed.initialize, in which case this monekypatch would have no effect.
>
> One alternative would be to put the deepspeed_config on the state during Event.INIT, so algorithms can modify it directly.
>
> If the decision has already been made to modify this private variable, please add a test case to ensure that the gradients are indeed being clipped (not just a functional test, but a test to ensure that doing it this way actually clips the gradients).

composer/algorithms/gradient_clipping/gradient_clipping.py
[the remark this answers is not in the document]
eracah: > I'm really not a fan of this, as it could cause a breaking change without any errors when we upgrade the deepspeed version. 
I'm not really convinced by this argument since plenty of our existing code could break if we upgrade deepspeed anyway. Not sure how this would be any different.

>In addition, how do we know this works? Deepspeed could be reading this variable during `deepspeed.initialize`, in which case this monekypatch would have no effect.

Good call. It should actually modify the optimizer directly. Fixed here:  3b66190

> One alternative would be to put the `deepspeed_config` on the state during `Event.INIT`, so algorithms can modify it directly. 

That is a certainly a valid solution, but would be more work and seemingly not necessary.

> If the decision has already been made to modify this private variable, please add a test case to ensure that the gradients are indeed being clipped (not just a functional test, but a test to ensure that doing it this way actually clips the gradients).

I think just verifying that the optimizer has optimizer.clip_grad == clipping_threshold is enough since we can trust that deepspeed works as written and that their tests are good.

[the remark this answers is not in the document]
eracah: ok that test is here: 
fdbb567

composer/trainer/trainer.py
eracah: done here: [715c203](https://github.com/mosaicml/composer/pull/1099/commits/715c20344830a1f5a31b8a2962186ac7103971ab)

composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -0,0 +1,227 @@
...
+
+        apply_gradient_clipping(parameters=state.model.parameters(),
+                                clipping_type=self.clipping_type,
+                                clipping_threshold=self.clipping_threshold)
+
+
+def _get_clipped_gradient_coeff(weights: torch.Tensor, grad: torch.Tensor, clipping_threshold: float = 0.01):
+    """Clips all gradients in model based on ratio of gradient norms to parameter norms.
+
+    Gradients whose norms exceed weight_norm * clipping_threshold are scaled down by
```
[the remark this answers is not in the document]
eracah: fix here: 
f56d4e5

tests/algorithms/test_gradient_clipping.py
```diff
@@ -58,56 +58,122 @@ def forward(self, x):
...
     state = Mock()
-    state.model = model
     state.profiler.marker = Mock(return_value=None)
     state.callbacks = []
-    state.algorithms = [AGC()]
+
+    # Enable deepspeed and set clipping_type to norm to ensure that apply_gradient_clipping
+    # is not called.
+    state.algorithms = [GradientClipping(clipping_type='norm', clipping_threshold=clipping_threshold)]
+    state.is_model_deepspeed = True
```
ravi-mosaicml, line 110: This is a read-only property so it cannot be set like this. Intead, the model will have to be wrapped with DeepSpeed engine. It would probably be easiest to construct an actual trainer here that uses deepspeed.

tests/algorithms/test_gradient_clipping.py
```diff
@@ -58,56 +58,122 @@ def forward(self, x):
...
 
-def test_AGC_algorithm(simple_model_with_grads):
-    # Get weight and gradients from the fixture model.
-    model = simple_model_with_grads
-    weights = next(model.parameters())
-    grad = weights.grad
-    expected_clipped_grad = grad.detach() * _get_clipped_gradient_coeff(weights, grad)
 
-    # Set up a mock engine.
+def test_gradient_clipping_algorithm_with_deepspeed_enabled(monkeypatch: pytest.MonkeyPatch):
```
ravi-mosaicml, line 99: This test will require `@pytest.mark.gpu` since deepspeed requires gpu.

tests/algorithms/test_gradient_clipping.py
```diff
@@ -58,56 +58,122 @@ def forward(self, x):
...
 
     # Run the Event that should cause AGC.apply to be called.
     engine.run_event(Event.AFTER_TRAIN_BATCH)
 
-    # Check that the gradients weights holds are equivalent to the calling
-    # grad.detach() * _get_clipped_gradient_coeff on the weights and grads.
+    # Make sure deepspeed engine's gradient_clipping_field is set properly.
+    assert state.optimizers.clip_grad == clipping_threshold
+    # Make sure apply_gradient_clipping is not called.
+    apply_gc_fn.assert_not_called()
```
ravi-mosaicml, line 123: I don't think this tests that gradients were actually clipped when using deepspeed. Would it be possible to inspect the optimizer weights?

tests/algorithms/test_gradient_clipping.py
eracah: fixed here: [cc8830d](https://github.com/mosaicml/composer/pull/1099/commits/cc8830dc422f9a472fc61fba673f07ba9e2d22b3). Using `state.deepspeed_enabled` which checks `state.deepspeed_config`

tests/algorithms/test_gradient_clipping.py
eracah: just ended up mocking most of it here: [cc8830d](https://github.com/mosaicml/composer/pull/1099/commits/cc8830dc422f9a472fc61fba673f07ba9e2d22b3)

tests/algorithms/test_gradient_clipping.py
eracah: now with setting deepspeed_config here: [cc8830d](https://github.com/mosaicml/composer/pull/1099/commits/cc8830dc422f9a472fc61fba673f07ba9e2d22b3), I don't think it's necessary to check the gradients are clipped if we trust the DeepSpeed API

composer/trainer/trainer.py
```diff
@@ -782,6 +786,33 @@ def __init__(
...
 
+        # Grad Clip Norm
+        if grad_clip_norm > 0:
+
+            warnings.warn(
+                DeprecationWarning((f"Using the 'grad_clip_norm' field in Trainer is deprecated. Please use"
+                                    'the GradientClipping Algorithm in composer.algorithms.gradient_clipping.')))
+
+            print_warning = False
+            if algorithms is not None:
```
ravi-mosaicml, line 797: nit: for brevity, can you do:

```if any(isinstance(alg, GradientClipping) for alg in ensure_tuple(algorithms)): warnings.warn(...)```

File: composer/trainer/trainer.py [hunk 24]
```diff
@@ -782,6 +786,33 @@ def __init__(
...
+                if isinstance(algorithms, Sequence):
+                    if any([isinstance(algo, GradientClipping) for algo in algorithms]):
+                        print_warning = True
+                elif isinstance(algorithms, GradientClipping):
+                    print_warning = True
+
+                if print_warning:
+                    warnings.warn(
+                        RuntimeWarning(
```
ravi-mosaicml, line 806: Can this be the default UserWarning rather than a RuntimeWarning?

composer/trainer/trainer.py
```diff
@@ -795,6 +826,8 @@ def __init__(
                            optimizers=optimizers,
                            run_name=run_name)
 
+        self.state.deepspeed_config = deepspeed_config
```
ravi-mosaicml, line 829: Can the deepspeed config be passed in when constructing the state?

composer/core/state.py
```diff
@@ -272,6 +272,7 @@ def __init__(
 
         # Set defaults for transient variables (to make pyright happy)
         self.batch: Any = None
+        self.deepspeed_config: Any = None
```
ravi-mosaicml, line 275: Can you add this argument (and a docstring entry) as part of the `__init__` for state, and type it as `Optional[Dict[str, Any]]`?

composer/core/state.py
```diff
@@ -412,6 +413,10 @@ def evaluators(self):
     def evaluators(self, evaluators: Union[Evaluator, Sequence[Evaluator]]):
         self._evaluators[:] = list(ensure_tuple(evaluators))
 
+    @property
+    def deepspeed_enabled(self):
```
ravi-mosaicml, line 417: Needs a docstring

ravi-mosaicml (commented):
Overall LGTM, see comments. Can you address in a follow-up PR?

PR #1102, opened 2022-05-27
Refactor algorithms from packages to modules

eracah:
CO-396

This task will change all algorithms from packages to modules and move the modules to be directly under the algorithms package.

mv composer/algorithms/{alg_name}/{alg_name}.py composer/algorithms/

mv composer/algorithms/{alg_name}/README.md docs/source/method_cards/{alg_name}.md

rm -rf composer/algorithms/{alg_name}/__init__.py

mv composer/algorithms/{alg_name}/metadata.json docs/source/metadata/{alg_name}.json

Doing this will remove two steps from this [list](https://www.notion.so/Checklist-for-implementing-an-algorithm-in-composer-2838052d244a44829d3c356665153c9f):

Remove the need for the user to write composer/algorithms/{alg_name}/__init__.py , which is unnecessary boilerplate

Remove the need for the user to create a symlink: ln -s composer/algorithms/{alg_name}/README.md docs/source/method_cards/{alg_name}.md

Removing 2 steps seems minimal, but if we want to get from 13 steps to 3 or 4, every step counts!


ravi-mosaicml (changes requested):
I understand and appreciate how this would help reduce the number of steps to write a new alg. However, for context, it was decided a while back to keep the `README.md` files next to the algorithm implementation to help if someone is browsing the GitHub (e.g. see https://github.com/mosaicml/composer/tree/dev/composer/algorithms/alibi)

Are there other ways we could simplify this process a bit? (Perhaps we could configure sphinx to search through composer/algorithms directly instead of requiring symlinks?) The algorithm implementation _could_ go in the `__init__.py` file within the algorithm folder, which would then save an additional file, while still keeping the folder structure.

CC @hanlint for thoughts.

hanlint:
yeah, agreed with @ravi-mosaicml here, I think we want to keep things readable (and the folder self-contained and portable)

eracah:
Ok. Makes sense. Davis and I have another approach, which keeps all the files together as is that we can discuss at the next design meeting.

PR #1103, opened 2022-05-27
Refactor StreamingDataset, cleaning up and adding public download().

knighton:


knighton:
The following took 85 sec the first run, then 0.04 sec on subsequent runs (local machine lol)

```
from time import time
from torchvision import transforms as tf
from typing import Any

from composer.datasets.streaming import StreamingImageClassDataset



class StreamingCIFAR(StreamingImageClassDataset):
    """Streaming CIFAR."""

    def decode_image(self, data: bytes) -> Any:
        arr = np.frombuffer(data, np.uint8)
        arr = arr.reshape(32, 32, 3)
        return Image.fromarray(arr)


remote = 's3://mosaicml-internal-dataset-cifar10/mds/1/train/'
local = '/tmp/mds-cache/mds-cifar10/train/'
shuffle = True
transform = tf.ToTensor()
dataset = StreamingCIFAR(remote, local, shuffle, transform)
print(len(dataset))
t0 = time()
dataset.download()
t = time() - t0
print('%.3f' % t)
```

dskhudia (approved):
Thanks for putting out so quickly.

dskhudia:
This test is failing for me as well on another PR and seems unrelated failure.

PR #1104, opened 2022-05-28
Move `run_name` into the state

mvpatel2000:
- supports storing and loading run_name in checkpoints
- makes providing run_name optional with autoresume if it isn't used by any of the provided names
- reuses run_name with autoresume, or otherwise generates new run_name. Adds tests to verify behavior

Note: I've removed the doctest for creating a runname because it doesn't actually test any functionality and it's now annoying since it would require an entire trainer init -- lmk if we want to keep it with trainer init


composer/trainer/trainer.py
```diff
@@ -1014,19 +1019,29 @@ def _check_for_autoresume(self, autoresume: bool, save_folder: Optional[str], sa
...
                     "save_latest_filename must be specified so autoresume knows where to load checkpoints from.")
-            if run_name is None:
-                raise ValueError("run_name must be specified so autoresume knows which run to load from.")
-            latest_filename_formatted = format_name_with_dist(save_latest_filename, run_name)
-            latest_checkpoint_path = os.path.join(save_folder, latest_filename_formatted)
+            latest_filename_formatted = save_latest_filename
+            save_folder_formatted = save_folder
+            latest_artifact_name = save_latest_artifact_name
+            # Check run_name is provided if necessary
+            if ("{run_name}" in save_folder or "{run_name}" in save_latest_filename or
```
ravi-mosaicml, line 1026: Instead of checking this directly, would it be possible to use https://docs.python.org/3/library/string.html#string.Formatter.parse instead?
Mainly concerned that this would not catch something like `{run_name:some_format_args}`, that would then cause this (manual) check to return an incorrect value.

composer/trainer/trainer.py
```diff
@@ -1014,19 +1019,29 @@ def _check_for_autoresume(self, autoresume: bool, save_folder: Optional[str], sa
...
+                    "{run_name}" in save_latest_artifact_name):
+                # Raise error if run_name is not provided
+                if run_name is None:
+                    raise ValueError("run_name must be specified so autoresume knows which run to load from.")
+                # Otherwise, format names
+                else:
```
ravi-mosaicml, line 1032: Can this else` be un-indented one level? The name should still be formatted if it contains other variables, like distributed variables.

composer/trainer/trainer.py
```diff
@@ -1014,19 +1019,29 @@ def _check_for_autoresume(self, autoresume: bool, save_folder: Optional[str], sa
...
+                    latest_filename_formatted = format_name_with_dist(save_latest_filename, run_name)
+                    save_folder_formatted = format_name_with_dist(save_folder, run_name)
+                    latest_artifact_name = format_name_with_dist(save_latest_artifact_name, run_name)
+            latest_checkpoint_path = os.path.join(save_folder_formatted, latest_filename_formatted)
             # If latest checkpoint is not saved locally, try to fetch from loggers
-            if not os.path.exists(latest_checkpoint_path) and save_latest_artifact_name is not None:
+            if not os.path.exists(latest_checkpoint_path):
```
ravi-mosaicml, line 1038: Curious why this change?

tests/trainer/test_checkpoint.py
```diff
@@ -476,6 +479,7 @@ def test_checkpoint_with_object_store_logger(
     pytest.param("resnet50_synthetic", marks=pytest.mark.daily),
     pytest.param("gpt2_52m", marks=pytest.mark.daily),
 ])
+@pytest.mark.parametrize("different_names", [pytest.param(True), pytest.param(False)])
```
ravi-mosaicml, line 482: Can this be its own test? Not a fan of expanding this parameterization matrix by 2x for this.

composer/trainer/trainer.py
mvpatel2000: its always non-None

mvpatel2000:
Working through flakey tests since timestamp changes for some of these checkpoint runs... but otherwise good for review

composer/loggers/logger.py
```diff
@@ -67,32 +65,7 @@ class Logger:
...
-            .. doctest:: composer.loggers.logger.Logger.__init__.run_name
-
-                >>> logger = Logger(state=state, destinations=[])
-                >>> logger.run_name
-                '1647293526-electric-zebra'
-
-            .. testcleanup:: composer.loggers.logger.Logger.__init__.run_name
-
-                time.time = original_time
+        run_name (str): The name for this training run.
```
ravi-mosaicml, line 68: Can we fully remove this field (and attribute) from the `Logger` class, since the `run_name` is now on the `state` attribute? I realize this will be a breaking change.

composer/trainer/trainer.py
```diff
@@ -388,9 +391,8 @@ class Trainer:
         loggers (LoggerDestination | Sequence[LoggerDestination], optional): The destinations to log training information to.
 
             .. seealso:: :mod:`composer.loggers` for the different loggers built into Composer.
-        run_name (str, optional): A name for this training run. If not specified, one will be generated automatically.
-
-            .. seealso:: :class:`~composer.loggers.logger.Logger` for a description of the run name.
+        run_name (str, optional): A name for this training run. If not specified, the timestamp will be combined with a
+            :doc:`coolname <coolname:index>`.
```
ravi-mosaicml, line 395: Makes sense to remove the doctest; but could you include a hard-coded example here of an auto-generated run name?

composer/trainer/trainer.py
```diff
@@ -1020,18 +1038,27 @@ def _determine_autoresume_load_path(
...
                 "save_latest_filename must be specified so autoresume knows where to load checkpoints from.")
+        # Check run_name is provided if necessary
         if run_name is None:
-            raise ValueError("run_name must be specified so autoresume knows which run to load from.")
-        latest_filename_formatted = format_name_with_dist(save_latest_filename, run_name)
-        latest_checkpoint_path = os.path.join(save_folder, latest_filename_formatted)
+            for save_arg in [save_folder, save_latest_filename, save_latest_artifact_name]:
+                for _, field_name, _, _ in string.Formatter().parse(save_arg):
+                    if field_name == "run_name":
+                        raise ValueError("run_name must be specified so autoresume knows which run to load from.")
```
ravi-mosaicml, line 1046:
```suggestion
                        raise ValueError("Since the `{save_arg}` contains the {{run_name}} variable, the `run_name` argument must be specified to determine the checkpoint path and load the checkpoint.")
```

composer/trainer/trainer.py
```diff
@@ -1020,18 +1038,27 @@ def _determine_autoresume_load_path(
         if save_latest_filename is None:
             raise ValueError(
                 "save_latest_filename must be specified so autoresume knows where to load checkpoints from.")
+        # Check run_name is provided if necessary
         if run_name is None:
-            raise ValueError("run_name must be specified so autoresume knows which run to load from.")
```
ravi-mosaicml, line 1024: (See slack), but can we continue to require this argument when using autoresume for the time being?

composer/trainer/trainer.py
```diff
@@ -944,6 +955,13 @@ def __init__(
                                               strict_model_weights=load_strict_model_weights,
                                               chunk_size=load_chunk_size,
                                               progress_bar=load_progress_bar)
+            # If using autoresume, use loaded run_name
+            if autoresume and run_name is not None:
```
ravi-mosaicml, line 959: For now, can we always override the run name to whatever was used for `Event.INIT`?

e.g. `self.state.run_name = generated_run_name`

ravi-mosaicml (commented):
Sorry for not catching this earlier, but I realize that to properly restore the run name from a checkpoint will require a bigger refactor of checkpoint loading. We need to restore the run name before `Event.INIT` (the only part of the checkpoint which should be restored then), so callbacks and loggers (e.g. wandb) that initialize based on the run name will have the correct run name then. (If the user does not supply a run name, the PR, in its current form, would call `Event.INIT` with the autogenerated run name, and change it to the run name in the saved checkpoint). This refactor to support this functionality can happen in a future PR.

 For now, can we limit this PR to moving the `run_name` into the `state` so it will be serialized, and out of the `logger`? Feel free to keep around the check for the `run_name` in the checkpoint paths.

composer/core/state.py
```diff
@@ -282,6 +287,7 @@ def __init__(
             "timestamp",
             "rank_zero_seed",
             "current_metrics",
+            "run_name",
```
ravi-mosaicml, line 290: Can you also add the `run_name` into the table of attributes that get serialized?

composer/profiler/torch_profiler.py
```diff
@@ -81,7 +81,7 @@ class TorchProfiler(Callback):  # noqa: D101
 
             Consider the following scenario, where:
 
-            *   The :attr:`~.Logger.run_name` is ``'awesome-training-run'``.
+            *   The :attr:`~.state.run_name` is ``'awesome-training-run'``.
```
ravi-mosaicml, line 84:
```suggestion
            *   The :attr:`~.State.run_name` is ``'awesome-training-run'``.
```

composer/trainer/trainer.py
```diff
@@ -750,16 +763,18 @@ def __init__(
         self.adaptive_gradient_accumulation = _is_adaptive_grad_accum(grad_accum, device=self._device)
         grad_accum = _get_initial_grad_accum(grad_accum)
 
+        # Run Name
+        generated_run_name = _get_run_name(run_name=run_name)
```
ravi-mosaicml, line 767: nit: since it could also be the user-supplied `run_name`, would prefer to keep the variable called `run_name`
```suggestion
        run_name = _get_run_name(run_name=run_name)
```

composer/trainer/trainer.py
```diff
@@ -750,16 +763,18 @@ def __init__(
...
-            optimizers=optimizers,
-        )
+        self.state = State(rank_zero_seed=rank_zero_seed,
+                           algorithms=algorithms,
+                           model=model,
+                           callbacks=callbacks,
+                           grad_accum=grad_accum,
+                           precision=precision,
+                           optimizers=optimizers,
+                           run_name=generated_run_name)
```
ravi-mosaicml, line 777:
```suggestion
                           run_name=run_name)
```

composer/trainer/trainer.py
```diff
@@ -944,6 +959,8 @@ def __init__(
                                               strict_model_weights=load_strict_model_weights,
                                               chunk_size=load_chunk_size,
                                               progress_bar=load_progress_bar)
+            # Always override run_name. In the future, we'll use the loaded name and not require run_name
+            self.state.run_name = generated_run_name
```
ravi-mosaicml, line 963:
```suggestion
            self.state.run_name = run_name
```

composer/trainer/trainer.py
```diff
@@ -944,6 +959,8 @@ def __init__(
                                               strict_model_weights=load_strict_model_weights,
                                               chunk_size=load_chunk_size,
                                               progress_bar=load_progress_bar)
+            # Always override run_name. In the future, we'll use the loaded name and not require run_name
```
ravi-mosaicml, line 962:
```suggestion
            # Always override run_name so it is consistent with what was used for Event.INIT. In the future, we'll use the loaded name and not require run_name
```

composer/trainer/trainer.py
ravi-mosaicml, line 963: If `autoresume and run_name != self.state.run_name`, can you raise an exception (for the time being), as the run name must be the same for autoresume to work properly?

tests/loggers/test_logger.py
```diff
@@ -3,24 +3,8 @@
...
 
-import pytest
-
 from composer.core.state import State
 from composer.loggers import Logger, LoggerDestination, LogLevel
-from composer.utils import dist, reproducibility
-
-
-@pytest.mark.world_size(2)
-def test_logger_run_name(dummy_state: State):
```
ravi-mosaicml, line 14: Can this test be moved into `test_trainer`?

ravi-mosaicml (approved):
LGTM! See comments

PR #1106, opened 2022-05-31
S3ObjectStore + streaming download change

anisehsani:


composer/utils/s3_object_store.py
```diff
@@ -0,0 +1,124 @@
...
+            s3_transfer_config: Optional[Dict[Any, Any]] = None) -> None:
+        self.bucket = bucket
+        config = Config(**s3_client_config)
+        self.client = boto3.client('s3', config=config)
+        self.s3_transfer_config = s3_transfer_config
+
+    def upload_object(self,
+                      file_path: str,
+                      object_name: str,
+                      bucket: Optional[str] = None,
```
ravi-mosaicml, line 30: Bucket is specified on `__init__` so I don't think it needs to be here?

composer/utils/s3_object_store.py
```diff
@@ -0,0 +1,124 @@
...
+            if self.bucket is None:
+                raise ValueError("No S3 bucket specified")
+            bucket = self.bucket
+
+        self.client.upload_file(Filename=file_path, Bucket=bucket, Key=object_name, ExtraArgs=extra)
+
+    def upload_object_via_stream(self,
+                                 obj: Union[bytes, Iterator[bytes]],
+                                 object_name: str,
+                                 bucket: Optional[str] = None,
```
ravi-mosaicml, line 57: Same thing

composer/utils/s3_object_store.py
```diff
@@ -0,0 +1,124 @@
...
+        del headers
+
+        if bucket is None:
+            if self.bucket is None:
+                raise ValueError("No S3 bucket specified")
+            bucket = self.bucket
+        
+        # add handling for other bytes-like types
+        if not isinstance(obj, bytes):
+            raise NotImplementedError("S3ObjectStore.upload_object_via_stream only takes a bytes object.")
```
ravi-mosaicml, line 80: Can we support `iterator[bytes]`? See https://stackoverflow.com/questions/62876879/how-to-create-a-python-file-like-object-from-an-iterator

composer/utils/s3_object_store.py
```diff
@@ -0,0 +1,124 @@
...
+                        delete_on_failure: bool = True):
+        """Download an object to the specified destination path.
+        """
+
+        if bucket is None:
+            if self.bucket is None:
+                raise ValueError("No S3 bucket specified")
+            bucket = self.bucket
+
+        self.client.download_file(Bucket=bucket, Key=object_name, Filename=destination_path)
```
ravi-mosaicml, line 103: Can `overwrite_existing` be implemented (e.g. error if it is false and the file already exists?)

In addition, can you download to a temporary file path (e.g. `destination_path + f".{time.time()}.tmp"`, and then once the file has been downloaded, `os.rename` it to `destination_path`)? If the download fails, then can we make a best ensure that the tempfile is automatically cleaned up (probably via try/finally?)

composer/utils/object_store.py
```diff
@@ -0,0 +1,46 @@
...
+                                 headers: Optional[Dict[str, str]] = None):
+        """Upload an object.
+        """
+        pass
+
+    def download_object(self,
+                        object_name: str,
+                        destination_path: str,
+                        overwrite_existing: bool = False,
+                        delete_on_failure: bool = True):
```
ravi-mosaicml, line 38: Can you remove the `delete_on_failure` argument? I realize this should never have been public in the libcloud class (the answer should always be yes).

composer/utils/s3_object_store.py
```diff
@@ -0,0 +1,124 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
```
ravi-mosaicml, line 2: Is it possible to add test cases for this? Take a look at https://github.com/spulec/moto for mocking out s3

ravi-mosaicml (commented):
Thanks for working on this. Would it be possible to refactor the base class changes into a seperate PR (so we can merge that in first)? In addition, please see comments on the s3 implementation.

abhi-mosaic:
Could someone help me understand with a quick code-snippet how a user would insantiate a dataset now? If pointing at a local streaming dataset, then no `object_store` would need to be provided, but otherwise we would need to import from Composer, instantiate an `object_store`, and pass?

I'm curious, in our `dataset_hparams` code blocks, how do we get the object store to pass to the dataset? Is there a shared object store for the Trainer, or do we have to include ObjectStoreHparams somehow?

ravi-mosaicml:
> Could someone help me understand with a quick code-snippet how a user would insantiate a dataset now? If pointing at a local streaming dataset, then no `object_store` would need to be provided, but otherwise we would need to import from Composer, instantiate an `object_store`, and pass?
> 
> I'm curious, in our `dataset_hparams` code blocks, how do we get the object store to pass to the dataset? Is there a shared object store for the Trainer, or do we have to include ObjectStoreHparams somehow?

Streaming datasets can use `get_file()`.

https://github.com/mosaicml/composer/blob/5fe3b59531a35f9a4ba6e09639e60c601840e508/composer/utils/file_helpers.py#L301.

I'm thinking that if `path` begins with a supported protocol (`https`, `http`, s3`, `gcs`, or `sftp`) and `object_store=None`, then it will delegate to the relevant object store implementation. Otherwise it will be treated as a local path.

If an `object_store` is specified (e.g. for custom authentication options, or some sort of advanced use case), then `path` will be treated as the key inside the bucket (for good ux, we should also support a full URI, and chop of the protocol and bucket).

There is no one object store for the trainer, so the object store may have to be passed around in multiple places. To avoid code duplication in the yaml, use yaml references for the object store hparams.

ravi-mosaicml:
Closing for #1147.

#1147 implements the s3 object store API, and the refactor of streaming to use the object store API should be done in a separate PR.

PR #1108, opened 2022-05-31
Updating `CONTRIBUTING.md` and `STYLE_GUIDE.md`

ravi-mosaicml:
- Add links to typing documentation in the style guide
- Added a note to document class `__init__` args under the class-level docstring
- Added in some notes about code structuring
- Fixed links (using relative links for a few items in the repo; switched the link in CONTRIBUTING.md to point to our style guide rather than the google one)
- Added a note about that callers, rather than calles, should check for argument values that result in no-ops


STYLE_GUIDE.md
```diff
@@ -36,12 +36,32 @@ pre-commit run isort --all-files  # for isort
 The configuration is stored in [pyproject.toml](pyproject.toml).
 
 
+### 2.4. Code Structure
+
+As a general rule of thumb,
+
+- Don't: use classes to make code look more like an outline
```
hanlint, line 43: what does this mean?

STYLE_GUIDE.md
```diff
@@ -274,6 +294,7 @@ Composer uses [Google Style Docstrings](https://sphinxcontrib-napoleon.readthedo
 
 The following guidelines apply to documentation.
 1.  Each function that needs a docstring must have its input arguments and return statement (if not None) annotated.
```
eracah, line 296: And `Raises`?

eracah (approved):

STYLE_GUIDE.md
ravi-mosaicml: Lol, good point. I think it had something to do with the last bullet, but yea let me remove it. Copied it from https://www.notion.so/Codebase-74ccd984a16c4095a368806225a555b4#23cafb24f4344751a700cd875dc095e9

STYLE_GUIDE.md
ravi-mosaicml: Added a comment that all custom exceptions should be documented.

STYLE_GUIDE.md
ravi-mosaicml: Fixed

PR #1109, opened 2022-05-31
Early Stopping and Threshold Stopping Notebook Example

anisehsani:
Add an example notebook showing how to use the Early Stopping and Threshold Stopping Callbacks


ravi-mosaicml (commented):
Can you merge in the latest dev and add the notebook to the `index.rst` so it will show up on the docs site? That'll make it easier to review. Thanks!

PR #1112, opened 2022-06-01
Remove `shell=True` from the launch script

ravi-mosaicml:
Setting `shell=True` in the launch script prevents the VSCode debugger from traversing subprocesses. See https://github.com/microsoft/vscode/issues/121884

We used `shell=True`, so the user could pass in variables like `$RANK` to the script, which would then be replaced with the actual rank of the process. To use `shell=False`, while still substituting variables like `$RANK`, required the following changes:

* Use `os.path.expandvars(os.path.expanduser(arg))` to expand any variables in the script args
* Since `os.path.expandvars` and `os.path.expanduser` reads from `os.environ`, the env vars must be patched in-place (instead of copying, updating, and passing through the copied dictionary)
* To ensure consistency and avoid polluting the env var namespace, the env vars for the parent process are reset to their original values after each sub-process is launched.

While it could work to treat each launch arg as a format string, that would introduce conflicts when format strings need to be passed into composer (e.g. for checkpoint loading), so would prefer to use shell variable syntax instead.


composer/cli/launcher.py
```diff
@@ -272,6 +273,13 @@ def _get_file(format: str):
             process.stdout = stdout_file
         processes[global_rank] = process
 
+        # Restore the env variables back to the original values
+        for var in ("RANK", "WORLD_SIZE", "LOCAL_RANK", "LOCAL_WORLD_SIZE", "NODE_RANK", "MASTER_ADDR", "MASTER_PORT"):
+            if var in original_env:
+                os.environ[var] = original_env[var]
+            else:
+                del os.environ[var]
```
hanlint, line 281: Should we be doing this? Or just restoring the distributed-related envvars? Regradless, there should be a `log.info` here

ravi-mosaicml: (I re-wrote this with a context manager to make it clearer what is happening)

hanlint: LGTM

hanlint (approved):
LGTM, but might want the original launch writer (cc: @jbloxham ) to take a look?

jbloxham (approved):
This looks fine to me.

PR #1113, opened 2022-06-01
Add GitHub.com links in the API Reference; Open External Links in New Tabs

ravi-mosaicml:
It's more useful to link the docs to the underlying source code rather than to a copy of it on the docs site. This PR replaces `sphinx-viewcode` with `sphinx-linkcode`, which adds links to each docstring entry to the source file and line number.

* For module-level attributes, line numbers are determined by parsing the ast tree. For functions, classes, and function methods, `inspect.getsourcelines` is used to determine the line number of the definition.
* `git` is used to determine the commit hash of the current build, so all docs links will be permalinks and work for both development and release branches.
* In addition, all external links (including existing external links) open in new tabs.

Example:
https://mosaicml-composer--1113.com.readthedocs.build/en/1113/api_reference/composer.profiler.json_trace_handler.html#composer.profiler.json_trace_handler.JSONTraceHandler


bandish-shah (approved):
LGTM

PR #1116, opened 2022-06-01
Embed Notebooks in the Docs

ravi-mosaicml:
* Configured `nbsphinx` to embed sphinx notebooks in the docs site
* Symlink the `examples/` folder in `docs/source/examples`, so notebooks can be included via docs
* Rename the `notebooks` folder to `examples`, to match the structure on the docs site. (Notebooks really are examples, anyways)
* Refactored some notebooks to clean up the headings -- there should only be one h1 (the title); otherwise, it will show up in the sidebar on the left
* Replaced `scripts/clean_notebooks.py` with the `nbstripout` pre-commit hook
* Removed the embedded image from the "custom speedup methods" notebook; instead, uploaded it to the google storage bucket, and linked to it from there, as changing this file was causing the "large file" pre-commit hook to fail


docs/source/conf.py
```diff
@@ -61,8 +70,28 @@
     'sphinx.ext.doctest',
     'sphinx_panels',
     'sphinxcontrib.images',
+    'nbsphinx',
 ]
 
+_GIT_COMMIT = "dev"  # TODO(ravi) -- replace this with the git commit
```
ravi-mosaicml, line 76: This will be fixed after #1113, which includes code to determine the git commit.

examples/custom_speedup_methods.ipynb
```diff
@@ -0,0 +1,837 @@
+{
```
ravi-mosaicml, line 1: This diff is large because I removed the embedded image, and instead put it in the google cloud bucket so it loads externally (large files should not be in the repo).

docs/source/conf.py
ravi-mosaicml: Done

A-Jacobson (approved):
https://mosaicml-composer--1116.com.readthedocs.build/en/1116/examples/medical_image_segmentation.html

LGTM super impressed with this.

PR #1117, opened 2022-06-01
Update code examples for progressive resizing and layer freezing

coryMosaicML:
Adds some missing example code snippets to the method cards.


jfrankle (approved):
LGTM

PR #1122, opened 2022-06-03
Add in the timestamp attributes as wandb artifact metadata

ravi-mosaicml:
Add in the timestamp attributes as wandb artifact metadata

Example artifact: https://wandb.ai/mosaic-ml/ravi-composer-examples/artifacts/ep2-ba58-rank0/1654289178-tuscan-hummingbird.checkpoints.ep2-ba58-rank0/d72a67bc4f0ea580764c/metadata

### DIFF
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -164,7 +164,7 @@ class WandBLogger(LoggerDestination):
             if state.dataloader is not state.train_dataloader:
                 # TODO If not actively training, then it is impossible to tell from the state whether
                 # the trainer is evaluating or predicting. Assuming evaluation in this case.
-                metadata = {f"eval_timestamp/{k}": v for (k, v) in state.eval_timestamp.state_dict().items()}
+                metadata.update({f"eval_timestamp/{k}": v for (k, v) in state.eval_timestamp.state_dict().items()})
 
             artifact = wandb.Artifact(
                 name=new_artifact_name,
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -158,7 +158,19 @@ class WandBLogger(LoggerDestination):
                                f"The artifact with name '{artifact_name}' will be stored as '{new_artifact_name}'."))
 
             extension = new_artifact_name.split(".")[-1]
-            artifact = wandb.Artifact(name=new_artifact_name, type=extension)
+
+            metadata = {f"timestamp/{k}": v for (k, v) in state.timestamp.state_dict().items()}
+            # if evaluating, also log the evaluation timestamp
+            if state.dataloader is not state.train_dataloader:
+                # TODO If not actively training, then it is impossible to tell from the state whether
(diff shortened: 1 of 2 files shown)


composer/loggers/wandb_logger.py
```diff
@@ -158,7 +158,19 @@ def log_file_artifact(self, state: State, log_level: LogLevel, artifact_name: st
...
 
             extension = new_artifact_name.split(".")[-1]
-            artifact = wandb.Artifact(name=new_artifact_name, type=extension)
+
+            metadata = {f"timestamp/{k}": v for (k, v) in state.timestamp.state_dict().items()}
+            # if evaluating, also log the evaluation timestamp
+            if state.dataloader is not state.train_dataloader:
+                # TODO If not actively training, then it is impossible to tell from the state whether
+                # the trainer is evaluating or predicting. Assuming evaluation in this case.
+                metadata = {f"eval_timestamp/{k}": v for (k, v) in state.eval_timestamp.state_dict().items()}
```
mvpatel2000, line 167: Comment is a bit confusing -- should it be logging both `timestamp/` and `eval_timestamp/` (which comment implies with "also") or just one of them? 

Also, if it's the latter, I'm personally in favor of doing something like
```
metadata = None
if state.dataloader is state.train_dataloader:
  metadata = ...
else:
  metadata = ...
```
so it only constructs the dict if necessary, but more minor nit and I'm not sure what's more consistent with rest of the code base

examples/run_composer_trainer.py
```diff
@@ -45,7 +45,7 @@ def _main() -> None:
         pass
     else:
         if wandb.run is not None:
-            wandb.config.update(**hparams.to_dict())
+            wandb.config.update(hparams.to_dict())
```
mvpatel2000, line 48: Why no more `**`?

mvpatel2000 (approved):
LGTM. Minor comments

moinnadeem (approved):
Ravi if I could hug you I would, but that'll have to wait for next week. LGTM!

examples/run_composer_trainer.py
ravi-mosaicml: Unpacking the dictionary didn't work; it threw this exception:

```
  File "/home/ravi/ravi-composer/examples/run_composer_trainer.py", line 70, in <module>
    _main()
  File "/home/ravi/ravi-composer/examples/run_composer_trainer.py", line 48, in _main
    wandb.config.update(**hparams.to_dict())
TypeError: Config.update() got an unexpected keyword argument 'model'
```
The docs here suggest that a dict should be passed through: https://docs.wandb.ai/guides/track/config#efficient-initialization

composer/loggers/wandb_logger.py
ravi-mosaicml: Ah good catch -- I meant to update the metadata, rather than override it. Thanks!

My thoughts are that if you are evaluating, then you would want both the training and evaluation timestamp. But if you're just training, then skip the evaluation timestamp.

PR #1123, opened 2022-06-03
Fix the Speed Monitor

ravi-mosaicml:
The speed monitor sometimes returns negative times (example: https://wandb.ai/mosaic-ml/regression-v0.7.0-RC1/table?workspace=user-ravimosaicml)

* Fixed by re-implementing the speed monitor to use the wall clock tracking built into `state.timestamp` and `state.eval_timestamp`. Using these variables ensures that all ranks have consistent timing information and simplifies the speed monitor implementation, as timestamp variables are already in duration units rather than wall clock timestamps
* Added a validation dataloader and checks to ensure no negative values in the test
* Logging wall clock time on every batch, rather than every epoch, to support NLP and single-epoch jobs
* (Coming along for the ride): Added better tests to catch downstream effects of how loggers handle callbacks that log data.


moinnadeem:
@ravi-mosaicml Good stuff! Do you have a example WandB report of what the speed monitor looks like now? It'll help evaluate the PR pretty easily -- the code looks good to me, but I want to check for downstream effects.

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
-        # log the total wall clock time that this program has been running for
-        self.wall_clock_total = 0.0
+        # Track the epoch num samples and wct to compute throughput over the entire epoch
+        self.epoch_start_num_samples = Time.from_sample(0)
+        self.epoch_start_wct = datetime.timedelta(0)
+
+        # Track the batch num samples and wct to compute throughput over a window of batches
+        self.batch_start_num_samples = Time.from_sample(0)
+        self.batch_start_wct = datetime.timedelta(0)
+        self.rolling_batch_times_and_num_samples: Deque[Tuple[datetime.timedelta,
```
hanlint, line 73: Let's simplify this to just `self.buffer`? The variable name is too long, makes the code hard to read.

Or alternatively, separate it out into a window buffer for wct and one for `num_samples` to mirror everything else here.

moinnadeem (commented):
@ravi-mosaicml Good stuff! Do you have a example WandB report of what the speed monitor looks like now? It'll help evaluate the PR pretty easily -- the code looks good to me, but I want to check for downstream effects.

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
+                                                              Time[int]]] = deque(maxlen=window_size)
+        self.window_size = window_size
 
-        # log the total wall clock_time that has been spent in training
-        self.wall_clock_train = 0.0
+        # To optimize performance for the the batch wct calculations, keep a running sum over the entire deque
+        self.running_wct = datetime.timedelta(0)
```
hanlint, line 78: we use `running` here but `total` for evaluation.. maybe just use `total`?

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
+            "batch_start_wct": self.batch_start_wct,
+            "rolling_batch_times_and_num_samples": self.rolling_batch_times_and_num_samples,
+            "running_wct": self.running_wct,
+            "running_num_samples": self.running_num_samples,
+            "total_eval_wct": self.total_eval_wct,
         }
 
     def load_state_dict(self, state: Dict[str, Any]) -> None:
-        self.loaded_state = state
+        self._loaded_state = state
```
hanlint, line 99: could you add a comment to explain why we don't just load the state direclty here, but instead wait for one of the events?

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
+            self.running_num_samples = self._loaded_state["running_num_samples"]
+            self.total_eval_wct = self._loaded_state["total_eval_wct"]
+            self._loaded_state = None
 
-    def batch_start(self, state: State, logger: Logger) -> None:
+    def epoch_start(self, state: State, logger: Logger):
         del logger  # unused
         self._load_state()
-        self.batch_start_num_samples = state.timestamp.sample
+        self.rolling_batch_times_and_num_samples.clear()
```
hanlint, line 119: do we need to clear across the epoch boundary?

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
-        self.epoch_start_time = time.time()
-        self.batch_end_times.clear()
-        self.batch_num_samples.clear()
-        self.train_examples_per_epoch = 0
+        self.batch_start_wct = state.timestamp.total_wct
+        self.batch_start_num_samples = state.timestamp.sample
 
     def batch_end(self, state: State, logger: Logger):
-        self.batch_end_times.append(time.time())
         new_num_samples = state.timestamp.sample
```
hanlint, line 130: new_num_samples is only used once, is this still needed?

composer/callbacks/speed_monitor.py
```diff
@@ -63,106 +63,106 @@ class SpeedMonitor(Callback):
...
+        batch_num_samples = new_num_samples - self.batch_start_num_samples
+        self.running_num_samples += batch_num_samples
+        batch_wct = state.timestamp.total_wct - self.batch_start_wct
+        self.running_wct += batch_wct
+
+        if len(self.rolling_batch_times_and_num_samples) == self.window_size:
+            # the buffer is full. subtract out the oldest num samples and wct from the running tallies, before the new entry is added
+            oldest_wct, oldest_num_samples = self.rolling_batch_times_and_num_samples.popleft()
+            self.running_num_samples -= oldest_num_samples
+            self.running_wct -= oldest_wct
```
hanlint, line 140: how much does this optimization save us? I think we are saving a .sum() of a few numbers here, if I understand this correctly. Maybe it's not worth having to keep extra variables and this logic here, and just let deque do its thing?

tests/callbacks/test_speed_monitor.py
```diff
@@ -31,6 +45,12 @@ def test_speed_monitor():
     wall_clock_total_calls = len(in_memory_logger.data['wall_clock/total'])
     throughput_step_calls = len(in_memory_logger.data['samples/step'])
     throughput_epoch_calls = len(in_memory_logger.data['samples/epoch'])
+    _assert_no_negative_values(in_memory_logger.data['wall_clock/train'])
```
hanlint, line 48: Did this test actually create negative values before this PR?

hanlint (approved):
Forgot to approve, most comments above are on improving code readability.

moinnadeem:
@ravi-mosaicml Glad I checked for downstream effects -- I get the following error when testing this PR:
```
  File "/root/composer/examples/run_composer_trainer.py", line 70, in <module>
    _main()
  File "/root/composer/examples/run_composer_trainer.py", line 66, in _main
    trainer.fit()
  File "/root/composer/composer/trainer/trainer.py", line 1293, in fit
    self._train_loop()
  File "/root/composer/composer/trainer/trainer.py", line 1507, in _train_loop
    self.engine.run_event(Event.BATCH_END)
  File "/root/composer/composer/core/engine.py", line 241, in run_event
    self._run_callbacks(event)
  File "/root/composer/composer/core/engine.py", line 361, in _run_callbacks
    cb.run_event(event, self.state, self.logger)
  File "/root/composer/composer/core/callback.py", line 97, in run_event
    return event_cb(state, logger)
  File "/root/composer/composer/callbacks/speed_monitor.py", line 151, in batch_end
    logger.data_batch({
  File "/root/composer/composer/loggers/logger.py", line 208, in data_batch
    self.data(LogLevel.BATCH, data)
  File "/root/composer/composer/loggers/logger.py", line 133, in data
    destination.log_data(self._state, log_level, data)
  File "/root/composer/composer/loggers/wandb_logger.py", line 106, in log_data
    wandb.log(data, step=int(state.timestamp.batch))
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/wandb_run.py", line 256, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/wandb_run.py", line 222, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/wandb_run.py", line 1548, in log
    self._log(data=data, step=step, commit=commit)
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/wandb_run.py", line 1339, in _log
    self._partial_history_callback(data, step, commit)
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/wandb_run.py", line 1228, in _partial_history_callback
    self._backend.interface.publish_partial_history(
  File "/usr/local/lib/python3.9/dist-packages/wandb/sdk/interface/interface.py", line 548, in publish_partial_history
    item.value_json = json_dumps_safer_history(v)
  File "/usr/local/lib/python3.9/dist-packages/wandb/util.py", line 820, in json_dumps_safer_history
    return json.dumps(obj, cls=WandBHistoryJSONEncoder, **kwargs)
  File "/usr/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.9/dist-packages/wandb/util.py", line 787, in default
    return json.JSONEncoder.default(self, obj)
  File "/usr/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type timedelta is not JSON serializable
```

composer/callbacks/speed_monitor.py
ravi-mosaicml: I like the latter suggustion

composer/callbacks/speed_monitor.py
ravi-mosaicml: I'll rename `running` to `window` to be more specific. The evaluation time is across all batches and epochs, not just the current window.

composer/callbacks/speed_monitor.py
ravi-mosaicml: Good question, I think this was a relic from the previous version where we stored the timestamp, rather than the duration, in the internal state. Now this is not needed anymore.

composer/callbacks/speed_monitor.py
ravi-mosaicml: Nope, don't need to -- this is how it was done before, but just removed it.

tests/callbacks/test_speed_monitor.py
ravi-mosaicml: I'm not sure. I never traced down the cause of the negative values.

moinnadeem (approved):
Much better, [it works](https://wandb.ai/mosaic-ml/composer-examples/runs/1v63b9h2?workspace=user-moin-mosaic)! Bellissimo, I like this behavior a lot better too.

PR #1125, opened 2022-06-04
Ignore Checkpoint Params

mvpatel2000:
* Add `load_ignore_keys` to support removing keys from a checkpoint via a list of glob strings or a filter function


composer/trainer/trainer.py
```diff
@@ -491,6 +491,25 @@ class Trainer:
...
+            of strings specifying the keys to index into ``state_dict``. If a prefix is provided, all children are also
+            ignored.
+
+            Example 1: `load_ignore_model_keys = [["state", "model", "classifier", "weights"], "state", "model", "classifier", "bias"]]`
+            would ignore the corresponding weights and biases of the classifier.
+
+            Example 2: In the above example, if these were the only parameters for the classifier, alternatively
+            `load_ignore_model_keys = [["state", "model", "classifier"]]` would have the same effect.
+
+            Example 3: `load_ignore_model_keys = [["state", "rank_zero_seed"], ["rng"]]` would reset all randomness when
```
hanlint, line 505: I don't see "rng" as part of the saved checkpoint format (e.g. https://docs.mosaicml.com/en/v0.7.0/trainer/checkpointing.html#anatomy-of-a-checkpoint). Is something missing form the docs?

composer/trainer/trainer.py
```diff
@@ -491,6 +491,25 @@ class Trainer:
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
         load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        load_ignore_keys (List[List[str]] | (Dict) -> None, optional): A list of paths for the ``state_dict``,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict``. If a prefix is provided, all children are also
+            ignored.
+
+            Example 1: `load_ignore_model_keys = [["state", "model", "classifier", "weights"], "state", "model", "classifier", "bias"]]`
```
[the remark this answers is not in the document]
ravi-mosaicml: YAHP won't be able to parse a nested list, so using a string here with a seperator would be preferred. However, thoughts about using `"/" to denote nested sub-dictionaries instead of "." to be more consistent with glob syntax (and I think pytorch natively uses "." in model parameter names?) Then, each path would be a glob string.

composer/utils/checkpoint.py
```diff
@@ -256,17 +276,39 @@ def _download_checkpoint(
     return composer_states_filepath, extracted_checkpoint_folder, extracted_rank_n
 
 
+def glob_filter(exclude_globs: List[List[str]]) -> Callable[[Dict], None]:
+    """Provides a function which deletes all subparts of a dictionary based on a list of paths."""
+
+    def filter_func(state_dict: Dict) -> None:
+        for exclude_glob in exclude_globs:
```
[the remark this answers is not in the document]
ravi-mosaicml: (See comment above, but would prefer '/' instead of '.' since individual parameter names could contain '.')

composer/trainer/trainer.py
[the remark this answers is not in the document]
hanlint: I recall in the discussion that `.` feels more natural to our users to align with how pytorch indexes into nested `nn.Module`. "/" might be confused with filepath directories.

composer/utils/checkpoint.py
```diff
@@ -256,17 +276,39 @@ def _download_checkpoint(
...
+            temp_dict = state_dict
+            for step in exclude_glob[:-1]:
+                temp_dict = temp_dict[step]
+            del temp_dict[exclude_glob[-1]]
```
ravi-mosaicml, line 287: Would it be possible to use https://docs.python.org/3.8/library/fnmatch.html to evaluate globs? (Though matching against '**' may require some custom logic -- it looks like fnmatch doesn't support that :'( )

ravi-mosaicml (commented):
I agree with @hanlint's to use `List[str]` instead of `List[List[str]]`, though would prefer to use `/` to denote nested dictionaries rather than `.`, so it will look more like a filepath string (which is what glob is meant to to be used for) and to avoid conflicts with parameter names that contain periods. 

For glob syntax parsing, would using https://docs.python.org/3.8/library/fnmatch.html work?

composer/trainer/trainer.py
mvpatel2000:
```
>>> list(state_dict)
['state', 'rng']
```
6th line in the code snippet in the link

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: Was basing this off Moin's example [here](https://github.com/mosaicml/composer/issues/497) -- will switch to layer1 instead for clarity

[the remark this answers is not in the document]
mvpatel2000: We don't support wildcards right now, but I can change it to do so

composer/trainer/trainer.py
```diff
@@ -491,6 +491,25 @@ class Trainer:
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
         load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        load_ignore_keys (List[List[str]] | (Dict) -> None, optional): A list of paths for the ``state_dict``,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict``. If a prefix is provided, all children are also
```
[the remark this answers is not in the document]
mvpatel2000: what example 2 is doing -- made this more clear

composer/trainer/trainer.py
moinnadeem: Hmm, this nested dictionary isn't actually super clear. I'm fine with it, but I don't think that it is the best possible UX. Could we do a string that separates on some character instead, or wildcard syntax?

composer/utils/checkpoint.py
mvpatel2000: I avoided this because we're indexing into keys, and I don't want to flatten the state_dict into a series of paths, match, and then go back and delete when recursing is a little simpler. Let me know if there's a particular reason you think this would be better?

mvpatel2000:
Switched to `/` since Ravi pointed out `.` can occur in param names.

moinnadeem (commented):
Implementation looks good to me, but we should discuss a little on the API or filter function design

composer/trainer/trainer.py
mvpatel2000: lol I think I pushed the change to `/` separator after you started the review but before you finished it -- tough thread race :'(

composer/trainer/trainer.py
```diff
@@ -491,6 +491,26 @@ class Trainer:
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
         load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        load_ignore_keys (List[str] | (Dict) -> None, optional): A list of paths for the ``state_dict`` of the checkpoint,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict`` joined together with `/` as a seperator.
+            If a prefix is provided, all children are also ignored (as demonstrated in Example 2). See :mod:`composer.core.state`
+            for the structure of state_dict.
+
+            Example 1: `load_ignore_model_keys = ["state/model/layer1/weights", "state/model/layer1/bias"]` would ignore
```
hanlint, line 500:
```suggestion
            Example 1: `load_ignore_keys = ["state/model/layer1/weights", "state/model/layer1/bias"]` would ignore
```

hanlint: Might be useful in the example to use exactly what the torch format is for model weights, from @ravi-mosaicml 's comment, seems like it should be "state/model/layer1.bias", but would be good to verify?

This would also make clear why we chose `/` instead of `.`

composer/trainer/trainer.py
```diff
@@ -491,6 +491,26 @@ class Trainer:
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
         load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        load_ignore_keys (List[str] | (Dict) -> None, optional): A list of paths for the ``state_dict`` of the checkpoint,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict`` joined together with `/` as a seperator.
+            If a prefix is provided, all children are also ignored (as demonstrated in Example 2). See :mod:`composer.core.state`
```
hanlint, line 497:
```suggestion
            If a prefix is provided, all children are also ignored (see Example 2). See :mod:`composer.core.state`
```

composer/trainer/trainer.py
```diff
@@ -491,6 +491,26 @@ class Trainer:
...
+            for the structure of state_dict.
+
+            Example 1: `load_ignore_model_keys = ["state/model/layer1/weights", "state/model/layer1/bias"]` would ignore
+            layer 1 weights and bias.
+
+            Example 2: `load_ignore_model_keys = ["state/model/layer1"]` would ignore layer1, which would have the same
+            effect as the previous example if `weights` and `bias` were the only parameters in layer`.
+
+            Example 3: `load_ignore_model_keys = ["state/rank_zero_seed", "rng"]` would reset all randomness when
```
hanlint, line 506: add wildcard example?

hanlint (approved):
LGTM, but should get @ravi-mosaicml 's approval as well. Main suggestion is to have the examples conform to exactly how we store it in composer.

composer/utils/checkpoint.py
```diff
@@ -125,6 +126,25 @@ def load_checkpoint(
             Ignored if the checkpoint is a local file path. (default: ``1_048_576`` bytes (1 MB))
         progress_bar (bool, optional): Whether or not to show a progress bar when downloading checkpoints.
             Ignored if the checkpoint is a local file path. (default: ``True``)
+        ignore_keys (List[str] | (Dict) -> None, optional): A list of paths for the ``state_dict`` of the checkpoint,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict`` joined together with `/` as a seperator.
+            If a prefix is provided, all children are also ignored (as demonstrated in Example 2). See :mod:`composer.core.state`
+            for the structure of state_dict.
+
+            Example 1: `load_ignore_model_keys = ["state/model/layer1/weights", "state/model/layer1/bias"]` would ignore
```
ravi-mosaicml, line 135: For readability, can these examples be converted into one big code block?
In addition, the variables here are `ignore_keys`, not `load_ignore_model_keys`.

composer/utils/checkpoint.py
```diff
@@ -256,17 +277,50 @@ def _download_checkpoint(
...
+    """Provides a function which deletes all subparts of a dictionary based on a list of paths."""
+
+    def filter_func(state_dict: Dict) -> None:
+
+        def recur_filter_func(state_dict: Dict, exclude_glob: List[str]):
+            # End of path, delete key in state_dict
+            if len(exclude_glob) == 1:
+                del state_dict[exclude_glob[0]]
+            # Wildcard, descend down all keys
+            elif exclude_glob[0] == "*":
```
ravi-mosaicml, line 290: Instead of manually parsing glob, can you use https://docs.python.org/3/library/fnmatch.html#module-fnmatch to support `?`, `[something]`, and `[!something]` notion?

composer/utils/checkpoint.py
```diff
@@ -256,17 +277,50 @@ def _download_checkpoint(
...
+                for key in state_dict.keys():
+                    recur_filter_func(state_dict[key], exclude_glob[1:])
+            # Step down through path
+            else:
+                recur_filter_func(state_dict[exclude_glob[0]], exclude_glob[1:])
```
ravi-mosaicml, line 295: If the `state_dict` does not have this key, then I think it's raise a hard-to-debug error (e.g. `KeyError 'bias' not found`). Could we make it more debuggable by re-raising the exception to include the full glob path that caused the error to be generated (e.g. `state/model/` has no key `conv1.bias`)

Another thought -- the implementation would raise an exception if one specifies an exclude glob that is not in the state dict. Should this be a warning or a silent ignore (for better portability?)

composer/utils/checkpoint.py
ravi-mosaicml: I think if one ends the path in a wildcard, then I presume that should remove all subkeys?

ravi-mosaicml: In addition,can we support using a wildcard as part of a key (e.g. `conv*` to remove all conv keys)?

composer/trainer/trainer.py
ravi-mosaicml: (I'd need to look at `model.named_parameters()` but that is my intuition)

ravi-mosaicml (commented):
See comments -- would prefer to use `fnmatch` for the glob parsing. There are some edge cases which manual parsing makes challenging.

composer/trainer/trainer.py
mvpatel2000: Yep, that's correct, can update examples accordingly.

composer/utils/checkpoint.py
mvpatel2000: Sorry, what exactly do you mean by convert into one big code block?

composer/utils/checkpoint.py
mvpatel2000: moving discussion offline

composer/utils/checkpoint.py
ravi-mosaicml: Nvm; I thought there would be more a a functional api we could call, but scratch this.

composer/trainer/trainer.py
```diff
@@ -491,6 +491,28 @@ class Trainer:
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``1,048,675``)
         load_progress_bar (bool, optional): Display the progress bar for downloading the checkpoint.
             Ignored if ``load_path`` is either ``None`` or a local file path. (default: ``True``)
+        load_ignore_keys (List[str] | (Dict) -> None, optional): A list of paths for the ``state_dict`` of the checkpoint,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict`` joined together with `/` as a seperator (as PyTorch
+            uses `.` in parameter names). If a prefix is provided, all children are also ignored (see Example 2). 
+            See :mod:`composer.core.state` for the structure of state_dict.
+
+            Example 1: `load_ignore_keys = ["state/model/layer1.weights", "state/model/layer1.bias"]` would ignore
```
ravi-mosaicml, line 500:
```suggestion
            Example 1: ``load_ignore_keys = ["state/model/layer1.weights", "state/model/layer1.bias"]`` would ignore
```

composer/trainer/trainer.py
```diff
@@ -491,6 +491,28 @@ class Trainer:
...
+            layer 1 weights and bias.
+
+            Example 2: `load_ignore_keys = ["state/model"]` would ignore the entire model, which would have the same
```
ravi-mosaicml, line 503:
```suggestion
            Example 2: ``load_ignore_keys = ["state/model"]`` would ignore the entire model, which would have the same
```

composer/trainer/trainer.py
```diff
@@ -491,6 +491,28 @@ class Trainer:
...
+            effect as the previous example if there was only 1 layer.
+
+            Example 3: `load_ignore_keys = ["state/model/*"]` would also ignore the entire model using wildcard syntax.
```
ravi-mosaicml, line 506:
```suggestion
            Example 3: ``load_ignore_keys = ["state/model/*"]`` would also ignore the entire model using wildcard syntax.
```

composer/trainer/trainer.py
```diff
@@ -491,6 +491,28 @@ class Trainer:
...
+
+            Example 4: `load_ignore_keys = ["state/rank_zero_seed", "rng"]` would reset all randomness when
```
ravi-mosaicml, line 508:
```suggestion
            Example 4: ``load_ignore_keys = ["state/rank_zero_seed", "rng"]`` would reset all randomness when
```

composer/utils/checkpoint.py
```diff
@@ -125,6 +126,27 @@ def load_checkpoint(
             Ignored if the checkpoint is a local file path. (default: ``1_048_576`` bytes (1 MB))
         progress_bar (bool, optional): Whether or not to show a progress bar when downloading checkpoints.
             Ignored if the checkpoint is a local file path. (default: ``True``)
+        ignore_keys (List[str] | (Dict) -> None, optional): A list of paths for the ``state_dict`` of the checkpoint,
+            which, when provided, will be ignored from the state_dict before a checkpoint is loaded. Each path is a list
+            of strings specifying the keys to index into ``state_dict`` joined together with `/` as a seperator (as PyTorch
+            uses `.` in parameter names). If a prefix is provided, all children are also ignored (see Example 2). 
+            See :mod:`composer.core.state` for the structure of state_dict.
+
+            Example 1: `ignore_keys = ["state/model/layer1.weights", "state/model/layer1.bias"]` would ignore
```
ravi-mosaicml, line 135: Same tick mark comments as above

composer/utils/checkpoint.py
mvpatel2000: Will rework to log ignored keys -- `log.info` for ones excluded and `log.warn` for passed in globs that don't match

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
...
+            if len(filtered_paths_from_glob) == 0:
+                log.warning(
+                    f"No parts from loaded checkpoint state_dict were ignored by load_ignore_key {exclude_glob}")
+            filtered_paths.extend(filtered_paths_from_glob)
+        filtered_paths = list(set(filtered_paths))
+        filtered_paths_str = " ".join(filtered_paths)
+        if len(filtered_paths_str) > 0:
+            log.info(f"Ignoring the following paths from the loaded checkpoint state_dict: {filtered_paths_str}")
+
+        def remove_path(state_dict: Dict, exclude_glob: List[str]):
```
ravi-mosaicml, line 315: Same comments as above re: moving this into a private helper and supporting lists.

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
...
 
+def glob_filter(exclude_globs: List[str]) -> Callable[[Dict], None]:
+    """Provides a function which deletes all subparts of a dictionary based on a list of paths."""
+
+    def filter_func(state_dict: Dict) -> None:
+
+        def flatten_state_dict(state_dict: Any, paths: List[str], existing_path: str):
+            """Recursively convert a dictionary into a set of paths."""
+            # Store path when we reach end, which is either non-Dict or empty Dict
+            if not isinstance(state_dict, Dict) or len(state_dict.keys()) == 0:
```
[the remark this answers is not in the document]
ravi-mosaicml: Re-wrote the implementation to support lists; this is outdated.

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
     return composer_states_filepath, extracted_checkpoint_folder, extracted_rank_n
 
 
+def glob_filter(exclude_globs: List[str]) -> Callable[[Dict], None]:
+    """Provides a function which deletes all subparts of a dictionary based on a list of paths."""
+
+    def filter_func(state_dict: Dict) -> None:
+
+        def flatten_state_dict(state_dict: Any, paths: List[str], existing_path: str):
```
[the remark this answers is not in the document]
ravi-mosaicml: Done.

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
...
+        # Flatten dictionary into paths
+        paths = []
+        flatten_state_dict(state_dict, paths, '/')
+        print(paths)
+
+        filtered_paths = []
+        for exclude_glob in exclude_globs:
+            filtered_paths_from_glob = fnmatch.filter(paths, exclude_glob)
+            if len(filtered_paths_from_glob) == 0:
+                log.warning(
```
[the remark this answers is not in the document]
ravi-mosaicml: Done.

composer/utils/checkpoint.py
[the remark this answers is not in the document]
ravi-mosaicml: Done

composer/utils/checkpoint.py
ravi-mosaicml: Done.

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
...
+                    f"No parts from loaded checkpoint state_dict were ignored by load_ignore_key {exclude_glob}")
+            filtered_paths.extend(filtered_paths_from_glob)
+        filtered_paths = list(set(filtered_paths))
+        filtered_paths_str = " ".join(filtered_paths)
```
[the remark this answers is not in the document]
ravi-mosaicml: Done.

composer/utils/checkpoint.py
```diff
@@ -255,17 +279,74 @@ def _download_checkpoint(
...
     extracted_checkpoint_folder: Optional[str],
     load_weights_only: bool,
     strict_model_weights: bool,
+    ignore_keys: Optional[Union[List[str], Callable[[Dict], None]]],
 ) -> Optional[List[Dict[str, Any]]]:
     """Restore a checkpoint into ``state`` and returns the rng state dicts (if ``load_weights_only`` is False)."""
     # Now, all ranks load the checkpoint that local rank zero downloaded
     state_dict = torch.load(composer_states_filepath, map_location='cpu')
+    if ignore_keys:
+        if state.is_model_deepspeed:
```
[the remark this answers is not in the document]
ravi-mosaicml: Done.

PR #1127, opened 2022-06-06
Abstract Object Store Class

anisehsani:
Abstract class for implementing object_store utilities like Libcloud, s3, sftp.


ravi-mosaicml (approved):
LGTM. We'll need to eventually add in an API for callers to handle retries (either an argument or some sort of uniform exception), but can add that in later.

PR #1128, opened 2022-06-06
(a) StreamingCIFAR10, (b) remove StreamingImageClassDataset.

knighton:
Add an implementation of the classic `CIFAR10` dataset using the MosaicML `StreamingDataset`.

### DIFF
--- a/composer/datasets/cifar.py
+++ b/composer/datasets/cifar.py
@@ -1,37 +1,24 @@
-# Copyright 2022 MosaicML Composer authors
-# SPDX-License-Identifier: Apache-2.0
-
-"""ImageNet classification streaming dataset.
-
-The most widely used dataset for Image Classification algorithms. Please refer to the `ImageNet 2012 Classification
-Dataset <http://image-net.org/>`_ for more details.
-"""
-
-from io import BytesIO
 import numpy as np
 import os
 from PIL import Image
-from typing import Any, List, Optional
+from typing import Any, Optional
 
-import torch
 from torchvision import transforms
 
 from composer.datasets.streaming import StreamingDataset
 
-__all__ = ["StreamingImageNet1k"]
+__all__ = ["StreamingCIFAR10"]
 
 
-class StreamingImageNet1k(StreamingDataset):
+class StreamingCIFAR10(StreamingDataset):
     """
-    Implementation of the ImageNet1k dataset using StreamingDataset.
+    Implementation of the CIFAR10 dataset using StreamingDataset.
 
     Args:
         remote (str): Remote directory (S3 or local filesystem) where dataset is stored.
         local (str): Local filesystem directory where dataset is cached during operation.
-        split (str): The dataset split to use, either 'train' or 'val'.
+        is_train (bool): Whether is training split.
         shuffle (bool): Whether to shuffle the samples in this dataset.
-        resize_size (int, optional): The resize size to use. Use -1 to not resize. Default: ``-1``.
-        crop size (int): The crop size to use. Default: ``224``.
         batch_size (Optional[int]): Hint batch_size that will be used on each device's DataLoader. Default: ``None``.
     """
 
@@ -44,7 +31,9 @@ class StreamingImageNet1k(StreamingDataset):
         Returns:
             Image: PIL image encoded by the bytes.
         """
-        return Image.open(BytesIO(data)).convert('RGB')
+        arr = np.frombuffer(data, np.uint8)
+        arr = arr.reshape(32, 32, 3)
+        return Image.fromarray(arr)
 
     def decode_class(self, data: bytes) -> np.int64:
         """Decode the sample class.
@@ -60,12 +49,11 @@ class StreamingImageNet1k(StreamingDataset):
     def __init__(self,
                  remote: str,
                  local: str,
-                 split: str,
+                 is_train: bool,
                  shuffle: bool,
-                 resize_size: int = -1,
-                 crop_size: int = 224,
                  batch_size: Optional[int] = None):
         # Build StreamingDataset
+        split = 'train' if is_train else 'val'
         decoders = {
             'x': self.decode_image,
             'y': self.decode_class,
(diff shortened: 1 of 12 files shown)


composer/datasets/cifar.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Args:
+            data (bytes): The raw bytes.
+
+        Returns:
+            np.int64: The class encoded by the bytes.
+        """
+        return np.frombuffer(data, np.int64)[0]
+
+    def __init__(self, remote: str, local: str, is_train: bool, shuffle: bool, batch_size: Optional[int] = None):
```
abhi-mosaic, line 52: Could we replace `is_train: bool` with a `split: str`? I know it's different from the original torchvision dataset but I think it's more clear.

composer/datasets/cifar_hparams.py
```diff
@@ -166,3 +168,25 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+        local (str): Local filesystem directory where dataset is cached during operation.
+    """
+
+    remote: str = hp.optional('Remote directory (S3 or local filesystem) where dataset is stored',
+                              default='s3://mosaicml-internal-dataset-cifar10/mds/1/')
+    local: str = hp.optional('Local filesystem directory where dataset is cached during operation',
+                             default='/tmp/mds-cache/mds-cifar10/')
+
+    def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHparams) -> DataLoader:
+        dataset = StreamingCIFAR10(self.remote, self.local, self.is_train, self.shuffle, batch_size)
```
abhi-mosaic, line 188: Could we use named args like `StreamingCIFAR10(remote=self.remote, local=self.local, ...)`? I get nervous because sometimes we add new things like retries, maybe a new transform arg, etc.

abhi-mosaic (commented):
Mostly LGTM, I agree that removing the `StreamingImageClassDataset` makes sense and improves readability for users who want to build their own datasets.

Could we add the `scripts/mds/...` file as well for generating the CIFAR dataset?

composer/datasets/cifar.py
knighton: Done

composer/datasets/cifar_hparams.py
knighton: Done

abhi-mosaic:
Btw @knighton could you assist @dskhudia with changes to the FFCV writer PR (https://github.com/mosaicml/composer/pull/1095/files) which currently uses `StreamingImageClassDataset`? I think the samples can be read as either a plain `StreamingDataset` class, or a subclass with the desired transforms.

Meta-comment -- we will want to adopt datapipe-style processing to allow users to reuse a dataset class and add new transforms on top, if they want to avoid subclassing.

scripts/ffcv/create_ffcv_datasets.py
```diff
@@ -140,6 +138,20 @@ def cache_streaming(remote, local, num_workers=64):
...
+        A subclass of StreamingDataset that asserts and returns samples that are
+        tuples (PIL.Image, np.int64).
+        Intended to enforce sample types for pipelines that expect the tuple format (image, class)
+    """
+    def __getitem__(self, idx: int) -> Tuple[Image.Image, np.int64]:
+        obj = super().__getitem__(idx)
+        assert 'x', 'y' in obj.keys()
+        assert isinstance(obj['x'], Image.Image)
+        assert isinstance(obj['y'], np.int64)
+        return (obj['x'], obj['y'])
```
abhi-mosaic, line 152: @dskhudia the FFCV writer, as well as our PIL image collator, expect samples to be tuples not dicts, so I had to add this step here.

scripts/ffcv/create_ffcv_datasets.py
```diff
@@ -158,10 +170,28 @@ def _main():
...
+                return Image.fromarray(arr)
+            def _decode_class(data):
+                return np.frombuffer(data, dtype=np.int64)[0]
+        elif args.dataset == 'imagenet':
+            def _decode_image(data):
+                return Image.open(BytesIO(data)).convert('RGB')
+            def _decode_class(data):
+                return np.frombuffer(data, dtype=np.int64)[0]
+        decoders = {'x': _decode_image, 'y': _decode_class}
+        dataset = PILImageClassStreamingDataset(remote=remote, local=local, decoders=decoders, shuffle=False)
```
abhi-mosaic, line 189: @knighton @dblalock just wanted to draw some eyes here for future UX

Some of this could potentially be swept away if `StreamingDataset` has an auto decoder, but still, I think there's some issue of modularity...

If a dataset author writes+uploads files, and then they write a dataset reader, they're probably going to want to add in all their training processing steps, and the object they build, like `StreamingImageNet1k` will have all that processing stuff.

If a second person comes along and wants to read the files in a custom way, right now they have to build their own reader class, like here.

If we want modularity, then the original dataset reader author has to either be proactive and provide args for doing so, or maybe they have to build multiple "pipes" like `ImageNet1kReaderPipe`, and `ImageNet1kAugmentationPipe`, `ImageNet1kTrainPipe`, etc. Where the writing burden should fall is not clear to me. 

The only thing I'm feeling more certain of is that enforcing encodings / sample types is really brittle. Even ImageNet / CIFAR10 do not store their data in the same format, despite both being image classification datasets! That's why we have different decoders here. And the sample type (PIL, class_id) format is a standard from `torchvision`, but it breaks almost immediately once you get out of image classification land. 

FWIW I don't think this problem is unique to us, it's the same issue motivating `torchdata` towards lots of little pipes... and I guess that philosophy would say that the second person should write their own custom reader. Food for thought...

scripts/ffcv/create_ffcv_datasets.py
dskhudia: Looks great. Thank you!

abhi-mosaic:
@dskhudia I tested FFCV conversion for both CIFAR10 and ImageNet but if you could please try out E2E convergence tests after PR merge to make sure nothing went wrong, that would be great.

dskhudia:
> @dskhudia I tested FFCV conversion for both CIFAR10 and ImageNet but if you could please try out E2E convergence tests after PR merge to make sure nothing went wrong, that would be great.

Sure. Will launch a full run once this gets merged.

abhi-mosaic (approved):
Gonna approve and push this through as it's blocking some time-sensitive stuff.

PR #1129, opened 2022-06-06
Libcloud downloads to a tempfile

ravi-mosaicml:
Modified `download_object` for the libcloud object store to first download to a tempfile and then rename if successful, as to eliminate partially downloaded corrupted files at the ``destination_path``.

### DIFF
--- a/composer/utils/libcloud_object_store.py
+++ b/composer/utils/libcloud_object_store.py
@@ -5,7 +5,6 @@
 import os
 import sys
 import tempfile
-import time
 import uuid
 from typing import Any, Dict, Iterator, Optional, Union
 
@@ -206,14 +205,14 @@ class LibcloudObjectStore:
 
         obj = self._get_object(object_name)
         # Download first to a tempfile, and then rename, in case if the file gets corrupted in transit
-        tmp_filepath = destination_path + f".{time.time()}.tmp"
+        tmp_filepath = destination_path + f".{uuid.uuid4()}.tmp"
         try:
             self._provider.download_object(
                 obj=obj,
                 destination_path=tmp_filepath,
             )
         except:
-            # The download failed for some reason. Make a best-effort attempt to remote the temporary file.
+            # The download failed for some reason. Make a best-effort attempt to remove the temporary file.
             try:
                 os.remove(tmp_filepath)
             except OSError:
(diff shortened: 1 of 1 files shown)


composer/utils/libcloud_object_store.py
```diff
@@ -183,29 +184,47 @@ def get_object_size(self, object_name: str) -> int:
...
+            raise FileExistsError(
+                f"destination_path {destination_path} exists and overwrite_existing was set to False.")
+
         obj = self._get_object(object_name)
-        self._provider.download_object(obj=obj,
-                                       destination_path=destination_path,
-                                       overwrite_existing=overwrite_existing,
-                                       delete_on_failure=delete_on_failure)
+        # Download first to a tempfile, and then rename, in case if the file gets corrupted in transit
+        tmp_filepath = destination_path + f".{time.time()}.tmp"
```
mvpatel2000, line 209: I would prefer generating random file names with `uuid.uuid4()` -- I believe certain systems only support second level granularity with `time.time()`, so this could (while extremely rare) have namespace collisions.

composer/utils/libcloud_object_store.py
```diff
@@ -183,29 +184,47 @@ def get_object_size(self, object_name: str) -> int:
...
+        try:
+            self._provider.download_object(
+                obj=obj,
+                destination_path=tmp_filepath,
+            )
+        except:
+            # The download failed for some reason. Make a best-effort attempt to remote the temporary file.
```
mvpatel2000, line 216: minor typo: `remote` -> `remove`

mvpatel2000 (approved):
LGTM! Minor comments

composer/utils/libcloud_object_store.py
```diff
@@ -183,29 +184,47 @@ def get_object_size(self, object_name: str) -> int:
...
+            try:
+                os.remove(tmp_filepath)
+            except OSError:
+                pass
+            raise
+
+        # The download was successful.
+        if overwrite_existing:
+            os.replace(tmp_filepath, destination_path)
```
anisehsani, line 225: I didn't know the difference between `os.rename` and `os.replace`, but looked it up and they do they same thing? Not sure it has to do with the `overwrite_existing` behavior, but just extra error handling on os.replace and os.replace is cross platform.

ravi-mosaicml: `os.replace` will override an existing file, vs `os.rename` would error if the existing file exists. I believe both functions are atomic which is awesome.

anisehsani (approved):

PR #1130, opened 2022-06-06
refactor bert and gpt

A-Jacobson:
I took a stab at converting `GPT2` and `BERT` to use `HuggingFaceModel` and deleting `transformer_shared` and `transformer_hparams`. I'd love for you @moinnadeem to check it out before I/or someone else tests them on the cluster as there are a few things that will probably explode despite the passing tests.

The main changes are in `bert/model.py` and `gpt/model.py` I converted the model creations to factory functions and moved all logic out of the hparams classes. 

I took some liberties that you may find.. questionable so I'm going to put them front and center so you don't have to dig. Please correct me if i'm very wrong in some of these assumptions.

Stuff that's probably going to break:
1. sequence length warmup - For some reason this isn't making my tests fail, but I can see that implementation relies on getting `model_inputs`. I didn't implement that in `HuggingFaceModel` because it seemed like it seemed like it was mostly for error checking and more related to the dataset being used. It's just not a great experience to have to manually specify these. Is this something we necessarily have to manually add or could it be inferred automagically from either the model or dataset?
2. default torchmetrics? - Maybe? I saw that the original `BERTModel` had these lines:
```      
# if we are in the single class case, then remove the classes dimension
        if output.shape[1] == 1:
            output = output.squeeze(dim=1)
```
HuggingFaceModel doesn't do this. Will this break GLUE? is this general enough to torchmetrics that i should add it to the base HuggingFaceModel?

Possible sus design choices:
1. I removed some of the validations on input args, mainly the one that prevents users from specifying a config and `pretrained=True`. Hugging Face does allow this behavior, they just load all the weights that are applicable and throw a warning. How do you feel about this?
2. Slimmed down the args these factories take, they no longer specify tokenizers and pretrained model names.I know these tokenizers are being used for some error validation but its pretty odd to specify them multiple times + preprocessing really does belong with the dataset.

Other thoughts:
1. I saw `assert transformers.AutoModelForMaskedLM.from_pretrained is not None, "from_pretrained should not be None"`, is this being used for anything other than making pyright happy?
2. With these changes, `bert-base` probably doesn't need multiple eval anymore. If that makes sense I can remove that from the yaml, thoughts?

* Addresses CO-386


moinnadeem (commented):
> re: sequence length warmup, Is this something we necessarily have to manually add or could it be inferred automagically from either the model or dataset?

`model_inputs` could be automatically inferred from the model!

> HuggingFaceModel doesn't do this. Will this break GLUE? is this general enough to torchmetrics that i should add it to the base HuggingFaceModel?
Yes!

> I removed some of the validations on input args, mainly the one that prevents users from specifying a config and pretrained=True. Hugging Face does allow this behavior, they just load all the weights that are applicable and throw a warning. How do you feel about this?
If a user specifies a config, then there will likely not be a pretrained model to pull -- I wanted to preempt this problem by raising an exception, do you see what I mean? Thoughts on keeping the validation?

> I know these tokenizers are being used for some error validation but its pretty odd to specify them multiple times + preprocessing really does belong with the dataset.
The trouble is that we need to make sure that the tokenizer's outputs are compatible with the model's inputs -- there is some level of codesign going on here. I agree that it is odd to use them multiple times, is it possible to centralize them somewhere?

> I saw assert transformers.AutoModelForMaskedLM.from_pretrained is not None, "from_pretrained should not be None", is this being used for anything other than making pyright happy?

Not that I can remember!

> With these changes, bert-base probably doesn't need multiple eval anymore. If that makes sense I can remove that from the yaml, thoughts?
Wait, why doesn't it need multiple evaluators anymore? I think I'm missing something.

ishanashastri:
> The trouble is that we need to make sure that the tokenizer's outputs are compatible with the model's inputs -- there is some level of codesign going on here. I agree that it is odd to use them multiple times, is it possible to centralize them somewhere?
Inferred from model type/pulled from tokenizer; future PR will handle adding proper tokenization entrypoints

moinnadeem (approved):
LGTM! Great work!

composer/models/bert/bert_hparams.py
```diff
@@ -5,130 +5,113 @@
...
 
 @dataclass
-class BERTForClassificationHparams(TransformerHparams):
-    """`YAHP <https://docs.mosaicml.com/projects/yahp/en/stable/README.html>`_ classification interface for
-    :class:`.BERTModel`.
+class BERTHparams(ModelHparams):
+    """`YAHP <https://docs.mosaicml.com/projects/yahp/en/stable/README.html>`_ interface for :class:`.BERTModel`.
 
     Args:
-        pretrained_model_name (str): Pretrained model name to pull from Hugging Face Model Hub.
```
moinnadeem, line 27: Wait, one quick question: why did we get rid of this? @ishanashastri @A-Jacobson 

How would we expect the user to change the underlying BERT model on the HF hub without another `pretrained_model_name`?

composer/models/gpt2/gpt2_hparams.py
```diff
@@ -3,63 +3,62 @@
...
+        if not self.use_pretrained and self.model_config == {}:
+            raise Exception('One of use_pretrained or model_config needed.')
+
+        if self.use_pretrained and self.model_config:
+            raise Exception('A model cannot load pretrained weights from configuration.')
 
         # setup the tokenizer in the hparams interface
-        if self.tokenizer_name is not None:
+        if self.tokenizer_name:
             tokenizer = transformers.GPT2Tokenizer.from_pretrained(self.tokenizer_name)
```
A-Jacobson, line 55: can we move this logic to the factory functions? I'd like to keep interface parity between them and these hparams classes.

A-Jacobson: same with the bert hparams. I think it's a better design to have all the logic for creating a model be in one place (factory function). and for the non-yaml and yaml args to be as closely aligned as possible.

composer/models/bert/bert_hparams.py
A-Jacobson: I removed these because I didn't see it being changed in any of the yaml configs and thought you were including it to enable having the same hparam class for multiple models. of course, I wasn't sure which is why you're tagged to review =). If changing the underlying bert model is something we want to enable I see two paths for that. either we make multiple factory functions `create_bert_base` `create_bert_large` or we just add the `pretrained_model_name` back in. no problem with either.

ishanashastri: yup, moving the `pretrained_model_name` arg back seems to make most sense without having to change the existing yamls

PR #1135, opened 2022-06-07
Composer ResNet50 recipes

growlix:
We need an artifact for our recipes that makes them as easy to run/replicate as possible.


composer/yamls/recipes/resnet50_medium.yaml
```diff
@@ -0,0 +1,94 @@
...
+optimizer:
+  decoupled_sgdw:
+    dampening: 0.0
+    lr: 2.048
+    momentum: 0.875
+    nesterov: false
+    weight_decay: 0.0005
+precision: AMP
+max_duration: 90ep
+scale_schedule_ratio: # Fraction of 90 epochs to train for. 0.75-2.6 for medium recipe
```
eracah, line 63: wait so why don't we set a value here? do we want this to fail by default unless a user explicitly edits the yaml? we want these to work out of the box right?

PR #1136, opened 2022-06-07
Adding InMemoryLogger to the Getting Started Notebook

ishanashastri:
Stores training data in InMemoryLogger instances and plots the results to compare speed updates

### DIFF
--- a/examples/getting_started.ipynb
+++ b/examples/getting_started.ipynb
@@ -59,7 +59,11 @@
     "import torch\n",
     "\n",
     "import composer\n",
+    "import matplotlib.pyplot as plt\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


ravi-mosaicml (approved):
LGTM, nice job! Though I am a bit concerned that the composer accuracy is _lower_ than baseline in the plot that this notebook generates. That's a separate issue which shouldn't hold up this PR, but we should investigate why or figure out how to fix it (do we need more epochs?)

ishanashastri:
I think the reason for only doing 3 epochs on the baseline run is to demonstrate that 3 epochs running without Composer takes approximately the same amount of time as 4 epochs running with Composer. I can still modify the baseline if demonstrating that speedup effect is not as important, though.

PR #1138, opened 2022-06-07
Notebook cleanup

ravi-mosaicml:
Fix some of the notebook tests that were failing as part of the daily test runs

1. Fix notebooks to run on both cpu and gpu
2. Fix monkeypatching of the trainer when running tests to do only 1 epoch, and 1 batch per epoch
5. Fix any errors in the tests
3. Stripping all metadata from the notebooks
4. Using `%pip` rather than `!pip` for better portability

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -95,6 +95,7 @@ repos:
         types:
           - "jupyter"
         args:
+          # Strip all the metadata that vscode or colab may add to a notebook
           - --extra-keys
           - >
             metadata.colab metadata.interpreter metadata.accelerator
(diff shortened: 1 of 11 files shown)


bandish-shah (approved):
LGTM.

For the metadata being stripped, is it possible that this would get added in again if someone edits and modifies the notebook?  In which case should we put out some guidelines on committing notebooks to the repo or possibly a pre-commit hook that automates the stripping?

ravi-mosaicml:
> LGTM.
> 
> For the metadata being stripped, is it possible that this would get added in again if someone edits and modifies the notebook? In which case should we put out some guidelines on committing notebooks to the repo or possibly a pre-commit hook that automates the stripping?

Yes, this is being done via pre-commit -- see https://github.com/mosaicml/composer/blob/2fc7e2d2f146721656269e8443f6f3a0b6874eca/.pre-commit-config.yaml#L101

PR #1140, opened 2022-06-08
Global shuffle, 1x egress

knighton:


composer/datasets/streaming/dataset.py
```diff
@@ -1,328 +1,467 @@
 # Copyright 2022 MosaicML Composer authors
 # SPDX-License-Identifier: Apache-2.0
 
-"""The :class:`StreamingDataset` class, used for building streaming iterable datasets.
```
ravi-mosaicml, line 4: Can you leave the class level docstring?

ravi-mosaicml (commented):
I didn't give this PR a close review, but at a high level, I am wondering if we can avoid setting up our own sockets. 

Instead, could we use `torch.dist` for all main-process to main-process communication, and then use `multiprocessing` for all main-process-to-dataloader-worker communication? Not a huge fan of setting up our own sockets when pytorch already provides (optimized) primitives to help with this.

ravi-mosaicml:
Decided against this -- won't be sharing shards across nodes.

PR #1141, opened 2022-06-08
SFTP handler for Streaming Datasets

abhi-mosaic:
Add simple SFTP support via the `StreamingDataset`'s download handlers.

### DIFF
--- a/composer/datasets/streaming/download.py
+++ b/composer/datasets/streaming/download.py
@@ -24,10 +24,8 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
         import boto3
         from botocore.config import Config
     except ImportError as e:
-        raise ImportError(
-            textwrap.dedent("""\
-            Composer was installed without streaming support. To use streaming with Composer, run: `pip install mosaicml
-            [streaming]` if using pip or `conda install -c conda-forge monai` if using Anaconda""")) from e
+        raise MissingConditionalImportError(extra_deps_group="streaming", conda_package="boto3") from e
+
 
     obj = urlparse(remote)
     if obj.scheme != 's3':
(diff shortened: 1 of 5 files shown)


composer/datasets/streaming/download.py
```diff
@@ -24,19 +24,48 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
...
+        raise ValueError(f"Expected obj.scheme to be 'sftp', got {obj.scheme} for remote={remote}")
+
+    local_tmp = local + ".tmp"
+    if os.path.exists(local_tmp):
+        os.remove(local_tmp)
+    
+    private_key_path = os.environ["STREAMING_DATASET_SSH_FILE"]
+    username = os.environ["STREAMING_DATASET_SSH_USERNAME"]
+    pkey = RSAKey.from_private_key_file(private_key_path)
+    transport = Transport(sock=(obj.netloc, 22))
```
ravi-mosaicml, line 61: do they have any context manager you can  use so `.close` does not have to be called manually?
If not, is it possible to use `with contextlib.closing(Transport(...)) as transport:` and same thing for the `connection` below?

composer/datasets/streaming/download.py
```diff
@@ -96,5 +126,6 @@ def download_or_wait(remote: str, local: str, wait: bool = False, max_retries: i
             error_msgs.append(e)
             last_error = e
             continue
-    if last_error:
-        raise RuntimeError(f"Failed to download {remote} -> {local}. Got errors:\n{error_msgs}") from last_error
+    if len(error_msgs):
+        print (f"Failed to download {remote} -> {local}. Got errors:\n{error_msgs}")
```
ravi-mosaicml, line 130: Can you leave this line as it was? Would prefer to re-raise the exception. This looks like debugging code?

ravi-mosaicml (approved):
LGTM; see comments

composer/datasets/streaming/download.py
abhi-mosaic: We've got an issue with child threads not killing the parent dataloader thread... so I added the `os._exit(1)` line which should work, but then I'm not sure if I can raise an Error right before calling that... do you know of a better way?

abhi-mosaic:
Just a quick comment -- want to wait on merging this until I can fix some sporadic SFTP slow downloads, and correct the timeout behavior.

What I saw on some ImageNet runs: a single SFTP download thread would run 100x slower than the rest, but neither it nor any sibling threads were successfully exiting past the `timeout` condition, so the whole training run was hanging.

ravi-mosaicml (commented):
(already reviewed; leaving a comment to dismiss it)

composer/datasets/streaming/download.py
abhi-mosaic: So the trouble I've been seeing is that exceptions raised in the child download threads are not killing the parent thread (a dataloader CPU worker). I think this is maybe expected, as `daemon=True` means that parent exit will cause child exit, but does not guarantee vice-versa.

My multithreading knowledge is very limited btw so it may be that there's a simple solution I'm not seeing.

abhi-mosaic: Here's where the child thread is created: https://github.com/mosaicml/composer/blob/dfc5a436df9fe2d045ef7ac52e314f728655173a/composer/datasets/streaming/dataset.py#L329

ravi-mosaicml: Ah yes. So, the main thread should periodically check to see if the child thread died. If it did, then it should exit. See an example here:

https://github.com/mosaicml/composer/blob/33acb2525c811fcb5cfdaba5c35d0f0c421b16ea/composer/loggers/object_store_logger.py#L267

abhi-mosaic: Reverted error handling to what it was doing before, and going to move this `_check_workers` refactor to a follow-up PR to avoid blocking.

PR #1142, opened 2022-06-08
Add devcontainer config

ravi-mosaicml:
* Add a devcontainer config for development in a container when using vscode. Makes it easier to develop on the docker image, which has system level dependencies installed.
* Using the `mosaicml` user to avoid permission issues that would occur when developing as root. However, `sudo` is still available if a user needs to switch into root inside the container.
* Using the `mosaicml/pytorch_vision:latest` image, so it will work on gpu machines and so it can run all tests (including vision ones).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+    "image": "mosaicml/pytorch_vision:latest",
+    "remoteUser": "mosaicml",
+    "postCreateCommand": "pip install -e .[all] && pre-commit install"
+}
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "image": "mosaicml/pytorch_vision:latest",
+    "remoteUser": "mosaicml",
```
bandish-shah, line 3: Is it possible to use `root` as is default in the image and just use a virtual environment?  I think we've gone back and forth on this enough, and it's better to keep it consistent across everything if possible.  I don't want developers to have to think about what they need to do differently if they want to run their code on the other images.

bandish-shah (changes requested):
See comments.

PR #1143, opened 2022-06-08
Sftp object store

anisehsani:
Object Store Implementation that uses SFTP as the backend.


composer/utils/iter_helpers.py
```diff
@@ -88,3 +89,37 @@ def iterate_with_pbar(iterator, progress_bar=None):
             yield x
             if pb is not None:
                 pb.update(len(x))
+
+
+class IteratorFileStream(io.RawIOBase):
```
ravi-mosaicml, line 94: I'm a bit confused; I thought this was already in the other PR?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        self.ssh_client = SSHClient()
+
+        # add AutoAddPolicy
+        if key_file_path is None:
+            self.ssh_client.load_system_host_keys()
+        else:
+            try:
+                self.ssh_client.load_system_host_keys(key_file_path)
+            except IOError:
+                warnings.warn(
```
ravi-mosaicml, line 32: Should this be a warning or `log.info`?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(self, host: str, port: int = 22, username: Optional[str] = None, key_file_path: Optional[str] = None):
+        self.ssh_client = SSHClient()
+
+        # add AutoAddPolicy
+        if key_file_path is None:
+            self.ssh_client.load_system_host_keys()
+        else:
+            try:
+                self.ssh_client.load_system_host_keys(key_file_path)
```
ravi-mosaicml, line 30: If `key_file_path` is specified and it doesn't exist, then can we leave the exception as-is? I think that should be whatever exception is raised.

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+                warnings.warn(
+                    "Host keys could not be read from {key_file_path}. Attempting to read keys from known hosts.")
+                self.ssh_client.load_system_host_keys()
+
+        # implement some retries here
+        self.ssh_client.connect(host, port, username)
+
+        self.sftp_client = self.ssh_client.open_sftp()
+
+    def close_client(self):
```
ravi-mosaicml, line 41: `.close` is the standard method for cleanup

```suggestion
    def close(self):
```

composer/utils/iter_helpers.py
anisehsani: Yeah - not sure why it's here, I merged from the other branch before it got merged into dev.

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Args:
+            file_path (str): Path to the object on disk.
+            object_name (str): Object name (i.e. where the object will be stored in the container.)
+        """
+        # add verification and retries/rename
+        tmp_path = object_name + f".{uuid.uuid4()}.tmp"
+
+        object_size = os.stat(file_path).st_size
+        attributes = (SFTPAttributes)(self.sftp_client.put(file_path, tmp_path))
```
ravi-mosaicml, line 55: Why does this need to be wrapped with `SFTPAttributes`?

ravi-mosaicml: Could we use `confirm=True` instead of manually verifying the file size?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        uploaded_bytes = attributes.st_size
+
+        if uploaded_bytes != object_size:
+            self.sftp_client.remove(tmp_path)
+            # Replace with custom exception later
```
ravi-mosaicml, line 60: (creating a GH comment to track this TODO)

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+            raise Exception("File sizes don't match uploading error")
+        else:
+            self.sftp_client.rename(tmp_path, object_name)
+
+    def upload_object_via_stream(self, obj: Union[bytes, Iterator[bytes]], object_name: str):
+        """Upload an object.
```
ravi-mosaicml, line 66: No need for the docstring,

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        self.ssh_client.connect(host, port, username)
+
+        self.sftp_client = self.ssh_client.open_sftp()
+
+    def close_client(self):
+        self.ssh_client.close()
+
+    def upload_object(self, file_path: str, object_name: str):
+        """Upload an object currently located on a disk.
+
```
ravi-mosaicml, line 46: No need for the docstring,

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+from paramiko import SFTPAttributes, SSHClient
+
+from composer.utils.iter_helpers import IteratorFileStream
+
+__all__ = ["SFTPObjectStore"]
+
+
+class SFTPObjectStore:
+    """Utility for uploading to and downloading to a server via SFTP."""
```
ravi-mosaicml, line 20: Can the `__init__` arguments be documented?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            try:
+                self.ssh_client.load_system_host_keys(key_file_path)
+            except IOError:
+                warnings.warn(
+                    "Host keys could not be read from {key_file_path}. Attempting to read keys from known hosts.")
+                self.ssh_client.load_system_host_keys()
+
+        # implement some retries here
+        self.ssh_client.connect(host, port, username)
```
ravi-mosaicml, line 37: Can we support all arguments through a `kwargs` dictionary?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+
+from composer.utils.iter_helpers import IteratorFileStream
+
+__all__ = ["SFTPObjectStore"]
+
+
+class SFTPObjectStore:
+    """Utility for uploading to and downloading to a server via SFTP."""
+
+    def __init__(self, host: str, port: int = 22, username: Optional[str] = None, key_file_path: Optional[str] = None):
```
ravi-mosaicml, line 22: Instead of manually enumerating some of the `connect` arguments, can we instead take a `connect_kwargs` dictionary, and pass through whatever the caller provides?

ravi-mosaicml, line 22: Is it possible to rename `key_file_path` to `host_key_filename`? I thought that this was the ssh private key (which it isn't).

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        self.ssh_client = SSHClient()
+
+        # add AutoAddPolicy
+        if key_file_path is None:
```
ravi-mosaicml, line 26: Do we need to do this check? Or can we do `ssh_client.load_system_host_keys(host_key_filename)`?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+            object_name (str): Object name (i.e. where the object will be stored in the container.)
+        """
+        tmp_path = object_name + f".{uuid.uuid4()}.tmp"
+
+        if isinstance(obj, bytes):
+            obj = iter(i.to_bytes(1, sys.byteorder) for i in obj)
+        file_obj = io.BufferedReader(IteratorFileStream(obj))
+        try:
+            self.sftp_client.putfo(fl=file_obj, remotepath=tmp_path, confirm=True)
+        except:
```
ravi-mosaicml, line 79: Can the except be more specific?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+            self.sftp_client.remove(tmp_path)
+        else:
+            self.sftp_client.rename(tmp_path, object_name)
+
+    def download_object(self, object_name: str, destination_path: str):
+        """Download an object to the specified destination path.
```
ravi-mosaicml, line 85: No need for the docstring.

composer/utils/sftp_object_store.py
ravi-mosaicml: (Also will need to wrap transient exceptions)

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        uploaded_bytes = attributes.st_size
+
+        if uploaded_bytes != object_size:
+            self.sftp_client.remove(tmp_path)
+            # Replace with custom exception later
+            raise Exception("File sizes don't match uploading error")
+        else:
+            self.sftp_client.rename(tmp_path, object_name)
+
+    def upload_object_via_stream(self, obj: Union[bytes, Iterator[bytes]], object_name: str):
```
ravi-mosaicml, line 65: If `object_name` contains a subfolder path, do we need to manually create the nested directories if the folders do not yet exist?

composer/utils/sftp_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # implement some retries here
+        self.ssh_client.connect(host, port, username)
+
+        self.sftp_client = self.ssh_client.open_sftp()
+
+    def close_client(self):
+        self.ssh_client.close()
+
+    def upload_object(self, file_path: str, object_name: str):
```
ravi-mosaicml, line 44: (Same comment below re: if the object name contains nested subfolders, do we need to create the folders if they do not exist?)

composer/utils/sftp_object_store.py
ravi-mosaicml, line 22: Can we add a "cwd" argument? Upon connection, we then `chdir` to this directory?

tests/utils/test_iter_helpers.py
```diff
@@ -30,3 +32,10 @@ def test_dict_to_tuple(x):
 @pytest.mark.parametrize("x", [torch.arange(3), np.arange(3)])
 def test_obj_to_tuple(x):
     assert ensure_tuple(x) == (x,)
+
+
+def test_iter_to_stream():
```
ravi-mosaicml, line 37: Same comment as before re: the previous PR.

ravi-mosaicml (commented):
Looking good! Seems pretty close; just left a bunch of thoughts.

composer/utils/object_store/object_store_hparams.py
```diff
@@ -210,7 +214,37 @@ def get_kwargs(self) -> Dict[str, Any]:
...
+
+    The :class:`.SFTPObjectStore` uses :mod:`paramiko` to upload and download files from SSH servers
+
+    .. note::
+
+        To follow best security practices, credentials shouldn't be specified as part of the hyperparameters.
+        Instead, please ensure that credentials are in the environment, which will be read automatically.
+
+    Args:
+        host (str): See :class:`.SFTPObjectStore`.
```
ravi-mosaicml, line 229: These args are incomplete....

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+__all__ = ['SFTPObjectStore']
+
+
+class SFTPObjectStore(ObjectStore):
+    """Utility for uploading to and downloading to a server via SFTP.
+
+    Args:
+        host (str): The server to connect to
+        port (int): The server port to connect to
```
ravi-mosaicml, line 20: Can this default to port 22?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+        username (str): The username to authenticate
```
ravi-mosaicml, line 21: Can this be optional?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+        key_file_path (str): The filename of private key
```
ravi-mosaicml, line 22: Can this be optional?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+        cwd (Optional[str]): The directory to navigate to upon creating the SSH connection.
```
ravi-mosaicml, line 23:
```suggestion
        cwd (str, optional): The directory to navigate to upon creating the SSH connection.
```

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+        try:
+            from paramiko import SSHClient
+        except ImportError as e:
+            raise MissingConditionalImportError(extra_deps_group='streaming', conda_package='paramiko') from e
+        self.ssh_client = SSHClient()
+
+        try:
+            self.ssh_client.load_system_host_keys(key_file_path)
+        except IOError:
```
ravi-mosaicml, line 41: No need to re-raise the exception here; it's ok to let the original exception propegate through. Would rather be specific that raise a general exception.

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+            raise Exception('Host keys could not be read from {key_file_path}.')
+
+        self.ssh_client.connect(host, port, username)
+        if cwd is not None:
+            self.ssh_client.exec_command(f'mkdir -p {cwd}')
```
ravi-mosaicml, line 46: Is it possible to use the `mkdir` and `chdir` commands? https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.mkdir and https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.chdir

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+            self.ssh_client.exec_command(f'cd {cwd}')
+
+        self.sftp_client = self.ssh_client.open_sftp()
+
+    def close(self):
+        self.ssh_client.close()
+
+    def get_uri(self, object_name: str) -> str:
+        return f'sftp://{self.host}:{self.port}/{object_name}'
```
ravi-mosaicml, line 55:
```suggestion
        return f'sftp://{self.host}:{self.port}/{cwd.rstrip("/") + '/' if self.cwd is not None else ''}{object_name}'
```

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def get_object_size(self, object_name: str) -> int:
+        return self.sftp_client.stat(object_name).st_size
+
+    def upload_object(self, file_path: str, object_name: str):
+        tmp_path = object_name + f'.{uuid.uuid4()}.tmp'
+
+        dirname = os.path.dirname(object_name)
+        self.ssh_client.exec_command(f'mkdir -p {dirname}')
```
ravi-mosaicml, line 64: Is it possible to use `mkdir`? https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.mkdir

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+        try:
+            self.sftp_client.put(file_path, tmp_path, confirm=True)
+        except IOError:
+            raise ObjectStoreTransientError
```
ravi-mosaicml, line 69: Are all IO errors always transient errors? And are there other types of non-io transient exceptions that could be raised? In addition, can we re-raise the exception?

```suggestion
            raise ObjectStoreTransientError from e
```

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def upload_object(self, file_path: str, object_name: str):
+        tmp_path = object_name + f'.{uuid.uuid4()}.tmp'
+
+        dirname = os.path.dirname(object_name)
+        self.ssh_client.exec_command(f'mkdir -p {dirname}')
+
+        try:
+            self.sftp_client.put(file_path, tmp_path, confirm=True)
+        except IOError:
```
ravi-mosaicml, line 68:
```suggestion
        except IOError as e:
```

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+            raise ObjectStoreTransientError
+        else:
+            self.sftp_client.rename(tmp_path, object_name)
+
+    def download_object(self, object_name: str, destination_path: str):
```
ravi-mosaicml, line 73: The signature should match that of the base class.

```suggestion
    def download_object(
        self,
        object_name: str,
        filename: Union[str, pathlib.Path],
        overwrite: bool = False,
        callback: Optional[Callable[[int, int], None]] = None,
    ) -> None:
```

Can the ``callback`` parameter be implemented?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+    def close(self):
+        self.ssh_client.close()
+
+    def get_uri(self, object_name: str) -> str:
+        return f'sftp://{self.host}:{self.port}/{object_name}'
+
+    def get_object_size(self, object_name: str) -> int:
+        return self.sftp_client.stat(object_name).st_size
+
+    def upload_object(self, file_path: str, object_name: str):
```
ravi-mosaicml, line 60: The signature should match that of the base class:

```suggestion
    def upload_object(
        self,
        object_name: str,
        filename: Union[str, pathlib.Path],
        callback: Optional[Callable[[int, int], None]] = None,
    ) -> None:
```

Can the ``callback`` parameter be implemented?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+        os.makedirs(dirname, exist_ok=True)
+
+        tmp_path = destination_path + f'.{uuid.uuid4()}.tmp'
+
+        try:
+            self.sftp_client.get(remotepath=object_name, localpath=tmp_path)
+        except IOError:
+            raise ObjectStoreTransientError
+        else:
+            os.replace(tmp_path, destination_path)
```
ravi-mosaicml, line 84: See the S3ObjectStore for how rename / replace is used depending on `overwrite`. In addition, it'd be good to check before downloading whether the localpath already exists.

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Args:
+        host (str): The server to connect to
+        port (int): The server port to connect to
+        username (str): The username to authenticate
+        key_file_path (str): The filename of private key
+        cwd (Optional[str]): The directory to navigate to upon creating the SSH connection.
+    """
+
+    def __init__(self, host: str, port: int, username: str, key_file_path: str, cwd: Optional[str] = None):
```
ravi-mosaicml, line 26: Can you add in a `kwargs` argument so the user can pass through additional options into paramiko?

tests/utils/object_store/object_store_settings.py
```diff
@@ -26,11 +27,20 @@
...
+    del paramiko
+except ImportError:
+    _PARAMIKO_AVAILABLE = False
+
 _object_store_marks = {
     LibcloudObjectStore: [pytest.mark.skipif(not _LIBCLOUD_AVAILABLE, reason='Missing dependency')],
     LibcloudObjectStoreHparams: [pytest.mark.skipif(not _LIBCLOUD_AVAILABLE, reason='Missing dependency')],
     S3ObjectStore: [pytest.mark.skipif(not _BOTO3_AVAILABLE, reason='Missing dependency')],
     S3ObjectStoreHparams: [pytest.mark.skipif(not _BOTO3_AVAILABLE, reason='Missing dependency')],
+    SFTPObjectStore: [pytest.mark.skipif(not _PARAMIKO_AVAILABLE, reason='Missing dependency')],
```
ravi-mosaicml, line 42: I think the tests will require some additional setup to be able to mock paramiko.

ravi-mosaicml (commented):
Looking nice! Mostly minor feedback. Main things are to implement the `callback` arg and configure the tests. Let's get this in tomorrow :)

anisehsani:
@ravi-mosaicml Thanks for the review. About the tests: Had written a bunch of tests today with the `mock-ssh-server` package. Unfortunately, I got a unsatisfiable dependency error so I removed them. Without a compatible mock server, unsure what the tests should actually do since I only really want to test the Paramiko implemented functions.

ravi-mosaicml:
> @ravi-mosaicml Thanks for the review. About the tests: Had written a bunch of tests today with the `mock-ssh-server` package. Unfortunately, I got a unsatisfiable dependency error so I removed them. Without a compatible mock server, unsure what the tests should actually do since I only really want to test the Paramiko implemented functions.

(Resolved offline, but I think this should be installable via pip):

```
(composer) ravi@dolphin:~/ravi-composer$ pip install mock-ssh-server
Collecting mock-ssh-server
  Downloading mock_ssh_server-0.9.1-py3-none-any.whl (14 kB)
Collecting paramiko
  Downloading paramiko-2.11.0-py2.py3-none-any.whl (212 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 212.9/212.9 kB 3.8 MB/s eta 0:00:00
Requirement already satisfied: cryptography>=2.5 in /home/ravi/venvs/composer/lib/python3.10/site-packages (from paramiko->mock-ssh-server) (37.0.2)
Collecting pynacl>=1.0.1
  Downloading PyNaCl-1.5.0-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl (856 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 856.7/856.7 kB 15.5 MB/s eta 0:00:00
Collecting bcrypt>=3.1.3
  Downloading bcrypt-3.2.2-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl (62 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 62.3/62.3 kB 4.7 MB/s eta 0:00:00
Requirement already satisfied: six in /home/ravi/venvs/composer/lib/python3.10/site-packages (from paramiko->mock-ssh-server) (1.16.0)
Requirement already satisfied: cffi>=1.1 in /home/ravi/venvs/composer/lib/python3.10/site-packages (from bcrypt>=3.1.3->paramiko->mock-ssh-server) (1.15.0)
Requirement already satisfied: pycparser in /home/ravi/venvs/composer/lib/python3.10/site-packages (from cffi>=1.1->bcrypt>=3.1.3->paramiko->mock-ssh-server) (2.21)
Installing collected packages: pynacl, bcrypt, paramiko, mock-ssh-server
Successfully installed bcrypt-3.2.2 mock-ssh-server-0.9.1 paramiko-2.11.0 pynacl-1.5.0
```

composer/utils/object_store/sftp_object_store.py
anisehsani: When you don't provide a username to Paramiko, it uses the current active username. I don't think this is correct so not sure about making it optional

composer/utils/object_store/sftp_object_store.py
anisehsani: Looking though the Paramiko functions I use - none have extra args to pass.

composer/utils/object_store/object_store_hparams.py
```diff
@@ -210,7 +214,40 @@ def get_kwargs(self) -> Dict[str, Any]:
...
+        port (int, optional): See :class:`.SFTPObjectStore`.
+        username (str, optional): See :class:`.SFTPObjectStore`.
+        key_file_path (str, optional): See :class:`.SFTPObjectStore`.
+        cwd (str, optional): See :class:`.SFTPObjectStore`.
+    """
+
+    host: str = hp.auto(SFTPObjectStore, 'host')
+    port: int = hp.optional(('port'), default=22)
```
ravi-mosaicml, line 237: For all these arguments, is it possible to use `hp.auto`? https://docs.mosaicml.com/projects/yahp/en/latest/api_ref/field.html#yahp.field.auto

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,110 @@
...
+                 key_file_path: Optional[str] = None,
+                 cwd: Optional[str] = None):
+        self.host = host
+        self.port = port
+        self.username = username
+        self.key_file_path = key_file_path
+        self.cwd = cwd
+
+        if '@' in host:
+            self.username, self.host, *_ = self.host.split('@')
```
ravi-mosaicml, line 42: Is it possible to use `urllib.parse.urlparse`? https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

tests/utils/object_store/test_object_store.py
```diff
@@ -81,6 +118,8 @@ def test_get_uri(self, object_store: ObjectStore):
             assert uri == 's3://my-bucket/tmpfile_object_name'
         elif isinstance(object_store, LibcloudObjectStore):
             assert uri == 'local://./tmpfile_object_name'
+        elif isinstance(object_store, SFTPObjectStore):
+            pytest.skip("SFTP object store testing uses a mock server without a uri")
```
ravi-mosaicml, line 122: This should still return something....may need to use a dummy uri in the test.

tests/utils/object_store/test_object_store.py
```diff
@@ -55,11 +90,13 @@ def object_store(request, monkeypatch: pytest.MonkeyPatch,
         os.makedirs(remote_dir)
         object_store_kwargs[request.param]['provider_kwargs']['key'] = remote_dir
         yield request.param(**object_store_kwargs[request.param])
+    elif request.param is SFTPObjectStore:
+        yield TestSFTPObjectStore("unused_test_hostname")
```
ravi-mosaicml, line 94: Instead of mocking out the entire init with `TestSFTPObjectStore`, would it be possible to monkeypatch the Server class to return the mock server class, so we can still test the original init?

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,115 @@
...
+    def close(self):
+        self.ssh_client.close()
+
+    def get_uri(self, object_name: str) -> str:
+        return f'sftp://{self.host}:{self.port}/{object_name}'
+
+    def get_object_size(self, object_name: str) -> int:
+        st_size = self.sftp_client.stat(object_name).st_size
+        if st_size is None:
+            raise ObjectStoreTransientError
```
ravi-mosaicml, line 75: 1. Do we need special handling for a `FileNotFound` error to re-wrap the exception? Or does Paramiko use the Python `FileNotFound exception?
2. If `st_size is None`: raise a RuntimeError?

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: Can we catch `SSHException` when it contains `Server connection dropped` (https://github.com/paramiko/paramiko/blob/609e01d550f62253c56c34fc5a49d01f1d49e6e2/paramiko/sftp_client.py#L854) for now? If you hit other transient errors when testing feel free to add those in too.

ravi-mosaicml: (This comment also applies to `get_file_size` and `download_object`)

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,115 @@
...
+        os.makedirs(dirname, exist_ok=True)
+
+        if os.path.exists(filename) and not overwrite:
+            raise FileExistsError(f'The file at {filename} already exists')
+
+        tmp_path = str(filename) + f'.{uuid.uuid4()}.tmp'
+
+        try:
+            self.sftp_client.get(remotepath=object_name, localpath=tmp_path, callback=callback)
+        except IOError:
```
ravi-mosaicml, line 104:
```suggestion
        except:
```

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,115 @@
...
+            # Make a best effort attempt to clean up the temporary file
+            try:
+                os.remove(tmp_path)
+            except OSError:
+                pass
+            raise ObjectStoreTransientError
```
ravi-mosaicml, line 110:
```suggestion
            raise
```

tests/utils/object_store/test_object_store.py
```diff
@@ -55,11 +88,13 @@ def object_store(request, monkeypatch: pytest.MonkeyPatch,
         os.makedirs(remote_dir)
         object_store_kwargs[request.param]['provider_kwargs']['key'] = remote_dir
         yield request.param(**object_store_kwargs[request.param])
+    elif request.param is SFTPObjectStore:
+        yield MockSFTPObjectStore('test_hostname', port=24, username='test_user')
```
ravi-mosaicml, line 92: Instead of creating the `MockSFTObjectStore`, can we use `monkepatch` and use the original object store class?

Something like:

```
elif request.param is SFTPObjectStore:
    # setup the server
    server = mockssh.Server(users={})
    private_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
    pem = private_key.private_bytes(encoding=serialization.Encoding.PEM,
                                    format=serialization.PrivateFormat.TraditionalOpenSSL,
                                    encryption_algorithm=serialization.NoEncryption())
    with tempfile.TemporaryDirectory() as tmpdir:
            tmppath = os.path.join(tmpdir, 'test_rsa_key')
            with open(tmppath, 'wb') as private_key_file:
                private_key_file.write(pem)
            server.add_user(uid=self.username, private_key_path=tmppath)
            with server:
                client = server.client(username)
                # Trick the SFTPObjectStore to create a client attached to the mock server
                monekpatch.setattr(composer.utils.object_store.sftp_object_store, "SSHClient", lambda: client)
                 yield SFTPObjectStore(**object_store_kwargs[request.param])

File: tests/utils/object_store/test_object_store.py [hunk 42]
```diff
@@ -30,6 +35,34 @@ def assert_all_data_transferred(self):
         assert self.total_num_bytes == self.transferred_bytes
 
 
+class MockSFTPObjectStore(SFTPObjectStore):
```
ravi-mosaicml, line 38: Is it possible to remove this class? See below for an alternative.

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: (Re the first comment, the test cases check for file not found errors, so this is a moot point if the test cases are passing)

tests/utils/object_store/test_object_store.py
ravi-mosaicml: It might be necessary to refactor this logic out into a shared helper, so it can also be used for the test_object_store_hparams_is_constructable test below.

anisehsani: I'm still going to monkeypatch the whole function because I also don't want the key_file_authentication commands to run. (They still would if I only patch the client). However, I will get rid of the class.

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,134 @@
...
+                 username: Optional[str] = None,
+                 key_file_path: Optional[str] = None,
+                 cwd: Optional[str] = None):
+        self.host = host
+        self.port = port
+        self.username = username
+        self.key_file_path = key_file_path
+        self.cwd = cwd
+
+        if host.startswith('sftp://'):
```
ravi-mosaicml, line 44: nit: Can we always attempt to `urllib.parse.urlparse(host)`, and then check to see if `scheme` is empty (in which case it's just the host), or `sftp` (in which case it's the URI)

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,134 @@
...
+            url = urllib.parse.urlparse(host)
+            if url.username:
```
ravi-mosaicml, line 46: For each of these fields, can you raise a `ValueError` if it is also specified via the init arguments?

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: (Bumping this)

composer/utils/object_store/sftp_object_store.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(
+        self,
+        host: str,
+        port: int = 22,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        known_hosts_filename: Optional[Union[pathlib.Path, str]] = None,
+        cwd: Optional[str] = None,
+        connect_kwargs: Optional[Dict[str, Any]] = None,
```
abhi-mosaic, line 64: The private key will need to be passed via `connect_kwargs` because we cannot rely on only one SSH key (`id_rsa`) being available. Would prefer a top level argument like `private_key_path`. Then you'll need:

```python
pkey = RSAKey.from_private_key_file(private_key_path)
if private_key_path:
            _set_kwarg(pkey, connect_kwargs, arg_name='private_key_path', kwarg_name='pkey')
```

abhi-mosaic (approved):
LGTM, trusting the tests and looking forward to E2E next weeks on 0.8.0.

Approving for expediency but I think there is still a blocker of providing a `private_key_path` argument.

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: Agreed; just added that in!

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: Done; removed the wrap here.

composer/utils/object_store/sftp_object_store.py
ravi-mosaicml: Resolved

composer/utils/sftp_object_store.py
ravi-mosaicml: Added in `connect_kwargs`

composer/utils/sftp_object_store.py
ravi-mosaicml: Added in `connect_kwargs`. Keeping the top-level arguments high level for visibility and autoyahp compatibility.

ravi-mosaicml: Done

composer/utils/sftp_object_store.py
ravi-mosaicml: Done.

composer/utils/sftp_object_store.py
ravi-mosaicml: Fixed

PR #1144, opened 2022-06-08
Abstract Retrying Logic and the Libcloud Object Store

ravi-mosaicml:
This PR abstracts out the retrying logic into its own exception class and helpers, and it abstracts away the libcloud object store throughout the trainer and loggers.

- Added the `ObjectStoreTransientError` class
- Made the `ObjectStoreLogger`, `Trainer`, and other classes support the abstract `ObjectStore` class rather than just the `LibcloudObjectStore`
- Abstracted out the `ObjectStoreHparams`; added a registry
- Added a `retry` helper to automatically retry functions with exponential backoff and/or jitter
- Removed the `GetFileNotFound` exception; replaced with the builtin `FileNotFound` exception
- Added retries to validate credentials
- Moved all object store helpers into `composer.utils.object_store` for better organization
- Make `libcloud` an optional dependency.

### DIFF
--- a/composer/loggers/object_store_logger.py
+++ b/composer/loggers/object_store_logger.py
@@ -20,8 +20,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 from composer.core.state import State
 from composer.loggers.logger import Logger, LogLevel
 from composer.loggers.logger_destination import LoggerDestination
-from composer.utils import ObjectStore, format_name_with_dist, get_file, retry
-from composer.utils.object_store.object_store import ObjectStoreTransientError
+from composer.utils import ObjectStore, ObjectStoreTransientError, dist, format_name_with_dist, get_file, retry
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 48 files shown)


composer/loggers/object_store_logger.py
```diff
@@ -42,21 +33,32 @@ def _always_log(state: State, log_level: LogLevel, artifact_name: str):
...
+
         object_store_logger = ObjectStoreLogger(
-            provider='s3',
-            container='my-bucket',
-            provider_kwargs={
-                'key': 'AKIA...',
-                'secret': '*********',
-                'region': 'ap-northeast-1',
+            object_store_cls=LibcloudObjectStore,
+            object_store_kwargs={
```
abhi-mosaic, line 54: Could you provide some context as to why we have to accept the class and kwargs, rather than an instance of the class?

abhi-mosaic (approved):
I glanced through the changes and I'm mostly trusting the tests for correctness (and E2E training runs to come?). Approving so other PRs stacked on top of this are unblocked.

I noticed that `ObjectStoreLogger` takes kwargs rather than an instance of `ObjectStore`, and I was just wondering what the logic behind that.

composer/loggers/object_store_logger.py
ravi-mosaicml: This is to support multiple upload workers. A different instance object store is constructed within each worker. It is not safe to share an `ObjectStore` instance across thread boundaries, and it would be impossible to share object stores across subprocesses. I'll add a comment to clarify what's going on.

PR #1145, opened 2022-06-09
Refactor the `ObjectStore` base class

ravi-mosaicml:
* Removed `upload_obj_as_stream` and `download_obj_as_stream` as nothing uses them that can't be replaced with files, and these signatures would be difficult to support with boto3 and paramiko (since they use file objects instead of iterators)
* Renamed arguments to `object_name` and `filename`, which the `object_name` always coming before the `filename` in API signatures. This ordering matches the boto3 API.
* Added in a `callback` argument into `upload_object` and `download_object`. The `callback` takes (`num_bytes_transferred`, `total_num_bytes`). It is called periodically. Both `boto3` and `paramiko` support a `callback` argument, and this is useful for showing things like progress bars
* Implemented the `callback` argument for libcloud, with test cases. As libcloud does not manually support `callback`, the implementation wraps the iterators to periodically call a `callback` function.
* Removed the `chunk_size` argument from `download_obj` and `upload_obj`. Instead, individual object stores should take `chunk_size` as an init argument. Making this change as both boto3 and paramiko use a client-wide chunk size argument that cannot be easily changed for individual uploads and downloads. Refactored libcloud to take the `chunk_size` on init.
* Implement `overwrite` for `get_file_artifact` to be consistent with the object store `download_object`
* Renamed `assert_yaml_loads` to `construct_from_yaml` in the tests, as I needed the constructed hparams dataclass in the object store tests.


abhi-mosaic (approved):
LGTM for now, looking forward to seeing E2E usage with YAMLs for training runs.

PR #1151, opened 2022-06-10
Require the `run_name` when `autoresume=True`

ravi-mosaicml:
This PR always requires the `run_name` to be specified when `autoresume = True`. This is because there is a nasty circular dependency when trying to automatically load the run name from the checkpoint with `autoresume=True`:

* Event.INIT needs the run name to be set.
* The run name could be in a cloud checkpoint, which would be fetched from the loggers.
* The Loggers are not fully initialized until Event.INIT runs

While it would be possible to autogenerate the run name AFTER Event.INIT if using autoresume, this would likely not help much in practice, as checkpoints would probably be stored in the cloud, where the run_name is being used. So, this PR requires that the `run_name` is always specified when `autoresume=True`.

Coming along for the ride: Some typing fixes


composer/utils/batch_helpers.py
```diff
@@ -53,8 +53,8 @@ def batch_get(batch: Any, key: Union[str, int, Callable, Any]):
             return itemgetter(*key)(batch)
         except TypeError:
             try:
-                return attrgetter(key)(batch)
-            except:
+                return attrgetter(cast(str, key))(batch)
+            except TypeError:
```
mvpatel2000, line 57: I assume this is intentional but just thrown into PR

mvpatel2000 (approved):
Lgtm. So were good with leaving run_name in state? Fwiw I think its probably nice to have it there

ravi-mosaicml:
> Lgtm. So were good with leaving run_name in state? Fwiw I think its probably nice to have it there

Yes, I like storing it in the state -- I think it should be part of the checkpoint.

composer/utils/batch_helpers.py
ravi-mosaicml: Yeah, this one came along for the ride haha. VSCode started to complain about this.

PR #1153, opened 2022-06-13
Use global batch size for dataloader length calculation

bandish-shah:
* Currently MLPerf callback prints MLLOG using device batch size, thus train/val sample counts are 8x off
* Fixes issue raised in https://github.com/mlcommons/submissions_training_2.0/issues/57


hanlint (approved):
LGTM

PR #1156, opened 2022-06-13
Use global batch size only in FFCV dataloader length calculation

### LINKED ISSUE #1154: MLPerf Callback does not use global batch size in dataset length calculation
The MLPerf callback release as part of Composer 0.7.1 calculates FFCV dataset length for device batch size only:

```
if isinstance(dataloader, ffcv.loader.Loader):
    return (dataloader.batch_size, len(dataloader) * dataloader.batch_size)  # type: ignore
```

This results in the MLPerf log printing a train and val length that is 8x off from actual:

```
:::MLLOG {"namespace": "", "time_ms": 1653030529909, "event_type": "POINT_IN_TIME", "key": "train_samples", "value": 160000, "metadata": {"file": "/usr/local/lib/python3.9/dist-packages/composer/callbacks/mlperf.py", "lineno": 235}}
:::MLLOG {"namespace": "", "time_ms": 1653030529909, "event_type": "POINT_IN_TIME", "key": "
```

bandish-shah:
* We should use the `dis.get_world_size()` only for FFCV dataloader sample size calculation
* Fixes #1154


composer/callbacks/mlperf.py
```diff
@@ -250,7 +250,9 @@ def _get_dataloader_stats(self, dataloader: Iterable):
             import ffcv  # type: ignore
 
             if isinstance(dataloader, ffcv.loader.Loader):
-                return (dataloader.batch_size, len(dataloader) * dataloader.batch_size)  # type: ignore
+                return (
+                    dataloader.batch_size,  # type: ignore
```
ravi-mosaicml, line 254: Curious why do we need type ignores?

composer/callbacks/mlperf.py
```diff
@@ -250,7 +250,9 @@ def _get_dataloader_stats(self, dataloader: Iterable):
...
+                    len(dataloader) * dataloader.batch_size * dist.get_world_size())  # type: ignore
```
ravi-mosaicml, line 255: Can we use this logic for a pytorch dataloader too (instead of just using it for FFCV)?

ravi-mosaicml (approved):
I think this partially fixes the bug -- it's possible that a pytorch dataset is pre-sharded as well, in which case the count will still be off. Is it always possible to use `len(dataloader) * dataloader.batch_size * dist.get_world_size()` to determine the number of samples, regardless of the dataloader type?

Otherwise LGTM.

composer/callbacks/mlperf.py
bandish-shah: This was there before, but I think it can't find the `batch_size` attribute since `dataloader` is typed as an iterable.

composer/callbacks/mlperf.py
bandish-shah: My understanding is `leng(dataloader)` returns number of batches.  So in theory yes, and we can just go back to the original fix of multiplying `num_samples` and `eval_num_samples` by `dist.get_world_size()` in the `_log_dict()`.  However I'm not sure if this will account for when the dataset is pre-sharded (e.g., across nodes).  @hanlint @abhi-mosaic any suggestions on how we should handle this?

composer/callbacks/mlperf.py
ravi-mosaicml: hmm, the isinstance check should tell it that it is a ffcv dataloader and not just an iterable. Not a big deal, but could you paste the pyright error in a comment?

composer/callbacks/mlperf.py
ravi-mosaicml: `len(dataloader)` is always an accurate estimate -- it accounts for the sampler, so it will be "correct" (within rounding error due to distributed or drop last) regardless of whether the dataset is pre sharded or not.

abhi-mosaic: Re. @ravi-mosaicml 's comment, whatever logic you decide on, it should work the same for FFCV and a traditional PyTorch dataloader. The sharding of the dataset should not affect things, best I can tell. It's up to the dataset author to make sure that `len(dataset)` is implemented correctly so that `len(dataloader)`, which calls the former, acts correctly.

Slightly tangential, I think it's important to separate:

* how many samples are in the dataset
* how many samples your training loop will consume within one call to `iter(dataloader)`

These values are not the same, because the latter depends on your batch size, the number of devices, and `drop_last` configuration. Which one does MLPerf expect you to report? 

Some illustrative cases:

num_samples = 25, devices=1, batch_size=8, drop_last=True ---> num_samples_consumed=24
num_samples = 25, devices=8, batch_size=8, drop_last=True ---> num_samples_consumed=24
num_samples = 25, devices=1, batch_size=8, drop_last=False ---> num_samples_consumed=25
num_samples = 25, devices=8, batch_size=8, drop_last=False ---> num_samples_consumed=32 ***

Basically things are tricky due to DDP constraints "equal batches per device". I think in the training loop when `drop_last: True` is used, you can get away with the `len(dataloader) * device_batch_size * world_size` hack, but for the eval loop, you've got to be more careful. I think you may again get lucky if `eval_dataset_samples % num_devices == 0`, so like 50000 ImageNet samples divided by 8 GPUs.

bandish-shah: Ok good point, I looked at NVIDIA's submission and it looks like they're reporting total number of samples in Imagenet.  So we're going to keep using `len(dataloader.dataset)` to compute train/eval_samples for dataloaders that are not FFCV.  For FFCV we will use `len(dataloader.next_traversal_order())` which is what FFCV uses in the length computation:
https://github.com/libffcv/ffcv/blob/main/ffcv/loader/loader.py#L254

composer/callbacks/mlperf.py
bandish-shah: Sure would be helpful to understand.  This is the pyright failure:
```
Found 359 source files
/Users/bandishshah/composer/composer/callbacks/mlperf.py
  /Users/bandishshah/composer/composer/callbacks/mlperf.py:254:36 - error: Cannot access member "batch_size" for type "Iterable[Unknown]"
    Member "batch_size" is unknown (reportGeneralTypeIssues)
  /Users/bandishshah/composer/composer/callbacks/mlperf.py:254:63 - error: Cannot access member "next_traversal_order" for type "Iterable[Unknown]"
    Member "next_traversal_order" is unknown (reportGeneralTypeIssues)
2 errors, 0 warnings, 0 informations
Completed in 9.713sec
```

hanlint (approved):

bandish-shah:
Had to multiply by world_size but other than that testing looks good, here's from the last run:
```
:::MLLOG {"namespace": "", "time_ms": 1655254045030, "event_type": "POINT_IN_TIME", "key": "train_samples", "value": 1281168, "metadata": {"file": "/usr/local/lib/python3.9/dist-packages/composer/callbacks/mlperf.py", "lineno": 235}}
:::MLLOG {"namespace": "", "time_ms": 1655254045031, "event_type": "POINT_IN_TIME", "key": "eval_samples", "value": 50000, "metadata": {"file": "/usr/local/lib/python3.9/dist-packages/composer/callbacks/mlperf.py", "lineno": 235}}
```

PR #1157, opened 2022-06-13
Added streaming dataloader notebook

milocress:
Added a notebook for the FaceSynthetics dataset with a streaming data loader


ravi-mosaicml (commented):
Awesome work :rocket:

A few comments:

1. Can you add an emoij the title? Maybe 🤖? 
2. Can the pip installs be 1 line? And use `%pip` instead of `!pip`? Also, curious what does `-U` do? See https://stackoverflow.com/questions/70602866/is-there-an-advantage-of-using-over-in-jupyter-when-the-same-command-exi
3. At the top of the notebook, could you mention that we have the `mosicml/pytorch_vision` dockerimage, which contains mmcv and mmsegmentation preinstalled? https://github.com/mosaicml/composer/tree/dev/docker#docker
4. Is it possible to use the python `requests` instead of `curl` for the download? Would prefer to avoid using shell magics when possible. In addition, can you skip the download if the file is already downloaded?
5. In the global settings, can `tqdm = True` instead of `tqdm = 1`? Can you also add a comment on for each of the items in `dataset parameters` (E.g. what are the units of shard size limit? what is training ratio?)
7. Can the `os.mkdir` be `os.makedirs(exist_ok=True)` (for idempotency)?
8. Can the `open(...)` use the context manager `with open(...) as f:`? This will ensure the files are properly closed.
9. For `def __getitem__(self, i):`, would `obj = super()[i]` work? (Prefer not to call dunder methods when possible)
10. In the same method, can we use the functional torchvision API?
11. In the dataset, no need to move the tensors to the device -- the pytorch dataloader and/or the trainer should take care of that automatically.
12. In cleanup, can you use python functions instead of `!rm dataset.zip` and `rm -rf dataset`?
13. In the next steps, the list isn't showing up as a list -- probably need another newline.

Overall, well explained and easy to follow -- nice work!

CC @abhi-mosaic and @knighton -- I didn't review how the streaming dataloader API is used. Can you double check that this is how we would expect a user to use it? (If there are things that are off, then we likely need better documentation, or need to change the API to make it more use friendly -- would be good to track in JIRA)

Before merging, please test all of this in colab to make sure it runs.

hanlint:
Robot is a bit overused.. maybe.. 🌊 ? 🌈?

knighton:
How about the emoji 🎢 interpreted as "streaming dataloader is the supports that hold up the model as it ascends a training curve" or something of that nature lol

Need to spell out variable names if we hope to understand this code in the future

You don't pass shuffle= in DataLoader init, because the shuffling happens in the StreamingDataset because IterableDataset

Why is batch_size = 2? What happens when it is left as 1?

milocress:
> 8. For `def __getitem__(self, i):`, would `obj = super()[i]` work? (Prefer not to call dunder methods when possible)

Agreed on the dunder methods but this seems to be the only way, the above does not work.

> 9. In the same method, can we use the functional torchvision API?

Are you referring to the one hot encoding? The only method I could find for this was in `torch.nn.functional`, if you know of a better one please lmk.

milocress:
> Need to spell out variable names if we hope to understand this code in the future

lmk if the most recent changes sufficiently address this, if not I'll take another pass.

> Why is batch_size = 2? What happens when it is left as 1?

```
ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 512, 1, 1])
```

ravi-mosaicml (approved):
LGTM! A few last comments:

1. for the unzip, can you use `zipfile`? https://docs.python.org/3/library/zipfile.html
2. "use the tqdm progress bar during training" -> "Show a progress bar while downloading"
3. At the end, can the `os.remove` be guarded with a `try / except FileNotFoundError` (for idempotency)

Let me know if you need help getting the tests to pass!

knighton:
tqdm = true shadows the module name (use_tqdm or progbar or some such?)

I am suspicious of 'device = "gpu" if torch.cuda.is_available() else "cpu"', should that be cuda or an actual torch Device?

d, m, n variables should be spelled out (dirname, offset, count)?

missing some typing annotations

I think y and num_classes are wrong -- the classes should be dense -- the ignore label 255 should be mapped down to "19"?

abhi-mosaic:
Sorry to pile on requests, happy to do this in a follow-up PR too if that's easier as I'm sure we'll find some more edits...

* could we pass `batch_size` down to the custom StreamingDataset class? It's needed to ensure correct number of incomplete batches at the end of the epoch (we carefully construct CPU workers' samples so that as workers are cycled over round-robin by the DataLoader, the right worker at the end has exactly 1 incomplete batch)

knighton:
can specialize Iterator[dict] to Iterator[Dict[str, bytes]]

knighton (approved):

milocress:
Added a notebook for the FaceSynthetics dataset with a streaming data loader

PR #1158, opened 2022-06-13
Updated cuda to 11.5.2 for pytorch 1.11

ravi-mosaicml:
Some cuda drivers require pytorch to be built against a later version of the cuda sdk.


ravi-mosaicml:
Closing this PR as I do not think this is needed to fix the underlying issue. I believe the pytorch image in the dockerimage was not being used.

PR #1159, opened 2022-06-13
Fix hf notebook

A-Jacobson:
changed `LanguageCrossEntropy(vocab_size=tokenizer.vocab_size)` to `CrossEntropy` in the nlp and huggingface notebooks. It was causing an error in validation when it tried to reshape a `(batch size, num_class)` shaped tensor to `(bs, vocab size)` The notebooks are showing a classification (fine-tuning) example so I'm not sure how this wasn't throwing an error before. Have you seen this before @Landanjs, I believe you wrote the initial nlp notebook that I shamelessly copied for HF.

```RuntimeError: shape '[-1, 30522]' is invalid for input of size 32```


ravi-mosaicml (commented):
(Going to dismiss myself as a reviewer; will defer to @Landanjs.)

If this notebook runs without error on Colab, then it's good with me!

Landanjs (approved):
Hmmm, my original collab had `LanguageCrossEntropyLoss` instead which is gone now. So I guess it got changed at some point and broke? This LGTM as long as it runs!

PR #1160, opened 2022-06-13
Skip libcloud tests

hanlint:
Currently with `pip install .[dev]`, the pytests will fail for tests that require `libcloud`. This PR adds a bunch of `pytest.importorskip('libcloud')` to the codebase.

Is there a cleaner way to do this, than to annotate every test? Seems like a lot of clutter.

After:
```
============= 1801 passed, 235 skipped, 440 deselected, 43 xfailed, 8 xpassed in 72.81s (0:01:12) =============
```


ravi-mosaicml (approved):
LGTM; though I'm confused how this was passing CI/CD before without these importskips :thinking:

PR #1161, opened 2022-06-13
Composer Dockerimages

ravi-mosaicml:
This PR adds Composer Dockerimages to the build matrix.

* Building Composer images on top of the `vision_stage` with `pip install mosaicml[all]` so all dependencies are installed
* Installing composer from pip, rather than the local folder, to ensure the image is the same as if a user installed composer in the `mosaicml/pytorch` image
* Removed `docker/Makefile`, since that is just a glorified wrapper around `docker build`, and it was difficult to keep up to date (it's not being used in CI/CD). Instead, the default values are set directly inside the `Dockerfile` so running `docker build` locally, without arguments, works, and will produce an image with composer installed.


docker/build_matrix.yaml
```diff
@@ -95,3 +95,55 @@
   - mosaicml/pytorch_vision:latest_cpu
   TARGET: vision_stage
   TORCHVISION_VERSION: 0.12.0
+- BASE_IMAGE: nvidia/cuda:11.3.1-cudnn8-devel-ubuntu20.04
```
ajaysaini725, line 98: In the docs it says the base image is `mosaicml/pytorch` but here the base images are all `nvidia/cuda` - how come?

ajaysaini725 (approved):
Just a couple points otherwise LGTM 👍

docker/build_matrix.yaml
ravi-mosaicml: We use staged builds, so the BASE_IMAGE arg is always the root base image, but the Dockerfile code path effectively builds it on top of mosaicml/pytorch_vision.

docker/README.md
```diff
@@ -1,12 +1,33 @@
...
-MM Segmentation, and FFCV dependencies.
+leverage.
 
-## Pytorch Images
+## Composer Images
 
-<!-- BEGIN_BUILD_MATRIX -->
+The [mosaicml/composer](https://hub.docker.com/r/mosaicml/composer) images contain all Composer pre-installed with all dependencies for both NLP and Vision models. They are built on top of the [`mosaicml/pytorch_vision`](https://hub.docker.com/r/mosaicml/pytorch_vision) family of images. (See the section on [MosaicML PyTorch Images](#pytorch-images) below.)
+
+**Note**: Only the Dockerimage for most recent version of Composer will be maintained. We recommend using `mosaicml/composer:latest` or `mosaicml/composer:latest_cpu`, which will always be up to date. Out-of-date images
```
ravi-mosaicml, line 10:
```suggestion
**Note**: Only the Dockerimage for most recent version of Composer will be maintained. We recommend using `mosaicml/composer:latest` or `mosaicml/composer:latest_cpu`, which will always be up to date.
```

docker/README.md
```diff
@@ -1,12 +1,33 @@
...
+may be available on DockerHub, but are not shown in the table below.
```
ravi-mosaicml, line 11:
```suggestion
```

docker/generate_build_matrix.py
```diff
@@ -140,17 +162,50 @@ def _main():
...
+        entry = {
+            'BASE_IMAGE': _get_base_image(cuda_version),
+            'CUDA_VERSION': cuda_version,
+            'CUDA_VERSION_TAG': cuda_version_tag,
+            'LINUX_DISTRO': 'ubuntu2004',
+            'PYTHON_VERSION': python_version,
+            'PYTORCH_VERSION': pytorch_version,
+            'TARGET': 'composer_stage',
+            'TORCHVISION_VERSION': _get_torchvision_version(pytorch_version),
+            'COMPOSER_BASE': 'vision_stage',
```
ravi-mosaicml, line 188:
```suggestion
```

PR #1163, opened 2022-06-14
Fix Dev Install Tests

ravi-mosaicml:
#1133 effectively disabled tests for conditional imports, Previously Jenkins was configured to use the "lint" image for a dev install. Because the lint image was changed to `pytorch_vision` in #1133, pytest was being invoked with `-m vision`, which effectively selected no tests.

This PR fixes that by explicitely specifying that dev install tests should be run on the `mosaicml/pytorch:latest_cpu` image, which should apply the marker `-m 'not vision'`

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -34,6 +34,7 @@ pytestGpuImageTagsToTest = [
     'mosaicml/pytorch:latest',
     'mosaicml/pytorch_vision:latest',
 ]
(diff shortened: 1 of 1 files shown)


ravi-mosaicml:
It appears this fixed the issue and the dev tests ran: https://ci.int.mosaicml.com/blue/organizations/jenkins/scratch%2Fcommand2/detail/command2/30649/tests

hanlint (approved):

PR #1166, opened 2022-06-14
Gradient Clipping Part 2: A Little Cleanup

eracah:
* clean up checking for existing gradient clipping
* deepspeed config passed to State constructor
* add doctring to `deepspeed_enabled`

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -101,7 +101,7 @@ class State(Serializable):
         scaler (torch.cuda.amp.GradScaler, optional): The gradient scaler in use for mixed precision training.
         algorithms (Algorithm | Sequence[Algorithm], optional): The algorithms used for training.
         callbacks (Callback | Sequence[Callback], optional): The callbacks used for training.
-        profiler (Optional[Profiler]): The Composer profiler.
+        deepspeed_config (Dict[str, Any], optional): The configuration dictionary for deepspeed.
 
     Attributes:
         batch (types.Batch): The batch. This will be the entire batch during the :attr:`.Event.AFTER_DATALOADER`, or a
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -237,6 +237,9 @@ class State(Serializable):
         # algorithms and callbacks
         algorithms: Optional[Union[Algorithm, Sequence[Algorithm]]] = None,
         callbacks: Optional[Union[Callback, Sequence[Callback]]] = None,
+
+        # deepspeed.
+        deepspeed_config: Optional[Dict[str, Any]] = None,
     ):
         self.rank_zero_seed = rank_zero_seed
         self.model = model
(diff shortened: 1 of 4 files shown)


composer/trainer/trainer.py
```diff
@@ -740,6 +740,8 @@ def __init__(
         # Profiling
         profiler: Optional[Profiler] = None,
     ):
+        algorithms = list(ensure_tuple(algorithms)) if algorithms is not None else None
```
ravi-mosaicml, line 743:
```suggestion
        algorithms = list(ensure_tuple(algorithms))
```

composer/trainer/trainer.py
```diff
@@ -791,25 +793,16 @@ def __init__(
                 DeprecationWarning((f"Using the 'grad_clip_norm' field in Trainer is deprecated. Please use"
                                     'the GradientClipping Algorithm in composer.algorithms.gradient_clipping.')))
 
-            print_warning = False
             if algorithms is not None:
```
ravi-mosaicml, line 796: With the above change, I think `algorithms` will always be a list now, and then this check can be removed?

composer/trainer/trainer.py
```diff
@@ -791,25 +793,16 @@ def __init__(
...
                         ))
-
-            else:
-                if algorithms is not None:
-                    algorithms.append(GradientClipping(clipping_type='norm', clipping_threshold=grad_clip_norm))
                 else:
-                    algorithms = [GradientClipping(clipping_type='norm', clipping_threshold=grad_clip_norm)]
+                    algorithms.append(GradientClipping(clipping_type='norm', clipping_threshold=grad_clip_norm))
+            else:
+                algorithms = [GradientClipping(clipping_type='norm', clipping_threshold=grad_clip_norm)]
```
ravi-mosaicml, line 805: I think this can be removed -- don't think this code path will be taken (see the above comment)

composer/trainer/trainer.py

composer/trainer/trainer.py

composer/core/state.py
```diff
@@ -270,9 +273,10 @@ def __init__(
 
         self.profiler: Optional[Profiler] = None
 
+        self.deepspeed_config = deepspeed_config
```
ravi-mosaicml, line 276: Needs docstring in the Attributes section

ravi-mosaicml (approved):
LGTM; See comments

tests/algorithms/test_gradient_clipping.py
```diff
@@ -86,6 +86,7 @@ def test_gradient_clipping_algorithm(monkeypatch, clipping_type, simple_model_wi
     state.profiler.marker = Mock(return_value=None)
```
ravi-mosaicml, line 86: Would it be possible to use a real state object instead of a mock here? Mainly concerned about this test still passing if we change the state object.

eracah: fixed here: [b95594b](https://github.com/mosaicml/composer/pull/1166/commits/b95594b13447fcfbfb20b6400456429e19e9493a)

composer/core/state.py
eracah: done

PR #1167, opened 2022-06-14
Add an option to run all ffcv transformations only on cpu

dskhudia:
with `ffcv_cpu_only` option for imagenet dataset, one can force all ffcv transformations to run only on cpu. 

@florescl requested for AMD GPUs. BTW @florescl only problematic part is the use of cupy in [NormalizeImage](https://github.com/libffcv/ffcv/blob/main/ffcv/transforms/normalize.py#L61) transform. If you install a [ROCm supported version of cupy](https://docs.cupy.dev/en/v8.6.0/install_rocm.html#building-cupy-for-rocm) in your environment, this change is not needed. Also I think cpu-only version of NormalizeImage relies on compiler for vectorization and may be slow if vectorization doesn't happen.


florescl (approved):
awesome thanks!! i'll give it a try!

PR #1168, opened 2022-06-14
Update DeepLabv3+ Baseline Hyperparameters

Landanjs:
I swear this is the last time. Last week, it was noticed the deeplabv3+ could be trained on 4x the batch size while still maintaining the same performance across SSRs. This gives a modest reduction in training time on a single node, but could pretty impactful in multi-node settings. This also lets us remove sync batch norm.

Wandb for SSR=1: https://wandb.ai/mosaic-ml/landan-dlv3p-baseline/groups/baseline-bs128-ssr1.0/workspace?workspace=user-landanjs

### DIFF
--- a/composer/yamls/models/deeplabv3_ade20k_optimized.yaml
+++ b/composer/yamls/models/deeplabv3_ade20k_optimized.yaml
@@ -16,9 +16,9 @@ val_dataset:
     shuffle: false
 optimizers:
   decoupled_sgdw:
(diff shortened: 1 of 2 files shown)


abhi-mosaic:
Big fan of bigger batch sizes :) Maybe we can collect some multi-node benchmarks soon.

coryMosaicML (approved):
Looks good to me!

PR #1170, opened 2022-06-15
Purifies Hparams class by removing some fields

milocress:
removed `datadir` and `is_test` fields from hparams and put them into individual classes


composer/datasets/dataset_hparams.py
```diff
@@ -120,15 +120,12 @@ class DatasetHparams(hp.Hparams, abc.ABC):
...
-    is_train: bool = hp.optional('Whether to load the training data (the default) or validation data.', default=True)
     drop_last: bool = hp.optional(textwrap.dedent("""\
         If the number of samples is not divisible by the batch size,
         whether to drop the last batch (the default) or pad the last batch with zeros."""),
                                   default=True)
     shuffle: bool = hp.optional('Whether to shuffle the dataset for each epoch. Defaults to True.', default=True)
 
-    datadir: Optional[str] = hp.optional('The path to the data directory', default=None)
-
     @abc.abstractmethod
```
ravi-mosaicml, line 129: Cam upi also remove the "datadir" and "is_train" arguments from the docstring above?

ravi-mosaicml (approved):
LGTM! See comment.

composer/datasets/dataset_hparams.py
milocress: yes, good catch

composer/datasets/c4_hparams.py
```diff
@@ -49,6 +49,8 @@ class StreamingC4Hparams(DatasetHparams):
     mlm: bool = hp.optional('Whether or not to use masked language modeling.', default=False)
     mlm_probability: float = hp.optional('If `mlm==True`, the probability that tokens are masked.', default=0.15)
 
+    datadir: Optional[str] = hp.optional('The path to the data directory', default=None)
```
ravi-mosaicml, line 52: I don't think this class uses the `datadir`.

ravi-mosaicml (commented):
Whoops -- forgot about the docstrings. Could you make sure that `datadir` and `is_train` have docstring entries for any classes where you added in these args?

composer/datasets/ade20k_hparams.py
```diff
@@ -48,6 +49,9 @@ class ADE20kDatasetHparams(DatasetHparams, SyntheticHparamsMixin):
     final_size: int = hp.optional('Final size of the image and target', default=512)
     ignore_background: bool = hp.optional('If true, ignore the background class in training loss', default=True)
 
+    datadir: Optional[str] = hp.optional('The path to the data directory', default=None)
+    is_train: bool = hp.optional('Whether to load the training data (the default) or validation data.', default=True)
```
ravi-mosaicml, line 53: I don't think this class needs `is_train`

milocress: the constructor is invoked with `is_train=False` in `test_dataset_registry.py`

ravi-mosaicml: Can we fix that in the tests? Would prefer not to keep non-test codes around to support tests.

milocress: yes

milocress:
removed `datadir` and `is_test` fields from `hparams` and put them into individual classes

PR #1171, opened 2022-06-16
Fix bug where algorithm names weren't being logged properly in traces

coryMosaicML:


coryMosaicML:
Turns out some of our tests rely on things being done the previous way, hence engine tests now fail, and this could cause some strange behaviors if the same algorithm is used more than once as in the tests. So, this will take some more work.

hanlint:
Nice -- has this now exposed a passing test is actually failing?

coryMosaicML:
No, things work correctly on current dev, only issue is some of our docstrings are misleading, and changing code to be in line with the docs causes things to break.

coryMosaicML:
Closing this for now until I fix it properly.

composer/core/engine.py
```diff
@@ -144,7 +146,7 @@ def _setup_trace(algorithms: Sequence[Algorithm], event: Event) -> Traces:
     The keys are of format ``<algorithm_name>/<event>`` (e.g.,  ``Blurpool/INIT``) and values are an instance of
     :class:`Trace`.
     """
-    return OrderedDict([(f'{algo}/{event}', Trace()) for algo in algorithms])
+    return OrderedDict([(f'{algo}/{event}', Trace(name=algo.__class__.__name__)) for algo in algorithms])
```
ravi-mosaicml, line 149: I don't think the `name` needs to be added to the trace, but rather, `{algo}` should be `{type(alg).__name__}`

PR #1175, opened 2022-06-16
Fix Printing of the Stack Trace on POSIX Signal Exits

### LINKED ISSUE #1165: Resnet benchmark crashed with exit code -6
**Environment**

- OS: [Ubuntu 20.04]
- Hardware (GPU, or instance type): [AMD Instinct/ROCm 5.1.1]

**To reproduce**

1.  Execute the recipe in https://github.com/mosaicml/benchmarks/tree/main/blogs/resnet. Specifically running the recipe: recipes/resnet50_hot.yaml
2. Benchmark runs for several epochs (and seems to be doing well) after which composer prints an error and terminates the run
ERROR:composer.cli.launcher:Rank 2 crashed with exit code -6

Dont see any error or stack trace on why the specific rank exited.  I enabled FileLogger and dont see any error / stack trace there as well. Here is the last few lines of the rank 2 log file that was generated. 

[EPOCH][batch=42500]: { "metrics/eval/Accuracy": 0.7575, "metrics/eval/CrossEntropy": 1.6468, }
[EPOCH][batch=42500]: { "epoch": 68, }
[EPOCH][batch=42500]: { "metrics/eval/Accuracy": 0.7575, "metrics/eval/CrossEntropy": 1.6468, }
[EPOCH][batch=42500]: { "epoch": 68, }
[stderr]: INFO:composer.algorithms.progressive_resizing.progressive_resizing:Applied Progressive Resizing with scale_factor=0.9820359281437125 and mode=resize.
[stderr]: Old input dimensions: (H,W)=(167, 167).
[stderr]: New input dimensions: (H,W)=(164, 164)

Is there a way to enable more logging or understand what the exit code -6 means? 

## Expected behavior

Benchmark runs to completion.

ravi-mosaicml:
When using the launcher script, the stack trace would not be printed when the process is terminated via a POSIX signal (e.g. `SIGABRT` -- see #1165). This is to prevent cluttering the logs when the launcher script terminates the process via SIGTERM or SIGKILL.

Instead of skipping the stack trace on all negative exit codes (e.g. all signals), this PR skips the stack trace only when the process was terminated via SIGKILL or SIGTERM, which are the two codes that the launcher script uses to terminate processes.

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -414,7 +414,7 @@ def _cleanup_processes(processes: Dict[int, subprocess.Popen]):
     for global_rank, process in processes.items():
         process.poll()
         if process.returncode is not None and process.returncode != 0:
-            if process.returncode < 0 and -process.returncode in (signal.SIGKILL, signal.SIGTERM):
+            if -process.returncode in (signal.SIGKILL, signal.SIGTERM):
                 # Negative return codes indicate the process was killed via a signal
                 # If the launcher script killed the training process (which would happen via SIGKILL or SIGTERM),
                 # then do not print the stack trace.
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -413,10 +413,14 @@ def _cleanup_processes(processes: Dict[int, subprocess.Popen]):
                         pass
     for global_rank, process in processes.items():
         process.poll()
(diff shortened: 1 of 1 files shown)


hanlint (approved):
This LGTM, but can you unlink the issue? This PR fixes the stack trace, but not the issue of resnet crashing on AMD chips.

PR #1176, opened 2022-06-16
Fix bug where algorithm names weren't being used to log traces

coryMosaicML:
This fixes the annoying issue of strange and unique to each run algorithm traces appearing in logs. However, if algorithms of the same name are used multiple times in the same run, logs might not be correct with this fix.

[A run showing what things look like after this change in wandb](https://wandb.ai/mosaic-ml/cory-debug/runs/u3mdtdmv)


ravi-mosaicml (approved):
LGTM. Down the line would like to significantly refactor or remove the Trace class, but if this helps the logging in the interim, all for it.

eracah:
CO-593

PR #1177, opened 2022-06-16
shardwise zip streaming datasets

milocress:
zips streaming datasets

* Addresses CO-550


ravi-mosaicml (commented):
Nice job!

1. Would it be possible to make enabling compression (and the compression level?) config options? Ok to default it to true. With checkpointing, we found that compressing significantly slows things down, so it may be helpful (when debugging) to disable it or if egress fees don't matter.
2. Can you add compression testing to the test matrix?

I'll take a closer look after these changes.

abhi-mosaic:
Adding on to Ravi's comments, I would like it if we can somehow auto-detect if the remote directory is compressed, and also maintain user readability of file types (`000007.mds` vs. `000007.mds.gz`). Anything we can do to avoid adding more args to the `StreamingDataset` init signature would be much appreciated :)

ravi-mosaicml (commented):
Thanks for updating this!

A few high level questions:
1. Can the download function automatically infer the compression alg, if any, via the file extension?
2. Do we need to store the compression alg in the index file? Mainly concerned about heterogenous shards where some might be compressed and some might not be compressed, or different compression algs could be used (e.g. when training on multiple datasets from different sources)
3. Can the index file also be compressed? (OK deferring this one to a future PR)

composer/datasets/streaming/format.py
```diff
@@ -183,8 +229,9 @@ def load(cls, fp: Union[BufferedReader, BytesIO]):
         bytes_per_sample = bytes_per_sample.astype(np.int64)
         num_fields, = read_array(fp, 1, np.int32)
         bytes_per_field = read_array(fp, num_fields, np.int32)
+        compression_scheme_id, = read_array(fp, 1, np.int8)
```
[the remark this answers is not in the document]
abhi-mosaic: Not making a comment amount about implementation, but of use cases --- I don't see any situation where some shards will be compressed and others not, as samples are randomly distributed amongst shards. I think that's just over complicating. Lets treat the common case and make it simple.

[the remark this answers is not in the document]
milocress: This sounds possible to represent as a union of multiple datasets? We can potentially add tooling for combining streaming datasets as monoids, but if you have three different sources it seems reasonable to have three different index files too.

tests/datasets/test_streaming.py
```diff
@@ -34,17 +34,22 @@ def get_fake_samples_decoders(num_samples: int) -> Tuple[List[Dict[str, bytes]],
...
+                                shard_size_limit=shard_size_limit,
+                                compression=compression) as writer:
         writer.write_samples(samples=samples)
 
 
 @pytest.mark.timeout(10)
 @pytest.mark.parametrize('num_samples', [100, 10000])
 @pytest.mark.parametrize('shard_size_limit', [1 << 8, 1 << 16, 1 << 24])
-def test_writer(remote_local: Tuple[str, str], num_samples: int, shard_size_limit: int) -> None:
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 50: Could this be its own test? Adding the parameterization here increases the number of tess across {num_samples} * {shard_size_limit} * {compression}; I don't believe that compression depends on the num samples are shard size limit?

tests/datasets/test_streaming.py
```diff
@@ -66,14 +74,19 @@ def test_writer(remote_local: Tuple[str, str], num_samples: int, shard_size_limi
 @pytest.mark.parametrize('batch_size', [None, 1, 2])
 @pytest.mark.parametrize('share_remote_local', [False, True])
 @pytest.mark.parametrize('shuffle', [False, True])
-def test_reader(remote_local: Tuple[str, str], batch_size: int, share_remote_local: bool, shuffle: bool):
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 77: Same thing as above

tests/datasets/test_streaming.py
```diff
@@ -106,15 +119,23 @@ def test_reader(remote_local: Tuple[str, str], batch_size: int, share_remote_loc
 @pytest.mark.timeout(10)
 @pytest.mark.parametrize('created_ago', [0.5, 3])
 @pytest.mark.parametrize('timeout', [1])
-def test_reader_after_crash(remote_local: Tuple[str, str], created_ago: float, timeout: float):
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 122: Do we need to test compression here?

tests/datasets/test_streaming.py
```diff
@@ -131,20 +152,24 @@ def test_reader_after_crash(remote_local: Tuple[str, str], created_ago: float, t
         pytest.param(False, marks=pytest.mark.xfail(reason='__getitem__ currently expects shards to exist')),
     ],
 )
-def test_reader_getitem(remote_local: Tuple[str, str], share_remote_local: bool):
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 155: Same as above; does compression affect `reader_get_item`?

tests/datasets/test_streaming.py
```diff
@@ -162,13 +187,17 @@ def test_reader_getitem(remote_local: Tuple[str, str], share_remote_local: bool)
         )),
 ])
 @pytest.mark.parametrize('shuffle', [False, True])
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 190: Same as above...

tests/datasets/test_streaming.py
```diff
@@ -234,8 +263,9 @@ def test_dataloader_single_device(remote_local: Tuple[str, str], batch_size: int
 @pytest.mark.parametrize('num_samples', [30, 31])
 @pytest.mark.parametrize('num_workers', [1, 3])
 @pytest.mark.parametrize('shuffle', [False, True])
+@pytest.mark.parametrize('compression', [None, 'gz:5', 'gz'])
```
ravi-mosaicml, line 266: Same...

composer/datasets/streaming/writer.py
```diff
@@ -60,9 +61,14 @@ class StreamingDatasetWriter(object):
         dirname (str): Directory to write shards to.
         fields: (List[str]): The fields to save for each sample.
         shard_size_limit (int): Maximum shard size in bytes. Default: `1 << 24`.
+        compression (Optional[str]): Compression algorithm and optional compression level. Currently supported: 'gz', 'gz:[1-9]' or None.
```
ravi-mosaicml, line 64:
```suggestion
        compression (str, optional): Compression algorithm and optional compression level. Currently supported: 'gz', 'gz:[1-9]' or None. Defaults to None, for no compression.
```

composer/datasets/streaming/writer.py
```diff
@@ -60,9 +61,14 @@ class StreamingDatasetWriter(object):
...
     """
 
-    def __init__(self, dirname: str, fields: List[str], shard_size_limit: int = 1 << 24) -> None:
+    def __init__(self,
+                 dirname: str,
+                 fields: List[str],
+                 shard_size_limit: int = 1 << 24,
+                 compression: Optional[str] = 'gz') -> None:
```
ravi-mosaicml, line 71: CC @abhi-mosaic, but should this default to gz or None? I think I would prefer None because of the additional compression/decompression overhead it can add, but can be convinced otherwise if using compression is strictly better. (i.e. we would still use compression on clouds where we would have zero egress and storage fees)

composer/datasets/streaming/writer.py
```diff
@@ -82,29 +88,63 @@ def __init__(self, dirname: str, fields: List[str], shard_size_limit: int = 1 <<
         self.new_samples = []
         self.new_shard_size = 0
 
+        # compression scheme for shards
+        self.compression_scheme = None
+        self.open_f = lambda n: open(n, 'xb')
```
ravi-mosaicml, line 93: Instead of defining `open_f` as a lambda, could you make this a private helper function? It's a bit unstandard to define a lambda as a self variable, rather than use a function, and makes the code less readable (typically one doesn't look at `__init__` to find where methods are defined).

For example:

E.g.

```python
def _open_file(self, filename: str):
    if self.compression_scheme.startswith("gz"):
        ...
        return gzip.open(filename, ...)
    if self.compression_scheme is None:
        return open(filename, ...)
    else:
        raise ValueError(f"Invalid compression scheme: {self.compression_scheme}")
```

composer/datasets/streaming/writer.py
```diff
@@ -82,29 +88,63 @@ def __init__(self, dirname: str, fields: List[str], shard_size_limit: int = 1 <<
...
+        self._set_compression_settings(compression)
```
ravi-mosaicml, line 94: I like that this parsing is done via a helper. But generally speaking, it'd be preferable to have a function do one thing and only one thing. Could this work?

```suggestion
        self.compression_scheme = _get_compression_scheme(compression)
```

Then, `_get_compression_scheme` could be a static helper function defined at the top of the file (need not be an instance method, but ok with me if it is)

composer/datasets/streaming/writer.py
```diff
@@ -82,29 +88,63 @@ def __init__(self, dirname: str, fields: List[str], shard_size_limit: int = 1 <<
...
+        with self.open_f(filename) as out:
             for data in self.new_samples:
                 out.write(data)
+
         self.samples_per_shard.append(len(self.new_samples))
         self.bytes_per_shard.append(self.new_shard_size)
         self.new_samples = []
         self.new_shard_size = 0
 
+    def _write_compression_scheme(self) -> None:
```
ravi-mosaicml, line 130: Curious why do we need to store the compression metadata in a separate file?

composer/datasets/streaming/format.py
```diff
@@ -21,25 +23,57 @@
 ]
 
 
-def get_index_basename() -> str:
+def split_compression_suffix(local_path: str) -> Tuple[str, Optional[str]]:
```
ravi-mosaicml, line 26: Can this be a private helper?

composer/datasets/streaming/format.py
```diff
@@ -21,25 +23,57 @@
...
+        Tuple[str, str]: tuple containing decompressed filename and compression suffix, if one exists
+    """
+    decompressed_path, ext = splitext(local_path)
+    if ext == '.mds' or ext == '.txt':
+        return local_path, None
+
+    return decompressed_path, ext[1:]
+
+
+def get_compression_scheme_basename() -> str:
```
ravi-mosaicml, line 42: (Same comment as below, but why do we need a compression scheme file?)

composer/datasets/streaming/download.py
```diff
@@ -94,7 +96,21 @@ def dispatch_download(remote, local, timeout: float):
         local (str): Local path (local filesystem).
         timeout (float): How long to wait for file to download before raising an exception.
     """
-    if os.path.exists(local):
+
+    def dispatch_decompress(local, local_decompressed, compression_scheme):
```
ravi-mosaicml, line 100: Can this function be a private helper, rather than being defined within a function? Also needs type annotations

Also, does this need to be a function, or can it be hardcoded on line 127? I think it is only used in one place?

composer/datasets/streaming/dataset.py
```diff
@@ -96,10 +97,19 @@ def __init__(self,
         self.timeout = timeout
         self.batch_size = batch_size
 
+        compression_basename = get_compression_scheme_basename()
+        compression_local = self._download_file(compression_basename, wait=(dist.get_local_rank() != 0))
+        with open(compression_local, 'r+') as fp:
+            compression_scheme = fp.read()
+            self.compression_scheme = compression_scheme if compression_scheme != '' else None
```
ravi-mosaicml, line 104: Curious why are we truncating the file?

composer/datasets/streaming/writer.py
milocress: if we store it in the index file, we wouldn't be able to compress the index file.

composer/datasets/streaming/dataset.py
milocress: local datasets have no compression scheme, so we empty the compression metadata file.

composer/datasets/streaming/format.py
milocress: I don't believe so - it's used in several different files

composer/datasets/streaming/writer.py
milocress: My intuition here is that on an ML box it would take longer to {up/down}load even a medium sized dataset than to [de]compress it -- happy to run benchmarks after this lands to test things though.

composer/datasets/streaming/writer.py
ravi-mosaicml: Hmm, could we have the user pass the path to the index file (including the compression extension, if any), rather than the path to the folder?

composer/datasets/streaming/dataset.py
ravi-mosaicml: Would it work to leave the file as-is if local?

composer/datasets/streaming/writer.py
milocress: > Anything we can do to avoid adding more args to the StreamingDataset init signature would be much appreciated :)
CC @abhi-mosaic I know we're hesitant to add/change the arguments to the streaming dataset function, what do you think of this?

composer/datasets/streaming/dataset.py
milocress: you're right, we can't support compression for `remote==local` anyway, I think we should throw some kind of error in `__init__` if the user attempts this.

composer/datasets/streaming/download.py
```diff
@@ -33,7 +36,10 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
 
     config = Config(read_timeout=timeout)
     s3 = boto3.client('s3', config=config)
-    s3.download_file(obj.netloc, obj.path.lstrip('/'), local)
+    try:
+        s3.download_file(obj.netloc, obj.path.lstrip('/'), local)
+    except DataNotFoundError:
```
ravi-mosaicml, line 41:
```suggestion
    except DataNotFoundError as e:
```

composer/datasets/streaming/download.py
```diff
@@ -33,7 +36,10 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
...
+        raise FileNotFoundError
```
ravi-mosaicml, line 42:
```suggestion
        raise FileNotFoundError from e
```

tests/datasets/test_streaming.py
```diff
@@ -330,3 +358,35 @@ def test_dataloader_multi_device(remote_local: Tuple[str, str], batch_size: int,
...
+                                      shard_size_limit=shard_size_limit,
+                                      compression=None)
+
+    dataset = StreamingDataset(remote=compressed, local=local, shuffle=shuffle, decoders=decoders)
+
+    list(x for x in dataset)  # download samples
+
+    dcmp = dircmp(remote, local)
+
+    def check_for_diff_files(dir: dircmp):
```
ravi-mosaicml, line 387: Could this be hardcoded, or a private helper rather than a function in function?

tests/datasets/test_streaming.py
```diff
@@ -330,3 +358,35 @@ def test_dataloader_multi_device(remote_local: Tuple[str, str], batch_size: int,
...
+                                      shard_size_limit=shard_size_limit,
+                                      compression=compression)
+    write_synthetic_streaming_dataset(dirname=remote,
+                                      samples=samples,
+                                      shard_size_limit=shard_size_limit,
+                                      compression=None)
+
+    dataset = StreamingDataset(remote=compressed, local=local, shuffle=shuffle, decoders=decoders)
+
+    list(x for x in dataset)  # download samples
```
ravi-mosaicml, line 383: The comment here is super helpful. Would prefer to iterate over it explicitly rather than convert it into a list (nit)
```suggestion
    for x in dataset:
        pass  # download sample
```

tests/datasets/test_streaming.py
```diff
@@ -330,3 +358,35 @@ def test_dataloader_multi_device(remote_local: Tuple[str, str], batch_size: int,
         assert sample_order != second_sample_order
     else:
         assert sample_order == second_sample_order
+
+
+@pytest.mark.timeout(10)
+@pytest.mark.parametrize('compression', [None, 'gz', 'gz:5'])
+def test_compression(compressed_remote_local: Tuple[str, str, str], compression: Optional[str]):
```
ravi-mosaicml, line 365: For `compression=None`, can you delete the compression file (for backwards compatibility)?

composer/datasets/streaming/writer.py
```diff
@@ -4,19 +4,39 @@
...
 
-from composer.datasets.streaming.format import (StreamingDatasetIndex, get_index_basename, get_shard_basename,
-                                                sample_dict_to_bytes)
+from composer.datasets.streaming.format import (StreamingDatasetIndex, get_compression_scheme_basename,
+                                                get_index_basename, get_shard_basename, sample_dict_to_bytes)
 
 __all__ = ['StreamingDatasetWriter']
 
 
+def _parse_compression_args(compression: Optional[str]) -> Dict[str, Any]:
```
ravi-mosaicml, line 22: Would prefer to return a tuple of (compression_scheme, compresslevel) rather than a dict -- since this is an internal API, we can change it later as we support additional compression options.

```suggestion
def _parse_compression_args(compression: Optional[str]) -> Tuple[str, int]:
```

ravi-mosaicml (commented):
Really close! Main change request is to skip writing the compression file if no compression is used -- this will help ensure backwards compatibility. Otherwise, just some nits.

tests/datasets/test_streaming.py
ravi-mosaicml: (NVM, with the requested change, if compression=None, then no file should be written)

tests/datasets/test_streaming.py
```diff
@@ -34,17 +44,24 @@ def get_fake_samples_decoders(num_samples: int) -> Tuple[List[Dict[str, bytes]],
...
+                                shard_size_limit=shard_size_limit,
+                                compression=compression) as writer:
         writer.write_samples(samples=samples)
 
 
 @pytest.mark.timeout(10)
 @pytest.mark.parametrize('num_samples', [100, 10000])
 @pytest.mark.parametrize('shard_size_limit', [1 << 8, 1 << 16, 1 << 24])
-def test_writer(remote_local: Tuple[str, str], num_samples: int, shard_size_limit: int) -> None:
+@pytest.mark.parametrize('compression', ['gz', None])
```
ravi-mosaicml, line 62: Can you remove this parameterization here? This will 2x the number of tests across num_samples and shard_size_limit. I believe the writer is tested below.

ravi-mosaicml (approved):
LGTM! See one comment re test cases. Nice work overall

PR #1184, opened 2022-06-17
Only catch OOM if `grad_accum='auto'`

hanlint:
This PR fixes a reported bug where gradient accumulation was always automatic.

Coming along for the ride -> some better logging of the grad accum increases.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1645,7 +1645,7 @@ class Trainer:
                         else:
                             optimizer.step()
             except RuntimeError as e:
-                if _is_cuda_oom(e):
+                if self.adaptive_gradient_accumulation and _is_cuda_oom(e):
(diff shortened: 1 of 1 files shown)


moinnadeem (approved):
Can confirm that I am seeing OOMs again

composer/trainer/trainer.py
```diff
@@ -1645,7 +1645,7 @@ def _train_batch(self, use_grad_scaling: bool):
                         else:
                             optimizer.step()
             except RuntimeError as e:
-                if _is_cuda_oom(e):
+                if self.adaptive_gradient_accumulation and _is_cuda_oom(e):
```
hanlint, line 1648: @ravi-mosaicml is it OK to do the check here? or should we always run the `should_handle_cuda_oom` logic, and then reraise it so that all the ranks fail?

hanlint:
Merging now -- created CO-535 to track test cases for automatic gradient accumulation.

PR #1185, opened 2022-06-17
Add notebook for auto grad accum

hanlint:
Adds a notebook demonstrating automatic gradient accumulation in action. The settings are designed to exceed memory limits in a colab notebook (e.g. K80 or V100 GPU), and users are encouraged to try and experiment with different settings.


examples/auto_grad_accum.ipynb
```diff
@@ -0,0 +1,257 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "colab": {
+     "base_uri": "https://localhost:8080/",
+     "height": 283,
+     "referenced_widgets": [
```
hanlint, line 170: hmm for some reason nbstrip did not remove these, will remove manually.

ravi-mosaicml (approved):
Overall LGTM!

1. Can `!pip install mosaicml` be `%pip install mosaicml`? (Would prefer to use the jupyter magic than a shell command)
2. Can the `data_directory` be `./data` to be consistent with the other notebooks?
3. Should we increase the gradient accumulation logs to loglevel.info? I think they are infrequent (and important) enough; don't mind if they are at a higher level.
4. Could the link to the Auto Grad Accum documentation be also at the top of the notebook (for better visibility)?
5. Can you include an optimizer in the trainer (to hide the no optimizer warning?)

ravi-mosaicml:
One more thing: Can the headings for "Set Up Our Workspace" and "Train a Baseline Model" be h2 or below? Otherwise it shows up in the sidebar on the docs site.

image

hanlint:
Thanks, will address these then merge!

PR #1187, opened 2022-06-17
Move grad_accum logging to every step

coryMosaicML:
Currently, the value of `grad_accum` is logged only when it is changed by `grad_accum="auto"`. This PR changes things such that the value of `grad_accum` is logged every step.

Example old behavior:
Screen Shot 2022-06-17 at 11 47 57 AM

Example new behavior:
Screen Shot 2022-06-17 at 11 48 52 AM


abhi-mosaic (approved):
Love it :) Thanks for fixing this @coryMosaicML !

PR #1189, opened 2022-06-17
A utility to fuse parallel linear layers in FX-traced models

dskhudia:
- All parallel linears will be fused. It's not limited to 2. 
- Either all or none should have the bias for as a pre-condition of fusion. 
- If no parallel linears are found, unmodified GraphModule is returned. 

Example,

Before:
Screen Shot 2022-06-17 at 9 53 03 AM

After:
Screen Shot 2022-06-17 at 9 53 22 AM

CO-402

### DIFF
--- a/composer/utils/fx_utils.py
+++ b/composer/utils/fx_utils.py
@@ -7,15 +7,19 @@ Provides utilities to do FX-based model transformations.
 """
 
 import logging
-from typing import Any, Callable, List, Union
+import operator
+from typing import Any, Callable, Dict, List, Mapping, Tuple, Union
 
+import torch
+import torch.nn as nn
+from torch.fx import Node
 from torch.fx.graph_module import GraphModule
 
 from composer.utils import ensure_tuple
 
 log = logging.getLogger(__name__)
 
-__all__ = ['count_op_instances', 'replace_op']
+__all__ = ['count_op_instances', 'replace_op', 'fuse_parallel_linears']
 
 
 def count_op_instances(gm: GraphModule, ops: Union[Callable, str, List[Union[Callable, str]]]) -> int:
@@ -51,7 +55,15 @@ def count_op_instances(gm: GraphModule, ops: Union[Callable, str, List[Union[Cal
         int: The number of instances of ``ops`` in ``gm``
     """
     ops = list(ensure_tuple(ops))
-    return sum(any(n.target == op for op in ops) for n in gm.graph.nodes)
+    all_modules = dict(gm.named_modules())
+    count = 0
+    for n in gm.graph.nodes:
+        for op in ops:
+            if n.target == op:
+                count += 1
+            elif n.op == 'call_module' and isinstance(op, type) and isinstance(all_modules[n.target], op):
+                count += 1
+    return count
 
 
 def replace_op(gm: GraphModule, src_ops: Union[Callable, str, List[Union[Callable, str]]],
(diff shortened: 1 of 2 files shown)


composer/utils/fx_utils.py
```diff
@@ -51,7 +54,15 @@ def count_op_instances(gm: GraphModule, ops: Union[Callable, str, List[Union[Cal
...
     """
     ops = list(ensure_tuple(ops))
-    return sum(any(n.target == op for op in ops) for n in gm.graph.nodes)
+    all_modules = dict(gm.named_modules())
+    count = 0
+    for n in gm.graph.nodes:
+        for op in ops:
+            if n.target == op:
+                count += 1
+            elif n.op == 'call_module' and isinstance(all_modules[n.target], op):
```
ravi-mosaicml, line 63: I'm a bit confused here when this isinstance check would return True if `op` is a string or a generic function. For example, the following errors:

```python
>>> isinstance(2, "hi")
  File "<stdin>", line 1, in <module>
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

composer/utils/fx_utils.py
```diff
@@ -123,13 +134,106 @@ def replace_residual_with_stochastic(gm: GraphModule):
     raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
 
 
-def fuse_parallel_linears(gm: GraphModule):
+def _can_linears_be_fused(linear_nodes: List, all_modules: Dict) -> bool:
+    """Check if all the linears have bias."""
+    bias = all_modules[linear_nodes[0].target].bias is None
+
+    ret_val = True
```
ravi-mosaicml, line 141:
```suggestion
    return all(bias == all_modules[node.target].bias is None for node in linear_nodes)
```

composer/utils/fx_utils.py
```diff
@@ -123,13 +134,106 @@ def replace_residual_with_stochastic(gm: GraphModule):
...
+    for node in linear_nodes:
+        ret_val = ret_val and bias == (all_modules[node.target].bias is None)
+    return ret_val
+
+
+def _create_fused_linear(linear_nodes: List, all_modules: Dict, keep_weights: bool = False) -> nn.Module:
```
ravi-mosaicml, line 147: Same as above; can the type annotations be more specific?

ravi-mosaicml (approved):
LGTM! This is neat, thanks! See comments.

composer/utils/fx_utils.py
dskhudia: Thanks for catching this error. This check should have only been executed when `op` is a class. Fixed it. It counts the instances of nn.Modules. For example, 

```
>>> class M(torch.nn.Module):
...   def __init__(self):
...     self.fc = nn.Linear(8, 8)
...     super().__init__()
...   def forward(self, x):
...     return self.fc(x)
>>> module = M()
>>> traced = symbolic_trace(module)
>>> count_op_instances(traced, torch.add)
0
>>> count_op_instances(traced, nn.Linear)
1
>>> count_op_instances(traced, "hi")
0
>>> count_op_instances(traced, nn.ReLU)
0
```

hanlint (commented):
Does this also need to change items in the optimizers?

dskhudia:
> Does this also need to change items in the optimizers?

It does. Let me do that in a followup PR.

PR #1190, opened 2022-06-18
ProgressBar Units

hanlint:
This PR:
* lightly refactors the ProgressBar for code clarity
* renders the progress bar in units of `max_duration`
* fixes interleaving of train/eval progress bars (see: #1191 )

## Refactor notes:
* Adds `current_pbar` property instead of using dictionaries for progress bars
* renames `_ProgressBarLoggerInstance` -> `_ProgressBar` for clarity/bervity. The object is a progress bar ala tqdm, not an instance of the logger itself. 
* Removes ProgressBarLoggerInstanceState, as unneccessary.
* Folds local rank zero guards into `show_pbar` property

## New behavior:

`max_duration=10ep`, `eval_interval=1ep`

```
Epoch     0 train 100%|█████████████████████████| 29/29 [00:09<00:00,  3.14ba/s, ...
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.35ba/s, ...
```

`max_duration=1000ba`, `eval_interval=10ba`
```
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.36ba/s, 
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.30ba/s, 
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.04ba/s, 
Batch       train   7%|█▋                       | 68/1000 [00:13<02:58,  5.21ba/s, loss/train=0.
```

`max_duration=1e6sp`, `eval_interval=1ep`
```
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.20ba/s, 
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.08ba/s,
Sample      train  15%|███▊                     | 150528/1000000 [00:22<02:06, 6725.71sp/s, 
# note different units for val vs. train. "ba/s" vs "sp/s". We have no access to the # of 
# samples in the eval dataloader ahead of time, so must count in batches.
```

Note: In cases where the `eval_interval` happens when the training bar is still active, the training bar will pause during eval, and then resume. The training bar will be sorted to the bottom, once training resumes, because terminals only support editing the last line (see last two examples above).

### DIFF
--- a/composer/loggers/progress_bar_logger.py
+++ b/composer/loggers/progress_bar_logger.py
@@ -12,6 +12,7 @@ from typing import Any, Callable, Dict, List, Optional, TextIO, Union
 import tqdm.auto
 
 from composer.core.state import State
+from composer.core.time import TimeUnit, Time
 from composer.loggers.logger import Logger, LogLevel, format_log_data_value
 from composer.loggers.logger_destination import LoggerDestination
 from composer.utils import dist
@@ -43,14 +44,16 @@ class _ProgressBar:
         file: TextIO,
         metrics: Dict[str, Any],
         keys_to_log: List[str],
+        unit: TimeUnit = TimeUnit.EPOCH,
     ) -> None:
         self.keys_to_log = keys_to_log
         self.metrics = metrics
         self.position = position
+        self.unit = unit
         self.pbar = tqdm.auto.tqdm(
             total=total,
             position=position,
-            bar_format=bar_format,  # f'{description} {{l_bar}}{{bar:25}}{{r_bar}}{{bar:-1b}}',
+            bar_format=bar_format,
             file=file,
             dynamic_ncols=True,
             postfix=metrics,
@@ -77,6 +80,7 @@ class _ProgressBar:
             'metrics': self.metrics,
             'keys_to_log': self.keys_to_log,
             'n': pbar_state['n'],
+            'unit': self.unit
         }
 
 
@@ -130,6 +134,9 @@ class ProgressBarLogger(LoggerDestination):
         self.eval_pbar: Optional[_ProgressBar] = None
         self.is_train: Optional[bool] = None
 
+        if isinstance(console_log_level, str):
+            console_log_level = LogLevel(console_log_level)
+
         if log_to_console is None:
             log_to_console = not progress_bar
 
@@ -138,9 +145,6 @@ class ProgressBarLogger(LoggerDestination):
             self.should_log = lambda state, ll: False
         else:
             # set should_log to a Callable[[State, LogLevel], bool]
-            if isinstance(console_log_level, str):
-                console_log_level = LogLevel(console_log_level)
-
             if isinstance(console_log_level, LogLevel):
                 self.should_log = lambda state, ll: ll <= console_log_level
             else:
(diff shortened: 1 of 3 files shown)


composer/loggers/progress_bar_logger.py
```diff
@@ -21,44 +21,67 @@
...
+    def __init__(
+        self,
+        total: Optional[int],
+        position: int,
+        bar_format: str,
+        file: TextIO,
+        metrics: Dict[str, Any],
+        keys_to_log: List[str],
+        unit: TimeUnit = TimeUnit.EPOCH,
+        epoch_style: bool = False,
```
ravi-mosaicml, line 35: Curious if we need the `epoch_style` argument? For evaluation, `batch_in_epoch` is always the same as `batch`, (same thing with samples and tokens), so I think it should be safe to always use the `x_in_epoch` value? 

I realize that `Timestamp.get(TimeUnit)` has no way to get the `x_in_epoch` -- that's an oversight. Thoughts about adding in a `get_in_epoch(TimeUnit)` helper?

composer/loggers/progress_bar_logger.py
```diff
@@ -21,44 +21,67 @@
...
 
     def log_data(self, data: Dict[str, Any]):
-        formatted_data = {k: format_log_data_value(v) for (k, v) in data.items() if k in self.state.keys_to_log}
-        self.state.epoch_metrics.update(formatted_data)
-        self.pbar.set_postfix(self.state.epoch_metrics)
+        formatted_data = {k: format_log_data_value(v) for (k, v) in data.items() if k in self.keys_to_log}
+        self.metrics.update(formatted_data)
+        self.pbar.set_postfix(self.metrics)
+
+    def update(self, n=1):
```
ravi-mosaicml, line 57: nit: do we need the default value, or is it always specified? 
```suggestion
    def update(self, n: int):
```

composer/loggers/progress_bar_logger.py
```diff
@@ -105,140 +128,178 @@ def __init__(
...
                 stream = sys.stderr
             else:
-                raise ValueError(
-                    f"Invalid stream option: Should be 'stdout', 'stderr', or a TextIO-like object; got {stream}")
+                raise ValueError(f'stream must be one of ("stdout", "stderr", TextIO-like), got {stream}')
+
         self.stream = stream
 
+    @property
+    def current_pbar(self) -> Optional[_ProgressBar]:
```
ravi-mosaicml, line 165: Curious do we need a public getter / setter here? If so, then can `_ProgressBar` be made public (it's a bit unconventional to have a public method return a private type...)

composer/loggers/progress_bar_logger.py
```diff
@@ -105,140 +128,178 @@ def __init__(
...
+        if self.is_train is None:
+            return None
+        return self.train_pbar if self.is_train else self.eval_pbar
+
+    @current_pbar.setter
+    def current_pbar(self, pbar: Optional[_ProgressBar]):
+        if self.is_train is None:
```
ravi-mosaicml, line 172: Since this is public, could this be a `ValueError` or a `RuntimeError`? Otherwise, if made private, could a standard `assert` statement be used?

composer/loggers/progress_bar_logger.py
```diff
@@ -105,140 +128,178 @@ def __init__(
...
+        This is often used to emit a pbar for each epoch, e.g.
+            Epoch     0 train 100%|█████████████████████████| 29/29
+            Epoch     1 train 100%|█████████████████████████| 29/29
+        """
+        position = 0 if is_train else 1
+        split = 'train' if is_train else 'val'
+
+        assert self.is_train is not None
+        if epoch_style:
+            assert state.dataloader_len is not None, 'dataloader_len should be set'
```
ravi-mosaicml, line 222: Can we be graceful for this case? I think tqdm supports progress bars for an unknown total length

ravi-mosaicml (approved):
Overall lgtm! See comments. Thanks for cleaning this up; the new implementation is much easier to follow.

> because terminals only support editing the last line (see last two examples above).
Hmm, I thought TQDM supported multiple progress bars? Curious what the underlying issue is here (if it's a hard fix, no need to do it in this pr, but please create a JIRA to track it. It feels a bit weird that training would be at the bottom under eval)

composer/loggers/progress_bar_logger.py
hanlint: `epoch_style` is also used during training when the `unit == TimeUnit.EPOCH`, so this is needed to switch that exception.

composer/loggers/progress_bar_logger.py
hanlint: To match the `tqdm.update` API as closed as possible, choose to have the default.

composer/loggers/progress_bar_logger.py
hanlint: Good point, made `current_pbar` private.

composer/loggers/progress_bar_logger.py
hanlint: done!

PR #1191, opened 2022-06-18
Fix interleaving train/eval progress bars

hanlint:
Before:
```
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.60it/s, metri
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.99it/s, metri
Epoch     0 train  34%|████████▌                | 10/29 [00:10<00:19,  1.01s/it, loss/
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 20.55it/s, metri
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.86it/s, metri
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.05it/s, metri
Epoch     1 train   3%|▊                        | 1/29 [00:09<04:29,  9.61s/it, loss/t
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.36it/s, metri
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.93it/s, metri
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.92it/s, metri
Epoch     2 train   7%|█▋                       | 2/29 [00:09<02:08,  4.75s/it, loss/t
Epoch     3 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.80it/s, metri
Epoch     3 val   100%|█████████████████████████| 10/10 [00:00<00:00, 22.03it/s, metri
Epoch     3 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.85it/s, metri
```

After:
```
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.68it/s, metri
Epoch     0 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.56it/s, metri
Epoch     0 train 100%|█████████████████████████| 29/29 [00:10<00:00,  2.80it/s, loss/
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 20.77it/s, metri
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.05it/s, metri
Epoch     1 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.61it/s, metri
Epoch     1 train 100%|█████████████████████████| 29/29 [00:09<00:00,  2.92it/s, loss/
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.87it/s, metri
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.52it/s, metri
Epoch     2 val   100%|█████████████████████████| 10/10 [00:00<00:00, 21.47it/s, metri
Epoch     2 train 100%|█████████████████████████| 29/29 [00:09<00:00,  2.94it/s, loss/
```

This is with `eval_interval=20ba`. Note: the train pbar for epoch 0 will show up after the eval pbar for epoch 0 (even though chronologically they are reversed). 

Depends on #1190

### DIFF
--- a/composer/loggers/progress_bar_logger.py
+++ b/composer/loggers/progress_bar_logger.py
@@ -179,7 +179,7 @@ class ProgressBarLogger(LoggerDestination):
             self.log_to_console(log_str)
 
     def log_to_console(self, log_str: str):
-        """ Logs to the console, avoiding interleaving with a progress bar"""
+        """Logs to the console, avoiding interleaving with a progress bar."""
         if self.current_pbar:
             # use tqdm.write to avoid interleaving
             self.current_pbar.pbar.write(log_str)
@@ -223,6 +223,7 @@ class ProgressBarLogger(LoggerDestination):
             self.current_pbar.update()
 
     def batch_end(self, state: State, logger: Logger) -> None:
+        self.is_train = True
         self._update()
 
     def eval_after_forward(self, state: State, logger: Logger) -> None:
--- a/composer/loggers/progress_bar_logger.py
+++ b/composer/loggers/progress_bar_logger.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import sys
 from dataclasses import asdict, dataclass
+from functools import wraps
 from typing import Any, Callable, Dict, List, Optional, TextIO, Union
 
 import tqdm.auto
@@ -21,8 +22,20 @@ __all__ = ['ProgressBarLogger']
 _IS_TRAIN_TO_KEYS_TO_LOG = {True: ['loss/train'], False: ['metrics/eval/Accuracy']}
 
 
+def rank_zero_only(fn: Callable) -> Callable:
+
+    @wraps(fn)
+    def wrapped_fn(*args: Any, **kwargs: Any) -> Optional[Any]:
+        if dist.get_local_rank() != 0:
+            return
+
+        return fn(*args, **kwargs)
+
+    return wrapped_fn
+
+
 @dataclass
-class _ProgressBarLoggerInstanceState:
+class _ProgressBarState:
     total: Optional[int]
     description: str
     position: int
@@ -31,9 +44,9 @@ class _ProgressBarLoggerInstanceState:
     epoch_metrics: Dict[str, Any]
 
 
-class _ProgressBarLoggerInstance:
+class _ProgressBar:
 
-    def __init__(self, file: TextIO, state: _ProgressBarLoggerInstanceState) -> None:
+    def __init__(self, file: TextIO, state: _ProgressBarState) -> None:
         self.state = state
         self.pbar = tqdm.auto.tqdm(
             total=state.total,
@@ -106,12 +119,17 @@ class ProgressBarLogger(LoggerDestination):
         stream: Union[str, TextIO] = sys.stderr,
     ) -> None:
         self.show_pbar = progress_bar
-        log_to_console = not progress_bar if log_to_console is None else log_to_console
-        if not log_to_console:
-            console_log_level = lambda state, ll: False
-        self.pbars: Dict[bool, _ProgressBarLoggerInstance] = {}
+        self.train_pbar: Optional[_ProgressBar] = None
+        self.eval_pbar: Optional[_ProgressBar] = None
         self.is_train: Optional[bool] = None
 
+        if log_to_console is None:
+            log_to_console = not progress_bar
+
+        if not log_to_console:
+            console_log_level = lambda state, ll: False
+
+        # create should_log callable based on console_log_level
         if isinstance(console_log_level, str):
             console_log_level = LogLevel(console_log_level)
         if isinstance(console_log_level, LogLevel):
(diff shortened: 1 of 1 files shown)


composer/loggers/progress_bar_logger.py
```diff
@@ -21,8 +22,20 @@
 _IS_TRAIN_TO_KEYS_TO_LOG = {True: ['loss/train'], False: ['metrics/eval/Accuracy']}
 
 
+def rank_zero_only(fn: Callable) -> Callable:
```
ravi-mosaicml, line 25: I do appreciate the simplicity of a decorator like this, but am concerned about a naive implementation, as it makes distributed errors much harder to debug in code like the following:

```python

def bar():
    dist.barrier()

@rank_zero_only
def foo():
   bar()
```
The above code snippet will hang, but it is unclear and hard to debug why. We had errors before like this, which is why I removed the `RankZeroCallback` class a while back.

That said, I do like the idea of having a rank zero context manager that causes all torch.dist calls to immediately error if used within context manager, and then we can replace all `if dist.get_global_rank() == 0` with a `with dist.global_rank_zero_only():`. I would prefer a context manager over a decorator, as it draws more visual attention to the code that it is being executed only on rank zero, but am also open to a function / class that could be used as both.

As for an implementation, it could look something like this:


```python
# composer/utils/dist.py

_DIST_ENABLED = True

@contextlib.contextmanager
def global_rank_zero_only(enabled: bool = True):
    global _DIST_ENABLED
    was_enabled = _DIST_ENABLED
    _DIST_ENABLED = enabled
    try:
        yield
    finally:
        _DIST_ENABLED = was_enabled

_DIST_DISABLED_ERROR_MSG = "Distributed functions cannot be called when used inside `global_rank_zero_only()`"

# Then, call dist wrappers (except the getters) would need to have something like this:
def barrier():
    global _DIST_ENABLED
    if not _DIST_ENABLED:
        raise RuntimeError(_DIST_DISABLED_ERROR_MSG)
    ...
```

Then, caller code could look like:

```python
def something():
    with global_rank_zero_only():
        print("Hello!")
```

Anyways, I think adding this decorator / context manager and replacing the existing `if dist.get_global_rank() == 0` checks is outside the scope of this PR.

composer/loggers/progress_bar_logger.py
```diff
@@ -123,47 +141,65 @@ def should_log(state: State, log_level: LogLevel, console_log_level: LogLevel =
...
-                stream = sys.stdout
-            elif stream.lower() == 'stderr':
-                stream = sys.stderr
-            else:
-                raise ValueError(
-                    f"Invalid stream option: Should be 'stdout', 'stderr', or a TextIO-like object; got {stream}")
+            stream = {
+                'stdout': sys.stdout,
+                'stderr': sys.stderr,
+            }[stream.lower()]
```
ravi-mosaicml, line 150: nit: Would prefer that a valueerror is raised on an invalid stream option (I agree with your comment: https://mosaicml.slack.com/archives/C0205HZTT9P/p1655512719555359?thread_ts=1655172522.700169&cid=C0205HZTT9P)

composer/loggers/progress_bar_logger.py
```diff
@@ -123,47 +141,65 @@ def should_log(state: State, log_level: LogLevel, console_log_level: LogLevel =
...
+        return self.train_pbar if self.is_train else self.eval_pbar
 
-        if not self.should_log(state, log_level):
-            return
-        data_str = format_log_data_value(data)
-        log_str = f'[{log_level.name}][batch={int(state.timestamp.batch)}]: {data_str}'
+    @current_pbar.setter
+    def current_pbar(self, pbar: Optional[_ProgressBar]):
+        if self.is_train is None:
+            raise AssertionError('Cannot set pbar if self.is_train is not set.')
```
ravi-mosaicml, line 162: Since this is a public setter, I think it should be a value error.

```suggestion
            raise RuntimeError('Cannot set pbar if self.is_train is not set.')
```

That said, do we need a public getter and setter for this?

composer/loggers/progress_bar_logger.py
```diff
@@ -123,47 +141,65 @@ def should_log(state: State, log_level: LogLevel, console_log_level: LogLevel =
...
+    def log_to_console(self, log_str: str):
+        """Logs to the console, avoiding interleaving with a progress bar."""
+        if self.current_pbar:
+            # use tqdm.write to avoid interleaving
+            self.current_pbar.pbar.write(log_str)
         else:
             # write directly to self.stream; no active progress bar
             print(log_str, file=self.stream, flush=True)
 
+    @rank_zero_only
```
ravi-mosaicml, line 190: On line 171, it is local rank zero -- here, it is global rank zero. Would it make sense for progress bars to show up on all local rank zeros? (Also, `rank_zero` is a bit of a confusing name because of this -- if you do keep the decorator approach, could it be called `_local_rank_zero_only`?)

hanlint:
Closing for #1190

PR #1192, opened 2022-06-20
Adding Gated Linear Units as an algorithm

moinnadeem:
Implements [GLU Variants Improve Transformer](https://arxiv.org/abs/2002.05202) for BERT Models


moinnadeem:
NOTE: Due to CO-382, GLU has been xfailed on `test_algorithm_resumption` due to issues with second-order optimizers. We should resolve this xfail before merging.

moinnadeem:
Ignore the algorithm card, I can write up the proper documentation in a bit!

composer/algorithms/gated_linear_units/gated_linear_unit_layers.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Callable
+
+import torch
+
+
+class DummyBERTIntermediateOutput(torch.nn.Module):
```
hanlint, line 9: Maybe a bit more descriptive..e.g. `IdentityLayer`, since its not specific to BERT really.

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,159 @@
...
+from composer.algorithms.warnings import NoEffectWarning
+from composer.core import Algorithm, Event, State
+from composer.loggers import Logger
+from composer.models import BERTModel
+from composer.utils import check_if_transformers_installed, module_surgery
+
+log = logging.getLogger(__name__)
+
+
+def from_BertOutput(layer: torch.nn.Module,
```
hanlint, line 30:
```suggestion
def from_BertOutput(layer: BertOutput,
```

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,159 @@
...
+                    module_index: int,
+                    act_fn: Callable,
+                    wi_0_bias: bool = False,
+                    wi_1_bias: bool = False) -> BERTGatedOutput:
+    """Defines a replacement policy from a `transformers.models.bert.modeling_bert.BertOutput` to a `composer.algorithms.gated_linear_units.gated_linear_unit_layers.BERTGatedOutput`"""
+    assert isinstance(
```
hanlint, line 36: This assert isn't needed once with the change above

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,159 @@
...
+    return BERTGatedOutput(d_embed=layer.dense.out_features,
+                           d_ff=layer.dense.in_features,
+                           dropout_rate=layer.dropout.p,
+                           act_fn=act_fn,
+                           layernorm_eps=layer.LayerNorm.eps,
+                           wi_0_bias=wi_0_bias,
+                           wi_1_bias=wi_1_bias)
+
+
+def from_BertIntermediate(layer: torch.nn.Module, module_index: int) -> DummyBERTIntermediateOutput:
```
hanlint, line 48:
```suggestion
def from_BertIntermediate(layer: BertIntermediate, module_index: int) -> DummyBERTIntermediateOutput:
```

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -33,7 +34,8 @@ def test_algorithm_resumption(
     if alg_cls is LayerFreezing:
         pytest.xfail('Known issues')
 
-    if alg_cls in (SAM, SqueezeExcite, StochasticDepth, Factorize, BlurPool):
+    # TODO (Moin): temporary xfail on this test to see if the rest pass. DO NOT MERGE with GLU xfailed.
```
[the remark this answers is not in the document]
moinnadeem: @hanlint This is related to the "we can't load a surgery algorithm checkpoints with Adam" -- I was hoping the fix would be in by the time we were merging this PR

tests/algorithms/test_gated_linear_units.py
```diff
@@ -0,0 +1,50 @@
...
+
+from composer.algorithms.gated_linear_units import GatedLinearUnits, apply_gated_linear_units
+from composer.algorithms.gated_linear_units.gated_linear_unit_layers import BERTGatedOutput
+from composer.core.event import Event
+from composer.loggers import Logger
+from composer.models import BERTModel
+from tests.fixtures.synthetic_hf_state import make_dataset_configs, synthetic_hf_state_maker
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
moinnadeem: Hmm, good question. I think that would be a good idea -- I can try to get that in today.

tests/algorithms/test_algorithm_resumption.py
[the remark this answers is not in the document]
hanlint: That seem unlikely, OK to keep the xfail for now

composer/algorithms/gated_linear_units/gated_linear_units.py
moinnadeem: Wait, for my education, does typing perform runtime checks? I thought it was just static analysis?

hanlint: Just static analysis. But this is already guaranteed by the replacement policy below, right? If someone wants to use this layer in some odd context, or where they wrote their own layer that should be replaced, I think we should allow it.

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,159 @@
...
+            model parameters.
+        act_fn (Callable, optional): Optionally, the activation function to use. If ``None``, the algorithm will
+            use the existing activation function in the model.
+        wi_0_bias (bool, optional): Whether to use biases in the linear layers within the GLU. Default: ``False``.
+        wi_1_bias (bool, optional): Whether to use biases in the linear layers within the GLU. Default: ``False``.
+    """
+    check_if_transformers_installed(TRANSFORMERS_INSTALLED)
+
+    # ensure that the model is an instance of a BERTModel, since our replacement policy is only defined for BERTs
+    if not isinstance(model, BERTModel):
```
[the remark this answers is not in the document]
moinnadeem: It would -- but we need to make sure that their BertIntermediate and BertOutput have the right overall structure (ie. it could be dangerous if their ATTN network does an operation that we consider part of our `BertGatedOutput`. 

For instance, HF had defined this class called [BertIntermediate](https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/modeling_bert.py#L437) that has what we would consider today as part of the FF network. 

This heavy-handed approach is to make sure that we can provide some sanity about the surgery-ified model, and that the composite model doesn't have characteristics that don't make sense for training. Part of that is knowing what their ATTN network looks like, so that's why I'm reluctant to support a random model.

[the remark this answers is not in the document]
hanlint: OK sounds good.

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,165 @@
...
+    if act_fn is None:
+        raise ValueError(
+            'Could not find an existing activation function to use, and no custom activation function was provided.')
+
+    # prepare the replacement policy and perform replacement
+    policy: Dict[Type[torch.nn.Module], module_surgery.ReplacementFunction] = {
+        BertIntermediate:
+            from_BertIntermediate,
+        BertOutput:
+            partial(from_BertOutput,
```
ravi-mosaicml, line 102: nit: For typing, would it be possible to define a sub-function here to bind these parameters, rather than use `functools.partial`? I don't believe pyright can validate functools.partial argument names or types...(I might be mistaken given the new ParamSpec in python 3.10, in which case ignore this comment)

composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class GatedLinearUnits(Algorithm):
+    """Replaces all instances of Linear layers in the feed-forward subnetwork with a `Gated Linear Unit <https://arxiv.org/abs/2002.05202>`_.
+    THe Gated Linear Units provide a more expressive form for the same number of parameters, and a slight degredation to throughput.
+
+    Runs on :attr:`~composer.core.event.Event.INIT`, so it can swap the Linear layers in the FFN for GLUs before the model is DDP wrapped.
+
+    Args:
+        act_fn (Callable, optional): Optionally, the activation function to use. If ``None``, the algorithm will
```
ravi-mosaicml, line 122: Same thing re: being more specific.

tests/algorithms/test_gated_linear_units.py
```diff
@@ -0,0 +1,87 @@
...
+
+def _layernorm(input_tensor, layernorm_eps):
+    mean = torch.mean(input_tensor, dim=-1, keepdim=True)
+    var = torch.square(input_tensor - mean).mean(dim=-1, keepdim=True)
+    return (input_tensor - mean) / torch.sqrt(var + layernorm_eps)
+
+
+@pytest.mark.parametrize('batch_size', [1])
+@pytest.mark.parametrize('seq_length', [128, 512])
+@pytest.mark.parametrize('d_embed', [768])
```
ravi-mosaicml, line 26: Nit: For parameters where there is just one value, can these be hardcoded?

tests/algorithms/test_gated_linear_units.py
```diff
@@ -0,0 +1,87 @@
...
+    hidden_states = hidden_states[1:]
+    manual_output = torch.matmul(hidden_states, gated_ff.gated_layer.weight.transpose(0, 1))
+    manual_output = act_fn(manual_output)
+    manual_output = manual_output * torch.matmul(hidden_states, gated_ff.non_gated_layer.weight.transpose(0, 1))
+    manual_output = torch.matmul(manual_output, gated_ff.wo.weight.transpose(0, 1)) + gated_ff.wo.bias
+    manual_output = _layernorm(manual_output + residual_connection, layernorm_eps)
+    assert torch.allclose(manual_output, model_output)
+
+
+@pytest.fixture()
```
ravi-mosaicml, line 53: These tests seem quite similar to other sugary tests -- curious if we could refactor them into shared helpers? Feel free to ignore if it is too much work / not easily doable.

ravi-mosaicml (approved):
Overall LGTM! See comments. Please double-check docstrings and type annotations -- didn't review them all.

tests/algorithms/test_gated_linear_units.py
moinnadeem: Thoughts on leaving this out of the scope of the PR and creating a JIRA? Feel free to assign to me -- I'm down to do it, but there's other priorities this week before the release is cut.

hanlint: Suggest we defer this until we define a more solid Surgery API

composer/algorithms/gated_linear_units/gated_linear_unit_layers.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    def forward(self, hidden_states):
+        """
+        A no-op forward that just returns the input.
+        """
+        return hidden_states
+
+
+class BERTGatedFFOutput(torch.nn.Module):
```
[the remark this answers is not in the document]
moinnadeem: It needs to because we make assumptions on the structure of the ATTN map, so it can't be used generically.

PR #1194, opened 2022-06-22
Add Tensorboard Logger

eracah:
CO-158

- [x] Add TensorboardLogger class
- [x] Local E2E test
- [x] Add docstrings
- [x] Add tests
- [x] Add run name compatibilty
- [x] Add artifact logging
- [x] Add flushing
~~Add hparams/config logging~~ (deferred to V2 of the logger)
- [x] Ravi and Hanlin's Comments
- [x] E2E Test with ResNet50:
Results: [tensorboard](https://tensorboard.dev/experiment/a2E7WXjhReKwhXrLD870Ow/#scalars&_smoothingWeight=0) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/15ermoz5)
- [x] E2E Test with Moin's BERT setup and compare wandb with tensorboard
Results: [tensorboard](https://tensorboard.dev/experiment/bHbWZjg1R6OPKUKSNSm0Ww/#scalars) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/3awal195)
- [x] Run full E2E bert production run as a test
Results: [tensorboard](https://tensorboard.dev/experiment/qim9YmEsQxCJE0GhwPCyaw/#scalars) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/1auufdfs)
- [x] E2E test with `speed_monitor` turned on
Results: [tensorboard](https://tensorboard.dev/experiment/2wkXxrAPSR2uphmOarhNcw/#scalars&_smoothingWeight=0) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/1ije9azo)
- [x] Test logging Tensorboard logs to S3 through `ObjectStoreLogger` works
- [x] Ravi's last comments


composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,32 @@
...
+class TensorboardLogger(LoggerDestination):
+    def __init__(self,
+                 log_dir: str=None,
+                 run_name: Optional[str]= None,
+                 rank_zero_only: bool = True):
+        
+        self.run_name = run_name if run_name is not None else ''
+        self.log_dir = log_dir
+        self.writer = SummaryWriter(log_dir=self.log_dir, filename_suffix=self.run_name)
+        self._enabled = (not rank_zero_only) or dist.get_global_rank() == 0
```
hanlint, line 24: `dist` will not have been enabled until later, so safer to do:
```
self.rank_zero_only = rank_zero_only
```

and then, down in `log_data`, do:
```
if (not rank_zero_only) or dist.get_global_rank() == 0:
```
```

Comment (eracah) [reply composer/loggers/tensorboard_logger.py]: oh good catch! done here: 
172c29b

File: composer/loggers/tensorboard_logger.py [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+        self.writer = SummaryWriter(log_dir=self.log_dir, filename_suffix=self.run_name)
+        self.rank_zero_only = rank_zero_only
+        self.run_name = run_name
+
+    def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
+        if (not self.rank_zero_only) or dist.get_global_rank() == 0:
+            for tag, data_point in data.items():
+                if isinstance(data_point, str):
+                    self.writer.add_text(tag, data_point)
+                self.writer.add_scalar(tag, data_point, global_step=int(state.timestamp.batch))
```
hanlint, line 33: since the data can be `Any`, do we want to check if its a scalar, and pass otherwise?

eracah:
@ravi-mosaicml, I am seeing some weird flushing/rank_zero_only behavior as of 
d786089

See below for a summary:

### num_ranks = 1
Screen Shot 2022-06-22 at 8 41 02 PM

Looks good.

### num_ranks=2, rank_zero_only=False
Screen Shot 2022-06-22 at 8 41 32 PM

Some weird plotting, where we have duplication of the same global step for multiple losses.

### num_ranks=2, rank_zero only=True
Screen Shot 2022-06-22 at 8 41 57 PM

Only gets a small fraction of the iterations.

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,103 @@
...
+        self.rank_zero_only = rank_zero_only
+        self.log_level = log_level
+
+    def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
+        del log_level
+
+        if (not self.rank_zero_only) or dist.get_global_rank() == 0:
+            for tag, data_point in data.items():
+                if isinstance(data_point, str):
+                    return
```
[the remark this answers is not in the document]
eracah: oh woow good catch. fixed here: 
53f8043

composer/trainer/trainer.py
```diff
@@ -1412,7 +1412,7 @@ def _accumulate_time_across_ranks(
     def _train_loop(self) -> None:
         """Run training for the specified number of epochs and log results."""
         # print training start
-        self.logger.data_fit({'trainer/algorithms': [str(algo) for algo in self.state.algorithms]})
+        self.logger.data_fit({algo.__class__.__name__: 1 for algo in self.state.algorithms})
```
ravi-mosaicml, line 1415: I presume this is because tensorboard unable to handle lists of strings?

tests/loggers/test_tensorboard_logger.py
```diff
@@ -0,0 +1,26 @@
...
+
+from composer.loggers import TensorboardLogger
+from tests.fixtures import dummy_fixtures
+
+# To satisfy pyright.
+dummy_state = dummy_fixtures.dummy_state
+composer_trainer_hparams = dummy_fixtures.composer_trainer_hparams
+
+
+def test_tensorboard_logger_trainer(monkeypatch: pytest.MonkeyPatch, composer_trainer_hparams):
```
ravi-mosaicml, line 16: Curious if we need this test? I think we already have coverage for this via https://github.com/mosaicml/composer/blob/16622c9457874f4f3811287fe933b0ec58107fc7/tests/callbacks/test_callbacks.py#L42 and https://github.com/mosaicml/composer/blob/16622c9457874f4f3811287fe933b0ec58107fc7/tests/callbacks/test_callback_hparams.py#L19

ravi-mosaicml (commented):
Thanks for working through this; see comments! Would also like to see a sample output and compare it against wandb

eracah:
> Thanks for working through this; see comments! Would also like to see a sample output and compare it against wandb

Here's a sample from resnet50: [tensorboard](https://tensorboard.dev/experiment/a2E7WXjhReKwhXrLD870Ow/#scalars&_smoothingWeight=0) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/15ermoz5)

composer/trainer/trainer.py
eracah: yes. plus this makes it more readable for any logger (it actually displays the class name and not the object name with memory address)

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,106 @@
...
+    def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
+        del log_level
+
+        if (not self.rank_zero_only) or dist.get_global_rank() == 0:
+            for tag, data_point in data.items():
+                if isinstance(data_point, str): # Will error out with weird caffe2 import error.
+                    continue
+                try:
+                    self.writer.add_scalar(tag, data_point, global_step=int(state.timestamp.batch))
+                except NotImplementedError:
```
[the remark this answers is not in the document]
eracah: If input is not scalar, string, numpy array nor tensor. See this:
```python
def make_np(x):
   if isinstance(x, np.ndarray):
       return x
   if isinstance(x, str):  # Caffe2 will pass name of blob(s) to fetch
       return _prepare_caffe2(x)
   if np.isscalar(x):
       return np.array([x])
   if isinstance(x, torch.Tensor):
       return _prepare_pytorch(x)
   raise NotImplementedError(
       'Got {}, but numpy array, torch tensor, or caffe2 blob name are expected.'.format(type(x)))```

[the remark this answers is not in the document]
Comment (eracah) [reply composer/loggers/tensorboard_logger.py:74]: I don't think it makes sense to pass through. It wouldn't make sense to crash someone's job because of the type of something being logged

File: composer/loggers/tensorboard_logger.py (hunk 2 above)
Comment (eracah) [reply composer/loggers/tensorboard_logger.py]: done

File: composer/loggers/tensorboard_logger.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (eracah) [reply composer/loggers/tensorboard_logger.py:74]: See this comment: https://github.com/mosaicml/composer/pull/1194#discussion_r904492634

File: composer/loggers/tensorboard_logger.py [hunk 7]
```diff
@@ -0,0 +1,103 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Log to `Tensorboard <https://www.tensorflow.org/tensorboard/get_started#:~:text=TensorBoard%20is%20a%20tool%20for,dimensional%20space%2C%20and%20much%20more./>`_."""
```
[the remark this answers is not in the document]
eracah: done

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,106 @@
...
+                 log_dir: Optional[str] = None,
+                 artifact_name: Optional[str] = None,
+                 flush_interval: int = 100,
+                 rank_zero_only: bool = True,
+                 log_level: LogLevel = LogLevel.BATCH):
+
+        self.artifact_name = artifact_name
+        self.flush_interval = flush_interval
+        self.log_dir = log_dir
+        self.writer: SummaryWriter
```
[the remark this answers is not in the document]
eracah: done

tests/loggers/test_tensorboard_logger.py
eracah: oh ok didn't see this test. ok removed

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,106 @@
...
+from composer.utils import dist
+
+__all__ = ['TensorboardLogger']
+
+
+class TensorboardLogger(LoggerDestination):
+    """Log to `Tensorboard <https://www.tensorflow.org/tensorboard/>`_.
+
+    Args:
+        log_dir (str, optional): The directory to store the logs. If
```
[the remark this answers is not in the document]
eracah: `run_name` has to be the child directory because otherwise the naming inside tensorboard gets messed up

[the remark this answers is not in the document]
eracah: changed to f'tensorboard/{state.run_name}'

moinnadeem (commented):
The tensorboard looks good to me! 
Two questions:
1) For our sanity, could we kick off a full pretraining run? I'm curious if any of the UX would change with more samples
2) I'm pretty sure this should be fine, but could we do a run with the `speed_monitor`? I want to see how that looks, but it is not blocking at all.

This has my BERT approval, but I would defer a full PR approval to @ravi-mosaicml and @hanlint

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,103 @@
...
+__all__ = ['TensorboardLogger']
+
+
+class TensorboardLogger(LoggerDestination):
+    """Log to `Tensorboard <https://www.tensorflow.org/tensorboard/>`_.
+
+    Args:
+        log_dir (str, optional): The path to the directory to put these logs in. If it is
+            None the logs will be placed in `./runs/{month}{day}{HH-MM-SS}_{device_name}.local
+        artifact_name (str, optional): Format string for the logfile's artifact name.
```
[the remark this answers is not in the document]
eracah: Ok I gave it some thought and I decided to reorganize the arguments a little. Here's what I did:

1. expose `run_name` as an argument to the user.
2. expose `log_dir` as the directory that stores all the other directories named after `run_name`.
3. Set the suffix of each log file will be the `run_name`
4. `artifact_name` is abstracted away from the user because there is pretty much no use case where they need to set this and it adds confusion. We then grab the stem of the log file path and set that as the `artifact_name`

Please refer to the docstring for more clarification

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,106 @@
...
+
+            The same format variables for ``filename`` are available. Setting this parameter to ``None``
+            (the default) will use the same format string as ``filename``. It is sometimes helpful to deviate
+            from this default. For example, when ``filename`` contains an absolute path, it is recommended to
+            set this parameter explicitly, so the absolute path does not appear in any artifact stores.
+
+            Leading slashes (``'/'``) will be stripped.
+
+            Default: ``None`` (which uses the same format string as ``filename``)
+        flush_interval (int, optional): How frequently to flush the log to the file,
```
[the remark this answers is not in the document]
eracah: I think this is unnecessary for V1. It runs at `eval_start` and `eval_end`, which I think is sufficient.

composer/loggers/tensorboard_logger.py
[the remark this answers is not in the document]
eracah: The log file path is configured by `run_name` and `log_dir` like so: `{log_dir}/{tensorboard_log_file_prefix_that_cannot_be_modified}{run_name}`

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,103 @@
...
+        self.log_dir = log_dir
+        self.writer: SummaryWriter
+        self.should_eval: Callable
+        self.rank_zero_only = rank_zero_only
+        self.log_level = log_level
+
+    def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
+        del log_level
+
+        if (not self.rank_zero_only) or dist.get_global_rank() == 0:
```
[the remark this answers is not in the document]
eracah: Fixed here: e9b7f5d.
DeMorgan's law ftw.

eracah:
Thanks, @moinnadeem!

> 1. For our sanity, could we kick off a full pretraining run? I'm curious if any of the UX would change with more samples

Ok I will fire one off. I'll let you know when it's done.

> 2. I'm pretty sure this should be fine, but could we do a run with the `speed_monitor`? I want to see how that looks, but it is not blocking at all.

Here you go: [tensorboard](https://tensorboard.dev/experiment/2wkXxrAPSR2uphmOarhNcw/#scalars&_smoothingWeight=0) and [wandb](https://wandb.ai/mosaic-ml/tb-test/runs/1ije9azo)

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,102 @@
...
+            try:
+                self.writer.add_scalar(tag, data_point, global_step=int(state.timestamp.batch))
+            except NotImplementedError:
+                pass
+
+    def init(self, state: State, logger: Logger) -> None:
+        if self.run_name is None:
+            self.run_name = f'{state.run_name}'
+        if self.log_dir is None:
+            self.log_dir = 'tensorboard_logs'
```
ravi-mosaicml, line 74:
```suggestion
            self.log_dir = os.path.join(self.run_name, 'tesnsorboard_logs')
```

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,102 @@
...
+        summary_writer_log_dir = Path(self.log_dir) / self.run_name
```
ravi-mosaicml, line 75: Would prefer not to subfolder the logdir by the run name if the user specifies the run dir explicitly -- seems like unintuitive behavior

composer/loggers/tensorboard_logger.py
ravi-mosaicml: This makes sense, but can you add a comment in the code explaining why we're passing on the not implemented exceptions?

tests/loggers/test_tensorboard_logger.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
```
ravi-mosaicml, line 2: Can you delete this file if it's not being used?

ravi-mosaicml (commented):
Overall looking good! Just a few comments

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def fit_end(self, state: State, logger: Logger) -> None:
+        # Flush the file on fit_end, in case if was not flushed on epoch_end and the trainer is re-used
+        # (which would defer when `self.close()` would be invoked)
+        self._flush(logger)
+
+    def _flush(self, logger: Logger):
+        self.writer.flush()
+        logger.file_artifact(LogLevel.FIT,
+                             artifact_name=self.run_name,
```
ravi-mosaicml, line 100: Can there be an artifact name prefix config variable, that defaults to '{run_name}/tensorboard_logs'?
The artifact name here should include the global rank of the process, so each rank will write to its own file.

composer/loggers/tensorboard_logger.py
eracah: done

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,102 @@
...
+                 run_name: Optional[str] = None,
+                 log_dir: Optional[str] = None,
+                 flush_interval: int = 100,
+                 rank_zero_only: bool = True):
+
+        self.run_name = run_name
+        self.log_dir = log_dir
+        self.flush_interval = flush_interval
+        self.rank_zero_only = rank_zero_only
+        self.artifact_name: str = None
```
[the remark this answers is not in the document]
eracah: good catch! whoops! fixed

composer/loggers/tensorboard_logger.py
eracah: it isn't necessary to add the global rank since the `file_artifact` function does that for you here: https://github.com/mosaicml/composer/blob/dev/composer/loggers/logger.py#L123

composer/loggers/tensorboard_logger.py
eracah: The way it is currently done is done for a reason: The name of the stem of the file path will be what the run is called in the tensorboard gui. You **must** have the stem of the log_dir be named by the run name or the run_name won't show up in the gui. If you set the log_dir to `{run_name}/{tensorboard_logs}` then the run name in the gui will be `"tensorboard_logs"`, which we don't want.

composer/loggers/tensorboard_logger.py
eracah: As mentioned below if we set the log_dir to `os.path.join(self.run_name, 'tensorboard_logs')` then the run name in the gui will be "tensorboard_logs". Having the log_dir be `f'{run_name}/{tensorboard_logs}/{run_name}'` is also undesirable because the user wants to compare various runs. If we set `summary_writer_log_dir = Path(self.log_dir) / self.run_name` then we get all runs saved inside the same director, self.log_dir and they are named properly.

eracah:
@ravi-mosaicml, I am sorry I failed to clarify how Tensorboard saves it's logs. I'll try to clarify below to communicate why some of my seemingly strange design decisions are actually very purposeful:

Let's say you specify `SummaryWriter(log_dir='foo/bar/baz')`:
1. Then, all log files for the current run will be saved to `'foo/bar/baz'`
2. If you want to view that run's logs, you point the tensorboard viewer at `'foo/bar'` (not `'foo/bar/baz'`!) like so: `tensorboard --logdir='foo/bar'`. Yes you must specify something different to `logdir` as you do for `log_dir`. It's very confusing, but that is out of our control.
3. Now, when you open the Tensorboard gui, the name of the run will be `'baz'`
4. If you want to see metrics from the `baz` run compared to metrics from another run (let's call it `'baz2'`), then you must save your log files for the the other run to `'foo/bar/baz2'`. That way when you point the tensorboard viewer at `'foo/bar'`, the logs from `baz` and `baz2` show up in the viewer because both subfolders are saved to `'foo/bar/'`.

composer/loggers/tensorboard_logger.py
eracah: that is a good point that I should probably name the `artifact_name=Path(self.writer.file_writer.event_writer._file_name).stem` or something because it is an actual tensorboard log file that gets saved as an artifact

eracah: Actually, Tensorboard won't read the log file unless it is named properly so calling it `Path(self.writer.file_writer.event_writer._file_name).stem` is essential. Fixed here: 
42426e6

eracah: update: I use the formatted string like so: `'tensorboard_logs/{run_name}/events.out.tfevents-{run_name}-{rank}'`

eracah:
Okay, I am finished on my side. All the tasks are finished and E2E tests have passed.

moinnadeem:
I reviewed the Tensorboard links earlier today, and am proud to say that Evan has done a great job!

All of the most important metrics [loss and MaskedAccuracy] are still accessible. Unfortunately, some of our metrics such as `wall_clock/train` and `samples/step` don't look great on [Tensorboard](https://tensorboard.dev/experiment/qim9YmEsQxCJE0GhwPCyaw/#scalars), but that is because of the fact that we might have slightly overfit to WandB and need to configure tensorboard in greater detail. WandB also just has better defaults than Tensorboard. For an example, look at the Y axis on `samples/step`.
image

Overall, this PR conveys all of the right information that I would want from a training run, and I think further improvements would be out of scope. LGTM from me, and let's see if setting axis scales and other items are possible in a future PR?

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,118 @@
...
+            except NotImplementedError:
+                pass
+
+    def init(self, state: State, logger: Logger) -> None:
+        if self.run_name is None:
+            self.run_name = state.run_name
+        if self.log_dir is None:
+            self.log_dir = 'tensorboard_logs'
+        summary_writer_log_dir = Path(self.log_dir) / self.run_name
+        self.writer = SummaryWriter(log_dir=summary_writer_log_dir, filename_suffix=self.run_name)
```
ravi-mosaicml, line 87: Curious why do we need `filename_suffix=self.run_name`?

In addition, is it possible to disable automatic flushing (since we do our own in self._flush)? Not sure if the flush interval could be set to a sentenal like -1, or whether we should just make it really big (like 2^63 - 1 -- assuming it's backed by a 64 bit int in tensorboard).

Mainly concerned if an automatic time-based flush coincides when the artifact logger is writing to the trace file when its copying it, resulting in a corrupted trace.

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,118 @@
...
+        # Flush the file on fit_end, in case if was not flushed on epoch_end and the trainer is re-used
+        # (which would defer when `self.close()` would be invoked)
+        self._flush(logger)
+
+    def _flush(self, logger: Logger):
+        # To avoid empty log artifacts for each rank.
+        if self.rank_zero_only and dist.get_global_rank() != 0:
+            return
+        self.writer.flush()
+        file_path = self.log_dir  # Initialize for pyright.
```
ravi-mosaicml, line 109: nit instead of the initialize and if check, add an assert is not None

```suggestion
        assert self.writer.file_writer is not None
        file_path = self.writer.file_writer.event_writer._file_name
```

ravi-mosaicml (approved):
LGTM; thanks for working through this!

See comments re: run name, flush interval, and a few other nits.

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Log to `Tensorboard <https://www.tensorflow.org/tensorboard/>`_."""
+
+from pathlib import Path
+from typing import Any, Dict, Optional
+
+from torch.utils.tensorboard import SummaryWriter
```
ravi-mosaicml, line 9: This will need to be a conditional import to get the dev tests to pass

ravi-mosaicml:
For the tests, the TensorBoardLogger will also need any entry in here for the skipif if tensorboard is not installed: https://github.com/mosaicml/composer/blob/16622c9457874f4f3811287fe933b0ec58107fc7/tests/callbacks/callback_settings.py#L86

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,118 @@
...
+                 run_name: Optional[str] = None,
+                 log_dir: Optional[str] = None,
+                 flush_interval: int = 100,
+                 rank_zero_only: bool = True):
+
+        self.run_name = run_name
+        self.log_dir = log_dir
+        self.flush_interval = flush_interval
+        self.rank_zero_only = rank_zero_only
+        self.writer: SummaryWriter = SummaryWriter()
```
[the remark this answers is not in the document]
ravi-mosaicml: (You may need `assert self.writer is not None` checks throughout the code)

composer/loggers/tensorboard_logger.py
eracah: It is explained a bit in the docstring, but I can put another more local comment

composer/loggers/tensorboard_logger.py
```diff
@@ -0,0 +1,118 @@
...
+    If you are accessing your logs from a cloud bucket, like S3, they will be
+    in `{your_bucket_name}/tensorboard_logs/{run_name}` with names like
+    `events.out.tfevents-{run_name}-{rank}`.
+
+    If you are accessing your logs locally (from wherever you are running composer), the logs
+    will be in the relative path: `tensorboard_logs/{run_name}` with names starting with
+    `events.out.tfevents.*`
+
+    Args:
+        run_name (str, optional): Tensorboard run name.
```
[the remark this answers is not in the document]
eracah: > I reviewed the Tensorboard links earlier today, and am proud to say that Evan has done a great job!

Thanks Moin! And thank you for all your help and putting up with me pestering you on slack.

> LGTM from me, and let's see if setting axis scales and other items are possible in a future PR?

Sounds good. I added that to the [jira for V2](https://mosaicml.atlassian.net/browse/CO-583)

composer/loggers/tensorboard_logger.py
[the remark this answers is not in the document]
eracah: done

composer/loggers/tensorboard_logger.py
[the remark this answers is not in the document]
eracah: thanks! done

composer/loggers/tensorboard_logger.py
eracah: > Curious why do we need `filename_suffix=self.run_name`?

good point. we don't need it. removed.


> In addition, is it possible to disable automatic flushing (since we do our own in self._flush)? Not sure if the flush interval could be set to a sentenal like -1, or whether we should just make it really big (like 2^63 - 1 -- assuming it's backed by a 64 bit int in tensorboard).
> 
> Mainly concerned if an automatic time-based flush coincides when the artifact logger is writing to the trace file when its copying it, resulting in a corrupted trace.

Don't think we can prevent the [first flush](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/summary/writer/event_file_writer.py#L207-L208), but I set `flush_secs` to `365 * 3600 * 24` (number of seconds in year). I don't think we will have any jobs running as long as a year!

composer/loggers/tensorboard_logger.py
eracah: done

composer/loggers/tensorboard_logger.py
eracah: Good call. I added some local comments

composer/loggers/tensorboard_logger.py
eracah: good catch. done

eracah:
> For the tests, the TensorBoardLogger will also need any entry in here for the skipif if tensorboard is not installed:
> 
> https://github.com/mosaicml/composer/blob/16622c9457874f4f3811287fe933b0ec58107fc7/tests/callbacks/callback_settings.py#L86

Done. Good call. Didn't know about that

composer/loggers/tensorboard_logger.py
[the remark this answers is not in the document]
ravi-mosaicml: The arg is still here...is it pushed?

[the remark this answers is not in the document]
eracah: ooh good catch. I removed the arg, but not the docstring. fixed here: 
0830416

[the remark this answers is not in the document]
eracah: are there any tests/pre-commits that can catch these types of mistakes?

PR #1195, opened 2022-06-22
Fix the Composer Launch Script for the Composer Dockerimage; Default `nproc = torch.cuda.device_count()` if not specified via env

ravi-mosaicml:
In the composer dockerimage, since the `composer` package is pip installed, the examples -- including the yaml entrypoint -- is not included. To fix, this PR moves the composer entrypoint into `composer.trainer.trainer_hparams` and  binds it to the `composer_train` CLI command.

So that `composer_train` can be used via the composer launcher script, this PR also adds in "command mode", where the user can pass an arbitrary executable that should be launched by the launcher script.

Put together, the user can now run when using the composer dockerimage:

```shell
composer -c composer_train -f composer/yamls/models/classify_mnist_cpu.yaml --train_dataset.mnist.datadir ~/datasets --val_dataset.mnist.datadir ~/datasets
```

Coming along for the ride:
* A fix in the docstrings that broke `composer_train --help`. Argparse treats help strings as python format strings, so all `%` will need to be escaped before passing it to argparse. This is a yahp bug which will be fixed later, but for now, just fixing the one offending docstring.
* Default `nproc` to `torch.cuda.device_count` if the env variable is not specified.


composer/trainer/trainer.py
```diff
@@ -612,7 +612,7 @@ class Trainer:
             .. note:: This is implemented by taking the batch yielded by the ``train_dataloader`` and splitting
                 it into ``grad_accum`` sections. Each section is of size ``train_dataloader // grad_accum``.
                 If the batch size of the dataloader is not divisible by ``grad_accum``,
-                then the last section will be of size ``batch_size % grad_accum``.
+                then the last section will be of size ``batch_size mod grad_accum``.
```
jbloxham, line 615: I don't think this is more correct. `mod` isn't really an operator in its mathematical sense. `%` is, and programmers know what it means.

composer/cli/launcher.py
```diff
@@ -268,6 +289,7 @@ def _launch_processes(
                 NODE_RANK=str(node_rank),
                 MASTER_ADDR=master_addr,
                 MASTER_PORT=str(master_port),
+                PYTHONUNBUFFERED='1',
```
jbloxham, line 292: What is this?

composer/trainer/trainer.py
jbloxham: Can we just escape the `%` instead?

ravi-mosaicml: It will then also escape in the docs build. I could do `%%` for now but don't want to have to push a new yahp version.

composer/cli/launcher.py
```diff
@@ -134,8 +147,11 @@ def _parse_args():
     args = parser.parse_args()
 
     # Default values to env vars if they are not provided
-    if args.nproc is None and 'LOCAL_WORLD_SIZE' in os.environ:
-        args.nproc = int(os.environ['LOCAL_WORLD_SIZE'])
+    if args.nproc is None:
+        if 'LOCAL_WORLD_SIZE' in os.environ:
+            args.nproc = int(os.environ['LOCAL_WORLD_SIZE'])
+        else:
+            args.nproc = max(1, torch.cuda.device_count())
```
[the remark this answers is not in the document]
ravi-mosaicml: This is safe to use without the check first. Tested locally:

```
Successfully installed torch-1.11.0+cpu torchaudio-0.11.0+cpu torchvision-0.12.0+cpu
(composer) ravi@dolphin:~/ravi-composer$ python3
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> torch.cuda.device_count()
0
```

ravi-mosaicml:
> Please don't delete an entrypoint (i.e. `examples/run_composer_trainer`) at such a critical time. I'm fine with making the shift to a new entrypoint, but we need to preserve the old one too. If you want to delete it, please do so only after it's had a deprecation notice and warning sitting on it for a minor version or two.

Fair point; added back the existing file with a deprecation warning.

composer/trainer/trainer.py
ravi-mosaicml: Discussed offline; will leave as-is for now. Created https://mosaicml.atlassian.net/browse/CO-558 to track it.

jbloxham (approved):
Thank you for restoring `run_composer_trainer`! That will go a long way towards making this a much smoother transition!

hanlint:
OK, I'll open a PR to change `model` -> `models`, `schedulers` -> `scheduler`, and `train_dataset` -> `train_datasets` 👿

ravi-mosaicml:
> OK, I'll open a PR to change `model` -> `models`, `schedulers` -> `scheduler`, and `train_dataset` -> `train_datasets` smiling_imp

PR #1199, opened 2022-06-22
Sequence length warmup update and tests

alextrott16:
- Add `preserve_end_of_sequence` option to sequence length warmup. This allows truncated batches to use the same end-of-sequence token as before truncation, which better supports token structures used in MLM (e.g., BERT) training.
- Allow memory pre-allocation to automatically increase `grad_accum` if the final batch*sequence size will produce OOM errors. This essentially inserts the automatic accumulation logic from the `trainer` class into the sequence length warmup algorithm's initialization.
- Minor refactors for readability.
- Add `tests/algorithms/test_seq_length_warmup.py` for unit testing sequence length warmup (functional and algorithm class).


composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,31 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
     """
 
     assert isinstance(batch, Mapping)
+    assert 'attention_mask' in batch
```
moinnadeem, line 70: Can we raise a ValueError if the `attention_mask` isn't in the batch?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,31 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
+        # eos_idx_truncated is the same thing, after truncation is applied
+        eos_idx_truncated = eos_idx.clamp(max=curr_seq_len - 1)
 
-        for k, v in batch.items():
-            if k == 'input_ids':
+        for k in batch.keys():
+            if len(batch[k].shape) < 2:
+                continue
+            eos_value = batch[k][r_idx, eos_idx]
+            batch[k] = batch[k][:, :curr_seq_len].contiguous()
```
moinnadeem, line 87: Do we need the contiguous here? I think it forces memory fragmentation, which is inefficient from a hardware perspective.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -266,11 +305,22 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
...
+        print('timestamp.batch', int(state.timestamp.batch))
+
+        curr_seq_len = self.step_size * (int(state.timestamp.batch) // update_every_n_steps) + self.min_seq_length
+        print('curr_seq_len', curr_seq_len)
         curr_seq_len = max(curr_seq_len, self.min_seq_length)
         curr_seq_len = min(curr_seq_len, self.max_seq_length)
 
-        state.batch = set_batch_sequence_length(state.batch, curr_seq_len, self.truncate)
+        if curr_seq_len != self._last_seq_len:
+            print(f'At batch {int(state.timestamp.batch)}, current sequence length = {curr_seq_len}.')
```
moinnadeem, line 320: Same comment about print statements

moinnadeem (commented):
So far, so good! We will need someone else with more `auto_grad_accum` experience to examine that chunk of the code.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Unfortunately, it is necessary. The BERT forward pass fails without the contiguous. I have expanded the test code to confirm that the model can do a forward/backward pass on the modified batch -- those tests will fail for BERT models if the contiguous is removed.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,31 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
-        tensor_len = (input_ids.shape[0] // curr_seq_len) * curr_seq_len
+    if curr_seq_len >= batch['attention_mask'].shape[1]:
+        return batch
 
-        input_ids = input_ids[:tensor_len]
-        input_ids = input_ids.view(-1, curr_seq_len)
-        batch['input_ids'] = input_ids
+    # Truncate, but preserve end-of-sequence tokens
+    if preserve_eos:
+        r_idx = torch.arange(batch['attention_mask'].shape[0])
```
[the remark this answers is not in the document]
alextrott16: Changed to "preserve_end_of_sequence"

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Done

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,31 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
+            batch[k][r_idx, eos_idx_truncated] = eos_value
+
+    else:
+        for k in batch.keys():
+            if len(batch[k].shape) < 2:
                 continue
-            v = v.view(-1)
-            v = v[:tensor_len]
-            batch[k] = v.view(-1, curr_seq_len)
+            batch[k] = batch[k][:, :curr_seq_len].contiguous()
```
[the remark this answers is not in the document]
hanlint: Surprised `continguous()` is needed here given that the previous code when `truncate=True` did not need it with the same indexing scheme

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -43,8 +45,9 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
     Args:
         batch (Dict[str, Tensor]): The input batch to the model, must be a dictionary.
         curr_seq_length (int): The desired sequence length to apply.
-        truncate (bool, optional): Truncate sequences early, or reshape tensors to create
-            new examples out of the extra tokens. Default: ``True``.
+        preserve_end_of_sequence (bool, optional): Preserve the end-of-sequence of the batch when
```
hanlint, line 48: Do we still want to preserve the `truncate=False` option for backwards compatibility?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -16,18 +16,20 @@
...
+    curr_seq_len: int,
+    preserve_end_of_sequence: bool = False,
+) -> Batch:
     """Set the sequence length of a batch.
 
     Changes the sequence length of all tensors in the provided dictionary
-    to ``curr_seq_len`` by either truncating the tensors (``truncate=True``)
-    or reshaping the tensors to create new examples from the extra tokens
-    (``truncate=False``).
+    to ``curr_seq_len`` by truncating the sequence tensors.
```
hanlint, line 32: Suggest the new option be explained in the docstrings perhaps with some example sequences?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,32 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
     """
 
     assert isinstance(batch, Mapping)
+    if 'attention_mask' not in batch:
```
hanlint, line 70: only if `preserve_end_of_sequence` is True though, right?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -64,25 +67,32 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
+        return batch
 
-        input_ids = input_ids[:tensor_len]
-        input_ids = input_ids.view(-1, curr_seq_len)
-        batch['input_ids'] = input_ids
+    # Truncate, but preserve end-of-sequence tokens
+    if preserve_end_of_sequence:
+        r_idx = torch.arange(batch['attention_mask'].shape[0])
+        # eos_idx should point to the final token index for each batch sample
+        eos_idx = batch['attention_mask'].sum(1).long() - 1
```
hanlint, line 80:
```suggestion
        eos_idx = batch['attention_mask'].sum(dim=1).long() - 1
```
nit: use `dim` kwargs for explicit

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -217,33 +210,75 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
...
-                loss = self._original_model.loss(outputs, model_inputs)
-
-            # since use_grad_scaling is in the Trainer, and we
-            # don't care about the loss values, skip scaling
-            for loss_item in ensure_tuple(loss):
-                loss_item.backward()
-
-            for optimizer in state.optimizers:
-                optimizer.zero_grad()
+
```
hanlint, line 213: some comment here on what the below loops are doing would be helpful for readers

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -217,33 +210,75 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
...
+            batch_clone = {}
+            device_batch_size = 0
+            device = None
+            for k, v in state.batch.items():
```
hanlint, line 217: is there a reason to use the `state.batch` directly, versus creating a random tensor as before?

hanlint (commented):
Defer to other for approval, but made a few comments/suggestions, specifically around preserving the old behavior (e.g. not contigious, truncating, using random tensor for the memory allocation) as much as possible

composer/algorithms/seq_length_warmup/seq_length_warmup.py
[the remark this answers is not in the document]
alextrott16: Could it be that the previous code was tested using a GPT model? The issue only arises when using a BERT model.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: It seemed more straightforward to use the `state.batch` since it reflects the structure of actual training batches. The alternative required inferring (or hardcoding) the batch structure in a way that seemed less flexible and/or harder to maintain if models change.
If I'm missing an upside to using random tensors, though, please let me know.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Good point. I guess we shouldn't assume that others aren't using `truncate=False` just because we aren't. I will add it back in.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Yes, good point. We can relax this requirement to that case. Will do.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
hanlint: OK that makes sense. One style nit is to break this up into a few logical bits, e.g. clone everything first, then truncate.
```
batch_clone = {k: torch.clone(v) for k, v in state.batch.items()]

# truncate all sequence-shaped tensors to the max sequence length
for k, v in batch_clone.items():
    if len(v.shape) > 1:
        batch_clone[v] = batch_clone[v][:, :self.max_seq_length]

device_batch_size = 
device = 
```

It's less efficient, but more readable

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -217,33 +211,78 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
...
+                if len(v.shape) > 1:
+                    seq_dim = min(self.max_seq_length, v.shape[1])
+                    batch_clone[k] = torch.clone(v[:, :seq_dim])
+                else:
+                    batch_clone[k] = torch.clone(v)
+                device_batch_size = v.shape[0]
+                device = v.device
+
+            grad_accum_successful = False
+            while not grad_accum_successful:
```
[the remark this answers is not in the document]
hanlint: The code looks ok, but i'm not the expert. It's sufficiently long that might want to break out into its own helper function for modularity. 

I'm also curious how well this works. We've heard sporadic reports of auto grad accum not freeing up all the memory. Would be interesting to experiment with/without using `torch.cuda.empty_cache()`

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Expanded the docstrings to have such an example

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Added comments to clarify

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Changes applied

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -65,10 +78,48 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
+                    'Sequence Length Warmup requires that the batch has "attention_mask" when using ``preserve_end_of_sequence=True``.'
+                )
+            r_idx = torch.arange(batch['attention_mask'].shape[0])
+            # eos_idx should point to the final token index for each batch sample
+            eos_idx = batch['attention_mask'].sum(dim=1).long() - 1
+            # eos_idx_truncated is the same thing, after truncation is applied
+            eos_idx_truncated = eos_idx.clamp(max=curr_seq_len - 1)
+
+            for k in batch.keys():
+                if len(batch[k].shape) < 2:
```
moinnadeem, line 106:
```suggestion
                if batch[k].ndim < 2:
```

 Can we use `ndim` here instead?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -65,10 +78,48 @@ def set_batch_sequence_length(batch: Dict[str, torch.Tensor], curr_seq_len: int,
...
+                    continue
+                eos_value = batch[k][r_idx, eos_idx]
+                batch[k] = batch[k][:, :curr_seq_len].contiguous()
+                batch[k][r_idx, eos_idx_truncated] = eos_value
+
+        else:
+            for k in batch.keys():
+                if len(batch[k].shape) < 2:
```
moinnadeem, line 114:
```suggestion
                if batch[k].ndim < 2:
```
See above

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -171,6 +237,103 @@ def __init__(
...
+                'Sequence Length Warmup requires the `state.dataloader` to have a `batch_size` attribute.') from e
+        if per_gpu_macrobatch is None:
+            raise RuntimeError('Sequence Length Warmup algorithm requires constant batch size.')
+
+        # truncate all sequence-shaped tensors to the max sequence length
+        batch_clone = {k: torch.clone(v) for k, v in state.batch.items()}
+        device_batch_size = 0
+        device = None
+        for k, v in batch_clone.items():
+            if len(v.shape) > 1:
```
moinnadeem, line 271:
```suggestion
            if v.ndim > 1:
```

moinnadeem (approved):
LGTM with a caveat on the `continuous` -- @hanlint does that have a throughput hit?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
alextrott16: Yes! Great suggestion. I will change these throughout. (I will incorporate all these changes in a single commit and resolve the individual suggestions separately.)

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -171,6 +237,103 @@ def __init__(
...
+
+            # This error/state.grad_accum handling mimics the logic in trainer._train_batch().
+            except RuntimeError as e:
+                if 'CUDA out of memory' in str(e):
+                    should_handle_cuda_oom = 1
+                elif 'Timed out' in str(e):
+                    # Catch timeout errors and only reraise if we did not encounter OOM on other ranks. Error
+                    # is likely transient if one rank OOMed, it likely did not reach a barrier. Note that if we
+                    # catch non-transient timeout errors they will be later reraised if no rank OOMed.
+                    caught_timeout_error = e
```
mvpatel2000, line 321: This can be deleted. It's for handling CUDA OOMs with gradient syncing, but the current design is flakey and we'll rework it. In either case, there's no optimizer step / gradient syncing here anyways so it's safe to remove

composer/algorithms/seq_length_warmup/seq_length_warmup.py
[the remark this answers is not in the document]
alextrott16: I played around a bit with `torch.cuda.empty_cache()` and couldn't find any clear advantages. It is possible (likely, even) that any memory issues with the auto grad accum will also occur here, since the logic is essentially identical.

At the very least, I took your suggestion wrt using a helper function.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -171,6 +237,103 @@ def __init__(
...
+                    state.grad_accum = min(2 * state.grad_accum, device_batch_size)
+                    logger.data_batch({'trainer/grad_accum': state.grad_accum})
+            elif caught_timeout_error:
+                # If not CUDA out of memory, raise exception to user. Note that this truncates the call stack
+                # back only to this newly raised error.
+                raise caught_timeout_error
+            else:
+                grad_accum_successful = True
+
+        self._activated = True
```
mvpatel2000, line 346: Minor Nit: I personally would prefer to remove `grad_accum_successful` and instead have:
```
while True:
    ...
    else:
        self._activated = True
        return
```
to be consistent with trainer `while` structure. Not strongly opinionated though

mvpatel2000 (commented):
Two minor comments on the auto grad accum part -- sorry I took a bit to look at this!

PR #1200, opened 2022-06-22
Fix module surgery for training resumptions with optimizers that save state

dskhudia:
Currently, when we attempt to save/load check points when:
(1) a surgery algorithm has been used that modifies the pytorch otpimizer state
(2) the trainer is using an optimizer that has state (.e.g., adam) 

it fails. 

This fixes it. However, SAM still fails with mismatch and not sure what's the root cause. 

CO-382

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -994,10 +994,15 @@ class Trainer:
                 progress_bar=load_progress_bar,
                 ignore_keys=load_ignore_keys,
             )
-            # Always ignore the run_name in the checkpoint so it is consistent with what was used for Event.INIT.
             self.state.run_name = run_name
-            log.info(f'Setting seed to {self.state.seed}')
-            reproducibility.seed_all(self.state.seed)
+
+        # reseed here. This helps with a couple of issues:
+        # 1. rng state may change at Event.INIT. For example, if an algorithm creates a new module and module
+        # parameters are initialized randomly, rng state will change. This reseeding nullifies such effects.
+        # 2. While resuming from a checkpoint, we want to spin dataloader and bring it back to the same state as at the time
+        # of the checkpoint. Therefore, spinning needs to start from the same rng state as in the original run.
+        log.info(f'Setting seed to {self.state.seed}')
+        reproducibility.seed_all(self.state.seed)
 
         # Move the model and optimizers to the specified device
         if not self.deepspeed_enabled:
--- a/tests/algorithms/test_algorithm_resumption.py
+++ b/tests/algorithms/test_algorithm_resumption.py
@@ -10,8 +10,7 @@ import pytest
 import torch
 
 from composer import Algorithm, Trainer
-from composer.algorithms import (SAM, BlurPool, Factorize, GatedLinearUnits, LayerFreezing, SqueezeExcite,
-                                 StochasticDepth)
+from composer.algorithms import SAM, LayerFreezing, StochasticDepth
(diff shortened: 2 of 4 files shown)


composer/trainer/trainer.py
```diff
@@ -996,7 +996,6 @@ def __init__(
             # Always ignore the run_name in the checkpoint so it is consistent with what was used for Event.INIT.
             self.state.run_name = run_name
             log.info(f'Setting seed to {self.state.seed}')
-            reproducibility.seed_all(self.state.seed)
```
ravi-mosaicml, line 999: Hmm, curious why was this line removed? I believe it is needed for the checkpointing tests...

ravi-mosaicml (approved):
Wow, good work -- thanks for solving this. LGTM; see comments.

composer/trainer/trainer.py
dskhudia: Random number generator state is restored from the checkpoint. We don't need to seed them again. If we do, resumption from a checkpoint won't match as the generated random numbers (If an algorithm uses them) would be different.

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -33,7 +33,7 @@ def test_algorithm_resumption(
     if alg_cls is LayerFreezing:
         pytest.xfail('Known issues')
 
-    if alg_cls in (SAM, SqueezeExcite, StochasticDepth, Factorize, BlurPool):
+    if alg_cls is SAM:
```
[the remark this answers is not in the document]
dskhudia: Existing tests cover this. They were disabled. I enabled them in this fix.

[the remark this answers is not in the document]
dskhudia: Added a standalone unit test for `module_surgery.update_params_in_optimizer`

It fails without the fix in this PR.

composer/trainer/trainer.py
dskhudia: Removed the log statement and comment as well.

hanlint (approved):
LGTM, thanks!

composer/trainer/trainer.py
ravi-mosaicml: Ah, I needed a night's sleep to remember the issue here:

The seed must be restored here, as the dataloader initializes its sampler on the first spin, not during Event.INIT. Without this, then the dataloader sampler will be initialized with a different random value than the seed in the checkpoint.

hanlint: Maybe we should add a comment there to record this for posterity :)

ravi-mosaicml: Agreed with hanlin this needs to be better documented haha :)

Following the code path in `trainer.py`, here's what happens w.r.t to seed and RNG state.

1. Run `Event.INIT` -- nothing should depend on the seed / rng here (line 874)
1. Load checkpoint (line 987)
2. Restore the seed (line 999)
3. Run `Event.FIT_START` (line 1420). Nothing should use the seed here.
4. Spin the dataloaders (line 1428). This will cause the dataloaders to read the RNG from the seed
5. Before the first batch that will be trained after the checkpoint, restore the RNG state, so it is accurate for events that will be used by the algorithms (line 1432 and line 1464, depending on whether it is a batch or epoch checkpoint)

So, in effect, the `state.seed` is used for the dataloaders, and `state.rng` is used for the algorithms

PR #1201, opened 2022-06-22
Fix multi-GPU conflicts when downloading `torchvision` datasets

abhi-mosaic:
This fixes a race condition when `download=True` where every GPU process on the node is trying to download to the same directory.


composer/datasets/cifar_hparams.py
```diff
@@ -162,12 +162,20 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
+                    self.datadir,
+                    train=self.is_train,
+                    download=download,
+                    transform=transformation,
+                )
+
+            if self.download:
+                if dist.get_local_rank() == 0:
+                    _get_dataset(download=True)
+                dist.barrier()
```
hanlint, line 176: This will fail if dist is not initialized. Suggestion, can we maybe use a contextmanager for this? something like:
```
    def __init__(download: bool):
        self.download = download
        self.wait_for_rank_zero = download and dist.initialized()
        
    def __enter__(self):
        if not self.wait_for_rank_zero: 
             return
        if dist.get_local_rank() == 0:
             self.download = True
         else:
             self.download = False
             dist.barrier()

def __exit__():
       if self.wait_for_rank_zero and dist.get_local_rank==0:
             dist.barrier()
           
```

so that this can be reused across datasets?

composer/datasets/mnist_hparams.py
```diff
@@ -52,12 +52,21 @@ def initialize_object(self, batch_size: int, dataloader_hparams: DataLoaderHpara
...
 
             transform = transforms.Compose([transforms.ToTensor()])
-            dataset = datasets.MNIST(
-                self.datadir,
-                train=self.is_train,
-                download=self.download,
-                transform=transform,
-            )
+
+            def _get_dataset(download: bool):
```
ravi-mosaicml, line 56: Nit: make this a lambda? E.g.
`_get_dataset = lambda download: datasets.MNIST(...)`

composer/datasets/cifar_hparams.py
ravi-mosaicml: `Dist` should be initialized when using the hparams codepath -- see https://github.com/mosaicml/composer/blob/81f3db523f6f7133262c5c4c8c53f7ef6060e1a2/composer/trainer/trainer_hparams.py#L424

ravi-mosaicml: Also, see nit belong re: making this a lambda.

ravi-mosaicml (approved):
LGTM; see comments

hanlint:
@ravi-mosaicml request a re-review as the implementation has changed significantly

composer/utils/dist.py
```diff
@@ -420,3 +421,35 @@ def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool, shuffle:
         num_replicas=get_world_size(),
         rank=get_global_rank(),
     )
+
+
+@contextmanager
+def wait_for_rank_zero():
```
ravi-mosaicml, line 427: (Discussed offline with hanlin, but would prefer that we make this shared helper more specific for torchvision, since this use case is very specific)

It's a much more common use case for rank zero to run some (extra) code, wait for a barrier, then all ranks run the same code rather than for rank zero to run code, wait, and then let all other ranks run the same code (while rank zero waits -- there should probably be another barrier at the end)

composer/utils/dist.py
```diff
@@ -420,3 +421,35 @@ def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool, shuffle:
...
+                ...,
+                download=True,
+            )
+
+    This prevents race conditions where multiple
+    ranks attempt to download the dataset to the
+    same location.
+    """
+    if not is_initialized():
+        return
```
Landanjs, line 447: Should this be yield? there might be issues with single gpu mnist example

Landanjs: Oh, @hanlint beat me to it

PR #1202, opened 2022-06-23
Fix the SFTP Object Store

ravi-mosaicml:
- AWS SFTP Gateway does not support directories, which breaks paramiko's `chdir` (it always returns a FileNotFound exception, as the `mkdir -p` is ineffective). Instead, store the cwd internally, and append it to any `ojbect_names`.
- Add in a config option for the `MissingHostKeyPolicy`

Closes https://mosaicml.atlassian.net/browse/CO-569

### DIFF
--- a/composer/utils/object_store/sftp_object_store.py
+++ b/composer/utils/object_store/sftp_object_store.py
@@ -60,7 +60,7 @@ class SFTPObjectStore(ObjectStore):
             :class:`paramiko.client.MissingHostKeyPolicy` to use for a missing host key. Defaults to ``'RejectPolicy'``.
 
             Built-in options:
-            *   ``'RejectPolicy'`` (the default), which will reject any host key authorized in the ``known_hosts_filename``.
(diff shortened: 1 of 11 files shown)


composer/utils/object_store/sftp_object_store.py
```diff
@@ -52,6 +56,15 @@ class SFTPObjectStore(ObjectStore):
             specified in ``~/.ssh/`` or via a SSH agent.
         known_hosts_filename (pathlib.Path | str, optional): The filename of the known hosts file. If not specified,
             the default SSH known hosts will be used.
+        missing_host_key_policy (str | paramiko.client.MissingHostKeyPolicy): The class name or instance of
+            :class:`paramiko.client.MissingHostKeyPolicy` to use for a missing host key. Defaults to ``'RejectPolicy'``.
+
+            Built-in options:
+            *   ``'RejectPolicy'`` (the default), which will reject any host key authorized in the ``known_hosts_filename``.
```
ravi-mosaicml, line 63:
```suggestion
            *   ``'RejectPolicy'`` (the default), which will reject any host key not authorized in the ``known_hosts_filename``.
```

PR #1204, opened 2022-06-23
Read from env variable if `run_name` not provided to Trainer

ajaysaini725:
Makes the Trainer check the env variable `COMPOSER_RUN_NAME` if `run_name` is not passed in. If the environment variable also does not exist, then a random run name is generated.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -53,6 +53,8 @@ __all__ = ['Trainer']
 # syntax to shorten the Scheduler type annoations
 Scheduler = Union[ComposerScheduler, PyTorchScheduler]
 
+_COMPOSER_RUN_NAME_ENV_KEY = 'COMPOSER_RUN_NAME'
+
 
 def _raise_missing_argument_exception(arg_name: str):
     raise ValueError((f'{arg_name} is a required argument and must be specified when constructing the '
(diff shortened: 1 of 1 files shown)


ravi-mosaicml (changes requested):
Hmm, I would prefer not to introduce env variables for every hook between composer and mcli -- I think we will need some sort of better integration here (perhaps a runtime that mcli automatically injects in the image, that then modifies composer as necessarry).

For now, would it work to use the env var `RUN_NAME`? YAHP should automatically parse and load from this -- it's totally undocumented, but see https://github.com/mosaicml/yahp/blob/a3d21be2c810b95e56492548b11583f37a41de72/yahp/create_object/create_object.py#L177

ajaysaini725:
> Hmm, I would prefer not to introduce env variables for every hook between composer and mcli -- I think we will need some sort of better integration here (perhaps a runtime that mcli automatically injects in the image, that then modifies composer as necessarry).
> 
> For now, would it work to use the env var `RUN_NAME`? YAHP should automatically parse and load from this -- it's totally undocumented, but see https://github.com/mosaicml/yahp/blob/a3d21be2c810b95e56492548b11583f37a41de72/yahp/create_object/create_object.py#L177

@ravi-mosaicml I agree that eventually MCLI should inject some runtime that modifies Composer - that is the ideal longer term solution.

Because we need to get this up and running in the next day and it's not a huge change can we stick with this env variable? I do want this to work even if the customer is not using YAHP so ideally we don't rely on any env var parsing yahp does. Doing this also won't affect any Composer users that are not using MCLI and also doesn't even change the current workflows for people who are using MCLI but have been specifying the run name in Composer because that takes precedence.

ravi-mosaicml:
Closing per offline discussion -- gonna only support auto-injection via yahp for now; will scope out a longer-term integration.

PR #1208, opened 2022-06-24
Fix the ObjectStoreLogger with Overwrite=True

ravi-mosaicml:
Where multiple workers uploading to the same object name, the actual latest version in the object store could be overridden by a earlier version that happened to have its upload finish later.
This happened in practice on a s3 backend, when uploading checkpoints every batch. The latest symlink file at the end of training was not indeed the latest symlink.

To fix, this PR keeps track of the enqueued object names in a set. When the upload api is called, add it to a `{object_name: latest_filepath}` dictionary that lives on the main thread. Periodically, also on the main thread, drain elements (that are not already in the enqueued object name set) from this dictionary onto the upload queue that is processed by the workers. Once a worker uploads the object, remove it from the enqueued object names set. In effect, each object name appears only once in the queue accessible by workers, and if an object is updated multiple times locally before it is given to the workers, then the intermediate versions won't be uploaded.

A test case is included to ensure this race condition fixed the issue.

Closes https://mosaicml.atlassian.net/browse/CO-570

Confirmed this fix worked -- see the strict version ordering on https://s3.console.aws.amazon.com/s3/object/mosaicml-internal-checkpoints-test?region=us-west-2&prefix=mnist_checkpoint_5%2Fravi-debugging%2Fcheckpoints%2Flatest-rank0.symlink&tab=versions

TODO:

- [x] Merge #1209. This PR is stacked on top of it. For the diff of just this PR, see https://github.com/ravi-mosaicml/ravi-composer/pull/12.


Averylamp (approved):
lgtm

bandish-shah (approved):
LGTM

PR #1209, opened 2022-06-24
Fix Symlinks on Non-Libcloud Object Stores

ravi-mosaicml:
As part of the object store refactor, forgot to fully abstract out the symlink logic. This PR fixes that.

* Add `composer.utils.create_symlink_file`: This API creates a symlink file that can be followed by `get_file`
* Moved the symlink following out of the libcloud object store and into `get_file`
* Remove the `logger.log_symlink_artifact` API. Instead, require the caller to use `composer.utils.create_symlink_file`, and then log that directly via the `logger.log_file_artifact` API

Closes https://mosaicml.atlassian.net/browse/CO-568


composer/utils/file_helpers.py
```diff
@@ -339,6 +341,64 @@ def get_file(
...
+    try:
+        _get_file(
+            path=path,
+            destination=destination,
+            object_store=object_store,
+            overwrite=overwrite,
+            progress_bar=progress_bar,
+        )
+    except FileNotFoundError as e:
+        if path.endswith('.symlink'):
```
milocress, line 377: nit: if I am reading your code correctly, this condition will always evaluate to `False`, because of the check above.

otherwise, lgtm

milocress (approved):

bandish-shah (approved):
LGTM

composer/utils/file_helpers.py
ravi-mosaicml: Thanks, good catch. I ended up refactoring the logic and forgot to remove this now-redundant check

ravi-mosaicml: Fixed

PR #1211, opened 2022-06-24
Remove the `composer_train` entrypoint; put it back in `examples`

ravi-mosaicml:
This PR partially reverts #1195 and removes the `composer_train` entrypoint from the pip package. Decided against including this in the pip package as it is very opinionated (e.g. with wandb config logging) and hides how users should use their own entrypoints.

It was also confusing to use it with the launcher script -- `composer -c composer_train` is unintuitive.

Instead, for users who want to use our example entrypoint with a `pip install` of composer, they can `wget` it from github.com at runtime.


moinnadeem (approved):
LGTM

hanlint (approved):
I concur

PR #1212, opened 2022-06-24
Set SFTP Object Store Private Key Filepath from an Environ

ravi-mosaicml:
This PR allows the SFTP Object Store filepath to be set via an environ when using the yahp codepath. The enviorn defaults to `COMPOSER_SFTP_KEY_FILE` to match what is used by streaming datasets; however, this variable can be overridden via the `key_filename_environ` argument (to support multiple sftp object stores with different keys).

Closes https://mosaicml.atlassian.net/browse/CO-573


abhi-mosaic (approved):
LGTM

PR #1215, opened 2022-06-24
Throughput metrics

linden-li:
Change `samples/sec` -> `throughput/samples_per_sec` to be consistent with the convention that `/` should indicate a hierarchy rather than a rate and eliminate `samples/epoch`


hanlint (approved):
LGTM, cc: @bandish-shah in case any profiler impacts here?

abhi-mosaic:
Looks good! There might be some unused variables that can be cleaned up though, like `self.epoch_start_num_samples`, will defer to Pyright

hanlint:
Ah good point -= @linden-li do you mind removing some of the unneeded variables? Pyright won't catch these, as they are `self.`

PR #1217, opened 2022-06-24
Cleanup SFTP url parsing for StreamingDataset

abhi-mosaic:
This PR allows SFTP urls passed to StreamingDataset to accept the same information (username, port, etc.) that we currently can for our SFTPObjectStore. This also eliminates the need for the `COMPOSER_SFTP_USERNAME` enviornment variable.

### DIFF
--- a/composer/datasets/streaming/download.py
+++ b/composer/datasets/streaming/download.py
@@ -45,7 +45,7 @@ def download_from_sftp(remote: str, local: str) -> None:
         local (str): Local path (local filesystem).
     """
     try:
-        from paramiko import AutoAddPolicy, RSAKey, SSHClient, SSHConfig
+        from paramiko import AutoAddPolicy, SSHClient
     except ImportError as e:
         raise MissingConditionalImportError(extra_deps_group='streaming', conda_package='paramiko') from e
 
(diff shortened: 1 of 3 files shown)


composer/datasets/streaming/download.py
```diff
@@ -66,7 +77,11 @@ def download_from_sftp(remote: str, local: str) -> None:
     with SSHClient() as ssh_client:
         # Connect SSH Client
         ssh_client.set_missing_host_key_policy(AutoAddPolicy)
```
ravi-mosaicml, line 79: Ooof ... I missed this line before, but do not think we should default to this. OK making this an option, but would strongly prefer to default to the RejectPolicy (as does paramiko).

CC @nlsapp @hanlint  for security review

ravi-mosaicml (approved):
LGTM, though see the (unrelated) comment re: the missing host key policy.

I'll approve this PR for now, though let's make sure to address it.

composer/datasets/streaming/download.py
abhi-mosaic: Ok, is this the same behavior with SFTPObjectStore? Like do we have host keys setup in the docker containers? I can test it.

ravi-mosaicml: The SFTPObjectStore defaults to reject, but has a config option to override this.

nlsapp: Should it reject by default and force the user to provide host keys as an argument to mcli (or like, an env var secret/configmap): Yes

Should we have an option, that allows a user to decide if hostkeys are ignored: Also yes

But def not the default and ideally we should warn them: something something "You are ignoring SFTP Host Keys, THIS IS UNSAFE!" or something

abhi-mosaic: Gonna address this in a followup PR, needs some coordination with the SFTPObjectStore and launch interface.

PR #1219, opened 2022-06-27
Update bert-base.yaml

moinnadeem:
Updates the BERT YAML to be streaming

### DIFF
--- a/composer/yamls/models/bert-base.yaml
+++ b/composer/yamls/models/bert-base.yaml
@@ -1,62 +1,78 @@
-train_dataset:
-  lm:
-    split: train
-    datadir:
-      - /datasets/wikipedia_saved_bert_128
-      - /datasets/bookcorpus_saved_bert_128
-    tokenizer_name: bert-base-uncased
-    seed: 19
-    shuffle: true
-    drop_last: false
-    subsample_ratio: 1.0
-    use_masked_lm: true
-    mlm_probability: 0.15
-    max_seq_length: 128
-evaluators:
-  - evaluator:
-      label: bert-base
-      eval_dataset:
-        lm:
-          split: validation
-          datadir:
-            - /datasets/wikipedia_saved_bert_128
-            - /datasets/bookcorpus_saved_bert_128
-          tokenizer_name: bert-base-uncased
-          seed: 19
-          shuffle: true
-          drop_last: false
-          subsample_ratio: 0.02
-          use_masked_lm: true
-          mlm_probability: 0.15
-          max_seq_length: 128
-      metric_names:
-        - LanguageCrossEntropy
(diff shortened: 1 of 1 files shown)


ravi-mosaicml:
Removing myself from reviews; research would be better to review hparams changes.

composer/yamls/models/bert-base.yaml
```diff
@@ -1,62 +1,78 @@
...
-  prefetch_factor: 2
-grad_accum: 2
-eval_interval: 1000ba
+
+precision: amp                     # Use mixed-precision training
+grad_clip_norm: -1.0               # Turn off gradient clipping
+grad_accum: 'auto'                 # Use automatic gradient accumulation to avoid OOMs
+
+save_folder: bert_checkpoints      # The directory to save checkpoints to
+save_interval: 3500ba              # Save checkpoints every 3500 batches
```
bandish-shah, line 74: [nit] In the other models we release with Composer we don't enable checkpointing.  Also not sure I want to enable this by default for regressions at this time.  I don't think this is a hard requirement but I'm thinking it might be nice to keep this bare minimum without imposing unnecessary resource requirements (disk space in this case) to users by default for these examples.  We could comment this out and put a link to the relevant Composer docs should a user want to enable it.

bandish-shah (approved):
LGTM, please see nit's for some discussion points.

composer/yamls/models/bert-base.yaml
moinnadeem: Haha, I was actually about to post this as a discussion point in Slack. YAMLs currently serve two purposes:
1. Let an end-user run a benchmark with as minimal configuration as possible
2. Let team members start working on the benchmark, by just pointing them to the default YAML for running that benchmark.

I was about to actually include object store logger hyperparameters as well, but figured that would serve Goal #2 at the expense of Goal #1. The best option is to add additional commented fields that help goal #2, and therefore wouldn't hurt Goal #1.

> In the other models we release with Composer we don't enable checkpointing
Checkpointing is enabled by default because it's required for BERT evaluation -- if someone performs a BERT run without checkpointing, they'll have to add it back in to actually evaluate it. Therefore, I'm of the strong opinion that it should be a default.

I don't want to build YAMLs to the specification required by regressions -- it is the regressions responsibility to delete fields that are incompatible. These YAMLs should be the YAMLs that provide Composer users with the best possible user experience.

composer/yamls/models/bert-base.yaml
```diff
@@ -1,62 +1,78 @@
...
+
+
+loggers:
```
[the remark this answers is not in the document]
moinnadeem: See the above comment!

PR #1222, opened 2022-06-27
Improve docstrings.

knighton:


ravi-mosaicml (approved):
LGTM; though please add another pair of tick marks for `fields` on https://mosaicml-composer--1222.com.readthedocs.build/en/1222/api_reference/composer.datasets.streaming.writer.html. It's in italics rn.

> "Given each sample, StreamingDatasetWriter only writes out the values for a subset of keys (_fields_) that are globally shared across the dataset."

PR #1223, opened 2022-06-27
Update vision-style `StreamingDataset`s to subclass `VisionDataset`

ravi-mosaicml:
Allows algorithms that expect a torchvision-style VisionDataset to work on the streaming dataset implementations.

Closes https://mosaicml.atlassian.net/browse/CO-608


knighton (approved):
Good times

PR #1224, opened 2022-06-28
Add `COMPOSER_KNOWN_HOSTS_FILENAME` for setting the sftp known hosts file environ

ravi-mosaicml:
Allow for setting the sftp known hosts filename via an environ, similar to how #1212 allows for setting the private key file via an environ. Remove `AutoAddPolicy` as the default SSH Private Key Policy.

Closes https://mosaicml.atlassian.net/browse/CO-614


abhi-mosaic (approved):
LGTM! The `StreamingDataset` download handler should probably also be updated to read this env var?

PR #1226, opened 2022-06-28
Fix mnist download for single GPU

hanlint:
When `dist` is not initialized, the context manager should yield instead of return

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -445,11 +445,12 @@ def run_local_rank_zero_first():
     """
     if not is_initialized():
         yield
+        return
+    
+    # hold non-zero ranks until rank zero done
+    if get_local_rank() != 0:
+        dist.barrier()
+        yield
     else:
-        # hold non-zero ranks until rank zero done
-        if get_local_rank() != 0:
-            dist.barrier()
-            yield
-        else:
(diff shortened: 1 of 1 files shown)


siriuslee (approved):
LGTM. Tested and it worked for me.

ravi-mosaicml (approved):
Fix LGTM, though I wonder should we raise an exception here if `dist.get_world_size() > 1`? If you have multiple ranks and dist is not initialized, then I think it would be unexpected to simply yield, since all ranks would immediately run whatever is in the context manager?

hanlint:
@ravi-mosaicml is there a use case where you have multiple ranks but `dist` is _not_ initialized?

ravi-mosaicml:
> @ravi-mosaicml is there a use case where you have multiple ranks but `dist` is _not_ initialized?

Yes, if using this function before `dist.initialize()` is called -- it's an easy mistake to make, especially if a user wants to write some dist code before using this function.

I incorporated this fix in https://github.com/mosaicml/composer/pull/1218/files#diff-cb8a6849a875ed68d93fc6beca6a6f24f7d0e520b87509433a4b94ce4bf09866, so going to close this PR, since it would now be 0-line diff.

PR #1228, opened 2022-06-28
StreamingDataset: correctly handle exceptions in child download thread.

knighton:


composer/datasets/streaming/dataset.py
```diff
@@ -209,14 +216,14 @@ def _insert_shard_samples(self, shard: int, part_min_id: int, part_max_id: int)
             # Extend and optionally reshuffle the remaining samples of any
             # epochs we have in progress.
             if self.shuffle:
-                if not self._is_downloaded:
+                if self._download_status == DownloadStatus.IN_PROGRESS:
                     self._downloaded_ids.extend(new_ids)
                     np.random.shuffle(self._downloaded_ids)
                 for todo_ids in self._epoch_to_todo_ids.values():
```
ravi-mosaicml, line 222: Do we need to handle DownloadStatus == FAILED and DownloadStatus.NOT_STARTED?

composer/datasets/streaming/dataset.py
```diff
@@ -227,8 +234,9 @@ def download(self) -> None:
             self._lock = Lock()
 
         with self._lock:
-            if self._is_downloaded:
+            if self._download_status != DownloadStatus.NOT_STARTED:
```
ravi-mosaicml, line 237: Would it make sense to call download a second time (i.e. if it failed) If so, then should this be check  to see if download_steatus != finished rather than not started?

ravi-mosaicml (approved):
See comments, but otherwise lgtm

composer/datasets/streaming/dataset.py
```diff
@@ -27,6 +27,13 @@ class DatasetCompressionException(Exception):
     pass
 
 
+class DownloadStatus(object):
```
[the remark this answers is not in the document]
knighton: Done

composer/datasets/streaming/dataset.py
knighton: currently having it die upon exception raised -- download failures should be forgiven at the point of retry/timeout parameters

making subsequent calls to download() start new download threads would expose us to races and likely not speed up downloading

composer/datasets/streaming/dataset.py
knighton: insert_shard_samples() is only called from download(), which does not call it in those states

PR #1230, opened 2022-06-28
Update CutMix to work with segmentation style labels

coryMosaicML:
This PR refactors CutMix to follow the way MixUp is done. This is mostly a usability change. The  goal is to accomplish 3 things:
1. Remove the need to specify `num_classes` as a hyper parameter. Now it should be inferred from the shape of the output tensor
2. Remove the requirement to use one-hot or dense labels by providing an `interpolate_loss` option which uses index labels and does interpolation directly on the loss, similar to the original implementation.
3. Enable the use of CutMix with segmentation style labels. 

Using `interpolate_loss=True` with both MixUp and CutMix can cause composition to behave strangely. A `NotIntendedUseWarning` should be raised in this case.

Runs:
[Current dev ResNet50-i1k Accuracy 77.41%](https://wandb.ai/mosaic-ml/cory-cutmix-update/runs/8leie5u3)
[This PR ResNet50-i1k + CutMix interpolate_loss=False Accuracy 77.47%](https://wandb.ai/mosaic-ml/cory-cutmix-update/runs/3ncrkg12)
[This PR ResNet50-i1k + CutMix interpolate_loss=True Accuracy 77.66%](https://wandb.ai/mosaic-ml/cory-cutmix-update/runs/p7ey8po6)
[This PR DeepLabV3+-ADE20k-webdataset interpolate_loss=False mIoU 44.3](https://wandb.ai/mosaic-ml/cory-cutmix-update/runs/398z1l04)
[Verification that the correct warning is raised when mixup and cutmix are run together with `interpolate_loss`](https://wandb.ai/mosaic-ml/cory-cutmix-update/runs/2z6o0dr5/logs)


Landanjs (approved):
I think it looks good! A few nits if you think they are worth addressing

composer/algorithms/cutmix/README.md
```diff
@@ -28,13 +29,9 @@ def training_loop(model, train_loader):
...
         for X, y in train_loader:
-            X_cutmix, y_cutmix = cf.cutmix_batch(X,
-                                                 y,
-                                                 num_classes=1000,
-                                                 alpha=1.0)
-
+            X_cutmix, y_perm, area, _ = cutmix_batch(X, y, alpha=0.2)
             y_hat = model(X_cutmix)
-            loss = loss_fn(y_hat, y_cutmix)
+            loss = area * loss_fn(y_hat, y) + (1 - area) * loss_fn(y_hat, y_perm)
```
eracah, line 34: Thanks for updating the readme to reflect the change! Are we okay with adding extra logic for the user (outside of just calling the algorithm's functional implementation)? Cory and Landan, you know better than me: is our philosophy that if you're using functional then you have to do a little extra work around the edges?

composer/algorithms/cutmix/README.md
```diff
@@ -28,13 +29,9 @@ def training_loop(model, train_loader):
 
     for epoch in range(num_epochs):
         for X, y in train_loader:
-            X_cutmix, y_cutmix = cf.cutmix_batch(X,
-                                                 y,
-                                                 num_classes=1000,
-                                                 alpha=1.0)
-
+            X_cutmix, y_perm, area, _ = cutmix_batch(X, y, alpha=0.2)
```
eracah, line 32:
```suggestion
            X_cutmix, y_perm, area, _ = cutmix_batch(X, y, alpha=0.2)
```
```suggestion
            X_cutmix, y_perm, area, _ = cf.cutmix_batch(X, y, alpha=0.2)
```

composer/algorithms/cutmix/cutmix.py
```diff
@@ -210,48 +193,103 @@ class CutMix(Algorithm):
...
+            self._indices = _gen_indices(input)
+
+            _cutmix_lambda = _gen_cutmix_coef(alpha)
+            self._bbox = _rand_bbox(input.shape[2],
+                                    input.shape[3],
+                                    _cutmix_lambda,
+                                    uniform_sampling=self._uniform_sampling)
+            self._adjusted_lambda = _adjust_lambda(input, self._bbox)
+
+            new_input, self._permuted_target, _, _ = cutmix_batch(input=input,
```
eracah, line 241: do we need `cutmix_batch` to return `bbox`? I don't see it being used anywhere in the code or examples

composer/algorithms/cutmix/cutmix.py
```diff
@@ -210,48 +193,103 @@ class CutMix(Algorithm):
...
+                                                                  target=target,
+                                                                  alpha=self.alpha,
+                                                                  bbox=self._bbox,
+                                                                  indices=self._indices,
+                                                                  uniform_sampling=self._uniform_sampling)
+
+            state.batch_set_item(key=self.input_key, value=new_input)
+
+        if not self.interpolate_loss and event == Event.BEFORE_LOSS:
```
eracah, line 250: does the interpolating of the target have to happen at BEFORE_LOSS. Can't it happen at BEFORE_FORWARD along with input transformation? Seems like that would make the logic simpler.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -210,48 +193,103 @@ class CutMix(Algorithm):
...
+            # Grab the loss function
+            if hasattr(state.model, 'loss'):
+                loss_fn = state.model.loss
+            elif hasattr(state.model, 'module') and hasattr(state.model.module, 'loss'):
+                if isinstance(state.model.module, torch.nn.Module):
+                    loss_fn = state.model.module.loss
+                else:
+                    raise TypeError('state.model.module must be a torch module')
+            else:
+                raise AttributeError('Loss must be accesable via model.loss or model.module.loss')
```
eracah, line 282:
```suggestion
                raise AttributeError('Loss must be accessible via model.loss or model.module.loss')
```

eracah (approved):
Thanks, @coryMosaicML! Looks good. I just had a few nits
I didn't know about this alternative approach to interpolate the loss. I'm curious what the benefit is to interpolate the loss is over interpolating the target. The math is the same right?

coryMosaicML:
> I didn't know about this alternative approach to interpolate the loss. I'm curious what the benefit is to interpolating the loss over interpolating the target. The math is the same right?

This is a great question. So, the math is the same if the loss function is cross entropy (or any other loss that is a linear function of the targets) but it won't be the same for other loss functions like MSE. There are two main reasons to make this change:
1. Running with `interpolate_loss=True` doesn't force the conversion to one-hot labels. This is a savings in memory and compute. For imagenet+resnet50 the cost of the conversion is negligible, but for cases where there are a large number of classes, or something like segmentation where there are many labels per image, there can be a noticeable slowdown and/or increased memory usage.
2. Running with `interpolate_loss=True` doesn't require the loss to accept one-hots. Frequently people use loss functions that expect index targets only. For example until recently this is how PyTorch's native cross entropy loss was implemented. So this change provides the option to use an existing loss function like that.

There are also a few big downsides to running with `interpolate_loss=True`:
1. This is not math equivalent for most loss functions, but things should still run just fine. I have no idea what this will do to generalization though, so care should be taken using this option with other loss functions.
2. Composition with other methods won't work as expected, specifically with our implementation of MixUp also running with `interpolate_loss=True`. Hence the warning should be raised in this case. Again, effects on generalization are unknown, though I don't think people frequently use CutMix and MixUp together like this.
3. Now, the loss function has to be computed twice, which could be undesirable if it's expensive.

composer/algorithms/cutmix/README.md
coryMosaicML: This is a good question, and I'm a little torn on it. This definitely makes the functional implementation less clean to allow for the functional and composer algorithm implementations to share the same code. The other option here would be to have a totally separate functional implementation that preserves the old behavior. I did things this way to allow for the `interpolate_loss=True` behavior to be replicated functionally if one so chooses, but open to other ideas here too.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -235,23 +224,74 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
-            indices=self._indices,
-        )
-
-        state.batch_set_item(key=self.input_key, value=new_input)
-        state.batch_set_item(key=self.target_key, value=new_target)
+        if event == Event.BEFORE_FORWARD:
+            if not isinstance(input, torch.Tensor):
+                raise NotImplementedError('Multiple tensors for inputs not supported yet.')
+            if not isinstance(target, torch.Tensor):
+                raise NotImplementedError('Multiple tensors for targets not supported yet.')
```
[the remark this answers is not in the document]
coryMosaicML: Done!

composer/algorithms/cutmix/cutmix.py
```diff
@@ -235,23 +224,74 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
+
+            # these are saved only for testing
+            self._indices = _gen_indices(input)
+
+            _cutmix_lambda = _gen_cutmix_coef(alpha)
+            self._bbox = _rand_bbox(input.shape[2],
+                                    input.shape[3],
+                                    _cutmix_lambda,
+                                    uniform_sampling=self._uniform_sampling)
+            self._adjusted_lambda = _adjust_lambda(_cutmix_lambda, input, self._bbox)
```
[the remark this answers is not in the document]
coryMosaicML: Done!

composer/algorithms/cutmix/cutmix.py
coryMosaicML: I think it needs to return `bbox` such that users of the functional interface can replicate what cutmix does to the target tensors when they have spatial dimensions. The composer algorithm implementation handles this automatically, but it would have to be done manually with the current functional implementation.

composer/algorithms/cutmix/cutmix.py
coryMosaicML: I think in principle it could be done on `BEFORE_FORWARD`, but doing it this way would require the user to specify `num_classes` as a hyperparam. Doing it on `BEFORE_LOSS` lets the number of classes be inferred automatically from the shape of the model outputs. Though, there's a tradeoff here between the ease of use of not having to specify `num_classes` and how robust the automatic conversion is.

composer/algorithms/cutmix/cutmix.py
```diff
@@ -235,23 +224,74 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
...
+                raise ValueError("Can't interpolate loss when target has the same height and width as the input")
+
+            # Grab the loss function
+            if hasattr(state.model, 'loss'):
+                loss_fn = state.model.loss
+            elif hasattr(state.model, 'module') and hasattr(state.model.module, 'loss'):
+                if isinstance(state.model.module, torch.nn.Module):
+                    loss_fn = state.model.module.loss
+                else:
+                    raise TypeError('state.model.module must be a torch module')
```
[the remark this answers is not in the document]
coryMosaicML: I think there was an edge case here with how the model was wrapped for DDP (possibly DeepSpeed?) but I'm not sure if it still exists.

PR #1231, opened 2022-06-28
Fix StreamingDataset compression with multi-rank

milocress:
This PR addresses 2 issues:

1) When multiple processes use a file, one of them downloads the file, and the others wait (by design there should be no download conflicts). When waiting for a compressed file, it is important to wait for the **uncompressed** local file to exist, not the compressed one, because that is when the shard is ready to use.

2) When a remote dataset is missing its`compression.txt` file, the rank0 process notices right away with a `FileNotFoundError` and proceeds. But the other ranks, which are waiting for rank0 to complete it's download, may wait up to 3min (3 tries * 60s) to understand that the file is missing and proceed. The `dist.broadcast` behavior change here will make the init time in this case much faster.

### DIFF
--- a/composer/datasets/streaming/dataset.py
+++ b/composer/datasets/streaming/dataset.py
@@ -138,10 +138,10 @@ class StreamingDataset(IterableDataset):
         self.timeout = timeout
         self.batch_size = batch_size
 
-        compression_basename = get_compression_scheme_basename()
-        if remote is not None:
+        self.compression_scheme = None
+        if remote is not None and dist.get_local_rank() == 0:
             try:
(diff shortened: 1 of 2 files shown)


abhi-mosaic:
Confirmed that this works with our multi-rank streaming tests:

```
 make test-dist WORLD_SIZE=2 EXTRA_ARGS="-s -m daily tests/datasets/test_streaming.py"      
```

As well as an end-to-end BERT run with 8 GPUs, with both a compressed and uncompressed remote dataset.

abhi-mosaic (approved):
LGTM, prioritizing speed here to unblock researchers working off `dev`.

PR #1232, opened 2022-06-28
Hotfix for S3 FileNotFoundError

abhi-mosaic:
Tested S3 dataset build success with 
``` 
pytest -s -m remote "tests/datasets/test_streaming_remote.py::test_streaming_remote_dataset[val-ade20k]"
```


milocress (commented):
This makes sense

PR #1234, opened 2022-06-29
Increase default timeout for `StreamingC4` to 120s

abhi-mosaic:
This PR increases the default timeout and makes it configurable in the `StreamingC4Hparams` class.

E2E test pending, waiting for available GPUs.


moinnadeem (approved):
LGTM, thanks Abhi! Took a nap and finally feeling better

PR #1236, opened 2022-06-29
Autograd accum fix

linden-li:
This PR is a first attempt at fixing a synchronization bug with automatic gradient accumulation. There is an issue where some ranks perform their micro-batching loops and block when reducing their gradients while another rank has OOM'd. The fix right now tries to call a single backward pass in `_train_microbatch` in every iteration. If there is a failure in any of the ranks, it raises a `CUDA out of memory` issue, which is caught in the original logic.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1656,11 +1656,7 @@ class Trainer:
                 else:
                     raise
 
-            # Propagate across all ranks if any rank hit CUDA OOM
-            should_handle_cuda_oom = self._device.tensor_to_device(
-                torch.tensor([should_handle_cuda_oom], dtype=torch.uint8))
-            dist.all_reduce(should_handle_cuda_oom, reduce_operation='MAX')
-            if int(should_handle_cuda_oom.item()) == 1:
+            if should_handle_cuda_oom == 1:
                 # If any rank hit CUDA OOM, update grad_accum and retry. Ignore any caught_timeout_error since
                 # it is likely transient, e.g. timeout because certain ranks OOMed and didn't reach barrier.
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -1717,8 +1713,21 @@ def _train_microbatches(self, microbatches: Sequence[Batch], ddp_sync: bool = Tr
...
+                    if self.adaptive_gradient_accumulation and _is_cuda_oom(e):
+                        fail = 1
+                    else:
+                        raise
+
+                fail = self._device.tensor_to_device(torch.tensor([fail], dtype=torch.int8))
+                dist.all_reduce(fail, reduce_operation='SUM')
+
+                if fail.item() > 0:
+                    raise RuntimeError('CUDA out of memory')
```
ravi-mosaicml, line 1730: Would it work to call the handle_cuda_oom here?

ravi-mosaicml (commented):
Can you add a test case that would cause a hang before, and then that passes now with this fix? For the test setup, I think it could be done through an algorithm / callback that (massively) increases the batch size on rank 1 while keeping it the same on rank 0.

If it can wait, going to defer to @mvpatel2000 for approval. Otherwise, let's meet and reason through the distributed logic together to make sure we fix it once and for all.

mvpatel2000:
Moving the discussion for the distributed logic offline -- it seems a little complicated

mvpatel2000:
Closing this PR -- we discussed offline about a couple subtleties this doesn't address. Looks like this is pretty tricky, so I'll throw up a design doc to discuss a few alternatives

PR #1239, opened 2022-06-29
Fix bug in streaming iteration/downloading, refactor

knighton:
This PR simplifies some of the core StreamingDataset logic and improves throughput in situations where the total # of shards and # samples is large. We also fix a bug that used to occur when the download status was set to `NOT_STARTED`.


abhi-mosaic:
Going to launch an E2E test with BERT+C4 as the shard insetion is going to be slower now, and I want to make sure that it doesn't affect throughput significantly. I don't expect it will, but just want to be certain.

abhi-mosaic (approved):
All tests are passing now, and E2E tests on both BERT+C4 and R50+ImageNet look ok so approving this PR.

PR #1243, opened 2022-06-30
[XS] Updating console progress_bar logger to use max_duration units

moinnadeem:
In #1190, we made a design decision to have the progress bar reflect whatever units `max_duration` was in. This just reflects that for console progress bars.

Tested on BERT YAMLs and everything looks right. 
Here is a sample output:
```
[FIT][SAMPLE=0/275184000]: { "trainer/algorithms": [], }
[EPOCH][SAMPLE=0/275184000]: { "epoch": 0, }
[BATCH][SAMPLE=0/275184000]: { "trainer/global_step": 0, "trainer/batch_idx": 0, }
[BATCH][SAMPLE=0/275184000]: { "trainer/grad_accum": 16, }
[BATCH][SAMPLE=0/275184000]: { "loss/train": 10.5169, }
[BATCH][SAMPLE=20480/275184000]: { "wall_clock/train": 113.6195, "wall_clock/val": 0.0000, "wall_clock/total": 113.6195, }
[BATCH][SAMPLE=20480/275184000]: { "lr-AdamW/group0": 0.0000, }
[BATCH][SAMPLE=20480/275184000]: { "trainer/global_step": 5, "trainer/batch_idx": 5, }
[BATCH][SAMPLE=20480/275184000]: { "trainer/grad_accum": 16, }
[BATCH][SAMPLE=20480/275184000]: { "loss/train": 10.4930, }
```


composer/loggers/progress_bar_logger.py
```diff
@@ -127,6 +133,7 @@ def __init__(
         log_to_console: Optional[bool] = None,
         console_log_level: Union[LogLevel, str, Callable[[State, LogLevel], bool]] = LogLevel.EPOCH,
         stream: Union[str, TextIO] = sys.stderr,
+        log_interval: int = 1,
```
ravi-mosaicml, line 136:
```suggestion
        log_interval: Union[LogLevel, str]= LogLevel.BATCH,
```

ravi-mosaicml (commented):
The formatting changes for samples/tokens are fine. However, the log interval changes also would need to be propegated through the trainer init signature and trainer hparams. I'd recommend leaving it out and instead just include the formatting changes in this PR.

composer/loggers/progress_bar_logger.py
```diff
@@ -189,7 +189,19 @@ def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]) -> N
         # log to console
         if self.should_log(state, log_level):
             data_str = format_log_data_value(data)
-            log_str = f'[{log_level.name}][batch={int(state.timestamp.batch)}]: {data_str}'
+            assert state.max_duration is not None, 'max_duration should be set'
```
ravi-mosaicml, line 192: If something is logged on `Event.INIT`, then max duration could be none, so probs best not to prepend it?

composer/loggers/progress_bar_logger.py
```diff
@@ -189,7 +189,19 @@ def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]) -> N
...
+
+            if state.max_duration.unit == TimeUnit.EPOCH:
+                assert state.dataloader_len is not None
```
ravi-mosaicml, line 195: Can we be graceful to dataloaders of unknown length?

abhi-mosaic: I think this may be OK though because if `max_duration` is in epochs, the dataloader is required to be sized right? I think when I use infinite-sized IterableDatasets, I would expect Trainer to forbid me from using `max_duration` in units of epochs.

bandish-shah:
Going to mark comments as resolved for now so we can get this into the release.  If you think there's an issue we need to address let's get the required Jira's filed and we can fix in 0.8.1 @moinnadeem @ravirahman @abhi-mosaic

bandish-shah (approved):
Let's go!

PR #1244, opened 2022-06-30
Add device_id param to DeviceGPU 

ishanashastri:
Allows one to specify an optional id of a GPU device to use for training by introducing an optional param in the `__init()__` function. Default behavior is unaffected.


ravi-mosaicml (approved):
Lgtm!

PR #1245, opened 2022-06-30
Catching ChannelErrors on SFTP Failures

moinnadeem:
Addresses https://mosaicml.atlassian.net/browse/CO-639, which contains the traceback that was caused by SFTP object store errors.

We should also verify that this is the right place to add the exception.

I was able to re-run and have not seen another crash in ~30 minutes, but this exception is pretty rare to being with.


abhi-mosaic (approved):
LGTM. My understanding is that this change will ensure that we go through the retry loop if `ChannelException` is raised. 

Best case, it works on the next try and we continue. Worst case, it fails `max_retries` in a row and we exit like before.

PR #1246, opened 2022-06-30
Make `StreamingDataset` compression file easier to write/read

abhi-mosaic:
Since our compression metadata is stored in a `.txt` file, it seems to imply that one could easily edit it. For example, I was trying to create some empty `compression.txt` files to upload to our older StreamingDatasets, for cleanliness.

But most text editors add a trailing newline when you try to save `.txt` files. They also display the file strangely if you open it and no newline exists.

What do you guys (@knighton, @milocress) think about adding this newline generally so that `compression.txt` is easily editable?

### DIFF
--- a/composer/datasets/streaming/dataset.py
+++ b/composer/datasets/streaming/dataset.py
@@ -139,11 +139,11 @@ class StreamingDataset(IterableDataset):
         self.batch_size = batch_size
 
         self.compression_scheme = None
-        if remote is not None and dist.get_local_rank() == 0:
+        if remote is not None and dist.get_global_rank() == 0:
             try:
                 compression_local = self._download_file(get_compression_scheme_basename())
                 with open(compression_local, 'r') as fp:
-                    compression_scheme = fp.read()
+                    compression_scheme = fp.read().rstrip('\n')
                     self.compression_scheme = compression_scheme if compression_scheme != '' else None
                     if remote == local and self.compression_scheme is not None:
                         raise DatasetCompressionException('cannot decompress when remote == local')
--- a/composer/datasets/streaming/dataset.py
+++ b/composer/datasets/streaming/dataset.py
@@ -143,7 +143,7 @@ class StreamingDataset(IterableDataset):
             try:
                 compression_local = self._download_file(get_compression_scheme_basename())
                 with open(compression_local, 'r') as fp:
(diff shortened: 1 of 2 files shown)


composer/datasets/streaming/dataset.py
```diff
@@ -139,11 +139,11 @@ def __init__(self,
         self.batch_size = batch_size
 
         self.compression_scheme = None
-        if remote is not None and dist.get_local_rank() == 0:
+        if remote is not None and dist.get_global_rank() == 0:
```
abhi-mosaic, line 142: Also, minor bug here that I realized as I was editing.

milocress:
I don't see a case where a user editing the `compression.txt` would end well. Potentially we could add a feature where modifying this file would cause the writer to {re/de}compress with a new alg, but I don't think we should give users a sense that this file is human writable.

+1 for the newline to make the file more readable, though.

composer/datasets/streaming/dataset.py
```diff
@@ -139,11 +139,11 @@ def __init__(self,
...
             try:
                 compression_local = self._download_file(get_compression_scheme_basename())
                 with open(compression_local, 'r') as fp:
-                    compression_scheme = fp.read()
+                    compression_scheme = fp.read().rstrip('\n')
```
knighton, line 146: or just `.rstrip()` (msft newlines)

knighton:
Sounds good!

knighton (approved):

milocress (approved):

abhi-mosaic:
> I don't see a case where a user editing the `compression.txt` would end well. Potentially we could add a feature where modifying this file would cause the writer to {re/de}compress with a new alg, but I don't think we should give users a sense that this file is human writable.

Yeah that's true... I think one small win is that I can easily modify an existing StreamingDataset and make it compressed with unix commands now, like `pigz *` and then `echo 'gz' >> compression.txt` and edit it. Which is pretty cool!

composer/datasets/streaming/dataset.py
abhi-mosaic: done

milocress:
> I can easily modify an existing StreamingDataset and make it compressed with unix commands now, like `pigz *` and then `echo gz >> compression.txt`. Which is pretty cool!

Agreed this is super cool

PR #1247, opened 2022-06-30
Add pytorch 1.12.0 docker image

linden-li:
This PR adds support for PyTorch 1.12 to the build matrix. The new build matrix is:

- Python 3.9 + torch 1.12
- Python 3.8 + torch 1.11
- Python 3.7 + torch 1.10

TODO:

- [x] Merge https://github.com/mosaicml/composer/pull/1275


ravi-mosaicml (commented):
The generate_build_matrix.py script will need to be changed. Run that locally then commit the result.

ravi-mosaicml (commented):
Since pytorch 1.9 is now 4 versions outdated, let's drop support for it?

Thoughts about making the build matrix something like this?  CC @bandish-shah 

* Python 3.9 + torch 1.12
* Python 3.8 + torch 1.11
* Python 3.7 + torch 1.10

If we do drop testing for torch 1.9, can you update the `setup.py` and `meta.yaml` to reflect the minimum torch version of 1.10?

I think the torchvision version will also need to be updated for pytorch 1.12 -- usually they release them in parallel. https://pypi.org/project/torchvision/0.13.0/

moinnadeem:
@ravi-mosaicml I wanted to flag a potential regression before we get far with PyTorch 1.12: https://mosaicml.atlassian.net/browse/CO-328

Basically, 1.12 disables TF32 matrix multiplications by default, so it might be slower. If we support 1.12, we should add a flag in Composer that checks your PyTorch version and re-enables TF32 (as long as we believe in it).

abhi-mosaic:
> @ravi-mosaicml the proposed build matrix looks good to me.
> 
> We should get TF32 default change in asap as this will likely lead to performance regressions.
> 
> Should we get the soft cross entropy deprecation warning in for 0.8.1 and deprecate our implementation in 0.9.0?

I'm working on a precision refactor, and was gonna add`fp32` and `tf32` options within Composer. Does this solve the issue? Or you still want to make all our Docker images set `tf32` by default at the image level?

hanlint:
This PR scope is to add the pytorch 1.12.0 docker image, lets not expand the scope to other work (e.g. deprecating pytorch versions) which is a composer-wide setting.

ravi-mosaicml:
I would normally agree, but upgrading the image to pytorch 1.12 will introduce a regression with the tf32 / fp32 change for anyone using the latest mosaicml/pytorch image. Synced with @abhi-mosaic this morning; thinking of just adding in `tf32` without the precision refactor; then can merge this in. The additional refactoring of the precision enum will come later and will not block the pytorch 1.12 image.

docker/build_matrix.yaml
```diff
@@ -36,46 +36,46 @@
...
   CUDA_VERSION: 11.3.1
   PYTHON_VERSION: '3.9'
-  PYTORCH_VERSION: 1.11.0
+  PYTORCH_VERSION: 1.12.0
   TAGS:
-  - mosaicml/pytorch:1.11.0_cu113-python3.9-ubuntu20.04
+  - mosaicml/pytorch:1.12.0_cu113-python3.9-ubuntu20.04
   - mosaicml/pytorch:latest
   TARGET: pytorch_stage
   TORCHVISION_VERSION: 0.12.0
```
ravi-mosaicml, line 44: I think the torchvision version will need an update -- they usually release a new torchvision version for each pytorch version.

docker/generate_build_matrix.py
```diff
@@ -21,7 +21,7 @@
 
 def _get_pytorch_version(python_version: str):
     if python_version == '3.9':
-        return '1.11.0'
+        return '1.12.0'
     if python_version in '3.8':
         return '1.10.2'
     if python_version == '3.7':
```
ravi-mosaicml, line 27: See the comment re the build matrix -- can we bump python 3.7 to torch 1.10, and python 3.8 to torch 1.11?

docker/generate_build_matrix.py
```diff
@@ -32,7 +32,7 @@ def _get_pytorch_version(python_version: str):
 def _get_torchvision_version(pytorch_version: str):
     if pytorch_version == '1.10.2':
         return '0.11.3'
-    if pytorch_version == '1.11.0':
+    if pytorch_version == '1.12.0':
         return '0.12.0'
```
ravi-mosaicml, line 36: My guess is we need 0.13.0 for torch 1.12?
```suggestion
        return '0.13.0'
```

ravi-mosaicml (commented):
See comments: re pytorch + python + torchvisions versions. Will hold off on approving until the tf32 commit is merged.

docker/generate_build_matrix.py
linden-li: Yes, torchvision should be upgraded to 0.13.0

docker/generate_build_matrix.py
```diff
@@ -21,21 +21,21 @@
...
 def _get_torchvision_version(pytorch_version: str):
-    if pytorch_version == '1.10.2':
-        return '0.11.3'
     if pytorch_version == '1.11.0':
         return '0.12.0'
-    if pytorch_version == '1.9.1':
-        return '0.10.1'
+    if pytorch_version == '1.12.0':
+        return '0.13.0'
+    if pytorch_version == '1.10.0':
```
ravi-mosaicml, line 37:
```suggestion
    if pytorch_version == '1.10.2':
```

Also nit: can you you sort these checks in order? A bit confusing to jump from 1.11 to 1.12 and back to 1.12.

ravi-mosaicml (approved):
LGTM! Thanks for doing this and dealing with all of the bugs that came up.

PR #1248, opened 2022-06-30
object stores for streaming datasets

milocress:
This PR 
* uses object stores for downloading streaming datasets
* allows users to specify an object store or connection string as a streaming upload destination.


composer/datasets/streaming/download.py
```diff
@@ -82,26 +67,16 @@ def download_from_sftp(remote: str, local: str) -> None:
     # Default port
     port = port if port else 22
 
-    # Local tmp
-    local_tmp = local + '.tmp'
-    if os.path.exists(local_tmp):
-        os.remove(local_tmp)
+    object_store = SFTPObjectStore(
```
ravi-mosaicml, line 70: Is it possible to incorporate the timeout into the SFTP object store? No worries if not...

composer/datasets/streaming/download.py
```diff
@@ -82,26 +67,16 @@ def download_from_sftp(remote: str, local: str) -> None:
...
+        host=hostname,
```
ravi-mosaicml, line 71: The SFTPObjectStore accepts taking a uri for the `host` argument, so the manual parsing logic above can be removed

composer/datasets/streaming/download.py
```diff
@@ -27,23 +27,13 @@ def download_from_s3(remote: str, local: str, timeout: float) -> None:
...
-        s3.download_file(obj.netloc, obj.path.lstrip('/'), local)
-    except ClientError as e:
-        _ensure_not_found_errors_are_wrapped(remote, e)
+    client_config = {'read_timeout': timeout}
+    bucket, key = obj.netloc, obj.path.lstrip('/')
+    object_store = S3ObjectStore(bucket=bucket, client_config=client_config)
+    object_store.download_object(key, local)
 
 
 def download_from_sftp(remote: str, local: str) -> None:
```
ravi-mosaicml, line 39: Curious if we still need the `download_from_sftp` / `download_from_s3` / `download_from_local` helper functions, or whether it would be possible to use `composer.utils.file_helpers.get_file` instead? Thinking that we could make the `get_file` helper function smarter to handle `s3://` and `sftp://` URIs (and auto-construct the appropriate object store if a URI begins with one of these schemes). Realize that this is slightly beyond the scope of this PR, so OK to defer to another one, but would like `get_file` to be smart and be able to handle `s3://` and `sftp://` URIs too (this will help with checkpointing).

composer/datasets/streaming/upload.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+import urllib.parse
+from typing import Optional
+
+from composer.utils.object_store.object_store import ObjectStore
+from composer.utils.object_store.s3_object_store import S3ObjectStore
+from composer.utils.object_store.sftp_object_store import SFTPObjectStore
+
+
+def get_object_store(remote: str, timeout: Optional[float]) -> ObjectStore:
```
ravi-mosaicml, line 13: Nit: this can be a private helper
```suggestion
def _get_object_store(remote: str, timeout: Optional[float]) -> ObjectStore:
```

composer/datasets/streaming/upload.py
```diff
@@ -0,0 +1,69 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import urllib.parse
+from typing import Optional
+
+from composer.utils.object_store.object_store import ObjectStore
+from composer.utils.object_store.s3_object_store import S3ObjectStore
+from composer.utils.object_store.sftp_object_store import SFTPObjectStore
+
```
ravi-mosaicml, line 11: Need `__all__`. Or thoughts about cominding this into the `writer.py` file?

composer/datasets/streaming/writer.py
```diff
@@ -12,9 +12,11 @@
 
 import numpy as np
 from tqdm import tqdm
+from upload import get_object_store
```
ravi-mosaicml, line 15: Nit: Use absolute imports
```suggestion
from composer.datasets.streaming.upload import get_object_store
```

composer/datasets/streaming/writer.py
```diff
@@ -90,7 +92,9 @@ def __init__(self,
                  dirname: str,
                  fields: List[str],
                  shard_size_limit: int = 1 << 24,
-                 compression: Optional[str] = default_compression) -> None:
+                 compression: Optional[str] = default_compression,
+                 remote: Union[ObjectStore, str, None] = None,
```
ravi-mosaicml, line 96: nit: prefer `Optional[...]` over `Union[None, ...]`

```suggestion
                 remote: Optional[Union[ObjectStore, str]] = None,
```

composer/datasets/streaming/writer.py
```diff
@@ -90,7 +92,9 @@ def __init__(self,
...
+                 timeout: Optional[float] = None) -> None:
```
ravi-mosaicml, line 97: This is more of a design question, but I'm not thrilled about propagating the timeout setting all the through. My thinking is that:

1. Not all object stores support timeout -- currently, just boto3 does.
2. There are other config settings of equal importance (namly, retries), and propagating all these settings through would complicate public API -- if one wants this amount of customization (i.e. more params than can be specified via the URI), then the can pass an already-constructed `ObjectStore` instance through.

(This comment applies to the other functions that take a `timeout`, except for `download_or_wait`.)

Thoughts?

ravi-mosaicml (commented):
Overall looks good! See comments.

Can you add some test cases? In addition, can you update the facesynthetics notebook to show this functionality?

composer/datasets/streaming/download.py
milocress: My understanding is the `get_file` utility function is being refactored by @eracah -- happy to take a look but I don't want to make a conflict unnecessarily, can we defer this to the next PR?

composer/datasets/streaming/writer.py
milocress: agreed, this makes sense. Let's have the connection strings and remove the `timeout` argument for now -- if the user wants to specify it manually, they can pass an `ObjectStore` as an argument.

tests/datasets/test_streaming.py
```diff
@@ -47,12 +47,14 @@ def get_fake_samples_decoders(num_samples: int) -> Tuple[List[Dict[str, bytes]],
 def write_synthetic_streaming_dataset(dirname: str,
                                       samples: List[Dict[str, bytes]],
                                       shard_size_limit: int,
-                                      compression: Optional[str] = StreamingDatasetWriter.default_compression) -> None:
+                                      compression: Optional[str] = StreamingDatasetWriter.default_compression,
```
ravi-mosaicml, line 50: Same comment re: hard-coding the default in here.

ravi-mosaicml (approved):
LGTM, see comments. Thanks!

PR #1252, opened 2022-07-01
BERT and C4 updates

abhi-mosaic:
I've got a couple BERT+C4 usability improvements here that I packaged to avoid 3 tiny PRs, but if there's a lot of discussion I can split it up. Opening in draft mode as well.

1) (@knighton) Make the `StreamingDatasetWriter` shard flushing async, which helps overlap work a bit when writing `.mds` datasets. I see a small but meaningful perf improvement, and no more intermittent "pausing" in the progress bar. @knighton

2) (@knighton) Improve our C4 to `.mds` conversion example so that it all happens via streaming and ~saves to compressed `gz` format.~ not doing this right now b/c our compression during writing is way too slow, and we need to use pigz-python.

I added a quick wrapper around the HF dataset to make it multi-worker compatible, and now we don't need to pre-download the dataset any more. The entire train dataset (360M samples, 770GB of text) can be converted in about 2hr with `num_workers=8`, and about 1.1hr with `num_workers=64`. Results may vary depending on system, and the latter result is probably an indicator of the limits of our single-stream writer, but this seems pretty good to me! 

 * Note: since HF Datasets 2.3.0, the streaming dataset multiprocessing issue [might be fixed upstream](https://github.com/huggingface/datasets/pull/4375), so we should investigate whether it's time to upgrade to HF datasets 2.0+ and refactor our HF C4 implementation to take advantage of this new feature. In general, we should keep a closer eye here... It appears HF even has a [streaming ImageNet1k dataset now.](https://huggingface.co/datasets/imagenet-1k). 

3) (@moinnadeem) Update the `bert-base-uncased` YAML to use a power-of-2 batch size, a max_duration that converts cleanly to 70k batches, and use the new compressed public C4 `StreamingDataset`

### DIFF
--- a/scripts/mds/c4.py
+++ b/scripts/mds/c4.py
@@ -86,7 +86,7 @@ def each(dataset: IterableDataset) -> Iterable[Dict[str, bytes]]:
 
 
 def main(args: Namespace) -> None:
-    """Main: create C4 streaming dataset, with 'gz' compression.
+    """Main: create C4 streaming dataset.
 
     Args:
         args (Namespace): Commandline arguments.
@@ -104,7 +104,7 @@ def main(args: Namespace) -> None:
         with StreamingDatasetWriter(dirname=os.path.join(args.out_root, split_new_name),
                                     fields=fields,
                                     shard_size_limit=args.shard_size_limit,
-                                    compression='gz') as out:
+                                    compression=None) as out:
             out.write_samples(samples=each(dataset), use_tqdm=bool(args.tqdm), total=expected_num_samples)
 
 
--- a/scripts/mds/c4.py
+++ b/scripts/mds/c4.py
@@ -13,6 +13,7 @@ from typing import Any, Dict, Iterable, List, Tuple
 import datasets
 import torch
 from datasets import Dataset
+from torch.utils.data import DataLoader, IterableDataset, get_worker_info
 
 from composer.datasets.streaming import StreamingDatasetWriter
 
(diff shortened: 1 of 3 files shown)


abhi-mosaic:
TODO: going to launch 3 seeds with the new `bert-base-uncased` YAML to make sure it matches expectations for convergence.

abhi-mosaic:
Ran 3 seeds each with AMP_FP16, AMP_BF16, and DeepSpeed_FP16, all runs look like they successfully converged, with MLM accuracy > 67.2%
Screen Shot 2022-07-05 at 10 50 06 AM

scripts/mds/c4.py
```diff
@@ -47,14 +67,17 @@ def each(dataset: Dataset) -> Iterable[Dict[str, bytes]]:
...
-    num_workers = 64
+    num_workers = min(64, dataset.num_shards())
     batch_size = 512
-    prefetch_factor = max(1, 2 * batch_size // num_workers)
-
-    loader = torch.utils.data.DataLoader(dataset=dataset,
-                                         batch_size=batch_size,
-                                         num_workers=num_workers,
-                                         prefetch_factor=prefetch_factor)
+    prefetch_factor = max(1, 2 * batch_size // num_workers) if num_workers > 0 else 2
```
knighton, line 72: explanatory comment here?

knighton (approved):

knighton:
Obligatory remark that we should eventually be thorough about our options wrt compression latency, compressed size, decompression latency

moinnadeem (approved):
LGTM! Could we re-run MNLI and GLUE numbers so we have an updated baseline to reference?

abhi-mosaic:
Reverting `writer.py` changes as I'm seeing test issues where the child threads have not completed writing their shards. Will follow up later with a more robust multiprocessing solution.

Will also kick off some new fine-tuning baseline runs but going to merge this now to get out of the way as other `StreamingDatset` stuff is landing soon.

PR #1253, opened 2022-07-01
Automatic Stochastic depth on residual blocks

dskhudia:
Detect residual blocks automatically and replace them with a stochastic version. See attached graphs for how transformation happens on ResNet50. 

~~TODO: Add tests.~~ Done

**Orig:** 
orig_resnet50

**Split into bottleneck and downsample:**
split_resnet50

**Converted with a call to block_stochastic_module:**
split_stochastic_resnet50

### DIFF
--- a/composer/utils/fx_utils.py
+++ b/composer/utils/fx_utils.py
@@ -8,18 +8,20 @@ Provides utilities to do FX-based model transformations.
 
 import logging
 import operator
-from typing import Any, Callable, Dict, List, Mapping, Tuple, Union
+import re
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
-from torch.fx import Node
-from torch.fx.graph_module import GraphModule
+from torch.fx import GraphModule, Node
+from torch.fx.passes.split_utils import split_by_tags
 
+from composer.algorithms.stochastic_depth.stochastic_layers import BlockStochasticModule
 from composer.utils import ensure_tuple
 
 log = logging.getLogger(__name__)
 
-__all__ = ['count_op_instances', 'replace_op', 'fuse_parallel_linears']
+__all__ = ['count_op_instances', 'replace_op', 'fuse_parallel_linears', 'apply_stochastic_residual']
 
 
 def count_op_instances(gm: GraphModule, ops: Union[Callable, str, List[Union[Callable, str]]]) -> int:
@@ -111,28 +113,138 @@ def replace_op(gm: GraphModule, src_ops: Union[Callable, str, List[Union[Callabl
     return gm
 
 
-def detect_residual_pattern(gm: GraphModule):
-    """Search and replace the pattern with another.
+def _get_ancestors(node: Node) -> List[Node]:
+    ancestorNodes = []
+    while node.op != 'placeholder':
+        ancestorNodes.append(node)
+        node = node.all_input_nodes[0]
+    return ancestorNodes
+
+
+def _get_residual_block_nodes(nodeLHS: Node, nodeRHS: Node) -> Tuple[List[Node], List[Node]]:
+    """Walk backwards from nodeLHS and nodeRSH to the root and construct lists of their parents.
 
     Arguments:
-        gm (GraphModule): The source FX-traced graph.
+        nodeLHS (Node): left-hand side node for a binary operator
+        nodeRHS (Node): right-hand side node for a binary operator
 
     Returns:
-        GraphModule: Modified GraphModule.
+        (lhsAncestors, rhsAncestors): Two lists of nodes containing ancestors for ``nodeLHS`` and ``nodeRHS`` with
+            their common ancestors removed.
     """
-    raise NotImplementedError('detect_residual_pattern is currently not implemented.')
+    lhsAncestors = _get_ancestors(nodeLHS)
+    rhsAncestors = _get_ancestors(nodeRHS)
+
+    # Iterate from back and eliminate common nodes
+    while lhsAncestors and rhsAncestors and lhsAncestors[-1] == rhsAncestors[-1]:
+        lhsAncestors.pop()
+        rhsAncestors.pop()
+    lhsAncestors.reverse()
+    rhsAncestors.reverse()
+    return lhsAncestors, rhsAncestors
 
 
-def replace_residual_with_stochastic(gm: GraphModule):
-    """Replaces residual pattern with their stoachstic equivalent.
+def _attach_tag(nodes: List[Node], tag: str):
+    """Attach tag to the given nodes for the splitter."""
+    for node in nodes:
+        node.tag = tag  # type: ignore[attr-defined]
+
+
+def _tag_residual_nodes(gm: GraphModule) -> Tuple[List[str], int]:
+    """Tag nodes for splitting."""
+    # all nodes that are not a part of the residual blocks are tagged with "mainN_{count}".
+    # a tag is required for all nodes by split_by_tags
+    # Also an earlier tag can be repeated for later nodes.
+    count = 0
+    all_tags = []
+    # In this pass over all nodes, we just tag them
+    for node in gm.graph.nodes:
+        default_tag = f'mainN_{count}'
+        node.tag = default_tag
+        if default_tag not in all_tags:
+            all_tags.append(default_tag)
+        if node.op == 'call_function' and node.target in [torch.add, operator.add]:
+            assert len(node.all_input_nodes) == 2
+            node0, node1 = node.all_input_nodes[0], node.all_input_nodes[1]
+            lhs_nodes, rhs_nodes = _get_residual_block_nodes(node0, node1)
+            if lhs_nodes or rhs_nodes:
+                if len(lhs_nodes):
+                    _attach_tag(lhs_nodes, f'non_res_{count}')
+                    all_tags.append(f'non_res_{count}')
+                if len(rhs_nodes):
+                    _attach_tag(rhs_nodes, f'residual_{count}')
+                    all_tags.append(f'residual_{count}')
+                add_tag = f'addN_{count}'
+                if add_tag not in all_tags:
+                    all_tags.append(add_tag)
+                node.tag = add_tag
+                count += 1
+    return all_tags, count
+
+
+def _get_residual_modules(gm: GraphModule, node: Node) -> Tuple[Optional[GraphModule], Optional[GraphModule], int]:
+    """Returns GraphModules for the main and residual branches.
+
+    node.op is assumed to be a call_module
+    """
+    pattern = re.compile(r'non_res_(\d+)|residual_(\d+)')
+    matches = pattern.match(str(node.target))
+    if matches:
+        idx = int(matches[1]) if matches[1] else int(matches[2])
+        main_submod = getattr(gm, f'non_res_{idx}')
+        residual_submod = getattr(gm, f'residual_{idx}', None)
+        return main_submod, residual_submod, idx
+    else:
+        return None, None, 0
+
+
+def _replace_residual_pattern(gm: GraphModule,
+                              original_node: Node,
+                              replacement_module: str,
+                              has_residual_ops: bool = False) -> None:
+    """Replaces main, residual and add_node with the ``replacement_module``.
+
+    ``replacement_module`` is already added to the gm.
+    """
+    insert_node = original_node.prev
+    add_node = original_node.next
+    if has_residual_ops:
+        add_node = original_node.next.next
+    with gm.graph.inserting_after(insert_node):
+        new_node = gm.graph.call_module(replacement_module, args=(insert_node,))  # type: ignore
+        add_node.replace_all_uses_with(new_node)
+        gm.graph.erase_node(add_node)
+        if has_residual_ops:
+            gm.graph.erase_node(original_node.next)
+        gm.graph.erase_node(original_node)
+    gm.graph.lint()
+
+
+def apply_stochastic_residual(gm: GraphModule, drop_rate: float = 0.2) -> Tuple[GraphModule, int]:
+    """Detect and replace residual pattern with their stochastic equivalent.
 
     Arguments:
-        gm (GraphModule): The source FX-traced graph.
+        gm (GraphModule): The source FX-traced graph. It can be the whole model symbolically traced.
 
     Returns:
-        GraphModule: Modified GraphModule.
+        GraphModule: Modified GraphModule that has stochastic residual connections.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    if not isinstance(gm, GraphModule):
+        raise ValueError(
+            f'Input to apply_stochastic_residual should be an instance of GraphModule. Received {type(gm)}')
+    all_tags, count = _tag_residual_nodes(gm)
+    split_gm = split_by_tags(gm, all_tags)
+    for node in split_gm.graph.nodes:
+        if node.op != 'call_module':
+            continue
+
+        main_submod, residual_submod, idx = _get_residual_modules(split_gm, node)
+        if main_submod:
+            residual_st_instance = BlockStochasticModule(main_submod, residual_submod, drop_rate)
+            split_gm.add_submodule(f'resi_st_{idx}', residual_st_instance)  # type: ignore
+            _replace_residual_pattern(split_gm, node, f'resi_st_{idx}', residual_submod is not None)
+    split_gm.recompile()
+    return split_gm, count
 
 
 def _can_linears_be_fused(linear_nodes: List[Node], all_modules: Mapping[str, nn.Module]) -> bool:
--- a/composer/algorithms/stochastic_depth/stochastic_layers.py
+++ b/composer/algorithms/stochastic_depth/stochastic_layers.py
@@ -3,8 +3,14 @@
 
 """Stochastic forward functions for ResNet Bottleneck modules."""
 
+from typing import Optional
+
 import torch
+import torch.nn as nn
+from torch.fx import GraphModule
 from torchvision.models.resnet import Bottleneck
+
+__all__ = ['make_resnet_bottleneck_stochastic', 'BlockStochasticModule']
 
 
 def block_stochastic_forward(self, x):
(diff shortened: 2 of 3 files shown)


Landanjs (commented):
This is awesome! It LGTM from the stochastic depth side. A few nits up to your discretion

I'm a lot less confident on the fx graph surgery implementation. It looks reasonable to me, but not sure if there are alternative implementations? It would be great to get someone from eng to review this as well.

One hesitation before approving: would it be possible to create a test for `apply_stochastic_residual`? It be nice to check that the graph surgery went as expected for say ResNet18 and ResNet50 e.g. counting the number of stochastic blocks after surgery  and checking it aligns with the number of residual blocks before surgery

Landanjs:
Oh, you just pushed a test 😂 Is this still being updated?

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+    return lhsAncestors, rhsAncesotors
+
+
+def _tag_nodes(nodes: List[Node], tag: str):
+    """Tag nodes for the splitter."""
+    for node in nodes:
+        node.tag = tag  # type: ignore[attr-defined]
+
+
+def apply_stochastic_residual(gm: GraphModule, drop_rate: float = 0.2) -> GraphModule:
```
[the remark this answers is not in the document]
dskhudia: Clarified in docstring.

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+
+    Arguments:
+        nodeLHS (Node): left-hand side node for a binary operator
+        nodeRHS (Node): right-hand side node for a binary operator
+
+    Returns:
+        (lhsAncestors, rhsAncesotors): Two lists of nodes containing ancestors for ``nodeLHS`` and ``nodeRHS`` with
+            their common ancestors removed.
+    """
+    lhsAncestors = _get_ancestors(nodeLHS)
```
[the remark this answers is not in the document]
dskhudia: It goes till the root node. This is not the most efficient way but I think it's ok because graphs will have 100s of nodes rather than 1000s or millions of nodes. If it becomes a problem, we can revisit.

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
         gm (GraphModule): The source FX-traced graph.
 
     Returns:
-        GraphModule: Modified GraphModule.
+        GraphModule: Modified GraphModule that has stochastic residual connections.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    assert isinstance(gm, GraphModule), 'Input to apply_stochastic_residual should be an instance of GraphModule'
+    count = 0
+    # all nodes that are not a part of the residual blocks are tagged with "main_0".
```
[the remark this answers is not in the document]
dskhudia: Correct.

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+        if default_tag not in all_tags:
+            all_tags.append(default_tag)
+        if node.op == 'call_function' and node.target in [torch.add, operator.add]:
+            assert len(node.all_input_nodes) == 2
+            node0, node1 = node.all_input_nodes[0], node.all_input_nodes[1]
+            lhs_nodes, rhs_nodes = _get_residual_path(node0, node1)
+            if lhs_nodes or rhs_nodes:
+                if len(lhs_nodes):
+                    _tag_nodes(lhs_nodes, f'bottleneck_{count}')
+                    all_tags.append(f'bottleneck_{count}')
```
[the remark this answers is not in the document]
dskhudia: Agreed. Let me change it to residual_bl

dskhudia:
> Oh, you just pushed a test 😂 Is this still being updated?

Just the test was missing so added it a little while ago.

dskhudia:
>  It be nice to check that the graph surgery went as expected for say ResNet18 and ResNet50 e.g. counting the number of stochastic blocks after surgery and checking it aligns with the number of residual blocks before surgery

Added a test for resnet18 that counts residual blocks as well as runs it with 0.0 drop_rate to make sure pre and post surgery outputs are the same.

composer/utils/fx_utils.py
```diff
@@ -111,6 +113,30 @@ def replace_op(gm: GraphModule, src_ops: Union[Callable, str, List[Union[Callabl
     return gm
 
 
+class block_stochastic_module(nn.Module):
```
ravi-mosaicml, line 116: nit: Use CamelCase for class names

```suggestion
class BlockStochasticModule(nn.Module):
```

And docstrings...

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+            their common ancestors removed.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    lhsAncestors = _get_ancestors(nodeLHS)
+    rhsAncesotors = _get_ancestors(nodeRHS)
+
+    # Iterate from back and eliminate common nodes
+    while lhsAncestors and rhsAncesotors and lhsAncestors[-1] == rhsAncesotors[-1]:  # type: ignore
+        del lhsAncestors[-1]
+        del rhsAncesotors[-1]
```
ravi-mosaicml, line 177:
```suggestion
        del rhsAncestors[-1]
```

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
-        GraphModule: Modified GraphModule.
+        (lhsAncestors, rhsAncesotors): Two lists of nodes containing ancestors for ``nodeLHS`` and ``nodeRHS`` with
+            their common ancestors removed.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    lhsAncestors = _get_ancestors(nodeLHS)
+    rhsAncesotors = _get_ancestors(nodeRHS)
+
+    # Iterate from back and eliminate common nodes
+    while lhsAncestors and rhsAncesotors and lhsAncestors[-1] == rhsAncesotors[-1]:  # type: ignore
```
ravi-mosaicml, line 175: curious about the need for a type ignore?

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+def apply_stochastic_residual(gm: GraphModule, drop_rate: float = 0.2) -> Tuple[GraphModule, int]:
+    """Detect and replace residual pattern with their stochastic equivalent.
+
+    Arguments:
+        gm (GraphModule): The source FX-traced graph. It can be the whole model symbolically traced.
+
+    Returns:
+        GraphModule: Modified GraphModule that has stochastic residual connections.
+    """
+    assert isinstance(gm, GraphModule), 'Input to apply_stochastic_residual should be an instance of GraphModule'
```
ravi-mosaicml, line 198: No need for this assert -- it is implied by the type annotation

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+    count = 0
+    # all nodes that are not a part of the residual blocks are tagged with "main_0".
+    # a tag is required for all nodes by split_by_tags
+    # Also an earlier tag can be repeated for later nodes.
+    all_tags = []
+    # In this first pass over all nodes, we just tag them
+    for node in gm.graph.nodes:
+        default_tag = f'mainN_{count}'
+        node.tag = default_tag
```
ravi-mosaicml, line 207: Curious if `.tag` is a part of the torchFX node spec? If not, is it possible to use a dictionary instead -- e.g. `Dict[Node, Tag]` instead of appending it onto the object?

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+
+
+def _tag_nodes(nodes: List[Node], tag: str):
+    """Tag nodes for the splitter."""
+    for node in nodes:
+        node.tag = tag  # type: ignore[attr-defined]
+
+
+def apply_stochastic_residual(gm: GraphModule, drop_rate: float = 0.2) -> Tuple[GraphModule, int]:
+    """Detect and replace residual pattern with their stochastic equivalent.
```
ravi-mosaicml, line 190: This function is a bit dense to read through. Is it possible to split it up -- e.g. `_tag_nodes`, `_replace_nodes_with_stochastic`?

composer/utils/fx_utils.py
```diff
@@ -123,16 +149,106 @@ def detect_residual_pattern(gm: GraphModule):
...
+        if default_tag not in all_tags:
+            all_tags.append(default_tag)
+        if node.op == 'call_function' and node.target in [torch.add, operator.add]:
+            assert len(node.all_input_nodes) == 2
+            node0, node1 = node.all_input_nodes[0], node.all_input_nodes[1]
+            lhs_nodes, rhs_nodes = _get_residual_path(node0, node1)
+            if lhs_nodes or rhs_nodes:
+                if len(lhs_nodes):
+                    _tag_nodes(lhs_nodes, f'non_res_{count}')
+                    all_tags.append(f'non_res_{count}')
```
ravi-mosaicml, line 217: Not a huge fan of tagging with a string and then regex parsing below to match against tags. Is it possible to a tuple or multiple tags instead? (Also see above comment re: using a dictionary)

ravi-mosaicml (commented):
Going to defer the correctness review to @Landanjs / @bandish-shah, but see comments re: code styling.

composer/utils/fx_utils.py
dskhudia: It's throwing the following error. Not sure what's the issue. 
 
  /workdisk/daya/composer/composer/utils/fx_utils.py:186:45 - error: Operator "==" not supported for types "Node | Unbound" and "Node | Unbound"
    Operator "==" not supported for types "Unbound" and "Unbound" when expected type is "List[Node]" (reportGeneralTypeIssues)

composer/utils/fx_utils.py
dskhudia: That just a static check and this API is used in non-type checked setting it will throw an unrelated error.

dskhudia:
```
>>> from composer.utils.fx_utils import apply_stochastic_residual
>>> x = 10
>>> apply_stochastic_residual(x)
  File "<stdin>", line 1, in <module>
  File "/workdisk/daya/composer/composer/utils/fx_utils.py", line 209, in apply_stochastic_residual
    assert isinstance(gm, GraphModule), 'Input to apply_stochastic_residual should be an instance of GraphModule'
AssertionError: Input to apply_stochastic_residual should be an instance of GraphModule
```

vs 

```
>>> from composer.utils.fx_utils import apply_stochastic_residual
>>> x = 10
>>> apply_stochastic_residual(x)
  File "<stdin>", line 1, in <module>
  File "/workdisk/daya/composer/composer/utils/fx_utils.py", line 216, in apply_stochastic_residual
    for node in gm.graph.nodes:
AttributeError: 'int' object has no attribute 'graph'
>>>
```

composer/utils/fx_utils.py
dskhudia: It is part of the torch.fx pass. See https://github.com/pytorch/pytorch/blob/master/torch/fx/passes/split_utils.py#L65

composer/utils/fx_utils.py
dskhudia: That's how the split utility in torch.fx is designed. https://github.com/pytorch/pytorch/blob/master/torch/fx/passes/split_utils.py#L65

composer/utils/fx_utils.py
dskhudia: Tagging part separated to _tag_nodes

composer/utils/fx_utils.py
```diff
@@ -111,6 +113,40 @@ def replace_op(gm: GraphModule, src_ops: Union[Callable, str, List[Union[Callabl
     return gm
 
 
+class BlockStochasticModule(nn.Module):
+    """A convenience class that stochastically executes the provided non-residual path.
+
+    Args:
+        lhs (GraphModule): Operators in the non-residual path of a residual block.
```
hanlint, line 120: nit: `lhs` and `rhs` are not descriptive -- consider using `main` vs `residual` as the arguments?

composer/utils/fx_utils.py
```diff
@@ -111,6 +113,40 @@ def replace_op(gm: GraphModule, src_ops: Union[Callable, str, List[Union[Callabl
     return gm
 
 
+class BlockStochasticModule(nn.Module):
```
hanlint, line 116: should this (and the corresponding code) be in the `stochastic_depth` algorithm instead? prefer to keep all algorithms as standalone as possible.

composer/utils/fx_utils.py
```diff
@@ -123,16 +159,111 @@ def detect_residual_pattern(gm: GraphModule):
...
+
+    Arguments:
+        gm (GraphModule): The source FX-traced graph. It can be the whole model symbolically traced.
+
+    Returns:
+        GraphModule: Modified GraphModule that has stochastic residual connections.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    assert isinstance(gm, GraphModule), 'Input to apply_stochastic_residual should be an instance of GraphModule'
+    all_tags, count = _tag_nodes(gm)
```
hanlint, line 241: `_tag_nodes` -> `_tag_residual_nodes`, as this util file may have additional tags required

composer/utils/fx_utils.py
```diff
@@ -123,16 +159,111 @@ def detect_residual_pattern(gm: GraphModule):
...
+    split_gm = split_by_tags(gm, all_tags)
+    pattern = re.compile(r'non_res_(\d+)|residual_(\d+)')
+    for node in split_gm.graph.nodes:
+        if node.op == 'call_module':
```
hanlint, line 245: nit: remove indentation with:
```
if node.op != call_module:
    continue

matches = pattern.match(node.target)
if not matches:
   continue 
```

composer/utils/fx_utils.py
```diff
@@ -123,16 +159,111 @@ def detect_residual_pattern(gm: GraphModule):
...
+            matches = pattern.match(node.target)
+            if matches:
```
hanlint, line 247: why could torch.fx's `replace_pattern` not work here? 

the readability of this code is a bit difficult, as the stochatic depth specific bits are mixed in with what seems like more general torch.fx graph manipulation methods. Is it possible to refactor this into helper functions for bette readability and also encapsulation (e.g. the regex pattern and getting the lsh_submod/rhs_submod should be encapsulated)? For example..
```
for node in split_gm.graph.nodes:
    if node.op != 'call_module':
        continue

    module = _get_residual_module(node)
    if module:
        main, residual = module
        new_module = BlockStochasticModule(main, residual, drop_rate)
        split_gm.add_module('...', new_module)

        _replace_call_module(split_gm, original=node, replace=new_module)
```   
or something like that?
     


```

**Review by hanlint — COMMENTED** [review]
Nice work, a few style nits, and a main few questions in the comments.

File: composer/utils/fx_utils.py (hunk 13 above)
Comment (dskhudia) [reply composer/utils/fx_utils.py:116]: Sure. Will move it there.

File: composer/utils/fx_utils.py (hunk 12 above)
Comment (dskhudia) [reply composer/utils/fx_utils.py:120]: How about residual and critical? critical path because it has more operations.

File: composer/utils/fx_utils.py (hunk 16 above)
Comment (dskhudia) [reply composer/utils/fx_utils.py:247]: `replace_pattern` doesn't work here for a number of reasons. The main challenging part in this was the dynamic conditional and is the reason why we need to split into subgraphs of graphModule so that we can insert conditional calls to them. 

- The pattern we want to replace has dynamic conditional. For `replace_pattern`, pattern + replacement both should be be FX-traceable. 
- replace_pattern replaces with new modules/functions so doesn't preserve weights. 

Will try to refactor but most of the refactored functions will not be reusable as they are very specific to stochastic depth. We will have our own `replace_pattern` implementation that takes care of optimizer state.

File: composer/utils/fx_utils.py (hunk 7 above)
Comment (ravi-mosaicml) [reply composer/utils/fx_utils.py]: Ah, it's probably coming from the lines below...instead of using `del` syntax, could you do `lhsAncestors.pop()`

File: composer/utils/fx_utils.py (hunk 11 above)
Comment (ravi-mosaicml) [reply composer/utils/fx_utils.py:219]: Wow, that seems like a whack design, but makes sense -- thanks!

File: composer/utils/fx_utils.py [hunk 17]
```diff
@@ -123,16 +159,111 @@ def detect_residual_pattern(gm: GraphModule):
...
+                    all_tags.append(f'residual_{count}')
+                add_tag = f'addN_{count}'
+                if add_tag not in all_tags:
+                    all_tags.append(add_tag)
+                node.tag = add_tag
+                count += 1
+    return all_tags, count
+
+
+def apply_stochastic_residual(gm: GraphModule, drop_rate: float = 0.2) -> Tuple[GraphModule, int]:
```
ravi-mosaicml, line 231: Would it be possible to split up this function into subroutines? It is quite dense to read through with all the nesting.

ravi-mosaicml (approved):
Overall looks good, thanks for doing this. See comments. I didn't check for correctness FYI; maybe wait for @hanlint or @Landanjs to approve before merging?

composer/utils/fx_utils.py
Landanjs: I expect the functionality here to replace the current stochastic depth implementation. Should that happen in this PR or a future PR?

dskhudia: This will happen in a followup PR.

composer/utils/fx_utils.py
dskhudia: Thanks. That fixed it. pyright is weird.

composer/utils/fx_utils.py
dskhudia: Renamed to residual and main

composer/utils/fx_utils.py
dskhudia: Moved to stochastic_layers.py

composer/utils/fx_utils.py
dskhudia: renamed.

composer/utils/fx_utils.py
dskhudia: done

composer/utils/fx_utils.py
dskhudia: refactored as suggested.

composer/utils/fx_utils.py
dskhudia: I separated part of it into another function earlier but now rest of the functionality is also separated into a couple of more functions as Hanlin suggested.

ravi-mosaicml (commented):
(Dismissing my review; don't have much to add)

composer/utils/fx_utils.py
```diff
@@ -123,16 +125,136 @@ def detect_residual_pattern(gm: GraphModule):
...
+    """Detect and replace residual pattern with their stochastic equivalent.
+
+    Arguments:
+        gm (GraphModule): The source FX-traced graph. It can be the whole model symbolically traced.
+
+    Returns:
+        GraphModule: Modified GraphModule that has stochastic residual connections.
     """
-    raise NotImplementedError('replace_residual_with_stochastic is currently not implemented.')
+    assert isinstance(gm, GraphModule), 'Input to apply_stochastic_residual should be an instance of GraphModule'
```
hanlint, line 244: since `apply_*` is part of the public API, this should raise a `ValueError` instead of an assert.

hanlint (approved):
LGTM, one small nit. Defer to @Landanjs to review for correctness. 

Right now `apply_stochastic_residual` is not actually used in the `stochastic depth` algorithm. I assume this will be deferred to follow-on PR? Should we do a training run with this to see if matches the performance of the old apply?

composer/utils/fx_utils.py
dskhudia: Converted it to valueerror

Landanjs (approved):
The stochastic depth implementation LGTM. Like Hanlin mentioned the best test would be before and after experiments, but those will happen in a follow up PR 🙂

PR #1254, opened 2022-07-03
Fix repeated words in event.py

isaac0804:


hanlint (approved):
Thank you @isaac0804 for fixing this!

PR #1255, opened 2022-07-05
Make progressive resizing quieter

coryMosaicML:
This PR removes the logging from the functional implementation of progressive resizing as it seems unnecessarily verbose


coryMosaicML:
> Maybe make this debug level instead?

Done

hanlint (approved):

PR #1256, opened 2022-07-05
Fix grad accum parsing in hparams

hanlint:
Currently, using `--grad_accum 1` in the arg parser fails. `grad_accum` can be an `int`, or ` str`, and so yahp does not attempt to coerce a string such as `"1"` into an int. This adds that logic into our `hparams` object.

This PR does not affect the non-YAHP Trainer API.


ravi-mosaicml (approved):
LGTM, though this is also a bug in yahp -- for unions with str, it should attempt to parse into other types first before assuming it is a string.

hanlint:
Agreed -- @ravi-mosaicml should we merge this anyway, or fix this directly in yahp?

hanlint:
Corresponding fix to YAHP here: https://github.com/mosaicml/yahp/pull/122

PR #1257, opened 2022-07-05
Fix recipe yamls for `v0.8`, add testing

hanlint:
This PR:
* fixes the recipe yamls to work with the latest version of COmposer
* expands our test coverage to include those yamls to prevent future regressions

Note: this is a UX change, as the recipe yamls will now have default ssr ratios, as well as datasets path. @growlix what do you prefer? It's easy to patch the yaml during the test if you would prefer to keep those unfilled.

### DIFF
--- a/composer/trainer/devices/device_gpu.py
+++ b/composer/trainer/devices/device_gpu.py
@@ -29,6 +29,8 @@ class DeviceGPU(Device):
     dist_backend = 'nccl'
 
     def __init__(self, device_id: Optional[int] = None):
+        if not torch.cuda.is_available():
+            raise ValueError('DeviceGPU cannot be created as torch.cuda is not available.')
         if not device_id:
             device_id = dist.get_local_rank()
         self._device = torch.device(f'cuda:{device_id}')
--- a/tests/models/test_model_yamls.py
+++ b/tests/models/test_model_yamls.py
@@ -2,13 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
-import sys
+from typing import Any, Dict
 
 import pytest
+import yaml
 
 import composer
 from composer.core.precision import Precision
-from composer.trainer.devices.device_cpu import DeviceCPU
 from composer.trainer.trainer_hparams import TrainerHparams
 from tests.common import configure_dataset_hparams_for_synthetic, configure_model_hparams_for_synthetic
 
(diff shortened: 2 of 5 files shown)


growlix (approved):
LGTM. I don't have strong feelings one way or another about leaving required but user-specific hparams (e.g. `datadir`) empty w/ a comment vs. providing a default that will need to be overridden.

composer/trainer/devices/device_gpu.py
```diff
@@ -29,6 +29,8 @@ class DeviceGPU(Device):
     dist_backend = 'nccl'
 
     def __init__(self, device_id: Optional[int] = None):
+        if not torch.cuda.is_available():
```
ravi-mosaicml, line 32: nit: Do we need this check? I believe it will error below on line 36 if cuda is not available.

tests/models/test_model_yamls.py
```diff
@@ -28,6 +29,11 @@ def walk_model_yamls():
 @pytest.mark.parametrize('hparams_file', walk_model_yamls())
 class TestHparamsCreate:
 
+    @pytest.fixture(scope='function', autouse=True)
+    def ensure_device_cpu(self, monkeypatch: pytest.MonkeyPatch):
```
ravi-mosaicml, line 33: Would prefer not to read `sys.argv` as they can contain pytest-specific args that would conflict with yahp. Instead, would it work to remove the `device` arg from the yaml files? The trainer will automatically use gpu if available; otherwise it will use CPU.

If we need to manually patch the device/precision, could we load the yaml into a dictionary, and then manually override `yaml_dict['device'] = {'cpu': {}}`, and then pass that into `TrainerHparams.create`? The create API supports reading from a dict as well as from a file.

ravi-mosaicml (commented):
Hparams changes look fine, though see comments re: test case changes.

composer/trainer/devices/device_gpu.py
hanlint: Yes, but the error is a bit obscure, so a check here would be more informative.

tests/models/test_model_yamls.py
hanlint: Done!

tests/models/test_model_yamls.py
```diff
@@ -64,7 +76,13 @@ def test_trainer_initialize(self, hparams_file: str):
         if 'deeplabv3' in hparams_file:
             pytest.importorskip('mmseg')
 
-        hparams = TrainerHparams.create(hparams_file, cli_args=False)
+        with open(hparams_file, 'r') as f:
+            yaml_dict = yaml.safe_load(f)
+            print(yaml_dict)
```
ravi-mosaicml, line 81: nit: looks like a debug statement
```suggestion
```

PR #1258, opened 2022-07-05
HTTPS downloads for streaming datasets

ravi-mosaicml:
This PR allows for streaming URIs to be specified as HTTP/HTTPs URIs. This allows for streaming public datasets from anywhere. It also provides a workaround to https://github.com/boto/boto3/issues/1200, where the dataset URI can be specified in https format rather than in s3 format to avoid having to configure unsigned credentials.

Closes https://mosaicml.atlassian.net/browse/CO-692


knighton (approved):

ravi-mosaicml:
Still need to add test cases; will merge after those are in.

abhi-mosaic (approved):
LGTM pending tests!

PR #1260, opened 2022-07-05
Create a new `boto3.Session()` per `S3ObjectStore` instance 

ravi-mosaicml:
`boto3` sessions are not thread safe. When used in the object store logger with `use_procs: False`, the default session was shared across threads, which caused us to run into https://github.com/boto/boto3/issues/1592. To fix, this PR creates a new session within each `S3ObjectStore` instance.

Closes https://mosaicml.atlassian.net/browse/CO-651


ajaysaini725 (approved):
Just verified this works with `use_procs: false` on our cluster

Thanks for getting this up so fast!

ajaysaini725 (approved):
Somehow approval from earlier didn't go through lol

PR #1262, opened 2022-07-06
Update to latest pytorch

mvpatel2000:
Upgrades syntax to latest pytorch version so we don't get warnings of deprecated syntax. Currently, tests fail on a fresh install because of these warnings

### DIFF
--- a/composer/models/resnet/model.py
+++ b/composer/models/resnet/model.py
@@ -4,10 +4,14 @@
 """A :class:`.ComposerClassifier` wrapper around the torchvision implementations of the ResNet model family."""
 
 import logging
+from packaging import version
 from typing import List, Optional
+import warnings
+
 
 from torchmetrics import MetricCollection
 from torchmetrics.classification import Accuracy
+import torchvision
 from torchvision.models import resnet
 
 from composer.loss import loss_registry
@@ -25,6 +29,7 @@ valid_model_names = ['resnet18', 'resnet34', 'resnet50', 'resnet101', 'resnet152
 def composer_resnet(model_name: str,
                     num_classes: int = 1000,
                     weights: Optional[str] = None,
+                    pretrained: bool = False,
                     groups: int = 1,
                     width_per_group: int = 64,
                     initializers: Optional[List[Initializer]] = None,
@@ -38,6 +43,8 @@ def composer_resnet(model_name: str,
             ``"resnet152"``].
         num_classes (int, optional): The number of classes. Needed for classification tasks. Default: ``1000``.
         weights (str, optional): If provided, pretrained weights can be specified, such as with ``IMAGENET1K_V2``. Default: ``None``.
+        pretrained (bool, optional): If True, use ImageNet pretrained weights. Default: ``False``. This parameter is deprecated and
+            will soon be removed.
         groups (int, optional): Number of filter groups for the 3x3 convolution layer in bottleneck blocks. Default: ``1``.
         width_per_group (int, optional): Initial width for each convolution group. Width doubles after each stage.
             Default: ``64``.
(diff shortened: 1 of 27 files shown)


composer/models/resnet/model.py
```diff
@@ -37,7 +37,7 @@ def composer_resnet(model_name: str,
         model_name (str): Name of the ResNet model instance. Either [``"resnet18"``, ``"resnet34"``, ``"resnet50"``, ``"resnet101"``,
             ``"resnet152"``].
         num_classes (int, optional): The number of classes. Needed for classification tasks. Default: ``1000``.
-        pretrained (bool, optional): If True, use ImageNet pretrained weights. Default: ``False``.
+        weights (str, optional): If provided, pretrained weights can be specified, such as with ``IMAGENET1K_V2``. Default: ``None``.
```
hanlint, line 40: any way we can avoid changing the API here?

mvpatel2000: Hm... we could leave it as if `pretrained=True`, we take `IMAGENET1K_V2` weights (the best ones). I'm pro changing though -- imo we should mirror the API to torchvision

mvpatel2000: So the issue is `weights` is only added in 0.13. I'm not sure if we've figured out an upgrade policy, but I see a few ways to proceed
(1) update `setup.py` to use the most recent `torchvision` release
(2) update `setup.py` to require `torchvision` that's at max 0.12
(3) update our class to identify the `torchvision` version being used and appropriately pass in the right args
I'm personally in favor of 1 > 3 > 2

ravi-mosaicml: Agreed; let's not force versions.

composer/models/resnet/model.py
```diff
@@ -76,9 +82,22 @@ def composer_resnet(model_name: str,
...
+            'The ``pretrained`` argument for composer_resnet is deprecated and will be removed in the future. Please use ``weights`` instead.'
+        )
+    if weights:
+        pretrained = True
+
     # Instantiate model
     model_fn = getattr(resnet, model_name)
-    model = model_fn(pretrained=pretrained, num_classes=num_classes, groups=groups, width_per_group=width_per_group)
+    model = None
+    if version.parse(torchvision.__version__) < version.parse('0.13.0'):
```
ravi-mosaicml, line 97: Curious if we can avoid parsing the version? Would prefer to try the newer syntax only if it is used. Thinking something along the lines of:
````python

if weights is None:
    if pretrained:
       warnings.warn(DeprecationWarning(...))
    model = model_fn(pretrained=pretrained, num_classes=num_classes, groups=groups, width_per_group=width_per_group)
else:
    if pretrained:
        raise ValueError("Cannot specify pretrained and weights")
    model = model_fn(weights=weights, num_classes=num_classes, groups=groups, width_per_group=width_per_group)
```

composer/models/resnet/model.py
```diff
@@ -76,9 +82,22 @@ def composer_resnet(model_name: str,
     if initializers is None:
         initializers = []
 
+    # Configure pretrained/weights based on torchvision version
+    if pretrained:
+        weights = 'IMAGENET1K_V2'
+        warnings.warn(
```
ravi-mosaicml, line 88: nit: Can this be a deprecation warning?

composer/models/resnet/model.py
```diff
@@ -76,9 +82,22 @@ def composer_resnet(model_name: str,
...
+            'The ``pretrained`` argument for composer_resnet is deprecated and will be removed in the future. Please use ``weights`` instead.'
+        )
+    if weights:
+        pretrained = True
```
ravi-mosaicml, line 92: Can we raise an exception if both are specified (i.e. pretrained is True and weights is not None)?

ravi-mosaicml (approved):
LGTM, see comments

ravi-mosaicml (approved):

composer/models/resnet/model.py
mvpatel2000: Using the deprecated path issues a warning, not a TypeError, so unfortunately we can't do that :(

composer/models/resnet/model.py
```diff
@@ -76,9 +82,27 @@ def composer_resnet(model_name: str,
...
+        raise ValueError(
+            'composer_resnet expects only one of ``pretrained`` or ``weights`` to be specified, but both were specified.'
+        )
+    if pretrained:
+        weights = 'IMAGENET1K_V2'
+        warnings.warn(
+            DeprecationWarning(
+                'The ``pretrained`` argument for composer_resnet is deprecated and will be removed in the future. Please use ``weights`` instead.'
+            ))
+    if weights:
```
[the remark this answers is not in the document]
mvpatel2000: Moved into check. I'm in favor of emitting a warning which indicates its falling back to `pretrained=True` and that user should upgrade `torchvision` version to use `weights`. 

> Also, if weights and weights is not True (i.e. it is a string / path), then would like to raise an exception, as we wouldn't be able to load a customized weights path.

I don't understand what exactly you mean here. It can be a string argument -- see https://pytorch.org/vision/stable/models.html#initializing-pre-trained-models

PR #1263, opened 2022-07-06
Glue Entrypoint

ishanashastri:
Allows pretraining and/or finetuning of NLP models to be conducted via a single entrypoint using multiprocessing. 

This is a fairly large, non-breaking PR, so I will attempt to break down what the structure and usage of this entrypoint is. The call structure is as follows: 

`python examples/run_nlp_trainer.py -f composer/yamls/models/bert-base.yaml --training_scheme {pretrain, finetune, all} --load_path {path/to/ckpts/folder}`

Note that `--load_path` should be set only in the case that `--training_scheme` is set to finetune. All checkpointing is handled internally by the entrypoint and in the case where `--training_scheme` is specified to be `all`, the entrypoint will automatically save and load all checkpoints either locally or via S3 and output a table of GLUE metrics at the very end of training. 

To use S3 or WandB, you must specify it within the loggers section of the bert-base.yaml (or whatever file you input as -f to the entrypoint).

Internally, for pretraining, the entrypoint will subprocess.Popen() a composer launcher instance to handle all the distributed pretraining of the given model. For finetuning, python multiprocessing will be used to spawn 1 GPU instances for each of the GLUE tasks. 

Todo: docs/notebook


examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        output_line.append("{:.2f}".format(glue_large/len(large_tasks)))
+        output_line.append("{:.2f}".format(glue_all/len(tasks)))
+        tb.append(output_line)
+    
+    print(tabulate(tb, headers='firstrow'))
+
+''' Set up CUDA environment and process pool for given finetuning jobs. '''
+def setup_finetuning_jobs(finetune_tasks: List[str], save_ckpts: List[bool], ckpt_load_path: str, ckpt_filename: str,
+                                         save_folder: str, save_locally: bool, bucket: Optional[str] = None, parent_ckpt: Optional[str] = None, load_ignore_keys: Optional[List[str]] = None) -> None:
+    cuda_envs = init_cuda_queue(len(finetune_tasks))
```
ravi-mosaicml, line 123: What if we have more finetune tasks than gpus available?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+    num_tasks = len(finetune_tasks)
+    
+    ''' Callback function for metric collection. '''
+    def log_metrics(metric) -> None:
```
ravi-mosaicml, line 127: Generally would try to avoid defining functions in functions. Can this be a parent function?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        if parent_ckpt: 
+            logged_ckpt_name = parent_ckpt
+        else:
+            logged_ckpt_name = ckpt_filename
+
+        if logged_ckpt_name not in GLUE_METRICS.keys():
+                GLUE_METRICS[logged_ckpt_name] = {}
+        task = list(metric.keys())[0]
+        for m in metric[task]:
+            formatted_task = list(metric.keys())[0].split('_')[1] # remove "glue_" prefix
```
ravi-mosaicml, line 137: I'm wondering whether the `GLUE_METRICS` dictionary is the right data structure -- there seems to be a lot of manual parsing and code that depends on the ordering of the keys. Curious if a `GlueState` object would help make it more readable? Eg.g

```python

@dataclass
class GlueState:
    task_names: List[str]
    task_to_metric: Dict[str, Dict[str, Any]]  # dict of task name to task metric names and values
```

I'm not sure what the right hierarchy is but something like that would help with readability.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+            if formatted_task not in GLUE_METRICS[logged_ckpt_name].keys():
+                GLUE_METRICS[logged_ckpt_name][formatted_task] = metric[task][m].item()
+            else:
+                GLUE_METRICS[logged_ckpt_name][formatted_task] = sum([GLUE_METRICS[logged_ckpt_name][formatted_task], metric[task][m].item()]) / 2  # automatically average scores 
+        
+    if parent_ckpt: 
+        wandb_group_name = parent_ckpt
```
ravi-mosaicml, line 144: Can we generalize this script so it doesn't rely on wandb?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        raise ValueError('load_path to checkpoint folder must be specified if finetuning a model')
+
+    if not args.save_locally and args.bucket is None:
+        raise ValueError('S3 bucket name must be specified if saving to cloud')
+
+''' Get TrainerHparams arguments from CLI and add NLP entrypoint specific arguments. ''' 
+def get_args():
+    parser = hp.get_argparse(TrainerHparams)
+    parser.add_argument('--training_scheme', help='training scheme used (one of "pretrain", "finetune", or "all")')
+    parser.add_argument('--save_locally', action=argparse.BooleanOptionalAction, help='save to local directory or to cloud using object store if --no-save-locally is enabled')
```
ravi-mosaicml, line 227: Curious why do we need save locally and bucket? Could one use the checkpointing args as part of TrainerHparams?

ravi-mosaicml (commented):
I wasn't able to go through it completely, but left some thoughts for now. I'll try to continue later today. Let's also sync up offline.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+                GLUE_METRICS[logged_ckpt_name][formatted_task] = sum([GLUE_METRICS[logged_ckpt_name][formatted_task], metric[task][m].item()]) / 2  # automatically average scores 
+        
+    if parent_ckpt: 
+        wandb_group_name = parent_ckpt
+    else:
+        wandb_group_name = ckpt_load_path
+
+    # finetuning from pretrained checkpoint
+    print(f"FINETUNING ON {ckpt_filename}!")
+    pool = mp.Pool(num_tasks, initializer=init_cuda_env, initargs=(cuda_envs,))
```
ravi-mosaicml, line 150: Fancy use of multiprocessing :)

```suggestion
    pool = mp.Pool(torch.cuda.device_count(), initializer=init_cuda_env, initargs=(cuda_envs,))
```

I presume that each mp worker will use the same device id, even if it is reused?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+    root_dir = os.path.join(os.path.dirname(composer.__file__), '..')
+    training_script = os.path.join(root_dir, 'examples/run_composer_trainer.py')
+    if args.save_locally:
+        args.save_folder = os.path.join(args.run_name, 'checkpoints')
+        if args.training_scheme == 'all':
+            new_save_folder = os.path.join(args.save_folder, 'pretrained')
+        else: # pretrain only
+            new_save_folder = args.save_folder
+        
+        proc = subprocess.Popen(['composer', training_script, '-f', f"{args.file}", '--save_folder', f"{new_save_folder}",
```
ravi-mosaicml, line 245: Can you create a jira to add a functional api to the composer launch script? This works but not a big fan of calling python via a subprocess. But there's no way around this for no so it's fine.

examples/run_nlp_trainer.py
ishanashastri: The entrypoint was built based off the assumption that we're running an 8GPU job for both pretraining and finetuning. How high of a priority is it to support other use cases? Doing so would likely require a more formal job/resource queue.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+            output_line.append("{:.2f}".format(GLUE_METRICS[ckpt][task]))
+            glue_all += GLUE_METRICS[ckpt][task]
+            if task in large_tasks:
+                glue_large += GLUE_METRICS[ckpt][task]
+        # GLUE Large and GLUE All
+        output_line.append("{:.2f}".format(glue_large/len(large_tasks)))
+        output_line.append("{:.2f}".format(glue_all/len(tasks)))
+        tb.append(output_line)
+    
+    print(tabulate(tb, headers='firstrow'))
```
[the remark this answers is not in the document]
ishanashastri: Done!

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+            
+    else:
+        if args.training_scheme == 'all':
+            ckpt_folder = f'{args.run_name}/{args.save_folder}'
+            save_folder = args.save_folder
+        else: # finetune only
+            ckpt_folder = args.load_path # load path should be full/path/to/folder/ containing all ckpts 
+            save_folder = args.run_name 
+
+        all_ckpts_list = get_all_s3_checkpoints(args.bucket, ckpt_folder)
```
ravi-mosaicml, line 285: Ah, I think I answer one of my questions from before about why we need to list the bucket. I see why that is needed now.

Instead of requiring a list bucket / list dir, could this function take the list of checkpoints to finetune?

ravi-mosaicml (commented):
Thanks for putting this together. Overall, this seems like a great MVP for being able to pretrain and fine tine from one command.

As for next steps, I would prefer for it to be less opinionated (e.g. s3, wandb, yaml paths) and also refactor the multiprocessing logic for readability. Lets sync offline.

examples/run_nlp_trainer.py
ishanashastri: These were created so that users don't have to manually modify all the GLUE yamls to set up and include `load_object_store` or `loggers` if they want to load/save checkpoints via cloud. If we're ok with throwing something into the yamls like how we did in the [bert yaml](https://github.com/mosaicml/composer/blob/2655cc0fc47755e8eb0e6dfeff3c300b99bd7a84/composer/yamls/models/bert-base.yaml#L80) then i can get rid of these arguments and just pull from there.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+    pool.close()
+    pool.join() # wait for join to proceed
+   
+    cuda_envs.close() 
+    cuda_envs.join_thread()
+    
+'''Run single instance of a finetuning job on given task. ''' 
+def train_finetune(task: str, save_ckpt: bool, load_path: str, wandb_group_name: str, save_folder: str, save_locally: bool, gpu_id: int, bucket: Optional[str] = None, load_ignore_keys: Optional[List[str]] = None):
+    # add dummy arg because arg parser removes first argument in tokenization
+    ft_hparams = TrainerHparams.create(cli_args=['-f',  f'./composer/yamls/models/glue/{task}.yaml',
```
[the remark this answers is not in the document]
ishanashastri: yes, this works too!

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        if args.training_scheme == 'all':
+            args.save_folder = 'pretrained'
+            save_artifact_name = '{run_name}/pretrained/ep{epoch}-ba{batch}-rank{rank}-wct{total_wct}'
+            proc = subprocess.Popen(['composer', training_script, '-f', f"{args.file}" , '--save_folder', args.save_folder, 
+                    '--save_artifact_name', save_artifact_name, '--save_num_checkpoints_to_keep', '0', '--loggers',
+                    'wandb', 'object_store', '--tags', 'pretrain', '--object_store_hparams', 's3', '--bucket', args.bucket], cwd=root_dir)                
+        
+        # only pretraining, save to default artifact name 
+        else: 
+            proc = subprocess.Popen(['composer', training_script, '-f', f"{args.file}", '--save_num_checkpoints_to_keep', '0',
```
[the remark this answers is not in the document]
ishanashastri: refactoring done.

subprocess.Popen was used to fwd the output to stdout in real time, which is likely useful for debugging a run and making sure it's going as expected. From my understanding, .run will wait until the end of the process to return the stdout result.

composer/trainer/trainer_hparams.py
```diff
@@ -238,6 +238,7 @@ class TrainerHparams(hp.Hparams):
 
         device (Device, optional): Hparams for constructing the device used for training.
             (default: ``None``)
+        device_id (int, optional): ID of GPU to train with. (default: ``None``)
```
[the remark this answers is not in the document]
ishanashastri: Correct, this was part of an old change before that PR. Resolved in cleanup.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        run_finetuner(args)
+        print('FINETUNING COMPLETE')
+
+    # output GLUE metrics if finetuning
+    if args.training_scheme != 'pretrain':
+        output_metrics()
+
+if __name__ == '__main__':
+    mp.set_start_method('spawn')
+    os.environ['WANDB_START_METHOD'] = 'thread' # to not interfere with multiprocessing
```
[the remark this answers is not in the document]
ishanashastri: Same issue as with the object store logger -- got an issue of 'daemonic processes are not allowed to have instances' that was interfering with the multiprocessing going on by the entrypoint. Based off a suggestion from the [wandb page](https://docs.wandb.ai/guides/track/advanced/distributed-training#hanging-at-the-beginning-of-training).

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+                glue_large += GLUE_METRICS[ckpt][task]
+        # GLUE Large and GLUE All
+        output_line.append("{:.2f}".format(glue_large/len(large_tasks)))
+        output_line.append("{:.2f}".format(glue_all/len(tasks)))
+        tb.append(output_line)
+    
+    print(tabulate(tb, headers='firstrow'))
+
+''' Set up CUDA environment and process pool for given finetuning jobs. '''
+def setup_finetuning_jobs(finetune_tasks: List[str], save_ckpts: List[bool], ckpt_load_path: str, ckpt_filename: str,
```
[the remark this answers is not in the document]
ishanashastri: Sorry I guess the logic is a bit convoluted here.` setup_finetuning_jobs()` launches the multiprocessing finetune jobs, which all have `log_metrics()` specified as a callback function. The `trainer.state.current_metrics` are returned from each of the individual finetune jobs and are then parsed by `log_metrics()` on callback to store it into `GLUE_METRICS`. I'm looking at your other comments and can try to rethink how `GLUE_METRICS` is defined so as to not be a global var.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: Sorry, what is "both" referring it?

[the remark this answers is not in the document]
ravi-mosaicml: the train_pretrain and train_finetune

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+    ckpt_names = []
+    for obj in bucket.objects.all():
+        if folder_name in obj.key:
+            obj_name = obj.key
+            ckpt_names.append(obj_name)
+
+    return ckpt_names
+
+''' Consolidate and prettify metrics. '''
+def output_metrics() -> None:
```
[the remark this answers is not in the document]
ishanashastri: yep can do that!

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+from composer.loggers.wandb_logger import WandBLogger
+from composer.trainer.trainer_hparams import TrainerHparams
+from composer.utils.object_store.object_store_hparams import S3ObjectStoreHparams
+from composer.utils.object_store.s3_object_store import S3ObjectStore
+
+# TODO: reset all configs and training lengths back to default
+# track GLUE Metrics
+GLUE_METRICS = {}
+
+def _warning_on_one_line(message: str, category: Type[Warning], filename: str, lineno: int, file=None, line=None):
```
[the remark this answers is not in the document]
ishanashastri: done

examples/run_nlp_trainer.py
ishanashastri: The reason it's defined within this function is because it's a callback that can only take one argument but it relies on other arguments already defined in this function. Will sync offline about this as there may be a workaround that I'm unaware of.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+
+''' List all the checkpoints in a given S3 bucket and folder. ''' 
+def get_all_s3_checkpoints(bucket_name: str, folder_name: str) -> List[str]:
+    # pulls AWS credentials from environment variables, assumes that aws_session_token is defined
+    session = boto3.Session()
+
+    s3 = session.resource('s3')
+    bucket = s3.Bucket(bucket_name)
+    ckpt_names = []
+    for obj in bucket.objects.all():
```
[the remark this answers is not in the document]
ishanashastri: Discussed offline -- not a huge priority to fix just yet. Will potentially remedy this using a text file list of checkpoints to load.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: Discussed offline -- difficult to do due since pretrain uses subprocess.run and finetune uses multiprocessing spawns. Code length and complexity will be cleaned up along with the other refactoring comments.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: can move most of them out, some of the args are conditional depending on the specified training scheme so those have to be provided as cli overrides

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+                '--device_id' , f"{gpu_id}",
+                "--log_to_console", "False",
+                "--progress_bar", "False",
+                "--load_ignore_keys", f"{load_ignore_keys}"])
+    
+    
+    ft_hparams.loggers = [WandBLogger(group=wandb_group_name, tags=['finetune', task], rank_zero_only=True)]
+    # load checkpoint to finetune on via cloud 
+    if not save_locally:
+        ft_hparams.load_object_store = S3ObjectStoreHparams(bucket=bucket)
```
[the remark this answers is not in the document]
ishanashastri: Users now have to specify loggers containing their object store and WandB args within the pretraining yaml in order to save those settings for the rest of the GLUE yamls.

examples/run_nlp_trainer.py
ishanashastri: Removed with enforcing shared settings to go in the pretrained yaml

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,365 @@
...
+        setup_finetuning_jobs(mnli_finetune_tasks, save_ckpts, ckpt_load_path, ckpt_filename, args.save_folder, glue_metrics, 
+            wandb_logger=args.wandb_logger, save_locally=save_locally, bucket=args.bucket, parent_ckpt=parent_ckpt, load_ignore_keys=["state/model/model.classifier*"], )  # type: ignore
+
+def _main() -> None:
+    warnings.formatwarning = _warning_on_one_line
+
+    if len(sys.argv) == 1:
+        sys.argv = [sys.argv[0], '--help']
+    
+    args = get_args()
```
moinnadeem, line 342: Could we add some asserts or exception raising here on default values? For example, raising an exception if the user has specified `training_scheme = finetune` but no path to pretrained checkpoints, etc.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,365 @@
...
+    -f composer/yamls/models/bert-base.yaml 
+    --training_scheme all 
+    --save_folder ~/checkpoints
+
+Example that finetunes a pretrained BERT:
+
+    >>> composer examples/run_nlp_trainer.py
+    -f composer/yamls/models/bert_base.yaml
+    --training_scheme finetune
+    --load_path ~/checkpoints
```
moinnadeem, line 24: Do we ever check if the `load_path` exists? I was able to run this command but it completed instantly since `~/checkpoints` does not exist.

moinnadeem (commented):
So far, LGTM! I'm now trying usability testing, and I run:
```
python examples/run_nlp_trainer.py -f composer/yamls/models/bert-base.yaml --training_scheme finetune --bucket s3://mosaicml-internal-checkpoints-bert/bert-base-128-jzf-seed-0-o0hc/
```

and I get 
```
  File "/root/composer/examples/run_nlp_trainer.py", line 364, in <module>
    _main()
  File "/root/composer/examples/run_nlp_trainer.py", line 354, in _main
    run_finetuner(args, glue_metrics)
  File "/root/composer/examples/run_nlp_trainer.py", line 294, in run_finetuner
    all_ckpts_list = get_all_s3_checkpoints(args.bucket, ckpt_folder)
  File "/root/composer/examples/run_nlp_trainer.py", line 79, in get_all_s3_checkpoints
    if folder_name in obj.key:
TypeError: 'in <string>' requires string as left operand, not _MISSING_TYPE
```

What am I missing? Additionally, could we provide validation for minimum sets of arguments required to run it? I am a bit unclear on what command line flags I need to fill out, and what they do.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+import subprocess
+import sys
+import warnings
+from typing import List, Optional
+
+import boto3
+import torch
+from composer.trainer.devices.device_gpu import DeviceGPU
+import yahp as hp
+from tabulate import tabulate
```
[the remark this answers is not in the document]
ishanashastri: I added it to the main composer install so this should be handled now!

examples/run_nlp_trainer.py
ishanashastri: I didn't check for this case in the entrypoint but composer should already have some sort of validation on this arg? cc @ravi-mosaicml

composer/trainer/trainer_hparams.py
```diff
@@ -414,7 +414,7 @@ def initialize_object(self) -> Trainer:
         # Device
         device = self.device
         if device is None:
-            device = DeviceGPU() if torch.cuda.is_available() else DeviceCPU()
```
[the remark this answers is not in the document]
ishanashastri: DeviceGPU by default used the local rank to initialize the device. #1244 allows us to specify a device to use now!

composer/yamls/models/bert-base.yaml
```diff
@@ -70,13 +70,17 @@ precision: amp                     # Use mixed-precision training
...
-  # Optional, some nice default parameters for object store checkpointing. Uncomment this if you want to checkpoint to cloud.
-  # object_store:
-  #  object_store_hparams:
-  #    s3:
-  #      bucket: mosaicml-internal-checkpoints-bert
+  # Optional, some nice default parameters for object store checkpointing and WandB logging. Uncomment these if you want to checkpoint to cloud and track training info.
+  object_store:
+    object_store_hparams:
+      s3:
+        bucket: mosaicml-internal-checkpoints-bert
```
alextrott16, line 83: Since this YAML is public facing, seems like maybe we would want to use a different bucket. Also this feels like the kind of thing that you wouldn't want to be part of a default BERT config. 

Is it possible to simply have the entrypoint activate this kind of storage, rather than make it active by default?

examples/run_nlp_trainer.py
alextrott16: It seems like a good goal to make a tool like this _not_ be overly resource-specific. But in the meantime it seems like a good idea to have the code verify that the available resources satisfy the requirements and raise the appropriate error(s) if they don't.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,366 @@
...
+            save_artifact_name = f'{{run_name}}/{task}/ep{{epoch}}-ba{{batch}}-rank{{rank}}'
+            save_latest_artifact_name = f'{{run_name}}/{task}/latest-rank{{rank}}'
+            ft_hparams.save_artifact_name = save_artifact_name
+            ft_hparams.save_latest_artifact_name = save_latest_artifact_name
+            ft_hparams.save_overwrite = True
+
+    print(f" --------\n SPAWNING NEW TASK: {task}\n DEVICE: {torch.cuda.current_device()}\n --------")
+
+    trainer = ft_hparams.initialize_object()
+    trainer.fit(duration='3ep')
```
alextrott16, line 207: Not sure we should hardcode the duration. I recommend to have the task YAMLs instead just use `max_duration: 3ep` so that it's at least more transparent where that duration comes from.

Also, are we sure all the fine-tuning tasks should run for 3 epochs? I seem to recall seeing task-specific durations.

composer/yamls/models/glue/cola.yaml
```diff
@@ -41,8 +41,8 @@ eval_batch_size: 32
 seed: 19
 dataloader:
   pin_memory: true
-  persistent_workers: true
-  num_workers: 8
+  persistent_workers: false
```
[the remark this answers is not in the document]
ishanashastri: Discussed offline. Change reverse by replacing mp.Pool with its non-daemonic parallel in concurrent.futures

examples/run_nlp_trainer.py
ishanashastri: Discussed offline. Changed by using the number of current devices to initialize the Pool.

composer/yamls/models/glue/cola.yaml
```diff
@@ -41,8 +41,8 @@ eval_batch_size: 32
 seed: 19
 dataloader:
   pin_memory: true
-  persistent_workers: true
-  num_workers: 8
```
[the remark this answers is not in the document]
ishanashastri: Python multiprocessing is a daemonic process that interfered with all distributed dataloader work so `num_workers` had to be changed to 0. This change was fixed by replacing mp.Pool with its non-daemonic parallel in concurrent.futures!

examples/run_nlp_trainer.py
ishanashastri: Sorry, this was just put in for testing purposes to shorten the amount of training required but will be fixed during cleanup. Thanks for pointing this out!!

composer/yamls/models/bert-base.yaml
ishanashastri: Sure, we can remove it. In the documentation, I can make a copy of this yaml to demonstrate how to use the entrypoint with S3/WandB.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+        sp.wait()
+    pool.close()
+    pool.join() # wait for join to proceed
+   
+    cuda_envs.close() 
+    cuda_envs.join_thread()
+    
+'''Run single instance of a finetuning job on given task. ''' 
+def train_finetune(task: str, save_ckpt: bool, load_path: str, wandb_group_name: str, save_folder: str, save_locally: bool, gpu_id: int, bucket: Optional[str] = None, load_ignore_keys: Optional[List[str]] = None):
+    ft_hparams = TrainerHparams.create(f=f'./composer/yamls/models/glue/{task}.yaml')
```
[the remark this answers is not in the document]
ishanashastri: Good catch!

composer/yamls/models/bert-base.yaml
alextrott16: OK great. In general, I think we want to avoid directly referencing this `composer/yamls/model/bert-base.yaml`, since it may get confusing when people are using this entrypoint with their own pre-training YAML.

alextrott16:
I'm excited about this code and so far it looks great!

I will admit that it is a bit tough to follow in places owing to the save/loading logic and how those behaviors are controlled by the user arguments and the supplied YAML. That probably just indicates the need for more thorough documentation and/or examples.

One issue I foresee is making the fine-tuning YAMLs reflect the pre-training. For example, if your pre-training model uses ALiBi as an algorithm, that should also be used in fine-tuning.

Might be good to sync up offline to discuss more.

examples/run_nlp_trainer.py
ishanashastri: Refactored out!

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+class GLUEMetricsLogger:
+    """ Class mapping all GLUE tasks to their respective average metric values. 
+
+    Args:
+        tasks_to_avg_metric dict(str, Any): dictionary mapping GLUE task names to their avg values.
+    
+    """
+    task_to_avg_metric: Dict[str, Any]  
+
+    def __init__(self, task_names) -> None:
```
ravi-mosaicml, line 21: When using a dataclass, no need to define `__init__` -- instead I'd make this a normal class, or use `__post_init__`.

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+    metric values. 
+    
+    Args: 
+        task_names list(str): the names of the GLUE tasks stored in the data struct
+        ckpt_to_tasks dict(str, GLUEMetricsLogger): dictionary mapping checkpoint names to GLUEMetricsLogger instances
+    """
+    task_names: List[str]
+    ckpt_to_tasks: Dict[str, GLUEMetricsLogger]
+        
+    def __init__(self, task_names, ckpt_to_tasks) -> None:
```
ravi-mosaicml, line 40: Same as above -- the default constructor for a dataclass will have this signature.

composer/trainer/trainer_hparams.py
```diff
@@ -414,7 +414,7 @@ def initialize_object(self) -> Trainer:
...
+            device = DeviceGPU(dist.get_local_rank()) if torch.cuda.is_available() else DeviceCPU()
```
ravi-mosaicml, line 417: Curious why is this change needed? I believe this is already default behavior?

composer/utils/entrypoint.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for defining custom entrypoints."""
+
+from typing import Type
+
+__all__ = ['_warning_on_one_line']
```
ravi-mosaicml, line 8: nit: private methods should not be exported.

Would it make sense to call this file as `warning_helpers.py`? idk, don't have strong opinions on the name.

composer/utils/entrypoint.py
```diff
@@ -0,0 +1,13 @@
...
+
+def _warning_on_one_line(message: str, category: Type[Warning], filename: str, lineno: int, file=None, line=None):
```
ravi-mosaicml, line 10:
```suggestion
def warning_on_one_line(message: str, category: Type[Warning], filename: str, lineno: int, file=None, line=None):
```

examples/run_nlp_trainer.py
ravi-mosaicml: Is `autoresume=True`? If so then it'll start from scratch. Otherwise it should crash.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+from composer.trainer.devices.device_gpu import DeviceGPU
+from composer.trainer.trainer_hparams import TrainerHparams
+from composer.utils.entrypoint import _warning_on_one_line
+from composer.utils.object_store.object_store_hparams import S3ObjectStoreHparams
+from composer.utils.object_store.s3_object_store import S3ObjectStore
+
+# TODO: reset all configs and training lengths back to default
+
+''' Generate a multiprocessing queue to store queue_size GPU IDs. '''
+def init_cuda_queue(queue_size: int) -> mp.Queue:
```
ravi-mosaicml, line 58: Based on our offline conversation, do we need the queue? It seems a bit heavy to have to use a queue to pass a unqiue id to each worker. Instead, could each worker call `multiprocessing.current_process()` (https://docs.python.org/3/library/multiprocessing.html#multiprocessing.current_process) within `init_cuda_env`?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+    cuda_envs = mp.Queue(queue_size)
+    cuda_envs_list = range(queue_size)
+    for e in cuda_envs_list:
+        cuda_envs.put(e)
+
+    return cuda_envs
+
+''' Set up a single GPU CUDA environment on initialization of a mp process pool. '''
```
ravi-mosaicml, line 66: nit: same comment from before re: putting docstrings inside the functions

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+        # future.add_done_callback(log_metrics)
+        futures.append(future)
+
+    executor.shutdown(wait=True) # wait for processes and callbacks to complete 
+  
+    cuda_envs.close() 
+    cuda_envs.join_thread()
+    
+'''Run single instance of a finetuning job on given task. ''' 
+def train_finetune(task: str, save_ckpt: bool, load_path: str, wandb_group_name: str, save_folder: str, save_locally: bool, wandb_logger: Union[WandBLogger, None], bucket: Optional[str] = None, load_ignore_keys: Optional[List[str]] = None):
```
ravi-mosaicml, line 168: Can we remove the `wandb` and `bucket` arguments here, and instead require that the user specify wandb and/or the object store logger as part of their yaml?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+    
+    if parent_ckpt: 
+        wandb_group_name = parent_ckpt
+    else:
+        wandb_group_name = ckpt_load_path
+
+    # finetuning from pretrained checkpoint
+    print(f"FINETUNING ON {ckpt_filename}!")
+    futures = []
+    log_metrics_partial = partial(log_metrics, parent_ckpt, ckpt_filename, glue_metrics)
```
ravi-mosaicml, line 154: nit: prefer lambda over functools.partial for pyright support.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+def train_finetune(task: str, save_ckpt: bool, load_path: str, wandb_group_name: str, save_folder: str, save_locally: bool, wandb_logger: Union[WandBLogger, None], bucket: Optional[str] = None, load_ignore_keys: Optional[List[str]] = None):
+    ft_hparams = TrainerHparams.create(cli_args=False, f=f'./composer/yamls/models/glue/{task}.yaml') 
+    ft_hparams.load_path = load_path
+    ft_hparams.device = DeviceGPU(torch.cuda.current_device())
+    ft_hparams.log_to_console = False 
+    ft_hparams.progress_bar = False
+    ft_hparams.load_ignore_keys = load_ignore_keys
+    
+    # add finetune-specific tags to the logger 
+    if wandb_logger:
```
ravi-mosaicml, line 177: Given the above comment, let's see if we can remove this line through line 209. Would prefer not to duplicate implementation that yahp already supports.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+
+    print(f"\n --------\n SPAWNING TASK {task.upper()}\n DEVICE: {torch.cuda.current_device()}\n --------")
+
+    trainer = ft_hparams.initialize_object()
+    trainer.fit()
+    print(f"\nFINISHED TRAINING TASK {task.upper()}\n")
+    return trainer.state.current_metrics
+
+'''Get the run name and S3 and WandB instances by pre-constructing a dummy TrainerHparams instance. ''' 
+def get_training_info():
```
ravi-mosaicml, line 219: Curious why do we need this function? Similar to the above, would like to not presume that wandb or the object store logger is being used.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+        if args.training_scheme == 'all':
+            args.save_folder = 'pretrained'
+            save_artifact_name = '{run_name}/pretrained/ep{epoch}-ba{batch}-rank{rank}-wct{total_wct}'
+            subprocess_args = ['composer', training_script, '-f', f"{args.file}" , '--save_folder', args.save_folder, 
+                    '--save_artifact_name', save_artifact_name, '--save_num_checkpoints_to_keep', '0']
+        # only pretraining, save to default artifact name 
+        else: 
+            subprocess_args = ['composer', training_script, '-f', f"{args.file}", '--save_num_checkpoints_to_keep', '0']
+    
+    proc = subprocess.Popen(subprocess_args)
```
ravi-mosaicml, line 277: nit: use subprocess.run(check=True) -- that will raise on errors, and then the call to wait wont be needed. Additional args might be needed to pass stdout/stderr through (but not sure)

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+    # saving on cloud
+    else:
+        if args.training_scheme == 'all':
+            args.save_folder = 'pretrained'
+            save_artifact_name = '{run_name}/pretrained/ep{epoch}-ba{batch}-rank{rank}-wct{total_wct}'
+            subprocess_args = ['composer', training_script, '-f', f"{args.file}" , '--save_folder', args.save_folder, 
+                    '--save_artifact_name', save_artifact_name, '--save_num_checkpoints_to_keep', '0']
+        # only pretraining, save to default artifact name 
+        else: 
+            subprocess_args = ['composer', training_script, '-f', f"{args.file}", '--save_num_checkpoints_to_keep', '0']
```
ravi-mosaicml, line 275: can we rely on the user's yaml to set save_num_checkpoints_to_keep? They might want to keep checkpoints locally.

ravi-mosaicml, line 275:
```suggestion
            subprocess_args = ['composer', training_script, '-f', args.file, '--save_num_checkpoints_to_keep', '0']
```

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+            new_save_folder = args.save_folder
+        
+        subprocess_args = ['composer', training_script, '-f', f"{args.file}", '--save_folder', f"{new_save_folder}"]
+       
+    # saving on cloud
+    else:
+        if args.training_scheme == 'all':
+            args.save_folder = 'pretrained'
+            save_artifact_name = '{run_name}/pretrained/ep{epoch}-ba{batch}-rank{rank}-wct{total_wct}'
+            subprocess_args = ['composer', training_script, '-f', f"{args.file}" , '--save_folder', args.save_folder,
```
ravi-mosaicml, line 271:
```suggestion
            subprocess_args = ['composer', training_script, '-f', args.file , '--save_folder', args.save_folder, 
```

Same theme of comment as below about not overriding the save arguments -- reuqire that the user specify it in the yaml. OK to default to a default if not specified.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+        if isinstance(l, WandBLogger):
+            wandb_logger = l 
+    return (run_name, save_locally, bucket, wandb_logger)
+
+''' Validate NLP specific entrypoint arguments. ''' 
+def validate_args(args) -> None:
+    if args.training_scheme == 'finetune' and isinstance(args.load_path, dataclasses._MISSING_TYPE): 
+        raise ValueError('load_path to checkpoint folder must be specified if finetuning a model')
+
+    if not isinstance(args.load_path, dataclasses._MISSING_TYPE) and args.training_scheme != 'finetune':
```
ravi-mosaicml, line 238: Could the load path be specified in the yaml and not in the cli args?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+        if isinstance(l, ObjectStoreLogger):
+            save_locally = False
+            bucket = l.object_store.bucket
+        if isinstance(l, WandBLogger):
+            wandb_logger = l 
+    return (run_name, save_locally, bucket, wandb_logger)
+
+''' Validate NLP specific entrypoint arguments. ''' 
+def validate_args(args) -> None:
+    if args.training_scheme == 'finetune' and isinstance(args.load_path, dataclasses._MISSING_TYPE):
```
ravi-mosaicml, line 235: Can we avoid checking against MISSING_TYPE?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,373 @@
...
+                glue_large += task_metric
+        # GLUE Large and GLUE All
+        output_line.append("{:.4f}".format(glue_large/len(large_tasks)))
+        output_line.append("{:.4f}".format(glue_all/len(tasks)))
+        tb.append(output_line)
+    
+    print(tabulate(tb, headers='firstrow'))
+
+''' Callback function for metric collection. '''
+def log_metrics(parent_ckpt: Optional[str], ckpt_filename: str, glue_metrics: GlueState, f_metric: Future) -> None:
```
ravi-mosaicml, line 120: Could we just take a `chkpt_filename` instead of parent_ckpt and ckpt_filename?

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def init_ckpt_dict(self, ckpt_name: str):
+        """ Create an empty GLUEMetricsLogger instance with all the tasks for a new checkpoint. """
+        self.ckpt_to_tasks[ckpt_name] = GLUEMetricsLogger(self.task_names)
+
+    def get_logged_ckpts(self) -> List[str]:
+        """ Get the list of currently logged checkpoints. """
+        return list(self.ckpt_to_tasks.keys())
+    
+    def set_task_metric(self, ckpt_name, task_name, metric) -> None:
```
ravi-mosaicml, line 52: Do we need this helper? Or can the code access the nested datastructure directly?

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+        """" Add a new metric value to the average value for a given task and checkpoint. """
+        self.ckpt_to_tasks[ckpt_name].task_to_avg_metric[task_name] = metric
+
+    def get_task_metric(self, ckpt_name, task_name) -> int:
```
ravi-mosaicml, line 56: Do we need this helper? Or can the code access the nested datastructure directly?

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+        
+    def __init__(self, task_names, ckpt_to_tasks) -> None:
+        self.task_names = task_names
+        self.ckpt_to_tasks = ckpt_to_tasks
+
+    def init_ckpt_dict(self, ckpt_name: str):
+        """ Create an empty GLUEMetricsLogger instance with all the tasks for a new checkpoint. """
+        self.ckpt_to_tasks[ckpt_name] = GLUEMetricsLogger(self.task_names)
+
+    def get_logged_ckpts(self) -> List[str]:
```
ravi-mosaicml, line 48: Do we need this helper? Or can the code access the nested datastructure directly?

composer/metrics/glue.py
```diff
@@ -0,0 +1,58 @@
...
+        ckpt_to_tasks dict(str, GLUEMetricsLogger): dictionary mapping checkpoint names to GLUEMetricsLogger instances
+    """
+    task_names: List[str]
+    ckpt_to_tasks: Dict[str, GLUEMetricsLogger]
+        
+    def __init__(self, task_names, ckpt_to_tasks) -> None:
+        self.task_names = task_names
+        self.ckpt_to_tasks = ckpt_to_tasks
+
+    def init_ckpt_dict(self, ckpt_name: str):
```
ravi-mosaicml, line 44: Since this is used in just one place, can it be hardcoded in?

ravi-mosaicml (commented):
Thanks for giving it another revision! I took another pass -- I didn't give the logic too close of a read, but curious if we can simplify it even more? Mainly wondering if we can remove the object store and wandb logic from the entrypoint. Also left some comments re general code style -- e.g. docstrings should go inside the functions; see if there are any helpers that could be removed if they are used in just one or two places.

composer/metrics/glue.py
ishanashastri: All the helper methods were created to try to improve readability as per your previous comment, but I can just refactor it back.

composer/utils/entrypoint.py
ishanashastri: yup renamed!

examples/run_nlp_trainer.py
ishanashastri: From my understanding, there is no way to extract a unique ID from mp.current_process(), which returns a Process object. The only way is to either somehow map process IDs to their GPU equivalents, or parse the worker number out of the Process name, neither of which seem to be less heavy or cleaner than the queue.

examples/run_nlp_trainer.py
ishanashastri: switched!

examples/run_nlp_trainer.py
ishanashastri: Right now, the user has to specify object store and loggers via their yaml. I removed the CLI args and instead only instantiate object store/wandb if it's present in the yaml.

examples/run_nlp_trainer.py
ishanashastri: As discussed offline, a lot of this logic is to make automation of pretraining + finetuning easier. If we were to remove all this logic and instead require the users to specify all the saving and artifact logic in their yamls, it would be confusing to them and it would be difficult to streamline the training process.

examples/run_nlp_trainer.py
ishanashastri: Discussed offline -- will keep the queue and add justification in documentation.

examples/run_nlp_trainer.py
ishanashastri: Discussed offline -- keeping to control task specific logging and saving!

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,350 @@
...
+''' Get TrainerHparams arguments from CLI and add NLP entrypoint specific arguments. ''' 
+def get_args():
+    parser = hp.get_argparse(TrainerHparams)
+    parser.add_argument('--training_scheme', help='training scheme used (one of "pretrain", "finetune", or "all")')
+    parser.add_argument('--save_locally', action=argparse.BooleanOptionalAction, help='save to local directory or to cloud using object store if --no-save-locally is enabled')
+    parser.add_argument('--bucket', help='S3 bucket name to pull from/save to if --no-save_locally is enabled')
+    
+    args = parser.parse_args()
+    validate_args(args)
+    return args
```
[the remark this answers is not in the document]
ishanashastri: What's the reasoning behind this? Is it not easier to just pass around the args object to the different helpers?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+            save_ckpts,
+            ckpt_load_path,
+            ckpt_filename,
+            finetune_hparams.save_folder,
+            glue_metrics,
+            args.file,
+            save_locally=save_locally,
+            parent_ckpt=parent_ckpt,
+            load_ignore_keys=['state/model/model.classifier*'],
+        )  # type: ignore
```
ravi-mosaicml, line 407: curious why do we need the type ignore?

ravi-mosaicml (commented):
Thanks for adding in the finetune hparams -- I think that helps clean it up a bunch. I gave it another pass -- it still seems to have some wandb/ s3 specific references. Ideally this script would not need to import wandb or boto3. In addition, is there a race condition with mnli? We can discuss more offline.

composer/yamls/models/glue/glue_example.yaml
```diff
@@ -0,0 +1,104 @@
...
+  model:
+    bert:
+      use_pretrained: false
+      tokenizer_name: bert-base-uncased
+      pretrained_model_name: bert-base-uncased
+
+  # Train the model on the English C4 corpus
+  train_dataset:
+    streaming_c4:
+      remote: s3://allenai-c4/mds/1/
```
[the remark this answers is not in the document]
ishanashastri: This is the same configuration that is used in bert-base.yaml.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    ft_hparams.log_to_console = False
+    ft_hparams.progress_bar = False
+
+    if ft_hparams.load_ignore_keys:
+        ft_hparams.load_ignore_keys.extend(load_ignore_keys)
+    else:
+        ft_hparams.load_ignore_keys = load_ignore_keys
+
+    # add finetune-specific tags to wandb if logger exists
+    if ft_hparams.loggers:
```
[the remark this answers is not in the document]
ishanashastri: I believe we've discussed this before, but this comes down to if it's a priority to not have to make the users do so much work to be able to use the entrypoint. I'm a little afraid of the documentation becoming a long list of nit things the user would have to do to make the output of the entrypoint usable (such as adding a tag in all 8 YAMLs but putting the rest of the wandb logic in the shared `finetune_hparams` section versus adding a few wandb specific lines into the entrypoint that handle all the small automation stuff.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+                if not logger._init_kwargs['tags']:
+                    logger._init_kwargs['tags'] = []
+                logger._init_kwargs['tags'].append(task)
+                logger._init_kwargs['group'] = wandb_group_name
+                logger._rank_zero_only = True
+
+    # saving single checkpoint at the end of training the task
+    if save_ckpt:
+        # add task specific artifact logging informatino
+        if save_locally:
```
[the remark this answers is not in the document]
ishanashastri: Also discussed offline previously but we can't do this because it's all task specific saving logic.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+                        required=True)
+
+    args = parser.parse_args()
+
+    return args
+
+
+def validate_args(args: argparse.Namespace, hp: NLPTrainerHparams) -> None:
+    """Validate CLI args as well as finetune-specific parameters."""
+    if not args.file:
```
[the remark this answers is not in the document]
ishanashastri: Without this, YAHP will spit out an error message saying`ValueError: The following required fields were not included in the yaml nor the CLI arguments: pretrain_hparams.model` which is potentially unhelpful unless you're familiar with composer?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+
+    return hparams
+
+
+def run_pretrainer(args) -> None:
+    """Logic for handling a pretraining job spawn based on storage and training settings."""
+    root_dir = os.path.join(os.path.dirname(composer.__file__), '..')
+    training_script = os.path.join(root_dir, 'examples/run_composer_trainer.py')
+
+    # manually copy pretrain_hparams to temporary file (workaround until CO-766 resolved)
```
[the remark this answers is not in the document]
ishanashastri: This is the issue I was referring to: https://mosaicml.atlassian.net/browse/CO-766. Originally the plan was to extract just the pretrain_hparams and then dump the `.to_yaml()` to a file (which is much simpler), but since YAHP has this bug I can't use that trick.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    # set automatic load and save paths
+    if args.training_scheme == 'all':  # load checkpoints from where pretraining saved them
+        ckpt_folder = args.save_folder
+    else:  # finetune only
+        ckpt_folder = finetune_hparams.load_path
+
+    if args.save_locally:
+        all_ckpts_list = [f for f in os.listdir(ckpt_folder) if not os.path.isdir(os.path.join(ckpt_folder, f))]
+    else:
+        all_ckpts_list = get_all_s3_checkpoints(args.bucket, ckpt_folder)
```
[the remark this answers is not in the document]
ishanashastri: I believe we also discussed this offline and agreed that using boto3 to find all the relevant checkpoints was not too expensive and that I can keep it without adding the manual list of checkpoints?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+
+    # finetune on every pretrained checkpoint
+    for ckpt_filename in all_ckpts_list:
+        parent_ckpt = ckpt_filename  # necessary for logging
+        if args.save_locally:
```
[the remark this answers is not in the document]
ishanashastri: The parent checkpoint is needed only for logging as described in one of the previous comments. It's not something used for loading/anything the hparams stores.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+                              ckpt_filename,
+                              ckpt_folder,
+                              glue_metrics,
+                              args.file,
+                              save_locally=save_locally,
+                              parent_ckpt=parent_ckpt,
+                              load_ignore_keys=['state/model/model.classifier*'])
+
+        # finetune on inference tasks using last mnli checkpoint
+        if args.save_locally:
```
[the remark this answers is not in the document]
ishanashastri: Yes! MNLI saves a checkpoint and then the last three finetune tasks finetune on MNLI's last checkpoint.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def log_metrics(ckpt_filename: str, glue_metrics: GlueState, f_metric: Future) -> None:
+    """Callback function for metric collection."""
+    metric = f_metric.result()
+
+    if ckpt_filename not in glue_metrics.get_logged_ckpts():
+        glue_metrics.init_ckpt_dict(ckpt_filename)
+
+    task = list(metric.keys())[0]
```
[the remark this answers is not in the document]
ishanashastri: It's less so the first element and moreso trying to extract the metric name if the metric is stored as `{'glue_cola': {'Accuracy': 0.98}} `

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    metric = f_metric.result()
+
+    if ckpt_filename not in glue_metrics.ckpt_to_tasks.keys():
+        glue_metrics.ckpt_to_tasks[ckpt_filename] = GLUEMetricsLogger(glue_metrics.task_names)
+
+    task = list(metric.keys())[0]
+    formatted_task = task.split('_')[1]  # remove "glue_" prefix
+    for key in metric.keys():  # handle case where mnli has glue_mnli and glue_mnli_mismatched
+        for m in metric[key]:
+            if not glue_metrics.ckpt_to_tasks[ckpt_filename].task_to_avg_metric[formatted_task]:
```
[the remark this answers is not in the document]
ishanashastri: Would it work to put back the helper methods that were previously defined in the GLUEState class? I think you mentioned previously that because it's only used in one or two places that I should just directly put it in the code.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    proc.communicate()  # block until training is complete, doesn't interfere with open file streaming
+
+
+def run_finetuner(args: argparse.Namespace, finetune_hparams, glue_metrics: GlueState) -> None:
+    """Logic for handling a finetuning job spawn based on storage and training settings."""
+    # set automatic load and save paths
+    if args.training_scheme == 'all':  # load checkpoints from where pretraining saved them
+        ckpt_folder = args.save_folder
+    else:  # finetune only
+        ckpt_folder = finetune_hparams.load_path
```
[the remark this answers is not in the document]
ishanashastri: folder

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def init_cuda_env(cuda_envs: mp.Queue) -> None:
+    """Set up a single GPU CUDA environment on initialization of a mp process pool."""
+    env = cuda_envs.get()
+    torch.cuda.set_device(env)
+
+    # fake a single node world
+    os.environ['WORLD_SIZE'] = '1'
+    os.environ['LOCAL_WORLD_SIZE'] = '1'
```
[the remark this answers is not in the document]
alextrott16: This might be a moot concern, but is hardcoding ports risky if two fine-tuning workloads were running on the same node?

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    formatted_task = task.split('_')[1]  # remove "glue_" prefix
+    for key in metric.keys():  # handle case where mnli has glue_mnli and glue_mnli_mismatched
+        for m in metric[key]:
+            if not glue_metrics.ckpt_to_tasks[ckpt_filename].task_to_avg_metric[formatted_task]:
+                glue_metrics.ckpt_to_tasks[ckpt_filename].task_to_avg_metric[formatted_task] = metric[key][m].item()
+
+            else:
+                old_val = glue_metrics.ckpt_to_tasks[ckpt_filename].task_to_avg_metric[formatted_task]
+                glue_metrics.ckpt_to_tasks[ckpt_filename].task_to_avg_metric[formatted_task] = sum(
+                    [old_val, metric[key][m].item()]) / 2  # automatically average scores
```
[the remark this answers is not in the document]
ishanashastri: Oops, nice catch! Running average needed for the GLUE scores but fixed during printing.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    trainer = ft_hparams.initialize_object()
+    trainer.fit()
+    print(f'\nFINISHED TRAINING TASK {task.upper()}\n')
+    return trainer.state.current_metrics
+
+
+def get_args() -> argparse.Namespace:
+    """Get NLPTrainerHparams arguments from CLI."""
+    parser = hp.get_argparse(NLPTrainerHparams)
+    parser.add_argument('--training_scheme',
```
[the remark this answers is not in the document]
ishanashastri: Sure, but we'll still need argparse to get the filename

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+
+    if args.training_scheme == 'all':  # extract run_name from trainer args for finetuning
+        hp = TrainerHparams.create(cli_args=False, f=tmp_file)
+        args.run_name = hp.run_name
+        if args.save_locally:
+            args.save_folder = hp.save_folder
+        else:
+            args.save_folder = os.path.join(args.run_name, hp.save_folder)
+
+    proc = subprocess.Popen(['composer', training_script, '-f', tmp_file, '--save_folder', args.save_folder])
```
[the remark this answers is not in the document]
ishanashastri: Resolved in one of the previous comments?

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: Would it make sense to put back the helper methods in the GlueState class? One of your previous comments suggested to just discard the helper methods and put it directly in the code so it seems like I'm going back and forth on this.

alextrott16 (commented):
Nice work! 

This generally looks good to me. One issue that I could imagine is that someone accidentally creates an enormous job by either storing too many checkpoints or pointing to a checkpoint directory with many files in it. A `max_checkpoints` argument (or something along those lines) might help to prevent this. Alternatively, something that subsamples available checkpoints so that only some fraction is used for fine-tuning (with that fraction being an argument).

While I feel good about the overall design, it remains challenging to mentally trace the logic that a hypothetical workload would follow. I think some stress testing is needed to expose any other issues or edge cases. If there is a way to merge this (so that testing can happen) while flagging this as still experimental with some warning, I'd say let's do that.

In the meantime, I'll defer to @ravi-mosaicml for approval since he has provided a lot of comments that still need to be addressed.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Hmm ok, does that mean there is a race condition here -- if these last finetuning tasks start before the mnli job finishes?

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Ah ok, if the dict is only expected to have a singular element, can you add a check for that, and error if it has multiple elements?

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Ah, by a variable, I mean something defined inline here, like:

```python
tasks = glue_metrics.ckpt_to_tasks[ckpt_filename]
metric = tasks.task_to_avg_metric[formatted_task]
```

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: I'm not a fan of wandb specific code, as customers may not be using wandb. That said, I do agree it provides some nice UX here, so I am OK having it until we have a better logging API (CC @eracah) that would not require any wandb specific code. I would like to keep the wandb and s3 specific code to a minimal, non-crucial cross-section, so the script would still work if someone is using, say, tensorboard over sftp.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: Hmm, ok, let's work through this later today. Sorry, I must be missing something here.

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ravi-mosaicml: hmm ok, can we declare the `pretrain_model` field as optional?

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: I dont think that condition exists because the finetune spawn calls `executor.shutdown()` which waits for all processes and callbacks to complete before returning, therefore the last 3 finetune tasks cannot start until that call finishes.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,434 @@
...
+import subprocess
+import sys
+import tempfile
+import warnings
+from concurrent.futures import Future
+from concurrent.futures import ProcessPoolExecutor as Pool
+from dataclasses import fields
+from typing import List, Optional, Union
+
+import boto3
```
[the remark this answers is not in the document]
ishanashastri: Not needed anymore with the new checkpoint saving changes.

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+    # pulls AWS credentials from environment variables, assumes that aws_session_token is defined
+    session = boto3.Session()
+
+    s3 = session.resource('s3')
+    bucket = s3.Bucket(bucket_name)
+    ckpt_names = []
+    for obj in bucket.objects.all():
+        if folder_name in obj.key:
+            obj_name = obj.key
+            ckpt_names.append(obj_name)
```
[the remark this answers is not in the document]
ishanashastri: No longer addressing after storing the list of checkpoints!

examples/run_nlp_trainer.py
```diff
@@ -0,0 +1,439 @@
...
+
+def run_finetuner(args: argparse.Namespace, finetune_hparams, glue_metrics: GlueState) -> None:
+    """Logic for handling a finetuning job spawn based on storage and training settings."""
+    # set automatic load and save paths
+    if args.training_scheme == 'all':  # load checkpoints from where pretraining saved them
+        ckpt_folder = args.save_folder
+    else:  # finetune only
+        ckpt_folder = finetune_hparams.load_path
+
+    if args.save_locally:
```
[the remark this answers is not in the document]
ishanashastri: No longer addressing with saving the list of checkpoints!

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: There's actually one edge case where `mnli` has two elements: `glue_mnli` and `glue_mnli_mismatched`. The task name doesn't change though so it's still extracted from the first.

composer/yamls/models/glue/glue_example.yaml
```diff
@@ -0,0 +1,104 @@
...
+  save_folder: checkpoints          # The directory to save checkpoints to
+  save_interval: 3500ba              # Save checkpoints every 3500 batches
+  save_artifact_name: '{run_name}/checkpoints/ep{epoch}-ba{batch}-rank{rank}-wct{total_wct}' # uncomment if you want to checkpoint to cloud 
+  save_overwrite: True
+
+  loggers:
+    object_store:
+      object_store_hparams:
+        s3:
+          bucket: ishana-bert-checkpoints
```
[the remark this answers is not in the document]
ishanashastri: Do we have a public bucket?

examples/run_nlp_trainer.py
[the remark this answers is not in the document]
ishanashastri: I'll file a JIRA to back fix this after the YAHP fixes are released!

alextrott16 (approved):
Thanks for addressing these issues! LGTM.

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,482 @@
...
+    warnings.formatwarning = warning_on_one_line
+
+    if len(sys.argv) == 1:
+        sys.argv = [sys.argv[0], '--help']
+
+    file, training_scheme = get_args()
+    finetune_hparams, save_locally = get_finetune_hparams(training_scheme)
+
+    # Pretrain
+    if training_scheme != 'finetune':
```
ravi-mosaicml, line 465: nit: for readability, can we do 
```suggestion
    if training_scheme in ('pretrain', 'all'):
```

And same thing below on line 472.

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,482 @@
...
+    if training_scheme != 'pretrain':
+        run_finetuner(training_scheme, file, save_locally, finetune_hparams.save_folder, finetune_hparams, glue_metrics)
+        print('FINETUNING COMPLETE')
+
+        # output GLUE metrics
+        print_metrics(glue_metrics)
+
+
+if __name__ == '__main__':
+    mp.set_start_method('spawn')
```
ravi-mosaicml, line 481: Instead of setting the mp start method globally, would it be possible to use a mp context? See the object store logger as an example...

ravi-mosaicml (commented):
Looking really close! Just a bunch of nits, and one concern re: determining the checkpoint names.

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,482 @@
...
+    ckpt_names = []
+    interval = Time.from_timestring(hp.save_interval)
+    duration = Time.from_timestring(hp.max_duration)
+
+    ep = 0
+    ba = 0
+    loop = True
+    save = False
+    save_last_batch = False
+    while loop:
```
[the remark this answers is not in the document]
ishanashastri: Discussed offline, will leave as is until object store logging is refactored!

composer/trainer/nlp_trainer_hparams.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def load(cls, model: str) -> NLPTrainerHparams:
+        model_hparams_file = os.path.join(
+            os.path.dirname(composer.__file__),
+            'yamls',
+            'models',
+            f'{model}.yaml',
+        )
+        trainer_hparams = NLPTrainerHparams.create(model_hparams_file, cli_args=False)
+        assert isinstance(trainer_hparams, NLPTrainerHparams), 'trainer hparams should return an instance of self'
```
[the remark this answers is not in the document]
ishanashastri: is in the equivalent method of `TrainerHparams` but not necessary

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,482 @@
...
+        ckpt_names.append(formatted_ckpt_name)
+        
+    return ckpt_names
+
+def run_pretrainer(training_scheme: str, file: str, save_locally: bool, finetune_hparams: GLUETrainerHparams) -> None:
+    """Logic for handling a pretraining job spawn based on storage and training settings."""
+    root_dir = os.path.dirname(__file__)
+    training_script = os.path.join(root_dir, 'run_composer_trainer.py')
+
+    # manually copy pretrain_hparams to temporary file (workaround until CO-766 resolved)
```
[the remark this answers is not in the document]
mvpatel2000: Yahp version released! Should be good to go

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,503 @@
...
+from dataclasses import dataclass
+from multiprocessing.context import BaseContext
+from typing import Dict, List, Optional, Tuple
+
+import torch
+import yahp as hp
+import yaml
+from tabulate import tabulate
+
+import examples.glue.nlp_trainer_hparams.GLUETrainerHparams as GLUETrainerHparams
```
ravi-mosaicml, line 42: Can you put this file in the `glue` folder, and try:
```suggestion
from .nlp_trainer_hparams import GLUETrainerHparams, NLPTrainerHparams
```

examples/run_glue_trainer.py
```diff
@@ -0,0 +1,482 @@
...
+
+    Args:
+        metric (Dict): Metrics returned from ``train_finetune()`` for a given GLUE task
+        ckpt_filename (str): Checkpoint to log metrics under
+        glue_metrics (GlueState): GlueState object storing all the glue metrics for the entrypoint's current run
+    """
+    if ckpt_filename not in glue_metrics.ckpt_to_tasks.keys():
+        glue_metrics.ckpt_to_tasks[ckpt_filename] = GLUEMetricsState(glue_metrics.task_names)
+
+    task = list(metric.keys())[0]
```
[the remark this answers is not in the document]
ravi-mosaicml: Ah sorry, missed the comment. OK, I see that the `task` variable is only used in formatting the task name. That's fine then.

composer/utils/misc.py
```diff
@@ -25,3 +28,17 @@ def is_notebook():
...
+    return f'{category.__name__}: {message} (source: {filename}:{lineno})\n'
+
+
+def get_free_tcp_port() -> int:
+    # from https://www.programcreek.com/python/?CodeExample=get+free+port
+    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    tcp.bind(('', 0))
+    _, port = tcp.getsockname()
+    tcp.close()
+    return port
```
ravi-mosaicml, line 44: nit: newline

composer/utils/misc.py
```diff
@@ -25,3 +28,17 @@ def is_notebook():
...
         return False
+        
+def warning_on_one_line(message: str, category: Type[Warning], filename: str, lineno: int, file=None, line=None):
+    """Force Python warnings to consolidate into one line."""
+    # From https://stackoverflow.com/questions/26430861/make-pythons-warnings-warn-not-mention-itself
+    return f'{category.__name__}: {message} (source: {filename}:{lineno})\n'
+
+
+def get_free_tcp_port() -> int:
+    # from https://www.programcreek.com/python/?CodeExample=get+free+port
```
ravi-mosaicml, line 39: Docstring

ravi-mosaicml (approved):
LGTM! :rocket: See comments. Feel free to merge after validating that the relative import from the examples folder works.

moinnadeem:
@ishanashastri Am I expected to be able to do something like `composer examples/glue/run_glue_trainer.py --help`? If not, do you think that would be in scope to add?

moinnadeem (commented):
Starting to use it now, and can give detailed feedback then! Once we address the `--help` feature, I think we should be able to give this an LGTM!

moinnadeem:
Quick question, when running `--help`, I see this:

```
/root/composer/composer/cli/launcher.py:223: UserWarning: AutoSelectPortWarning: The distributed key-value port was auto-selected. This may lead to race conditions when launching multiple training processes simultaneously. To eliminate this race condition, explicitly specify a port with --master_port PORT_NUMBER
  warnings.warn('AutoSelectPortWarning: The distributed key-value port was auto-selected. '
usage: run_glue_trainer.py [-h] [-s [stdout]] [-i] [-c] [-f FILE] [-d [stdout]] [--validate] [--training_scheme TRAINING_SCHEME] [--pretrain_hparams PRETRAIN_HPARAMS] [--finetune_hparams FINETUNE_HPARAMS]

optional arguments:
  -h, --help            show this help message and exit
  -s [stdout], --save_template [stdout]
                        Generate and dump a YAML template to the specified file (defaults to `stdout`) and exit.
  -i, --interactive     Whether to generate the template interactively. Only applicable if `--save_template` is present.
  -c, --concise         Skip adding documentation to the generated YAML. Only applicable if `--save_template` is present.
  -f FILE, --file FILE  Load data from this YAML file into the Hparams.
  -d [stdout], --dump [stdout]
                        Dump the resulting Hparams to the specified YAML file (defaults to `stdout`) and exit.
  --validate            Whether to validate YAML against Hparams.

NLPTrainerHparams:
  --training_scheme TRAINING_SCHEME
                        <str> training scheme used (one of "pretrain", "finetune", or "all") (Default: all).
  --pretrain_hparams PRETRAIN_HPARAMS
                        <Optional[TrainerHparams]> Pretraining hyperparameters (Default: None).
  --finetune_hparams FINETUNE_HPARAMS
                        <Optional[GLUETrainerHparams]> GLUE Finetuning hyperparameters (Default: None).
```

However, `pretrain_hparams` doesn't tell me anything: what can I actually fill in there?
Is this in scope to fix?

PR #1265, opened 2022-07-07
Add combo loss to DeepLabv3+

Landanjs:
This PR adds a slight variation of the [MONAI dice loss](https://docs.monai.io/en/stable/_modules/monai/losses/dice.html#DiceLoss) to composer and provides the option to use both cross entropy and dice loss when training DeepLabv3+. Cross entropy and dice can be selected by providing `cross_entropy_weight` and `dice_weight` arguments that are > 0. A helper function is included in DeepLabv3+ to return a list of losses based on the weight for the cross entropy and dice losses. Note: it may be easier if `ComposerClassifier` accepted a sequence of callables, but TBD if we want to continue to extend this class.

Some changes were made to `ensure_targets_one_hot`. The first change required cloning targets since in-place operations on `target` prevented `target` from being used with multiple losses. Second, PyTorch's `one_hot` function seems to be ~10% slower than the operation in MONAI, so I switched to a variation similar to MONAI.

I did an initial pass at tests by checking two extreme scenarios: completely correct predictions and completely incorrect predictions. Please let me know if more tests are needed or if the tests can be organized another way

[Most recent WandB runs](https://wandb.ai/mosaic-ml/dlv3p-ade20k-dice/groups/dice-0.25-0.75-ignore-neg-inds/workspace?workspace=user-landanjs)


composer/loss/utils.py
```diff
@@ -47,23 +46,37 @@ def ensure_targets_one_hot(input: torch.Tensor,
...
         if targets.min() < 0:
             warnings.warn('Negative label indices are being ignored in conversion to one-hot labels')
+
+            # Add new dimension for the class vectors
+            class_dim = 1
+            targets = targets.clone().unsqueeze(class_dim).long()
+
             # Map all negative indicies to a class to drop.
             targets[targets < 0] = num_classes
-            targets = F.one_hot(targets, num_classes=num_classes + 1)
```
hanlint, line 58: Prefer to use the torch one_hot, I'm not sure a 10% optimization on the loss calculation (on your benchmarked tensor sizes) is worth the implementation complexity? or at minimum, refactor the MONAI-based implementation into a `def _one_hot(..)` method for readability?

otherwise, it's unclear here what changes here are refactoring the one hot comptuation, and which are net new functional changes.

composer/models/deeplabv3/model.py
```diff
@@ -186,14 +193,28 @@ def composer_deeplabv3(num_classes: int,
...
+        [CrossEntropy(ignore_index=ignore_index),
+         MIoU(num_classes, ignore_index=ignore_index)])
+    val_metrics = MetricCollection(
+        [CrossEntropy(ignore_index=ignore_index),
+         MIoU(num_classes, ignore_index=ignore_index)])
+
+    ce_loss_fn = functools.partial(soft_cross_entropy, ignore_index=ignore_index)
+    dice_loss_fn = DiceLoss(softmax=True, batch=True, ignore_absent_classes=True)
+
+    def _combo_loss(output, target):
```
hanlint, line 206: curious why multiple losses are required here, versus a single loss:
```
loss_fn = lambda output, target: ce_loss_fn(output, target) * cross_entropy_weight + dice_loss_fn(output, target) * dice_weight
```

tests/test_loss.py
```diff
@@ -81,3 +81,55 @@ def test_soft_cross_entropy(self, tensors, use_weights, reduction):
...
+        input = F.one_hot(input)
+        input = torch.movedim(input, 3, 1)
+        return input
+
+    @pytest.fixture()
+    def incorrect_input(self, correct_input):
+        return correct_input[[3, 2, 1, 0]]
+
+    @pytest.mark.filterwarnings('ignore::UserWarning')
+    @pytest.mark.parametrize('squared_pred', [True, False])
```
hanlint, line 107: These parameterize that are shared can be put on the class itself:
```
@pytest.mkar.parametrize(..)
@pytest.mark.parametrize(..)
class TestDiceLoss:
    ...
```

tests/test_loss.py
```diff
@@ -81,3 +81,55 @@ def test_soft_cross_entropy(self, tensors, use_weights, reduction):
...
+    def test_correct_prediction(self, correct_input: torch.Tensor, target: torch.Tensor, squared_pred: bool,
+                                jaccard: bool, batch: bool, ignore_absent_classes: bool, reduction: str):
+        dice_loss = DiceLoss(squared_pred=squared_pred,
+                             jaccard=jaccard,
+                             batch=batch,
+                             ignore_absent_classes=ignore_absent_classes,
+                             reduction=reduction)
+
+        loss = dice_loss(correct_input, target)
+        assert loss == 0.0
```
hanlint, line 121:
```suggestion
        assert dice_loss(correct_input, target) == 0.0
```

composer/loss/loss.py
```diff
@@ -157,6 +158,123 @@ def soft_cross_entropy(input: Tensor,
         raise ValueError(f'Unrecognized target type {target_type}')
 
 
+class DiceLoss(_Loss):
```
hanlint, line 161: does this need to be a class? The other losses here are functions.

composer/loss/utils.py
Landanjs: The throughput without dice is ~210 images/sec. With the dice implementation that uses `F.one_hot`, it is ~170 images/sec. With the dice implementation with the new one hot conversion, it is ~190 images/sec. To me, this seems like a significant difference in throughput to justify the implementation change . Similarly, we see significant throughput differences with mixup and label smoothing, but the new implementation should see less impact. 

I can move the implementation to separate function.

hanlint: OK didn't realize the 10% was in overall throughput, not just the loss calculation itself. Then yes that's significant -- can you move to a separate function, thanks!

composer/models/deeplabv3/model.py
Landanjs: With the lambda above, dice would be calculated even if the weight is 0, so we will get a 10% throughput hit for a few lines of code, did not seem worth it to me.

I think there are two benefits to returning a tuple of losses:
1. We should be plotting individual losses instead of total loss. This is a standard practice to measure optimization on the individual losses. We aren't doing this in the trainer yet, but we absolutely should.
2. There are a few potential algorithms that do automatic multiple loss reweighting. Although there is a lot of skepticism around these algorithms, it would be great if we could test them by implementing them as an `Algorithm`. If we sum the losses within the model, I imagine we would need to implement these algorithms within the model.

hanlint: Thanks for the explanation, very helpful!

composer/loss/loss.py
Landanjs: I was torn on this, so happy to hear your opinion. This could be turned into a function, but I would need to use `functools.partial` in DeepLabv3's `model.py` since the loss needs new default arguments. I thought it would be cleaner to make a class, then to have a function and use `functools.partial`. But I don't have a strong opinion here

composer/loss/utils.py
Landanjs: I ended up moving most of the `ensure_targets_one_hot` functionality to the `_one_hot` function since it seemed to fit. Let me know if it matches expectations

tests/test_loss.py
Landanjs: Awesome, done! thanks 🙂

tests/test_loss.py
Landanjs: Done!

Landanjs:
Huh, it turns out our previous one-hot conversion method lead to noncontiguous tensors. Could this have been the cause of the slowdown?

The label smoothing tests use `F.one_hot()` and `.movedim()` to produce the reference labels. I needed to add `.contiguous()`, so the strides would match between the new and old methods.

Landanjs:
Hmm, getting a fused layer norm timeout. HuggingFace timeouts Hanlin mentioned?

composer/loss/utils.py
hanlint: Ideally `_one_hot` would be a drop-in replacement for `F.one_hot` that is faster, and respects the same semantics.  That also makes it easy to add tests, as there exists a reference implementation.

Is that possible in this case (e.g. keep the same code, but just replace `F.one_hot` -> `_one_hot`?), or is the optimization mixed in with some of the other bushiness logic?

hanlint (commented):
LGTM, one last question on `_one_hot`. In addition, since this is a significant change to a core component, could you add some testing around `_one_hot`?  Ideally can compare against the reference `F.one_hot`.

hanlint (approved):
Discussed offline -- giving preliminary approval.

Landanjs:
Added:
- Lots of tests for `_one_hot`
- `num_classes=-1` functionality
- `dim` argument to specify where the new class dimension will be specified
- Check in `_one_hot` that the input tensor is an index tensor
- Check in `_one_hot` that `num_classes` >= `tensor.max() + 1`

PR #1266, opened 2022-07-07
Add callback for visualizing image inputs and outputs

coryMosaicML:
This PR adds an ImageMonitor callback which logs input images during training/evaluation. Images are logged just before the forward pass so that the aggregate effect of all data transforms can be visualized. Additionally, there is a segmentation mode that allows visualization of segmentation masks.

[Example for `mode="input"` from our spicy ResNet50 recipe](https://wandb.ai/mosaic-ml/cory-image-monitor/runs/1qmdpgd7)
[Example for `mode="segmentation"` from DeepLabV3+ with cutmix added](https://wandb.ai/mosaic-ml/cory-image-monitor/runs/21a9rh4x)

Images are currently logged in wandb's image format, so this is really intended to be used with the `WandBLogger` but it should still log correctly with the others.


composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'input' and state.timestamp.get(self.interval.unit).value % self.interval.value == 0:
+            self._log_inputs(state, logger, 'Images/Train')
+
+    def eval_before_forward(self, state: State, logger: Logger):
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'input' and state.eval_timestamp.get(TimeUnit.BATCH).value == 0:
+            self._log_inputs(state, logger, 'Images/Eval')
+
+    def before_loss(self, state: State, logger: Logger):
```
eracah, line 142: Nit: is there a reason you chose BEFORE_LOSS instead of AFTER_FORWARD? It seems like no events happen in between them. Is there an algorithm that could run at AFTER_FORWARD that we want to run first? Maybe for clarity, we could do AFTER_FORWARD since we do EVAL_AFTER_FORWARD unless there is a reason not to.

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+        outputs += 1
+    table = wandb.Table(columns=['Image'])
+    for image, target, prediction in zip(images, targets, outputs):
+        mask = {'ground truth': {'mask_data': target}, 'prediction': {'mask_data': prediction}}
+        img_mask_pair = wandb.Image(image, masks=mask)
+        table.add_data(img_mask_pair)
+    return table
+
+
+def _check_for_image_format(data: torch.Tensor) -> bool:
```
eracah, line 193: Maybe here you could add the check for numpy array and PIL Image. Also the checking for multiple tensors can be checked here?

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+import torch
+
+from composer.core import Callback, State, Time, TimeUnit
+from composer.loggers import Logger
+from composer.utils.import_helpers import MissingConditionalImportError
+
+__all__ = ['ImageMonitor']
+
+
+class ImageMonitor(Callback):
```
eracah, line 16: What are your thoughts on `ImageVisualizer`? It seems more informative here for a user look for a callback. `ImageMonitor` could mean many things, whereas `ImageVisualizer` is a bit more specific.

eracah: or `ImageLogger`?

eracah (commented):
Hey, Cory, looks pretty good. I just had a few nits with regards to the acceptable image types and the name of the callback.

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+            raise ValueError(f'Invalid time string for parameter interval') from error
+
+        # Verify that the interval has supported units.
+        if self.interval.unit not in [TimeUnit.BATCH, TimeUnit.EPOCH]:
+            raise ValueError(f'Invalid time unit for parameter interval: '
+                             f'{self.interval.unit}')
+
+    def _log_inputs(self, state: State, logger: Logger, key: str):
+        inputs = state.batch_get_item(key=self.input_key)
+        if not isinstance(inputs, torch.Tensor):
```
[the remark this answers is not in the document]
Landanjs: This check might because `_make_input_images` does `images = inputs[0:num_images].data.cpu().permute(0, 2, 3, 1).numpy()` which would only be possible for torch tensors. How desirable would it be to have PIL images and numpy arrays separated? It would add some checking to `_make_input_images`, right?

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+            self.interval = Time.from_timestring(interval)
+        except ValueError as error:
+            raise ValueError(f'Invalid time string for parameter interval') from error
+
+        # Verify that the interval has supported units.
+        if self.interval.unit not in [TimeUnit.BATCH, TimeUnit.EPOCH]:
+            raise ValueError(f'Invalid time unit for parameter interval: '
+                             f'{self.interval.unit}')
+
+    def _log_inputs(self, state: State, logger: Logger, key: str):
```
Landanjs, line 108: Nit: `key` confused me a bit, would something like `data_name` or `input_name` be better?

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+                              num_images: int,
+                              shift_targets: bool = False):
+    import wandb
+    if min([inputs.shape[0], targets.shape[0], outputs.shape[0]]) < num_images:
+        num_images = min([inputs.shape[0], targets.shape[0], outputs.shape[0]])
+    images = inputs[0:num_images].data.cpu().permute(0, 2, 3, 1).numpy()
+    targets = targets[0:num_images].data.cpu().numpy()
+    outputs = outputs[0:num_images]
+    # Convert outputs to segmentation masks. Assume channels are first dim
+    outputs = outputs[0:num_images]
```
Landanjs, line 178: Accidental duplicate?

Landanjs:
```suggestion
    outputs = outputs[0:num_images]
    # Convert outputs to segmentation masks. Assume channels are first dim
```

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+def _make_segmentation_images(inputs: torch.Tensor,
+                              targets: torch.Tensor,
+                              outputs: torch.Tensor,
+                              num_images: int,
+                              shift_targets: bool = False):
+    import wandb
+    if min([inputs.shape[0], targets.shape[0], outputs.shape[0]]) < num_images:
+        num_images = min([inputs.shape[0], targets.shape[0], outputs.shape[0]])
+    images = inputs[0:num_images].data.cpu().permute(0, 2, 3, 1).numpy()
+    targets = targets[0:num_images].data.cpu().numpy()
```
Landanjs, line 175: Could targets be one-hot? Like for mixup? Would that work?

tests/callbacks/test_image_monitor.py
```diff
@@ -0,0 +1,45 @@
...
+    import wandb
+    _WANDB_INSTALLED = True
+    del wandb  # unused
+except ImportError:
+    _WANDB_INSTALLED = False
+
+
+@pytest.mark.skipif(not _WANDB_INSTALLED, reason='Wandb is optional')
+@pytest.mark.parametrize('interval', ['9ba', '90ba'])
+def test_image_monitor(interval: str):
```
Landanjs, line 24: Cool way to test the logging!

Landanjs (commented):
This is awesome!! A few suggestions, but the main things:
1. Would the segmentation logging work as expected if mixup or label smoothing is used? Maybe it is too complex of a feature to get working, but should at least through a warning about one hot targets (likely from algorithms used)
2. Could you add this callback to the main Callbacks document?
3. Extra indent for `doctest` so example is displayed correctly

composer/callbacks/image_monitor.py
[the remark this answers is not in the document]
Landanjs: This check might because `_make_input_images` does `images = inputs[0:num_images].data.cpu().permute(0, 2, 3, 1).numpy()` which would only be possible for torch tensors. How desirable would it be to have PIL images and numpy arrays separated? It would add some checking to `_make_input_images`, right?

composer/callbacks/image_monitor.py
coryMosaicML: I chose BEFORE_LOSS because I want to log the targets as the loss function sees them, so this seems safest to me. Some algorithms (cutmix for example) modify the targets on the BEFORE_LOSS event, so logging images on AFTER_FORWARD here wouldn't pick up those changes.

composer/callbacks/image_monitor.py
coryMosaicML: Agree, `ImageVisualizer` is better, I'll make that change.

composer/callbacks/image_monitor.py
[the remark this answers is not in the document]
coryMosaicML: Oh this is interesting. Do we ever expect it to be the case that an input or output to the network will not be a torch tensor on BEFORE_FORWARD or BEFORE_LOSS? A tensor conversion has to happen by the time all state changes have finished on BEFORE_FORWARD or the network will crash, right? Or, is there some case where one might handle the conversion to tensor in the forward function itself?

composer/callbacks/image_monitor.py
eracah: ok makes sense. thanks!

composer/callbacks/image_monitor.py
eracah: nice!

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+                self.interval.unit).value % self.interval.value == 0:
+            self._log_segmented_inputs(state, logger, 'Images/Train')
+
+    def eval_after_forward(self, state: State, logger: Logger):
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'segmentation' and state.eval_timestamp.get(TimeUnit.BATCH).value == 0:
+            self._log_segmented_inputs(state, logger, 'Images/Eval')
+
+
+def _make_input_images(inputs: torch.Tensor, num_images: int):
```
eracah, line 154: you could change the inputs type to `Union[torch.Tensor, np.ndarray, PIL.Image]` and do some `if isinstance` or `try... except` logic and then cast it to `torch.tensor`.

eracah (approved):

composer/callbacks/image_monitor.py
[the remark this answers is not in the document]
coryMosaicML: Thats a good point, thanks for bringing it up! I'll add a check, and then in another PR I can add the support for PIL and numpy images.

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+            batches. Eval images are logged once at the start of each eval. Default: ``"100ba"``.
+        mode (str, optional): How to log the image labels. Valid values are ``"input"`` (input only)
+            and "segmentation" which also saves segmentation masks. Default: ``"input"``.
+        num_images (int, optional): Number of images to log. Should be less than or equal to than the microbatch size.
+            If there are not enough images in the microbatch, all the images in the microbatch will be logged.
+            Default: ``8``.
+        shift_targets (bool): For use when ``mode="segmentation"``. Optionally shifts the target classes up by one to
+            allow for ignored/background classes with negative values. If ``shift_targets=True``, all class ids will be
+            incremented by 1, and any pixels with negative target indices will be assigned a target index of zero.
+            Default: ``False``.
```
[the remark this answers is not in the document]
coryMosaicML: I agree with this. It should be changed to this behavior now, and as a bonus you no longer need to specify a `shift_targets` option because targets are never shifted.

composer/callbacks/image_monitor.py
coryMosaicML: You're right, one-hots weren't being handled properly before. Fixed it so things should work correctly now.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+                 input_key: Union[str, int, Tuple[Callable, Callable], Any] = 0,
+                 target_key: Union[str, int, Tuple[Callable, Callable], Any] = 1):
+        self.interval = interval
+        self.mode = mode
+        self.num_images = num_images
+        self.input_key = input_key
+        self.target_key = target_key
+
+        try:
+            import wandb
```
ravi-mosaicml, line 81: Hmm, I'm not thrilled about having this callback only be compatible with wandb `wandb` here. That said, I realize that there isn't another way around this until we have the logger refactor.

I presume this PR will be ready before the logger refactor is in, and don't want that to hold this up. So, can you do the following?

1. Can you create a JIRA to track that ImageVisualizer will need to be refactored after the LoggerDesign refactor is implemented and assign it to @eracah?
2. Can you add a comment here -- e.g. `TODO(Evan): Generalize as part of the logger refactor`
3. Can you add a `.. note::` to the docstring that this callback only works with `wandb` logging for now
4. In the implementation, instead of calling `logger.data_batch()`, can you log the object directly to the wandb logger? E.g. `for dest in ensure_tuple(logger.destinations): if isinstance(dest, WandBLogger): dest.log_data(data)`? Mainly concerned about attempting to log a wandb.image to the file logger, and then having it either crash or blow up the logfile with a binary image.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Check that the output mode is valid
+        if self.mode.lower() not in ['input', 'segmentation']:
+            raise ValueError(f'Invalid mode: {mode}')
+
+        # Check that the interval timestring is parsable and convert into time object
+        try:
+            self.interval = Time.from_timestring(interval)
+        except ValueError as error:
+            raise ValueError(f'Invalid time string for parameter interval') from error
```
ravi-mosaicml, line 96: nit: no need to re-raise the exception; the stack trace should make it obvious that it was the parameter interval that was a bad timestring

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+        if self.interval.unit not in [TimeUnit.BATCH, TimeUnit.EPOCH]:
+            raise ValueError(f'Invalid time unit for parameter interval: '
+                             f'{self.interval.unit}')
+
+    def _log_inputs(self, state: State, logger: Logger, key: str):
+        inputs = state.batch_get_item(key=self.input_key)
+        if not isinstance(inputs, torch.Tensor):
+            raise NotImplementedError('Multiple input tensors not supported yet')
+        # Verify inputs is a valid shape for conversion to an image
+        if _check_for_image_format(inputs):
```
ravi-mosaicml, line 108: Should we raise an exception if the batch is not in image format?

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+            element is the target. Default: ``1``.
+    """
+
+    def __init__(self,
+                 interval: str = '100ba',
+                 mode: str = 'input',
+                 num_images: int = 8,
+                 input_key: Union[str, int, Tuple[Callable, Callable], Any] = 0,
+                 target_key: Union[str, int, Tuple[Callable, Callable], Any] = 1):
+        self.interval = interval
```
ravi-mosaicml, line 74: nit: I'd remove this line here; then, the `assert isinstance(self.interval, Time)` checks later won't be required.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+        if not isinstance(targets, torch.Tensor):
+            raise NotImplementedError('Multiple target tensors not supported yet')
+        if not isinstance(outputs, torch.Tensor):
+            raise NotImplementedError('Multiple output tensors not supported yet')
+
+        table = _make_segmentation_images(inputs, targets, outputs, self.num_images)
+        logger.data_batch({key: table})
+
+    def before_forward(self, state: State, logger: Logger):
+        assert isinstance(self.interval, Time)
```
ravi-mosaicml, line 128: nit: see comment above about removing this line.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+            element is the input. Default: ``0``.
+        target_key (str | int | Tuple[Callable, Callable] | Any, optional): A key that indexes to the target
+            from the batch. Can also be a pair of get and set functions, where the getter
+            is assumed to be first in the pair. The default is 1, which corresponds to any sequence, where the second
+            element is the target. Default: ``1``.
+    """
+
+    def __init__(self,
+                 interval: str = '100ba',
+                 mode: str = 'input',
```
ravi-mosaicml, line 70: Thoughts about removing this parameter, and instead conditioning on whether `target_key` is set? If `target_key = None`, then it's input only; otherwise, then it's segmentation?

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'input' and state.timestamp.get(self.interval.unit).value % self.interval.value == 0:
+            self._log_inputs(state, logger, 'Images/Train')
+
+    def eval_before_forward(self, state: State, logger: Logger):
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'input' and state.eval_timestamp.get(TimeUnit.BATCH).value == 0:
+            self._log_inputs(state, logger, 'Images/Eval')
+
+    def before_loss(self, state: State, logger: Logger):
```
ravi-mosaicml, line 137: To be parallel with evaluation, would it make sense to use the after forward event instead of before loss?

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+        assert isinstance(self.interval, Time)
+        if self.mode.lower() == 'segmentation' and state.eval_timestamp.get(TimeUnit.BATCH).value == 0:
+            self._log_segmented_inputs(state, logger, 'Images/Eval')
+
+
+def _make_input_images(inputs: torch.Tensor, num_images: int):
+    import wandb
+    if inputs.shape[0] < num_images:
+        num_images = inputs.shape[0]
+    images = inputs[0:num_images].data.cpu().permute(0, 2, 3, 1).numpy()
```
ravi-mosaicml, line 153: Do we need to have `channels_last` as a config option? Or is there a standard way to represent an image when training? Mainly thinking if there are cases where the permute isn't needed.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+    for image in images:
+        img = wandb.Image(image)
+        table.add_data(img)
+    return table
+
+
+def _make_segmentation_images(inputs: torch.Tensor, targets: torch.Tensor, outputs: torch.Tensor, num_images: int):
+    import wandb
+    if min([inputs.shape[0], targets.shape[0], outputs.shape[0]]) < num_images:
+        num_images = min([inputs.shape[0], targets.shape[0], outputs.shape[0]])
```
ravi-mosaicml, line 164: Shouldn't `inputs.shape[0] == targets.shape[0] == outputs.shape[0]` all the time?

ravi-mosaicml (approved):
Overall LGTM; code is very readable -- thanks!

Main comments re: wandb and cleaning up the API if possible. Nothing major.

composer/callbacks/image_monitor.py
coryMosaicML: I added the check, but decided to add it earlier in the code.

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+        >>> from composer.callbacks import ImageMonitor
+        >>> # constructing trainer object with this callback
+        >>> trainer = Trainer(
+        ...     model=model,
+        ...     train_dataloader=train_dataloader,
+        ...     eval_dataloader=eval_dataloader,
+        ...     optimizers=optimizer,
+        ...     max_duration="1ep",
+        ...     callbacks=[ImageMonitor()],
+        ... )
```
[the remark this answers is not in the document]
coryMosaicML: Should be fixed now

composer/callbacks/image_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Check that the output mode is valid
+        if self.mode.lower() not in ['input', 'segmentation']:
+            raise ValueError(f'Invalid mode: {mode}')
+
+        # Check that the interval timestring is parsable and convert into time object
+        try:
+            self.interval = Time.from_timestring(interval)
+        except ValueError as error:
+            raise ValueError(f'Invalid time string for parameter interval') from error
```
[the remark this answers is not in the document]
coryMosaicML: Done.

composer/callbacks/image_monitor.py
coryMosaicML: Changed this to `data_name`

composer/callbacks/image_monitor.py
coryMosaicML: Yes, was an accident. Fixed!

composer/callbacks/image_monitor.py
coryMosaicML: Ok, I added a check for numpy arrays and PIL images, so at least the appropriate error should get raised. I'm going to hold off on support for these types for a future PR.

composer/callbacks/image_visualizer.py
```diff
@@ -0,0 +1,189 @@
...
+            is assumed to be first in the pair.  The default is 0, which corresponds to any sequence, where the first
+            element is the input. Default: ``0``.
+        target_key (str | int | Tuple[Callable, Callable] | Any, optional): A key that indexes to the target
+            from the batch. Can also be a pair of get and set functions, where the getter
+            is assumed to be first in the pair. The default is 1, which corresponds to any sequence, where the second
+            element is the target. Default: ``1``.
+    """
+
+    def __init__(self,
+                 interval: str = '100ba',
```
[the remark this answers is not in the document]
coryMosaicML: Made this change, also included using an int with default batch units.

composer/callbacks/image_visualizer.py
coryMosaicML: Yeah I don't love it either. I'd like to get this in soon, probably before the logger refactor because I think we might want to use it to demo our segmentation work. 

2-4 should be done in the code now. Will make the JIRA too.

composer/callbacks/image_visualizer.py
coryMosaicML: Cleaned this up

composer/callbacks/image_visualizer.py
coryMosaicML: I decided not to, as training presumably will still work fine unless there's other unrelated bugs, just images won't be logged. I don't want to interrupt a training run if someone is just using the callback when they shouldn't. Happy to make the change though if you think we should raise an exception here.

composer/callbacks/image_visualizer.py
coryMosaicML: Fixed these.

composer/callbacks/image_visualizer.py
coryMosaicML: I think there might be an edge case here if someone does odd things either downsampling or upsampling the data as part of the forward pass. But, 99.9% of the time these will probably all be the same.

composer/callbacks/image_visualizer.py
coryMosaicML: Running before loss instead of after forward lets one see the results of target modifications that happen on the before loss event. A few algorithms (such as cutmix) modify the targets here, and those changes wouldn't be able to be visualized otherwise.

composer/callbacks/image_visualizer.py
coryMosaicML: In the future I anticipate there being other visualization modes (like object detection when we get there) which will also need targets, so I'd like to leave it as a parameter so that other modes can be added easily.

composer/callbacks/image_visualizer.py
coryMosaicML: You're right, sometimes images will be in the `channels_last` format. I added an option for it now.

PR #1267, opened 2022-07-07
ALiBi for BERT and ALiBi testing

alextrott16:
- Refactors the ALiBi implementation to embrace the (unavoidable) fact that there is not a one-size-fits-all approach to ALiBi model surgery. The refactor provides a way to define a registry of model surgery replacement functions associated with various transformer modules.
- Design choices from the original implementation are largely still intact, but this refactor greatly reduces the complexity of the parameter configuration the user needs to specify. This reduction is enabled by the registry-based approach.
- Expands ALiBi algorithm to support BERT models. This involves a modified implementation of the `forward` method in the `BertSelfAttention` class.
- Also supports RoBERTa models (this exploits the fact that HuggingFace RoBERTa models use copied but renamed implementations of the BERT classes).
- Adds unit testing for the ALiBi algorithm and the functional implementation, as well as tests covering warnings for unsupported model types as well as registering custom alibi surgery functions.
- Updated the README code examples and added a section covering supported models, which now includes an example demonstration of how to register alibi surgery functions.


composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+                        dtype=old_weight.dtype,
+                        layout=old_weight.layout,
+                        device=old_weight.device))
+        new_weight.requires_grad = False
+        setattr(pos_embedding_module, 'weight', new_weight)
+
+        log.info(f' Position embedding expanded to sequence length {max_sequence_length}, zeroed, and frozen')
+
+    except AttributeError:
+        # TODO: This should perhaps result in an exception rather than a warning
```
growlix, line 113: It does now! Hooray!

composer/algorithms/alibi/attention_surgery_functions/_bert.py
```diff
@@ -0,0 +1,157 @@
...
+from transformers.models.bert.modeling_bert import BertSelfAttention, BertEmbeddings
+from transformers.models.roberta.modeling_roberta import RobertaSelfAttention, RobertaEmbeddings
+
+@register_surgery_function_builder(BertEmbeddings, RobertaEmbeddings)
+def build_bert_embedding_converter(max_sequence_length: int):
+    """Builds a function to remove positional embeddings.
+    
+    The built function also expands `position_ids` buffer to support `max_sequence_length` tokens.
+    """
+    def convert_position_embeddings_and_ids(module: torch.nn.Module, module_index: Optional[int] = None):
```
growlix, line 25: Should `module`'s type specifically be `Union[BertEmbeddings, RobertaEmbeddings]`? Or can we rely on that based on how the registry is implemented?

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    else:  # e.g., for BERT
+        # Following https://github.com/ofirpress/attention_with_linear_biases/issues/5 (Implementation 1)
+        # In the causal case, you can exploit the fact that softmax is invariant to a uniform translation
+        # of the logits, which makes the math work out *after* applying causal masking. If no causal masking
+        # will be applied, it is necessary to construct the diagonal mask.
+        context_position = torch.arange(max_token_length)[:, None]
+        memory_position = torch.arange(max_token_length)[None, :]
+        relative_position = torch.abs(memory_position - context_position)
+        # [n_heads, max_token_length, max_token_length]
+        relative_position = relative_position.unsqueeze(0).expand(n_heads, -1, -1)
```
growlix, line 143: This is the kind of thing I would write a test for, but I'm also not very good at mentally reshaping tensors.

composer/algorithms/alibi/attention_surgery_functions/_bert.py
```diff
@@ -0,0 +1,157 @@
...
+        # positional_embedding = positional_embedding.to(dtype=query_layer.dtype)  # fp16 compatibility
+        #
+        # if self.position_embedding_type == "relative_key":
+        #     relative_position_scores = torch.einsum("bhld,lrd->bhlr", query_layer, positional_embedding)
+        #     attention_scores = attention_scores + relative_position_scores
+        # elif self.position_embedding_type == "relative_key_query":
+        #     relative_position_scores_query = torch.einsum("bhld,lrd->bhlr", query_layer, positional_embedding)
+        #     relative_position_scores_key = torch.einsum("bhrd,lrd->bhlr", key_layer, positional_embedding)
+        #     attention_scores = attention_scores + relative_position_scores_query + relative_position_scores_key
+        ########################################
```
growlix, line 123: Take a hike, position embeddings

composer/algorithms/alibi/attention_surgery_functions/_bert.py
```diff
@@ -0,0 +1,157 @@
...
+        module.register_buffer(
+            "position_ids", torch.arange(max_sequence_length).expand((1, -1)).to(module_device))
+        return module
+    return convert_position_embeddings_and_ids
+
+
+@register_surgery_function_builder(BertSelfAttention, RobertaSelfAttention)
+def build_bert_attention_converter(max_sequence_length: int):
+    """Builds a function that does model surgery to add ALiBi to Bert-style SelfAttention."""
+    def convert_attention(module: torch.nn.Module, module_index: Optional[int] = None):
```
growlix, line 40: Same as above

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+        log.info(f' Position embedding expanded to sequence length {max_sequence_length}, zeroed, and frozen')
+
+    except AttributeError:
+        # TODO: This should perhaps result in an exception rather than a warning
+        log.error(f'Unable to zero and freeze position embeddings. Module '
+                  f'{module} may lack attribute {position_embedding_attribute}, or position '
+                  f"embeddings may lack attribute 'weight'.")
+
+
+def register_alibi(module: torch.nn.Module, n_heads: int, max_token_length: int, causal: bool) -> torch.nn.Module:
```
growlix, line 119: I love it

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+# Initialize the empty registry that will be filled by using the `register_surgery_function_builder` decorator.
+SurgeryFunctionBuilder = TypeVar('SurgeryFunctionBuilder', bound=Callable[[int], ReplacementFunction])
+replacement_policy_mapping_builder: Dict[torch.nn.Module, SurgeryFunctionBuilder] = dict()
+
+def register_surgery_function_builder(*modules: torch.nn.Module) -> Callable[[SurgeryFunctionBuilder], SurgeryFunctionBuilder]:
+    """This decorator builds a registry that maps torch modules to their applicable SurgeryFunctionBuilder.
+
+    To accommodate the specifics of composer's model surgery, the register is used to create
```
growlix, line 21:
```suggestion
    To accommodate the specifics of composer's model surgery, the registry is used to create
```

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+            # Builds a function (`convert_attention`) that does model surgery any GPT2Attention modules in the model.
+            
+            def convert_attention(module: GPT2Attention, module_index: Optional[int] = None):
+                # Do surgery (change `module` or generate a new `module` instance to return)
+                # Note that this function can (and often should for ALiBi) depend on `max_sequence_length`
+                ...
+                return module
+            return convert_attention
+
+    In the above example, by decorating `build_gpt2_attention_converter` (which is an instances of a `SurgeryFunctionBuilder`
```
growlix, line 62:
```suggestion
    In the above example, by decorating `build_gpt2_attention_converter` (which is an instance of a `SurgeryFunctionBuilder`
```

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    function) with `@register_surgery_function_builder(GPT2Attention)`, the ALiBi algorithm will now apply model surgery to any
+    instances of `GPT2Attention` within the model, and will apply surgery on those instances using the `convert_attention` function
+    returned by `build_gpt2_attention_converter`.
+
+    Note that `convert_attention` follows the specific signature of a `ReplacementFunction`, which has specific arguments. However,
```
growlix, line 67: Is there model surgery documentation that can be linked to here?

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    often the model surgery functions used by ALiBi should depend on a `max_sequence_length` argument. Since this argument is not used
+    by a `ReplacementFunction`, we make use of these `SurgeryFunctionBuilder` functions so that the exact functions used in model surgery
+    can properly refer to the `max_sequence_length` argument, which is not provided until runtime (see `../alibi.py`).
+    """
```
growlix, line 71: Through documentation!

growlix (approved):
Excellent refactor. A few small comments but overall I think your approach is great!

composer/algorithms/alibi/attention_surgery_functions/utils.py
alextrott16: Hah, indeed it does!

composer/algorithms/alibi/attention_surgery_functions/_bert.py
alextrott16: We can rely on that based on how `alibi.py` will use the registry, but if someone wanted to hack into the registry to get a hold of these functions then they could pass arbitrary `torch.nn.Module` instances. I think this hypothetical person understands the risk of their misadventure, but, for the sake of good measure, I'll set the more restrictive type (which will also elucidate the way the code works).

composer/algorithms/alibi/attention_surgery_functions/_bert.py
alextrott16: So 2018

composer/algorithms/alibi/attention_surgery_functions/utils.py
alextrott16: Added some asserts to confirm sizing. Any sizing mistakes should also cause unit testing to fail.

composer/algorithms/alibi/attention_surgery_functions/_bert.py
alextrott16: I actually had to change the type signatures here to address issues with pyright. For example, `convert_attention` will still use `module: torch.nn.Module`, but I added asserts throughout to make sure that `module` is always an instance of one of the classes in the decorator.

composer/algorithms/alibi/attention_surgery_functions/utils.py
alextrott16: Added a reference to the `module_surgery.py` file

setup.py
```diff
@@ -126,6 +126,7 @@ def package_files(prefix: str, directory: str, extension: str):
     'mock-ssh-server==0.9.1',
     'cryptography==37.0.2',
     'pytest-httpserver>=1.0.4,<1.1',
+    'transformers>=4.11,<5',
```
ravi-mosaicml, line 129: Since transformers is a conditional dependency, all of its import usages need to be conditional. See https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#51-external-dependencies

ravi-mosaicml (commented):
See comment re:conditional dependencies.

setup.py
alextrott16: @ravi-mosaicml I've removed this from `setup.py` and caught transformers-related import errors as suggested above. Can you take a quick look at `composer/algorithms/alibi/attention_surgery_functions/__init__.py` and `composer/algorithms/alibi/alibi.py` (lines 100-117) to let me know if there are issues?

moinnadeem (commented):
I did a first pass and it looks good to me so far, but why do we override the forward method instead of replacing all instances of a particular attention mechanism? (cc @hanlint -- thoughts on the design?)

composer/algorithms/alibi/alibi.py
```diff
@@ -26,15 +23,10 @@
...
 def apply_alibi(
     model: torch.nn.Module,
-    heads_per_layer: int,
-    max_sequence_length: int,
-    position_embedding_attribute: str,
-    attention_module: Type[torch.nn.Module],
-    attr_to_replace: str,
-    alibi_attention: Callable,
-    mask_replacement_function: Optional[Callable[[torch.nn.Module, int], torch.nn.Module]] = None,
+    max_sequence_length: int = 8192,
```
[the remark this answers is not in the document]
alextrott16: 8192 was the default when I picked up this code, but I'm glad you bring this up because I'm realizing that we should lower that substantially. The size of the buffers become substantial at that size. If someone wants to use 8192 (or larger), they should do so by choice rather than by default.

Thinking more, perhaps this shouldn't have a default value. `max_sequence_length` changes the maximum length the model can process, and would possibly even shorten it if set too low. As such, if we give default that is too small, users may accidentally lessen the maximum length. On the flip side, if we give a default that is too large, it might be needlessly wasteful.

I'll make this a required argument and make sure the docstring is appropriately clear!

composer/algorithms/alibi/attention_surgery_functions/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Import files that add functions to the `replacement_policy_mapping_builder` registry in order to actually
+# register those functions.
+try:
+    from composer.algorithms.alibi.attention_surgery_functions import _bert, _gpt2  # pyright: reportUnusedImport=none
```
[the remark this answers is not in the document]
alextrott16: Populating the registry requires importing the source files that add to it (i.e. that decorate the surgery functions with `@register_surgery_function_builder`). To pyright, these appear to be unused imports, but they do serve a purpose.

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,179 @@
...
+    This is intended to be used specifically for "removing" positional embeddings.
+    """
+    try:
+        pos_embedding_module = attrgetter(position_embedding_attribute)(module)
+        old_weight = getattr(pos_embedding_module, 'weight')
+        if not isinstance(old_weight, torch.nn.Parameter):
+            raise TypeError(f'Module {module._get_name()}, position embedding {position_embedding_attribute}, '
+                            f"'weight' attribute must be of type torch.nn.Module")
+        new_weight = torch.nn.Parameter(
+            torch.zeros((max_sequence_length, old_weight.shape[1]),
```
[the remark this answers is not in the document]
alextrott16: Unless the position embeddings are 0 (or otherwise uniform at each position) each token's embedding will still be position-dependent, which we don't want.

Zeroing and setting `required_grad = False` is the simplest way (I think) to effectively make the position embedding module a no-op.

alextrott16:
> I did a first pass and it looks good to me so far, but why do we override the forward method instead of replacing all instances of a particular attention mechanism? (cc @hanlint -- thoughts on the design?)

@moinnadeem Great question. This was my plan early on but I switched gears. There are a few main reasons.
1. Replacing modules with new module instances would require additional boilerplate code to define the replacement class, which is very likely going to be only slightly different than the class type it replaces during surgery.
2. It is not necessarily trivial to know what parameter values to pass when instantiating the replacement class. By modifying the module instance rather than replacing it you avoid the problem of having to correctly initializing each new instance.
3. Overriding the module function (rather than replacing the module instance itself) means that module surgery leaves the module's parameter weights intact. This is desirable, e.g., if the pre-surgery module's parameters are already trained -- for instance, if the module weights had been loaded in from some checkpoint.

hanlint:
Nice work! I like the way you've structured the registries, so that new models can self-register themselves. I would like to find a way to avoid creating the alibi model via monkey-patching the forward pass. I'll reach out offline to maybe brainstorm if any alternate approaches would work.

composer/algorithms/alibi/alibi.py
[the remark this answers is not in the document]
moinnadeem: Sounds good!

composer/algorithms/alibi/attention_surgery_functions/_bert.py
```diff
@@ -0,0 +1,163 @@
...
+                                max_token_length=max_sequence_length,
+                                causal=False)
+        setattr(module, 'forward', MethodType(forward, module))
+
+        return module
+
+    return convert_attention
+
+
+# pyright: reportGeneralTypeIssues = none
```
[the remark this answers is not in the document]
moinnadeem: Is there a reason we decided to override the forward function rather than replace the class instances? (cc @hanlint , I think he wrote the same comment and reached out offline)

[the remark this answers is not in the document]
alextrott16: A few reasons. They're listed out in a separate comment below. Copied here:

>> I did a first pass and it looks good to me so far, but why do we override the forward method instead of replacing all instances of a particular attention mechanism? (cc @hanlint -- thoughts on the design?)

> @moinnadeem Great question. This was my plan early on but I switched gears. There are a few main reasons.
> 
> Replacing modules with new module instances would require additional boilerplate code to define the replacement class, which is very likely going to be only slightly different than the class type it replaces during surgery.
> It is not necessarily trivial to know what parameter values to pass when instantiating the replacement class. By modifying the module instance rather than replacing it you avoid the problem of having to correctly initializing each new instance.
> Overriding the module function (rather than replacing the module instance itself) means that module surgery leaves the module's parameter weights intact. This is desirable, e.g., if the pre-surgery module's parameters are already trained -- for instance, if the module weights had been loaded in from some checkpoint.

composer/algorithms/alibi/attention_surgery_functions/utils.py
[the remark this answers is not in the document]
moinnadeem: Makes sense, good call! This is because we don't want to override the code that sums the position embeddings, right?

moinnadeem (commented):
thinking about the forward function monkey patching:

> Replacing modules with new module instances would require additional boilerplate code to define the replacement class, which is very likely going to be only slightly different than the class type it replaces during surgery.

Hmmm I am amenable to this argument

> It is not necessarily trivial to know what parameter values to pass when instantiating the replacement class. By modifying the module instance rather than replacing it you avoid the problem of having to correctly initializing each new instance.

I usually copy over the old values when instantiating the new one, and this makes it more clear imo when you add new specific parameters (such as `max_sequence_length` to the `__init__`). 

> Overriding the module function (rather than replacing the module instance itself) means that module surgery leaves the module's parameter weights intact. This is desirable, e.g., if the pre-surgery module's parameters are already trained -- for instance, if the module weights had been loaded in from some checkpoint.

However, given that we now changed the model architecture and the way it works, is this a priority? Aren't the original weights useless? I almost view that one as a feature.

Overall, I'm pretty ambivalent here -- @hanlint do you have any thoughts? I think our current surgery implementation is a bit half-baked anyways, so don't have strong opinions. (cc @dblalock you might have some opinions?)

moinnadeem (approved):
From a science side, this is an LGTM to me! I have some questions about monkey patching versus replacing the entire instance, but Alex raises some good questions that it feels like it involves a bigger discussion. I'll defer to @hanlint and @dblalock on monkey patching surgery versus replacing instances.

moinnadeem:
The one question I have: the rest of Composer does surgery via replacing class instances, is it weird if ALiBi does surgery via monkey patching instead? I wonder if consistency is important here.

dblalock:
Some attempts at useful comments re: module replacement vs method replacement:

- The main reason we've done module replacement in the past is because we have standalone `nn.Module`s that implement methods. E.g., BlurPool2d, SqueezeExcite2d, etc. This is so that users can use these methods directly without having to go through the Trainer + an Algorithm. Once we already have such a module, swapping it in via surgery involves the least duplicate code and (arguably) the most debugability.
- I don't see any reason to not have an Alibi attention nn.Module; supporting speedups without the Trainer is still a design goal AFAIK.
- Just looking at the surgery in isolation, replacing one method (and maybe deleting one attribute) is simpler and will require less duplicate code than replacing the whole module.
- I spent a few minutes thinking about composition and I _suspect_ that it's about equally difficult in both cases.

composer/algorithms/alibi/attention_surgery_functions/utils.py
```diff
@@ -0,0 +1,184 @@
...
+                        dtype=old_weight.dtype,
+                        layout=old_weight.layout,
+                        device=old_weight.device))
+        new_weight.requires_grad = False
+        setattr(pos_embedding_module, 'weight', new_weight)
```
dblalock, line 119: Do we need to overwrite this with an empty tensor? Can we just delete the attribute?

alextrott16: I think that would cause problems because the attribute is referenced in the `forward` method of the parent module. We could replace the `forward` method so that this attribute isn't referenced, but that would add boilerplate.

dblalock: wait, does the parent forward() get called? I'm probably just not understanding the code fully.

alextrott16: The parent `forward` gets called when the full model is executed. Maybe one point of confusion is that, in this implementation, none of the modules are swapped out with new modules (which is I think the typical way surgery works). Instead, the modules are modified. 
Alternatively, I could have misunderstood your original comment and created additional confusion with my response.

dblalock: Possible point of confusion: is `parent` referring to the owning module or a parent class?

I was understanding your statement as referring to the latter. Since:

1. for every model we modify, the `forward()` gets completely replaced with the `forward()` function in alibi, and
2. This function doesn't reference `super().forward()`,

I expect no forward() will ever try to access the parameter in question.

But if the owning module accesses this parameter within its child, then that makes total sense.

alextrott16: I think the intended interpretation is that "the owning module accesses this parameter within its child." ... I think.

Note, though, that in the course of alibi surgery on a Bert model, we do surgery on the `BertEmbedding` module, but we do not replace the `forward` function when modifying that module type. Instead, we just zero/freeze the weights of its `position_embedding` attribute. `BertEmbedding.forward()` (which references its `self.position_embedding`) remains intact. (see `composer/algorithms/alibi/attention_surgery_functions/_bert.py`)

hanlint (commented):
LGTM, a few comments:

* I'm OK with the monkeypatching for now. Eventually we should replace this with `torch.fx` magic (cc: @dskhudia for comment there).
* We should be able to remove the `_check_builder_mapping` function and the `try.. except` in `attention_surgery_functions/__init__.py`. Instead, do something like:
```
def apply_alibi(..):
    try:
        import attention_surgery_functions as asf
    except ImportError as e and 'transformers' in e:
        policy_registry = {}
    else:
         policy_registry = asf.policy_registry
```
* We can also potentially simplify the registry / builder functions by having the `convert` methods also take in a sequence length argument. Then, during `apply_alibi`, do something like:
```
policies = {}
for module, policy in policy_registry:
    policies[module] = lambda m: nn.Module, index: int functools.partial(policy, sequence_length=16)
```

and a nit: could you shorten the naming of some of the variables? e.g. `replacement_policy_mapping_builder` -> `policy_registry` (replacement is assumed, and `mapping` is expressed by registry, perhaps builder can be removed?).

composer/algorithms/alibi/README.md
```diff
@@ -94,7 +81,7 @@ trainer.fit()
 
 ALiBi is implemented as follows. On `Event.INIT`:
 1. The model's position embeddings are expanded to accommodate sequences of up to length `max_sequence_length` and then "bypassed" by setting them to zero and freezing them.
-2. The attribute that computes the self-attention (`attr_to_replace`) in the model's self-attention modules (`attention_module_name`) is replaced with an ALiBi-enabled self-attention method (`alibi_attention`) using graph surgery.
+2. The attribute that computes the self-attention in the model's self-attention modules is replaced with an ALiBi-enabled self-attention method using graph surgery. Our implementation builds a registry that maps module types to their graph surgery functions. **Note:** you may need to add to the registry if your model's self-attention and embedding modules are not included in the registry.
```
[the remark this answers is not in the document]
alextrott16: Yep, added that section, which describes current support and provides an example pattern for registering your own implementation. Also added a unit test to confirm that at-runtime registry works.

composer/algorithms/alibi/alibi.py
```diff
@@ -26,15 +23,10 @@
...
-    heads_per_layer: int,
     max_sequence_length: int,
-    position_embedding_attribute: str,
-    attention_module: Type[torch.nn.Module],
-    attr_to_replace: str,
-    alibi_attention: Callable,
-    mask_replacement_function: Optional[Callable[[torch.nn.Module, int], torch.nn.Module]] = None,
     optimizers: Optional[Union[Optimizer, Sequence[Optimizer]]] = None,
-) -> None:
+    output_replaced_pairs: bool = False,
```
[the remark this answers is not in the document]
alextrott16: Removed this. Replaced it with a warning that tells the user that Alibi had no effect. Also added tests to confirm that the warning triggers correctly.

composer/algorithms/alibi/attention_surgery_functions/_bert.py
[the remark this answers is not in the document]
alextrott16: Done!

PR #1268, opened 2022-07-08
Add `scale_warmup` argument to schedulers

hanlint:
Based on request from @codestar12 , this PR adds the option to also scale the warmup period for our schedulers with the `scale_schedule_ratio`. 

For example:
```
    from composer.optim.scheduler import MultiStepWithWarmupScheduler

    scheduler = MultiStepWithWarmupScheduler(
        milestones=["10ep", "20ep"],
        t_warmup="4ep",
        scale_warmup=True,
    )
```

With `scale_schedule_ratio=0.5`, this scheduler will warmup for 2 epochs, then step at 5 and 10 epochs. `scale_warmup` defaults to `False` to preserve the current behavior.

During implementation, I observed an unintuitive default behavior with our warmup. Suppose we have `max_duration=100ba`. If we define a scheduler with a warmup in batches (e.g. `t_warmup=10ba`), and apply an ssr, the warmup period will not be scaled, per our default behavior. However, if the warmup was defined in duration (e.g. `t_warmup=0.1dur`), and apply an ssr, then the warmup period will always be scaled (as we scale `max_duration` in the trainer beforehand). This PR respects that behavior, but perhaps we should attempt a fix separately.


ravi-mosaicml:
CC @jbloxham and @jfrankle since they originally wrote the schedulers

composer/optim/scheduler.py
```diff
@@ -537,19 +537,25 @@ class MultiStepWithWarmupScheduler(ComposerScheduler):
         rate multiplier until the warmup has completed.
 
     .. warning::
```
ravi-mosaicml, line 539: nit: Thoughts about making it a `.. note::`? (Same comments for other schedulers) IMO warning feels a bit strong?

codestar12: It was warning before these changes right? Before the option to scale warmup it was very important to know how SSR changed the schedule for correctness. I feel like that is still true.

hanlint: Correct, it was warning before, so agree with @codestar12 to keep this as warning.

PR #1271, opened 2022-07-11
Add packaging to `meta.yaml`; add `py-cpuinfo` max version

ravi-mosaicml:
* Follow up on #1612 to add packaging to the `meta.yaml`.
* Constrain `py-cpuinfo` to a max version of 9.0.0


mvpatel2000 (approved):
Good catch! Didn't realize there was another spot besides setup

PR #1272, opened 2022-07-11
Fix Flaky Tests

ravi-mosaicml:
* Tests involving the file logger require writing and flushing to disk, which can be slow on Jenkins and lead to sporadic timeouts in unrelated tests. Increasing the timeout by default to 30 seconds for all tests involving the file logger.
* Fixed `TestThresholdStopper` to use smaller (and fewer) batches, since the metrics were manually set

Closes https://mosaicml.atlassian.net/browse/CO-664

### DIFF
--- a/composer/models/resnet/README.md
+++ b/composer/models/resnet/README.md
@@ -13,7 +13,7 @@ from composer.models import composer_resnet
 model = composer_resnet(
     model_name="resnet50",
     num_classes=1000,
-    pretrained=False
+    weights=None
 )
 ```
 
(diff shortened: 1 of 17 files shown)


bandish-shah (approved):
Approving to unblock, looks like a timeout in a Jenkins test not sure if that's related to this or another timeout issue.

eracah (approved):

hanlint:
The offending test (def test_threshold_stopper_eval) can probably be shortened significantly by monkeypatching `state.current_metrics` to return a 100.0

ravi-mosaicml:
> The offending test (def test_threshold_stopper_eval) can probably be shortened significantly by monkeypatching `state.current_metrics` to return a 100.0

This test doesn't require training convergence to pass; the `MetricSetterCallback` patches `state.current_metrics` according to the `metric_sequence` fixture. I changed the data configuration so training is effectively a no-op.

PR #1273, opened 2022-07-11
Hotfix CO-658

milocress:
Fixes https://mosaicml.atlassian.net/browse/CO-658

In some cases, `dist.broadcast_object_list` was called before `dist.init`, resulting in an error. We factor out this call by downloading the compression file to each rank.


ravi-mosaicml:
https://github.com/mosaicml/composer/pull/1272 should fix the flaky tests ... that seems to be failing on every PR now.

ravi-mosaicml (approved):
LGTM -- Thanks! I like how it downloads to a well known location only on local rank zero; that should minimize duplicate downloads for multi-rank training.

abhi-mosaic (changes requested):
Adding a quick block here while I debug the multi-gpu tests, I'm seeing it pass on `dev` but not this branch

abhi-mosaic:
Ok I figured out the bug, but it's gonna be a bit tricky to resolve, will defer to you @milocress and review after it's fixed :)

* the compression file is downloaded by local rank0 and waited upon by other ranks
* but in the current code, rank0 also deletes the file right away with `os.remove`
* so we need to wait to delete this file until all devices have read it
* or we need to download it to a tempfile somewhere that is **not named `compression.txt`** because the local cache should not be compressed.

To verify if the is working, all cases for this test should pass:
```
make test-dist WORLD_SIZE=2 EXTRA_ARGS="-s -vv -m daily tests/datasets/test_streaming.py::test_dataloader_multi_device"
```

composer/datasets/streaming/download.py
```diff
@@ -116,7 +116,8 @@ def dispatch_download(remote: Optional[str], local: str):
         else:
             raise NotImplementedError
         os.rename(tempfile, local_decompressed)
-        os.remove(local)
+        if local != local_decompressed:
+            os.remove(local)
```
abhi-mosaic, line 120: Is this still needed?

composer/datasets/streaming/dataset.py
```diff
@@ -139,26 +139,20 @@ def __init__(self,
         self.batch_size = batch_size
 
         self.compression_scheme = None
-        if remote is not None and dist.get_global_rank() == 0:
+        if remote is not None:
             try:
-                compression_local = self._download_file(get_compression_scheme_basename())
+                compression_local = self._download_file(get_compression_scheme_basename(),
+                                                        wait=False,
```
abhi-mosaic, line 145: I think this should be a `wait=(dist.get_local_rank() != 0)`, right?

composer/datasets/streaming/download.py
milocress: I believe so, otherwise the local file might be deleted as soon as it's downloaded.

abhi-mosaic: But I think this is inside the decompression block, so in this block the local file will never be the same as `local_decompressed` right?

milocress: oh good catch

abhi-mosaic (approved):
Looks good to me! Seeing the test cases all pass now, and thank you for reducing the network calls to `local_rank_zero` only :)

PR #1274, opened 2022-07-12
Adds S3ObjectStore import to util __init__.py 

codestar12:
Adds S3ObjectStore import to util __init__.py so that S3ObjectStore closure is constant with LibCloudObjectStore

### DIFF
--- a/composer/utils/__init__.py
+++ b/composer/utils/__init__.py
@@ -10,7 +10,7 @@ from composer.utils.file_helpers import (create_symlink_file, ensure_folder_has_
                                          get_file, is_tar)
 from composer.utils.import_helpers import MissingConditionalImportError, import_object
 from composer.utils.iter_helpers import IteratorFileStream, ensure_tuple, map_collection
-from composer.utils.object_store import LibcloudObjectStore, ObjectStore, ObjectStoreTransientError, S3ObjectStore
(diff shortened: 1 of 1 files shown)


composer/utils/__init__.py
```diff
@@ -23,6 +23,7 @@
     'ObjectStore',
     'ObjectStoreTransientError',
     'LibcloudObjectStore',
+    'S3ObjectStore'
     'MissingConditionalImportError',
```
hanlint, line 27: Could you add `SFTPObjectStore` as well? 🙏

hanlint (approved):

ravi-mosaicml (approved):
LGTM, thanks!

composer/utils/__init__.py
```diff
@@ -23,6 +23,7 @@
     'ObjectStore',
     'ObjectStoreTransientError',
     'LibcloudObjectStore',
+    'S3ObjectStore'
```
hanlint, line 26:
```suggestion
    'S3ObjectStore',
```

composer/utils/__init__.py
codestar12: done

composer/utils/__init__.py
codestar12: nice catch. Fixed when adding SFTPObjectStore

PR #1278, opened 2022-07-13
Allow object name prefix for S3ObjectStore

abhi-mosaic:
This makes it easy to specify a folder-like prefix for objects that are stored in S3. 

This will be pretty important for users who share a bucket for artifacts like checkpoints, but use folders like `my-name/my-project/` within to separate work. 

It will also be important for this to land before connecting ObjectStore with StreamingDatasets (#1248) because the basic unit being referenced is a "folder" that is rarely an entire bucket.

### DIFF
--- a/composer/utils/object_store/s3_object_store.py
+++ b/composer/utils/object_store/s3_object_store.py
@@ -83,8 +83,13 @@ class S3ObjectStore(ObjectStore):
             from botocore.config import Config
         except ImportError as e:
             raise MissingConditionalImportError('s3', 'boto3') from e
-        self.bucket = bucket
-        self.prefix = prefix
+
+        # Format paths
+        self.bucket = bucket.strip('/')
+        self.prefix = prefix.strip('/')
+        if self.prefix:
+            self.prefix += '/'
+
         if client_config is None:
             client_config = {}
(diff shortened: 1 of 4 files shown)


composer/utils/object_store/s3_object_store.py
```diff
@@ -99,7 +102,7 @@ def __init__(
         self.transfer_config = TransferConfig(**transfer_config)
 
     def get_uri(self, object_name: str) -> str:
-        return f's3://{self.bucket}/{object_name}'
+        return os.path.join('s3://', self.bucket, self.prefix, object_name)
```
ravi-mosaicml, line 105: Since object store paths must use forward slashes, would prefer to use manual string building for the URI rather than `os.path.join`, which could result in a backwards slash (`\`) on windows. If you want to use a helper, take a look at https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit

ravi-mosaicml (approved):
LGTM, but see comment.

ravi-mosaicml:
Forgot to mention but can you add a quick test case with the `prefix` argument to ensure that the URL formats properly?

PR #1280, opened 2022-07-13
Fix S3 remote paths for StreamingDataset download

abhi-mosaic:
Forward slash must be stripped, otherwise the `S3ObjectStore` fails to download files.

Is this bug intentional by the way? I tried doing some local tests and it seems like all our ObjectStore implementations don't do any stripping of the `object_name`.

Will follow up in Slack, but this is also a good reminder that when modifying `StreamingDataset` code, it's always necessary to run the remote tests which actually do a full E2E download from S3 (and can be run even from your laptop!). This is how I noticed something was failing:

```
pytest -s -m remote "tests/datasets/test_streaming_remote.py::test_streaming_remote_dataset[val-cifar10]"            
```


milocress (approved):
Good catch!

PR #1281, opened 2022-07-14
AMD integration & yamls

florescl:
Did some profiling on amd gpus with rocm containers and these specific yamls.


composer/trainer/devices/device_hparams_registry.py
```diff
@@ -18,4 +18,5 @@
 device_registry: Dict[str, Union[Type[Device], Type[hp.Hparams]]] = {
     'gpu': DeviceGPU,
     'cpu': DeviceCPU,
+    'amd_gpu': DeviceGPU,
```
abhi-mosaic, line 21: Is this necessary? If we're mapping directly to `DeviceGPU` it seems to imply we might be able to just use 'gpu' for both NVIDIA and AMD :D

florescl: right, but i want to still have this option to clearly distinguish runs.

composer/yamls/hwbenchmark/bert-stream.yaml
```diff
@@ -0,0 +1,75 @@
+train_dataset:
```
hanlint, line 1: We can probably leave the YAMLs out of the PR.. does it "just work" like this? If so, maybe we don't even need a PR then for integration? Maybe a notebook tutorial suffices?

florescl: I can add a notebook later but I'd like to have this folder in the yamls for all the hw benchmarking (not just amd) i'm doing.

composer/trainer/devices/device_hparams_registry.py
ravi-mosaicml: Agree with @abhi-mosaic; this is a bit confusing / redundant...

Curious what is the use case to distinguish? is it for logging? If so, then can we have the trainer add a log statement for the cuda device id or the like?

ravi-mosaicml (commented):
Thanks for testing everything out on AMD GPUs -- I'm guessing there's a lot more that went into that than the size of this PR diff would indicate!

Overall I agree with the comments already made. Happy to add in more logging statements if that would be helpful, but would prefer not to have duplicate entries in the device registry.

composer/trainer/devices/device_hparams_registry.py
bandish-shah: +1 on removing this and just using the `gpu` device if possible, we can either log or find another way to capture the device specific information.

How does this work?  The `DeviceGPU` class defined in the `device_gpu.py` module is a CUDA specific implementation.  My understanding is that AMD uses a different backend called ROCm so wouldn't this need to change?

composer/yamls/hwbenchmark/bert.yaml
```diff
@@ -0,0 +1,83 @@
+# Use a bert-base model, initialized from scratch
```
bandish-shah, line 1: This YAML looks almost identical to https://github.com/mosaicml/composer/blob/dev/composer/yamls/models/bert-base.yaml.  Let's not duplicate.

hanlint:
Discussed offline - closing this for now, and moving the benchmark YAMLs elsewhere, consistent with our move towards removing most yamls from the composer repo.

bandish-shah (changes requested):
It's great that AMD works with Composer with basically no modification.

Please see my question regarding how this works with the existing CUDA-centric implementation of the `DeviceGPU` class.  

Since the YAML's in this PR very closely resemble other existing YAML's, I'm in favor of excluding them from this PR for now.  Let's figure out offline where we can stash these.

PR #1282, opened 2022-07-14
Fix Tensorboard with ObjectStore quadratic complexity issue and switch from flush_interval -> flush_secs

eracah:
CO-690

This PR:
- [x] creates and logs to a new file for every flush
- [x] re-initializes a new SummaryWriter for each flush
- [x] names the artifact based on flush count to prevent overwriting on S3 or other objectstore
- [x] changes flushing interval specification to by time instead of number of batches (`flush_interval` -> `flush_secs`)

* Logging to a new file for every flush solves the quadratic complexity issue, where the same file was sent to S3, while it gradually increased in size.
* Re-initializing a new SummaryWriter fixed various weird issues (dropping of datapoints) when changing a log file name midway through the life of a SummaryWriter object's life.
* Append flush_count to artifact_name, so we create a unique artifact after every flush and don't overwrite the previous one.
* flushing interval changed to seconds because it is more intuitive, easier to tune, and holds the same meaning regardless of workload.

[wandb report](https://wandb.ai/mosaic-ml/evan-tb-test/reports/Tensorboard-Logger-Fix---VmlldzoyMzEzNzIx?accessToken=r57f3nfquc4hv5qp28h413hc8rmr33zof2aznuh47cm2u6l9160o8gojqx827li5) showing scaling of Tensorboard logging over the course of training (linear increase in bytes sent, and no increase in thropughput is good!)

### DIFF
--- a/composer/loggers/tensorboard_logger.py
+++ b/composer/loggers/tensorboard_logger.py
@@ -143,6 +143,7 @@ class TensorboardLogger(LoggerDestination):
             file_path=file_path,
             overwrite=True)
 
+        self.flush_count += 1
         # Close writer and reinitialize it to ensure no strange issues with dropping
         # of points.
         self.writer.close()
@@ -152,5 +153,3 @@ class TensorboardLogger(LoggerDestination):
         # Give event_file a unique name to avoid appending to the same file on every flush.
         self.writer.file_writer.event_writer._file_name = self.event_file_base_file_path + f'-{self.flush_count}'
         self.writer.file_writer.event_writer._async_writer._writer._writer.filename = self.event_file_base_file_path + f'-{self.flush_count}'
-
-        self.flush_count += 1
--- a/composer/loggers/tensorboard_logger.py
+++ b/composer/loggers/tensorboard_logger.py
@@ -12,7 +12,6 @@ from composer.loggers.logger import Logger, LogLevel
 from composer.loggers.logger_destination import LoggerDestination
 from composer.utils import dist
 from composer.utils.import_helpers import MissingConditionalImportError
-import copy
 
 __all__ = ['TensorboardLogger']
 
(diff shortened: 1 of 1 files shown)


eracah:
Open question @abhi-mosaic, do we make a new file and corresponding artifact after EVERY flush or every few flushes, so we don't overload the metadata server OR do we just recommend a pretty high `flush_secs` interval?

composer/loggers/tensorboard_logger.py
```diff
@@ -31,19 +32,19 @@ class TensorboardLogger(LoggerDestination):
             will be saved. This is also the value that should be specified when starting
             a tensorboard server. e.g. `tensorboard --logdir={log_dir}`. If not specified
             `./tensorboard_logs` will be used.
-        flush_interval (int, optional): How frequently by batch to flush the log to a file.
-            For example, a flush interval of 10 means the log will be flushed to a file
-            every 10 batches. The logs will also be automatically flushed at the start and
-            end of every evaluation phase (`EVENT.EVAL_START` and `EVENT.EVAL_END` ),
-            the end of every epoch (`EVENT.EPOCH_END`), and the end of training
-            (`EVENT.FIT_END`). Default: ``100``.
+        flush_secs (int, optional): How frequently in seconds to flush the log to a file
```
ravi-mosaicml, line 35: discussed offline, but could we put the change to make the flush interval in seconds to a seperate pr? would like to cherry-pick the fix into 0.8.1, but changing the interval to seconds would be a breaking change.

ravi-mosaicml (commented):
Thanks for getting this fix out! Will review later tonight or first thing tmrw am.

eracah:
new PR is https://github.com/mosaicml/composer/pull/1283

composer/loggers/tensorboard_logger.py
eracah: https://github.com/mosaicml/composer/pull/1283

PR #1283, opened 2022-07-14
Fix TB Logger + ObjectStore quadratic complexity issue by doing 1 file per flush

eracah:
CO-690

This PR:
- [x] creates and logs to a new file for every flush by calling `writer.close()`
- [x] names the artifact based on the tfevents file to prevent overwriting on S3 or other objectstore

* Logging to a new file for every flush solves the quadratic complexity issue, where the same file was sent to S3, while it gradually increased in size.

[wandb report](https://wandb.ai/mosaic-ml/evan-tb-test/reports/Tensorboard-Fix-Plots--VmlldzoyMzE0MTEy?accessToken=bnqi0jaawy5xubsfcck9mz1izabzdyom29o34sdp2nn9ovy5vb8o2naq3aeo0rcb)

### DIFF
--- a/composer/loggers/tensorboard_logger.py
+++ b/composer/loggers/tensorboard_logger.py
@@ -3,9 +3,9 @@
 
 """Log to `Tensorboard <https://www.tensorflow.org/tensorboard/>`_."""
 
-import time
 from pathlib import Path
 from typing import Any, Dict, Optional
+import uuid
 
 from composer.core.state import State
 from composer.loggers.logger import Logger, LogLevel
@@ -55,8 +55,6 @@ class TensorboardLogger(LoggerDestination):
         self.flush_interval = flush_interval
         self.rank_zero_only = rank_zero_only
         self.writer: Optional[SummaryWriter] = None
-        self.flush_count = 0
-        self.event_file_base_file_path: Optional[str] = None
         self.run_name: Optional[str] = None
 
     def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
@@ -80,7 +78,7 @@ class TensorboardLogger(LoggerDestination):
     def init(self, state: State, logger: Logger) -> None:
         self.run_name = state.run_name
 
-        # We set the log_dir to a constant, so all runs can be co-located together.
+        # We fix the log_dir, so all runs are co-located.
         if self.log_dir is None:
             self.log_dir = 'tensorboard_logs'
 
@@ -98,13 +96,6 @@ class TensorboardLogger(LoggerDestination):
         # To disable automatic flushing we set flushing to once a year ;)
         flush_secs = 365 * 3600 * 24
         self.writer = SummaryWriter(log_dir=summary_writer_log_dir, flush_secs=flush_secs)
-
-        if self.event_file_base_file_path is None:
-            self.event_file_base_file_path = self.writer.file_writer.event_writer._file_name
-
-        # Give event_file a unique name to avoid appending to the same file on every flush.
-        self.writer.file_writer.event_writer._file_name = self.event_file_base_file_path + f'-{self.flush_count}'
-        self.writer.file_writer.event_writer._async_writer._writer._writer.filename = self.writer.file_writer.event_writer._file_name
 
     def batch_end(self, state: State, logger: Logger) -> None:
         if int(state.timestamp.batch) % self.flush_interval == 0:
@@ -124,35 +115,22 @@ class TensorboardLogger(LoggerDestination):
         if self.rank_zero_only and dist.get_global_rank() != 0:
             return
 
+        # Skip if no writes occurred since last flush.
+        if not self.writer.file_writer:
+            return
+
+        self.writer.flush()
         assert self.writer.file_writer is not None
         file_path = self.writer.file_writer.event_writer._file_name
 
-        # If no writes have happened since the last flush, then file_path won't exist, so
-        # we should skip doing flushing and skip filing the artifact since it will error
-        # out anyway (given that the file_path doesn't exist).
-        if not Path(file_path).exists():
-            return
-
-        assert self.writer is not None
-        self.writer.flush()
-
-
         logger.file_artifact(
             LogLevel.FIT,
-            # For a file to be readable by Tensorboard, it must start with
-            # 'events.out.tfevents'. Child directory is named after run_name, so the logs
-            # are named properly in the Tensorboard GUI.
             artifact_name=(
                 'tensorboard_logs/{run_name}/events.out.tfevents-{run_name}-{rank}'
                 # Append flush_count, so artifact has unique name.
-                + f'-{self.flush_count}'),
+                + f'-{uuid.uuid4()}'),
             file_path=file_path,
             overwrite=True)
         
-        # Close writer and reinitialize it with a new log file path. This ensures that 
-        # we have one file per flush, which means `file_artifact` is always copying 
-        # data points that have never been copied before. Also this ensures no strange 
-        # issues with dropping of points when setting a new log file path.
+        # Close writer, which creates new log file.
         self.writer.close()
-        self.flush_count += 1
-        self._initialize_summary_writer()
--- a/composer/loggers/tensorboard_logger.py
+++ b/composer/loggers/tensorboard_logger.py
@@ -3,6 +3,7 @@
 
 """Log to `Tensorboard <https://www.tensorflow.org/tensorboard/>`_."""
 
+import time
 from pathlib import Path
 from typing import Any, Dict, Optional
 
@@ -55,6 +56,9 @@ class TensorboardLogger(LoggerDestination):
         self.flush_interval = flush_interval
         self.rank_zero_only = rank_zero_only
         self.writer: Optional[SummaryWriter] = None
+        self.flush_count = 0
+        self.event_file_base_file_path: Optional[str] = None
+        self.run_name: Optional[str] = None
 
     def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
         del log_level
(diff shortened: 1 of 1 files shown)


composer/loggers/tensorboard_logger.py
```diff
@@ -75,20 +79,34 @@ def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
...
 
         # To disable automatic flushing we set flushing to once a year ;)
         flush_secs = 365 * 3600 * 24
         self.writer = SummaryWriter(log_dir=summary_writer_log_dir, flush_secs=flush_secs)
 
+        if self.event_file_base_file_path is None:
+            self.event_file_base_file_path = self.writer.file_writer.event_writer._file_name
+
+        # Give event_file a unique name to avoid appending to the same file on every flush.
+        self.writer.file_writer.event_writer._file_name = self.event_file_base_file_path + f'-{self.flush_count}'
```
hanlint, line 107: wouldn't a unique filename for every flush lead to creation a ton of files / storage consumption for a single run? If the observed throughput degradation could be fixed by increasing the flush_interval, is that not sufficient?

I'm also not thrilled about accessing private fields of the `SummaryWriter` in this way.

composer/loggers/tensorboard_logger.py
```diff
@@ -115,11 +131,32 @@ def _flush(self, logger: Logger):
         assert self.writer.file_writer is not None
         file_path = self.writer.file_writer.event_writer._file_name
 
+        # If no writes have happened since the last flush, then file_path won't exist, so
+        # we should skip doing flushing and skip filing the artifact since it will error
+        # out anyway (given that the file_path doesn't exist).
+        if not Path(file_path).exists():
+            return
+
+        assert self.writer is not None
```
hanlint, line 140: given that we do `        assert self.writer.file_writer is not None` above, is this still neccessary?

Also, I see that the writer has already been flushed in lines 128-129, so is this second flush necessary given that nothing new had changed about the writer?

composer/loggers/tensorboard_logger.py
```diff
@@ -115,11 +131,32 @@ def _flush(self, logger: Logger):
         assert self.writer.file_writer is not None
         file_path = self.writer.file_writer.event_writer._file_name
 
+        # If no writes have happened since the last flush, then file_path won't exist, so
```
hanlint, line 134: This comment is too verbose

composer/loggers/tensorboard_logger.py
```diff
@@ -115,11 +131,32 @@ def _flush(self, logger: Logger):
...
             file_path=file_path,
             overwrite=True)
+        
+        # Close writer and reinitialize it with a new log file path. This ensures that 
+        # we have one file per flush, which means `file_artifact` is always copying 
+        # data points that have never been copied before. Also this ensures no strange 
+        # issues with dropping of points when setting a new log file path.
+        self.writer.close()
+        self.flush_count += 1
+        self._initialize_summary_writer()
```
hanlint, line 162: Seems convoluted to me that with every call to `flush()`, we are basically closing the writer and creating a new one (to a new file). I don't think this complicated handling is really needed, but rather just set a higher `flush_interval`?

composer/loggers/tensorboard_logger.py
```diff
@@ -75,20 +79,34 @@ def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
...
+        assert self.run_name is not None
+        assert self.log_dir is not None
         # We name the child directory after the run_name to ensure the run_name shows up
         # in the Tensorboard GUI.
-        summary_writer_log_dir = Path(self.log_dir) / state.run_name
+        summary_writer_log_dir = Path(self.log_dir) / self.run_name
 
         # To disable automatic flushing we set flushing to once a year ;)
         flush_secs = 365 * 3600 * 24
         self.writer = SummaryWriter(log_dir=summary_writer_log_dir, flush_secs=flush_secs)
```
hanlint, line 101: why can't we just use SummaryWriter's own flushing frequency? Can we add a concise comment here?

composer/loggers/tensorboard_logger.py
```diff
@@ -75,20 +79,34 @@ def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
                 pass
 
     def init(self, state: State, logger: Logger) -> None:
-        from torch.utils.tensorboard import SummaryWriter
+        self.run_name = state.run_name
 
         # We set the log_dir to a constant, so all runs can be co-located together.
```
hanlint, line 84:
```suggestion
        # fixed log_dir so all runs are co-located
```

composer/loggers/tensorboard_logger.py
```diff
@@ -115,11 +131,32 @@ def _flush(self, logger: Logger):
...
+
+        assert self.writer is not None
+        self.writer.flush()
+
+
         logger.file_artifact(
             LogLevel.FIT,
             # For a file to be readable by Tensorboard, it must start with
             # 'events.out.tfevents'. Child directory is named after run_name, so the logs
             # are named properly in the Tensorboard GUI.
```
hanlint, line 148:
```suggestion
```

hanlint (commented):
I think more context is needed here as to why we need to create a new file for every flush (and also re-initialize the logger). This seems convoluted and counter to how `SummaryWriter` is supposed to be used?

composer/loggers/tensorboard_logger.py
```diff
@@ -55,6 +56,9 @@ def __init__(self, log_dir: Optional[str] = None, flush_interval: int = 100, ran
         self.flush_interval = flush_interval
         self.rank_zero_only = rank_zero_only
         self.writer: Optional[SummaryWriter] = None
+        self.flush_count = 0
+        self.event_file_base_file_path: Optional[str] = None
```
ravi-mosaicml, line 60: Curious do we need to bind this variable to `self`? It seems to be read only in the same function where it is set.

ravi-mosaicml (commented):
Thanks for fixing this! Overall LGTM, though see comments. Curious if you had a chance to test this manually when saving TB traces to S3 buckets, and then having TB stream from S3? Would be good to validate that this doesn't break anything, even when it works locally. Happy to help test this tomorrow.

ravi-mosaicml:
> This seems convoluted and counter to how `SummaryWriter` is supposed to be used?

My guess is that the authors of the SummaryWriter assumed that tensorboard users would be training and running tensorboard locally, not saving traces to an object store. In this use case, then appending to one file is simpler and does not have O(n^2) overhead, as linux appends do not require the entire file to be overridden.

composer/loggers/tensorboard_logger.py
hanlint: Thanks for the explanation, I understand the need. However, there may be a simpler way here to have multiple event files, given that `SummaryWriter` will automatically re-open new `EventWriter`, which creates a new event file inside the logdir, e.g.
```
writer = SummaryWriter(log_dir='test_tb')
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.close()
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.close()
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.add_scalar(tag='hello', scalar_value=5)
writer.close()
```

will populate the `log_dir` with:
```
events.out.tfevents.1657784030.Hanlins-M1-Macbook-Pro.local.216.0
events.out.tfevents.1657784030.Hanlins-M1-Macbook-Pro.local.216.1
events.out.tfevents.1657784030.Hanlins-M1-Macbook-Pro.local.216.2
```

Is that sufficient to implement this behavior?

ravi-mosaicml: Ooh interesting, if `.close()` does that for us, then yes, that would be way simpler.

composer/loggers/tensorboard_logger.py
ravi-mosaicml: The tl;dr is the summarywriter flush happens in a background thread, and that could corrupt the tracefile if the background thread is flushing at the same time that the object store logger is staging the trace file for upload. Agree this should be elaborated on in a comment.

composer/loggers/tensorboard_logger.py
eracah: sorry this was just a mistaken duplicate line when transposing the old PR to remove flush_secs. fixed. good catch

composer/loggers/tensorboard_logger.py
hanlint: Yup it's handled by the `SummaryWriter`, and the tensorbord event writer ensures the files are unique with a guid, this might simplify the implementation significantly.

eracah: Great find, Hanlin! Changed it to use close. Greatly simplified a lot of logic. thanks. fixed

composer/loggers/tensorboard_logger.py
```diff
@@ -55,6 +56,9 @@ def __init__(self, log_dir: Optional[str] = None, flush_interval: int = 100, ran
         self.flush_interval = flush_interval
         self.rank_zero_only = rank_zero_only
         self.writer: Optional[SummaryWriter] = None
+        self.flush_count = 0
```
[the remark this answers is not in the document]
eracah: using the event file name, which contains a uuid.

composer/loggers/tensorboard_logger.py
eracah: using the Hanlin's `close()` idea instead.

composer/loggers/tensorboard_logger.py
eracah: done

eracah:
> Thanks for fixing this! Overall LGTM, though see comments. Curious if you had a chance to test this manually when saving TB traces to S3 buckets, and then having TB stream from S3? Would be good to validate that this doesn't break anything, even when it works locally. Happy to help test this tomorrow.

Works with streaming tensorboard logs from s3!

hanlint (approved):
code quality looks good to me, but didn't test this PR myself. 

only nit is: Did we have some functionality where the objectstorelogger would upload every new file in a folder, possibly used by the profiler traces? That might obviate the need to call a private `_file_name` variable.

ravi-mosaicml (approved):
LGTM; this looks much simpler than before. Thanks! Approving; assuming we tested this locally and when "streaming" tb logs over a bucket.

PR #1285, opened 2022-07-14
Bugs in getting_started.ipynb

rahulvigneswaran:
There seem to be several bugs in `getting_started.ipynb`. Most of them are due to changes in keywords.


hanlint (commented):
Thanks for putting in the fix! Could you revert the `ComposerResNet` change, as explained in the comments?

hanlint (approved):

PR #1287, opened 2022-07-14
Checkpoint backwards compatibility for ProgressBar

hanlint:
Makes the progress bar deserialization compatible with `mosaicml<=0.8.0` checkpoints.


ravi-mosaicml (approved):
I'll presume you tested locally and this works, thanks!

PR #1290, opened 2022-07-14
Fix device inference in module surgery

hanlint:
During module surgery, the new layers need a device assignment. The previous code to infer that device had a few bugs:
(1) If the original layer did not have any `parameters()` but had `buffers()`, no device would be found.
(2) If the original layer had no parameters or buffers, no replacement would be done.

This leads to a device mismatch on algorithms such as BlurPool.

This PR will infer the device from the original layer's `parameters` _and_ `buffers`. If no device can be inferred, then the device inferred from the whole model will be used instead.

Also adds a test case for this edge condition.

### DIFF
--- a/composer/utils/module_surgery.py
+++ b/composer/utils/module_surgery.py
@@ -221,8 +221,8 @@ def _infer_device(module: torch.nn.Module) -> Optional[torch.device]:
         p = next(itertools.chain(module.parameters(), module.buffers()))
     except StopIteration:
         return None
-
-    return p.device
+    else:
+        return p.device
 
 
 def count_module_instances(module: torch.nn.Module, module_class: Union[Type[torch.nn.Module],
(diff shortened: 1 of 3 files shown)


composer/utils/module_surgery.py
```diff
@@ -210,6 +215,16 @@ def replace_module_classes(
...
 
 
+def _infer_device(module: torch.nn.Module) -> Optional[torch.device]:
+    """Attempt to infer a module's device by inspecting its parameters and buffers."""
+    try:
+        p = next(itertools.chain(module.parameters(), module.buffers()))
+    except StopIteration:
+        return None
+
+    return p.device
```
mvpatel2000, line 225: Nit: Prefer putting `return p.device` inside `try` -- even though there's only one code path, imo it's better to return while its clear `p` is in scope so its more readable.

mvpatel2000 (approved):
LGTM!

composer/utils/module_surgery.py
hanlint: done!

PR #1279, opened 2022-07-13
Enable README Doctests with GPUs

mvpatel2000:
Modifies readmes to turn on GPU tests. Shoutout to all the research people I pinged for helping me understand / fix some of the tests

Looks like blurpool is broken, causing issues with README and blurpool tests. Will still ask for review in meantime, but we should hold off merging until its fixed

PR #1292, opened 2022-07-15
Matthew/swa readme

growlix:
Updated SWA method card to current format and added readme. Typo fix to gradient clipping readme.


composer/algorithms/swa/README.md
```diff
@@ -0,0 +1,71 @@
...
+|:--|
+|*Visualization of SWA from an extensive [PyTorch blogpost about SWA](https://pytorch.org/blog/stochastic-weight-averaging-in-pytorch/).*|
+
+
+## How to Use
+
+### Composer Trainer
+
+<!-- TODO: Address timeouts -->
+<!--pytest-codeblocks:skip-->
```
mvpatel2000, line 19: Can you enable this test? See https://github.com/mosaicml/composer/pull/1279 which should be merged any sec for a few examples

growlix: Lmk if this looks right

composer/algorithms/swa/README.md
```diff
@@ -0,0 +1,80 @@
...
+swa_algorithm = SWA(
+    swa_start="1ep",
+    swa_end="2ep",
+    update_interval='5ba',
+)
+trainer = Trainer(
+    model=model,
+    train_dataloader=train_dataloader,
+    eval_dataloader=eval_dataloader,
+    max_duration="3ep",
```
mvpatel2000, line 46: Nit: can we make 2ep? Slightly faster and consistent with other tests

composer/algorithms/swa/README.md
```diff
@@ -0,0 +1,80 @@
...
+    algorithms=[swa_algorithm]
+)
+
+trainer.fit()
+
```
mvpatel2000, line 51: Nit: extra line after `trainer.fit()`

mvpatel2000 (approved):
LGTM! The original PR has one small thing I want to run by Ravi before merging, so this will have to be held off for a bit, but should be good to go once thats one.

composer/algorithms/swa/README.md
growlix: Due to the nature of the algorithm, it doesn't make sense to run it for less than 3 epochs. I know this is an example, but I think the example should reflect the best practices of the method. Can go into detail if you're curious, but happy to change if necessary.

PR #1294, opened 2022-07-15
model export for inference. Functional API

dskhudia:
Functional version of model export for inference API

~~Todo: loading from checkpoint part.~~ Done

### DIFF
--- a/composer/utils/__init__.py
+++ b/composer/utils/__init__.py
@@ -9,7 +9,9 @@ from composer.utils.file_helpers import (create_symlink_file, ensure_folder_has_
                                          ensure_folder_is_empty, format_name_with_dist, format_name_with_dist_and_time,
                                          get_file, is_tar)
 from composer.utils.import_helpers import MissingConditionalImportError, import_object
+from composer.utils.inference import export_for_inference
 from composer.utils.iter_helpers import IteratorFileStream, ensure_tuple, map_collection
+from composer.utils.misc import is_model_deepspeed
 from composer.utils.object_store import (LibcloudObjectStore, ObjectStore, ObjectStoreTransientError, S3ObjectStore,
                                          SFTPObjectStore)
 from composer.utils.retrying import retry
@@ -28,11 +30,13 @@ __all__ = [
     'SFTPObjectStore',
     'MissingConditionalImportError',
     'import_object',
+    'is_model_deepspeed',
     'StringEnum',
     'load_checkpoint',
     'save_checkpoint',
     'ensure_folder_is_empty',
     'ensure_folder_has_no_conflicting_files',
+    'export_for_inference',
     'format_name_with_dist',
     'format_name_with_dist_and_time',
     'is_tar',
--- a/composer/utils/inference.py
+++ b/composer/utils/inference.py
@@ -1,0 +1,166 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Inference-related utility functions for model export and optimizations.
+
+Used for exporting models into various formats such ONNX, torchscript etc. and apply optimizations such as fusion.
+"""
+
+import contextlib
+import copy
+import logging
+import os
+import tempfile
+from typing import Any, Callable, Optional, Sequence, Union
+
+import torch
+import torch.nn as nn
+
+from composer.utils import dist
+from composer.utils.checkpoint import download_checkpoint
+from composer.utils.iter_helpers import ensure_tuple
+from composer.utils.misc import is_model_deepspeed
+from composer.utils.object_store import ObjectStore
+from composer.utils.string_enum import StringEnum
+
+log = logging.getLogger(__name__)
+
+__all__ = ['export_for_inference', 'ExportFormat']
+
+
+class ExportFormat(StringEnum):
+    """Enum class for the supported export formats.
+
+    Attributes:
+        torchscript: Export in "torchscript" format.
+        onnx:  Export in "onnx" format.
+    """
+    TORCHSCRIPT = 'torchscript'
+    ONNX = 'onnx'
+
+
+def export_for_inference(
+    model: nn.Module,
+    target_format: Union[str, ExportFormat],
+    store_path: str,
+    store_object_store: Optional[ObjectStore] = None,
+    sample_input: Optional[Any] = None,
+    surgery_algs: Optional[Union[Callable[[nn.Module], nn.Module], Sequence[Callable[[nn.Module], nn.Module]]]] = None,
+    transforms: Optional[Union[Callable[[nn.Module], nn.Module], Sequence[Callable[[nn.Module], nn.Module]]]] = None,
+    load_path: Optional[str] = None,
+    load_object_store: Optional[ObjectStore] = None,
+    strict: bool = False,
+) -> None:
+    """Export a model for inference.
+
+    Args:
+        model (nn.Module): An instance of nn.Module. Please note that model is not modified inplace.
+            Instead, export-related transformations are applied to a  copy of the model.
+        target_format (Union[str, ExportFormat]):  Format to export to. Either ``"torchscript"`` or ``"onnx"``.
+        store_path: (str): The path for storing the exported model. It can be a path to a file on the local disk,
+        a URL, or if ``store_object_store`` is set, the object name
+            in a cloud bucket. For example, ``my_run/exported_model``.
+        store_object_store (ObjectStore, optional): If the ``store_path`` is in an object name in a cloud bucket
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~.ObjectStore` which will be used
+            to store the exported model. Set this to ``None`` if ``store_path`` is a local filepath.
+            (default: ``None``)
+        sample_input (Any, optional): Example model inputs used for tracing. This is needed for "onnx" export.
+            The ``sample_input`` need not match the batch size you intend to use for inference. However, the model
+            should accept the ``sample_input`` as is. (default: ``None``)
+        surgery_algs (Union[Callable, Sequence[Callable]], optional): Algorithms that should be applied to the model
+            before loading a checkpoint. Each should be callable that takes a model and returns modified model.
+            ``surgery_algs`` are applied before ``transforms``. (default: ``None``)
+        transforms (Union[Callable, Sequence[Callable]], optional): transformations (usually optimizations) that should
+            be applied to the model. Each should be a callable that takes a model and returns a modified model.
+            ``transforms`` are applied after ``surgery_algs``. (default: ``None``)
+        load_path (str): The path to an existing checkpoint file.
+            It can be a path to a file on the local disk, a URL, or if ``load_object_store`` is set, the object name
+            for a checkpoint in a cloud bucket. For example, run_name/checkpoints/ep0-ba4-rank0. (default: ``None``)
+        load_object_store (ObjectStore, optional): If the ``load_path`` is in an object name  in a cloud bucket
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~.ObjectStore` which will be used to retreive the checkpoint.
+            Otherwise, if the checkpoint is a local filepath, set to ``None``. (default: ``None``)
+        strict (bool): Whether the keys (i.e., model parameter names) in the model state dict should
+            perfectly match the keys in the model instance. (default: ``False``)
+
+    Returns:
+        None
+    """
+    target_format = ExportFormat(target_format)
+
+    if is_model_deepspeed(model):
+        raise ValueError(f'Exporting for deepspeed models is currently not supported.')
+
+    # Only rank0 exports the model
+    if dist.get_global_rank() != 0:
+        return
+
+    # make a copy of the model so that we don't modify the original model
+    model = copy.deepcopy(model)
+
+    # Apply surgery algorithms in the given order
+    for alg in ensure_tuple(surgery_algs):
+        model = alg(model)
+
+    if load_path is not None:
+        # download checkpoint and load weights only
+        log.debug('Loading checkpoint at %s', load_path)
+        with tempfile.TemporaryDirectory() as tempdir:
+            composer_states_filepath, _, _ = download_checkpoint(path=load_path,
+                                                                 node_checkpoint_folder=tempdir,
+                                                                 object_store=load_object_store,
+                                                                 progress_bar=True)
+            state_dict = torch.load(composer_states_filepath, map_location='cpu')
+            missing_keys, unexpected_keys = model.load_state_dict(state_dict['state']['model'], strict=strict)
+            if len(missing_keys) > 0:
+                log.warning(f"Found these missing keys in the checkpoint: {', '.join(missing_keys)}")
+            if len(unexpected_keys) > 0:
+                log.warning(f"Found these unexpected keys in the checkpoint: {', '.join(unexpected_keys)}")
+
+    model.eval()
+    # Apply transformations (i.e., inference optimizations) in the given order
+    for transform in ensure_tuple(transforms):
+        model = transform(model)
+
+    is_remote_store = store_object_store is not None
+    tempdir_ctx = tempfile.TemporaryDirectory() if is_remote_store else contextlib.nullcontext(None)
+    with tempdir_ctx as tempdir:
+        if is_remote_store:
+            local_store_path = os.path.join(str(tempdir), 'model.export')
+        else:
+            local_store_path = store_path
+
+        if target_format == ExportFormat.TORCHSCRIPT:
+            export_model = None
+            try:
+                export_model = torch.jit.script(model)
+            except Exception as e:
+                log.warning(
+                    'Scripting with torch.jit.script failed with the following exception. Trying torch.jit.trace!',
+                    exc_info=True)
+                if sample_input is not None:
+                    export_model = torch.jit.trace(model, sample_input)
+                else:
+                    raise RuntimeError(
+                        'Scripting with torch.jit.script failed and sample inputs are not provided for tracing with torch.jit.trace'
+                    ) from e
+
+            if export_model is not None:
+                torch.jit.save(export_model, local_store_path)
+
+        if target_format == ExportFormat.ONNX:
+            if sample_input is None:
+                raise ValueError(f'sample_input argument is required for onnx export')
+            sample_input = ensure_tuple(sample_input)
+            torch.onnx.export(
+                model,
+                sample_input,
+                local_store_path,
+                input_names=['input'],
+                output_names=['output'],
+            )
+
+        # upload if required.
+        if is_remote_store:
+            store_object_store.upload_object(store_path, local_store_path)
--- a/tests/utils/test_inference.py
+++ b/tests/utils/test_inference.py
@@ -1,0 +1,83 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Test inference APIs.
+"""
+import os
+import tempfile
+from functools import partial
+
+import pytest
+import torch
+
+from composer.models import composer_resnet
+from composer.utils import export_for_inference
+
+
+@pytest.mark.timeout(10)
+@pytest.mark.parametrize(
+    'model_cls, sample_input',
+    [
+        (partial(composer_resnet, 'resnet18'), (torch.rand(4, 3, 224, 224), torch.randint(10, (4,)))),
+    ],
+)
+def test_export_for_inference_torchscript(model_cls, sample_input):
+    model = model_cls()
+    model.eval()
+
+    orig_out = model(sample_input)
+
+    target_format = 'torchscript'
+    with tempfile.TemporaryDirectory() as tempdir:
+        store_path = os.path.join(tempdir, f'model.pt')
+        export_for_inference(model=model, target_format=target_format, store_path=store_path)
+        loaded_model = torch.jit.load(store_path)
+        loaded_model.eval()
+        loaded_model_out = loaded_model(sample_input)
+
+        assert torch.allclose(
+            orig_out,
+            loaded_model_out), f'mismatch in the original and exported for inference model outputs with {target_format}'
+
+
+@pytest.mark.timeout(10)
+@pytest.mark.parametrize(
+    'model_cls, sample_input',
+    [
+        (partial(composer_resnet, 'resnet18'), (torch.rand(4, 3, 224, 224), torch.randint(10, (4,)))),
+    ],
+)
+def test_export_for_inference_onnx(model_cls, sample_input):
+    pytest.importorskip('onnx')
+    import onnx  # type: ignore
+    import onnxruntime as ort  # type: ignore
+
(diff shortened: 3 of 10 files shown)


ravi-mosaicml (commented):
Overall looking good! See comments.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    Returns:
+        None
+    """
+    if target_format not in ['torchscript', 'onnx']:
+        raise ValueError(
+            f'Currenlty only torchscript and onnx as target_format are supported. Received {target_format}.')
+
+    # Only rank0 exports the model
+    if dist.get_global_rank() != 0:
+        return
```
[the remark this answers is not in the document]
dskhudia: Let us do deepspeed in a separate PR. Also let us discuss this offline.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # make a copy of the model so that we don't modify the original model
+    model = copy.deepcopy(model)
+
+    # Apply surgery algorithms in the given order
+    for alg in ensure_tuple(surgery_algs):
+        model = alg(model)
+
+    if load_path is not None:
```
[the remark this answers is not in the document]
dskhudia: yes. We would need to refactor first.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+            raise ValueError(f'sample_input argument is required for onnx export')
+        sample_input = ensure_tuple(sample_input)
+        torch.onnx.export(
+            model,
+            sample_input,
+            store_path,
+            input_names=['input'],
+            output_names=['output'],
+        )
+
```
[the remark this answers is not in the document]
dskhudia: Let us keep that for future.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+            pass
+
+    model.eval()
+    # Apply transformations (i.e., inference optimizations) in the given order
+    for transform in ensure_tuple(transforms):
+        model = transform(model)
+
+    if target_format == 'torchscript':
+        model = torch.jit.script(model)
+        torch.jit.save(model, store_path)
```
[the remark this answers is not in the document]
dskhudia: Sure.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+log = logging.getLogger(__name__)
+
+__all__ = [
+    'export_for_inference',
+]
+
+
+def export_for_inference(
+    model: ComposerModel,
+    target_format: str,
```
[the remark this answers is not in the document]
dskhudia: will do

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    store_path: str,
+    store_object_store: Optional[ObjectStore] = None,
+    sample_input: Optional[Union[torch.Tensor, Sequence[torch.Tensor]]] = None,
+    surgery_algs: Optional[Union[Callable, Sequence[Callable]]] = None,
+    transforms: Optional[Union[Callable, Sequence[Callable]]] = None,
+    load_path: Optional[str] = None,
+    load_object_store: Optional[ObjectStore] = None,
+) -> None:
```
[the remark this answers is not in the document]
dskhudia: We require the store path to be provided. Model will be stored at the path. We can return it too and make store path optional.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    """Export model for inference from a local file, URI, or cloud object store.
+
+    Args:
+        model (ComposerModel): An instance of ComposerModel. Please note that model is not modified inplace.
```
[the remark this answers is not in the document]
dskhudia: We should move it to CPU. I have `state_dict = torch.load(composer_states_filepath, map_location='cpu')`

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if load_path is not None:
+        # download checkpoint and load weights only
+        tempdir_ctx = tempfile.TemporaryDirectory()
+        with tempdir_ctx as tempdir:
+            #rank_0_checkpoint_filepath = os.path.join(tempdir, 'rank0_checkpoint')
+            #get_file(destination=rank_0_checkpoint_filepath, path=load_path, object_store=load_object_store)
+            #state_dict = torch.load(composer_states_filepath, map_location='cpu')
+            #model.load_state_dict(state_dict['model'], strict=False)
+            pass
```
dblalock, line 90: Probably this is just WIP code, but gonna point out that this needs to be uncommented (or replaced)

tests/utils/test_inference.py
```diff
@@ -0,0 +1,85 @@
...
+)
+def test_export_for_inference_torchscript(model_cls, sample_input):
+    model = model_cls()
+    model.eval()
+
+    orig_out = model(sample_input)
+
+    tempdir_ctx = tempfile.TemporaryDirectory()
+    target_format = 'torchscript'
+    with tempdir_ctx as tempdir:
```
dblalock, line 33: Super minor, but can this just be:
```suggestion
    target_format = 'torchscript'
    with tempfile.TemporaryDirectory() as tempdir:
```
When I see a tempfile call outside of a `with` statement I immediately skip ahead in the function looking for where it gets cleaned up, whereas it's immediately clear that it does get cleaned up when instantiated directly in a `with` statement. Also just saves a line.

(applies in a couple other places as well)

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+from composer.models.base import ComposerModel
+from composer.utils import ObjectStore, dist
+from composer.utils.file_helpers import get_file, is_tar
+from composer.utils.iter_helpers import ensure_tuple
+
+log = logging.getLogger(__name__)
+
+__all__ = [
+    'export_for_inference',
+]
```
dblalock, line 26: nit: single line?

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    surgery_algs: Optional[Union[Callable, Sequence[Callable]]] = None,
+    transforms: Optional[Union[Callable, Sequence[Callable]]] = None,
+    load_path: Optional[str] = None,
+    load_object_store: Optional[ObjectStore] = None,
+) -> None:
+    """Export model for inference from a local file, URI, or cloud object store.
+
+    Args:
+        model (ComposerModel): An instance of ComposerModel. Please note that model is not modified inplace.
+            A copy of the model is made and export related transforms work on this copy.
```
dblalock, line 44:
```suggestion
            Instead, export-related transforms are applied to a copy of the model.
```

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+        target_format (str):  format to export to. For example, torchscript, onnx, etc.
```
dblalock, line 45:
```suggestion
        target_format (str): Format to export to. Either ``"torchscript"`` or ``"onnx"``.
```

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+        store_path: (str): The path string for storing the exported model.
+        store_object_store (ObjectStore, optional): If the ``path`` is in an object store
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~.ObjectStore` which will be used
+            to store the exported model. Otherwise, if the store_path is a local filepath, set to ``None``.
```
dblalock, line 50: Ambiguous whether "set to None" is an instruction the to the reader or a statement about the internal logic of the function.
```suggestion
            to store the exported model. Set this to ``None`` if ``store_path`` is a local filepath.
```

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+        sample_input (Union[torch.Tensor, Sequence[torch.Tensor]], optional): Example model inputs used for tracing.
```
dblalock, line 51: Is there any requirement that the input have a particular batch size (or satisfy any other constraint, like matching the intended dtype)? E.g., if I plan on doing inference with a batch size of 1, do I need to make sure my example input also has a batch size of 1?

I actually don't know the answer to this so might be helpful to add a sentence about it.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+        surgery_algs (Union[Callable, Sequence[Callable]], optional): Algorithms that should be applied to the model
+            before loading a checkpoint. Each should be callable that takes a model and returns modified model.
+        surgery_algs (Union[Callable, Sequence[Callable]], optional): transformations (usually optimizations) that should be applied to the model
+            before loading a checkpoint. Each should be a callable that takes a model and returns a modified model.
+        load_path (str): The path string to an existing checkpoint file.
```
dblalock, line 56:
```suggestion
        load_path (str): The path to an existing checkpoint file.
```

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+            It can be a path to a file on the local disk, a URL, or if ``load_object_store`` is set, the object name
+            for a checkpoint in a cloud bucket. For example, my_model/ep1-rank0.tar
+        load_object_store (ObjectStore, optional): If the ``path`` is in an object store
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~.ObjectStore` which will be used to retreive the checkpoint.
+            Otherwise, if the checkpoint is a local filepath, set to ``None``. (default: ``None``)
```
dblalock, line 62: Ideally we should be consistent about adding (default: ``whatever``) to either all args or none of them. (We used to do all, but maybe we stopped that?)

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Returns:
+        None
+    """
+    if target_format not in ['torchscript', 'onnx']:
+        raise ValueError(
+            f'Currenlty only torchscript and onnx as target_format are supported. Received {target_format}.')
```
dblalock, line 69:
```suggestion
            f'Currently only "torchscript" and "onnx" as target_format are supported. Received {target_format}.')
```

dblalock (commented):
Basically LGTM except for a couple nitpicks and one chunk of code that I assume is still WIP.

composer/utils/__init__.py
```diff
@@ -9,6 +9,7 @@
                                          ensure_folder_is_empty, format_name_with_dist, format_name_with_dist_and_time,
                                          get_file, is_tar)
 from composer.utils.import_helpers import MissingConditionalImportError, import_object
+#from composer.utils.inference import export_for_inference
```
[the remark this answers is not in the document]
dskhudia: will do. was running into a circular dependency issue but resolved now.

composer/utils/__init__.py
```diff
@@ -33,6 +34,7 @@
     'save_checkpoint',
     'ensure_folder_is_empty',
     'ensure_folder_has_no_conflicting_files',
+    #'export_for_inference',
```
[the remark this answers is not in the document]
dskhudia: will do.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+
+log = logging.getLogger(__name__)
+
+__all__ = [
+    'export_for_inference',
+]
+
+
+def export_for_inference(
+    model: ComposerModel,
```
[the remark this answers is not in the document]
dskhudia: Made it nn.Module to keep it generic.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    target_format: str,
+    store_path: str,
+    store_object_store: Optional[ObjectStore] = None,
+    sample_input: Optional[Union[torch.Tensor, Sequence[torch.Tensor]]] = None,
```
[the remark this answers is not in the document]
dskhudia: will make it Any

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    surgery_algs: Optional[Union[Callable, Sequence[Callable]]] = None,
```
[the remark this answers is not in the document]
dskhudia: Do you mean this should accept, for example, ``BlurPool()``? If yes, we can't do it here since we don't have ``state`` to pass it to ``apply`` method of algorithms. 

> mention in docstring. 
Will do

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: BTW torch.onnx.export doesn't return anything. It only stores at the specified path. So if we want to return it, we will have to load the model and return it.

composer/utils/inference.py
dskhudia: I missed adding default. Will add for all of them.

composer/utils/__init__.py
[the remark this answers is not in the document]
dskhudia: Done.

composer/utils/__init__.py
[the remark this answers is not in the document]
dskhudia: Done.

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: Done. Made it nn.Module.

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: Left the class form part as is and addressed the other concern.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+    transforms: Optional[Union[Callable, Sequence[Callable]]] = None,
+    load_path: Optional[str] = None,
+    load_object_store: Optional[ObjectStore] = None,
+) -> None:
+    """Export model for inference from a local file, URI, or cloud object store.
```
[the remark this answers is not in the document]
dskhudia: fixed.

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: Currently we don't return anything. Let me know if this is a deal breaker.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+        target_format (str):  format to export to. For example, torchscript, onnx, etc.
+        store_path: (str): The path string for storing the exported model.
+        store_object_store (ObjectStore, optional): If the ``path`` is in an object store
+            (i.e. AWS S3 or Google Cloud Storage), an instance of
+            :class:`~.ObjectStore` which will be used
+            to store the exported model. Otherwise, if the store_path is a local filepath, set to ``None``.
+        sample_input (Union[torch.Tensor, Sequence[torch.Tensor]], optional): Example model inputs used for tracing.
+        surgery_algs (Union[Callable, Sequence[Callable]], optional): Algorithms that should be applied to the model
+            before loading a checkpoint. Each should be callable that takes a model and returns modified model.
+        surgery_algs (Union[Callable, Sequence[Callable]], optional): transformations (usually optimizations) that should be applied to the model
```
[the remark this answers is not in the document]
dskhudia: Fixed.

composer/utils/inference.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def export_for_inference(
+    model: ComposerModel,
+    target_format: str,
+    store_path: str,
+    store_object_store: Optional[ObjectStore] = None,
+    sample_input: Optional[Union[torch.Tensor, Sequence[torch.Tensor]]] = None,
+    surgery_algs: Optional[Union[Callable, Sequence[Callable]]] = None,
+    transforms: Optional[Union[Callable, Sequence[Callable]]] = None,
```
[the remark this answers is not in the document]
dskhudia: Fixed it.

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: Added an error for deepspeed. `is_model_deepspeed` is copied from `State`. Ok do duplicate this code?

composer/utils/inference.py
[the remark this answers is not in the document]
dskhudia: Actually `_download_checkpoint` can be used as is. Renaming it to `download_checkpoint`

composer/utils/inference.py
dskhudia: Yes. This was WIP. Fixed it.

tests/utils/test_inference.py
dskhudia: Fixed it here and at a couple of other places.

composer/utils/inference.py
dskhudia: Fixed.

composer/utils/inference.py
dskhudia: Thanks. Fixed.

composer/utils/inference.py
dskhudia: Fixed.

composer/utils/inference.py
dskhudia: Fixed.

composer/utils/inference.py
dskhudia: There is no such requirement. Added a comment clarifying this in the docstring for sample_input.

composer/utils/inference.py
dskhudia: Fixed.

composer/utils/inference.py
dskhudia: Added defaults.

composer/utils/inference.py
dskhudia: This got removed.

composer/utils/inference.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Attributes:
+        torchscript: Export in "torchscript" format.
+        onnx:  Export in "onnx" format.
+    """
+    TORCHSCRIPT = 'torchscript'
+    ONNX = 'onnx'
+
+
+def is_model_deepspeed(model) -> bool:
```
ravi-mosaicml, line 41: Could we move this into a shared helper?

ravi-mosaicml, line 41:
```suggestion
def is_model_deepspeed(model: torch.nn.Module) -> bool:
```

composer/utils/inference.py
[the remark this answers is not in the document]
ravi-mosaicml: Ah ok, makes sense not to return then.

composer/utils/inference.py
```diff
@@ -0,0 +1,175 @@
...
+        if is_remote_store:
+            local_store_path = os.path.join(str(tempdir), 'model.export')
+        else:
+            local_store_path = store_path
+
+        if target_format == ExportFormat.TORCHSCRIPT:
+            export_model = None
+            try:
+                export_model = torch.jit.script(model)
+            except Exception as e:
```
ravi-mosaicml, line 147: Can we let the exception propagate through?

composer/utils/inference.py
```diff
@@ -0,0 +1,175 @@
...
+                log.warning(
+                    f'Scripting with torch.jit.script failed with the following exception. Trying torch.jit.trace!'
+                    f' {str(e)}')
+            if sample_input is not None:
```
ravi-mosaicml, line 151: Should we attempt a torch.jit.script export when a sample input is provided? Or in this case should we just use a trace export?

In addition, script is successful, then I don't think we need to attempt a jit?

composer/utils/inference.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if target_format == ExportFormat.ONNX:
+            if sample_input is None:
+                raise ValueError(f'sample_input argument is required for onnx export')
+            sample_input = ensure_tuple(sample_input)
+            torch.onnx.export(
+                model,
+                sample_input,
+                local_store_path,
+                input_names=['input'],
```
ravi-mosaicml, line 169: What do the input_names and output_names do? Would the user want to customize it?

ravi-mosaicml (approved):
LGTM! See comment re: torchscript trace/jit logic.

composer/utils/inference.py
dskhudia: These will be the name of the input and output for the onnx graph. A user may want to customize these but let us leave it as is until we have such a request.

composer/utils/inference.py
dskhudia: What do you mean? If script fails, we have to try tracing so need to catch exception.

composer/utils/inference.py
dskhudia: script captures more of the model if it works. So IMO that should be the default.

composer/utils/inference.py
dskhudia: Added a standalone functions and replaced all uses.

composer/utils/inference.py
dskhudia: Do you mean to rethrow if tracing also fails?

composer/utils/inference.py
dskhudia: > In addition, script is successful, then I don't think we need to attempt a jit?

Good catch. Fixed it.

composer/utils/inference.py
ravi-mosaicml: Ah I see, I think I missed the indentation. This logic works, though thoughts about doing something like this?

```python
if sampel_input is None:
    model = torch.jit.script(model)
else:
    model = torch.jit.trace(model, sample_input)
```

My slight preference is for the above (I think it's slightly easier to follow / reason though), but also understand if torch jit script is superior and we should use that whenever possible.

composer/utils/inference.py
```diff
@@ -0,0 +1,166 @@
...
+            local_store_path = os.path.join(str(tempdir), 'model.export')
+        else:
+            local_store_path = store_path
+
+        if target_format == ExportFormat.TORCHSCRIPT:
+            export_model = None
+            try:
+                export_model = torch.jit.script(model)
+            except Exception as e:
+                log.warning(
```
ravi-mosaicml, line 139: nit: instead of stringifying the exception (which would lose the stack trace), set `exc_info=True` -- see https://docs.python.org/3/library/logging.html#logging.debug

composer/utils/inference.py
```diff
@@ -0,0 +1,166 @@
...
+                    f'Scripting with torch.jit.script failed with the following exception. Trying torch.jit.trace!'
+                    f' {str(e)}')
+                if sample_input is not None:
+                    export_model = torch.jit.trace(model, sample_input)
+                else:
+                    log.error(
```
ravi-mosaicml, line 145: in this case can we raise an exception and propagate the original exception? e.g. `raise RuntimeError('Scripting with torch.jit.script failed and sample inputs are not provided for tracing with torch.jit.trace') from e`

composer/utils/inference.py
```diff
@@ -0,0 +1,166 @@
...
+                        f'Scripting with torch.jit.script failed and sample inputs are not provided for tracing with torch.jit.trace'
+                    )
+
+            if export_model is not None:
```
ravi-mosaicml, line 149: if export_model is None, then should we raise an exception?

ravi-mosaicml (commented):
Just a few more comments -- feel free to merge after addressing.

composer/utils/inference.py
dskhudia: Control wouldn't reach here if export_model is None. Earlier code path will have already raise an exception.

PR #1295, opened 2022-07-16
Add monitored barrier for autograd accum

mvpatel2000:
Resolves this [jira](https://mosaicml.atlassian.net/jira/software/c/projects/CO/boards/12?modal=detail&selectedIssue=CO-676&assignee=61fc1878aab3620070f699bf), which means users now get an error and a helpful message if a deadlock is encountered with auto grad accum

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -793,6 +793,8 @@ class Trainer:
         # Grad Accum
         self.adaptive_gradient_accumulation = _is_adaptive_grad_accum(grad_accum, device=self._device)
         grad_accum = _get_initial_grad_accum(grad_accum)
+        # Dynamic time estimate for forward and backward pass. Used for monitored_barrier to avoid deadlocks
+        self.batch_compute_time = 300
 
         # Grad Clip Norm
         if grad_clip_norm > 0:
@@ -1634,6 +1636,7 @@ class Trainer:
 
         # Retry until we successfully complete training and return loss
         while True:
+            start_time = time.time()
             total_loss = None
             # Note: We use uint8 instead of bool as BOR is not supported on all torch.distributed backends
             should_handle_cuda_oom = 0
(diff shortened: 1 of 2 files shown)


mvpatel2000:
Hm... looks like I'm not getting a deadlock with the test example Linden gave. Will try to repro issue so I can test fix

composer/trainer/trainer.py
```diff
@@ -750,7 +750,8 @@ def __init__(
         self._device = _get_device(device)
 
         # Distributed
-        if deepspeed_enabled or dist.get_world_size() > 1:
+        self.use_dist = deepspeed_enabled or dist.get_world_size() > 1
```
ravi-mosaicml, line 753: Curious why do we need this variable on self?

composer/trainer/trainer.py
```diff
@@ -1662,14 +1662,21 @@ def _train_batch(self, use_grad_scaling: bool):
...
-                    # catch non-transient timeout errors they will be later reraised if no rank OOMed.
-                    caught_timeout_error = e
                 else:
                     raise
 
+            # Use monitored barrier to error on deadlock. If one rank OOMs and another doesn't and gets stuck
+            # on a dist reduction in gradient syncronization, the monitored barrier will fail after the timeout.
+            if self.use_dist:
+                try:
+                    group_gloo = torch.distributed.new_group(backend='gloo')
```
ravi-mosaicml, line 1672: Is it possible to re-use the existing distributed group?

ravi-mosaicml (commented):
Thanks for adding this in; see comments.

composer/trainer/trainer.py
mvpatel2000: Monitored barrier requires `gloo`. Added in comments

mvpatel2000:
Test results showing monitored_barrier works. 

image

composer/trainer/trainer.py
```diff
@@ -1662,14 +1662,21 @@ def _train_batch(self, use_grad_scaling: bool):
...
+                    torch.distributed.monitored_barrier(group=group_gloo, timeout=datetime.timedelta(seconds=300))
+                except:
```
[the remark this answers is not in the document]
mvpatel2000: yep, runtime error

composer/trainer/trainer.py
```diff
@@ -1662,14 +1661,19 @@ def _train_batch(self, use_grad_scaling: bool):
...
-                    # is likely transient if one rank OOMed, it likely did not reach a barrier. Note that if we
-                    # catch non-transient timeout errors they will be later reraised if no rank OOMed.
-                    caught_timeout_error = e
                 else:
                     raise
 
+            # Use monitored barrier to error on deadlock. If one rank OOMs and another doesn't and gets stuck
+            # on a dist reduction in gradient syncronization, the monitored barrier will fail after the timeout.
+            try:
+                dist.monitored_barrier(timeout=datetime.timedelta(seconds=300))
```
ravi-mosaicml, line 1670: 5 minutes seems....like a lot? Curious if we could shorten the time for a better UX for debugging? Would it make sense to set the timeout dynamically (say, `min(300, max(10, 0.5 * forward_backwards_pass_time))`)?

composer/utils/dist.py
```diff
@@ -164,6 +165,28 @@ def barrier() -> None:
...
+    """Synchronizes all processes.
+
+    This function blocks until all processes reach this function. Unlike `barrier`, `monitored_barrier`
+    times out and raises an error if not all ranks reach this function by `timeout`.
+
+    .. seealso:: :func:`torch.distributed.barrier`
+    """
+    if dist.is_available() and dist.is_initialized():
+        # monitored_barrier requires gloo backend
+        group_gloo = dist.new_group(backend='gloo')
```
ravi-mosaicml, line 178: Instead of creating the group here, could you create the a gloo group as part of `initailize_dist`, and assign it to a global variable?

ravi-mosaicml (approved):
Overall LGTM! See comments.

PR #1296, opened 2022-07-16
Require JIRA Links

ravi-mosaicml:
This PR requires that all (future) PRs include a JIRA link in the PR description.

Linking each PR to a JIRA issue will streamline release management. Specifically, it will help identify which commits should be cherry-picked in for each release based on the "fix version" field in Jira. In the future, we could write a script to automate this process. If we later added a standardized formatting for a release notes section in the PR description too, then we could also automate the release note generation.

For external contributors, we will need to manually create a Jira story for them, and add the link to the PR description, before they can merge. Tests should still run, however, without this link, since this check is performed last. Features and bug fixes completed by external contributors should still have a corresponding JIRA story, since these features should be included in release notes, too!

Closes https://mosaicml.atlassian.net/browse/CO-729

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -296,12 +296,7 @@ stage('Prepare') {
             // off of the "fix version" field in JIRA
 
             if (!pullRequest.body.contains('https://mosaicml.atlassian.net/browse/CO-')) {
-                pullRequest.createStatus(
-                    status: 'error',
-                    context: 'ci/jenkins/pr-merge/has-jira-link',
-                    description: 'The PR description must contain a link to the corresponding JIRA bug or story',
-                    targetUrl: "${env.JOB_URL}/testResults",
-                )
+                pullRequest.comment('The PR description must contain a link to the corresponding JIRA bug or story')
             }
 
         }
@@ -447,6 +442,10 @@ stage('Build') {
                         fingerprintArtifacts: true,
                         optional: true,
                     )
+                }
+
+                if (env.CHANGE_ID && !pullRequest.body.contains('https://mosaicml.atlassian.net/browse/CO-')) {
+                    error('The PR description must contain a link to the corresponding JIRA bug or story')
                 }
 
                 sh "mkdir -p $buildOutputFolder"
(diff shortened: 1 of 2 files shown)


hanlint:
This seems too heavy-handed, couldn't we just tag the PRs in github with the release number?

abhi-mosaic:
> I like what this PR does in theory but I also don't want to risk this grinding dev to a halt. Is there a way we can phase this in? For example, as an independent check that is not initially required? We can then communicate to everyone that this check will be required on a specific date so the team has some time to transition?

Also would prefer something like this where you get a "JIRA tag missing" warning that is non-blocking at first.

> This seems too heavy-handed, couldn't we just tag the PRs in github with the release number?

Also this... are there more things we would gain from JIRA links other than the release #?

ravi-mosaicml:
> This seems too heavy-handed, couldn't we just tag the PRs in github with the release number?

I'd rather not have the release number in two different places (github and jira); it makes things difficult to keep in sync.

ravi-mosaicml:
> > I like what this PR does in theory but I also don't want to risk this grinding dev to a halt. Is there a way we can phase this in? For example, as an independent check that is not initially required? We can then communicate to everyone that this check will be required on a specific date so the team has some time to transition?
> 
> Also would prefer something like this where you get a "JIRA tag missing" warning that is non-blocking at first.

I posted a message in slack. I think it's a very minor ask to have the PR author include a simple link to relieve hours from the release captain's job of trying to figure out which commits to cherry-pick in.

> 
> > This seems too heavy-handed, couldn't we just tag the PRs in github with the release number?
> 
> Also this... are there more things we would gain from JIRA links other than the release #?

Sure, Jira now knows the development history for each commit. We can configure automations to automatically close an issue when the corresponding PR is merged; track the actual complexity of each feature (e.g. slocs or actual development time), and have better visibility in JIRA about which PRs are actually "in progress". All of this would require additional JIRA configuration, but having the link would allow this to be possible.

ravi-mosaicml:
I'll see if I can re-write this as a github action as a separate (non required) check; we can require this check later.

bandish-shah:
Closing out for now, we can revive this in another incarnation if needed.

PR #1300, opened 2022-07-19
Retry SFTP on Size Mismatch

ravi-mosaicml:
On occasion, the SFTP Object Store could raise a file size mismatch.

This PR fixes that by manually validating the file size, and raising an `ObjectStoreTransientError` if the remote file size differs from the local file size. This behavior replicates the `confirm=True` flag that paramiko provides, but instead raises an error message that we can retry on.

Closes https://mosaicml.atlassian.net/browse/CO-642


anisehsani (approved):
lgtm! test failure doesn't seem related

ravi-mosaicml:
Going to wait to merge until I can test locally or via integration testing.

ravi-mosaicml:
> Looks good - is file size a sufficient metric for comparison?
> 
> Would it be too expensive to run a hash and compare it to a server side hash file? (this would require changing the object store API to store hashes on remote automatically, this is a potentially useful flag anyway...)

Hashes would definitely be more robust, but that would be object store dependent. Definitely doable for s3 (in fact, the hash is already stored as part of the metadata via the `ETag`). Not really doable for sftp though (see https://stackoverflow.com/questions/28967969/how-to-check-if-paramiko-successfully-uploaded-a-file-to-an-sftp-server)

eracah (approved):

ravi-mosaicml:
(Going to merge this after adding in integration testing for sftp)

ravi-mosaicml:
#1322 includes the integration tests to validate that this change doesn't break stuff. Once that pr is merged, then merge dev into this PR, retrigger Jenkins to run the integration tests, and merge!

PR #1302, opened 2022-07-20
[xs] Store the Composer version in one place.

ravi-mosaicml:
This PR places the composer version number in just one place -- `composer._version`.

Closes https://mosaicml.atlassian.net/browse/CO-556

### DIFF
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.8.0'
+__version__ = '0.8.0-dev'
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.8.0-dev'
+__version__ = '0.8.0'
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -1,0 +1,6 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""The Composer Version."""
+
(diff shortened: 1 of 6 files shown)


milocress (approved):

composer/utils/collect_env.py
```diff
@@ -348,7 +348,7 @@ def print_env(file: Optional[TextIO] = None) -> None:
 
         Composer information
         --------------------
-        Composer version: 0.8.0
+        Composer version: 0.4.2
```
ravi-mosaicml, line 351: Instead of trying to keep this up to date, let's make it a meme...

composer/_version.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""The Composer Version."""
+
+__version__ = '0.8.0'
```
dskhudia, line 6: I think by default we should append it with 'dev' or something similar so that users know if they are on dev or a released version.

composer/utils/collect_env.py
hanlint: Yes you could do:
```
__doc__ = """blah blah {blah}""".format(blah='blah')
```

But this is fine too :)

milocress: @hanlint just out of curiosity, is python smart enough to unroll the constants like that at function definition time rather than at function evaluation?

composer/_version.py
ravi-mosaicml: Discussed offline; gonna leave this out cause python doesn't respect semantic versioning.

PR #1304, opened 2022-07-21
Fix `get_file_artifact` in the WandBLogger to work on all ranks

ravi-mosaicml:
Fix`get_file_artifact` to work on all ranks, not just the rank zero even if `rank_zero_only=True`

This allows for resuming from checkpoints stored in WandB in multi-node settings, even if `rank_zero_only=True`.

Added a `@pytest.mark.remote` test to verify the fix. Verified that this test passed locally.

Closes https://mosaicml.atlassian.net/browse/CO-640

### DIFF
--- a/composer/loggers/wandb_logger.py
+++ b/composer/loggers/wandb_logger.py
@@ -95,6 +95,10 @@ class WandBLogger(LoggerDestination):
         self._init_kwargs = init_kwargs
         self._is_in_atexit = False
 
+        # These variables are set from global rank 0 to all ranks
+        self.entity = None
+        self.project = None
+
     def _set_is_in_atexit(self):
         self._is_in_atexit = True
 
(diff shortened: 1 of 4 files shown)


abhi-mosaic:
I'm going to launch a quick run with multi-gpu and auto-resumption to verify that this works.

hanlint:
cc: @eracah for a review as the logger whiz

eracah:
> cc: @eracah for a review as the logger whiz

sorry for the delay! I will take a look now

eracah (approved):
LGTM. Thanks, for fixing this issue, @ravi-mosaicml

abhi-mosaic (changes requested):
I'm running into an issue when I launch a run with `autoresume: true` and there are no artifacts existing yet, I see ranks failing with 
```
wandb.errors.CommError: Invalid artifact path: mosaic-ml/abhi-autoresume/gpt3-125m-safu/checkpoints/latest-rank7:latest
```

abhi-mosaic:
OK ran into many, many errors along the way, but I managed to launch an 8-GPU run with rank-zero-only checkpointing to WandB, followed by a second 8-GPU run with carefully constructed parameters, and the resume worked. So the direct issue of multi-gpu resumption from WandB is fixed.

Screen Shot 2022-07-22 at 4 16 43 PM


However there are **a lot** of UX improvements left, and I'll make a JIRA issue to document them. The only one that is immediately relevant to this PR though: Can we remove the `extension` handling for now as it looks really ugly in WandB due to our default `save_artifact_name` not having an extension at the end.

Screen Shot 2022-07-22 at 4 25 30 PM

abhi-mosaic:
This is the cleanest I was able to get WandB artifact logging to look, and I was able to run the same YAML on both run0 and run1. However there are still unique YAMLs for each run, which I think is kinda unavoidable for now since it lives in our entrypoint.

Screen Shot 2022-07-22 at 4 54 00 PM


And here are the settings:
```yaml
  autoresume: true
  save_artifact_name: "gpt3-125m-fixed-run-name/checkpoints"
  save_latest_artifact_name: "gpt3-125m-fixed-run-name/latest"

  save_folder: checkpoints
  save_interval: 20ba

  loggers:
    wandb:
      log_artifacts: true
      rank_zero_only: true
```

abhi-mosaic (approved):
I think this is fine to merge for now, but looking forward to making WandB artifact logging more visually friendly soon.

ravi-mosaicml:
Yes, please document the issues to clean up the WandB ux in a jira. Happy to address in another PR.

PR #1305, opened 2022-07-21
Integration Testing

ravi-mosaicml:
Run integration tests (those marked with `@pytest.mark.remote` on each PR)

Closes https://mosaicml.atlassian.net/browse/CO-627

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -52,7 +52,7 @@ condaTimeout = '7200' // timeout for conda builds, in seconds
 
 // Jenkins settings
 numDaysOfBuildsToKeep = '7' // number of days to keep builds (so Jenkins doesn't crash)
-jenkinsShellJobName = 'scratch/commandNext' // The jenkins job name used to spawn sub-jobs
+jenkinsShellJobName = 'scratch/command2' // The jenkins job name used to spawn sub-jobs
 gitCredentialsId = '9cf9add1-2cdd-414b-8160-94bd4ac4a13d' // Jenkins credential ID to use for git clones
 wandbCredentialsId = 'e6fbcd33-e739-483f-a4a9-9b6815572c8b'
 awsCredentialsId = 'd931e1a8-f356-42f4-bc4d-bc8582c3bf9e'
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -167,6 +167,7 @@ void runLint(String pDockerImage) {
             string(name: 'P_GIT_REPO', value: gitUrl),
             string(name: 'P_GIT_COMMIT', value: gitCommit),
             string(name: 'P_EPHEMERAL_STORAGE_LIMIT', value: lintMemLimit),
+            booleanParam(name: 'P_USE_TMP_RAMDISK', value: true),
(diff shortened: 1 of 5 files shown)


.ci/test.sh
```diff
@@ -8,6 +8,13 @@ set -exuo pipefail
 EXTRA_DEPS="$1"
 MARKERS="$2"
 
+# Integration test settings
+export WANDB_ENTITY='mosaicml-public-integration-tests'
+export WANDB_PROJECT='integration-tests'
```
bandish-shah, line 13: Does each test create a run?  Wondering if we should have a separate project with a unique ID or timestamp for each invocation of the test suite if that's the case.

tests/utils/object_store/test_s3_object_store.py
```diff
@@ -21,17 +21,13 @@ def _worker(bucket: str, tmp_path: pathlib.Path, tid: int):
 # This test requires properly configured aws credentials; otherwise the s3 client would hit a NoCredentialsError
 # when constructing the Session, which occurs before the bug this test checks
 @pytest.mark.remote
-def test_s3_object_store_multi_threads(tmp_path: pathlib.Path):
+def test_s3_object_store_multi_threads(tmp_path: pathlib.Path, s3_bucket: str):
```
bandish-shah, line 24: Just a general question on `ObjectStoreLogger` tests, do we have any tests that vary `flush_interval`?  I noticed while working on a demo that setting this value to a small value can cause some hangs.  Probably a topic for another issue but just curious.

bandish-shah (approved):
LGTM, few minor comments and questions.

.ci/Jenkinsfile
```diff
@@ -52,8 +52,11 @@ condaTimeout = '7200' // timeout for conda builds, in seconds
 
 // Jenkins settings
 numDaysOfBuildsToKeep = '7' // number of days to keep builds (so Jenkins doesn't crash)
-jenkinsShellJobName = 'scratch/command2' // The jenkins job name used to spawn sub-jobs
+jenkinsShellJobName = 'scratch/commandNext' // The jenkins job name used to spawn sub-jobs
 gitCredentialsId = '9cf9add1-2cdd-414b-8160-94bd4ac4a13d' // Jenkins credential ID to use for git clones
+wandbCredentialsId = 'e6fbcd33-e739-483f-a4a9-9b6815572c8b'
+awsCredentialsId = 'd931e1a8-f356-42f4-bc4d-bc8582c3bf9e'
+sshCredentialsId = 'f71fda19-c867-4e7b-b6e8-3894e8288076'
```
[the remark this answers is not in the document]
ravi-mosaicml: Correct; these are Jenkins IDs corresponding to the secrets.

tests/utils/object_store/test_s3_object_store.py
ravi-mosaicml: I don't think we vary the flush interval.
Can you file a bug report for an example where it hangs?

.ci/test.sh
ravi-mosaicml: Only tests marked as `@remote` that use wandb will create a run. Good point -- will create a seperate project for each build

ravi-mosaicml: Resolved.

PR #1306, opened 2022-07-21
[Hot fix] Remove compression from Bert example

bandish-shah:
* Using compressed version of the C4 streaming dataset is causing regression failures
* Pointing the example to the un-compressed version for v0.8.1
* Fixes CO-763


abhi-mosaic (approved):
LGTM

PR #1307, opened 2022-07-22
Add a `return_outputs` flag to `predict()`

ravi-mosaicml:
This PR adds a `return_outputs` flag to `predict()`, which when set, will move all batch outputs to cpu and accumulate them in a python list when running `predict()`.

Closes https://mosaicml.atlassian.net/browse/CO-765

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1836,6 +1836,54 @@ class Trainer:
     ):
         """Output model prediction on the provided data.
 
+        There are two ways to access the prediction outputs.
+
+        1.  With ``return_outputs`` set to True, the batch predictions will be collected into a list and returned.
+        2.  Via a custom callback, which can be used with ``return_outputs`` set to False.
+
+            This technique can be useful if collecting all the outputs from the dataloader would exceed available memory,
+            and you want to write outputs directly to files. For example:
+
+            .. testsetup::
+
+                predict_dl = train_dataloader
+
+            .. testcode::
+
+                import os
+                import torch
+
+                from torch.utils.data import DataLoader
+
+                from composer import Trainer, Callback
+                from composer.loggers import Logger, LogLevel
+
+                class PredictionSaver(Callback):
+                    def __init__(self, folder: str):
+                        self.folder = folder
+                        os.makedirs(self.folder, exist_ok=True)
+
+                    def predict_batch_end(self, state: State, logger: Logger) -> None:
+                        name = f'batch_{int(state.predict_timestamp.batch)}.pt'
+                        filepath = os.path.join(self.folder, name)
+                        torch.save(state.outputs, filepath)
+
+                        # Also log the outputs as an artifact
+                        logger.file_artifact(LogLevel.BATCH, artifact_name=name, file_path=filepath)
+
+                trainer = Trainer(
+                    ...,
+                    callbacks=PredictionSaver('./predict_outputs'),
+                )
+
+                trainer.predict(predict_dl, return_outputs=False)
+
+                print(sorted(os.listdir('./predict_outputs')))
+
+            .. testoutput::
+
+                ['batch_1.pt', ...]
+
         Args:
             dataloader (DataLoader | DataSpec): The :class:`.DataLoader` or
                 :class:`.DataSpec` for the prediction data.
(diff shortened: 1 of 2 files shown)


mvpatel2000 (approved):
LGTM

composer/trainer/trainer.py
```diff
@@ -1680,7 +1680,7 @@ def _train_batch(self, use_grad_scaling: bool):
                     'least one rank encountered a CUDA OOM error.') from e
             # Synchronize new batch compute time
             batch_compute_time = end_time - start_time
-            batch_compute_time = self._device.tensor_to_device(torch.tensor([batch_compute_time], dtype=torch.uint8))
+            batch_compute_time = self._device.tensor_to_device(torch.tensor([batch_compute_time], dtype=torch.float))
```
mvpatel2000, line 1683: good catch

composer/trainer/trainer.py
```diff
@@ -1884,6 +1946,9 @@ def predict(self, dataloader: Union[DataLoader, DataSpec], subset_num_batches: i
                 self.state.outputs = self.state.model(self.state.batch)
                 self.engine.run_event(Event.PREDICT_AFTER_FORWARD)
 
+                if return_outputs:
+                    outputs.append(cpu_device.batch_to_device(self.state.outputs))
```
hanlint, line 1950: should we concatenate along the batch dimension so its one tensor/np array?

composer/trainer/trainer.py
```diff
@@ -1836,6 +1842,59 @@ def predict(self, dataloader: Union[DataLoader, DataSpec], subset_num_batches: i
...
+
+                    If collecting all outputs from the dataloader would exceed available memory, we recommend
+                    to set `return_outputs` to False and use your own callback to write the outputs from each
+                    prediction batch to a file. For example:
+
+                    .. testsetup::
+
+                        predict_dl = train_dataloader
+
+                    .. testcode::
```
hanlint, line 1859: should we just make this its won callback and backlink it here?

hanlint: if not, we should move this into the main body of the docstrings as a "There are two ways to access the outputs of `predict`. With `return_outputs=True`, you can use the list of answers and save them to disk yourself, e.g. `outputs = trainer.predict(..); torch.save(outputs, ..)`.

Or if the total outputs are too large, use `return_outputs=False`, and add the predictionCallback to save things to disk or object store  with each batch.

composer/trainer/trainer.py
eracah: +1

tests/trainer/test_predict.py
```diff
@@ -28,6 +33,21 @@ def _assert_predict_events_called_expected_number_of_times(
         assert expected == actual, f'Event {event} expected to be called {expected} times, but instead it was called {actual} times'
 
 
+class PredictionSaver(Callback):
+
+    def __init__(self, folder: str):
+        self.folder = folder
+        os.makedirs(self.folder, exist_ok=True)
+
+    def predict_batch_end(self, state: State, logger: Logger) -> None:
```
eracah, line 42: comment: there might be a situation, where the logger has a `log_table` function that we call here

eracah (approved):

composer/trainer/trainer.py
ravi-mosaicml: I was thinking about this.....and was leaning against it, as the outputs _could_ be different shapes. Or perhaps the output could be a dictionary. IDK, thoughts?

composer/trainer/trainer.py
ravi-mosaicml: I'd rather leave it as a docstring for now until we can write a robust callback. But agreed, I'll put this high level in the docstring.

composer/trainer/trainer.py
hanlint: Oh, good point... Don't we have smart infer for concat based on type, or is that just for split? I dont think this is a blocker for merging

eracah: what about using a nestedtensor for outputs of different shapes? https://pytorch.org/docs/stable/nested.html

hanlint: I'd hesitate to have our public API depends on something in prototype stage at this point.

composer/trainer/trainer.py
ravi-mosaicml: Doctests were failing locally for me...

composer/trainer/trainer.py
ravi-mosaicml: I'll leave this as-is for now. We can always add in a collation argument later to try to consolidate it into one data structure. The user can pass a custom collate fn as part of their data spec, but we don't have a generic collate fn like we do with a batch split one.

composer/trainer/trainer.py
ravi-mosaicml: Updated the docstring.

PR #1308, opened 2022-07-22
[xs] Build the Composer Docker Image only on `dev` branch merges

ravi-mosaicml:
The composer dockerimage should only be built when merging into dev (or on a pr that does that); it should not be built on release branches, as the correct composer version is not yet published on pypi (and the build will thus fail!)

Closes https://mosaicml.atlassian.net/browse/CO-761


bandish-shah (approved):
LGTM

PR #1309, opened 2022-07-22
Enforce FusedLayerNorm is ordered last

alextrott16:
Adds a single line of logic into `composer.core.engine` so that `FusedLayerNorm` is last in the algorithm execution order.

This will ensure that any `torch.nn.LayerNorm` modules which are added by other model surgery algorithms are caught by `FusedLayerNorm` and converted into the fused versions.

As a motivating use case, when `GatedLinearUnits` and `FusedLayerNorm` are used together on a BERT model, having `GatedLinearUnits` perform surgery last will result in a model that is roughly 50/50 fused and unfused layernorm instances.

### DIFF
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -321,7 +321,7 @@ class Engine():
         Returns:
             Sequence[Algorithm]: Modified sequence of algorithms.
         """
-        from composer.algorithms import CutMix, MixUp, SelectiveBackprop, StochasticDepth, FusedLayerNorm
+        from composer.algorithms import CutMix, FusedLayerNorm, MixUp, SelectiveBackprop, StochasticDepth
 
         # Move selective backprop to the beginning while maintaining order of other algorithms
         algorithms = sorted(algorithms_to_run,
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -321,11 +321,14 @@ class Engine():
         Returns:
             Sequence[Algorithm]: Modified sequence of algorithms.
         """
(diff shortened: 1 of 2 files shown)


composer/core/engine.py
```diff
@@ -321,12 +321,15 @@ def _compile(
...
         """
-        from composer.algorithms import CutMix, MixUp, SelectiveBackprop, StochasticDepth
+        from composer.algorithms import CutMix, MixUp, SelectiveBackprop, StochasticDepth, FusedLayerNorm
 
         # Move selective backprop to the beginning while maintaining order of other algorithms
         algorithms = sorted(algorithms_to_run,
                             key=lambda x: not isinstance(x, SelectiveBackprop) and not isinstance(x, StochasticDepth))
 
+        # Move fused layernorm to the end while maintaining order of other algorithms (FLN only does surgery on leaf modules)
+        algorithms = sorted(algorithms_to_run, key=lambda x: isinstance(x, FusedLayerNorm))
```
ravi-mosaicml, line 331: I presume we want to keep the previous sort from line 327?

```suggestion
        algorithms = sorted(algorithms, key=lambda x: isinstance(x, FusedLayerNorm))
```

ravi-mosaicml (approved):
LGTM; see comment.

ravi-mosaicml:
At some point I think we'll need to write a DAG scheduler for the algorithms....but that's a future problem :)

composer/core/engine.py
alextrott16: Yes, thank you!

hanlint (approved):
LGTM

PR #1314, opened 2022-07-25
Add json schemas to docs

mvpatel2000:
Adds json schemas to docs. Adds a test to make sure the generated files are the most recent version


mvpatel2000:
Still need to verify docs generate properly. Tagging for review for discussion of questions

ravi-mosaicml (commented):
Thanks for adding this! Wow, this will be powerful.

Discussed offline, but can we use jsonschema `$defs` or `$refs` to reduce the file size of the json schemas?

Afterwards, I'll take a look on how this shows on the docs site.

moinnadeem:
Is this related to CO-669, CO-670, or CO-671? @ravi-mosaicml @mvpatel2000

mvpatel2000:
> Is this related to CO-669, CO-670, or CO-671? @ravi-mosaicml @mvpatel2000

CO-671. The first two are done, though there's some changes we had to make as a result of this PR [here](https://github.com/mosaicml/yahp/pull/124).

.ci/test_lint_doctests.py
```diff
@@ -81,6 +82,36 @@ def test_docker_build_matrix():
         assert existing_build_matrix == f.read()
 
 
+@pytest.mark.timeout(30)
+def test_json_schemas():
```
ravi-mosaicml, line 86: In addition to this check that validates the json schemas are up to date, can we use a pre-commit hook to validate the yamls all pass the jsonschema check? See https://github.com/python-jsonschema/check-jsonschema

composer/trainer/trainer_hparams.py
```diff
@@ -262,6 +262,8 @@ class TrainerHparams(hp.Hparams):
         grad_clip_norm (float, optional): See :class:`.Trainer`.
 
         profiler (ProfilerHparams, optional): Profiler hyperparameters.
+
+    .. jsonschema:: ../json_schemas/trainer_hparams.json
```
ravi-mosaicml, line 266: Instead of generating a new jsonschema for each hparams class, and linking in the docs to that file, can we link the the definition of that hparams class within the experiment hparams json schema? The experiment hparams is the "root", and all possible hparams should be used within it. The JSONSchema plugin supports JSONPointer notation.

For example:

```suggestion
    .. jsonschema:: ../json_schemas/experiment_hparams.json#/defs/trainer_hparams
```

This should help reduce the duplication, and make the docs site more accessible (since there will be just one jsonschema entry)

.ci/test_lint_doctests.py
```diff
@@ -81,6 +82,36 @@ def test_docker_build_matrix():
...
+    for filename in sorted(os.listdir(schemas_folder)):
+        if filename != 'generate_json_schemas.py':
+            with open(os.path.join(schemas_folder, filename), 'r') as f:
+                existing_schemas.append(json.load(f))
+
+    # Run the script
+    check_output(
+        subprocess.run(['python', 'generate_json_schemas.py'], cwd=schemas_folder, capture_output=True, text=True))
+
+    def sort_json(json):
```
ravi-mosaicml, line 101: nit: can you move this to a parent function? not a fan of functions in functions.

composer/json_schemas/generate_json_schemas.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generate JSON Schemas for all Hparam classes."""
+
+import os
+
+# Dataset Hparams
+from composer.datasets.ade20k_hparams import ADE20kDatasetHparams, StreamingADE20kHparams
```
ravi-mosaicml, line 9: Adding on to the comment below, I think we will only need to generate the JSONSchema for the experiment_hparams.py class, since all other hparam classes should be accessible within (that hparams class serves as the root, if you will).

composer/json_schemas/generate_json_schemas.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2022 MosaicML Composer authors
```
ravi-mosaicml, line 1: nit: can you add a shebang? `#!/usr/bin/env python3` should suffice

ravi-mosaicml (commented):
Overall this is awesome -- see comments.

In the JSONSchema generation (likely in the YAHP PR), one more area to make it more compact:

is it possible to specify properties and patternPropertiesin the same object? Trying to reduce some duplication -- e.g. this seems redundant

```
{
                    "additionalProperties": false,
                    "properties": {
                        "deeplabv3": {
                            "$ref": "#/$defs/DeepLabV3Hparams"
                        }
                    },
                    "type": "object"
                },
                {
                    "additionalProperties": false,
                    "patternProperties": {
                        "^deeplabv3\\+": {
                            "$ref": "#/$defs/DeepLabV3Hparams"
                        }
                    },
                    "type": "object"
                },
```

Before approving, would like to fix the docs build + bump the new yahp version and upgrade yahp in the dependencies.

composer/json_schemas/generate_json_schemas.py
mvpatel2000: is there a reason why? It looks like no other file has this?

composer/trainer/trainer_hparams.py
mvpatel2000: You're robbing me of my 25k lines committed though :(

mvpatel2000: re offline discussion, we actually do need all the files to enable autocomplete

mvpatel2000:
Note: All glue related yaml files are failing because of [this](https://mosaicml.atlassian.net/browse/CO-766).

mvpatel2000:
this is not really human readable sad. Discussing alternatives offline

mvpatel2000:
Closing this PR -- after a lot of offline discussion, we've decided not to insert JSON Schema into the docs. We'll instead do a skinny PR with just the test infra.

PR #1316, opened 2022-07-26
Logger Destination Refactor

eracah:
Closed. See https://github.com/mosaicml/composer/pull/1416

- [x] add new methods (`log_metrics`, `log_hyperparameters`) to `LoggerDestination`
- [x] add corresponding methods to `Logger`
- [x] Add corresponding methods to WandB logger
- [x] Add corresponding methods to TB logger
- [x] Add corresponding methods to File logger
- [x] Add corresponding methods to Progress Bar logger
- [x] Add corresponding methods to in memory logger
- [x] refactor all `Logger` calls in `Trainer` to use new methods
- [x] refactor all `Logger` calls in `Algorithms` to use new methods
- [x] refactor all `Logger` calls elsewhere to use new methods
- [x] remove `Logger.data()` method
- [x] remove all `data_fit` 
- [x] remove all `data_batch`
- [x] remove all `data_epoch`
- [x] Remove `log_data` methods everywhere
- [x] Test progress bar in notebook
- [ ] Test progress bar on cluster/k8's
- [x] Remove loglevel everywhere
- [ ] update docs with changes
- [ ] update notebooks
- [ ] update tests
- [x] add log_traces
- [ ] test with all loggers on on the cluster

### DIFF
--- a/composer/loggers/logger.py
+++ b/composer/loggers/logger.py
@@ -79,18 +79,18 @@ class Logger:
         self.destinations = ensure_tuple(destinations)
         self._state = state
 
-    def data(self, log_level: Union[str, int, LogLevel], data: Dict[str, Any]) -> None:
-        """Log data to the :attr:`destinations`.
+    def log_traces(self, traces: Dict[str, Any]):
+        for destination in self.destinations:
+            destination.log_traces(traces)
 
-        Args:
-            log_level (str | int | LogLevel): The log level, which can be a name, value, or instance of
-                :class:`LogLevel`.
-            data (Dict[str, Any]): The data to log.
-        """
-        log_level = LogLevel(log_level)
+    def log_hyperparameters(self, parameters: Dict[str, Any]):
+        for destination in self.destinations:
+            destination.log_hyperparameters(parameters)
+
+    def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
 
         for destination in self.destinations:
-            destination.log_data(self._state, log_level, data)
+            destination.log_metrics(metrics, step)
 
     def file_artifact(
         self,
@@ -127,18 +127,6 @@ class Logger:
                 overwrite=overwrite,
             )
 
-    def data_fit(self, data: Dict[str, Any]) -> None:
-        """Helper function for ``self.data(LogLevel.FIT, data)``."""
-        self.data(LogLevel.FIT, data)
-
-    def data_epoch(self, data: Dict[str, Any]) -> None:
-        """Helper function for ``self.data(LogLevel.EPOCH, data)``."""
-        self.data(LogLevel.EPOCH, data)
-
-    def data_batch(self, data: Dict[str, Any]) -> None:
-        """Helper function for ``self.data(LogLevel.BATCH, data)``."""
-        self.data(LogLevel.BATCH, data)
-
     def has_file_artifact_destination(self) -> bool:
         """Determines if the logger has a destination which supports logging file artifacts.
 
@@ -151,7 +139,6 @@ class Logger:
             if destination.can_log_file_artifacts():
                 return True
         return False
-
 
 def format_log_data_value(data: Any) -> str:
     """Recursively formats a given log data value into a string.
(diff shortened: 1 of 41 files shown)


eracah:
It's a little rough around the edges, but I figured I'd get a little feedback first.

composer/loggers/logger.py
```diff
@@ -79,18 +79,18 @@ def __init__(
         self.destinations = ensure_tuple(destinations)
         self._state = state
 
-    def data(self, log_level: Union[str, int, LogLevel], data: Dict[str, Any]) -> None:
-        """Log data to the :attr:`destinations`.
+    def log_traces(self, traces: Dict[str, Any]):
+        for destination in self.destinations:
+            destination.log_traces(traces)
```
mvpatel2000, line 84: Why does this have to be different from metrics?

composer/loggers/wandb_logger.py
```diff
@@ -98,11 +99,16 @@ def __init__(
...
-    def log_data(self, state: State, log_level: LogLevel, data: Dict[str, Any]):
+    def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
         import wandb
-        del log_level  # unused
-        if self._enabled:
-            wandb.log(data, step=int(state.timestamp.batch))
+        wandb.config.update(hyperparameters)
+
+    def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+        import wandb
```
mvpatel2000, line 107: Why are these `import wandb` in line?

composer/loggers/wandb_logger.py
```diff
@@ -137,6 +143,7 @@ def init(self, state: State, logger: Logger) -> None:
         if self._enabled:
             wandb.init(**self._init_kwargs)
             atexit.register(self._set_is_in_atexit)
+        print(wandb.run.dir)
```
mvpatel2000, line 146: flagging for deletion

composer/trainer/trainer.py
```diff
@@ -1537,6 +1528,12 @@ def _train_loop(self) -> None:
                     # `now` is actually in the past, but want to include the time it takes to perform this reduction
                     last_wct = now
 
+                    if self.train_metrics is not None:
+                        self._compute_and_log_metrics(
+                            dataloader_label='train',
+                            metrics=self.train_metrics,
+                        )
+
```
mvpatel2000, line 1536: Why is this moved? Is it because you want the older timestamp?

tests/callbacks/test_image_visualizer.py
```diff
@@ -18,7 +18,7 @@
 except ImportError:
     _WANDB_INSTALLED = False
 
-
+#@pytest.mark.skip
```
mvpatel2000, line 21: marking for deletion

mvpatel2000 (commented):
This looks great! Only major comment in initial pass: can you explain why `log_traces` has to be different? 

Also, it'd be awesome if you could include pictures of testing when you get to that point.

composer/loggers/logger.py
eracah: It's for logging [engine traces,](https://github.com/mosaicml/composer/blob/dev/composer/core/engine.py#L310) which we want in the `FileLogger`, but we don't want in `WandBLogger`, so `WandBLogger` doesn't implement `log_traces`

composer/loggers/wandb_logger.py
eracah: because it doesn't get imported anywhere else. We don't want it as a top-level import because the user may not have it installed

composer/trainer/trainer.py
eracah: yes exactly. train_metrics had an off by one error

tests/callbacks/test_image_visualizer.py
eracah: good catch! done

composer/loggers/wandb_logger.py
eracah: nice spot. done

eracah:
This PR is moved to https://github.com/mosaicml/composer/pull/1416

PR #1318, opened 2022-07-26
composer on tpus

florescl:
some wandb runs https://wandb.ai/mosaic-ml/tpu-perf?workspace=user-laura.
Design doc incoming.


composer/trainer/devices/common_tpu.py
```diff
@@ -0,0 +1,92 @@
...
+    futures = {executor.submit(fn, index=i): i for i in range(threads)}
+
+    results = {
+        futures[f]: f.result() for f in concurrent.futures.as_completed(futures)
+    }
+
+  return results
+
+
+def run_pjrt_multiprocess(fn: Callable, *args,
```
ravi-mosaicml, line 62: I presume the design doc will help clarify this, but I am confused on why we need to spawn threads / processes for TPU devices. Curious why the composer launcher script is insufficient?

composer/trainer/devices/device_tpu.py
```diff
@@ -0,0 +1,46 @@
...
+"""The TPU device used for training."""
+
+from __future__ import annotations
+
+import logging
+from contextlib import contextmanager
+from typing import Generator, TypeVar, Union
+
+import torch
+import torch_xla
```
ravi-mosaicml, line 13: This will need to be a conditional import.

composer/trainer/trainer.py
```diff
@@ -770,7 +777,7 @@ def __init__(
         if isinstance(precision, str):
             precision = Precision(precision)
 
-        _validate_precision(precision, self._device, deepspeed_enabled)
+        #_validate_precision(precision, self._device, deepspeed_enabled)
```
ravi-mosaicml: Can we keep this line? Feel free to fix the function to make it work on TPU.

```suggestion
        _validate_precision(precision, self._device, deepspeed_enabled)
```

composer/trainer/devices/common_tpu.py
```diff
@@ -0,0 +1,92 @@
...
+    fn: Function to run on all devices
+    args: args to pass to `fn`
+    kwargs: kwargs to pass to `fn`
+
+  Returns:
+    Dict of the form {process_rank: {thread_rank: return_value}}, where
+    return_value is the result of calling `fn`.
+  '''
+  if pjrt_device() == 'TPU':
+    processes = num_visible_tpu_chips()
```
ravi-mosaicml, line 76: Should this error if not being used on TPU?

composer/trainer/devices/common_tpu.py
```diff
@@ -0,0 +1,92 @@
...
+  else:
+    processes = 1
+
+  with concurrent.futures.ProcessPoolExecutor(
+      max_workers=processes) as executor:
+    futures = {
+        executor.submit(run_thread_per_device, i, processes,
```
ravi-mosaicml, line 83: Can you elaborate why we spawn multiple threads per process?

composer/trainer/trainer.py
```diff
@@ -1439,7 +1446,11 @@ def _train_loop(self) -> None:
         # surpressing GradScaler warnings as they are always created
         # self._use_grad_scaling() will raise a RuntimeError if grad scaling is not available when it is required
         warnings.filterwarnings(action='ignore', message='torch.cuda.amp.GradScaler')
-        self.state.scaler = ClosureGradScaler() if self._use_closures() else GradScaler()
+        if type(self._device) == DeviceTPU:
+            import torch_xla
+            self.state.scaler = torch_xla.amp.GradScaler
```
ravi-mosaicml, line 1449: Can we error if one is using a closure with TPU?

Also, do we need to instantiate it?

```suggestion
            self.state.scaler = torch_xla.amp.GradScaler()
```

composer/trainer/trainer.py
```diff
@@ -1652,15 +1663,21 @@ def _train_batch(self, use_grad_scaling: bool):
                             total_loss = self.state.scaler.step(
                                 optimizer, closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs))
                         else:
-                            total_loss = optimizer.step(
-                                closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs).item())
+                            if type(self._device) == DeviceTPU:
```
ravi-mosaicml, line 1664: nit: prefer `isinstance(..., xxx)` rather than `type(...) is xxx` as the former allows for subclassing 

```suggestion
                            if isinstance(self._device, DeviceTPU):
```

composer/trainer/trainer.py
```diff
@@ -1652,15 +1663,21 @@ def _train_batch(self, use_grad_scaling: bool):
...
+                                total_loss = xm.optimizer_step(optimizer)
```
ravi-mosaicml, line 1665: I'm not thrilled about device-specific code in the training loop. Thoughts about adding a `step(self, optimizer)` onto the `Device` class? Also not thrilled about this alternative, as then the code would be the same for cpu and gpu. IDK, thoughts?

composer/trainer/trainer.py
```diff
@@ -1652,15 +1663,21 @@ def _train_batch(self, use_grad_scaling: bool):
...
+                            else:
+                                total_loss = optimizer.step(
+                                    closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs).item())
                 else:
                     total_loss = self._train_microbatches(microbatches)
                     for optimizer in self.state.optimizers:
-                        if use_grad_scaling:
-                            self.state.scaler.step(optimizer)
+                        if type(self._device) == DeviceTPU:
+                            xm.optimizer_step(optimizer)
```
ravi-mosaicml, line 1673: How do we handle grad scaling if using tpu? I saw we create a XLA scaler earlier but I don't see that being stepped here?

composer/trainer/trainer.py
```diff
@@ -2124,6 +2142,7 @@ def _use_grad_scaling(self, precision: Union[str, Precision], scaler: Optional[G
         precision = Precision(precision)
         use_grad_scaling = precision == Precision.AMP
 
+        # will need fix for tpu
```
ravi-mosaicml, line 2147: (Comment to track this TODO...)

examples/run_composer_trainer_tpu.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Type
+
+from composer.loggers.logger import LogLevel
+from composer.trainer.trainer_hparams import TrainerHparams
+from composer.utils import dist
+import torch
+
+def train():
+
+    if len(sys.argv) == 1:
```
ravi-mosaicml, line 15: Curious why do we need a separate entrypoint for TPUs? I think with the composer launcher script we shouldn't need one?

ravi-mosaicml (commented):
Thanks for adding in TPU support. Leaving some feedback for now -- happy to take another look with the design doc in hand and once the PR is out of draft mode.

setup.py
```diff
@@ -78,7 +78,7 @@ def package_files(prefix: str, directory: str, extension: str):
     'torchmetrics>=0.7.0,<0.8',
     'torch_optimizer>=0.1.0,<0.2',
     'torchvision>=0.10.0',  # torchvision has strict pytorch requirements
-    'torch>=1.10,<2',
+    'torch>=1.9,<2',
```
hanlint, line 81: We recently removed support for torch 1.9, is there a reason we need to re-introduce here?

composer/yamls/tpu/resnet50.yaml
```diff
@@ -0,0 +1,40 @@
+callbacks:
```
hanlint, line 1: This `yaml` doesn't seem correct -- synthetic data, etc. I also don't think we need a device-specific YAML, right? We should just be able to use the `resnet50.yaml` and change the device to `tpu`. Prefer not to maintain duplicate yamls unless neccessary.

composer/trainer/devices/common_tpu.py
florescl: resolving these as the common_tpu.py file is removed and has been added by the xla team to torch_xla.experimental.

setup.py
florescl: the code was segfaulting on the interactive tpu instance and this is the change that fixed it. I can look into there might be wrong interactions and will revert this for now.

composer/trainer/trainer_hparams.py
```diff
@@ -428,8 +432,6 @@ def initialize_object(self) -> Trainer:
 
         # Device
         device = self.device
-        if device is None:
```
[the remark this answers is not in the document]
florescl: added it back as it was.

composer/yamls/tpu/resnet50.yaml
florescl: happy to remove this one and add it to the private repo.

examples/run_composer_trainer_tpu.py
florescl: I added some details in the design doc, if you have any other questions we can discuss offline.

composer/trainer/devices/device_tpu.py
```diff
@@ -0,0 +1,46 @@
...
+
+class DeviceTPU(Device):
+    """An extension of :class:`~composer.trainer.devices.device.Device` for TPUs
+    
+    When running on TPUVMs, you need to `export PJRT_DEVICE=TPU`.
+    More details.
+    """
+    dist_backend = "xla-tpu"
+    def __init__(self):
+        import torch_xla.core.xla_model as xm
```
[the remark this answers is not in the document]
florescl: fixed

composer/trainer/devices/device_tpu.py
florescl: fixed

composer/trainer/trainer.py
florescl: fixed

composer/trainer/trainer.py
florescl: fixed

composer/trainer/trainer.py
florescl: I think since some code is device specific in a few places + otherwise we'd have to right the same code for cpu, gpu, this is fine for now. As we add support for different types of hardware we can refactor.

composer/trainer/trainer_hparams.py
```diff
@@ -72,6 +73,7 @@
 def _initialize_dataloader(
     dataset_hparams: Optional[DatasetHparams],
     dataloader_label: str,
+    device: Optional[DeviceHparams],
```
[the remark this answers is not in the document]
florescl: it was leftover from a previous implementation, removed.

composer/trainer/trainer.py
```diff
@@ -103,8 +103,6 @@ def _get_training_metrics(model: ComposerModel):
 def _validate_precision(precision: Precision, device: Device, deepspeed_enabled: bool):
     if isinstance(device, DeviceCPU) and precision != Precision.FP32:
         raise ValueError(f'{precision} is not supproted for CPU training.')
-    if not deepspeed_enabled and precision == Precision.FP16:
-        raise ValueError('FP16 precision is only supported when training with DeepSpeed.')
```
bandish-shah, line 107: Why is this check removed?  Should we qualify it for TPU instead?

composer/trainer/trainer.py
```diff
@@ -302,7 +303,6 @@ class Trainer:
 
             This label is used to index the training metrics (if ``compute_training_metrics`` is True) in
             :attr:`.State.current_metrics`.
-
```
bandish-shah, line 305: [nit] should this newline be removed?

bandish-shah (commented):
Added a few additional comments/questions.

setup.py
```diff
@@ -151,6 +151,8 @@ def package_files(prefix: str, directory: str, extension: str):
     'scikit-learn>=1.0.1,<2',
 ]
 
+extra_deps['tpu'] = ['torch_xla==1.10.0']
```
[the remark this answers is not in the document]
florescl: will open a separate pr

setup.py
hanlint: OK, but does that mean for TPU, the user needs to install `torch==1.9`? and that newer versions of torch are not supported?

PR #1319, opened 2022-07-26
Avoid Deadlock with Autograd

mvpatel2000:
Currently, using `grad_accum=auto` has a deadlock risk if a subset of ranks OOM. In this scenario, some ranks start to retry the batch, while others are stuck in the gradient reduce after backprop.

In order to fix this, we use [DDP Communication Hooks](https://pytorch.org/docs/stable/ddp_comm_hooks.html) which let us control how gradients are reduced across ranks. We create a custom hook to insert a `monitored_barrier` before gradient reductions. If all ranks fail to enter the `monitored_barrier` in a specified `timeout` window, the barrier raises an error, implying at least one rank encountered a CUDA OOM and never reached the gradient reduction step. Other ranks which did not OOM can use this error to abort reduction and then raising a timeout error. This timeout error is caught at the same level as CUDA OOMs and silently ignored if at least one rank OOMed. In theory, this avoids deadlock by allowing ranks which did not OOM to still abort to top level if another rank OOMed.

Things I've "fixed" which may have better solutions:
- PyTorch C++ internals get unhappy and raise errors if you abort before a gradient reduction and then try to do another backward pass. To allow PyTorch to clean up internals, we do a no-op reduction (each rank keeps its gradients) before raising an error
- `monitored_barrier` requires Gloo backend. Additionally, if a `monitored_barrier` fails, PyTorch (or whatever underlying software it uses) shuts down the Gloo process group. So, if we want to catch the barrier fail and continue, for each backward pass, we need a new Gloo process group. We inject this through a state variable the DDP Comm Hook accepts.
- Gloo initialization fails by default on >8 GPUs because it looks at the wrong network device (but NCCL succeeds somehow). We need to explicitly set `GLOO_SOCKET_IFNAME` to the right device if we use Gloo on 9+ GPUs (ty Abhi for figuring this out). Abhi has a hacky script which fixes this, but we want a more permanent solution

**Things I can't figure out (where helpful insights are appreciated):**
- Currently, if a rank has less memory and is the bottleneck rank which OOMs, we manage to recover and start the next batch with a higher `grad_accum` value, but the rank which OOMed gets stuck on backprop.*

*Repro steps: 2 GPUs with mcli interactive. Run `torch.zeros([2000, 2000, 1100]).to('cuda:1')` to eat up memory on rank 1. Run a large model (GPT 3, 1 billion parameters). Both ranks hit CUDA OOMs until `grad_accum=16`. Rank 0 is able to work through the batch but rank 1 OOMs. Rank 0 hits the `monitored_barrier`, raises an error, and both ranks restart with `grad_accum=32`. Rank 0 works through the batch but rank 1 hangs on `final_loss.backward(create_graph=self._backwards_create_graph)` -- it reaches the line before that but does not reach the first line of the DDP Comm hook. Rank 0 goes through the batch, fails on the `monitored_barrier` because rank 1 is stuck on backprop on the first minibatch (rank 0 thinks rank 1 hit OOM and hence never reached `monitored_barrier`).

If anyone has insights on what's going on, feel free to chime in! Otherwise, I'll be crawling through PyTorch source code (again).

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -342,8 +342,8 @@ stage('Build') {
         if (isComposerImage && buildArgs['COMPOSER_INSTALL_COMMAND'] == 'mosaicml[all]GIT_COMMIT') {
             // If this is a composer image and it is the 'GIT_COMMIT' entry in the matrix
             // then build and publish the image on the staging repo
-            if (!isMergeCommit) {
-                // Skip composer images except on merge commits -- it's slow!
+            if (!shouldPushToDestinations) {
+                // Skip composer images except on merge commits on the dev branch -- it's slow!
                 return
             }
             String stagingTagName = "mosaicml/composer_staging:${gitCommitHash[0..6]}"
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -321,11 +321,14 @@ class Engine():
         Returns:
             Sequence[Algorithm]: Modified sequence of algorithms.
         """
-        from composer.algorithms import CutMix, MixUp, SelectiveBackprop, StochasticDepth
+        from composer.algorithms import CutMix, FusedLayerNorm, MixUp, SelectiveBackprop, StochasticDepth
 
         # Move selective backprop to the beginning while maintaining order of other algorithms
         algorithms = sorted(algorithms_to_run,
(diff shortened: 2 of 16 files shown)


mvpatel2000:
Closing as won’t do. Looks like cache clear / fragmentation fixed most of the issues. We’ll still see problems if you artificially eat up mem on a particular rank, but it’s not worth supporting this.

PR #1320, opened 2022-07-27
[xs] Fix pytest test filtering; Bump the minimum pytorch version to 1.10

ravi-mosaicml:
1. Fix pytest filtering in `get_algs_with_marks` which caused the entire test suite to be skipped if `transformers` was not installed. Now, only tests that depend on `transformers` are skipped.
2. Bump the minimum pytorch version to 1.10. We stopped testing on pytorch 1.9 as part of https://github.com/mosaicml/composer/pull/1247. Updated the requirements in setup.py / meta.yaml, and removed the runtime checks if we are running on torch 1.9.

Closes https://mosaicml.atlassian.net/browse/CO-709


bandish-shah (approved):
LGTM

PR #1321, opened 2022-07-27
Enable gloo backend

mvpatel2000:
Originally thought gloo group initialization selectively failed on a subset of ranks when using >8 GPUs. It actually fails on all of them, so we can turn on the monitored barrier without nccl syncing gloo initialization success

Also fixes a bug in computing batch time, where it computed time for OOMed iterations


mvpatel2000:
Transient test failures -- will force reruns till it passes once its reviewed

abhi-mosaic:
Hey @mvpatel2000 could you clarify a bit here what will happen when we are in a multi-node setting? Will GLOO fail to initialize and then monitored_barrier will be a no-op? I'm a bit confused

mvpatel2000:
> Hey @mvpatel2000 could you clarify a bit here what will happen when we are in a multi-node setting? Will GLOO fail to initialize and then monitored_barrier will be a no-op? I'm a bit confused

Yea sorry, I should have written something more descriptive in PR header. The latter is correct -- if Gloo fails to initialize, monitored_barrier will become a no-op. If we're not in a multi-node setting, it will be enabled.

Updated comments to be more clear

abhi-mosaic (approved):
LGTM! Thanks for the batch time fix as well.

ravi-mosaicml:
> Transient test failures -- will force reruns till it passes once its reviewed

Pytest timeout is gone :) a bit concerned we still have transient failures...

mvpatel2000:
> > Transient test failures -- will force reruns till it passes once its reviewed
> 
> Pytest timeout is gone :) a bit concerned we still have transient failures...

lol apparently its not transient 🙃 I just assumed it was

PR #1322, opened 2022-07-27
Integration Tests for Object Stores and Loggers

ravi-mosaicml:
1. Add remote object store tests
2. Add remote logger tests

Coming along for the ride: some typing fixes

Closes https://mosaicml.atlassian.net/browse/CO-768


bandish-shah (commented):
Is this ready to review?  Started looking and then saw the failed tests, looks like this might need some more work.

ravi-mosaicml:
> Is this ready to review? Started looking and then saw the failed tests, looks like this might need some more work.

Yes, the core logic is here. Fixed (most, if not all) of the failing tests.

tests/callbacks/test_callbacks.py
```diff
@@ -107,7 +107,7 @@ def test_multiple_init_and_close(self, cb_cls: Type[Callback], dummy_state: Stat
 
 
 @pytest.mark.parametrize('cb_cls', get_cbs_and_marks(callbacks=True, loggers=True, profilers=True))
-@pytest.mark.parametrize('grad_accum', [1, 2])
+@pytest.mark.parametrize('grad_accum,remote', [(1, False), (2, False), pytest.param(1, True, marks=pytest.mark.remote)])
```
dskhudia, line 110: remote is not used in this class?

mvpatel2000:
Discussed offline about `remote`, it's used to toggle `pytest.mark.remote`. Requested comments explaining + changing to `_remote` to indicate its unused variable

mvpatel2000 (commented):
No-op review so I can get re-requested for review when ready

composer/core/state.py
```diff
@@ -10,6 +10,7 @@
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Sequence, Union, cast
 
 import torch
+import torch.nn
```
mvpatel2000, line 13: why is there an import change but no changes in the file?

mvpatel2000 (approved):
LGTM besides previous comment. Approving to unblock

composer/core/state.py
ravi-mosaicml: Pyright in vscode doesn't like implicit imports of submodules for some reason...

ravi-mosaicml: (This is coming along for the ride...)

ravi-mosaicml:
> Discussed offline about `remote`, it's used to toggle `pytest.mark.remote`. Requested comments explaining + changing to `_remote` to indicate its unused variable

Added in a comment + renamed the variable.

tests/callbacks/test_callbacks.py
ravi-mosaicml: This is a bit confusing, but it is used to parameterize the test markers. The value of the variable is unused. Added a comment to explain this.

PR #1323, opened 2022-07-28
Add export for inference callback

nik-mosaic:
Closes https://mosaicml.atlassian.net/browse/CO-702


composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,68 @@
...
+        self.save_object_store = save_object_store
+        self.transforms = transforms
+
+    def fit_end(self, state: State, logger: Logger):
+        del logger
+        self.export_model(state)
+
+    def export_model(self, state: State):
+        if state.dataloader is not None:
+            sample_input = next(iter(state.dataloader))
```
dskhudia, line 60: By the time, this gets called dataloader iter will have already reached `StopIteration` I think we should make a copy of the inputs when training loop is running and store it on state during _train_loop call somewhere [here](https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L1476).

dskhudia:
Overall looks great. Thanks for getting it done so quickly. Just one issue with sample inputs and some minor nits.

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,68 @@
...
+
+__all__ = ['ExportForInference']
+
+
+class ExportForInference(Callback):  # noqa: D101
+    """Callback to export model for inference.
+
+    Args:
+        save_format (Union[str, ExportFormat]):  Format to export to. Either ``"torchscript"`` or ``"onnx"``.
+        out_filename (str): The path for storing the exported model. It can be a path to a file on the local disk,
```
dskhudia, line 29: nit: Change this parameter name to be the same as in `export_for_inference`, i.e., `save_path`.

dskhudia:
Can you also test locally to make sure `save_object_store` works as intended? Here is an example how I used it to test `export_for_inference` functional api. 
```
from composer.utils.object_store import S3ObjectStore

object_st = S3ObjectStore(
    bucket="mosaicml-internal-checkpoints-bert",
)

export_for_inference(
    model=model,
    target_format="torchscript",
    store_path="daya_resnet_test/model.pt",
    store_object_store=object_st,
)
```

tests/callbacks/test_inference.py
```diff
@@ -0,0 +1,101 @@
...
+        trainer.fit()
+
+        model.eval()
+        orig_out = model(sample_input)
+
+        loaded_model = torch.jit.load(save_path)
+        loaded_model.eval()
+        loaded_model_out = loaded_model(sample_input)
+
+        assert torch.allclose(
```
hanlint, line 51: could just use `torch.testing.assert_close` here to give an informative output. (https://pytorch.org/docs/stable/testing.html)

tests/callbacks/test_inference.py
```diff
@@ -0,0 +1,101 @@
...
+    'model_cls, sample_input',
+    [
+        (partial(composer_resnet, 'resnet18'), (torch.rand(1, 3, 224, 224), torch.randint(10, (1,)))),
+    ],
+)
+def test_inference_callback_onnx(model_cls, sample_input):
+    pytest.importorskip('onnx')
+    pytest.importorskip('onnxruntime')
+    import onnx  # type: ignore
+    import onnxruntime as ort  # type: ignore
```
hanlint, line 66: Curious why these type ignores are needed?

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,68 @@
...
+from composer.loggers import Logger
+from composer.utils.inference import ExportFormat, export_for_inference
+from composer.utils.object_store import ObjectStore
+
+log = logging.getLogger(__name__)
+
+__all__ = ['ExportForInference']
+
+
+class ExportForInference(Callback):  # noqa: D101
```
hanlint, line 24: Since we also have the `export_for_inference` method, this might be worth being a bit more verbose, e.g. `ExportForInferenceCallback`

composer/callbacks/__init__.py
```diff
@@ -24,6 +25,7 @@
     'CheckpointSaver',
     'MLPerfCallback',
     'EarlyStopper',
+    'ExportForInference',
```
hanlint, line 28: Could you add this to https://github.com/mosaicml/composer/blob/dev/docs/source/trainer/callbacks.rst as well?

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,68 @@
...
+        transforms (Union[Callable, Sequence[Callable]], optional): transformations (usually optimizations) that should
+            be applied to the model. Each should be a callable that takes a model and returns a modified model.
+    """
+
+    def __init__(
+        self,
+        save_format: Union[str, ExportFormat],
+        out_filename: str,
+        save_object_store: Optional[ObjectStore] = None,
+        transforms: Optional[Union[Callable[[nn.Module], nn.Module], Sequence[Callable[[nn.Module],
```
hanlint, line 46: Could we simplify the typing here with an alias? e.g., up top to 
```
Transform = Callable[[nn.Module],nn.Module]
```

then

```
transforms: Union[Transform, Sequence[Transform]]
```

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,68 @@
...
+        if state.dataloader is not None:
+            sample_input = next(iter(state.dataloader))
+            export_for_inference(model=state.model,
+                                 save_format=self.save_format,
+                                 save_path=self.out_filename,
+                                 save_object_store=self.save_object_store,
+                                 sample_input=(sample_input,),
+                                 transforms=self.transforms)
+        else:
+            raise RuntimeError('Exporing model failed because state has no dataloader.')
```
hanlint, line 68:
```suggestion
            raise RuntimeError('Exporting model failed because state has no dataloader.')
```

composer/callbacks/export_for_inference.py
hanlint: Might be better for the user to provide the sample_input to the Callback.

For example, if we take progressive resizing, then the exported models over training will have different input resolutions.

If we do want to copy the inputs, `state.batch` should already be populated, or you can have the callback run on `before_forward`

tests/callbacks/test_inference.py
```diff
@@ -0,0 +1,101 @@
...
+            None,
+            {'input': sample_input[0].numpy()},
+        )
+
+        torch.testing.assert_close(
+            orig_out.detach().numpy(),
+            loaded_model_out[0],
+            rtol=1e-4,  # lower tolerance for ONNX
+            atol=1e-3,  # lower tolerance for ONNX
+            msg=f'mismatch in the original and exported for inference model outputs with {save_format}')
```
hanlint, line 101: msg is probably not needed, as pytest errors will already have descriptions attached.

hanlint (commented):
Looks great, just a few minor stylistic nits!

composer/callbacks/export_for_inference.py
nik-mosaic: We can definitely do that. However, this still won't fix the error I'm currently seeing when running test_callback_hparams_is_constructable
`TypeError: A union of multiple classes without a primitive or Enum are not supported`

composer/callbacks/export_for_inference.py
nik-mosaic: Changed!

composer/callbacks/__init__.py
nik-mosaic: Added.

tests/callbacks/test_inference.py
nik-mosaic: Changed to assert_close so both the torchscrip and onnx test match.

composer/callbacks/export_for_inference.py
nik-mosaic: Changed!

composer/callbacks/export_for_inference.py
hanlint: Ah must be a `yahp` limitation. Can you try `Union[Callable[], Callable[], None]`?

nik-mosaic: Unfortunately, yahp defines [primitive types](https://github.com/mosaicml/yahp/blob/bbe1a826e7c492722199e303bc1868d47664a4dd/yahp/utils/type_helpers.py#L25) as bool, float, ints, or strings. So None does not work. If there's an ignore flag we can use to get yahp to skip this, that could work.

nik-mosaic:
>

composer/callbacks/export_for_inference.py
dskhudia: How about we add `sample_input` as an argument and if it's not provided we try to use from `state.batch`?

composer/callbacks/export_for_inference.py
nik-mosaic: Resolved by using Optional[Sequence[Transform]]

dskhudia (approved):

composer/callbacks/export_for_inference.py
nik-mosaic: Updated to populate sample_input from state.batch at after_dataloader event.

ravi-mosaicml (approved):
LGTM; see comments.

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,84 @@
...
+        self.save_format = save_format
+        self.save_path = save_path
+        self.save_object_store = save_object_store
+        self.sample_input = sample_input
+        self.transforms = transforms
+
+    def after_dataloader(self, state: State, logger: Logger) -> None:
+        del logger
+        if self.sample_input is None:
+            self.sample_input = state.batch
```
ravi-mosaicml, line 72: If the batch is modified in place, then algs like progressive resizing could still be caught in the sample batch here. Should this be a deepcopy? That said, a deepcopy would have memory (and by extension, throughput) overhead, so not thrilled about that either. This concern could be mitigated by doing the deepcopy only once (e.g on the first batch)

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,84 @@
...
+        del logger
+        self.export_model(state)
+
+    def export_model(self, state: State):
+        export_for_inference(model=state.model,
+                             save_format=self.save_format,
+                             save_path=self.save_path,
+                             save_object_store=self.save_object_store,
+                             sample_input=(self.sample_input,),
+                             transforms=self.transforms)
```
ravi-mosaicml, line 84: nit: newline

composer/callbacks/export_for_inference.py
```diff
@@ -0,0 +1,84 @@
...
+            (default: ``None``)
+        transforms (Union[Callable, Sequence[Callable], None]): transformations (usually optimizations) that should
+            be applied to the model. Each should be a callable that takes a model and returns a modified model.
+    """
+
+    def __init__(
+        self,
+        save_format: Union[str, ExportFormat],
+        save_path: str,
+        save_object_store: Optional[ObjectStore] = None,
```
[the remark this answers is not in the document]
dskhudia: We talked about this and we can share this functionality with Checkpoint API. Need to refactor that code and reuse here. Current plan is to leave it for a future PR.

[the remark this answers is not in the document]
dskhudia: https://mosaicml.atlassian.net/browse/CO-805

composer/callbacks/export_for_inference.py
nik-mosaic: Since sample_input will only be None once, adding deepcopy() to line 72 won't be a perf issue.

tests/callbacks/test_inference.py
```diff
@@ -0,0 +1,99 @@
...
+from tests.common.datasets import RandomImageDataset
+
+
+@pytest.mark.parametrize(
+    'model_cls, sample_input',
+    [
+        (partial(composer_resnet, 'resnet18'), (torch.rand(1, 3, 224, 224), torch.randint(10, (1,)))),
+    ],
+)
+def test_inference_callback_torchscript(model_cls, sample_input):
```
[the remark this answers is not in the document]
nik-mosaic: I've updated this PR to mock the `export_for_inference` method and ensure that it's called exactly once with the expected arguments.

PR #1326, opened 2022-07-28
Fix WandB load from checkpoint

abhi-mosaic:
This PR fixes an issue with loading from WandB artifact checkpoints. Previously we had fixed the auto-resume pathway but not a direct load from checkpoint. 

We should probably set up a design review soon @eracah as the WandB Artifact storage is more tricky to deal with than our other object stores.


growlix (approved):
lgtm. tyvm for unblocking avalara

PR #1328, opened 2022-07-29
Fix a race condition in the object store logger

ravi-mosaicml:
The Object Store Logger had a race condition in `post_close` where the workers could be shut down before all files were enqueued, due to the same flag being used multiple times. Specifically, if all workers saw an empty queue after the shutdown flag was set while the enqueueing thread was sleeping, then the worker pool would shut down, even if there were objects left to enqueue.

Fixed by adding a second flag to trigger shutdown of the enquing thread that is seperate from the flag used to shutdown the worker pool.

Did not add a new test case to verify the "fix", as it is impossible to deterministically trigger this race condition via the public API.

Closes https://mosaicml.atlassian.net/browse/CO-638


milocress (approved):

composer/loggers/object_store_logger.py
```diff
@@ -402,23 +415,32 @@ def post_close(self):
...
 
+        if len(self._enqueued_objects) > 0 or len(self._logged_objects) > 0:
+            # Warn on all objects that have not been uploaded
+            object_names = list(self._enqueued_objects)
+            object_names.extend(self._logged_objects.keys())
+            warnings.warn(
+                RuntimeWarning('The following objects may not have been uploaded, likely due to a worker crash: ' +
+                               ', '.join(self._enqueued_objects)))
+
+        # Reset all variables
```
jbloxham, line 436: Why is this necessary?

jbloxham (approved):
This looks good to me. Can't think of any race conditions exposed here.

composer/loggers/object_store_logger.py
ravi-mosaicml: This is needed if you are creating a new trainer with the same callback instance.

PR #1333, opened 2022-08-01
Add save fine-tune model to HuggingFace example

lupesko:
HuggingFace example did not include a demonstration of saving the fine-tuned PyTorch model parameters.
This commit addresses this and adds saving of the model parameters.
Tested on Colab with `%pip install "mosaicml[nlp] @ git+https://github.com/mosaicml/composer@dev"`, ran successfully.

### DIFF
--- a/examples/huggingface_models.ipynb
+++ b/examples/huggingface_models.ipynb
@@ -286,6 +286,29 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Save Fine-Tuned Model"
+   ]
+  },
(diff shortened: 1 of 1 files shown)


moinnadeem (approved):
LGTM, but do we want to support easily loading this back into the HuggingFace framework? If so, then we might want to provide an example of saving this into a directory with a `config.json` as well.

lupesko:
@moinnadeem - this is a good idea, we can consider this as a follow up and I will file a task for this.

PR #1347, opened 2022-08-02
Add Precision Contexts

mvpatel2000:
As title. Was causing problems with methods which require certain precisions.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1952,7 +1952,8 @@ class Trainer:
                 self.engine.run_event(Event.PREDICT_BATCH_START)
 
                 self.engine.run_event(Event.PREDICT_BEFORE_FORWARD)
-                self.state.outputs = self.state.model(self.state.batch)
+                with get_precision_context(self.state.precision):
+                    self.state.outputs = self.state.model(self.state.batch)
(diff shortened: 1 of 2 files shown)


dskhudia:
LGTM! Could you add a test for it?

ravi-mosaicml (commented):
Code LGTM, though will defer to @dskhudia  or @abhi-mosaic for approval -- they're the precision masters. Agree having a test case would be nice, especially if there was an error beforehand. Take a look at the existing precision test case which measure memory.

mvpatel2000:
Tests on dev 
image

dskhudia (approved):

mvpatel2000:
@dskhudia looks like tests are failing in CI because the percentage memory usage w precision is only ~85%. On an interactive instance, I'm seeing ~65%. I bumped up thresholds from 70% to 90% for predict/eval, but I don't want to merge unless you think this seems fine with you.

dskhudia:
> @dskhudia looks like tests are failing in CI because the percentage memory usage w precision is only ~85%. On an interactive instance, I'm seeing ~65%. I bumped up thresholds from 70% to 90% for predict/eval, but I don't want to merge unless you think this seems fine with you.

Is there a difference between the number of GPUs used on CI vs interactive instance?

Issue #1331, opened 2022-07-30
Assorted Issues


vedantroy:
** Environment **
- OS: Ubuntu 22.04 LTS
- GPU:
```
  *-display                 
       description: VGA compatible controller
       product: GP102 [GeForce GTX 1080 Ti]
       vendor: NVIDIA Corporation
       physical id: 0
       bus info: pci@0000:01:00.0
       version: a1
       width: 64 bits
       clock: 33MHz
       capabilities: pm msi pciexpress vga_controller bus_master cap_list rom
       configuration: driver=nvidia latency=0
       resources: irq:58 memory:fa000000-faffffff memory:c0000000-cfffffff memory:d0000000-d1ffffff ioport:e000(size=128) memory:c0000-dffff
  *-graphics
       product: EFI VGA
       physical id: 2
       logical name: /dev/fb0
       capabilities: fb
       configuration: depth=32 resolution=1024,768
```
- Cuda: 11.7
- Composer: 0.8.2

1. `loss` should specify micro_batch instead of batch
2. Cannot log multiple losses, allow me to return dictionary of losses
3. Cannot log things at batch level (not micro-batch level) inside of `loss` method
4. (Bug) `grad_accum` fails with CUDA OOM even though `batch_size=1` w/ no `grad_accum` works
5. (Bug) Nothing is printed indicated that composer is restarting the forward method when `grad_accum="auto"` is set to `True`

hanlint:
Thanks @vedantroy for reporting these, we'll take a look at each carefully.

For #5 above, you should see something in the logs (e.g.):
```
INFO:composer.trainer.trainer:CUDA out of memory detected.
Gradient Accumulation increased from 1 -> 2, and the batch
will be retrained.
```

Let us know if that's not the case!

vedantroy:
> Thanks @vedantroy for reporting these, we'll take a look at each carefully.
> 
> For #5 above, you should see something in the logs (e.g.):
> 
> ```
> INFO:composer.trainer.trainer:CUDA out of memory detected.
> Gradient Accumulation increased from 1 -> 2, and the batch
> will be retrained.
> ```
> 
> Let us know if that's not the case!

I do not see that 😔, hence why I added it to the list of issues.

hanlint:
Sounds good, tagging @mvpatel2000 for auto grad accum. 

[...]

PR #1349, opened 2022-08-02
Refactor multiple losses to support dictionaries and fix discrepancies

Landanjs:
Addresses #1331 request to be able to train on a dictionary of losses.

The issue came from `self.state.loss = cast(torch.Tensor, self.state.scaler.scale(self.state.loss))`. If `GradScaler.scale` takes in a dictionary, it will output a map object that cannot be properly parsed by `ensure_tuple`.

A quick fix would be to use `map_collection(self.state.loss, self.state.scaler.scale)`, but @abhi-mosaic recommended adding the multiple losses before the loss scaling for numeric stability. I noticed there were some inconsistencies with the order of operation when deepspeed is turned off/on, so I refactored to make it consistent. Specifically:

Without deepspeed:
1. Scale by grad accum
2. Add to total loss
3. `AFTER_LOSS` and `BEFORE_BACKWARD` events
4. GradScaler
5. Sum losses
6. Backward

With deepspeed:
1. `AFTER_LOSS` and `BEFORE_BACKWARD` events
2. GradScaler
3. Scale by grad accum (within deepspeed backwards)
4. Backward
5. Add to total loss

So `self.state.loss` may or may not be scaled by grad_accum at `AFTER_LOSS` and `BEFORE_BACKWARD` events depending on deepspeed. Also, loss isn't summed before deepspeed, so not sure if multiple losses would work with deepspeed.

New order:
1. `AFTER_LOSS` and `BEFORE_BACKWARD` events
2. Sum losses
3. Add grad accum scaled loss to total loss (does not inplace change loss)
4. GradScaler
5. Scale loss by grad accum (either manually or through deepspeed)
6. Backward

Let me know if none of this makes sense!

TODO:

- [x] Resolve comments
- [x] Test loss dictionary
- [x] Test loss grad accum scaling
- [x] BERT Convergence
- [ ] DeepLabv3+ Convergence

Edit:
- I also changed the DeepLabv3+ loss to output dictionaries instead of tuples

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1795,39 +1795,30 @@ class Trainer:
             with get_precision_context(self.state.precision):
                 self.state.loss = self._original_model.loss(self.state.outputs, self.state.batch)
 
-            # We always want to scale loss by the grad_accum before the backwards pass and
-            # also for sake of metrics. Complicating matters, the DeepSpeed engine does its
-            # own scaling when we call `.backward`, but this isn't in place so we still need
-            # to scale for sake of metrics after the `.backward` call.
-
-            # Loss is added to losses with clone to not scale the loss for the step printout
-            # Likely need to look into the performance impact
-            if not self.deepspeed_enabled:
-                for loss in ensure_tuple(self.state.loss):
-                    loss.mul_(microbatch_num_samples / current_batch_size)
-                    total_loss += loss.detach().clone()
-
             assert self.state.loss is not None
             self.engine.run_event(Event.AFTER_LOSS)
 
             # backward
             self.engine.run_event(Event.BEFORE_BACKWARD)
 
+            # Sum individual losses
+            microbatch_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
+            for loss in ensure_tuple(self.state.loss):
+                microbatch_loss += loss
+
+            # Loss used for logging, scaled by grad_accum for correctly calculating metrics
+            total_loss += microbatch_loss.detach().clone().mul(microbatch_num_samples / current_batch_size)
+
             if use_grad_scaling:
-                self.state.loss = cast(torch.Tensor, self.state.scaler.scale(self.state.loss))
+                microbatch_loss = cast(torch.Tensor, self.state.scaler.scale(microbatch_loss))
 
             if self.deepspeed_enabled:
-                self.state.deepspeed_model.backward(self.state.loss)
+                self.state.deepspeed_model.backward(microbatch_loss)
 
-                # This is the same loss scaling and reporting we skipped earlier.
-                for loss in ensure_tuple(self.state.loss):
-                    loss.mul_(microbatch_num_samples / current_batch_size)
-                    total_loss += loss.detach().clone()
             else:
-                final_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
-                for loss in ensure_tuple(self.state.loss):
-                    final_loss += loss
-                final_loss.backward(create_graph=self._backwards_create_graph)
+                # Scale loss by grad_accum before backwards pass
+                microbatch_loss.mul_(microbatch_num_samples / current_batch_size)
+                microbatch_loss.backward(create_graph=self._backwards_create_graph)
 
             self.engine.run_event(Event.AFTER_BACKWARD)
 
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1802,7 +1802,9 @@ class Trainer:
             self.engine.run_event(Event.BEFORE_BACKWARD)
 
             # Sum individual losses
-            microbatch_loss = torch.stack(ensure_tuple(self.state.loss)).sum()
+            microbatch_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
+            for loss in ensure_tuple(self.state.loss):
+                microbatch_loss.add_(loss.mean())
 
             # Loss used for logging, scaled by grad_accum for correctly calculating metrics
             total_loss += microbatch_loss.detach().clone() * (microbatch_num_samples / current_batch_size)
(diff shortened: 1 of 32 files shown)


composer/trainer/trainer.py
```diff
@@ -1795,39 +1795,30 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
             assert self.state.loss is not None
             self.engine.run_event(Event.AFTER_LOSS)
 
             # backward
             self.engine.run_event(Event.BEFORE_BACKWARD)
 
+            # Sum individual losses
+            microbatch_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
+            for loss in ensure_tuple(self.state.loss):
+                microbatch_loss += loss
```
abhi-mosaic, line 1807: I always get a bit uneasy with operations like `+=` with autograd... could we do something like 
```python
microbatch_loss = torch.stack(ensure_tuple(self.state.loss)).sum()
```

composer/trainer/trainer.py
```diff
@@ -1795,39 +1795,30 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
+
+            # Loss used for logging, scaled by grad_accum for correctly calculating metrics
+            total_loss += microbatch_loss.detach().clone().mul(microbatch_num_samples / current_batch_size)
```
abhi-mosaic, line 1810: This scaling should be applied to the actual loss tensor that is being back propagated (if not using deepspeed), not just the logged value. This helps maintain consistent numerics for batch-averaged loss functions (most common), because if a whole batch is being backpropagated there would be a `(1 / batch_size)` applied to each samples activation gradient.  If using microbatches, there is only a `(1 / microbatch_size)` factor, so this scaling corrects that difference in the activation gradients.

composer/trainer/trainer.py
Landanjs: Sounds good to me!

composer/trainer/trainer.py
abhi-mosaic, line 1810: I'm not 100% sure if this works, but I think there might be a more PyTorch way to do this now:

```python
with torch.no_grad():
  total_loss += microbatch_loss * (microbatch_num_samples / current_batch_size)
```

composer/trainer/trainer.py
```diff
@@ -1795,39 +1795,30 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
-                # This is the same loss scaling and reporting we skipped earlier.
-                for loss in ensure_tuple(self.state.loss):
-                    loss.mul_(microbatch_num_samples / current_batch_size)
-                    total_loss += loss.detach().clone()
             else:
-                final_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
-                for loss in ensure_tuple(self.state.loss):
-                    final_loss += loss
-                final_loss.backward(create_graph=self._backwards_create_graph)
+                # Scale loss by grad_accum before backwards pass
```
abhi-mosaic, line 1819: Nitpick, but this value might not be exactly `grad_accum`, since we can have incomplete microbatches. Maybe something like "Scale loss to maintain activation gradient numerics... "?

abhi-mosaic (commented):
This tentatively looks good to me, made a few minor comments. I think for something as core as `loss.backward()` though I'd like to do a couple convergence runs (maybe ADE20k and BERT) before approving though, what do you think @Landanjs ?

abhi-mosaic:
Also almost forgot but super glad that we can pass loss dictionaries now! I can see a lot of VAE and diffusion style models benefitting from this.

composer/trainer/trainer.py
Landanjs: Hmm, this would allow `self.state.loss` to contain sequences of non-scalar tensors. Do you think this would be a desirable behavior or would it be best to check that each tensor is a scalar to avoid bugs?

Landanjs:
Yeah, I definitely want to do convergence runs as well, but there is a big compute crunch that would delay getting results by 1-2 days. That's probably for the best though

composer/trainer/trainer.py
Landanjs: Agreed

Landanjs: Let me know if the new comment is enough

composer/trainer/trainer.py
ravi-mosaicml: As an alternative, you can also do the inplace operator:

```suggestion
            microbatch_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
            for loss in ensure_tuple(self.state.loss):
                microbatch_loss.add_(loss)
```

My preference is for whatever is more generalizable

ravi-mosaicml (commented):
Looks fine to me, but can we add a test case to verify this dictionaries of losses work as expected? In addition do we already have test cases to ensure correct loss scaling with grad accum? If not, can we add those in?

abhi-mosaic (approved):
I think this looks pretty good to me now, just want to see some convergence curves before merging. Maybe running ADE20k with grad_accum=1 and grad_accum=4? And also with the new combo loss?

Landanjs:
@ravi-mosaicml I believe `TestTrainerEquivalence.test_grad_accum` checks if the model with/without grad accum is the same, so that should make sure grad accum is working as expected. Let me know if this isn't right

@ravi-mosaicml I added a test to check that a model trained with an equivalent dict loss results in the same model. Let me know if it should be reorganized though :).

I'll do quick convergence runs next!

ravi-mosaicml (approved):
LGTM! Thanks for adding in the test

Landanjs:
BERT train loss after 1000 steps:
image

Landanjs:
[Deeplabv3+ training loss](https://wandb.ai/mosaic-ml/landan-debug/groups/dlv3p-test/workspace?workspace=user-landanjs) (a lot noisier so not as clear): 
image

PR #1350, opened 2022-08-03
Fix scheduler edge cases

### LINKED ISSUE #1077: Warmup schedulers crash if `t_warmup` == `max_duration`
## To reproduce

1. Use a scheduler like `cosine_decay_with_warmup` and set `t_warmup` == `max_duration` == `100ba`. 
2. Launch a training run, it will run until the last step and then crash with a `ZeroDivisionError`:

```
  File "/root/composer/examples/run_composer_trainer.py", line 67, in <module>
    main()
  File "/root/composer/examples/run_composer_trainer.py", line 63, in main
    trainer.fit()
  File "/root/composer/composer/trainer/trainer.py", line 1289, in fit
    self._train_loop()
  File "/root/composer/composer/trainer/trainer.py", line 1497, in _train_loop
    scheduler.step()
  File "/usr/local/lib/python3.9/dist-packages/torch/optim/lr_scheduler.py", line 154, in step
    values = self.get_lr()
  File "/usr/local/lib/python3.9/dist-packages/torch/optim/lr_scheduler.py", line 252, in get_lr
    return [base_lr * lmbda(self.last_epoch)
  File "/usr/local/lib/python3.9/dist-packages/torch/optim/lr_scheduler.py", line 252, in <listcomp>
    return [base_lr * lmbda(self.last_epoch)
  File "/root/composer/composer/optim/scheduler.py", line 190, in scheduler_fn
    return scheduler(state, ssr)
  File "/root/composer/composer/optim/scheduler.py", line 697, in __call__
    frac_of_total = ((current_time - t_warmup) / (t_max - t_warmup)).value
  File "/root/composer/composer/core/time.py", line 308, in __truediv__
    return Time(self.value / other.value, TimeUnit.DURATION)
ZeroDivisionError: division by zero
```

## Expected behavior

We should either allow this setting, and just not attempt to increment the schedule past the warmup... or we should catch this edge case and raise a `ValueError` out on the scheduler's `__init__`.

abhi-mosaic:
This PR should close #1077 and add extra tests to catch the edge case when `t_warmup` == `t_max`.

### DIFF
--- a/composer/optim/scheduler.py
+++ b/composer/optim/scheduler.py
@@ -692,7 +692,7 @@ class LinearWithWarmupScheduler(ComposerScheduler):
 
         t_max = _convert_time(self.t_max, state, ssr=ssr)
         current_time = state.timestamp.get(t_warmup.unit)
-        frac_of_total = ((current_time - t_warmup) / (t_max - t_warmup)).value if (t_max > t_warmup) else 1.0
+        frac_of_total = ((current_time - t_warmup) / (t_max - t_warmup)).value if (t_max > t_warmup) else 0.0
         frac_of_total = min(1.0, frac_of_total)
 
         current_factor = self.alpha_i + frac_of_total * (self.alpha_f - self.alpha_i)
@@ -760,7 +760,7 @@ class CosineAnnealingWithWarmupScheduler(ComposerScheduler):
 
         t_max = _convert_time(self.t_max, state, ssr=ssr)
         current_time = state.timestamp.get(t_warmup.unit)
-        frac_of_total = ((current_time - t_warmup) / (t_max - t_warmup)).value if (t_max > t_warmup) else 1.0
+        frac_of_total = ((current_time - t_warmup) / (t_max - t_warmup)).value if (t_max > t_warmup) else 0.0
         frac_of_total = min(1.0, frac_of_total)
 
         return _cosine_anneal(x=frac_of_total, min_y=self.alpha_f)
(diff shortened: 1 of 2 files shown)


composer/optim/scheduler.py
```diff
@@ -132,13 +132,12 @@ def _convert_time(time: Union[str, Time[int], Time[float]], state: State, ssr: f
     assert state.max_duration is not None, 'max_duration should be set whenever schedulers are invoked'
 
     if time.unit == TimeUnit.DURATION:
-        if state.dataloader_len is None:
-            raise RuntimeError('Cannot convert time, as state.dataloader_len is None.')
         if state.max_duration.unit == TimeUnit.EPOCH:
+            if state.dataloader_len is None:
+                raise RuntimeError('Cannot convert time, as state.dataloader_len is None.')
             return Time(int(time.value * int(state.dataloader_len) * state.max_duration.value), TimeUnit.BATCH)
         return Time(int(time.value * state.max_duration.value), state.max_duration.unit)
```
abhi-mosaic, line 139: Some context here, we don't actually need the dataloader length if the `state.max_duration` is in units like batches or samples or tokens.

jbloxham (approved):
This could be fixed slightly more concisely by changing all the `if state.timestamp < t_warmup:` checks to use `<=` instead of `<`, but that technically assumes that the main scheduler outputs an alpha of 1 for t=0. This looks good to me - and thanks for updating the tests!

abhi-mosaic:
> This could be fixed slightly more concisely by changing all the `if state.timestamp < t_warmup:` checks to use `<=` instead of `<`, but that technically assumes that the main scheduler outputs an alpha of 1 for t=0. This looks good to me - and thanks for updating the tests!

Yup this was my first thought as well and then I got worried thinking that someone might want to construct some wacky LR schedule that is discontinuous at `t_warmup`.

PR #1355, opened 2022-08-03
Update Export for Inference methods

nik-mosaic:
Closes https://mosaicml.atlassian.net/browse/CO-703.
Includes:
- Adding a Trainer wrapper over the `composer.utils.export_for_inference` method, trainer.export_for_inference().
- Adding tests to ensure DDP models are exported properly
- Updating ExportForInferenceCallback to be compatible with DDP models
- Use loggers to export model from Trainer wrapper and Callback instead of just an the ObjectStore argument.


composer/trainer/trainer.py
```diff
@@ -2193,3 +2194,39 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+            sample_input (Any, optional): Example model inputs used for tracing. This is needed for "onnx" export.
+                The ``sample_input`` need not match the batch size you intend to use for inference. However, the model
+                should accept the ``sample_input`` as is. (default: ``None``)
+            transforms (Sequence[Transform], optional): transformations (usually optimizations) that should
+                be applied to the model. Each Transform should be a callable that takes a model and returns a modified model.
+
+        Returns:
+            None
+        """
+        inference.export_for_inference(model=self.state.model,
```
dskhudia, line 2227: Let us use `sample_input` from `state.batch` as we do in the callback if not provided.

composer/trainer/trainer.py
```diff
@@ -2193,3 +2194,39 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+                                       save_format=save_format,
+                                       save_path=save_path,
+                                       save_object_store=save_object_store,
```
dskhudia, line 2230: Also if save_object_store is not given, it should default to the one given to trainer. See 'CheckpointSaver' callback on how it stores checkpoint.

composer/trainer/trainer.py
```diff
@@ -2193,3 +2195,41 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+            None
+        """
+        if sample_input == None:
+            sample_input = deepcopy(self.state.batch)
+        inference.export_for_inference(model=self.state.model,
+                                       save_format=save_format,
+                                       save_path=save_path,
+                                       save_object_store=save_object_store,
+                                       sample_input=(sample_input,),
+                                       transforms=transforms)
```
dskhudia, line 2235: nit: newline. Seems like something is wrong with your editor setting. Your previous PR also had this.

nik-mosaic: Sorry about that, will fix.

composer/trainer/trainer.py
nik-mosaic: If we do this, it could be helpful & convenient to users. However, it will break the pattern of the other ExportForInference functions ([composer.callbacks.ExportForInferenceCallBack](https://github.com/mosaicml/composer/blob/4f00c2205c800b4962bfeb110153ee566db79e15/composer/callbacks/export_for_inference.py#L49) and [composer.utils.export_for_inference](https://github.com/mosaicml/composer/blob/4f00c2205c800b4962bfeb110153ee566db79e15/composer/utils/inference.py#L68)), for which save_object_store is either an ObjectStore object (if the user is saving remotely), or None, if the user wishes to save locally.
Implementing it this way would result in different behavior for
```
trainer = Trainer(
    model=MNIST_Classifier(num_classes=10),
		callbacks=ExportForInferenceCallback(save_path='model/",save_format='torchscript')
		load_path='/tmp/ep1-rank0',
		loggers='ObjectStoreLogger('s3',..)'
)
```
and
```
trainer = Trainer(
    model=MNIST_Classifier(num_classes=10),
		load_path='/tmp/ep1-rank0',
		loggers='ObjectStoreLogger('s3',...)'
)
trainer.export_for_inference(save_path='model/",save_format='torchscript')
```
One would be saved locally and one would not.

nik-mosaic: @ravi-mosaicml or @hanlint. Any thoughts?

dskhudia: Discussed offline but leaving my thoughts here too: IMO this discrepancy is ok since` trainer.export_for_infernce ` method is tied to trainer and user is most likely to use the same object store for both.

 
If the trainer is saving checkpoints locally, `trainer.export_for_inference` will also export locally

if the trainer is saving checkpoints in a remote store, `trainer.export_for_inference` will also export to the same remote store

hanlint: Looking at the code above, I can't actually tell which example would save locally and which one would not, which seems to me we should maintain consistent behavior across both. In both cases, I would expect the method to only save locally, and never export? 

Better to require explicit, rather than be convenient but unclear?

composer/trainer/trainer.py
```diff
@@ -2193,3 +2195,41 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+            sample_input (Any, optional): Example model inputs used for tracing. This is needed for "onnx" export.
+                The ``sample_input`` need not match the batch size you intend to use for inference. However, the model
+                should accept the ``sample_input`` as is. (default: ``None``)
+            transforms (Sequence[Transform], optional): transformations (usually optimizations) that should
+                be applied to the model. Each Transform should be a callable that takes a model and returns a modified model.
+
+        Returns:
+            None
+        """
+        if sample_input == None:
```
hanlint, line 2228: `.. and save_format == 'onnx'` ?

composer/trainer/trainer.py
```diff
@@ -2193,3 +2195,41 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+            sample_input = deepcopy(self.state.batch)
+        inference.export_for_inference(model=self.state.model,
```
hanlint, line 2230: do we need to worry if `self.state.model` is DDP or deepspeed wrapped? might be worth running a quick test here

dskhudia: For deepspeed, we have an error in `export_for_inference` functional API https://github.com/mosaicml/composer/blob/dev/composer/utils/inference.py#L94  

Will be great to test DDP wrapped module.

composer/trainer/trainer.py
dskhudia: @nik-mosaic In your examples above, since loggers is S3 object store both would store using this logger. Isn't it?

composer/trainer/trainer.py
nik-mosaic: Writing a DDP test now.

composer/trainer/trainer.py
nik-mosaic: Because save_object_store is not passed into the ExportForInferenceCallback, it will be none, which will call composer.utils.export_for_inference with save_object_store=None, which will then cause it to be stored locally, since we have `is_remote_store = save_object_store is not None`.

dskhudia: Oh! I thought callback also uses loggers. I think we should use loggers for `ExportForInferenceCallback` too.

dskhudia: By doing so we will make exporting functionality similar to checkpoint_saver functionality that also uses logger for checkpoint saving.

composer/trainer/trainer.py
dskhudia: Correct. Please go ahead and add it for that functional API

composer/trainer/trainer.py
nik-mosaic: I can definitely see the benefit of making the decision to use loggers across the board, but that's probably best done in a separate PR to keep the scope of this one limited to the trainer Export wrapper.

dskhudia: Sure. feel free to do it in a separate PR. Could you please create a JIRA so that we don't forget? I was hoping we could add it all for the release today along with notebook update https://github.com/mosaicml/composer/pull/1360

composer/trainer/trainer.py
```diff
@@ -2193,3 +2195,41 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
...
+                The ``sample_input`` need not match the batch size you intend to use for inference. However, the model
+                should accept the ``sample_input`` as is. (default: ``None``)
+            transforms (Sequence[Transform], optional): transformations (usually optimizations) that should
+                be applied to the model. Each Transform should be a callable that takes a model and returns a modified model.
+
+        Returns:
+            None
+        """
+        if sample_input == None and save_format == 'onnx':
+            sample_input = deepcopy(self.state.batch)
```
ravi-mosaicml, line 2229: Curious why do we need to deepcopy the input? Should this copy be done in the `export_for_inference` function?

composer/trainer/trainer.py
ravi-mosaicml: I don't think we should need to take an object store here; instead, this function can call `logger.file_artifact(save_path)`. That will then use any/all artifact loggers specified as part of the trainer.

ravi-mosaicml (commented):
Didn't give a close review, but see comments re: deepcopy and object store. Otherwise will defer to Daya / Hanlin.

composer/trainer/trainer.py
dskhudia: This is just for overriding, in case, a user want to export model to a specific object store separate from training checkpoints. if None, it will use `logger.file_artifact`.

composer/trainer/trainer.py
nik-mosaic: Added two tests; one with Onnx export and one with torchscript.

composer/trainer/trainer.py
nik-mosaic: sample_input could be modified inside export_for_inference; e.g. sample_input = ensure_tuple(sample_input) in the onnx part of the method. We can do the copy inside of export_for_inference since we also copy the model there.

nik-mosaic: sample_input could be modified inside export_for_reference(). However, it does make more sense to do the copy inside of the method before the rest of the logic.

composer/callbacks/export_for_inference.py
```diff
@@ -9,6 +9,8 @@
 from copy import deepcopy
 from typing import Any, Optional, Sequence, Union
 
+from torch.nn import Module
```
dskhudia, line 12: nit: standard way of importing this is import torch.nn as nn and then use nn.Module. Change here and other places you use it.

composer/utils/inference.py
```diff
@@ -101,6 +101,9 @@ def export_for_inference(
     # make a copy of the model so that we don't modify the original model
```
dskhudia, line 101: Let us raise an error in this function if the model is ddp wrapped (same as we do for deepspeed wrapped).

composer/callbacks/export_for_inference.py
nik-mosaic: Done!

dskhudia (approved):

dskhudia:
Great work!

Issue #1270, opened 2022-07-10
Composer profiler not supporting multi-node settings


YilunKuang:
**Environment**

- OS: [CentOS Linux 7 (Core)]
- Hardware: [8 NVIDIA A100-SXM4-40GB, two nodes with four GPUs each]

**Setting**
I was running composer trainer for ResNet50 training with profiler in a multi-node environment (two nodes, each with 4 GPUs) using PyTorch with SLURM. All of the environment variables (`RANK`, `WORLD_SIZE`, `LOCAL_RANK`, `LOCAL_WORLD_SIZE`, `NODE_RANK`, `MASTER_ADDR`, `MASTER_PORT`) are properly set and the code runs smoothly without profiler. 

```python
composer_profiler = Profiler(
                                    trace_handlers=JSONTraceHandler(folder=comprof_folder, overwrite=True),
                                    schedule=cyclic_schedule(
                                        wait=0,
                                        warmup=1,
                                        active=4,
                                        repeat=1,
                                    ),
                                    torch_prof_folder=torch_prof_folder,
                                    torch_prof_overwrite=True,
                                )
```

I am able to get a merge trace file from processes 0-3 on the first node saved as `merged_trace.json`, but processes 4 on the second node give the following error

```bash
Traceback (most recent call last):
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/site-packages/submitit/core/_submit.py", line 11, in <module>
    submitit_main()
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/site-packages/submitit/core/submission.py", line 72, in submitit_main
    process_job(args.folder)
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/site-packages/submitit/core/submission.py", line 65, in process_job
    raise error
  File "/mnt/home/ykuang/miniconda3/envs/mcmc2/lib/python3.10/site-packages/submitit/core/submission.py", line 54, in process_job
    result = delayed.result()
[...]

PR #1358, opened 2022-08-03
Fix the profiler on multi-node training

ravi-mosaicml:
https://github.com/mosaicml/composer/issues/1270 identified how the trace merger did not work on multi-node training. This PR fixes that by using the local rank zero, rather than the global rank zero, to synchronize timestamps.

Did not add test cases as we do not support multi-node testing.

Closes https://github.com/mosaicml/composer/issues/1270
Closes https://mosaicml.atlassian.net/browse/CO-674


mvpatel2000 (approved):
LGTM

ravi-mosaicml:
Manually tested by running the `examples/profiler_demo.py` script.

PR #1361, opened 2022-08-04
Move metrics to context

mvpatel2000:


Landanjs (approved):
LGTM

composer/trainer/trainer.py
```diff
@@ -1497,7 +1497,11 @@ def _train_loop(self) -> None:
                                     self.state.batch, self.state.grad_accum):
                                 # TODO: Detect if self.run_event(Event.AFTER_DATALOADER) changes the training
                                 # data and if so print a warning that metrics may return unexpected results
-                                self.train_metrics.update(*self._original_model.validate(eval_microbatch))
+                                with get_precision_context(self.state.precision):
+                                    outputs, targets = self._original_model.validate(eval_microbatch)
+                                # Cast to full precision to avoid NaN
+                                with torch.cuda.amp.autocast(enabled=True, dtype=torch.float32):
```
dskhudia, line 1503: You can just do `torch.autocast(enabled=False)` to make it device and dtype agnostic.

mvpatel2000: ([JIRA](https://mosaicml.atlassian.net/browse/CO-827)) so the issue is if training happens in lower precision, metrics might blow up if not computed in same precision. Tried always setting it to float 32, but realizing actually there might be stuff that would require lower floats, so I'm moving this to same precision context as `validate()`

abhi-mosaic: Are there really metrics that would be computed in low precision? I feel like it kinda makes sense to always do FP32... I even see a comment in the Trainer that we revert metrics to FP32 when DeepSpeed accidentally downcasts them.

Also @dskhudia I think that `torch.autocast(enabled=False)` would be like a no-op, but what we want is to force the inputs to metrics like softmax to be upcast to `torch.float32` (they are coming in as 16 bit), so I think the dtype might be necessary?

abhi-mosaic: Alternatively, if we have some general function that can just convert a batch to a dtype, then all we really need to do is call `upcast(outputs)` and `upcast(targets)`

dskhudia: `torch.autocast(enabled=False)` will disable autocasting but inputs under this context should be upcast. So your `upcast(outputs)` makes sense and a reasonable way to do it.

dskhudia: Also agree with doing metrics update with fp32.

mvpatel2000: TLDR: It looks like its pretty hard to autocast / upcast to fp32 on CPU. You have to manually cast each tensor, but model output can be really messy. Torch handles a lot of stuff internally well (eg Abhi found + casts to fp32), so after offline discussion, i think we should leave PR as is

mvpatel2000: @dskhudia if this sounds reasonable to you, can you approve + resolve comments? need approval from code owner

dskhudia: sounds good.

dskhudia (approved):

PR #1362, opened 2022-08-04
Update CUDA and remove NCCL downgrade from Dockerfile

abhi-mosaic:
This PR updates CUDA versions for our images and remove the NCCL downgrade that was previously done for GCP support. If we revisit GCP and have issues we can make changes at that time specifically for GCP (not the default images).

New build matrix:
* torch 1.10.2 + cuda 11.3.1
* torch 1.11.0 + cuda 11.5.2
* torch 1.12.0 + cuda 11.6.2

[OLD] I am having trouble with BF16 NCCL communication (necessary for good FSDP perf) on our MosaicML images, I have a feeling it might be due to the NCCL downgrade hardcoded in our Dockerfile, so I'm going to try removing that and see if it fixes anything.


abhi-mosaic:
@bandish-shah I reinstalled NCCL from within one of our images and confirmed that upgrading to `2.13.4` does fix the issue. Is there any harm to just removing these lines? And building a separate GCP-specific image if we need it? Maybe GCP is fixed now as well?

I'm also itching to switch to CUDA 11.6 if you have any thoughts.

bandish-shah (approved):
LGTM!

Should we remove the `mosaicml/pytorch:1.12.0_cu113-python3.9-ubuntu20.04`, `mosaicml/pytorch:1.11.0_cu113-python3.8-ubuntu20.04` and `mosaicml/pytorch_vision:1.12.0_cu113-python3.9-ubuntu20.04` images from Docker Hub?  Don't want to confuse folks since these are no longer in the README.

ravi-mosaicml:
> LGTM!
> 
> Should we remove the `mosaicml/pytorch:1.12.0_cu113-python3.9-ubuntu20.04`, `mosaicml/pytorch:1.11.0_cu113-python3.8-ubuntu20.04` and `mosaicml/pytorch_vision:1.12.0_cu113-python3.9-ubuntu20.04` images from Docker Hub? Don't want to confuse folks since these are no longer in the README.

Those are the up to date images. there are others that are not up to date on dockerhub.

docker/Dockerfile
```diff
@@ -124,24 +124,6 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-##################################################
-# Change the NCCL version to fix NVLink errors
-# This is required to train on Nvidia A100s in GCP
-##################################################
-RUN if [ -n "$CUDA_VERSION" ] ; then \
```
ravi-mosaicml, line 131: Do we want to make this a conditional thing that can be controlled via a build arg?
I have a slight preference to leaving the code around, until GCP fixes the underlying issue.

ravi-mosaicml (approved):
Changes lgtm, though see comment.

docker/Dockerfile
abhi-mosaic: I think we're not sure if GCP still has the issue, someone's gotta take some time and try it out again... maybe it's been fixed.

I can keep the code here but commented out if that's preferrable?

abhi-mosaic:
I think there may be some documentation to update too with docker image tags, I will grep through the codebase to double check.

docker/Dockerfile
bandish-shah: Let's remove it, this was a total hack and support isn't scalable longer term especially as NCCL versions update with subsequent CUDA versions.  We can consider adding it back if someone raises an issue but let's keep it simple and let this be use case driven.

PR #1363, opened 2022-08-04
Refactor GLUE fine-tune queuing to improve efficiency and add task-specific seed sweeps

alextrott16:
- Runs fine tuning in 2 rounds of parallel jobs (as opposed to `n_pretrain_checkpoints * 2` rounds of parallel jobs). This enables more efficient parallelization, especially when running on machines with `>= 4` GPUs. Concretely, this puts all fine-tuning runs that do not depend on each other in the same round of parallel jobs:
  - The first round of fine-tuning runs train on COLA, MNLI, QNLI, QQP, and SST-2 for all pre-train checkpoints and saves checkpoints for MNLI runs.
  - The second round of fine-tuning trains on RTE, MRPC, and STS-B for all the pre-train checkpoints' downstream MNLI checkpoints (which are only available after the first round).

- Adds `seed_overrides` to the fine-tuning Hparams so the user can specify to sweep over seeds for certain tasks that tend to have noisy outcomes. For example, for each pre-train checkpoint COLA fine-tuning should happen 4 times, each using a different seed.

- This PR includes fixes for a number of "breaking changes" that happened during its development, including issues with multiprocessing, checkpoint naming, and metrics.


examples/glue/run_glue_trainer.py
```diff
@@ -185,41 +185,62 @@ def merge_hparams(hparams: TrainerHparams, override_hparams: GLUETrainerHparams)
...
-        task = finetune_tasks[rank]
-        future = executor.submit(train_finetune, base_yaml_file, task, task_to_save_ckpt[task], ckpt_load_path,
-                                 wandb_group_name, ckpt_save_folder, save_locally, free_port + rank, load_ignore_keys)
-        future.add_done_callback(done_callback)
+
+    def make_callback(parent_ckpt, glue_metrics):
+        return lambda future: log_metrics(future.result(), ckpt_filename=parent_ckpt, glue_metrics=glue_metrics)
+
+    # Fine-tune from pre-trained checkpoint(s)
+    for parent_idx, (ckpt_load_path, parent_ckpt) in enumerate(zip(ckpt_load_paths, parent_ckpts)):
```
moinnadeem, line 224: Can we move the `zip` and `(ckpt_load_path, parent_ckpt)` into their own line? This line has a lot of work going on.

examples/glue/run_glue_trainer.py
```diff
@@ -274,17 +301,23 @@ def train_finetune(
...
         ft_hparams.save_latest_artifact_name = save_latest_artifact_name
 
         if save_locally:
             if not os.path.exists(ft_hparams.save_folder):
-                os.mkdir(ft_hparams.save_folder)
+                os.makedirs(ft_hparams.save_folder)
+
+    else:
+        # Disable saving
+        ft_hparams.save_folder = None
```
moinnadeem, line 316: good call on this one

composer/yamls/models/glue/stsb.yaml
```diff
@@ -52,3 +52,4 @@ callbacks:
 load_path: https://storage.googleapis.com/llm_checkpoints/bert_checkpoint/bert_checkpoints/ep7.pt
 load_weights_only: true
 load_strict_model_weights: false
+precision: fp32 # Eval requires FP32
```
moinnadeem, line 55: Is this only required for STS-B? Or is this an impact of the precision refactor and we should avoid this long-term?

examples/glue/run_glue_trainer.py
```diff
@@ -185,41 +185,62 @@ def merge_hparams(hparams: TrainerHparams, override_hparams: GLUETrainerHparams)
...
+            if task not in seed_overrides:
+                # Run 1 fine-tune trainer from this checkpoint, with no seed overriding.
+                future = executor.submit(train_finetune, base_yaml_file, task, save_ckpt, ckpt_load_path, parent_ckpt,
+                                         parent_idx, ckpt_save_folder, save_locally, free_port + rank, load_ignore_keys)
+                future.add_done_callback(make_callback(parent_ckpt, glue_metrics))
+                rank += 1
+            else:
+                # Run multiple fine-tune trainers from this checkpoint, using a different seed override for each
+                for seed in seed_overrides[task]:
+                    future = executor.submit(train_finetune, base_yaml_file, task, save_ckpt, ckpt_load_path,
```
moinnadeem, line 239: should we just deepcopy `train_finetune` and update the seed here, instead of adding an extra parameter to the signature and updating the seed there?

moinnadeem (approved):
LGTM; thanks for doing this!

moinnadeem (changes requested):
Switching to request changes on the `precision:` stuff since that might be non-trivial for long term

composer/yamls/models/glue/stsb.yaml
alextrott16: It seems to be a consequence of the precision refactor (STS-B wasn't causing problems before that), but the issue remains after merging in dev after a fix was pushed. It's unclear to me what the root of the issue is, but it has something to do with the `SpearmanCorrCoeff` metric wanting the model predictions to be `fp32`. I'm not sure how to prevent this issue at the code level, but using `fp32` prevents the error and adds only a small overhead for this quick fine-tuning task.

Given the minor impact, I was hoping we could leave this as a TODO. If it needs to be dealt with now, I'm open to suggestions on how to proceed :)

examples/glue/run_glue_trainer.py
alextrott16: Given the way the `merge_hparams` function works, I'm not sure it would work to do the deepcopy strategy. My thinking was also that, because the downstream `train_finetune` function already handles several overwrites, it would be cleaner to handle the seed overwrite in that area of the code.

composer/yamls/models/glue/stsb.yaml
moinnadeem: What concerns me isn't the small change on this PR, but rather, it indicates that something about the root cause of the issue still isn't fixed, and it'll be way worse if an open-source user is running into this -- it's highly less likely that they won't know what to do! In other words, we can continue being productive but Composer overall will suffer.

Since we can run our weekend runs without this, @mvpatel2000 thoughts on working with Alex on root-causing and finding an appropriate Composer solution that doesn't require `precision: fp32` here?

examples/glue/run_glue_trainer.py
moinnadeem: That is fair, I can go ahead and resolve this one!

examples/glue/run_glue_trainer.py
```diff
@@ -185,41 +185,62 @@ def merge_hparams(hparams: TrainerHparams, override_hparams: GLUETrainerHparams)
...
                     initializer=init_cuda_env,
                     initargs=(cuda_envs, free_port),
                     mp_context=ctx)
-    for rank in range(num_tasks):
-        task = finetune_tasks[rank]
-        future = executor.submit(train_finetune, base_yaml_file, task, task_to_save_ckpt[task], ckpt_load_path,
-                                 wandb_group_name, ckpt_save_folder, save_locally, free_port + rank, load_ignore_keys)
-        future.add_done_callback(done_callback)
+
+    def make_callback(parent_ckpt, glue_metrics):
```
ishanashastri, line 220: Curious why we need this wrapper for the lambda function?

ishanashastri, line 220: Curious why we need this wrapper for the lambda?

composer/yamls/models/glue/stsb.yaml
alextrott16: Very good points!

examples/glue/run_glue_trainer.py
```diff
@@ -185,41 +185,62 @@ def merge_hparams(hparams: TrainerHparams, override_hparams: GLUETrainerHparams)
...
+        return lambda future: log_metrics(future.result(), ckpt_filename=parent_ckpt, glue_metrics=glue_metrics)
+
+    # Fine-tune from pre-trained checkpoint(s)
+    for parent_idx, (ckpt_load_path, parent_ckpt) in enumerate(zip(ckpt_load_paths, parent_ckpts)):
+        # `ckpt_load_path` provides the path to the checkpoint from which we load the starting weights used when fine-tuning
+        # `parent_ckpt` keeps track of the original pre-training checkpoint, for tasks with multiple fine-tuning stages (e.g., RTE)
+        # `parent_idx` is used for bookkeeping, so `parent_ckpt` can be internally recovered from the path used to save fine-tune checkpoints
+        rank = 0
+        for task, save_ckpt in task_to_save_ckpt.items():
+            if task not in seed_overrides:
```
ishanashastri, line 230: Would it make sense to instead put all the tasks in `seed_overrides,` even if it only needs a single seed, and then remove this extra if-else loop? This also may make yaml vs entrypoint details less confusing. See mostly duplicate code so just thinking about how to collapse.

examples/glue/run_glue_trainer.py
```diff
@@ -274,17 +301,23 @@ def train_finetune(
     # saving single checkpoint at the end of training the task
     if save_ckpt:
         # add task specific artifact logging information
-        ft_hparams.save_folder = f'{save_folder}/{task}'
-        save_artifact_name = f'{save_folder}/{task}/ep{{epoch}}-ba{{batch}}-rank{{rank}}'  # ignored if not uploading
-        save_latest_artifact_name = f'{save_folder}/{task}/latest-rank{{rank}}'
+        ft_hparams.save_folder = f'{save_folder}/{task}-{parent_idx:03d}'
+        save_artifact_name = f'{save_folder}/{task}-{parent_idx:03d}/ep{{epoch}}-ba{{batch}}-rank{{rank}}'  # ignored if not uploading
```
ishanashastri, line 305: Do we also need to save the seed information here and in wandb?

examples/glue/run_glue_trainer.py
```diff
@@ -185,41 +185,62 @@ def merge_hparams(hparams: TrainerHparams, override_hparams: GLUETrainerHparams)
...
 ) -> None:
-    """Set up CUDA environment and process pool for given finetuning jobs."""
+    """Set up CUDA environment and process pool for given finetuning jobs and wait for them to complete."""
+    if parent_ckpts:
+        assert len(parent_ckpts) == len(ckpt_load_paths), 'Must supply one parent_ckpt per ckpt_load_path'
+    else:
+        parent_ckpts = ckpt_load_paths  # By default, the "parent checkpoint" is logged simply as the checkpoint
+
+    # To reduce noisiness with these tasks, expand the evaluation to multiple fine-tuning seeds per pre-train checkpoint
+    seed_overrides = {
```
ishanashastri, line 203: Wondering if it makes sense to make a field in `finetune_hparams` to store these overrides along with all the other overrides? That way, someone is not bound to spawning all these seeds if they don't need to run all the different seedings.

ishanashastri (commented):
Overall looks really good, thanks for putting this together so quickly! Mostly comments about seeds and where those overrides should lie.

examples/glue/run_glue_trainer.py
ravi-mosaicml: For readability, I think it would be helpful to hardcode the lambda directly in twice, even though it would be repeated code.

ravi-mosaicml (approved):
Overall looks fine; I trust that this works, but see comments re: readability.
The inline comments are very helpful.

examples/glue/run_glue_trainer.py
alextrott16: Added this change

examples/glue/run_glue_trainer.py
alextrott16: Removed the wrapper

examples/glue/run_glue_trainer.py
ishanashastri: Will sync offline about this!

moinnadeem (approved):
LGTM, let's move the precision context into a new issue so we can unblock. We do need to address the precision for sure

hanlint (approved):

PR #1368, opened 2022-08-05
Clean up unused exception

mvpatel2000:
As title. Diff is large, but I just removed the exception and unindented. We can add this back if we want to use it, but I'm pro simplifying code and not leaving in dead snippets. Fixes CO-864


hanlint:
Hmm I believe we needed this for the `Benchmarker` code to work, cc: @abhi-mosaic ?

ravi-mosaicml (approved):
LGTM; per the PR comment I did not validate the diff.

PR #1369, opened 2022-08-05
Fix default precision

A-Jacobson:
fixes a bug causing default precision to be fp32 if trainer `device='gpu'` and AMP if `device=DeviceGPU()`


mvpatel2000 (approved):
This is a great catch!

hanlint:
interesting, this seems to have surfaced an odd failure in our tests:
> AssertionError: /scaler/GradScaler/_growth_tracker differs: 4 != 2

Going to take a look at this a bit later.

PR #1370, opened 2022-08-05
add filename extension

mvpatel2000:
Changes default name to have `.pt`.  Fixes CO-850.


ravi-mosaicml (approved):
LGTM; thanks!

ravi-mosaicml:
Forgot to mention, but do we want the latest symlink to also have an extension?

mvpatel2000:
> Forgot to mention, but do we want the latest symlink to also have an extension?

Its in the PR -- yea I think we do so it looks like a checkpoint file

PR #1371, opened 2022-08-05
JSON Schemas

mvpatel2000:
With YAHP 0.1.3 supporting JSON Schema generation, we now want to add JSON Schemas to Composer. This PR adds tests which automatically validate example YAMLs we provide (and corrects a few issues in them).

PR #1373, opened 2022-08-05
JSON Schemas pt 2

mvpatel2000:
Follow up on https://github.com/mosaicml/composer/pull/1371 which got automerged


ravi-mosaicml (approved):
LGTM; though not sure why the diff is so large considering the previous PR. Didn't check this closely but assume it was just formatting / list changes.

PR #1374, opened 2022-08-05
Perf Analyzer

mvpatel2000:
Closes CO-749. Step 1 of autoperf. 

Adds a new type of trace handler which searches for dataloader bottlenecks. Reworks abstractions to push action information through to tracehandler (so we can give warnings on the last step of profile window)

Along with mock pytests, I've run a variety of loads playing around with thresholds and what feels like the right borderline, eg various ResNet recipes which let me control bottleneck %. This looks like it's in a pretty good state and fairly robust for what we want to open source. Also changes default profile windows to be more reasonable.

### DIFF
--- a/composer/profiler/performance_analysis_trace_handler.py
+++ b/composer/profiler/performance_analysis_trace_handler.py
@@ -66,10 +66,11 @@ class PerformanceAnalyzerTraceHandler(TraceHandler):
             self._record_wait_time(is_start, self.dataloader_wait_times, timestamp.batch.value, wall_clock_time_ns)
         elif 'batch' in categories:
             self._record_wait_time(is_start, self.batch_wait_times, timestamp.batch.value, wall_clock_time_ns)
-        if action == ProfilerAction.ACTIVE_AND_SAVE and self.is_dataloader_bottlenecked():
-            warnings.warn(
-                f'The current training run is dataloader bottlenecked. {self._average_dataloader_wait_time()} additional time is being taken per batch waiting for the dataloader to fetch more data.'
-            )
+            # Only check for bottleneck on end of batch
+            if not is_start and action == ProfilerAction.ACTIVE_AND_SAVE and self.is_dataloader_bottlenecked():
+                warnings.warn(
+                    f'The current training run is dataloader bottlenecked. {self._average_dataloader_wait_time()} additional time is being taken per batch waiting for the dataloader to fetch more data.'
+                )
 
     def _average_batch_wait_time(self) -> float:
         return self._pruned_mean(list(self.batch_wait_times.values()))
--- a/composer/profiler/performance_analysis_trace_handler.py
+++ b/composer/profiler/performance_analysis_trace_handler.py
@@ -72,8 +72,12 @@ class PerformanceAnalyzerTraceHandler(TraceHandler):
             self._record_wait_time(False, self.batch_wait_times, timestamp.batch.value, wall_clock_time_ns)
             # Only check for bottleneck on end of batch
             if not is_start and action == ProfilerAction.ACTIVE_AND_SAVE and self.is_dataloader_bottlenecked():
+                dataloader_time = self._average_dataloader_wait_time()
+                batch_time = self._average_batch_wait_time()
+                percentage = format(dataloader_time / batch_time * 100, '.2f')
+                additional_latency = format(dataloader_time / 1e9, '.3f')
                 warnings.warn(
-                    f"The current training run is dataloader bottlenecked. Waiting {format(self._average_dataloader_wait_time() / 1e9, '.3f')}s per batch for dataloader."
+                    f'The current training run is dataloader bottlenecked. The dataloader takes {percentage}% more time to compute a batch, adding {additional_latency}s per batch for dataloader.'
                 )
 
     def _average_batch_wait_time(self) -> float:
@@ -90,7 +94,7 @@ class PerformanceAnalyzerTraceHandler(TraceHandler):
         mstdevs = deviations_from_median / median_deviation if median_deviation else 0.
         filtered_arr = arr[mstdevs < mstdev_threshold]
         # Return mean of filtered_arr if it exists, otherwise return median of original array
-        return float(np.mean(filtered_arr)) if len(filtered_arr) > 0 else np.median(arr)
+        return float(np.mean(filtered_arr)) if len(filtered_arr) > 0 else float(np.median(arr))
 
     def is_dataloader_bottlenecked(self) -> bool:
         """Bottlenecked if batch yield time is more than 0.1% of batch compute time with >=10 datapoints."""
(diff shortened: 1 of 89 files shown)


composer/profiler/profiler_schedule.py
```diff
@@ -12,10 +12,10 @@
 
 
 def cyclic_schedule(
-    skip_first: int = 0,
+    skip_first: int = 1,
     wait: int = 0,
     warmup: int = 1,
-    active: int = 4,
+    active: int = 50,
```
ravi-mosaicml, line 18: CC @bandish-shah  on these changes -- they look fine to me.

composer/profiler/performance_analysis_trace_handler.py
```diff
@@ -0,0 +1,114 @@
...
+
+class PerformanceAnalyzerTraceHandler(TraceHandler):
+    """Meaures if trainer run is model or dataloader bottlenecked."""
+
+    def __init__(self):
+        # Stores wait times for the current epoch
+        self.dataloader_wait_times: Dict[int, Any] = {}
+        self.batch_wait_times: Dict[int, Any] = {}
+        self.last_start_time = None
+        # Dataloader is bottlenecked if yield takes more than 1% of batch compute time and at least
```
ravi-mosaicml, line 30: Should the batch yield time depend on the size of the batch? From the pytorch DL source code, I don't think it does -- it just moves pointers around and increments / decrements counters: https://github.com/pytorch/pytorch/blob/1cd1f48d9311561920c841d322a98b805e49dcc7/torch/utils/data/dataloader.py#L676

Thinking then would the constant overhead check be sufficient? Curious if you found in practice it does depend on the batch size?

I do think it would still be super helpful to report how long (as a %) we spent with dataloader yielding, but on second thought, I'm not sure whether that percentage should be used to justify whether we are DL bottlenecked. CC @bandish-shah?

composer/profiler/performance_analysis_trace_handler.py
```diff
@@ -0,0 +1,114 @@
...
+
+MINIMUM_DATAPOINTS_NEEDED = 10
+
+
+class PerformanceAnalyzerTraceHandler(TraceHandler):
+    """Meaures if trainer run is model or dataloader bottlenecked."""
+
+    def __init__(self):
+        # Stores wait times for the current epoch
+        self.dataloader_wait_times: Dict[int, Any] = {}
```
ravi-mosaicml, line 27: nit: can the typing be more specific -- e.g. `Dict[int, float]`?

composer/profiler/performance_analysis_trace_handler.py
```diff
@@ -0,0 +1,114 @@
...
+        elif 'after_train_batch' in categories and 'PerformanceAnalyzerTraceHandler' in categories:
+            self._record_wait_time(False, self.batch_wait_times, timestamp.batch.value, wall_clock_time_ns)
+            # Only check for bottleneck on end of batch
+            if not is_start and action == ProfilerAction.ACTIVE_AND_SAVE and self.is_dataloader_bottlenecked():
+                dataloader_time = self._average_dataloader_wait_time()
+                batch_time = self._average_batch_wait_time()
+                # Log percentage overhead and additional time in milliseconds
+                percentage = format(dataloader_time / batch_time * 100, '.2f')
+                additional_latency = format(dataloader_time / 1e6, '.3f')
+                warnings.warn(f'The current training run is dataloader bottlenecked, adding {additional_latency}ms and '
```
ravi-mosaicml, line 88: nit: the formating can go directly in the format string below; no need to call `format` manually -- e.g.

 ```suggestion
                 warnings.warn(f'The current training run is dataloader bottlenecked, adding {additional_latency:.3f}ms and '
```

File: composer/profiler/performance_analysis_trace_handler.py [hunk 5]
```diff
@@ -0,0 +1,114 @@
...
+                              f'increasing batch time by {percentage}%. This warning may be spurious or noisy if the '
+                              '`active` period of the profiler schedule is too small.')
+
+    def _average_batch_wait_time(self) -> float:
+        return self._pruned_mean(list(self.batch_wait_times.values()))
+
+    def _average_dataloader_wait_time(self) -> float:
+        return self._pruned_mean(list(self.dataloader_wait_times.values()))
+
+    def _pruned_mean(self, data: List[float], mstdev_threshold=2.0) -> float:
```
ravi-mosaicml, line 98:
```suggestion
    def _pruned_mean(self, data: List[float], mstdev_threshold: float = 2.0) -> float:
```

ravi-mosaicml: This math seems fine though curious about the rational to use this formula to exclude outliers? Is this the standard way of doing it?

ravi-mosaicml (commented):
Overall lgtm -- see comments. Thanks for adding this in!

Will hold off on approval re: the discussion on whether this should live in the private or public repo.

composer/profiler/performance_analysis_trace_handler.py
mvpatel2000: https://en.wikipedia.org/wiki/Median_absolute_deviation -- this is one way I've seen using stdev style filtering to be more robust with outliers. median stdevs are convergent in probability to stdev, but slightly more robust. Not an expert on this (recalling things I heard somewhere in a stat class)

composer/profiler/performance_analysis_trace_handler.py
mvpatel2000: My understanding (haven't traced through the PyTorch code though) is if a dataset is written in a way where decoding happens on fetching an item (eg for images), the yield time becomes a fn of batch size

composer/profiler/profiler_schedule.py
bandish-shah: This is probably fine, the one issue I can think of is if someone enables the Torch profiler.  The wall clock time is significant when we turn it on so if a user runs with the defaults and enables the Torch profiler, it could slow down the workload substantially.  Is there a smaller value, maybe 10-20 that would work?  Alternatively could we have a schedule per trace handler vs a global schedule?  My guess is that would require a significant amount of re-factor to the profiler to support.

composer/profiler/performance_analysis_trace_handler.py
bandish-shah: In general, we should assume that we are dataloader bottlenecked any time the model is idle because we are waiting for dataloader to return a batch.  The dataloader latency should include any Python control overhead, which I believe can be dataloader implementation specific.  At some point it would be good to provide a breakdown indicating what exactly the dataloader is spending time on, we can include control overhead as a part of this break down.   

Might be good to chat about this offline but personally I'm not a huge fan of static thresholds.  If the implementation requires them, we should make them configurable by the user because it's not clear if we can come up with good defaults that work in all cases.

mvpatel2000:
Closing based on offline discussions

PR #1375, opened 2022-08-06
Log multiple losses

Landanjs:
Resolves CO-842. Adds logic to log multiple losses. Also, adds loss dictionary to DeepLabv3+ as an example. A couple of scenarios:

### 1. Scalar tensor loss

This will be logged as `loss/train/total`. Bert example:

### 2. Tuple of tensor losses

Individual losses will be logged as `'loss/train/loss{i}'` where i is the index of the individual loss. There will also be `'loss/train/total'` which is the sum of the individual losses.

### 3. Dictionary of losses without total key

Individual losses will be logged as `'loss/train/{loss_name}'`. There will also be `'loss/train/total'` which is the sum of the individual losses.

### 4. Dictionary of losses with total key

Individual losses will be logged as `'loss/train/{loss_name}'`. Loss given at `'loss/train/total'` will be used as the loss for the backpropagation, individual losses are not summed when `'total'` is present.

### Questions

- Assumes returned loss is a `torch.Tensor`, is this fine?
- Should these different scenarios be documented somewhere?


Landanjs:
Ah, I think I skipped over closures... Need some time

abhi-mosaic:
Is there any way we can get rid of the trailing `/` when there is only 1 loss value, like `loss/train` instead of `loss/train/` ? I'm just wondering if there could be some parsing issues later on.

Also, I wonder if we can drop the `_loss` suffix for things like `loss/train/dice` rather than `loss/train/dice_loss`, as it seems a bit repetitive.

Landanjs:
Thanks for the suggestion Abhi, good points! Yeah, that would be cleaner, let me try to set it up 🙂

composer/trainer/trainer.py
```diff
@@ -1518,14 +1518,16 @@ def _train_loop(self) -> None:
...
-                    full_loss = total_loss.cpu().item()
-                    self.logger.data_batch({'loss/train': full_loss / dist.get_world_size()})
+                    if isinstance(total_loss, dict):
+                        map_collection(total_loss, dist.all_reduce)
+                        total_loss = {k: loss.cpu().item() / dist.get_world_size() for k, loss in total_loss.items()}
+                        self.logger.data_batch(total_loss)
+                    if isinstance(total_loss, torch.Tensor):
+                        dist.all_reduce(total_loss, reduce_operation='SUM')
+                        full_loss = total_loss.cpu().item()
+                        self.logger.data_batch({'loss/train/total': full_loss / dist.get_world_size()})
```
Landanjs, line 1530: I thought total_loss should always be a dict, but pyright wasn't convinced. I added this check to appease pyright, does this look fine?

composer/trainer/trainer.py
```diff
@@ -1653,8 +1655,8 @@ def _train_batch(self, use_grad_scaling: bool):
                             total_loss = self.state.scaler.step(
                                 optimizer, closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs))
                         else:
-                            total_loss = optimizer.step(
-                                closure=lambda **kwargs: self._train_microbatches(microbatches, **kwargs).item())
+                            total_loss = optimizer.step(closure=lambda **kwargs: self._train_microbatches(
+                                microbatches, **kwargs))  # type: ignore
```
Landanjs, line 1659: Must closures exclusively return floats? Pyright says yes, is it okay to ignore this? The closure in this case will return a dict

abhi-mosaic: I think we should probably not be passing around the dictionary to any autograd / optimizer objects, we should just be passing them the total loss Tensor, because that's the only thing that should get `.backward()` called on it, and stuff like that, right? So for logging we pass around the dictionary, but not for any real optimization math, if that makes sense

Landanjs: I should read up on closures, but before this PR, `self._train_microbatches` returned `total_loss` which was a detached, cloned copy of the loss, so it didn't have any impact on optimization, right?

Also, if we need the closure to return a single loss value, I could do `self._train_microbatches(microbatches, **kwargs)['total']`, but `total_loss` would not contain the additional losses, making multiple loss logging impossible when using closures. Unless there is a pretty decent refactor of how we pass around `total_loss`

abhi-mosaic: Ok gotcha, I think I understand the issue a bit better now... so the thing that worries me a bit is that when optimizers are handed closures, I think they can potentially (e.g. SAM) have some logic inside that checks whether the return value is None, like they expect it to be a float or a None if something went wrong.

I think maybe what we need to do is make the loss stuff an explicit torchmetric (outside the closure) and have the metric.update(...) call happen inside the closure, for each key in the loss dict. Let's brainstorm tomorrow on this if possible!

Landanjs: I'll be out Wednesday, let's chat on Thursday if you will be in 🙂

eracah (commented):
''

mvpatel2000:
`I don't understand pyright` same :')

composer/trainer/trainer.py
```diff
@@ -1675,32 +1672,32 @@ def _train_batch(self, use_grad_scaling: bool):
...
             start_time = time.time()
-            total_loss = None
+            total_loss_dict = {'loss/train/total': self._device.tensor_to_device(torch.zeros(size=(1,)))}
             # Note: We use uint8 instead of bool as BOR is not supported on all torch.distributed backends
             should_handle_cuda_oom = 0
             try:
                 assert self.state.scaler is not None
                 microbatches = self._train_data_spec.split_batch(device_batch, self.state.grad_accum)
-                if self.deepspeed_enabled:
-                    total_loss = self._train_microbatches(microbatches)
```
Landanjs, line 1685: Does removing this and adding the if statement in `else:` work? To me, this makes this section a lot cleaner...

mvpatel2000: CC: @jbloxham  for deepspeed question

composer/trainer/trainer.py
```diff
@@ -1846,13 +1853,30 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int, tot
...
             self.engine.run_event(Event.BEFORE_BACKWARD)
 
-            # Sum individual losses
-            microbatch_loss = self._device.tensor_to_device(torch.zeros(size=(1,)))
-            for loss in ensure_tuple(self.state.loss):
-                microbatch_loss.add_(loss.mean())
+            microbatch_loss_dict = {}
+            # If total loss key is present, copy loss
+            if isinstance(self.state.loss, dict) and ('total' in self.state.loss):
+                microbatch_loss = self.state.loss['total']  # type: ignore
```
Landanjs, line 1859: I'm getting this error from pyright, any fixes?
```
/workdisk/landan/landan-composer/composer/trainer/trainer.py
  /workdisk/landan/landan-composer/composer/trainer/trainer.py:1859:35 - error: Argument of type "Literal['total']" cannot be assigned to parameter "indices" of type "_int | slice | Tensor | List[Unknown] | Tuple[Unknown, ...] | None" in function "__getitem__"
    Type "Literal['total']" cannot be assigned to type "_int | slice | Tensor | List[Unknown] | Tuple[Unknown, ...] | None"
      Type cannot be assigned to type "None"
      "Literal['total']" is incompatible with "_int"
      "Literal['total']" is incompatible with "slice"
      "Literal['total']" is incompatible with "Tensor"
      "Literal['total']" is incompatible with "List[Unknown]"
      "Literal['total']" is incompatible with "Tuple[Unknown, ...]" (reportGeneralTypeIssues)
  /workdisk/landan/landan-composer/composer/trainer/trainer.py:1859:35 - error: Argument of type "Literal['total']" cannot be assigned to parameter "index" of type "slice" in function "__getitem__"
    "Literal['total']" is incompatible with "slice" (reportGeneralTypeIssues)
```

composer/trainer/trainer.py
Landanjs: Refactored everything, resolving for the next review :)

abhi-mosaic (approved):
This looks great to me!!

@Landanjs I'm assuming you have some convergence tests on DeepLabV3 that work? And hopefully we have some trainer tests that do loss as a Tensor, tuple, dict? 

Approving now for velocity and trusting you on the tests :)

composer/trainer/trainer.py
```diff
@@ -1675,32 +1672,32 @@ def _train_batch(self, use_grad_scaling: bool):
...
+                                microbatches, total_loss_dict, **kwargs).item())
                 else:
-                    total_loss = self._train_microbatches(microbatches)
-                    for optimizer in self.state.optimizers:
-                        if use_grad_scaling:
-                            self.state.scaler.step(optimizer)
-                        else:
-                            if isinstance(self._device, DeviceTPU):
-                                xm.optimizer_step(optimizer, barrier=True)
+                    self._train_microbatches(microbatches, total_loss_dict)
```
abhi-mosaic, line 1691: should this be `total_loss = ...`?

Landanjs:
I added some convergence tests for tuple losses and dict losses with `'total'`. TBH I do not know how to make logging, but I'm running DeepLabv3 and ResNet experiments ([wandb here](https://wandb.ai/mosaic-ml/landan-test?workspace=user-landanjs) group by group). The results appear to be align with dev.

Landanjs:
I mentioned this to Evan, but just for the record, it looks like i need someone from @mosaicml/composer-team-eng to review as well

eracah (approved):

composer/trainer/trainer.py
```diff
@@ -1724,32 +1723,32 @@ def _train_batch(self, use_grad_scaling: bool):
         # Retry until we successfully complete training and return loss
```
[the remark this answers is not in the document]
Landanjs: For the docstring, would this work:

```
Returns:
     Dict[str, torch.Tensor]: a dictionary containing the total loss and individual losses if available
```

PR #1376, opened 2022-08-06
Fix AMP resumption with grad scaler

hanlint:
Fixes a bug where, for `amp`, the GradScaler's `_growth_factor` would not have been resumed properly. This was because `state.scaler` was not created _before_ the `load_checkpoint` call, so the `GradScaler` state would not have been loaded properly.

This fix moves the grad scaler creation from the `_train_loop` -> `__init__`.


composer/trainer/trainer.py
```diff
@@ -1343,6 +1348,9 @@ def fit(
             _validate_precision(precision, self._device, self.deepspeed_enabled)
             self.state.precision = precision
 
+            # update scaler since precision was provided
+            self.state.scaler = ClosureGradScaler() if self._use_closures() else GradScaler()
```
ravi-mosaicml, line 1352: This would wipe the grad scaler state if one sets the precision in `fit()`, even if it is the same as in `__init__`. Should this line be guarded with a `if precision != state.precision:` check?

ravi-mosaicml, line 1352: In addition, if `reset_time` is set, would it make sense to also wipe the grad scaler state?

tests/algorithms/test_algorithm_resumption.py
```diff
@@ -42,7 +42,7 @@ def test_algorithm_resumption(
         'max_duration': '2ep',
         'save_filename': 'ep{epoch}-rank{rank}',
         'train_subset_num_batches': 2,
-        'precision': 'fp32',
+        'precision': 'amp',
```
ravi-mosaicml, line 45: Would it work to add a seperate test case to test that the grad scaler state is being properly restored? (I presume this change was made here to test that, but would prefer to have a separate test in `test_trainer` if that's not too much work.)

ravi-mosaicml (approved):
LGTM; see comments. Thanks for fixing this!

tests/algorithms/test_algorithm_resumption.py
hanlint: Actually, this is reverting a change made in https://github.com/mosaicml/composer/pull/1369.

composer/trainer/trainer.py
hanlint: That sounds unnecessary.

PR #1379, opened 2022-08-08
Remove Unused Exception

mvpatel2000:
Reopening since the last got automerged. As title. Diff is large, but I just removed the exception and unindented. We can add this back if we want to use it, but I'm pro simplifying code and not leaving in dead snippets. Fixes CO-864

From last PR:
> Hmm I believe we needed this for the Benchmarker code to work, cc: @abhi-mosaic ? Is that still used for data collection?


abhi-mosaic:
I believe this was mostly used for the `Benchmarker`, but I also don't see that callback in the codebase anymore... so I guess it's OK to remove for now. I think since we now can modify `state.timestamp` we have other ways of quitting an epoch early.

PR #1380, opened 2022-08-08
Auto Grad Accum Cache Clearing

mvpatel2000:
Adds cache clearing to auto grad accum, which should help reduce the fragmentation issues people have been seeing. Used a [simple function](https://github.com/pytorch/pytorch/issues/29554) which approximates fragmentation. Picture below shows fragmentation dropping as we call cache clear
```
def compute_memory_fragmentation():
                            snapshot = torch.cuda.memory_snapshot()
                            return sum(b['allocated_size'] for b in snapshot) / sum(b['total_size'] for b in snapshot)
```
image


hanlint (approved):
This looks good to me, thanks!

PR #1385, opened 2022-08-09
Cleaner API reference part 1: references with minimal import paths

dblalock:
To address [CO-789](https://mosaicml.atlassian.net/browse/CO-789) (showing preferred import paths in the docs rather than private ones), we need to change our internal references to not break when the paths change. E.g., we need to replace 
```rst
:class:`~composer.algorithms.blurpool.blurpool.BlurPool`
```
with 
```rst
:class:`.BlurPool`
```
so that, when we have the docs just show `composer.algorithms.BlurPool`, the reference still works.

This PR touches a lot of files but is 90% just the result of `sed` commands, with a little bit of manual correction for isolated cases. Also I fixed a typo in [docs/source/composer_model.rst](https://github.com/mosaicml/composer/pull/1385/files#diff-fd86e5afd002ce9ef1291de49c2a22b77e6e303e582a8452753ad5135661c22f) and a missing reference in [docs/source/trainer/using_the_trainer.rst](https://github.com/mosaicml/composer/pull/1385/files#diff-5f973a074ae8beaa91023affdd5e757f71702611ae61005cbdff5c8ce090cf49).

### DIFF
--- a/composer/algorithms/__init__.py
+++ b/composer/algorithms/__init__.py
@@ -34,7 +34,7 @@ For example, a simple algorithm that shortens training:
         def apply(self, state: State, event: Event, logger: Logger):
             state.max_duration /= 2  # cut training time in half
 
-For more information about events, see :class:`~composer.core.event.Event`.
+For more information about events, see :class:`.Event`.
 """
 
 from composer.algorithms.alibi import Alibi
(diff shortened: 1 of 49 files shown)


hanlint:
Merging this now so other contributors working on documentation can base off of this commit.

PR #1387, opened 2022-08-09
Print the microbatch size when using Adaptive Gradient Accumulation

hanlint:
Unsure if microbatch size is a well known term but this information is helpful for folks setting the `eval_batch_size`.


hanlint:
Also, is the batch _retrained_, or is it _skipped_?

mvpatel2000:
> Also, is the batch _retrained_, or is it _skipped_?

It's retrained

mvpatel2000 (approved):
Lgtm.

PR #1388, opened 2022-08-09
Add Event.BEFORE_DATALOADER

mvpatel2000:
Currently, the only point in which `Profiler` adds a marker outside the engine is for dataloader yields. This is messy agent wise because accessing dataloader yield times (for checking if we're dataloader bottlenecked) requires monkeypatching `Profiler`. 

In order to simplify the control plane, we've now added `Event.BEFORE_DATALOADER`. The engine should have enough entry points to control the entire flow without relying on anything else, and this new event helps complete that.

More broadly, `Profiler` is really just a wrapper around inserting callbacks. We should refactor away `Profiler` into callbacks so each trace handler can have its own schedule. This is a broader redesign though that's lower priority since it's mostly to clean up Composer internals as opposed to user-facing changes.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -31,7 +31,7 @@ from composer.core import (Algorithm, Callback, DataSpec, Engine, Evaluator, Eve
                            ensure_data_spec, ensure_evaluator, ensure_time)
 from composer.core.precision import get_precision_context
 from composer.core.time import TimeUnit
-from composer.core.types import Batch, PyTorchScheduler
+from composer.core.types import Batch, PyTorchScheduler, TrainerMode
 from composer.loggers import Logger, LoggerDestination, LogLevel, ProgressBarLogger
 from composer.models.base import ComposerModel
 from composer.optim.decoupled_weight_decay import DecoupledSGDW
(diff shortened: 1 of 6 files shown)


mvpatel2000:
image
Picture of updated traces showing profiling works with new event format

composer/trainer/trainer.py
```diff
@@ -2156,15 +2156,16 @@ def _iter_dataloader(self):
...
             try:
+                # [BEFORE/AFTER]_DATALOADER only runs while training
+                if trainer_mode == TrainerMode.TRAIN:
+                    self.engine.run_event(Event.BEFORE_DATALOADER)
                 batch = next(dataloader_iter)
             except StopIteration:
+                # [BEFORE/AFTER]_DATALOADER only runs while training
+                if trainer_mode == TrainerMode.TRAIN:
+                    self.engine.run_marker_only_event(Event.AFTER_DATALOADER)
                 break
```
hanlint, line 2168: A little confused by this.. what happened to: 
https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L1513

(the run_event(AFTER_DATALOADER)), and why is the marker here? is it to more precisely measure the yield time?

mvpatel2000: Added a comment -- sorry this was unclear. Basically, in the yield where we dont actually get anything (end of loop), we need to have a corresponding marker.finish() call

hanlint (approved):
LGTM, thanks for the clarification!

PR #1389, opened 2022-08-09
remove private s3 paths

A-Jacobson:
removing links to private s3 buckets in yamls/scripts. The proper s3 paths will have to be injected through mcli for internal use.

Closes CO-397


ravi-mosaicml (approved):
LGTM; thanks!

PR #1390, opened 2022-08-10
Decouple GLUE entry point saving and loading logic 

ishanashastri:
Currently, the GLUE entry point doesn't allow one to load from object store but save locally -- the saving and loading logic are tied to each other. This is a use case for our open-source users who may want to load a pre-trained checkpoint from object store but do not possess the credentials to write back to object store.

This PR removes this coupling, allowing for as much flexibility as the original composer entry point allows for.


lupesko (changes requested):
LGTM.

Two questions though before approving:
(1) Did you test this locally with the new load locally functionality? It is always good to state this explicitly in the PR description.
(2) Do we have any test coverage for this entry point? I was hoping to see a new/updated test with this PR that confirms the change is tested. If we do not have a test we probably need to add one, even if basic, so that this entry point is not accidentally broken in the future. You can also add test coverage in another PR if you prefer.

ishanashastri:
> (1) Did you test this locally with the new load locally functionality? It is always good to state this explicitly in the PR description.
Yes, tested locally with the new functionality. Will update the PR description!

> (2) Do we have any test coverage for this entry point? I was hoping to see a new/updated test with this PR that confirms the change is tested. If we do not have a test we probably need to add one, even if basic, so that this entry point is not accidentally broken in the future. You can also add test coverage in another PR if you prefer.
Discussed with others offline and decided that unit tests were not necessary for v0 of the entrypoint since most of the tests would be on argparsing correctly. Will file a JIRA to make a more thorough testing scheme for this entrypoint.

lupesko (approved):

bandish-shah (approved):
LGTM

[nit] Please move the todo to a Jira issue and remove the todo from the code base before merging.

PR #1391, opened 2022-08-10
Only sync batchnorm statistics within a node for deeplab

Landanjs:
Two adjustments to sync BN in deeplabv3:
1. **Necessary**: Ignore `sync_bn` when only one process is present. This avoids an ambiguous error being thrown
2. **Optional**: Only sync BN with ranks within a node. 
    - **Pro**: Improves throughput by ~5% on a two node system and should slightly improve convergence. Might have a bigger impact on throughput with larger systems.
    - **Cons**: 
        - For large number of nodes (I think 16 nodes with 8 gpus), syncing within node would be insufficient. 
        - batch norm batch size will change as the number of nodes increases even with sync BN. This could improve or hurt performance depending on the batch size and number of GPUs
        - Adds some complexity to code

Somewhat random thought: the second option is a simpler case for a more general feature which would be adjusting batch norm batch size to be a particular size using ghost BN and sync BN. But at that point, could we just use weight standardization + group norm and tune the group norm hparam? This would have nicer properties such as reducing train/val discrepancy and issues with BN transfer learning.


composer/models/deeplabv3/model.py
```diff
@@ -94,7 +97,12 @@ def deeplabv3(num_classes: int,
              https://download.openmmlab.com/mmcv/dist/{cu_version}/{torch_version}/index.html where {cu_version} and
              {torch_version} refer to your CUDA and PyTorch versions, respectively. To install mmsegmentation, please
              run pip install mmsegmentation==0.22.0 on command-line.""")) from e
-    norm_type = 'SyncBN' if sync_bn else 'BN'
+
+    world_size = dist.get_world_size()
+    if sync_bn and world_size == 1:
+        warnings.warn('sync_bn was true, but only one process is present for training. sync_bn will be ignored.')
+
+    norm_type = 'SyncBN' if sync_bn and world_size > 1 else 'BN'
```
hanlint, line 105: if the `norm_cfg` is already set to `SyncBN`, do we still need to run the `convert_sync_batchnorm` call below?

composer/models/deeplabv3/model.py
```diff
@@ -133,8 +141,20 @@ def deeplabv3(num_classes: int,
...
 
-    if sync_bn:
-        model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model)
+    if sync_bn and world_size > 1:
+        local_world_size = dist.get_local_world_size()
+
+        # List of ranks for each node, assumes that each node has the same number of ranks
+        num_nodes = world_size // local_world_size
+        process_group = None
+        if num_nodes > 1:
```
hanlint, line 150: instead of creating all the process groups for each node, and then selecting based on the node_rank, can this be simplified to:
```
node_rank = dist.get_node_rank()
ranks = list(range(node_rank * local_world_size, (node_rank + 1) * local_world_size))
process_group = torch.dist.new_group(ranks)
...
```
```

File: composer/models/deeplabv3/model.py (hunk 1 above)
Comment (Landanjs) [reply composer/models/deeplabv3/model.py:105]: `norm_cfg` is only for the head of network i.e. DeepLabv3+. The `convert_sync_batchnorm` is primarily for the backbone, i.e. ResNet. It may be the case that we only need `convert_sync_batchnorm`, but since mmseg has a lot of their own bespoke layers, I'm not sure `convert_sync_batchnorm` detects the custom layers. I can test this if needed.

Comment (hanlint) [reply composer/models/deeplabv3/model.py:105]: ok thanks. not neccessary, better to be safe I suppose!

File: composer/models/deeplabv3/model.py (hunk 2 above)
Comment (Landanjs) [reply composer/models/deeplabv3/model.py:150]: Yeah, that should work haha. I think I tried to follow the pytorch example too closely, thanks!

**Comment by dskhudia** [comment]
> For large number of nodes (I think 16 nodes with 8 gpus), syncing within node would be insufficient.

Do you expect batch size per node to shrink as you scale?

**Comment by Landanjs** [comment]
It is TBD if convergence stays the same when doubling batch size and learning rate only. I saw a slight degradation at double the current batch size, but unsure if this was due to the batch norm batch size or due to worse optimization. Once we have more options for multinode, we could test larger batch sizes :)

File: composer/models/deeplabv3/model.py (hunk 2 above)
Comment (Landanjs) [reply composer/models/deeplabv3/model.py]: Hmmm, I'm getting a weird NCCL error with the change. Maybe torch.distributed needs to know that all of the processes are assigned to certain groups even in each process? Not really sure.

```
  File "/composer/examples/run_composer_trainer.py", line 78, in <module>
    _main()
  File "/composer/examples/run_composer_trainer.py", line 74, in _main
    trainer.fit()
  File "/usr/local/lib/python3.9/dist-packages/composer/trainer/trainer.py", line 1354, in fit
    self._train_loop()
  File "/usr/local/lib/python3.9/dist-packages/composer/trainer/trainer.py", line 1521, in _train_loop
    total_loss = self._train_batch(use_grad_scaling)
  File "/usr/local/lib/python3.9/dist-packages/composer/trainer/trainer.py", line 1664, in _train_batch
    total_loss = self._train_microbatches(microbatches)
  File "/usr/local/lib/python3.9/dist-packages/composer/trainer/trainer.py", line 1761, in _train_microbatches
    self._train_microbatch(use_grad_scaling, current_batch_size, total_loss, is_final_microbatch)
  File "/usr/local/lib/python3.9/dist-packages/composer/trainer/trainer.py", line 1798, in _train_microbatch
    self.state.outputs = self.state.model(self.state.batch)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/parallel/distributed.py", line 1008, in forward
    output = self._run_ddp_forward(*inputs, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/parallel/distributed.py", line 971, in _run_ddp_forward
    return module_to_run(*inputs, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/composer/models/tasks/classification.py", line 89, in forward
    outputs = self.module(inputs)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/composer/models/deeplabv3/model.py", line 35, in forward
    features = self.backbone(x)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/torchvision/models/_utils.py", line 69, in forward
    x = module(x)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/batchnorm.py", line 748, in forward
    return sync_batch_norm.apply(
  File "/usr/local/lib/python3.9/dist-packages/torch/nn/modules/_functions.py", line 56, in forward
    dist._all_gather_base(combined_flat, combined, process_group, async_op=False)
  File "/usr/local/lib/python3.9/dist-packages/torch/distributed/distributed_c10d.py", line 2136, in _all_gather_base
    work = group._allgather_base(output_tensor, input_tensor)
RuntimeError: NCCL error in: ../torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:1191, unhandled system error, NCCL version 2.10.3
ncclSystemError: System call (e.g. socket, malloc) or external library call failed or device error. It can be also caused by unexpected exit of a remote peer, you can check NCCL warnings for failure reason and see if there is connection closure by a peer.
```

hanlint: odd, ok. this is just a nit, so i'm OK merging in its current form.

PR #1393, opened 2022-08-10
Add remove cols last

mvpatel2000:
For autoperf, we'll need ways to start enabling and disabling methods. This will happen with the functional API since it'll run post-init. 

We're going to need to start adding remove methods


composer/algorithms/channels_last/channels_last.py
```diff
@@ -30,6 +30,18 @@ def apply_channels_last(model: torch.nn.Module) -> None:
     model.to(memory_format=torch.channels_last)  # type: ignore
 
 
+def remove_channels_last(model: torch.nn.Module) -> None:
```
hanlint, line 33: A few considerations here:
* should this be a functional API, or a method of the `Algorithm` class? e.g. in case some algorithms may need to store some state to know how to revert
* I am assuming the remove should only apply changes that were done on `INIT`? And assume that subsequently, the algorithm itself is removed from running?

mvpatel2000: Lets discuss offline -- I think we might need to rework parts of the Algorithm class to be reversible

mvpatel2000: The former I think the answer is yes.

The latter also yes

dskhudia:
I have doubts about going the route of `remove_<algo>`. The main reason:

We won't be able do it for all the algorithms. It may be ok to do this for certain algorithms. I just want to make sure that this is a conscious decision and we are not building agent based on the assumption that algorithms will be reversible. An alternative is recreating the model and applying algorithm(s) again. However, for this route interaction with optimizer is something we will need to think about carefully. 

For large models, if memory is an issue we can use `meta` device so that we [don't materialize weight tensors](https://pytorch.org/tutorials/prototype/skip_param_init.html#skipping-module-parameter-initialization).

mvpatel2000:
Closing based on offline discussion

PR #1394, opened 2022-08-10
Fix hparams error

mvpatel2000:
Hparams by default instantiates `train_subset_num_batches` to -1. However, we only check if its not None when giving a warning

### DIFF
--- a/composer/trainer/trainer_hparams.py
+++ b/composer/trainer/trainer_hparams.py
@@ -86,10 +86,9 @@ def _initialize_dataloader(
 
         train_device_batch_size = batch_size // dist.get_world_size()
         if dataset_hparams.shuffle and subset_num_batches is not None and subset_num_batches != -1:
-            warnings.warn(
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer_hparams.py
```diff
@@ -85,7 +85,7 @@ def _initialize_dataloader(
             )
 
         train_device_batch_size = batch_size // dist.get_world_size()
-        if dataset_hparams.shuffle and subset_num_batches is not None:
+        if dataset_hparams.shuffle and subset_num_batches is not None and subset_num_batches != -1:
             warnings.warn(
                 (f'SubsetNumBatchesWarning: When specifying `subset_num_batches` for the {dataloader_label} dataset, '
```
dskhudia, line 90: Should this be `[train|eval]_subset_num_batches` in the warning as for a user there is not variable named subset_num_batches?

mvpatel2000: yes let me update

dskhudia (approved):

PR #1396, opened 2022-08-11
Expand evaluation doc

alextrott16:
Uses evaluation logic pseudocode to provide a better intuition for the logic and the relationship between `ComposerModel.validate` and `ComposerModel.metrics`.

Adds some exposition around using the `Evaluators` class.

Closes CO-861


alextrott16:
@hanlint Not sure if this addresses all the aspects of CO-861, but AFAICT the `validate` method should be the same regardless of what style of `metric` it feeds into (i.e., the output tuple from `validate` will always be used as the positional arguments for `metric.update` regardless of whether it represents a single metric or a list-style or dict-style metric collection).

docs/source/trainer/evaluation.rst
```diff
@@ -59,3 +79,12 @@ can be specified as in the following example:
 
 In this case, the metrics from :meth:`.ComposerModel.metrics` will be ignored
 since they are explicitly provided above.
+
+Note that a single :class:`.Evaluator` can be also be passed to ``eval_dataloader``.
```
hanlint, line 83: I think this section is a bit confusing. The above example already shows passing evaluator objects to eval_dataloader, so why is the evaluation logic different? 

Instead, maybe something like: Note that for multiple evaluators, the interface between validate and metrics is the same, we call:
```
output, targets = model.validate(batch)
for evaluator in evaluators:
    evaluator.metrics.update(output, targets)
    
```

hanlint (commented):
one comment, perhaps that last section is unneeded? Otherwise, looks good to me.

docs/source/trainer/evaluation.rst
alextrott16: I might be a bit confused here. 

The example in the first section doesn't use a `Evaluator` explicitly -- it shows a dataloader being passed to `eval_dataloader`. So, I figured it might not be clear how the single-dataloader vs multiple-evaluators logics compare.

Regarding the multiple-evaluators logic: isn't a separate evaluation loop invoked for each `Evaluator`? My understanding is that such is necessary because different evaluators can have different underlying datasets. So you would need the `for evaluator in evaluators` to be the outer loop rather than the inner loop.

Also it looks like the `Evaluator` class changed, so is there even `evaluator.metrics` anymore?

I'm happy to clear up the examples, but want to make sure I'm properly understanding first!

hanlint (approved):

PR #1399, opened 2022-08-11
Update DeepLabv3 pretrained weight interface to work with PyTorch 1.12

Landanjs:
This is supposed to be similar to the ResNet PR here: https://github.com/mosaicml/composer/pull/1262. 

The `is_backbone_pretrained` and `backbone_url` are replaced in favor of `backbone_weights` which uses PyTorch's new multiple weights API. When torchvision is below 0.13, the respective urls for `IMAGENET1K_V1` and `IMAGENET1K_V2` are hardcoded.

For the future, we need a better interface to be able to use arbitrary pre-trained weights i.e. be able to use weights locally or ones stored in the cloud. I just wanted to get a quick fix to use PyTorch 1.12, but will create a jira for this feature

### DIFF
--- a/composer/models/deeplabv3/README.md
+++ b/composer/models/deeplabv3/README.md
@@ -11,8 +11,7 @@ from composer.models import composer_deeplabv3
 
 model = composer_deeplabv3(num_classes=150,
                            backbone_arch="resnet101",
-                           is_backbone_pretrained=True,
-                           backbone_url="https://download.pytorch.org/models/resnet101-cd907fc2.pth",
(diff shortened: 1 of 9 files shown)


hanlint (approved):
lgtm!

hanlint:
Perhaps have someone in vision cc: @A-Jacobson take a quick look.

Landanjs:
I can't interpret why the test failed. Is something abnormal happening or am I missing the failure?

A-Jacobson (approved):
hey! looks good to me. like the new pytorch weight format much better.

PR #1401, opened 2022-08-12
Add informative error for infer batch size issues

hanlint:
This PR:
* adds an informative error if the batch yielded by the dataloader has a non-Tensor element (e.g. cannot retrieve the shape)
* improves test coverage of the `_default_num_samples_in_batch` function.

### DIFF
--- a/tests/trainer/test_dataspec.py
+++ b/tests/trainer/test_dataspec.py
@@ -27,6 +27,7 @@ class TestDefaultGetNumSamples:
         (torch.rand(N, 8), torch.rand(N, 64)),  # tuple
         [torch.rand(N, 8), torch.rand(N, 64)],  # list
         torch.rand(N, 8),  # tensor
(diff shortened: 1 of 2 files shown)


tests/trainer/test_dataspec.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestDefaultGetNumSamples:
+
+    @pytest.fixture
+    def dataspec(self):
+        dataloader = DataLoader(RandomClassificationDataset())
+        return DataSpec(dataloader=dataloader)
+
+    # yapf: disable
+    @pytest.mark.parametrize('batch', [
```
dskhudia, line 24: Does it work on > 2-d tensors?

dskhudia (approved):

tests/trainer/test_dataspec.py
hanlint: added a 4-dim tensor test!

PR #1402, opened 2022-08-12
current metrics docs

A-Jacobson:
added a line about getting current eval metrics from the trainer after training.

stopgap for CO-756. Metrics can also be accessed from the composermodel, but i believe this is the most logical way to access them.

### DIFF
--- a/docs/source/trainer/using_the_trainer.rst
+++ b/docs/source/trainer/using_the_trainer.rst
@@ -78,9 +78,14 @@ objects.
 
     trainer.fit()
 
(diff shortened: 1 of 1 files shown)


hanlint (approved):
LGTM

docs/source/trainer/using_the_trainer.rst
```diff
@@ -78,10 +78,15 @@ objects.
 
     trainer.fit()
 
+When training is complete, metrics can be accessed from the trainer state.
```
hanlint, line 81: `trainer` should be :class:`.Trainer` instead

PR #1403, opened 2022-08-12
add install from dev instructions

A-Jacobson:
install from dev instructions for all notebooks


examples/glue/glue_example.yaml
```diff
@@ -33,7 +33,7 @@ pretrain_hparams:
         label: bert_pre_training
         eval_dataset:
           streaming_c4:
-            remote: s3://allenai-c4/mds/1/
+            remote: my-s3-bucket-name
```
alextrott16, line 36: Since this change seems a bit out of scope, I'm wondering what the motivation is? Also are we sure we don't want to point to the public C4 bucket in our basic GLUE example?

At the very least I might add a comment here flagging that the user needs to set up the remote data source and replace the placeholder.

alextrott16 (commented):
I left a comment about the GLUE YAML, but otherwise looks good!

examples/glue/glue_example.yaml
A-Jacobson: Whoops this is a mistake guess I had some carried over changes from another branch. Nice catch

alextrott16 (approved):

hanlint (approved):

hanlint:
@A-Jacobson some conflicts to resolve, but otherwise ready to merge

PR #1406, opened 2022-08-12
merge nlp+hf notebooks

A-Jacobson:
by "merge" we mean delete the nlp notebook. the HF notebook does the same thing as the nlpnotebook and is now our recommended path to train with HF models.


alextrott16 (approved):
LGTM!

hanlint (commented):
Is there anything from the NLP notebook that we want to transfer over?

A-Jacobson:
I think we're good with this. I transferred the tensorboard logging in this pr. The only difference between the two notebooks are that this one uses the huggingfacemodel class.

hanlint (approved):

PR #1407, opened 2022-08-12
Consistent Method Card Style

growlix:
This is the first pass through the Method Cards, making sure they meet the specifications as listed [here](https://www.notion.so/Documentation-Sprint-August-2022-09b39ac2ed0c4f5abb950c5ce0f465e0). I was able to implement most fixes myself, but I assigned a few larger issues (e.g. example images, working code) to the method cards owners.

Updates:
 - I merged the API link update into this PR (see below)
 - i added the decoupled weight decay method card

The following are _not_ addressed in this PR:
- ~update method cards to have links to the API ([[Jira](https://mosaicml.atlassian.net/browse/CO-787)](https://mosaicml.atlassian.net/browse/CO-787))~ I merged the API link update into this PR
- update method cards with warnings if they may have different behavior on different GPUs of a multi-GPU workload ([[Jira](https://mosaicml.atlassian.net/browse/CO-409)](https://mosaicml.atlassian.net/browse/CO-409))
- ~adding _new_ method cards~ I added decoupled weight decay


dskhudia:
Are we going with `Mathematically Equivalent` ? Can't think of a better short name but something like `convergence preserving` would be more accurate.

jfrankle:
Let's stick with mathematically equivalent for now as a placeholder for
something better as we come up with more ideas.

growlix:
> Are we going with `Mathematically Equivalent` ? Can't think of a better short name but something like `convergence preserving` would be more accurate.

Good question. I feel like `Convergence Preserving` is more accurate but slightly more jargony.  Do people have strong preferences from {`Convergence`, `Math`, `Mathematically`}, {`Preserving`, `Equivalent`}?

@hanlint @jfrankle

hanlint:
I think `MathEq` is fine, we can just define it somewhere (e.g. in the algorithms table). Favor readability / accessibility over technical accuracy here.

hanlint:
I don't like "Changes the Training Algorithm" though, that's pretty vague and hard to understand.

growlix:
> I don't like "Changes the Training Algorithm" though, that's wordy, vague, and hard to understand. Can we just exclude that tag? Anything that doesn't have the `MathEq` tag is by default changing the math?

Sounds good to me!

composer/algorithms/fused_layernorm/README.md
```diff
@@ -73,8 +73,18 @@ Fused LayerNorm does not have any hyperparameters. It utilizes the existing `nor
 
 APEX's FusedLayerNorm achieves a substantial speedup over PyTorch by doing a few things:
 1. Instead of a naive implementation, which requires two passes over the input in order to estimate variances, it uses [Welford's Online Algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) to estimate the variances in a single step, creating a substantive wall-clock speedup.
-2. Instead of requiring multiple CUDA kernel launches, it computes everything in a single kernel launch, therefore improving GPU utilization.
+2. Instead of requiring multiple CUDA kernel launches, it computes everything in a single
+   kernel launch, therefore improving GPU utilization.
```
dblalock, line 77:
```suggestion
2. Instead of requiring multiple CUDA kernel launches, it computes everything in a single
   kernel launch, improving GPU utilization.
```

composer/algorithms/ghost_batchnorm/README.md
```diff
@@ -93,17 +93,13 @@ The Composer implementation of GhostBatchNorm uses model surgery to replace `Bat
...
 >
 > This yields slightly different mean and variance statistics compared to using a normal BatchNorm module.
 > The difference stems from the moving average over a sequence of groups not being equal to the true average of the groups.
 
 For small ghost batch sizes, this method might run more slowly than normal batch normalization. This is because our implementation uses a number of operations proportional to the number of ghost batches, and each PyTorch operation has a small amount of overhead. This overhead is inconsequential when doing large chunks of “work” per operation (i.e., operating on large inputs), but can matter when the inputs are small.
 
->❗ Ghost BatchNorm Slows Down Training
->
-> We observed throughput decreases of around 5% fewer samles per second for ResNet-50 on ImageNet.
-
```
dblalock, line 106: Is this not true?

composer/algorithms/mixup/README.md
```diff
@@ -169,7 +174,13 @@ Doing so avoids putting additional work on the CPU (since augmentation occurs on
...
 >❗ When `interpolate_loss=True` MixUp interpolates the loss rather than the targets.
 >
-> This is fine for loss functions that are linear in the targets, such as cross entropy, but may produce unexpected results for other loss functions.
+> This is fine for loss functions that are linear in the targets, such as cross entropy,
+> but may produce unexpected results for other loss functions.
+
+> 🚧 MixUp May Interact with Other Methods that Modify Targets
+>
+> This method interacts with other methods (such as Label Smoothing) that alter the targets.
+> While such methods may still compose well with label smoothing in terms of improved accuracy, it is important to ensure that the implementations of these methods compose.
```
dblalock, line 183: Does composer not take care of this? wording doesn't make clear what I have to do differently / worry about as a user

composer/algorithms/randaugment/README.md
```diff
@@ -102,11 +102,11 @@ The class form of RandAugment runs on `Event.FIT_START` and inserts `RandAugment
...
+As per [Cubuk et al. (2020)](https://openaccess.thecvf.com/content_CVPRW_2020/html/w40/Cubuk_Randaugment_Practical_Automated_Data_Augmentation_With_a_Reduced_Search_Space_CVPRW_2020_paper.html), we found that `depth=1` (applying a chain of two augmentations to each image) and `severity=9` (each augmentation is applied quite strongly) works well for different models of the ResNet family on ImageNet. For `depth≥3`, we find diminishing accuracy gains (due to over-regularization) and substantial training slowdown (due to the CPU becoming a bottleneck because of the amount of augmentation it must perform). We also recommend `augmentation_set=all` (using all available augmentation techniques).
```
dblalock, line 105:
```suggestion
As per [Cubuk et al. (2020)](https://openaccess.thecvf.com/content_CVPRW_2020/html/w40/Cubuk_Randaugment_Practical_Automated_Data_Augmentation_With_a_Reduced_Search_Space_CVPRW_2020_paper.html), we found that `depth=1` (applying a single augmentation to each image) and `severity=9` (each augmentation is applied quite strongly) works well for different models of the ResNet family on ImageNet. For `depth≥3`, we find diminishing accuracy gains (due to over-regularization) and substantial training slowdown (due to the CPU becoming a bottleneck because of the amount of augmentation it must perform). We also recommend `augmentation_set=all` (using all available augmentation techniques).
```

docs/source/method_cards/decoupled_weight_decay.md
```diff
@@ -4,50 +4,68 @@
 
 # 🏋️‍♀️ Decoupled Weight Decay
 
-Tags: `Best Practice`, `Increased Accuracy`, `Regularization`
-
-## TL;DR
+[\[How to Use\]](#how-to-use) - [\[Suggested
+Hyperparameters\]](#suggested-hyperparameters) - [\[Technical
+Details\]](#technical-details) - [\[Attribution\]](#attribution) - [\[API Reference\]](#api-reference) - [\[API Reference\]](#api-reference)
```
dblalock, line 9:
```suggestion
Details\]](#technical-details) - [\[Attribution\]](#attribution) - [\[API Reference\]](#api-reference)
```

docs/source/method_cards/decoupled_weight_decay.md
```diff
@@ -4,50 +4,68 @@
...
 
-## Considerations
+## Suggested Hyperparameters
 
-There are no known negative side effects to using decoupled weight decay once it is properly tuned, as long as the original base optimizer is either `torch.optim.Adam` or `torch.optim.SGD`.
+Optimizers with decoupled weight decay can be used as drop-in replacements for their
+non-decoupled counterparts, however the optimal `weight_decay` value for decoupled
+optimizers will typically be larger than for their non-decoupled counterparts, because
+decoupled weight decay is not scaled by the learning rate.
+## Technical Details
```
dblalock, line 63:
```suggestion
Optimizers with decoupled weight decay can be used as drop-in replacements for their
non-decoupled counterparts. However, the optimal `weight_decay` value for decoupled
optimizers will typically be larger than for their non-decoupled counterparts, because
decoupled weight decay is not scaled by the learning rate.
## Technical Details
```
Also, is it larger values and not smaller? With an lr of .01 I would expect the weight decay to be equivalent when 100x smaller, but I also didn't think about this too carefully just now.

docs/source/method_cards/decoupled_weight_decay.md
```diff
@@ -4,50 +4,68 @@
...
 
-## Composability
+There are no known negative side effects to using decoupled weight decay once it is properly tuned, as long as the original base optimizer is either `torch.optim.Adam` or `torch.optim.SGD`.
 
 Weight decay is a regularization technique and thus is expected to yield diminishing returns when composed with other regularization techniques.
+
+## Attribution
+
+*[Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)*, by Ilya Loshchilov and Frank Hutter. Published as a conference paper at ICLR 2019.
```
dblalock, line 71: Missing API reference below this?

dblalock (commented):
I mostly just looked at my methods + prose. Only real issue is that API reference links are all broken, since we no longer autogenerate a docs page for each algorithm directory. Needs links to the Algorithm and functional API directly.

composer/algorithms/cutmix/README.md
```diff
@@ -101,7 +102,7 @@ CutMix is intended to improve generalization performance, and we empirically fou
 
 > 🚧 Composing Regularization Methods
 >
-> As general rule, composing regularization methods may lead to diminishing returns in quality improvements. CutMix is one such regularization method.
+> As general rule, composing regularization methods may lead to diminishing returns in quality improvements. CutMix is one such regularization method. We do not see improvements when combining CutMix with other regularization and augmentation methods such as Mixup and Label Smoothing.
```
coryMosaicML, line 105: I don't think this is true? Ex: over the vanilla ResNet50 baseline, I think CutMix + Label Smoothing gives an improvement. And although we can't do it, it's pretty common to use CutMix + Mixup together in other recipes, alternating between the two, which is supposed to give an improvement

coryMosaicML (approved):
Looks good to me! Mostly looked over changes to the things I'm familiar with.

moinnadeem (approved):
LGTM! Could we add a comment to the Scale Schedule Ratio that it is only required on the vision models?

dblalock:
> nit: could we add a comment to the Scale Schedule Ratio that it is only required on the vision models?

@moinnadeem, don't the chinchilla results contradict this? Or really, would it be safer to just make the more specific statement that this isn't necessary when using a constant LR?

composer/algorithms/blurpool/README.md
```diff
@@ -1,6 +1,8 @@
 # 🏊 BlurPool
 
-[\[How to Use\]](#how-to-use) - [\[Suggested Hyperparameters\]](#suggested-hyperparameters) - [\[Technical Details\]](#technical-details) - [\[Attribution\]](#attribution)
+[\[How to Use\]](#how-to-use) - [\[Suggested
+Hyperparameters\]](#suggested-hyperparameters) - [\[Technical
+Details\]](#technical-details) - [\[Attribution\]](#attribution) - [\[API Reference\]](#api-reference) - [\[API Reference\]](#api-reference)
```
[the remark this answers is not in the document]
growlix: I think it's just here, but checking every one to be sure. Good catch!

composer/algorithms/ghost_batchnorm/README.md
growlix: Not sure why that was removed! Reverted, ty!

composer/algorithms/mixup/README.md
growlix: Good question. I saw it in the label smoothing readme, so I added it to this. @coryMosaicML, do you have any insights here?

docs/source/method_cards/decoupled_weight_decay.md
growlix: It's definitely smaller, good catch

composer/algorithms/mixup/README.md
coryMosaicML: Composer takes care of this if everything is run with one hot rather than index labels. However, by popular demand MixUp and CutMix have an index label mode that can be used. The bad interaction here is when MixUp and CutMix are both used in index label mode (`interpolate_loss=True`), which will not compose correctly. There used to be a warning that got raised if someone tried to do this, but it looks like it got removed at some point?!

alextrott16 (commented):
ALiBi and SLW look good to me. I'll forgo a formal "Approve" since I only reviewed 2, but please consider my approval given.

Landanjs:
When looking at the API Reference I see this:

image

This isn't right, right?

dskhudia:
Since others have already reviewed, I am just rubber stamping it.

dskhudia (approved):

PR #1409, opened 2022-08-12
Upgrade to torch 1.12.1

abhi-mosaic:
1.12.0 -> 1.12.1 comes with some bugfixes, particularly Adam optimizer resumption, that will be good to have.


bandish-shah (changes requested):
For all the PyTorch 1.12.1 updates I think we might need to update Torchvision to 0.13.1.  Seems to be the case with the CPU image at least.

abhi-mosaic:
@bandish-shah I updated to torchvision 0.13.1 and all the checks seem to pass

bandish-shah (approved):
LGTM thanks

PR #1411, opened 2022-08-13
Metrics refactor pt1

ishanashastri:
This PR implements part 1 of the Metrics Refactor, which entails removing metrics from the `Evaluator` class and storing raw deep-copied metrics in the `State` class. The `Evaluator` class now stores evaluation metric _names_ instead of metric instances, which are matched against the model-defined `model.val_metrics` to indicate which metrics will be computed at eval time. Additionally, instead of storing all computed metrics as part of `state.current_metrics`, both the raw non-computed training and validation metrics are now stored separately as part of `state.train_metrics` and `state.eval_metrics`. 

All tests updated and passing, regression tests will be conducted once PR 2 is ready.

Note: PR for part 2 will also touch a lot of the same code, so expect some refactors and clean up (mostly in the `Trainer` class) after this PR is merged in.

Closes [CO-679](https://mosaicml.atlassian.net/browse/CO-679)


composer/trainer/trainer.py
```diff
@@ -956,8 +959,10 @@ def __init__(
         if eval_dataloader is None:
             evaluators: List[Evaluator] = []
         else:
+            model_metric_names = [str(k) for k in model.metrics(train=False).keys()]  # type: ignore
             evaluators = [
-                ensure_evaluator(evaluator, model.metrics(train=False)) for evaluator in ensure_tuple(eval_dataloader)
+                ensure_evaluator(evaluator, default_metric_names=model_metric_names)
```
hanlint, line 964: I'm a bit confused by the logic here -- if a user has different metrics they want to apply to different evaluators, how does that work? Seems like here we just run every metric against every eval dataset?

composer/trainer/trainer.py
```diff
@@ -1349,10 +1354,12 @@ def fit(
 
         # Evaluators
         if eval_dataloader is not None:
+            assert isinstance(self._original_model, ComposerModel)
```
hanlint, line 1357:
```suggestion
```

This should not be necessary with the `state.model` check above

composer/trainer/trainer.py
```diff
@@ -1406,19 +1413,32 @@ def _ensure_metrics_device_and_dtype(self, metrics: MetricCollection):
...
         self.logger.data(
             log_level=log_level,
             data={f'metrics/{dataloader_label}/{name}': val for (name, val) in computed_metrics.items()},
         )
-        self.state.current_metrics[dataloader_label] = computed_metrics
+
+        # store raw metrics
+        for metric_name, metric in metrics.items():
+            if dataloader_label == 'train':
+                assert isinstance(metric, Metric)
```
hanlint, line 1435: nit: can do a single assert after the for loop line

composer/trainer/trainer.py
```diff
@@ -2078,6 +2098,24 @@ def eval(
 
             self.engine.run_event(Event.EVAL_START)
 
+            # extract model metrics based on provided names
+            # TODO (Ishana): refactor as part of CO-251
+            if not metric_names:
+                metrics = self._original_model.metrics(train=False)
+            else:
+                # filter metrics based on globs
+                metrics = []
```
hanlint, line 2107: For this PR, can we at least move this to its own helper function (e.g.):
```
metrics = self._filter_metrics(metrics=self._original_model.metrics(train=False), names=metric_names)
```

hanlint (approved):
Overall LGTM, a few questions and suggested code changes for style nits.

composer/core/state.py
```diff
@@ -120,8 +120,26 @@ class State(Serializable):
             ...     eval_dataloader=eval_dataloader,
             ... )
             >>> trainer.fit()
-            >>> trainer.state.current_metrics
-            {'train': {'Accuracy': tensor(...)}, 'eval': {'CrossEntropy': tensor(...), 'Accuracy': tensor(...)}}
+            >>> trainer.state.train_metrics
+            {'Accuracy': Accuracy()}
+
+        eval_metrics (Dict[str, Dict[str, Metric]]): The current raw evaluation metrics, organized
+            by dataloader label and then by metric name. If not using an :class:`.Evaluator`,
```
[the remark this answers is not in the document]
ishanashastri: > * Can we add something here about how the dataloader label is typically used for multiple eval datasets, and maybe cross-link to https://docs.mosaicml.com/en/v0.8.2/trainer/evaluation.html#multiple-datasets?

Good point, added the documentation!

> * The use of "raw" is a bit confusing to me, maybe remove that adjective?

👍🏽

> * [Q]: are the examples here automatically doc-tested, or do we need to add `.. doctest::` ?

They are automatically doc-tested.

composer/trainer/trainer.py
ishanashastri: The `ensure_evaluator()` function actually does two things: if the given `evaluator` is an `Evaluator` instance (meaning it has its own set of metrics), the `evaluator` is just returned. Otherwise, if the given `evaluator` is just a dataloader, it is converted to an `Evaluator` that evaluates against all the given `default_metric_names` (which by default are just set to be all the evaluation metrics defined in the model).

composer/trainer/trainer_hparams.py
```diff
@@ -792,7 +791,7 @@ def initialize_object(self, model: ComposerModel, dataloader_hparams: DataLoader
 
         # Metrics
 
-        # TODO(Ravi): Cleanup this code as part of the MetricsModule. This code was copied
+        # TODO(Ishana): Cleanup this code as part of the MetricsModule. This code was copied
```
bandish-shah, line 794: Is there a Jira issue to track this TODO?  If not can we file an issue and remove the TODO from code?

bandish-shah (approved):
LGTM, please see my nit about the TODO

composer/core/state.py
```diff
@@ -295,11 +316,21 @@ def __init__(
...
 
-        self.current_metrics: Dict[str, Dict[str, Any]] = {}
+        self.train_metrics: Dict[str, Metric] = {}
+        self.eval_metrics: Dict[str, Dict[str, Metric]] = {}
+
+    @property
+    def current_metrics(self):
+        warnings.warn(
+            DeprecationWarning(
+                'The ``current_metrics`` argument for a :class:`Trainer`. state is deprecated and will be removed in the future. Please use ``train_metrics`` and'
```
hanlint, line 331: DeprecationWarnings are not rendered with RST, so can just do:

The current_metrics property for Trainer.state is deprecated and will be removed in the future. Please use train_metrics or eval_metrics instead.

hanlint (commented):
One more quick change, and I think this is ready to merge!

PR #1412, opened 2022-08-15
Use state algos

mvpatel2000:
Use state algos, so engine changes to algos can adjust these variables as needed. Needed for agent

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -973,8 +973,8 @@ class Trainer:
         self.state.evaluators = evaluators
 
         # Some algorithms require specific settings
-        self._backwards_create_graph = any(map(lambda x: x.backwards_create_graph, ensure_tuple(algorithms)))
-        self._find_unused_parameters = any(map(lambda x: x.find_unused_parameters, ensure_tuple(algorithms)))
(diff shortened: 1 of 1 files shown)


mvpatel2000:
This is going to change following offline discussion

mvpatel2000:
Oops closed wrong PR

composer/trainer/trainer.py
```diff
@@ -973,8 +973,8 @@ def __init__(
         self.state.evaluators = evaluators
 
         # Some algorithms require specific settings
-        self._backwards_create_graph = any(map(lambda x: x.backwards_create_graph, ensure_tuple(algorithms)))
-        self._find_unused_parameters = any(map(lambda x: x.find_unused_parameters, ensure_tuple(algorithms)))
+        self._backwards_create_graph = any(map(lambda x: x.backwards_create_graph, ensure_tuple(self.state.algorithms)))
+        self._find_unused_parameters = any(map(lambda x: x.find_unused_parameters, ensure_tuple(self.state.algorithms)))
```
dskhudia, line 977: ensure_tuple is not needed. self.state__init__ takes care of it.

dskhudia (approved):

composer/trainer/trainer.py
mvpatel2000: Good catch!

PR #1413, opened 2022-08-16
Have docs use preferred import paths, not longest import paths

dblalock:
Addresses [CO-789](https://mosaicml.atlassian.net/browse/CO-789) and [CO-794](https://mosaicml.atlassian.net/browse/CO-794). 

However, we'll need two follow-up changes:
 - Since our notebooks use hardcoded URLs as docs links, we'll need another PR to un-break all these links.
 - Some logic in certain algorithm modules isn't included in `composer.algorithms` (e.g., blurpool `nn.Module`s). This needs to get added to the algorithms `__init__.py` or something.

Design choices:

- Follows PyTorch docs templates + API reference structure, except that it's still possible to expand/collapse modules in the sidebar. PyTorch has this disabled, and I could [disable it easily](https://github.com/executablebooks/sphinx-remove-toctrees), but I left it as is to minimize the amount of logic changed in this PR.
    - In particular, there's now a unique page for each class and function, rather than one big page for everything in a module. This single page is also consistent with basically the whole pydata ecosystem.
- We manually choose which modules to include in the API reference, as well as optional members to blacklist to avoid redundancy.
- The actual .rst file for a given module is autogenerated using a simplified version of our current logic for adding autosummary tables to module docstrings. I was writing them out manually, but then realized that I'm literally just pasting stuff from the `__init__.py` files in a way I could easily automate.
- Everything available under `composer` is only shown in `composer`. This mostly means that the `composer.core` page is pretty empty.

### Before:
image

### After:
image

### Before:
_What you get when you click the top-level Alibi link in the sidebar:_
image

_Actual Alibi algorithm, in the middle of composer.algorithms.alibi.alibi module docs:_
image

### After:
_Alibi algorithm docs, which you're taken to immediately from the compser.algorithms sidebar link:_
image


hanlint:
🤔 is there some code missing from this push? The generated docs don't have an API reference anymore, and in the PR diff I don't see the modules being manually entered, just a refactored `conf.py` file?

dblalock:
Fixed the API reference not showing up (and then not building when run locally, which caused the unit test failure).

hanlint:
this looks 🔥 , defer to @bandish-shah for a careful look though

hanlint:
The last line in the API docs is too long.. should we increase the sidebar, or rename the utils to something shorter?

Screen Shot 2022-08-18 at 1 59 01 PM

dblalock:
^ Fixed the sidebar by injecting a little custom CSS (seems to be the only way to do this with the Furo theme).

@bandish-shah This should be ready for review

bandish-shah (approved):
LGTM, thanks for doing this it's awesome and so much simpler.

PR #1414, opened 2022-08-16
Fix pre-commit checks failing on fresh checkout of dev

dblalock:
Fixes [CO-883](https://mosaicml.atlassian.net/browse/CO-883) using a `type: ignore` as suggested by @hanlint.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -168,7 +168,7 @@ extra_deps['coco'] = [
 
 extra_deps['nlp'] = [
     'transformers>=4.11,<5',
-    'datasets>=1.14,<2',
+    'datasets>=2.4.0,<3',
 ]
 
 extra_deps['mlperf'] = [
--- a/composer/datasets/glue_hparams.py
+++ b/composer/datasets/glue_hparams.py
@@ -113,7 +113,8 @@ class GLUEHparams(DatasetHparams, SyntheticHparamsMixin):
             tokenizer = transformers.AutoTokenizer.from_pretrained(self.tokenizer_name)  #type: ignore (thirdparty)
 
             log.info(f'Loading {self.task.upper()} on rank {dist.get_global_rank()}')
-            download_config = datasets.utils.DownloadConfig(max_retries=self.max_network_retries)
+            download_config = datasets.utils.DownloadConfig(  # type: ignore (thirdparty)
+                max_retries=self.max_network_retries)
             dataset = datasets.load_dataset('glue', self.task, split=self.split, download_config=download_config)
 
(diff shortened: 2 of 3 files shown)


mvpatel2000:
Hm... I'm not getting this on dev... any ideas why?
image

hanlint:
@mvpatel2000 I think you can replicate the issue if you try: `pip install --upgrade datasets` (to >=version 2)

mvpatel2000:
image
I'm getting several issues with upgrade... @dblalock  are you also seeing these?

mvpatel2000:
@dblalock could you actually just resolve all the lint errors with datasets 2 and then update meta.yaml / setup.py to upgrade to datasets 2?

mvpatel2000:
CC: @moinnadeem do we need any tests for upgrading datasets?

moinnadeem:
Let's hold off on tests for updating `datasets` (ie. I can't think of anything that'll go wrong by updating that)

We'll need regression tests for updating `tokenizers` and `transformers` because of potential throughput and accuracy degradations respectively. If updating `datasets` requires updating `transformers` or `tokenizers`, then we should hold off and regression test. Otherwise, LGTM!

mvpatel2000:
@dblalock I put a bit on your plate, so if you don't have the bandwidth for this, lmk and I'm happy to help out. I'd rather just do the full upgrade now though given its something we need to get to anyways

mvpatel2000 (approved):
LGTM! Thanks for handling the upgrade :)

dblalock:
@bandish-shah can you approve / review this? You appear to be the only remaining member of "composer-team-admins"

setup.py
```diff
@@ -168,7 +168,7 @@ def package_files(prefix: str, directory: str, extension: str):
 
 extra_deps['nlp'] = [
     'transformers>=4.11,<5',
-    'datasets>=1.14,<2',
+    'datasets>=2.4.0,<3',
```
hanlint, line 171: shoulds this be >1.14 < 3?

and if it is >=2.4, we shoudl remove the `type: ignore`, and update the syntax to the datasets2 syntax?

dblalock: I removed the type: ignore and switched the name to `datasets.DownloadConfig`, which fixes the pyright error. I'm not sure what you mean by datasets2 syntax. Haven't really used `datasets` much but, in case this is relevant, [load_dataset](https://huggingface.co/docs/datasets/package_reference/loading_methods#datasets.load_dataset) seems to still be happy with the same API.

dblalock: Gonna call this resolved for the sake of getting this PR across the finish line. Happy to change something if you want though

hanlint (approved):

PR #1417, opened 2022-08-17
eval_batch_size=auto

mvpatel2000:
- Adds eval_batch_size= auto, expanding grad_accum=auto
- removes monitored_barrier

Interestingly, eval_batch_split is often far smaller than grad_accum (with resnet batch size 8k, I was seeing 2 vs. 8), so making them two different variables was the right call.


mvpatel2000:
Oops meant to open draft PR -- not ready for review yet!

mvpatel2000:
I _think_ this works -- tested on resnet with large batch size, but I'd love to get some help testing this on more workloads.

dskhudia:
Is this ready for review?

mvpatel2000:
> Is this ready for review?

Yes! :)

composer/trainer/trainer.py
```diff
@@ -180,7 +180,91 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
-    return 'CUDA out of memory' in str(e)
+    if 'CUDA out of memory' in str(e):
+        return True
+    # With batch_norm, large batch sizes sometimes result in cuDNN instead of Cuda OOMs.
+    if 'cuDNN error: CUDNN_STATUS_NOT_SUPPORTED. This error may appear if you passed in a non-contiguous input.' in str(
+            e):
+        warnings.warn(
+            f'Encountered "cuDNN error: CUDNN_STATUS_NOT_SUPPORTED. This error may appear if you passed in a non-contiguous input." This can happen when the batch_size is too large for the GPU so auto grad_accum will rerun with a higher grad_accum value, but there may be a user error with non-contiguous inputs.'
+        )
+        return True
```
mvpatel2000, line 191: Is there a better way to handle this?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -310,21 +310,16 @@ def _activate_model(self, state: State, logger: Logger) -> None:
...
-            else:
+            # In-line to avoid circular dependency
+            from composer.trainer.trainer import _get_device, _handle_cuda_oom
+
+            device_arg = None
+            if device is not None:
+                device_type = device.type
+                if 'cuda' in device.type:
+                    device_type = 'gpu'
+                device_arg = _get_device(device_type)
```
mvpatel2000, line 321: is there a better way to convert torch.device to a composer device? Maybe we move device into state?

composer/trainer/trainer.py
hanlint: This error isn't specific to CUDA OOM, so I am concerned about catching non-OOM errors with this check. Did we encounter use cases with this, and how common is this?

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -310,21 +310,16 @@ def _activate_model(self, state: State, logger: Logger) -> None:
...
+            if not _handle_cuda_oom(state, should_handle_cuda_oom, device_arg, device_batch_size, is_train=True):
```
hanlint, line 322: The side effect is a bit obfuscated here (there's no training happening here, yet `is_train=True`), what about something like:
```
    if device='gpu' and any_rank_oom(cuda_oom):
        adjust_grad_accum(state, device_batch_size)
```

composer/trainer/trainer.py
```diff
@@ -180,7 +180,91 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
+    return False
+
+
+def _handle_cuda_oom(state: State, should_handle_cuda_oom: int, device: Optional[Device], device_batch_size: int,
```
hanlint, line 195: This function is probably trying to do too much, suggest splitting it up per above usage in SLW. CUDA OOMing is only for `gpu`, so we can assume that any calls to this method are already using GPU (therefore device is not needed).

composer/trainer/trainer.py
```diff
@@ -1691,6 +1766,40 @@ def _train_loop(self) -> None:
...
         self._run_evaluators(Event.FIT_END, log_level=LogLevel.FIT)
 
+    def _eval_train_metrics(self):
+        assert self._train_data_spec is not None, 'The train data spec should be set on __init__ or fit()'
+        assert self.train_metrics is not None, 'The train metrics should be set on __init__ or fit()'
+
+        self.state.model.eval()
+        with torch.no_grad():
+            # Cache the device batch, because `self.state.batch` gets overridden in microbatching loop
+            device_batch = self.state.batch
```
hanlint, line 1776: the `device_batch` isn't used below, maybe cache the `device_batch_size` instead?

composer/trainer/trainer.py
```diff
@@ -1691,6 +1766,40 @@ def _train_loop(self) -> None:
...
+
+            # Retry until we successfully complete evaluation
+            while True:
+                # Note: We use uint8 instead of bool as BOR is not supported on all torch.distributed backends
```
hanlint, line 1780:
```suggestion
```

composer/trainer/trainer.py
```diff
@@ -1691,6 +1766,40 @@ def _train_loop(self) -> None:
...
+                should_handle_cuda_oom = 0
```
hanlint, line 1781: perhaps better to describe what happened, not what-to-do, e.g. `found_cuda_oom`, or simply `cuda_oom`.

composer/trainer/trainer.py
```diff
@@ -180,7 +180,91 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
+    dist.all_reduce(should_handle_cuda_oom_tensor, reduce_operation='MAX')
+    if int(should_handle_cuda_oom_tensor.item()) == 1:
+        # If any rank hit CUDA OOM, update grad_accum and retry. Raise runtime error if training
+        # 1 sample at a time still resulted in CUDA out of memory.
+        if is_train:
+            _handle_train_cuda_oom(state, device_batch_size)
+        else:
+            _handle_eval_cuda_oom(state, device_batch_size)
+        # Empty cache if on GPU, which will help reduce fragmentation
+        if isinstance(device, DeviceGPU):
```
hanlint, line 224: no need to guard, `torch.cuda.empty_cache()` is a no-op on CPUs.

composer/trainer/trainer.py
```diff
@@ -1716,12 +1825,10 @@ def _train_batch(self, use_grad_scaling: bool):
         assert self._train_data_spec is not None, 'The train data spec should be set on __init__ or fit()'
 
         # Cache the device batch, because `self.state.batch` gets overridden in microbatching loop
-        # TODO: fix this name collision!
         device_batch = self.state.batch
 
         # Retry until we successfully complete training and return loss
         while True:
-            start_time = time.time()
```
hanlint, line 1724: did we accidentally rm the `batch_compute_time` ? Or was that intentional?

composer/trainer/trainer.py
```diff
@@ -2204,6 +2255,41 @@ def eval(
         if original_num_batches is not None:
             self.state.dataloader_len = original_num_batches
 
+    def _eval_batch(self, data_spec: DataSpec, metrics: MetricCollection):
```
hanlint, line 2258: prefer to keep this in the `eval` method if possible, so that all the events are shown linearly in the eval loop

composer/trainer/trainer_hparams.py
```diff
@@ -423,6 +448,11 @@ def initialize_object(self) -> Trainer:
 
         # ensure grad_accum is 'auto' or an integer
         grad_accum = _parse_grad_accum(self.grad_accum)
+        eval_batch_size = _parse_eval_batch_size(self.eval_batch_size)
+        # _parse_eval_batch_size ensures eval_batch_size is auto if set to str. We need to check if
+        # it's a str and not eval_batch_size == 'auto' or pyright complains it might be a str still.
+        if isinstance(eval_batch_size, str):
+            eval_batch_size = self.train_batch_size
```
hanlint, line 455: wouldn't this then just ignore `eval_batch_size=auto` completely? Or is that fact that auto was selected recorded somewhere else?

composer/trainer/trainer.py
```diff
@@ -876,6 +959,7 @@ def __init__(
                            model=model,
                            callbacks=callbacks,
                            grad_accum=grad_accum,
+                           eval_batch_split=eval_batch_split,
```
hanlint, line 962: You probably vetted this with research team, but wouldn't we want something like `auto_eval_batch_size: bool` instead to be more discoverable a feature?

hanlint (commented):
Overall good, a few suggestions about the abstraction / making the side effects clear.

composer/trainer/trainer.py
mvpatel2000: Can you explain why? I think it's a useful comment

composer/trainer/trainer.py
mvpatel2000: I'm encountering this consistently with resnet50 on a100 with in batchnorm with 8k batch size. I'm happy to skip though since its an edge case

composer/trainer/trainer.py
mvpatel2000: Intentional, it was previously used for monitored_barrier

composer/trainer/trainer.py
mvpatel2000: Research preferred having grad_accum=auto affect both eval and train

composer/trainer/trainer_hparams.py
mvpatel2000: if grad_accum=auto, we turn it on for both train/eval (per research's feedback). So, whether you set eval_batch_size explicitly or not, its enabled with grad_accum=auto. 

As a further feature, if you don't want to worry about eval batch size at all (most use cases), setting it to auto lets you ignore it completely.

composer/trainer/trainer.py
hanlint: ah ok sounds good, then does `eval_batch_split` even need to be an argument to the trainer?

composer/trainer/trainer.py
mvpatel2000: can you elaborate? this is the same decomposition as the train loop -- I prefer at least some decomposition over one massive function

composer/trainer/trainer.py
mvpatel2000: oops! forgot to use it

composer/trainer/trainer.py
mvpatel2000: I don't think so. The goal here is to add no trainer args. 

It doesn't really make sense for someone to intentionally set eval_batch_split and a high batch size (since theres no grad accum happening)

hanlint: oops, my bad I misread this code; thought it was the `__init__` signature and didn't expand the context.

dskhudia:
Do we clear the already computed gradients if we run out of memory while backward is partially executed? PyTorch will do += for such already computed gradients on retry.

composer/trainer/trainer.py
hanlint: OK I'm fine keeping it in, that seems like a reasonable case.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -306,27 +307,25 @@ def _activate_model(self, state: State, logger: Logger) -> None:
...
+            # In-line to avoid circular dependency
+            from composer.trainer.trainer import _handle_train_cuda_oom
+
+            # Auto grad accum only supported on GPU
+            if device and device.type == 'gpu':
+                devicegpu = DeviceGPU()
+                # Propagate across all ranks if any rank hit CUDA OOM
+                found_cuda_oom = devicegpu.tensor_to_device(torch.tensor([found_cuda_oom], dtype=torch.uint8)).item()
+                if found_cuda_oom == 1:
+                    _handle_train_cuda_oom(state, device_batch_size)
```
hanlint, line 323: personal preference, so feel free to ignore, but prefer not obfuscating the side effect and being more transparent on what is happening, e.g. `_adjust_grad_accum`. otherwise, the reader needs to check another function to actually understand what is happening.

composer/trainer/trainer.py
```diff
@@ -180,7 +180,62 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
+    if 'cuDNN error: CUDNN_STATUS_NOT_SUPPORTED. This error may appear if you passed in a non-contiguous input.' in str(
+            e):
+        warnings.warn(
+            f'Encountered "cuDNN error: CUDNN_STATUS_NOT_SUPPORTED. This error may appear if you passed in a non-contiguous input." This can happen when the batch_size is too large for the GPU so auto grad_accum will rerun with a higher grad_accum value, but there may be a user error with non-contiguous inputs.'
+        )
+        return True
+    return False
+
+
+def _handle_train_cuda_oom(state: State, device_batch_size):
```
hanlint, line 195: personal preference, if the docstrings say "adjust grad_accum", maybe the function name should just reflect that to be more readable.

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -306,27 +307,25 @@ def _activate_model(self, state: State, logger: Logger) -> None:
...
-            else:
-                self._activated = True
-                return
+            # In-line to avoid circular dependency
+            from composer.trainer.trainer import _handle_train_cuda_oom
+
+            # Auto grad accum only supported on GPU
+            if device and device.type == 'gpu':
+                devicegpu = DeviceGPU()
+                # Propagate across all ranks if any rank hit CUDA OOM
```
hanlint, line 320: i think the `dist.all_reduce` code is now missing

composer/trainer/trainer.py
hanlint: Generally favor terse comments, unless its a significant (and unexpected) bit of code. Suggest moving this to an actual check inside the whichever helper does the reduction, or even just convert the bool -> uint8.

If the comment has to be here, suggest something more terse:
```
found_cuda_oom = 0  # int since bool not supported on all torch.distributed backends
```

composer/trainer/trainer.py
hanlint: The train loop structure was done more for code readability/reuse, but never purely for decomposition:
* `_train_batch:` prevent excessive indentation (while, try/catch, etc.)
* `_train_microbatches`: multiple calls depending on closures/etc.
* `_train_microbatch`: prevent excessive indentation (using sync contexts)

As `_eval_batch` is not that long block of code anyway, prefer not to disrupt the readability of EVAL_* events in the code + add another layer in the stack trace.

hanlint (commented):
Close, i think we can close this out soon.

composer/trainer/trainer.py
```diff
@@ -180,7 +180,62 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
+    # at a time still resulted in CUDA out of memory.
+    if state.eval_batch_split == device_batch_size:
+        raise RuntimeError(('CUDA out of memory. The train loop failed with an internal microbatch of size 1.'
+                            'The GPU does not have enough memory to process even 1 sample.'))
+    else:
+        original_eval_batch_split = state.eval_batch_split
+        state.eval_batch_split = min(2 * state.eval_batch_split, device_batch_size)
+        warnings.warn(
+            RuntimeWarning('CUDA out of memory detected. Eval batch split '
+                           f'increased from {original_eval_batch_split} -> {state.eval_batch_split}, '
```
bandish-shah, line 235: [nit] Instead of "Eval batch split" in this warning message could we just call this "Eval microbatch"?  We refer to the eval "micro-batchsize" below so this would be more consistent messaging to users.  Alternatively we could just refer to it as "batch splits", just looking for consistency.

composer/core/state.py
```diff
@@ -82,6 +82,8 @@ class State(Serializable):
         run_name (str): The name for this training run.
         grad_accum (int, optional): The number of gradient accumulation steps to use. With this argument, micro batch
             size for each device becomes ``microbatch_size = train_batch_size / (num_devices * grad_accum)``.
+        eval_batch_split (int, optional): The mirror of grad_accum for eval. With this argument, micro batch
+            size for each device becomes ``microbatch_size = eval_batch_size / (num_devices * grad_accum)``.
```
bandish-shah, line 86: [nit] Could be mistaken but should the denominator have `grad_accum`?  I thought these were independent so `eval_microbatch_size = eval_batch_size // eval_batch_split`?

composer/trainer/trainer.py
```diff
@@ -180,7 +180,65 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
...
+        original_grad_accum = state.grad_accum
+        state.grad_accum = min(2 * state.grad_accum, device_batch_size)
+        warnings.warn(
+            RuntimeWarning('CUDA out of memory detected. Gradient Accumulation, the number of train microbatches, '
+                           f'increased from {original_grad_accum} -> {state.grad_accum}, '
+                           'and the batch will be retrained with a '
+                           f'micro-batchsize of {device_batch_size // state.grad_accum}'))
+    # Clear gradients in case failure happened during backwards pass
+    for optimizer in state.optimizers:
+        optimizer.zero_grad()
```
mvpatel2000, line 217: @dskhudia  this should resolve the += gradient issue, correct? Not as familiar with internals so I'd like to verify

dskhudia: This should do it.

hanlint (approved):
LGTM, thanks! Please merge when you are ready.

dskhudia (approved):

Issue #467, opened 2022-02-11
avoid double forward pass for train metrics


A-Jacobson:
Our current training loop requires an extra forward pass to compute training metrics which creates a considerable slowdown. Since training metrics are pretty useful, this slowdown could be a dealbreaker for some. It also seems completely avoidable, I see two options:

current loop..ish
```python

def fit():
    while timer < max_duration:
        model.train()  # set model to train mode
        for batch in train_dataloader:

            if self.compute_training_metrics:
                model.eval()
                with torch.no_grad():
                    outputs, targets = model.validate(batch)
                    train_metrics.update(outputs, targets)
                model.train()

            optimizer.zero_grad()
            outputs = model.forward(batch)
            loss = model.loss(outputs, batch)
            loss.backward()
            scheduler.step()

        ## Eval
        model.eval()
        eval_metrics = model.metrics(train=False)  # get eval metrics
        with torch.no_grad():
            for batch in eval_dataloader:
                outputs, targets = model.validate(batch)
                eval_metrics.update(outputs, targets)
            eval_metrics.compute()  # compute and sync metrics across devices
```
###  Option 1 :
 re-use outputs from forward and access targets from `state.batch`. 

upside:
-  solves the problem
-  no changes to the model class

downside:
- we need to know how to unpack the batch =(

```python
  for batch in train_dataloader:
      optimizer.zero_grad()
      outputs = model.forward(batch)
      loss = model.loss(outputs, batch)
      if self.compute_training_metrics:
          train_metrics.update(outputs.detach(), state.batch[1]) # tuple batch
      loss.backward()
      scheduler.step()
 ```
 ### Option 2:
 return `outputs` and `targets` from `model.forward`
upside:
-  solves the problem
- user specifies how to unpack the batch
- potential to unify `forward` and `validate` by having `validate` call forward by default and allowing users to override if their validation loop requires special logic such as beamsearch decoding for nlp. 

downside:
- targets return from `forward` is sort of strange, but we cant use `forward` for prediction for our current setup anyway as it expects a full batch (with targets). maybe we rename this method as there's already an expectation (though not convention) in pytorch that `forward` takes model input and returns output?
- `validate` overrides would either be ignored in the training loop, or require two forward steps yet again.

```python
  for batch in train_dataloader:
      optimizer.zero_grad()
      outputs, targets = model.forward(batch)
      loss = model.loss(outputs, batch)
      if self.compute_training_metrics:
          train_metrics.update(outputs.detach(), targets)
[...]

PR #1419, opened 2022-08-17
Metrics Refactor Part 2

ishanashastri:
This PR implements the second half of the metrics refactor (#1411) that updates the training loop and all the Composer models. 

The training loop now uses the previously added `state.train_metrics` and `state.eval_metrics` from #1411 to perform all training and evaluation on the correct set of metrics.

The main changes with respect to models is that the models now have split up `validate()` into `eval_forward()` and `update_metrics()` methods, which run an evaluation forward pass and update the metrics with the outputs of the evaluation forward pass respectively. This is mainly to get rid of the double forward pass (#467). 

TODO before merging: full regression tests

Closes [CO-15](https://mosaicml.atlassian.net/browse/CO-15) and [CO-251](https://mosaicml.atlassian.net/browse/CO-251)


hanlint (commented):
A few initial comments -- I also think this needs to be re-merged with `dev`now that the eval oom catching is in, likely some conflicts there.

composer/models/huggingface.py
```diff
@@ -88,23 +89,41 @@ def forward(self, batch):
...
+
+        if not metrics:
+            return {}
+
+        model_metrics = {}
+        for name, metric in metrics.items():
+            assert isinstance(metric, Metric)
+            model_metrics[name] = metric
+
+        return model_metrics
```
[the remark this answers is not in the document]
ishanashastri: Good catch!

composer/models/tasks/classification.py
```diff
@@ -81,15 +82,27 @@ def loss(self, outputs: Tensor, batch: Tuple[Any, Tensor], *args, **kwargs) -> T
...
+
+        if isinstance(metrics, Metric):
+            metrics_dict = {metrics._get_name(): metrics}
+        else:
+            metrics_dict = {}
+            for name, metric in metrics.items():
+                assert isinstance(metric, Metric)
+                metrics_dict[name] = metric
+
+        return metrics_dict
```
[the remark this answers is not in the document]
ishanashastri: How do we feel about modifying the `ComposerClassifier` signature to only accept a `MetricCollection` and not a `Union[Metric, MetricCollection]`. Taking it one step further, what about forcing subclasses of `ComposerClassifier` to accept a pre-constructed dictionary? This would resolve all the typing conversions in this function.

composer/core/evaluator.py
```diff
@@ -118,13 +120,22 @@ def __init__(
...
+        metrics: Optional[Union[Metric, MetricCollection]] = None,
         subset_num_batches: Optional[int] = None,
         eval_interval: Optional[Union[int, str, Time, Callable[[State, Event], bool]]] = None,
     ):
         self.label = label
         self.dataloader = ensure_data_spec(dataloader)
 
-        self.metric_names = metric_names if metric_names else []
+        if metric_names or metrics:
+            assert (metric_names is None and metrics) or (metrics is None and metric_names)
```
hanlint, line 131: I think this should properly throw an error instead of an internaly-only assert

composer/core/evaluator.py
```diff
@@ -118,13 +120,22 @@ def __init__(
...
+            if metric_names:
+                self.metric_names = metric_names
+            elif metrics:
+                if isinstance(metrics, Metric):
+                    self.metric_names = [metrics._get_name()]
+                else:
+                    self.metric_names = [str(k) for k, _ in metrics.items()]
```
hanlint, line 138: if not provided, shouldn't `self.metric_names` be initialized to `[]`?

composer/core/evaluator.py
```diff
@@ -118,13 +120,22 @@ def __init__(
...
         self.label = label
         self.dataloader = ensure_data_spec(dataloader)
 
-        self.metric_names = metric_names if metric_names else []
+        if metric_names or metrics:
+            assert (metric_names is None and metrics) or (metrics is None and metric_names)
+            if metric_names:
+                self.metric_names = metric_names
+            elif metrics:
+                if isinstance(metrics, Metric):
```
hanlint, line 135: Let's throw a DeprecationWarning for this one, tell folks to use metric_names instead in the future

composer/models/base.py
```diff
@@ -195,7 +196,10 @@ def __init__(self):
             def metrics(self, train: bool = False):
                 return self.train_acc if train else MetricCollection([self.val_acc, self.val_loss])
         """
-        raise NotImplementedError('Implement metrics in your ComposerModel to run validation.')
+        warnings.warn(
```
hanlint, line 199: Can this be an explicit `DeprecationWarning`?

composer/models/huggingface.py
```diff
@@ -70,12 +70,11 @@ def __init__(self,
...
 
         self.train_metrics = None
-        self.valid_metrics = None
+        self.val_metrics = None
 
         if metrics:
-            metric_collection = MetricCollection(metrics)
-            self.train_metrics = metric_collection.clone(prefix='train_')
-            self.valid_metrics = metric_collection.clone(prefix='val_')
+            self.train_metrics = {metric._get_name(): metric for metric in metrics}
```
hanlint, line 76: We shouldn't rely on a private method from torchmetrics here.. can this just be the class name?

composer/models/huggingface.py
```diff
@@ -88,23 +87,33 @@ def forward(self, batch):
...
 
-    def validate(self, batch):
+    def eval_forward(self, batch, outputs: Optional[Any] = None):
         if self.use_logits:
-            labels = batch.pop('labels')
-            output = self.forward(batch)
+            if outputs:
+                output = outputs
+            else:
+                output = self.forward(batch)
```
hanlint, line 95:
```suggestion
            output = outputs if outputs else self.forward(batch)
```

hanlint: This can probalby be i the outer loop actually, something like:
```
def eval_forward(..):
    if self.use_logits:
        output = output['logits']..
        ...
         return output
     return output

File: composer/models/huggingface.py [hunk 9]
```diff
@@ -88,23 +87,33 @@ def forward(self, batch):
...
         else:
-            output = self.forward(batch)
-            return output, None
+            return outputs if outputs else self.forward(batch)
 
-    def metrics(self, train: bool = False):
-        return self.train_metrics if train else self.valid_metrics
+    def get_metrics(self, is_train: bool = False) -> Dict[str, Metric]:
+        if is_train:
+            metrics = deepcopy(self.train_metrics)
```
hanlint, line 108: why do we need to deepcopy here?

composer/models/tasks/classification.py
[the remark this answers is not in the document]
hanlint: cc: @Landanjs and @coryMosaicML for some input on how to handle this under the new Metrics API

composer/models/tasks/classification.py
```diff
@@ -81,15 +82,35 @@ def loss(self, outputs: Tensor, batch: Tuple[Any, Tensor], *args, **kwargs) -> T
...
         inputs, _ = batch
         outputs = self.module(inputs)
         return outputs
 
-    def validate(self, batch: Tuple[Tensor, Tensor]) -> Tuple[Tensor, Tensor]:
-        _, targets = batch
-        outputs = self.forward(batch)
-        return outputs, targets
+    def eval_forward(self, batch: Any, outputs: Optional[Any] = None) -> Any:
+        if outputs is not None:
```
hanlint, line 111: one-liner?
```
return outputs if outputs is not None else self.forward(batch)
```

composer/trainer/trainer.py
```diff
@@ -996,9 +975,43 @@ def __init__(
...
+        # Set state.train_metrics and state.eval_metrics here to allow callbacks / algs to potentially
+        # change the model, which could change what metrics are computed
+        self.state.train_metrics = self.state.model.get_metrics(is_train=True)
+        self.state.eval_metrics = {}
+        if eval_dataloader is None:
+            evaluators: List[Evaluator] = []
+        else:
+            eval_metrics = self.state.model.get_metrics(is_train=False)
+            model_metric_names = [str(k) for k in eval_metrics.keys()]
+            evaluators = []
```
hanlint, line 994: i think more readable here to split the code here by action:
```
eval_metrics = ..
names = [str(k) for k in eval_metrics.keys()]

evaluators = [
    ensure_evaluator(evaluator, default_metric_names=model_metric_names)
    for evaluator in ensure_tuple(eval_dataloader)
]


self.state.eval_metrics = {
    evaluator.label: _filter_metrics(...)
    for evaluator in evaluators
]
```

composer/trainer/trainer.py
```diff
@@ -1435,17 +1415,18 @@ def fit(
...
-                # Need to use the `original_model` rather than `state.model`, as `state.model`
-                # could be DDP / DeepSpeed wrapped.
-                ensure_evaluator(evaluator, default_metric_names=metric_names)
-                for evaluator in ensure_tuple(eval_dataloader)
-            ]
+            # Need to use the `original_model` rather than `state.model`, as `state.model`
+            # could be DDP / DeepSpeed wrapped.
+            eval_metrics = self._original_model.get_metrics(is_train=False)
+            metric_names = [str(k) for k in eval_metrics.keys()]
+            evaluators = []
```
hanlint, line 1422: same comment here as above regarding the code separation

composer/models/base.py
```diff
@@ -232,4 +236,60 @@ def validate(self, batch): # batch is the output of the dataloader
 
             metrics.compute() # compute final metrics
         """
-        raise NotImplementedError('Implement validate in your ComposerModel to run validation.')
+        return None, None
```
hanlint, line 239: I think just delete this method?

composer/trainer/trainer.py
```diff
@@ -1759,9 +1743,22 @@ def _eval_train_metrics(self):
                         # TODO: Detect if self.run_event(Event.AFTER_DATALOADER) changes the training
                         # data and if so print a warning that metrics may return unexpected results
                         with get_precision_context(self.state.precision):
-                            outputs, targets = self._original_model.validate(eval_microbatch)
-                            # Run in same precision context to avoid NaNs
-                            self.train_metrics.update(outputs, targets)
+                            if any(self._original_model.validate(eval_microbatch)):  # backwards compatibility check
```
hanlint, line 1746: This would cause calling `validate` twice. How about removing `validate` from `CompsoerModel` entirely, and then doing a:
```
if hasattr(self._original_model, 'validate'):
    warnings.warn(DeprecationWarning(...)
    eval_outputs, target = self...
    for _, metric in..
        metric.update(eval_outputs, target)
```

Note: I think the metric updating step was missing in this code block (I added back in)

composer/trainer/trainer.py
```diff
@@ -2218,17 +2209,32 @@ def eval(
                         for eval_microbatch in data_spec.split_batch(self.state.batch, self.state.eval_batch_split):
                             self.engine.run_event(Event.EVAL_BEFORE_FORWARD)
                             with get_precision_context(self.state.precision):
-                                self.state.outputs, targets = self._original_model.validate(eval_microbatch)
+                                if any(self._original_model.validate(
```
hanlint, line 2212: same comment as above re: deprecation

composer/trainer/trainer.py
```diff
@@ -2218,17 +2209,32 @@ def eval(
...
                                 if isinstance(self._device, DeviceMPS):
                                     # torchmetrics math has numerical errors on M1 devices
                                     # running the compute on CPU instead
-                                    metrics.update(self.state.outputs.cpu(), targets.cpu())
+                                    outputs = self.state.outputs.cpu()
                                 else:
-                                    metrics.update(self.state.outputs, targets)
+                                    outputs = self.state.outputs
+                                for _, metric in metrics.items():
+                                    self._original_model.update_metric(
```
hanlint, line 2232: For backwards compatibility though, this method wouldn't exist right?

composer/models/huggingface.py
ishanashastri: if there are multiple evaluators, we want to make sure that each evaluator only updates its own set of metrics which are set using this method.

hanlint: ah, is this something we can handle internally? Users implementing their models may not realize they need to return a deepcopy.

ishanashastri: yep, good point -- will move to inside the trainer

composer/trainer/trainer_hparams.py
```diff
@@ -822,32 +817,29 @@ def initialize_object(self, model: ComposerModel, dataloader_hparams: DataLoader
...
+            metrics = {}
             for metric_name in self.metric_names:
-                try:
-                    metric = model_metrics[metric_name]
-                except KeyError as e:
+                for k in model_metrics.keys():
+                    matches = re.match(f'.*{metric_name}.*', k, re.IGNORECASE)
+                    if matches:
+                        metrics[k] = model_metrics[k]
+                else:
```
hanlint, line 832: is this `else` indented improperly?

docs/source/composer_model.rst
```diff
@@ -100,25 +100,27 @@ To compute metrics during training, implement the following methods:
 
 .. code:: python
 
-   def validate (self, batch) -> outputs, targets:
+   def eval_forward (self, batch, outputs) -> outputs:
        ...
 
-   def metrics(self, train=False) -> Metrics:
+   def get_metrics(self, is_train=False) -> Dict[str, Metric]:
```
hanlint, line 106: missing description of `update_metrics`?

hanlint: This was marked as resolved, but I don't see additional text here? Is there a commit missing?

composer/trainer/trainer.py
```diff
@@ -1759,9 +1750,23 @@ def _eval_train_metrics(self):
                         # TODO: Detect if self.run_event(Event.AFTER_DATALOADER) changes the training
                         # data and if so print a warning that metrics may return unexpected results
                         with get_precision_context(self.state.precision):
-                            outputs, targets = self._original_model.validate(eval_microbatch)
-                            # Run in same precision context to avoid NaNs
-                            self.train_metrics.update(outputs, targets)
+                            if hasattr(self._original_model, 'validate'):  # backwards compatibility check
+                                warnings.warn(
+                                    'Using ``validate()`` is no longer supported and will be removed in a future version. Please use ``eval_forward()`` instead.'
```
hanlint, line 1755:
```suggestion
                                    'Using validate() is no longer supported and will be removed in a future version. Please use eval_forward() instead.'
```

hanlint (approved):
THis LGTM, thank for putting this together! Pending your end2end testing, I think we are good to merge this.

PR #1421, opened 2022-08-17
Add default ignore index

moinnadeem:
Adds a default ignore index value. This has been shown to be difficult to use by both external users and internal users.

Why -100? That is the default value used for BERT, as well as the [value PyTorch uses for `ignore_index` internally](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html).


codestar12 (approved):
Looks good to me. Very helpful to have this default.

alextrott16 (approved):
LGTM

hanlint (approved):

PR #1427, opened 2022-08-19
Simplify helper for conflicting files

hanlint:
* Simplifies the helper function
* Removes checks for `_in_epoch` conflicts, since its not possible for files to have the same `batch` counter, but have different `epoch_in_batch` counters.


composer/utils/file_helpers.py
```diff
@@ -99,33 +99,17 @@ def ensure_folder_has_no_conflicting_files(folder_name: Union[str, pathlib.Path]
...
-                        match.group('batch_in_epoch'))) > timestamp.batch_in_epoch:
-                    valid_match = False
-            if not valid_match:
-                raise FileExistsError(
-                    f'{os.path.join(folder_name, file)} exists and conflicts in namespace with a future checkpoint of the current run.'
-                )
+            match = match.groupdict()
+            for unit, value in match.items():
+                if unit.endswith('_in_epoch'):
+                    continue  # _in_epoch counters already accounted for by the units
```
mvpatel2000, line 107: What about something like `{epoch}_{batch_in_epoch}`? I think `_in_epoch` are only accounted for if the corresponding unit is in the pattern.

mvpatel2000 (commented):
This is a lot cleaner! I'm not sure if we can skip the `_in_epoch` though -- see comment.

composer/utils/file_helpers.py
hanlint: Ah right, sure I can add the logic back in

composer/utils/file_helpers.py
```diff
@@ -99,33 +99,20 @@ def ensure_folder_has_no_conflicting_files(folder_name: Union[str, pathlib.Path]
...
-                raise FileExistsError(
-                    f'{os.path.join(folder_name, file)} exists and conflicts in namespace with a future checkpoint of the current run.'
-                )
+            match = match.groupdict()
+            for unit, value in match.items():
+                if unit.endswith('_in_epoch'):
+                    if 'epoch' not in match:
+                        raise ValueError(f'{filename} has {{unit}} but not {{epoch}}. Add {{epoch}} for uniqueness.')
+                    if int(match['epoch']) != timestamp.epoch:
+                        continue  # only check _in_epoch if both files have same epoch count
```
mvpatel2000, line 110: Why is this necessary?

mvpatel2000 (approved):
Approving to unblock

composer/utils/file_helpers.py
hanlint: If the epochs don't match, `continue` will skip the check of the `_in_epoch` property in the following lines. Let me know if the flow is a bit unclear.

PR #1433, opened 2022-08-19
Dynamic quantization + minor improvements in inference APIs

dskhudia:
Adds 

1. Helper wrapper for dynamic quantization 
2. If scripting fails, we were printing an exception stack trace but still continuing with tracing. Fix this. 
3. Add a test for dynamic quantization. 

closes https://mosaicml.atlassian.net/browse/CO-798

Test:
python -m pytest -k "test_dynamic_quantize" -s

### DIFF
--- a/composer/utils/inference.py
+++ b/composer/utils/inference.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import contextlib
 import copy
+import functools
 import logging
 import os
 import tempfile
@@ -29,9 +30,20 @@ if TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
-__all__ = ['export_for_inference', 'ExportFormat', 'export_with_logger']
+__all__ = ['export_for_inference', 'ExportFormat', 'export_with_logger', 'quantize_dynamic']
 
 Transform = Callable[[nn.Module], nn.Module]
+
+# This is the most common way to use dynamic quantization.
+#  Example:
+#    from composer.utils import quantize_dynamic
+#    export_for_inference(
+#        ...
+#        transforms = [quantize_dynamic],
+#        ...
+#    )
+#  A user can always redefine it with extra options. This also serves as an example of what to pass to transforms.
+quantize_dynamic = functools.partial(torch.quantization.quantize_dynamic, qconfig_spec={torch.nn.Linear})
 
 
 class ExportFormat(StringEnum):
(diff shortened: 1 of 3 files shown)


composer/utils/inference.py
```diff
@@ -29,10 +30,21 @@
...
+# This is the most common way to use dynamic quantization.
+#  Example:
+#    from composer.utils import quantize_dynamic
+#    export_for_inference(
+#        ...
+#        transforms = [quantize_dynamic],
+#        ...
+#    )
+#  A user can always redefine it with extra options. This also serves as an example of what to pass to transforms.
+quantize_dynamic = functools.partial(torch.quantization.quantize_dynamic, qconfig_spec={torch.nn.Linear})
```
nik-mosaic, line 46: Is there a reason we only apply this to linear layers?

tests/utils/test_inference.py
```diff
@@ -275,3 +275,41 @@ def test_export_with_other_logger(model_cls, sample_input):
...
+
+@pytest.mark.parametrize(
+    'model_cls',
+    [
+        (LinModel),
+    ],
+)
+def test_dynamic_quantize(model_cls):
+    model = model_cls()
+    model.eval()
```
lupesko, line 301: Is this needed? doesn't `export_for_inference()` implementation call `.eval()` after deep copying the model?

composer/utils/inference.py
```diff
@@ -148,16 +160,15 @@ def export_for_inference(
...
                 else:
-                    raise RuntimeError(
-                        'Scripting with torch.jit.script failed and sample inputs are not provided for tracing with torch.jit.trace'
-                    ) from e
+                    log.warning(
+                        'Scripting with torch.jit.script failed and sample inputs are not provided for tracing '
+                        'with torch.jit.trace',
+                        exc_info=True)
 
             if export_model is not None:
```
lupesko, line 173: Should we raise an exception in case `export_model` is None, to avoid silent failure?

composer/utils/inference.py
dskhudia: Dynamic quantization only works for linear layers and LSTMs. For us, only linears are important.

tests/utils/test_inference.py
dskhudia: It's not needed but also doesn't hurt.

dskhudia: Will remove

composer/utils/inference.py
nik-mosaic: Got it; thanks.

composer/utils/inference.py
dskhudia: Done. However, jit or script would have already raised an error.

lupesko:
Looks good! Thanks for addressing the comments.

nik-mosaic (approved):
Approved!

PR #1434, opened 2022-08-20
ResNet50 on ImageNet training script example

Landanjs:
## Motivation

A few goals for this script:
1. Replace several of the yamls we use for benchmarking purposes.
2. Example for users to quickly get a sense of how to use a variety of features in composer (general training, saving/loading checkpoints, create recipes, etc.)
3. Hackable example for users to quickly swap in their own model, dataset, or other training examples

 For 1, yamls this should cover:
- `resnet18.yaml`
- `resnet34.yaml`
- `resnet50.yaml`
- `resnet101.yaml`
- `resnet152.yaml`
- `resnet50_mild.yaml`
- `resnet50_medium.yaml`
- `resnet50_spicy.yaml`

Additionally, there are a few features that are not in the yamls:
- SpeedMonitor 
- LRMonitor
- Save checkpoint
- (Only) Resume training from a checkpoint

## Results

Need to run...

## Questions

1. Thoughts on having example commands as a comment at the top of the document? I could add these to a README, but I thought it would be nice to have the script be completely standalone
2. What do people think about how the recipes are instantiated? Is it okay that there are no argparse for the individuals methods in the recipe?
3. Should there be warnings about the crop resolution, loss function, and duration if a recipe is set, but these args are set to their default value?
4. Is it necessary to implement FFCV or Dali now to alleviate dataloader bottleneck? Or can this wait until a separate PR?


examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+    composer train_resnet_imagenet1k.py /path/to/imagenet --recipe_name mild --train_crop_size 176 --val_crop_size 224
+    --max_duration 36ep  --loss_name binary_cross_entropy
+
+Medium ResNet recipe highest accuracy with similar training time as baseline:
+    composer train_resnet_imagenet1k.py /path/to/imagenet --recipe_name medium --train_crop_size 176 --val_crop_size 224
+    --max_duration 135ep  --loss_name binary_cross_entropy
+
+Spicy ResNet recipe for our most accurate ResNet over a long training schedule:
+    composer train_resnet_imagenet1k.py /path/to/imagenet --recipe_name spicy --train_crop_size 176 --val_crop_size 224
+    --max_duration 270ep  --loss_name binary_cross_entropy
```
dblalock, line 34: Can these have a `\` at the end of the first line so that you can copy-paste these into a terminal?

Same for the wandb command above.

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+parser.add_argument('--grad_accum', help='Gradient accumulation either an int or "auto"', default='auto')
+parser.add_argument('--max_duration',
+                    help='Duration to train specified in terms of Time',
+                    type=Time.from_timestring,
+                    default='90ep')
+parser.add_argument('--eval_interval',
+                    help='How frequently to run the evaluation datasets',
+                    type=Time.from_timestring,
+                    default='1ep')
+parser.add_argument('--device', help='Device to run training on', choices=['gpu', 'cpu', 'tpu', 'mps'], default='gpu')
```
Landanjs, line 129: Err, maybe should swap this out for a check with `torch.cuda.is_available()`

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This script demonstrates how to train torchvision ResNet architectures on ImageNet using Composer's Trainer. The
+default settings replicate our baseline results. Below illustrates how some arguments can be set to replicate
+our fastest ResNet recipe, our highest accuracy ResNet recipe, and several other settings. This script contains several
+additional features such as saving checkpoints, resume training from a saved checkpoint, speed monitor to track
+training time and throughput, and a learning rate monitor to track learning rate.
```
dblalock, line 9:
```suggestion
"""
This script demonstrates how to train torchvision ResNet architectures on ImageNet using Composer's Trainer. The
default settings replicate our baseline results. Below illustrates how some arguments can be set to replicate
our fastest ResNet recipe, our highest accuracy ResNet recipe, and several other settings. This script contains several
additional features such as saving checkpoints, resuming training from a saved checkpoint, logging training speed over time, and a logging learning rate over time.
"""
```

Ignore the closing """ at the end; I just put that there so the diff figures out it's a docstring and becomes legible.

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+# LR scheduler arguments
+parser.add_argument('--t_warmup',
+                    help='Duration of learning rate warmup specified by a Time string',
+                    type=Time.from_timestring,
+                    default='8ep')
+parser.add_argument('--t_max',
+                    help='Duration to cosine decay the learning rate',
+                    type=Time.from_timestring,
+                    default='1dur')
+parser.add_argument('--alpha_f', help='Learning rate multiplier to decay to', type=Time.from_timestring, default=0)
```
dblalock, line 93: Can this arg just be the minimum learning rate to decay to? I know this framing matches the scheduler param name, but I'd rather make the top-level interface cleaner and just have a line that converts between them in the code.

Also feels ripe for a usability JIRA issue...

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+
+# Save checkpoint arguments
+parser.add_argument('--save_checkpoint_dir',
+                    help='Directory to save checkpoints',
```
dblalock, line 97:
```suggestion
                    help='Directory in which to save model checkpoints',
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+# Logger parameters: progress bar logging is used by default
+# Only has Weights and Biases option to reduce the number of arguments. Other loggers can be substituted in the script
+parser.add_argument('--wandb_logger', help='Whether or not to log results to Weights and Biases', action='store_true')
+parser.add_argument('--wandb_entity', help='WandB entity name', type=str)
+parser.add_argument('--wandb_project', help='WandB project name', type=str)
+parser.add_argument('--wandb_run_name', help='WandB run name', type=str)
+
+# Trainer arguments
+parser.add_argument('--seed', help='Random seed', type=int, default=17)
+parser.add_argument('--grad_accum', help='Gradient accumulation either an int or "auto"', default='auto')
```
dblalock, line 120: Hard to fit in one line so maybe not worth it, but would be great to specify what the int options mean. Like, does grad accum = 0 turn it off (no extra batches get accumulated), or does grad accum = 1 turn it off (one batch per update)?

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+parser.add_argument('--max_duration',
+                    help='Duration to train specified in terms of Time',
```
dblalock, line 122: Can we have a consistent phrasing for the args that take a time? E.g., `'specified in epochs (e.g., "5ep") or full training duration (e.g., "1dur")'`.

Though admittedly the `dur` version makes no sense for this particular arg.

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+# Divide batch sizes by number of devices if running multi-gpu training
+if dist.get_world_size():
+    args.train_batch_size = args.train_batch_size // dist.get_world_size()
+    args.eval_batch_size = args.eval_batch_size // dist.get_world_size()
+
+IMAGENET_CHANNEL_MEAN = (0.485, 0.456, 0.406)
+IMAGENET_CHANNEL_STD = (0.229, 0.224, 0.225)
+
+# Train dataset
+logging.info('Build train dataloader')
```
dblalock, line 147:
```suggestion
logging.info('Building train dataloader')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+    collate_fn=pil_image_collate,
+    persistent_workers=True,  # Reduce overhead of creating new workers at the expense of using slightly more RAM
+)
+# DataSpec allows for on-gpu transformations, marginally relieving dataloader bottleneck
+train_dataspec = DataSpec(dataloader=train_dataloader,
+                          device_transforms=NormalizationFn(mean=IMAGENET_CHANNEL_MEAN, std=IMAGENET_CHANNEL_STD))
+logging.info('Built train dataloader')
+
+# Validation dataset
+logging.info('Build evaluation dataloader')
```
dblalock, line 171:
```suggestion
logging.info('Building evaluation dataloader')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+    sampler=eval_sampler,
+    collate_fn=pil_image_collate,
+    persistent_workers=True,  # Reduce overhead of creating new workers at the expense of using slightly more RAM
+)
+eval_dataspec = DataSpec(dataloader=eval_dataloader,
+                         device_transforms=NormalizationFn(mean=IMAGENET_CHANNEL_MEAN, std=IMAGENET_CHANNEL_STD))
+logging.info('Built evaluation dataloader')
+
+# Instantiate torchvision ResNet model
+logging.info('Build Composer model')
```
dblalock, line 194:
```suggestion
logging.info('Building Composer model')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+def weight_init(w: torch.nn.Module):
+    if isinstance(w, torch.nn.Linear) or isinstance(w, torch.nn.Conv2d):
+        torch.nn.init.kaiming_normal_(w.weight)
+    if isinstance(w, torch.nn.BatchNorm2d):
+        w.weight.data = torch.rand(w.weight.data.shape)
+        w.bias.data = torch.zeros_like(w.bias.data)
+    # When using binary cross entropy, set the classification layer bias to -log(num_classes)
+    # to ensure the initial probabilities are approximately 1 / num_classes
+    if args.loss_name == 'binary_cross_entropy' and isinstance(w, torch.nn.Linear):
+        w.bias.data = torch.ones(w.bias.shape) * -torch.log(torch.tensor(w.bias.shape[0]))
```
dblalock, line 209: Are we using binary xent loss / is this needed to repro our numbers? Just looking for ways to simplify the example

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+model.apply(weight_init)
+
+# Performance metrics to log outside of training loss
```
dblalock, line 214:
```suggestion
# Performance metrics to log other than training loss
```
"outside of" kinda sounds like it's talking about scoping / where the logic runs

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+    loss_fn = binary_cross_entropy_with_logits
+else:
+    loss_fn = soft_cross_entropy
+
+# Wrapper function to convert a classification PyTorch model into a Composer model
+composer_model = ComposerClassifier(model, train_metrics=train_metrics, val_metrics=val_metrics, loss_fn=loss_fn)
+logging.info('Built Composer model')
+
+# Optimizer
+logging.info('Build optimizer and learning rate scheduler')
```
dblalock, line 230:
```suggestion
logging.info('Building optimizer and learning rate scheduler')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+                          lr=args.learning_rate,
+                          momentum=args.momentum,
+                          weight_decay=args.weight_decay)
+
+# Learning rate scheduler: LR warmup for 8 epochs, then cosine decay for the rest of training
+lr_scheduler = CosineAnnealingWithWarmupScheduler(t_warmup=args.t_warmup, t_max=args.t_max, alpha_f=args.alpha_f)
+logging.info('Build optimizer and learning rate scheduler')
+
+# Callbacks for logging
+logging.info('Build SpeedMonitor, LRMonitor, and CheckpointSaver callbacks')
```
dblalock, line 241:
```suggestion
logging.info('Building SpeedMonitor, LRMonitor, and CheckpointSaver callbacks')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+speed_monitor = SpeedMonitor(window_size=50)
+lr_monitor = LRMonitor()
+
+# Callback for checkpointing
+checkpoint_saver = CheckpointSaver(folder=args.save_checkpoint_dir, save_interval=args.checkpoint_interval)
+logging.info('Build SpeedMonitor, LRMonitor, and CheckpointSaver callbacks')
+
+# Recipes for training ResNet architectures on ImageNet in order of increasing training time and accuracy
+logging.info('Build algorithm recipes')
```
dblalock, line 250:
```suggestion
logging.info('Building algorithm recipes')
```

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+if args.recipe_name == 'mild':
+    algorithms = [
+        BlurPool(),  # Add anti-aliasing filters to strided convs and maxpools
+        ChannelsLast(),
```
dblalock, line 254: We should probably either explain all of the algorithms or none of them. And I'm not sure there's a concise way to explain them all. Maybe method card links? If we go that route, might look cleaner to deduplicate these lines, so that there's only one line instantiating a BlurPool(), one for a ChannelsLast(), etc.
```suggestion
        BlurPool(),
        ChannelsLast(),
```

dblalock (approved):
LGTM assuming it actually runs and reproduces our numbers (which I didn't check).

I had various suggestions for polishing it slightly, but nothing major.

dblalock:
To answer your questions:


**Thoughts on having example commands as a comment at the top of the document?**

This or putting them in a README seems fine. The one thing I'd say is that, assuming we keep this in its own directory, *something* has to go in the README and example commands seem like the most natural content. I'd also vote against replicating them in two places.

**What do people think about how the recipes are instantiated? Is it okay that there are no argparse for the individuals methods in the recipe?**

Seems good to me. This both constrains the user's choice (good for an example) and suggests a helpful pattern (thinking in recipes rather than individual algos).

**Should there be warnings about the crop resolution, loss function, and duration if a recipe is set, but these args are set to their default value?**

Can we just make it do the right thing by default? E.g., check the 'recipe' arg earlier in the script?

**Is it necessary to implement FFCV or Dali now to alleviate dataloader bottleneck? Or can this wait until a separate PR?**

Fine to put in a separate PR. Only reason to fuse it would be to save you work + compute time, and it sounds like you already have it working this way.

examples/benchmarks/train_resnet_imagenet1k.py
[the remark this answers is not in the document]
A-Jacobson: + 1 either a note that its experimental or leave it out. Thought mps had a lot of convergence problems.

A-Jacobson (approved):
Also LGTM. Great that you've included saving and loading too.

- on recipe overrides/hparams - i think what you've done here is a good 80%/20% effort. It's enough for reproduction and hackability but if someone wants to do hardcore research sweeps on nested hparams they should probably be using whichever config manager we land on. Otherwise the argparse is going to get out of hand fast.

- on readmes - I like having a readme vs comments at the top though that would require us to have a directory per model. It's mostly a personal preference for me since it would be easier to read in github. I don't feel strongly about it.

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+                    default='90ep')
+parser.add_argument('--eval_interval',
+                    help='How frequently to run the evaluation datasets',
+                    type=Time.from_timestring,
+                    default='1ep')
+parser.add_argument('--device', help='Device to run training on', choices=['gpu', 'cpu', 'tpu', 'mps'], default='gpu')
+parser.add_argument('--precision',
+                    help='Numerical precision for training',
+                    choices=['fp32', 'fp16', 'amp'],
+                    default='amp')
```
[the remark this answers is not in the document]
Landanjs: Great points! I didn't want to hide certain aspects of the configuration, so the user knows what is happening under-the-hood and knows what can be changed. Let me know if you don't think that is worthwhile though!

Alternatively, I can do `precision = 'amp' if torch.cuda.is_available() else 'fp32'`

examples/benchmarks/train_resnet_imagenet1k.py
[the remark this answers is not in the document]
Landanjs: I think it will be better to remove the device arparse in favor of selecting device based on `torch.cuda.is_available()`

examples/benchmarks/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,320 @@
...
+
+# Trainer arguments
+parser.add_argument('--seed', help='Random seed', type=int, default=17)
+parser.add_argument('--grad_accum', help='Gradient accumulation either an int or "auto"', default='auto')
+parser.add_argument('--max_duration',
+                    help='Duration to train specified in terms of Time',
+                    type=Time.from_timestring,
+                    default='90ep')
+parser.add_argument('--eval_interval',
+                    help='How frequently to run the evaluation datasets',
```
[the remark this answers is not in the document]
Landanjs: Done!

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Yeah, I hate this. I might just get rid of it since it doesn't need to be completely configurable and it is confusing

Landanjs: Jira created

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Done!

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: I will remove this argument as well in favor of fixing `grad_accum='auto'` since there doesn't seem to be an easy way for arg parse to take two argument types? (I didn't dig too deep...)

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Yep, this is needed to reproduce our recipes. Do you think it is okay to include?

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Yeah, I was ambitiously going to describe them all, but gave up quickly 😅 What if I add a link to our method card documentation in a comment before the if-statement?

eracah:
This is really cool! I love the fully-working example without any YAHP. Great work, @Landanjs!

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Moved these to README instead

examples/benchmarks/train_resnet_imagenet1k.py
Landanjs: Good point... Would it be good enough to add "specified as a Time string" to each argument and expect the user to read about our Time conventions? I'm not sure if I could sufficiently convey information in the argparse description

Landanjs:
OMG, that was a real struggle with the doc tests. I wound adding the markdown in `exclude_parameters` in conf.py since I didn't want to include this in any toctree yet. Is it okay as a temporary solution? 

Otherwise, hopefully I addressed all comments

examples/imagenet/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+These examples illustrate how to train models on ImageNet-1k using Composer.
+
+Prerequisites:
+- Install Compose: `pip install mosaicml`
+- ImageNet already downloaded on the system used for training
+
+## ResNet
+
+The `train_resnet_imagenet1k.py` script trains models from the ResNet architecture family e.g. ResNet-50, ResNet-101.
```
dblalock, line 11:
```suggestion
The `train_resnet_imagenet1k.py` script trains models from the ResNet architecture family (e.g. ResNet-50, ResNet-101).
```

examples/imagenet/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Few details about the script:
+- Models are pulled directly from torchvision, then wrapped into a `ComposerModel`
+- Saves checkpoints to "checkpoints/{run_name}" by default where `run_name` is random if not specified
+- Logs time-to-train, throughput, and learning rate throughout training.
```
dblalock, line 16:
```suggestion
This script:
- Pulls models directly from torchvision, then wraps them in a `ComposerModel`
- Saves checkpoints to "checkpoints/{run_name}" by default, where `run_name` is random if not specified
- Logs time-to-train, throughput, and learning rate throughout training.
```

examples/imagenet/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Example configurations:
+
+<!--pytest.mark.skip-->
```
dblalock, line 20: I'm assuming this doesn't do anything given the conf.py change.
```suggestion
```

examples/imagenet/README.md
```diff
@@ -0,0 +1,43 @@
...
+# Single/Multi GPU training (infers the number of GPUs available)
+composer train_resnet_imagenet1k.py /path/to/imagenet
+
+# Manually specify number of GPUs to use:
+composer -n $N_GPUS train_resnet_imagenet1k.py /path/to/imagenet
+
+# Mild ResNet recipe for fastest training to ~76.5% accuracy:
+composer train_resnet_imagenet1k.py /path/to/imagenet --recipe_name mild --train_crop_size 176 --eval_crop_size 224 --max_duration 36ep  --loss_name binary_cross_entropy
+
+# Medium ResNet recipe highest accuracy with similar training time as baseline:
```
dblalock, line 38:
```suggestion
# Medium ResNet recipe for highest accuracy with similar training time as the baseline:
```

examples/imagenet/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,297 @@
...
+parser.add_argument('--load_checkpoint_path', help='Path to the checkpoint to load', type=str)
+
+# Recipes
+parser.add_argument('--recipe_name',
+                    help='Either "mild", "medium" or "spicy" in order of increasing training time and accuracy',
+                    type=str,
+                    choices=['mild', 'medium', 'spicy'])
+
+# Logger parameters: progress bar logging is used by default
+# Only has Weights and Biases option to reduce the number of arguments. Other loggers can be substituted in the script
```
dblalock, line 85:
```suggestion
# To use a logger other than Weights and Biases, you'll need to modify the script appropriately
```

examples/imagenet/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,297 @@
...
+                    default='1ep')
+
+# Local storage checkpointing
+args = parser.parse_args()
+
+
+def _main():
+
+    # Divide batch sizes by number of devices if running multi-gpu training
+    if dist.get_world_size():
```
dblalock, line 110: Our get_world_size() [docs](https://docs.mosaicml.com/en/v0.9.0/api_reference/composer.utils.dist.html?highlight=get_world_size#composer.utils.dist.get_world_size) aren't that informative, but the torch function that this wraps seems to return either an int >= 1 if run in part of a process group and -1 otherwise. Current logic would always evaluate to True and choke on the -1.

```suggestion
    if dist.get_world_size() > 1:
```

examples/imagenet/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # Train dataset
+    logging.info('Building train dataloader')
+    train_transforms = transforms.Compose([
+        transforms.RandomResizedCrop(args.train_crop_size, scale=(0.08, 1.0), ratio=(0.75, 4.0 / 3.0)),
+        transforms.RandomHorizontalFlip(),
+    ])
+    train_dataset = ImageFolder(os.path.join(args.data_dir, 'train'), train_transforms)
+    # Nifty function to instantiate a PyTorch DistributedSampler based on your hardware setup
+    train_sampler = dist.get_sampler(train_dataset, drop_last=True, shuffle=True)
```
dblalock, line 125: Do we need this? I don't have this in the science repo but also wouldn't be _that_ surprised if I've been sampling wrong this whole time.

examples/imagenet/train_resnet_imagenet1k.py
```diff
@@ -0,0 +1,297 @@
...
+    logging.info('Built train dataloader\n')
+
+    # Validation dataset
+    logging.info('Building evaluation dataloader')
+    eval_transforms = transforms.Compose([
+        transforms.Resize(args.eval_resize_size),
+        transforms.CenterCrop(args.eval_crop_size),
+    ])
+    eval_dataset = ImageFolder(os.path.join(args.data_dir, 'val'), eval_transforms)
+    # Nifty function to instantiate a PyTorch DistributedSampler based on your hardware setup,
```
dblalock, line 148:
```suggestion

```

dblalock (commented):
LGTM. Just some minor edits. For some reason the github UI isn't giving me the option to approve, request changes, etc, though.

dblalock:
Oops. I opened this tab >3h ago and didn't see that it got merged. Well, now we have suggestions for the next PR.

PR #1443, opened 2022-08-24
Algolia doc search

nqn:
Prototype of real time search with Algolia docsearch
Screen Shot 2022-08-24 at 3 27 18 PM


docs/source/_templates/base.html
```diff
@@ -0,0 +1,115 @@
...
+    {%- block regular_scripts -%}
+    {% for path in script_files -%}
+      {{ js_tag(path) }}
+    {% endfor -%}
+    {%- endblock regular_scripts -%}
+
+    {# Theme-related JavaScript code #}
+    {%- block theme_scripts -%}
+    {%- endblock -%}
+
```
lupesko, line 112: nit: is this empty line needed?

lupesko (approved):
Looks good beyond the small nit

nqn:
There is an issue with require.min.js and load order (it messes with algolia's library). So need to resolve that before we can land this.

It is tricky, as the issue occurs once require.min.js has been cached

PR #1447, opened 2022-08-25
Upgrade timm and deepspeed

mvpatel2000:
Upgrade these packages and switch deepspeed to `<=, >=` instead of `==`. This will let dependabot only open PRs when a new version is released as opposed to every upgrade


mvpatel2000:
I'm closing because I want dependabot to do it directly

PR #1448, opened 2022-08-25
Create dependabot.yml

mvpatel2000:


hanlint (commented):
to avoid spam and timeliness, can we limit it to maybe `open-pull-requests-limit` to 5? Otherwise, expect a lot of PR spam :)

mvpatel2000:
> to avoid spam and timeliness, can we limit it to maybe `open-pull-requests-limit` to 5? Otherwise, expect a lot of PR spam :)

Done. We probably won't get too many because we don't hard pin versions with `==`. Eg we'll only get 5 when this turns on. But yes, agree let's add it

hanlint (approved):

PR #1449, opened 2022-08-26
Fix incorrect deps group in `streaming` requirement

hanlint:
Should be `streaming` not `s3`, as users should do `pip install mosaicml[streaming]`


eracah (approved):
Good catch! LGTM

PR #1455, opened 2022-08-26
Bump dependencies to next major version

mvpatel2000:
- bump moto[s3]


mvpatel2000:
Actually just editing dependabot PR

Issue #1429, opened 2022-08-19
Evaluation API


ananyahjha93:
## 🚀 Feature Request

Adding more ways to use the current eval API.

## Motivation

The eval API feels a bit restrictive and requires us to write duplicate code. For large eval harnesses, or places where only eval is required (nearest neighbor augmented LMs, a whole bunch of large pre-trained LMs etc.); passing ``eval_dataloader`` to trainer is not sufficient for calling ``trainer.eval()``.

If ``eval_dataloader`` is passed to the trainer, I am assuming the Composer model will already have ``validate`` and ``metric`` functions.

```
trainer = Trainer(
  eval_dataloader=...,
)

# pass dataloader and metric again
# can it reuse metric defined in composer model?
trainer.eval(dataloder, metric)
```

Simply calling ``trainer.eval()`` without the need for passing additional ``dataloader``, ``metric`` will also be useful for the case where the trainer already has been passed multiple evaluators (this is specially a really nice API design). 

```
eval1 = Evaluator(...)
eval2 = Evaluator(...)

trainer = Trainer(
eval_dataloader=[eval1, eval2],
)

trainer.eval()
```

## Additional context

Initial discussion on slack in composer channel.

ananyahjha93:
@hanlint happy to continue the discussion here and if this change makes sense for the API.

hanlint:
Thanks @ananyahjha93, I'm thinking we can revamp the `eval` API to mirror the `trainer` API, so that users can either do:
```
trainer = Trainer(eval_dataloader=[eval1, eval2, ..])
trainer.eval()
```
or
```
trainer = Trainer(...)
trainer.eval(
    eval_dataloader=[eval1, eval2]
)
```

With that design, the eval API would be something like:
```
def eval(self, 
    eval_dataloader: Union[DataLoader, Evaluator], 
    metrics: Optional[Metric] = None
):
```
where user can either provide an `dataloader` + `metrics`, or just provide evaluators.

Would that make sense?

ananyahjha93:
@hanlint yes, that definitely makes the API uniform

hanlint:
OK great, we'll implement this in the next release (v0.10), while maintaining some backwards compatibilty with the current API.

PR #1479, opened 2022-08-29
Update Eval API

hanlint:
Based on discussions in #1429 , it's clear that our Eval API should be updated to be consistent with the Trainer API, eg.
```
trainer = Trainer(
    eval_dataloader=...
)
trainer.eval()
```
or
```
trainer = Trainer(
    ...
)

trainer.eval(
    eval_dataloader=...
)
```

Instead, currently the `eval` API is fairly constrained to evaluating a single dataloader/metric, and also requires re-passing in the metrics, even though they have already been defined by the user in the `model`. This PR implements the new Evaluation API.

Todo:
- [x] update docstrings / docs
- [x] update tests

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1767,10 +1767,10 @@ class Trainer:
             assert evaluator.eval_interval is not None, 'eval_interval should have been set on __init__() or fit()'
             assert evaluator.subset_num_batches is not None, 'subset_num_batches should have been set on __init__() or fit()'
             if evaluator.eval_interval(self.state, event):
-                self.eval_loop(dataloader=evaluator.dataloader,
-                               dataloader_label=evaluator.label,
-                               subset_num_batches=evaluator.subset_num_batches,
-                               metrics=self.state.eval_metrics[evaluator.label])
+                self._eval_loop(dataloader=evaluator.dataloader,
+                                dataloader_label=evaluator.label,
+                                subset_num_batches=evaluator.subset_num_batches,
+                                metrics=self.state.eval_metrics[evaluator.label])
 
     def _train_batch(self, use_grad_scaling: bool) -> Dict[str, torch.Tensor]:
         """Compute loss by training on a full batch of data.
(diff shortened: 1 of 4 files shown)


composer/trainer/trainer.py
```diff
@@ -978,7 +978,7 @@ def __init__(
         if eval_dataloader is None:
             evaluators: List[Evaluator] = []
         else:
-            eval_metrics = self.state.model.get_metrics(is_train=False)
+            eval_metrics = deepcopy(self.state.model.get_metrics(is_train=False))
```
A-Jacobson, line 981: Doing this would make it so that the metrics on the model aren't updated. is this intended?

hanlint: Yes, in the case of multiple metrics, since the model metrics only returns `Accuracy()`, we need to duplicate them.

A-Jacobson (approved):
LGTM, left note about copying metrics. copying them feels strange to me since the model parameters are updated in place + if the metrics belong and are meant to be accessed from the trainer.. why are they defined on the model?

composer/trainer/trainer.py
```diff
@@ -2140,12 +2140,155 @@ def predict_batch_end(self, state: State, logger: Logger) -> None:
...
 
     def eval(
+        self,
+        eval_dataloader: Optional[Union[Iterable, DataSpec, Evaluator, Sequence[Evaluator]]] = None,
+        subset_num_batches: int = -1,
+        # deprecated arguments below
+        dataloader: Optional[Any] = None,
+        dataloader_label: Optional[Any] = None,
+        metrics: Optional[Any] = None,
+        log_level: Optional[Any] = None,
```
eracah, line 2150:
```suggestion
        self,
        eval_dataloader: Optional[Union[Iterable, DataSpec, Evaluator, Sequence[Evaluator]]] = None,
        subset_num_batches: int = -1,
        # deprecated arguments below
        **kwargs,
```
How about we just set this to kwargs and then we check if any of the deprecated arguments are specified inside the kwargs? This way we don't have args in the signature that we don't want people to use, which would be confusing.

composer/trainer/trainer.py
```diff
@@ -2140,12 +2140,155 @@ def predict_batch_end(self, state: State, logger: Logger) -> None:
...
+                ``eval_dataloader`` provided to the trainer init().
+            subset_num_batches (int, optional): Evaluate on this many batches. Default to ``-1`` (the entire
+                dataloader. Can also be provided in the trainer init()as ``eval_subset_num_batches``.
+            dataloader: Removed, do not use.
+            dataloader_label: Removed, do not use.
+            metrics: Removed, do not use.
+            log_level: Removed, do not use.
+
+        """
+        if any([dataloader, dataloader_label, metrics]):
```
eracah, line 2239:
```suggestion
        if any([k in kwargs for k in ['dataloader', 'dataloader_label', 'metrics', 'log_level']]):
```
see above comment for context. Basically, we have a catch-all kwargs that we check for deprecated arguments.

eracah (approved):
Looks pretty good, but maybe using a catch-all kwargs instead of deprecated arguments is easier?

composer/trainer/trainer.py
hanlint: Ah i like this better, thanks!

PR #1484, opened 2022-08-30
Clean up `save_checkpoint`

hanlint:
This PR cleans up the code in `save_checkpoint` for readability, and simplifies the logic.


composer/trainer/trainer.py
```diff
@@ -2396,7 +2402,12 @@ def save_checkpoint(self, name: str = 'ep{epoch}-ba{batch}-rank{rank}', *, weigh
         Returns:
             str or None: See :func:`.save_checkpoint`.
         """
-        return save_checkpoint(state=self.state, filename=name, weights_only=weights_only)
+        return save_checkpoint(
```
eracah, line 2405: any way we can fix both functions have the same name?

hanlint: I changed it to be:
```
def save_checkpoint(..):
    return checkpoint.save_checkpoint(...)
```

to clarify the save_checkpoint namespace a bit, does that work?

eracah (approved):
LGTM

mvpatel2000 (approved):
LGTM. I like the checkpoint.fn_name -- good idea

PR #1486, opened 2022-08-30
Remove hparams from `test_precision` and `test_state`

hanlint:
This PR removes the requirement to use yahp in `test_precision` and `test_state`. The state test is also refactored for readability.

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -3,18 +3,16 @@
 
 import pathlib
 import random
-from unittest.mock import Mock
 
+import pytest
 import torch
 import torch.nn.functional as F
-from torch.functional import Tensor
+from torch.utils.data import DataLoader
 
-from composer.algorithms import ChannelsLast
-from composer.core import DataSpec, Precision, State
-from composer.core import state as state_module
+from composer.core import Precision, State
 from composer.core.types import Batch
-from composer.datasets.dataset_hparams import DataLoaderHparams, DatasetHparams
 from tests.common import SimpleModel, assert_state_equivalent
+from tests.common.datasets import RandomClassificationDataset
 
 
 def random_tensor(size=(4, 10)):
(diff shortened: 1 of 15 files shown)


tests/test_state.py
```diff
@@ -24,14 +22,15 @@ def random_tensor(size=(4, 10)):
...
     model = SimpleModel()
     optimizers = torch.optim.Adadelta(model.parameters())
-    state = State(model=model,
-                  run_name=f'{random.randint(0, 100)}',
-                  grad_accum=random.randint(0, 100),
-                  rank_zero_seed=random.randint(0, 100),
-                  precision=Precision.AMP,
-                  max_duration=f'{random.randint(0, 100)}ep',
-                  optimizers=optimizers,
-                  algorithms=[ChannelsLast()])
```
eracah, line 34: why do we remove `ChannelsLast`?

tests/test_state.py
```diff
@@ -58,89 +57,70 @@ def train_one_step(state: State, batch: Batch) -> None:
...
 
-def test_state_batch_set_item(monkeypatch):
-    state = get_dummy_state()
-    state.batch = [1, 2]
-    mock_batch_set = Mock()
-    mock_batch_set.return_value = [7, 10]
-    monkeypatch.setattr(state_module, 'batch_set', mock_batch_set)
-    state.batch_set_item(key=1, value=154)
-    mock_batch_set.assert_called_once_with([1, 2], key=1, value=154)
+    assert state.batch_get_item(key) == val
```
eracah, line 110: good work. These mock tests were a bit overkill. The guy who wrote them was a little overzealous 😉
I think I put the mocks because I already tested `batch_get`, but probably easier just to test without mocks here.

eracah (approved):
LGTM! Good work starting the de-hparam-ification, @hanlint !

tests/test_state.py
hanlint: The `SimpleModel` has no convolution layers, so ChannelsLast has no effect

PR #1487, opened 2022-08-30
Random Names with Fixed Seed

mvpatel2000:
- adds test to verify names are different even with seed fixed
- verifies test failed before and now passes


mvpatel2000:
Enabling automerge since its a small change

hanlint:
Does this mean that for say a distributed training run, each rank would get a different run_name?

mvpatel2000:
> Does this mean that for say a distributed training run, each rank would get a different coolname? So we would get:

Hm... ok yea... I agree that's bad. Let me change it to do only on rank0 and then share

composer/trainer/trainer.py
```diff
@@ -278,6 +279,8 @@ def _get_ddp_sync_strategy(ddp_sync_strategy: Optional[Union[str, DDPSyncStrateg
 
 
 def _generate_run_name() -> str:
+    # change coolname randomness for different names with same seed
+    coolname.replace_random(random.Random(os.urandom(128)))
     # prefixing with the time so experiments sorted alphabetically will have the latest experiment last
     generated_run_name = str(int(time.time())) + '-' + coolname.generate_slug(2)
     run_name_list = [generated_run_name]
```
mvpatel2000, line 286: @hanlint  did you get those run_names from running it? I believe the lines below here (github wont let me highlight for some reason) should take `run_name` generated on rank0 and override the `run_name`s generated by other ranks

hanlint: Ah, no I took the opportunity to make those up :) If it's already handled/tested with DDP, then great

mvpatel2000: LOL

mvpatel2000: I was very unsure because the names were solid

mvpatel2000: I'm re-enabling automerge and requesting review!

hanlint (approved):

Issue #1473, opened 2022-08-28
Logging issue with 0.9.0 and current dev branch


ananyahjha93:
** Environment **

- OS: Ubuntu 20.04
- Hardware (GPU, or instance type): 8xA100
- cuda: 11.3
- cudnn: 8
- pytorch: 1.12.1
- composer: dev branch installed from source/0.9.0 installed from pip
- transformers: 4.21.2

** To reproduce

I have the following definition of bloom model, mostly copied from the GPT2 definition within composer.

```
def create_bloom(
    model_name: str,
    tokenizer_name: str,
    use_pretrained: Optional[bool] = False,
    model_config: Optional[dict] = None,
    gradient_checkpointing: Optional[bool] = False,
) -> ComposerModel:

    if not model_config:
        model_config = {}

    if use_pretrained:
        model = transformers.AutoModelForCausalLM.from_pretrained(model_name, **model_config)
    else:
        config = transformers.AutoConfig.from_pretrained(model_name, **model_config)
        model = transformers.AutoModelForCausalLM.from_config(config)

    tokenizer = transformers.AutoTokenizer.from_pretrained(tokenizer_name)

    if gradient_checkpointing:
        model.gradient_checkpointing_enable()

    return HuggingFaceModel(model=model, tokenizer=tokenizer, metrics=[HFCrossEntropy(), Perplexity()])
```

There are 2 issues, one with the 0.9.0 release and the other with the dev branch.

1. Running LM training with grad accumulation with 0.9.0 doesn't plot HF metrics in wandb, but has correct step counts while logging metrics.

You can see that the logs don't show Perplexity and CrossEntropy metrics.
![image](https://user-images.githubusercontent.com/7491256/187063210-18da4fdd-381e-4255-af27-8ab55bf12d05.png)
![image](https://user-images.githubusercontent.com/7491256/187063229-3227b5d6-e9be-419c-8772-0eb8f303292b.png)

2. Running LM training with grad accumulation with the dev branch plots HF metrics but gets the step count while plotting these metrics completely wrong.

You can see metrics being plotted for 266 step with only 38 batches being trained.
![image](https://user-images.githubusercontent.com/7491256/187063267-d8be0b1d-4d5f-4e7c-a68e-cc3ebda9db5e.png)
![image](https://user-images.githubusercontent.com/7491256/187063280-921fe86e-d57a-4ec0-9424-fd697febc32d.png)

[...]

PR #1488, opened 2022-08-31
Fix WandBLogger bug with inaccurate step count

eracah:
CO-995
Fixes https://github.com/mosaicml/composer/issues/1473

Adds manual step specification for WandB to fix inaccurate step bug.

[Before](https://wandb.ai/mosaic-ml/evan-composer/runs/325coxke)
[After](https://wandb.ai/mosaic-ml/evan-composer/runs/2pxt6yrn)


mvpatel2000 (approved):
LGTM

PR #1490, opened 2022-09-01
Add CometML Support

eracah:
CO-1027

* Add `CometMLLogger`
* Along for the ride:
    * log hyperparameters for all loggers in entrypoint (not just wandb)

Manual/E2E Tests:
### Local Custom Entrypoint
[Comet results](https://www.comet.com/mosaic-ml/evan-test/d7da1dce90024c26932b73fbfa4d5963?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)
[WandB results for comparison](https://wandb.ai/mosaic-ml/evan-composer/runs/80u1a9v9)

### Local YAHP Entrypoint
[CometML](https://www.comet.com/mosaic-ml/evan-test/6c5d1344af754c98a105f141e5e3352a?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)
[WandB](https://wandb.ai/mosaic-ml/evan-tb-test/runs/38jrwz77)

### Remote YAHP Entrypoint
[CometML](https://www.comet.com/mosaic-ml/evan-test/40b2778f2d7c4a3bac402ba533b7e01f)
[WandB](https://wandb.ai/mosaic-ml/evan-tb-test/runs/1zpccxcq)

### Remote Custom Entrypoint
[Remote Bert Run Results with Comet](https://www.comet.com/mosaic-ml/evan-test/f5e1877a22ef4c65a9906611f03b7393?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)
[Corresponding WandB run](https://wandb.ai/mosaic-ml/evan-tb-test/runs/3hvqimh3?workspace=user-evan-mosaicml)
* note: hyperparameters will be different because we are using Moin's custom entrypoint which explicitly logs hyperparameters to wandb


composer/loggers/cometml_logger.py
```diff
@@ -0,0 +1,102 @@
...
+        name: Optional[str] = None,
+        rank_zero_only: bool = True,
+        exp_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        try:
+            from comet_ml import Experiment
+        except ImportError as e:
+            raise MissingConditionalImportError(extra_deps_group='comet_ml',
+                                                conda_package='comet_ml',
+                                                conda_channel='conda-forge') from e
```
mvpatel2000, line 57: Note: composer currently doesn't support `conda`. What exactly does this do?

composer/loggers/cometml_logger.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        if self._enabled:
+            assert self.experiment is not None
+            self.experiment.log_metrics(dic=metrics, step=step)
+
+    def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
+        if self._enabled:
+            assert self.experiment is not None
+            self.experiment.log_parameters(hyperparameters)
```
mvpatel2000, line 102: I haven't looked through the comet API, but just want to double check this doesnt cause internal step issues?

docs/source/notes/run_name.md
```diff
@@ -53,13 +53,11 @@ See {class}`~.CheckpointSaver` for more information on specifying the arguments
...
+#### Experiment Tracking Loggers
 
-The {class}`~.TensorboardLogger` will save all the logs for a run to a folder called `run_name` and the name of each run in the Tensorboard GUI will be `run_name`.
-
-#### Weights and Biases Logger
-
-The `run_name` you specify will be used by the {class}`~.WandBLogger` as the run name for Weights and Biases.
+* The {class}`~.TensorboardLogger` will save all the logs for a run to a folder called `run_name` and the name of each run in the Tensorboard GUI will be `run_name`.
+* The `run_name` you specify will be used by the {class}`~.WandBLogger` as the run name for Weights and Biases.
+* The `run_name` you specify will be used by the {class}`~.CometMLLogger` as the run name for your Comet experiment.
```
mvpatel2000, line 60: Looks like this is deleting the old docs for TB and W&B?

setup.py
```diff
@@ -144,6 +144,8 @@ def package_files(prefix: str, directory: str, extension: str):
     'wandb>=0.13.2,<0.14',
 ]
 
+extra_deps['comet_ml'] = ['comet_ml>=3.31.12']
```
mvpatel2000, line 147:
```suggestion
extra_deps['comet_ml'] = ['comet_ml>=3.31.12,<4.0.0']
```
Dependabot will auto-bump, and we want unit tests on major changes.

mvpatel2000 (commented):
Are integration tests here feasible? Ideally we'd have something set up

composer/loggers/cometml_logger.py
```diff
@@ -0,0 +1,102 @@
...
+        exp_kwargs['log_code'] = log_code
+        exp_kwargs['log_graph'] = log_graph
+
+        self.name = name
+        self._rank_zero_only = rank_zero_only
+        self._exp_kwargs = exp_kwargs
+        self.experiment: Optional[Experiment] = None
+
+    def init(self, state: State, logger: Logger) -> None:
+        import comet_ml
```
karan6181, line 79: I see that only `Experiment` is used in `comet_ml` package. Is it better to just import `Experiment` from `comet_ml`? For example, `from comet_ml import Experiment`.

composer/loggers/cometml_logger.py
```diff
@@ -0,0 +1,102 @@
...
+        if not self._rank_zero_only:
+            self.name += f'-rank{dist.get_global_rank()}'
+
+        if self._enabled:
+            self.experiment = comet_ml.Experiment(**self._exp_kwargs)
+            self.experiment.set_name(self.name)
+
+    def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        if self._enabled:
+            assert self.experiment is not None
```
karan6181, line 96: Suggestion is to raise a `ValueError` or something similar Exception rather than using `assert` which gives end-user a proper error code with descriptive error message.

karan6181: Also, in what scenario will `self.experiment` be `None` ? I see that it is getting initialized during `init()` method? will the `init()` method will always gets called before `log_metrics()`?

composer/loggers/cometml_logger.py
```diff
@@ -0,0 +1,102 @@
...
+            self.experiment.log_metrics(dic=metrics, step=step)
+
+    def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
+        if self._enabled:
+            assert self.experiment is not None
```
karan6181, line 101: Same comment as in `log_metrics()` method.

docs/source/notes/run_name.md
```diff
@@ -53,13 +53,11 @@ See {class}`~.CheckpointSaver` for more information on specifying the arguments
 
 In addition to checkpointing, loggers also use the `run_name` for default logging.
 
-#### Tensorboard Logger
+#### Experiment Tracking Loggers
```
karan6181, line 56: Heading changed to `Experiment Tracking Loggers` from `Tensorboard Logger` but the below comment line still refers to old name `TensorboardLogger`. Do we also need to change that?

karan6181 (commented):
Thank you for your Contribution. Left some comments. Also, do you have a plan to add a unit test for this change in `tests/loggers/test_cometml_logger.py` ?

docs/source/notes/run_name.md
eracah: No it's fine. It just moves the Weights and Biases line down to new category `Experiment Tracking Loggers`

composer/loggers/cometml_logger.py
eracah: I don't think it makes much difference. I guess slightly faster to do your suggestion. I figured it's a bit more readable to show the namespace of `Experiment`

composer/loggers/cometml_logger.py
eracah: Yes you're right. this assert statement will never be False. It is solely there to please the `pyright` type checker. We have to explicitly show that `self.experiment` is not None to the type checker. See the TensorboardLogger code.

eracah: `init` usually gets called first because it happens at the beginning of trainer initiailziation. See the pseudocode [here](https://docs.mosaicml.com/en/stable/getting_started/welcome_tour.html#events-engines-and-state)

eracah: However, you raise a good point. You can edit the run name with the `comet_ml.Experiment`, so we don't have to wait until `init` to create `Experiment`. I will change that. Thanks, @karan6181 !

docs/source/notes/run_name.md
eracah: the three comments below it actually refer to the Tensorboard, WandB, and CometML, which all fall under the category of Experiment tracking loggers.

composer/loggers/cometml_logger.py
eracah: `log_metrics` could, but we explicitly pass `step`, so seems to be fine. `log_parameters` is just for hyperparameters. I'm not sure why they allow you to pass step to `log_parameters`

composer/loggers/cometml_logger.py
eracah: yeah it's a weird practice I'm following from our [style guide](https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#51-external-dependencies)

composer/loggers/cometml_logger.py
mvpatel2000: ok, resolving. Just wanted to double check there's no issues if we don't pass in `step` here. Agree its a bit odd that it takes it

eracah: yeah I checked the dashboard and it just logged them in a table, like configs in wandb

composer/loggers/cometml_logger.py
eracah: ok fixed!

composer/loggers/cometml_logger.py
eracah: fixed

.ci/Jenkinsfile
```diff
@@ -271,6 +272,10 @@ void runPytest(String pDockerImage, String markers, String extraDeps, Boolean is
                 name: 'P_JENKINS_USERNAME_PASSWORD_CREDENTIALS',
                 value: "IGNORED_WANDB_USERNAME,WANDB_API_KEY=$wandbCredentialsId",
             ),
+            string(
+                name: 'P_JENKINS_USERNAME_PASSWORD_CREDENTIALS',
+                value: "WANDB_API_KEY=$cometmlCredentialsId",
```
hanlint, line 277: should this be different, given the above for wandb?

eracah: good catch! fixed

eracah:
Ok, @karan6181, @mvpatel2000, I added a unit test. Let me know how it looks and if this PR is ready for merge.

mvpatel2000 (approved):
LGTM! Ideally we get integration tests at some point soon, but that's a later issue

PR #1491, opened 2022-09-01
Remove hparams from checkpoint tests

hanlint:
This PR refactors the checkpoint tests to not use hparams objects, for the following tests:
* test_autoresume
* test_load_weights
* test_load_weights_object_store
* test_save_overwrite
* test_checkpoint
* test_different_run_name

These tests are pretty critical to our checkpointing infra, so I ask the reviewers to please review the logic carefully and make sure the tests are functionally identical to before the refactor.

The existing tests had some functionality (e.g. testing event calls) that were more appropriate to be put in `test_events.py`

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -2,10 +2,129 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pytest
+import torch
+from torch.utils.data import DataLoader
 
-from composer.core import Event
+from composer import Trainer
+from composer.core import Event, Time
+from composer.core.time import TimeUnit
+from tests.common import RandomClassificationDataset, SimpleModel
+from tests.common.events import EventCounterCallback
 
 
 @pytest.mark.parametrize('event', list(Event))
 def test_event_values(event: Event):
     assert event.name.lower() == event.value
+
+
+class TestEventCalls:
+
+    eval_subset_num_batches = 5
+    train_subset_num_batches = 5
+
+    def get_trainer(self, **kwargs):
+        model = SimpleModel()
+        optimizer = torch.optim.Adam(model.parameters())
+
+        return Trainer(
+            model=model,
+            train_dataloader=DataLoader(
+                dataset=RandomClassificationDataset(),
+                batch_size=8,
+                shuffle=False,
+            ),
+            eval_dataloader=DataLoader(
+                dataset=RandomClassificationDataset(),
+                batch_size=16,
+                shuffle=False,
+            ),
+            grad_accum=2,
+            precision='fp32',
+            train_subset_num_batches=self.train_subset_num_batches,
+            eval_subset_num_batches=self.eval_subset_num_batches,
+            max_duration='2ep',
+            optimizers=optimizer,
+            callbacks=[EventCounterCallback()],
+            **kwargs,
+        )
+
+    @pytest.mark.parametrize('world_size', [
+        pytest.param(1),
+        pytest.param(2, marks=pytest.mark.world_size(2)),
+    ])
+    @pytest.mark.parametrize('device,deepspeed_zero_stage', [
+        pytest.param('cpu', None, id='cpu-ddp'),
+        pytest.param('gpu', None, id='gpu-ddp', marks=pytest.mark.gpu),
+        pytest.param('gpu', 0, id='deepspeed-zero0', marks=pytest.mark.gpu),
+        pytest.param('gpu', 1, id='deepspeed-zero1', marks=pytest.mark.gpu),
+        pytest.param('gpu', 2, id='deepspeed-zero2', marks=pytest.mark.gpu),
+    ])
+    @pytest.mark.parametrize('save_interval', ['1ep', '1ba'])
+    def test_event_calls(self, world_size, device, deepspeed_zero_stage, save_interval):
+        save_interval = Time.from_timestring(save_interval)
+
+        if deepspeed_zero_stage:
+            deepspeed_config = {'zero_optimization': {'stage': deepspeed_zero_stage}}
+        else:
+            deepspeed_config = None
+
+        trainer = self.get_trainer(
+            device=device,
+            deepspeed_config=deepspeed_config,
+            save_interval=save_interval,
+            eval_interval=save_interval,
+        )
+        trainer.fit()
+
+        self._assert_expected_event_calls(trainer, save_interval, num_epochs=2)
+
+    def _assert_expected_event_calls(self, trainer: Trainer, eval_interval: Time, num_epochs: int):
+        state = trainer.state
+
+        assert state.dataloader_len is not None
+        total_steps = num_epochs * int(state.dataloader_len)
+        total_microbatches = total_steps * state.grad_accum
+
+        if eval_interval.unit == TimeUnit.BATCH:
+            total_evals = total_steps // int(eval_interval)
+        elif eval_interval.unit == TimeUnit.EPOCH:
+            total_evals = num_epochs // int(eval_interval)
+        else:
+            total_evals = 0
+
+        if trainer.state.evaluators:
+            steps_per_eval = self.eval_subset_num_batches
+            total_eval_steps = total_evals * steps_per_eval * len(trainer.state.evaluators)
+        else:
+            total_eval_steps = 0
+
+        expected_num_calls = {
+            Event.INIT: 1,
+            Event.EPOCH_START: num_epochs,
+            Event.BATCH_START: total_steps,
+            Event.AFTER_DATALOADER: total_steps,
+            Event.BEFORE_FORWARD: total_microbatches,
+            Event.AFTER_FORWARD: total_microbatches,
+            Event.BEFORE_LOSS: total_microbatches,
+            Event.AFTER_LOSS: total_microbatches,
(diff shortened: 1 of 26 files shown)


tests/trainer/test_checkpoint.py
```diff
@@ -230,275 +182,48 @@ def test_ignore_params(remove_field_paths: List[List[str]], filter_params: List[
...
-    # load only model weights
-    second_trainer_hparams.load_path = checkpoint_a_file_path[0]
-    second_trainer_hparams.load_weights_only = True
-    second_trainer_hparams.load_strict_model_weights = True
-    # setup a new optimizer
-    second_trainer_hparams.optimizers = AdamWHparams()
-
-    # setup a new LR scheduler
-    assert isinstance(second_trainer_hparams.max_duration, str)
-    second_trainer_hparams.schedulers = [CosineAnnealingScheduler(t_max=second_trainer_hparams.max_duration)]
```
hanlint, line 269: I removed the check for the different scheduler type, as I wasn't sure what its purpose is in this test.. does keeping the same scheduler somehow change how the weights are loaded?

tests/trainer/test_checkpoint.py
```diff
@@ -230,275 +182,48 @@ def test_ignore_params(remove_field_paths: List[List[str]], filter_params: List[
...
-
-    # Create checkpoint in seperate folder to load. Optionally delete original checkpoint by moving it.
-    if delete_local_checkpoint:
-        shutil.move(checkpoint_a_folder, checkpoint_b_folder)
-    else:
-        shutil.copytree(checkpoint_a_folder, checkpoint_b_folder, symlinks=True)
-    # Recreate symlink in new folder
-    new_latest_path = os.path.join(checkpoint_b_folder, latest_checkpoint)
-    os.remove(new_latest_path)
-    os.symlink(final_checkpoint, new_latest_path)
```
hanlint, line 350: I also removed this logic here, and removed the use of a `second` folder. Am I missing something here?

hanlint:
I bit unsure why these tests are failing.. unable to reproduce locally and seem unrelated to my PR's changes.

tests/test_events.py
```diff
@@ -2,10 +2,130 @@
...
+from tests.common import RandomClassificationDataset, SimpleModel
+from tests.common.events import EventCounterCallback
 
 
 @pytest.mark.parametrize('event', list(Event))
 def test_event_values(event: Event):
     assert event.name.lower() == event.value
+
+
+@pytest.mark.skip
```
mvpatel2000, line 20: Why is there a skip?

tests/test_events.py
```diff
@@ -2,10 +2,130 @@
...
+class TestEventCalls:
```
mvpatel2000, line 21: Why does this need to be a class?

tests/trainer/test_checkpoint.py
```diff
@@ -230,275 +135,49 @@ def test_ignore_params(remove_field_paths: List[List[str]], filter_params: List[
...
-    del device  # unused. Set automatically
-    if not isinstance(composer_trainer_hparams.train_dataset, SyntheticHparamsMixin):
-        pytest.skip('Checkpointing tests require synthetic data')
-        return
-    if not isinstance(composer_trainer_hparams.val_dataset, SyntheticHparamsMixin):
-        pytest.skip('Checkpointing tests require synthetic data')
-        return
-    if use_object_store:
-        pytest.importorskip('libcloud')
+class TestCheckpointLoading:
```
mvpatel2000, line 138: same question, why does this need to be a class?

tests/trainer/test_checkpoint.py
mvpatel2000: I believe it's to test `load_weights` doesn't override other params, eg scheduler? I'm not super sure. it might be nice to include in the test though to verify `load_weights` doesn't touch other parts

tests/trainer/test_checkpoint.py
mvpatel2000: So what's happening is `autoresume` first checks for local checkpoints, and if there aren't any, checks the object stores for checkpoints to resume from. 

The logic here is to say: 1) backup checkpoint to a second folder so the original trainer can reload it and 2) (optionally) delete the checkpoint so we can test if creating a new trainer reloads from object store if local checkpoint is missing.

We don't need the second folder anymore since we can keep an instantiated copy of the original trainer to use for comparisons. However, we do still need to check if deleting the original file (`delete_local_checkpoint`) works. 

I believe your refactor looks like it handles this correctly to me :)

tests/trainer/test_checkpoint.py
```diff
@@ -507,290 +186,333 @@ def test_checkpoint_with_object_store_logger(
...
-
-    assert_checkpoints_equivalent(
-        checkpoint_file_a=first_trainer_final_checkpoint_filepath,
-        checkpoint_file_b=second_trainer_final_checkpoint_filepath,
+        # TODO (mihir): should be checking the entire state
+        # not just the model?
+        self._assert_weights_equivalent(
+            trainer_1.state.model,
+            trainer_2.state.model,
+        )
```
mvpatel2000, line 280:
```suggestion
        self._assert_weights_equivalent(
            trainer_1.state.model,
            trainer_2.state.model,
        )
```
LGTM! yes, it should be comparing the entire state, since with `autoresume` there should be no difference from continuing the original vs. restarting from the new one

tests/trainer/test_checkpoint.py
```diff
@@ -507,290 +186,333 @@ def test_checkpoint_with_object_store_logger(
...
+        trainer_1.fit()
+        trainer_1.close()
+
+        trainer_2 = self.get_trainer(
+            load_path=os.path.join('first', 'ep2.pt'),
+            seed=67890,
+        )
+
+        assert trainer_1.state.run_name != trainer_2.state.run_name
```
mvpatel2000, line 298:
```suggestion
    def test_different_run_names(self):

        trainer_1 = self.get_trainer(
            save_folder='first/',
            seed=12345,
        )
        trainer_1.fit()
        trainer_1.close()

        trainer_2 = self.get_trainer(
            load_path=os.path.join('first', 'ep2.pt'),
            seed=12345,
        )

        assert trainer_1.state.run_name != trainer_2.state.run_name
```
I believe it should now generate different names even if the seed is the same. It might be worth changing this test to use the same seed and then deleting the new test file I added `tests/trainer/test_run_name.py` to simplify tihs

tests/trainer/test_checkpoint.py
```diff
@@ -507,290 +186,333 @@ def test_checkpoint_with_object_store_logger(
...
+        )
+        trainer_1.fit()
+        trainer_1.close()
+
+        if save_overwrite:
+            ctx = contextlib.nullcontext
+        else:
+            ctx = functools.partial(pytest.raises, FileExistsError)
+
+        with ctx():  # expect FileExistsError if save_overwrite=False
```
mvpatel2000, line 316: (minor) nit: would prefer to one-line or define `ctx = None` at top. Surprised lint isn't complaining about this.

mvpatel2000 (commented):
Wow!! Amazing work 🚀 . This is _so_ much cleaner without Hparams. I left a few comments, biggest one being I don't get the reason why we have classes used everywhere.

Because this is very subtle and critical code, I would prefer to do a second pass once comments are resolved, especially because I might spot something on a second read through I missed the first time.

tests/trainer/test_checkpoint.py
hanlint: Since it's an `if .., else`, the `ctx = None` would always be overridden, so it should be ok?

mvpatel2000: My personal preference is to always put a `ctx=None` even if its always overridden to make scope clear. Even though python doesn't scope indents, most people tend to think in those scope levels, so its nice to be explicit about it. With that said, this is a minor nit you are free to ignore without further discussion :)

tests/trainer/test_checkpoint.py
hanlint: Good point, I added a new test to check that the statefulcallback was not touched.

tests/trainer/test_checkpoint.py
hanlint: OK understood, added the `ctx = None`.

tests/test_events.py
```diff
@@ -56,9 +55,6 @@ def get_trainer(self, **kwargs):
     @pytest.mark.parametrize('device,deepspeed_zero_stage', [
         pytest.param('cpu', None, id='cpu-ddp'),
         pytest.param('gpu', None, id='gpu-ddp', marks=pytest.mark.gpu),
-        pytest.param('gpu', 0, id='deepspeed-zero0', marks=pytest.mark.gpu),
```
hanlint: There is an issue with deepspeed here -- the tests pass, but somehow the errors overflow into other tests.

hanlint:
Thanks for the review, I addressed the comments here: https://github.com/mosaicml/composer/pull/1491/commits/492cef36febab6a657b42a66d004bde41cc1d10d

mvpatel2000 (approved):
LGTM. Left one minor comment. I'll let you track and address the deepspeed issue separately

PR #1493, opened 2022-09-01
update model token embeddings according to tokenizer len

### LINKED ISSUE #1492: HuggingFace model should update word embeddings dimension according to tokenizer
** To reproduce

Right now if I call 

```
model = transformers.AutoModelForCausalLM.from_pretrained('bigscience/bloom-560m')
tokenizer = transformers.AutoTokenizer.from_pretrained('gpt2')

model = HuggingFaceModel(model=model, tokenizer=tokenizer)
```

The bloom model will have a [250880, 1024] dim word embedding matrix and a vocab size of 250880 even though a gpt2 tokenizer with vocab size of 50257 is used.

## Expected behavior

The class HuggingFaceModel needs to update word embedding matrix if a tokenizer is supplied.

ananyahjha93:
tell me if I need to add any tests for this. Closes #1492

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from torchmetrics import Metric
@@ -14,6 +15,8 @@ from composer.utils.import_helpers import MissingConditionalImportError
 
 if TYPE_CHECKING:
     import transformers
+
+log = logging.getLogger(__name__)
 
 __all__ = ['HuggingFaceModel']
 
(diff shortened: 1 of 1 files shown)


moinnadeem (commented):
@ananyahjha93 Great contribution! 

I want to be careful when we sliently change the model without the user necessarily knowing (to help avoid user error): can you precede this with a `log.warning(f"The number of tokens in the tokenizer and the number of tokens in the model are different. Resizing the model tokenizer to from {config.vocab_size} to {len(tokenizer}."` that gets printed if `config.vocab_size != len(tokenizer)`? 

That way, if this is unexpected (or undesired), the user at least knows. Does that make sense?

Happy to give an LGTM once that is in there!

ananyahjha93:
@moinnadeem done!

moinnadeem:
@ananyahjha93 Awesome, thanks1 One last part of the requested change that was unaddressed:
> that gets printed if `config.vocab_size != len(tokenizer)`?

As it currently stands, the user would be seeing this if the `config.vocab_size==len(tokenizer)`, which could lead to confusion. We should only print the warning (and resize) when the sizes are different. Does this make sense? Let me know if the motivation isn't clear

Thanks for the PR btw, I've been meaning to do this 😅

ananyahjha93:
@moinnadeem sorry, missed that part, its a bit early in the morning here in Seattle 😝 
fixed it

moinnadeem (approved):
Hahaha, I'm in PST as well so I know the feeling 😅

Great work! Congratulations on your first commit to Composer! 😎🤩

ananyahjha93:
thanks!

PR #1495, opened 2022-09-01
BERT classifier metrics depend on num_labels

alextrott16:
This simply adds some logic to `composer.models.create_bert_classifier` to reflect that `num_labels=1` is a regression case and requires different metrics. As part of this change, regression models use MSE and SpearmanCorrCoef, whereas classification models (when `num_labels > 1`) use Accuracy, MatthewsCorrCoef, and (if appropriate) BinaryF1Score.

The docstring is updated with a note to explain this.

Closes CO-1007.


moinnadeem (approved):
LGTM! Thanks for this 🙂

mvpatel2000 (approved):
LGTM

PR #1496, opened 2022-09-01
Reset train metrics every batch

abhi-mosaic:
Change behavior for `train_metrics` to reset every batch, rather than every epoch.


alextrott16 (approved):
LGTM!

PR #1497, opened 2022-09-02
Squelch Engine debug logs

hanlint:
The engine currently emits extremely verbose debug logs. `pytest` errors invoke debug logging levels, which ends up polluting the pytest output even when engine errors were not the issue.

Instead, this PR hides the debug log behind an `ENGINE_DEBUG` environment variable, which can be set if specialized debugging of the engine is required. This is consistent with our frameworks that use envvars to control specific debugging modes.


mvpatel2000 (approved):
> This is consistent with our frameworks that use envvars to control specific debugging modes.
Can you elaborate? I don't quite understand this -- what frameworks are you referencing?

Approving to unblock since it's mostly a question for me and not a disagreement on using envvars.

PR #1504, opened 2022-09-06
Raise ValueError for Profiler + Auto Grad Accum

mvpatel2000:
As title. It's very complicated to support profiler and auto grad accum since profiler requires all "close" events, eg `AFTER_X`, run. This is hard to do and not worth it right now.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -141,8 +141,12 @@ def _set_evaluator_interval_and_subset_num_batches(
             evaluator.eval_interval = eval_interval
 
 
-def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device):
+def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device, profiler: Optional[Profiler]):
     if grad_accum == 'auto':
+        if profiler:
+            raise ValueError("`grad_accum='auto' is not compatible with the profiler. It is recommended to run "
+                             "a mini-run with `grad_accum='auto'` to identify the optimal grad_accum value and "
+                             'then manually specify that in a second run with profiler.')
         warnings.warn(("Setting `grad_accum='auto'` is an experimental feature which may cause "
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -141,8 +141,12 @@ def _set_evaluator_interval_and_subset_num_batches(
             evaluator.eval_interval = eval_interval
 
 
-def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device):
+def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device, profiler: Optional[Profiler]):
     if grad_accum == 'auto':
```
karan6181, line 145: [Optional]: Though it is not necessary as part of this PR, but just wondering to add a method comments to understand the code better. Thanks!

karan6181 (approved):
LGTM. Thank you for your contribution!

composer/trainer/trainer.py
```diff
@@ -141,8 +141,13 @@ def _set_evaluator_interval_and_subset_num_batches(
             evaluator.eval_interval = eval_interval
 
 
-def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device):
+def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device, profiler: Optional[Profiler]):
```
hanlint, line 144: seems a bit odd that this method returns whether something is grad accum, but then has this weird side effect of checking the profiler. Can we just add this into the main trainer init, and accept the duplication?

composer/trainer/trainer.py
```diff
@@ -142,6 +142,7 @@ def _set_evaluator_interval_and_subset_num_batches(
 
 
 def _is_adaptive_grad_accum(grad_accum: Union[int, str], device: Device):
+    """Checks if `grad_accum='auto'` and if its compatible with current run configurations."""
```
hanlint, line 145:
```suggestion
```

hanlint (approved):

PR #1505, opened 2022-09-06
add yahp deprecation warnings

hanlint:
This PR adds deprecation warnings whenever a user invokes:
* `run_composer_trainer.py` entrypoint
* `TrainerHparams` creation

There are many places in our codebase where we have hparams objects, but this should be sufficient to cover most use cases.


mvpatel2000 (approved):
Do we also need to put this warning in the other Hparams classes in `trainer_hparams.py`?

Approving to unblock.

hanlint:
I don't think so? All roads lead to the `TrainerHparams` anyway.

mvpatel2000:
Yep you're right. Lgtm!

PR #1506, opened 2022-09-07
Raise ValueError if missing a surgery algorithm

mvpatel2000:
- raise ValueError if missing surgery algorithm
- checks exact count of surgery algos because some (eg Factorize) might require multiple copies
- add new test for expected behavior. Confirmed test fails before change with ungraceful error (optimizer doesnt have all the params it expects)
- Note: This does not work with user-created algorithms, but I think that's fine

TODO / Looking for Feedback: 
- how do we actually know the right params to feed into a surgery algo? I don't think this is possible with the current design... should we require surgery algos store init params in `state_dict`?

Along for the ride:
- remove `test_run_name.py`, which is now replaced by the run_name test in the refactored `test_checkpoint.py` file


mvpatel2000:
This works as a v1, so im requesting reviews. 

There's some edge cases where you might have 2+ copies of an algorithm. I also am not sure how to cleanly give the init args to user, so ill talk with research.

mvpatel2000:
Requesting review because I think it's in a mergable state. Ideally, we also give the user the arguments required for surgery so its really easy to turn it on if missing, but it seems algos don't currently store enough information for this in the checkpoint (unless I'm missing something)

dskhudia:
> how do we actually know the right params to feed into a surgery algo? I don't think this is possible with the current design... should we require surgery algos store init params in state_dict?

So far we require that the list of algorithm and init args should be provided by the user when constructing trainer. However, a user who created the checkpoint may be different from the one who is using a checkpoint and have no idea what init args were used for surgery algorithms. In such cases, I think the only way is to store init args in the checkpoint and at checkpoint loading time give the error about which algorithm should be applied with what init args. Alternatively we can apply it automatically once we know the missing algorithm and init args but this may be too much of automagic under the hood.

composer/algorithms/alibi/alibi.py
```diff
@@ -164,6 +164,10 @@ def __init__(self, max_sequence_length: int, train_sequence_length_scaling: floa
         self.train_sequence_length_scaling = train_sequence_length_scaling
         self._applied = False
 
+    @staticmethod
+    def is_model_surgery() -> bool:
```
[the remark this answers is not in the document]
hanlint: `is_model_surgery` sounds more idiomatic, and should also cover graph transformation?

composer/core/algorithm.py
```diff
@@ -56,13 +56,22 @@ def find_unused_parameters(self) -> bool:
 
     @property
     def backwards_create_graph(self) -> bool:
-        """Return ``True`` to indicate that this algorithm requires a second derivative to be computed. Defaults to ``False``.
+        """Return ``True`` to indicate this algorithm requires a second derivative to be computed. Defaults to ``False``.
```
hanlint, line 59: "that" is the right grammar?

composer/core/state.py
```diff
@@ -505,9 +505,48 @@ def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
         """
         state = _ensure_backwards_compatible_checkpointing(state)
 
+        # Verify surgery algorithms from checkpoint are also enabled in current run
```
hanlint, line 508: If a comment is needed here, then this should be a helper function, e.g. something like:
```
_verify_required_algorithms_enabled
```

composer/core/state.py
```diff
@@ -505,9 +505,48 @@ def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
...
+        if 'algorithms' in state:
+            import composer.algorithms as algorithms
+
+            # Get the types and counts of existing algorithms
+            existing_algorithm_counts = {}
```
hanlint, line 513: can use `defaultdict(int)` for this, e.g.
```
algo_count = defaultdict(int)
for algo in self.algorithms:
    algo_count[type(algo)] += 1
```

style -- also suggest more compact variable names.

composer/core/state.py
```diff
@@ -505,9 +505,48 @@ def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
...
+            for algorithm in self.algorithms:
+                if algorithm not in existing_algorithm_counts:
+                    existing_algorithm_counts[type(algorithm)] = 0
+                existing_algorithm_counts[type(algorithm)] += 1
+
+            # Get the types and counts of checkpoint algorithms
+            checkpoint_algorithm_counts = {}
+            for algorithm in state['algorithms'].keys():
```
hanlint, line 521: similiar to above, can simply via:

```
checkpoint_algo_counts = defaultdict(int)

for algo in state['algorithms'].keys():
    try:
        if getattr(algorithms, algo).is_model_surgery:
            checkpoint_algo_counts[algo] += 1
     except AttributeError:
         # do warning and pass

```

composer/core/state.py
hanlint: But also.. Do we need to count?

I don't see surgery being applied twice. We can future proof this, but actually I think the bigger edge case here is if someone applied the same surgery algorithm, but with different input parameters.

hanlint (commented):
A few initial comments and suggestions!

composer/algorithms/alibi/alibi.py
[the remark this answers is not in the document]
dskhudia: In addition to `surgery` sounding more of an afterthought, my only other concern was `surgery` being too closely tied to `composer/utils/module_surgery.py` way of transforming model. Something more generic would imply if any way of changing model architecture is used, this method should return true.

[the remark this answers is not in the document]
mvpatel2000: `is_model_changing`?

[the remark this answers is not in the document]
hanlint: what about `required_on_load` ? to convey not the nature of the algorithm, but that it is required to be enabled before loading? Or is there some other unique handling for surgery that goes beyond this?

composer/core/state.py
```diff
@@ -505,9 +505,48 @@ def load_state_dict(self, state: Dict[str, Any], strict: bool = False):
...
+                        if algorithm_cls not in checkpoint_algorithm_counts:
+                            checkpoint_algorithm_counts[algorithm_cls] = 0
+                        checkpoint_algorithm_counts[algorithm_cls] += 1
+                else:
+                    logger.warning(
+                        f'Found algorithm of unknown type: {algorithm}. Skipping check for if it is required when loading checkpoint.'
+                    )
+
+            missing_surgery_algos = []
+            for algorithm in checkpoint_algorithm_counts.keys():
```
hanlint, line 536: this can also be simplified with `.items()`:

```
for algo, count in checkpoint_algo_count.items():
    if count != algo_counts.get(algo, 0):
        missing_algos.append(...)

if missing_algos:
    raise ...
```

hanlint (commented):
Took a final pass. 

For comparing surgery algorithm settings, we should implement `__repr__` for the algorithms and do that comparison.

composer/core/state.py
mvpatel2000: I'd rather future proof for now

composer/core/state.py
mvpatel2000: +1 to default dict, but want to push back on `try except`. It's a little longer code, but that's ok imo. I'd rather not rely on try catch like this just to skip an additional check

composer/algorithms/alibi/alibi.py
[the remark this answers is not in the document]
dskhudia: `required_before_restore`?

composer/core/state.py
hanlint: that's fine, its a EAFP preference to be explicit.

mvpatel2000:
Reworked to use `__repr__`. Added test to make sure all surgery algorithms have `__repr__`

tests/algorithms/test_required_on_load.py
```diff
@@ -0,0 +1,32 @@
...
+
+from typing import Type
+
+import pytest
+
+from composer import algorithms
+from composer.core import Algorithm, Time, TimeUnit  # type: ignore imports used in `eval(representation)`
+
+
+def initialize_algorithm(algo_cls: Type):
```
hanlint, line 12: We do have `algorithm_settings` that hold the minimal set of required `kwargs` if that helps

hanlint (approved):
LGTM

composer/algorithms/alibi/alibi.py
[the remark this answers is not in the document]
mvpatel2000: I like `on_load` more unless Daya you strongly object

dskhudia (approved):

PR #1508, opened 2022-09-07
Move logic from `initialize_object` to object store class

hanlint:
This PR moves some of the environment loading logic from the hparams class into the object store proper, while retaining backwards compatibility.

With this change, some additional hparams-specific tests can also be moved into the `test/hparams` folder.

This PR also adds a `dist.barrier()` to `Trainer.close()` which will wait for all ranks to close before exiting.


composer/utils/object_store/object_store_hparams.py
```diff
@@ -161,20 +161,24 @@ def get_object_store_cls(self) -> Type[ObjectStore]:
...
+            provider_kwargs['port'] = self.port
+        provider_kwargs.update(self.extra_init_kwargs)
+
+        return {
             'provider': self.provider,
             'container': self.container,
-            'provider_kwargs': {},
+            'chunk_size': self.chunk_size,
+            'key_environ': self.key_environ,
+            'secret_environ': self.secret_environ,
```
mvpatel2000, line 179: why are these moved out of provider

mvpatel2000 (approved):
LGTM but not super familiar with this

composer/utils/object_store/object_store_hparams.py
hanlint: They weren't -- originally those were used inside `initialize_object` to read from envvars if they exist and populate `provider`. That functionality is now in the object store's own `__init__`.

mvpatel2000: Right, so before they were added to the `provider` object but now they are moved outside `provider` into the parent json (if im understanding this correctly). Just making sure this is a conscious choice

hanlint: Ah, let me clarify!

The `provider` object has not been changed. Before/after it is still `{'key': ..., 'secret': ..}`.

Pseudo-code before:
```
provider = {
'key', ...,
'secret',... }

if key_environ in os.environ:
    provider['key'] = # read from envvar

# create object
return ObjectStore( provider..)

```

After:
```
provider = {
'key', ...,
'secret',... }


# create object
return ObjectStore( provider, key_envirion)

class ObjectStore

def __init__(provider, key_envirion):

    # logic now lives here
    if key_environ in os.environ:
        provider['key'] = ...

```

mvpatel2000:
Discussed offline, things are clear to me. Confident in lgtm

PR #1516, opened 2022-09-10
Use eval context

mvpatel2000:
Fixes dev regressions. These were caused by the model being in eval mode for all but the first microbatch.

Now, we set train mode only once (at start of training) and all eval happens within a context which automatically cleans up eval mode.

Regression runs verifying fix: https://wandb.ai/mosaic-ml/mihir-metrics-debug?workspace=user-mihirmosaic (pre-metrics vs. fix-medium)


hanlint (approved):
LGTM, thanks for chasing down this bug!

PR #1517, opened 2022-09-10
Rework auto grad accum checks

mvpatel2000:
As is, this would silently skip certain things (eg metrics calculation) for non-CPU / non-GPU devices. It would catch OOM and then just... return. We should be checking device type at top.

Note: I am assuming AGA doesn't work on TPU and disabling, but @hanlint  feel free to change if you think it should work / is something you've tested

Update: Also moved aga check to state so seq length warmup does this properly

### DIFF
--- a/composer/algorithms/seq_length_warmup/seq_length_warmup.py
+++ b/composer/algorithms/seq_length_warmup/seq_length_warmup.py
@@ -310,13 +310,13 @@ class SeqLengthWarmup(Algorithm):
 
             # This error/state.grad_accum handling mimics the logic in trainer._train_batch().
             except RuntimeError as e:
-                if state.adaptive_gradient_accumulation and _is_cuda_oom(e):
+                if state.auto_grad_accum and _is_cuda_oom(e):
                     log.debug((f"Rank {dist.get_global_rank()} OOM'd."))
                     found_cuda_oom = 1
                 else:
                     raise
 
-            if state.adaptive_gradient_accumulation:
+            if state.auto_grad_accum:
                 devicegpu = DeviceGPU()
                 # Propagate across all ranks if any rank hit CUDA OOM
                 found_cuda_oom = devicegpu.tensor_to_device(torch.tensor([found_cuda_oom], dtype=torch.uint8))
--- a/composer/algorithms/seq_length_warmup/seq_length_warmup.py
+++ b/composer/algorithms/seq_length_warmup/seq_length_warmup.py
@@ -3,6 +3,7 @@
 
 """Core code for sequence length warmup."""
 
+import logging
 import textwrap
 from math import ceil
 from typing import Dict, Mapping, Optional
@@ -18,6 +19,8 @@ from composer.loggers import Logger
 from composer.models import HuggingFaceModel
 from composer.trainer.devices import DeviceGPU
 from composer.utils import dist, ensure_tuple
+
+log = logging.getLogger(__name__)
 
 __all__ = ['SeqLengthWarmup', 'set_batch_sequence_length']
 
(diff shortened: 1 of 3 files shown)


composer/trainer/trainer.py
```diff
@@ -1739,16 +1739,14 @@ def _eval_train_metrics(self, device_batch):
...
-                    # Propagate across all ranks if any rank hit CUDA OOM
-                    found_cuda_oom = self._device.tensor_to_device(torch.tensor([found_cuda_oom], dtype=torch.uint8))
-                    dist.all_reduce(found_cuda_oom, reduce_operation='MAX')
-                    if found_cuda_oom.item() == 1:
-                        device_batch_size = self._train_data_spec.get_num_samples_in_batch(device_batch)
-                        _adjust_eval_batch_split(self.state, device_batch_size)
-                        # Skip return and rerun after handling oom
-                        continue
+                # Propagate across all ranks if any rank hit CUDA OOM
+                found_cuda_oom = self._device.tensor_to_device(torch.tensor([found_cuda_oom], dtype=torch.uint8))
```
hanlint, line 1743: I find this still a bit odd, that we are doing `found_cuda_oom` reductions for _any_ hardware device (even non-CUDA ones). Since `found_cuda_oom` will always be zero for non-GPU devices, why not add a guard here to skip this step?

mvpatel2000: Updated. PR now ensures check is same as guard (or it will break for same reason described in description).

mvpatel2000: NVM I see the TPU check, but I think this is still cleaner + now properly handles seq length warmup

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -306,16 +310,13 @@ def _activate_model(self, state: State, logger: Logger) -> None:
 
             # This error/state.grad_accum handling mimics the logic in trainer._train_batch().
             except RuntimeError as e:
-                if 'CUDA out of memory' in str(e):
+                if state.adaptive_gradient_accumulation and _is_cuda_oom(e):
+                    log.debug((f"Rank {dist.get_global_rank()} OOM'd."))
                     found_cuda_oom = 1
                 else:
                     raise
```
hanlint, line 317: cc: @alextrott16 on whether they want to require `grad_accum=auto` for seq length warmup oom protection.

If so, then maybe we should print something here about setting ga=auto for SLW once it OOMs.

composer/trainer/trainer.py
```diff
@@ -904,16 +902,19 @@ def __init__(
...
-                           optimizers=optimizers,
-                           run_name=run_name,
-                           deepspeed_config=deepspeed_config)
+        self.state = State(
+            rank_zero_seed=rank_zero_seed,
+            algorithms=algorithms,
+            model=model,
+            callbacks=callbacks,
+            grad_accum=grad_accum,
+            adaptive_gradient_accumulation=adaptive_gradient_accumulation,
```
hanlint, line 911: nit: very long variable name.. maybe `auto_grad_accum` since we use `grad_accum` above? and users set `grad_accum=auto`, not `grad_accum=adaptive`?

regardless, might be confusing for `state` to have both grad accum and auto_grad_accum attributes. trying to think of better naming here.. 🤔

composer/algorithms/seq_length_warmup/seq_length_warmup.py
mvpatel2000: Note that SLW works without `grad_accum=auto` (it existed before aga did) and I don't think we should require it

hanlint (approved):

composer/trainer/trainer.py
```diff
@@ -1436,8 +1437,8 @@ def fit(
 
         # Grad Accum
         if grad_accum is not None:
-            self.adaptive_gradient_accumulation = _is_adaptive_grad_accum(grad_accum, device=self._device)
-            if self.adaptive_gradient_accumulation and self.state.profiler:
+            self.state.auto_grad_accum = _is_adaptive_grad_accum(grad_accum, device=self._device)
```
dskhudia, line 1440: since you are renaming here, also rename `_is_adaptive_grad_accum` to `_is_auto_grad_accum`?

dskhudia (approved):

PR #1518, opened 2022-09-11
Fix SAM loss

mvpatel2000:
Fixes spikes in loss logs from SAM (yellow line) where every SAM pass has 2x the loss logged. Loss should not be computed from the second pass. Bug was introduced [here](https://github.com/mosaicml/composer/commit/8f0d7601eb5a29773f36a5b1aed9c9d8273f94ae) when loss became computed by passing by reference
image


hanlint (approved):
LGTM

dskhudia (approved):

PR #1523, opened 2022-09-12
Update evaluate_periodically() when eval interval is of type Duration

karan6181:
## Description
- User can now pass the `'<*>dur'` or `Time(<value>, TimeUnit.DURATION)` as `eval_interval` during training.
- For example, `eval_interval` as `0.1dur` with `max_duration` as `20ep` (20 Epochs) means that evaluation would get executed at every 2 epochs with total of 10 evaluations during the whole training process.
- There are multiple ways where user can define a combination of `eval_interval` and `max_duration`.

| No.  | eval_interval                 | max_duration | Description                              |
| ---- | ----------------------------- | ------------ | ---------------------------------------- |
| 1    | '0.1dur'                      | '50ep'       | Perform evaluation at every 5 epochs     |
| 2    | '0.05dur'                     | '1000ba'     | Perform evaluation at every 50 batches   |
| 3    | '0.2dur'                      | '6000sp'     | Perform evaluation at every 1200 samples |
| 4    | '0.1dur'                      | '10000tok'   | Perform evaluation at every 1000 tokens  |
| 5    | Time(0.25, TimeUnit.DURATION) | '40ep'       | Perform evaluation at every 10 epochs    |

## Testing
# Unit test
- Added a unit test to test the functionality for the above usecase except the 4th.
- [TODO] Add a unittest for NLP workloads (when `max_duration` is `'<val>tok'`)

# Integration Test
- Ran the MNIST example to different `max_duration` usecase such as `'<value>ep'`, `'<value>ba'`, and `'<value>sp'`
1. `max_duration: 10ep`,  `eval_interval: 0.2dur`, and `train_batch_size: 2048`
    1. Evaluation has been called 5 times as expected which is at epoch 1, 3, 5, 7, 9 (epoch number starts with 0)
    ```
    [epoch=1][batch=29/10]: metrics/eval/CrossEntropy: 0.1552
    [epoch=1][batch=29/10]: metrics/eval/Accuracy: 0.9490
    ....
    [epoch=3][batch=29/10]: metrics/eval/CrossEntropy: 0.0755
    [epoch=3][batch=29/10]: metrics/eval/Accuracy: 0.9770
    ....
    [epoch=5][batch=29/10]: metrics/eval/CrossEntropy: 0.0604
    [epoch=5][batch=29/10]: metrics/eval/Accuracy: 0.9806
    ....
    [epoch=7][batch=29/10]: metrics/eval/CrossEntropy: 0.0540
    [epoch=7][batch=29/10]: metrics/eval/Accuracy: 0.9828
    ....
    [epoch=9][batch=29/10]: metrics/eval/CrossEntropy: 0.0531
    [epoch=9][batch=29/10]: metrics/eval/Accuracy: 0.9831
    ``` 
2. `max_duration: 100ba`,  `eval_interval: 0.1dur`, and `train_batch_size: 2048`
    1.  Evaluation has been called 10 times as expected which is at batch 10, 20, 30, 40, 50, 60, 70, 80, 90, 100
    ```
    [batch=10/100]: metrics/eval/CrossEntropy: 2.1561
    [batch=10/100]: metrics/eval/Accuracy: 0.3948
    ....
    [batch=20/100]: metrics/eval/CrossEntropy: 1.0348
    [batch=20/100]: metrics/eval/Accuracy: 0.7864
    ....
    [batch=30/100]: metrics/eval/CrossEntropy: 0.5125
    [batch=30/100]: metrics/eval/Accuracy: 0.9109
    ....
    [batch=40/100]: metrics/eval/CrossEntropy: 0.2912
    [batch=40/100]: metrics/eval/Accuracy: 0.9279
    ....
    <output curtailed>
    ....
    [batch=100/100]: metrics/eval/CrossEntropy: 0.1161
    [batch=100/100]: metrics/eval/Accuracy: 0.9644
    ```
3. `max_duration: 60000sp`,  `eval_interval: 0.10dur`, and `train_batch_size: 2048`
    1. Evaluation has been called 10 times as expected which is at samples 6144, 12288, 18432, 24576, 30720, 36864, 43008, 49152, 55296, 61440.
    ```
    [sample=6144/60000]: metrics/eval/CrossEntropy: 2.3010
    [sample=6144/60000]: metrics/eval/Accuracy: 0.1135
    ....
    [sample=12288/60000]: metrics/eval/CrossEntropy: 2.2809
    [sample=12288/60000]: metrics/eval/Accuracy: 0.1451
    ....
    [sample=18432/60000]: metrics/eval/CrossEntropy: 2.2158
    [sample=18432/60000]: metrics/eval/Accuracy: 0.2579
    ....
    [sample=24576/60000]: metrics/eval/CrossEntropy: 2.0613
    [sample=24576/60000]: metrics/eval/Accuracy: 0.5095
    ....
    <output curtailed>
    ....
    [sample=61440/60000]: metrics/eval/CrossEntropy: 1.0423
    [sample=61440/60000]: metrics/eval/Accuracy: 0.8067
    ```

## Upcoming
As part of next PR
- Add a support when `eval_interval` is `dur` and max_duration is `tok`. 
- Add a unittest for the same
Tracking Jira for the same [CO-1084](https://mosaicml.atlassian.net/browse/CO-1084)

### DIFF
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -77,10 +77,31 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
                     state.max_duration.value * eval_interval.value) == 0 and event == Event.BATCH_END:
                 last_batch_seen = state.timestamp.batch
                 return True
-            elif state.max_duration.unit == TimeUnit.SAMPLE and int(state.timestamp.sample) % math.ceil(
-                    state.max_duration.value * eval_interval.value) == 0 and event == Event.BATCH_END:
-                last_batch_seen = state.timestamp.batch
-                return True
+            elif state.max_duration.unit == TimeUnit.SAMPLE and event == Event.BATCH_END:
+                # If the last global sample in a current batch is evenly divisible (zero remainder) by a evaluation interval duration
+                # during each evaluation cycle.
+                # E.g.: If batch_size = 2, max_duration.value = 50, eval_interval = 0.2
+                #       then, last sample in a batch: 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, ....... , 50
+                #       eval interval duration (theoretically): 10, 20, 30, 40, 50
+                #       eval interval duration (practically):   10, 20, 30, 40, 50
+                if int(state.timestamp.sample) % math.ceil(state.max_duration.value * eval_interval.value) == 0:
+                    last_batch_seen = state.timestamp.batch
+                    return True
+                # If the last global sample in a current batch is NOT evenly divisible (zero remainder) by a evaluation interval duration
+                # during each evaluation cycle, then it performs evaluation during first batch of next evaluation cycle.
+                # Logic: It should perform evaluation if the co-efficient of last global sample in a current batch by evaluation interval
+                #        duration does not match with last global sample in a previous batch by evaluation interval duration.
+                # E.g.: If batch_size = 2, max_duration.value = 50, eval_interval = 0.1
+                #       then, last sample in a batch: 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, ....... , 50
+                #       eval interval duration (theoretically): 5, 10, 15, 20, 25, 30, 35, 40, 45, 50
+                #       eval interval duration (practically):   6, 10, 16, 20, 26, 30, 36, 40, 46, 50
+                elif int(state.timestamp.batch) != 0:
+                    samples_in_a_batch = int(state.timestamp.sample) // int(state.timestamp.batch)
+                    if int(state.timestamp.sample) // math.ceil(state.max_duration.value * eval_interval) != int(
+                            state.timestamp.sample - samples_in_a_batch) // math.ceil(
+                                state.max_duration.value * eval_interval):
+                        last_batch_seen = state.timestamp.batch
+                        return True
             elif state.max_duration.unit == TimeUnit.TOKEN and int(state.timestamp.token) % math.ceil(
                     state.max_duration.value * eval_interval.value) == 0 and event == Event.BATCH_END:
                 last_batch_seen = state.timestamp.batch
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -85,7 +85,7 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
                         last_batch_seen = state.timestamp.batch
                         return True
             elif state.max_duration.unit == TimeUnit.TOKEN and event == Event.BATCH_END:
-                raise ValueError(f'Evaluation interval of type `dur` is not support yet for max_duration as `tok`')
+                raise ValueError(f'Evaluation interval of type `dur` is not supported yet for max_duration as `tok`')
         return False
 
     return should_eval
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -75,23 +75,8 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
                     state.max_duration.value * eval_interval.value) == 0 and event == Event.BATCH_END:
                 last_batch_seen = state.timestamp.batch
                 return True
-            elif (state.max_duration.unit == TimeUnit.SAMPLE or
-                  state.max_duration.unit == TimeUnit.TOKEN) and event == Event.BATCH_END:
-                # Scenarios:
-                # (1) If the last global sample in a current batch is evenly divisible (zero remainder) by a evaluation interval duration
-                # during each evaluation cycle.
-                # E.g.: If batch_size = 2, max_duration.value = 50, eval_interval = 0.2
-                #       then, last sample in a batch: 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22,......., 50
-                #       eval interval duration (theoretically): 10, 20, 30, 40, 50
-                #       eval interval duration (practically):   10, 20, 30, 40, 50
-                # (2) If the last global sample in a current batch is NOT evenly divisible (zero remainder) by a evaluation interval duration
-                # during each evaluation cycle, then it performs evaluation during first batch of next evaluation cycle.
-                # E.g.: If batch_size = 2, max_duration.value = 50, eval_interval = 0.1
-                #       then, last sample in a batch: 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22,......., 50
-                #       eval interval duration (theoretically): 5, 10, 15, 20, 25, 30, 35, 40, 45, 50
-                #       eval interval duration (practically):   6, 10, 16, 20, 26, 30, 36, 40, 46, 50
-                # Logic: It should perform evaluation if the co-efficient of last global sample in a current batch by evaluation interval
-                #        duration does not match with last global sample in a previous batch by evaluation interval duration.
+            elif state.max_duration.unit == TimeUnit.SAMPLE and event == Event.BATCH_END:
+                # If last sample in batch is not evenly divisible by eval_interval, perform evaluation in next batch
                 if int(state.timestamp.batch) > 0:
                     samples_in_a_batch = int(state.timestamp.sample) // int(state.timestamp.batch)
                     if int(state.timestamp.sample) // math.ceil(state.max_duration.value * eval_interval) != int(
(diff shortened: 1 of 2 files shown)


composer/core/evaluator.py
```diff
@@ -35,13 +36,14 @@ def evaluate_periodically(eval_interval: Union[str, Time, int], eval_at_fit_end:
...
-    if eval_interval.unit not in (TimeUnit.EPOCH, TimeUnit.BATCH):
-        raise ValueError('The `eval_interval` must have units of EPOCH or BATCH, or be a function.')
+    if eval_interval.unit not in (TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.DURATION):
+        raise ValueError('The `eval_interval` must have units of EPOCH, BATCH, DURATION or be a function.')
 
     last_batch_seen = -1
 
     def should_eval(state: State, event: Event):
-        if int(eval_interval) <= 0:
+        # `TimeUnit.Duration` value is a floating value and ranges from `[0.0, 1.0)`
```
mvpatel2000, line 45:
```suggestion
        # `TimeUnit.Duration` value is a float from `[0.0, 1.0)`
```

composer/core/evaluator.py
```diff
@@ -59,6 +61,46 @@ def should_eval(state: State, event: Event):
             last_batch_seen = state.timestamp.batch
             return True
 
+        if eval_interval.unit == TimeUnit.DURATION:
+            if state.max_duration is None:
+                raise ValueError(f'`max_duration` is a required argument and must be specified when constructing the '
+                                 f'`Trainer()` or `duration` when calling `Trainer.fit()`')
```
mvpatel2000, line 67: Since we check this in `Trainer`, it's fine to skip this and not have to raise a ValueError. If it's because of pyright typing, `assert state.max_duration != None` is fine.

mvpatel2000 (commented):
This looks like the right approach to me, but I think we can make this a little more general while we're already refactoring it. It looks like you're basically casing duration into the 4 other types and then (approximately) converting into those units. 

Can we instead rewrite this logic to say if the time unit is a duration, convert it into the type of max duration at the top of the function, and then follow the codepath for the respective type we've converted it to? This gives us the added benefit of being able to use the same logic you've added for tokens and samples to allow users to submit eval_interval in those units as well.

composer/core/evaluator.py
karan6181: Done in https://github.com/mosaicml/composer/pull/1523/commits/7fb861f1fd1c6b4461132c9092ec826f48a00071.

composer/core/evaluator.py
karan6181: Done in https://github.com/mosaicml/composer/pull/1523/commits/7fb861f1fd1c6b4461132c9092ec826f48a00071.

karan6181:
> This looks like the right approach to me, but I think we can make this a little more general while we're already refactoring it. It looks like you're basically casing duration into the 4 other types and then (approximately) converting into those units.
> 
> Can we instead rewrite this logic to say if the time unit is a duration, convert it into the type of max duration at the top of the function, and then follow the codepath for the respective type we've converted it to? This gives us the added benefit of being able to use the same logic you've added for tokens and samples to allow users to submit eval_interval in those units as well.

If the `eval_interval.unit` is of type `TimeUnit.SAMPLE` or `TimeUnit.TOKEN`, then the calculation of evaluation interval is different than the current implementation of `eval_interval.unit` as `TimeUnit.DURATION` and `max_duration.unit` as `TimeUnit.SAMPLE`. 

For example, `batch_size = 2`, `max_duration = 5000sp`, and `eval_interval = 100sp`
- then we perform evaluation at every 100 samples.
- However, the calculation is different than the current approach in this PR where we don't have to do `math.ceil(state.max_duration.value * eval_interval)` instead we can just use `eval_interval`

Looks we have to handle all the scenarios case by case basis. 
Please let me know if I didn't understand your point clearly.

mvpatel2000 (approved):
Ideally, we'd just convert `eval_interval` from duration unit to whatever unit type `max_duration` at top of function for simplicity. More broadly, we want a util function set that lets us easily convert between different units. I don't think its necessary for getting this in before the next release though.

After this goes in, can you please make a new JIRA and track it for adding NLP support with tokens and unit tests for that scenario?

composer/core/evaluator.py
```diff
@@ -59,6 +61,44 @@ def should_eval(state: State, event: Event):
...
+                #       eval interval duration (theoretically): 10, 20, 30, 40, 50
+                #       eval interval duration (practically):   10, 20, 30, 40, 50
+                # (2) If the last global sample in a current batch is NOT evenly divisible (zero remainder) by a evaluation interval duration
+                # during each evaluation cycle, then it performs evaluation during first batch of next evaluation cycle.
+                # E.g.: If batch_size = 2, max_duration.value = 50, eval_interval = 0.1
+                #       then, last sample in a batch: 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22,......., 50
+                #       eval interval duration (theoretically): 5, 10, 15, 20, 25, 30, 35, 40, 45, 50
+                #       eval interval duration (practically):   6, 10, 16, 20, 26, 30, 36, 40, 46, 50
+                # Logic: It should perform evaluation if the co-efficient of last global sample in a current batch by evaluation interval
+                #        duration does not match with last global sample in a previous batch by evaluation interval duration.
```
mvpatel2000, line 94:
```suggestion
                # If last sample in batch is not evenly divisible by eval_interval, perform evaluation in next batch
```
We generally want to avoid long comments in the code. While this is very valuable for me to understand it in a PR review, I don't think we need to give examples and detailed explanations. The code speaks for itself, and our style guide dictates minimal comments. Lets just leave a small comment for the edge case in scenario 2

composer/core/evaluator.py
```diff
@@ -59,6 +61,31 @@ def should_eval(state: State, event: Event):
...
+                # If last sample in batch is not evenly divisible by eval_interval, perform evaluation in next batch
+                if int(state.timestamp.batch) > 0:
+                    samples_in_a_batch = int(state.timestamp.sample) // int(state.timestamp.batch)
+                    if int(state.timestamp.sample) // math.ceil(state.max_duration.value * eval_interval) != int(
+                            state.timestamp.sample - samples_in_a_batch) // math.ceil(
+                                state.max_duration.value * eval_interval):
+                        last_batch_seen = state.timestamp.batch
+                        return True
+            elif state.max_duration.unit == TimeUnit.TOKEN and event == Event.BATCH_END:
+                raise ValueError(f'Evaluation interval of type `dur` is not support yet for max_duration as `tok`')
```
mvpatel2000, line 88:
```suggestion
                raise ValueError(f'Evaluation interval of type `dur` is not supported yet for max_duration as `tok`')
```

composer/core/evaluator.py
```diff
@@ -59,6 +61,44 @@ def should_eval(state: State, event: Event):
...
+                    state.timestamp.batch) % math.ceil(state.max_duration.value * float(eval_interval) *
+                                                       state.dataloader_len) == 0 and event == Event.BATCH_END:
+                last_batch_seen = state.timestamp.batch
+                return True
+            elif state.max_duration.unit == TimeUnit.BATCH and int(state.timestamp.batch) % math.ceil(
+                    state.max_duration.value * eval_interval.value) == 0 and event == Event.BATCH_END:
+                last_batch_seen = state.timestamp.batch
+                return True
+            elif (state.max_duration.unit == TimeUnit.SAMPLE or
+                  state.max_duration.unit == TimeUnit.TOKEN) and event == Event.BATCH_END:
```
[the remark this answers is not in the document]
karan6181: Removed as part https://github.com/mosaicml/composer/pull/1523/commits/24a6f7ea7f073bc06805508ce7f9d2205fd97d46 and added a ValueError for the same. Thanks!

composer/core/evaluator.py
karan6181: Cool. Done in https://github.com/mosaicml/composer/pull/1523/commits/24a6f7ea7f073bc06805508ce7f9d2205fd97d46.

karan6181:
> Ideally, we'd just convert `eval_interval` from duration unit to whatever unit type `max_duration` at top of function for simplicity. More broadly, we want a util function set that lets us easily convert between different units. I don't think its necessary for getting this in before the next release though.
> 
> After this goes in, can you please make a new JIRA and track it for adding NLP support with tokens and unit tests for that scenario?

Sounds good. Thanks for the explanation!

composer/core/evaluator.py
karan6181: Thanks! Done in https://github.com/mosaicml/composer/pull/1523/commits/55badd6111d6c79b705205444aa0d898df42b554.

PR #1525, opened 2022-09-13
Add Low Precision LayerNorm

nik-mosaic:
Fused LayerNorm's speedup over PyTorch's native torch.nn.LayerNorm comes from running in low precision mode. (Torch's Layernorm is already fused under the hood, so it is just as fast as Apex, but it does not run in FP16/BF16 while autocast is on.) We should switch to using native torch.nn.LayerNorm, cast to low precision, whenever possible, to reduce our dependency on Apex.

This algorithm is a low precision version of `torch.nn.LayerNorm`. It is just a thin wrapper that disables autocast and forces the layer into FP16 (if running in AMP mode) or BF16 (if running in amp_bf16 mode). Equivalent converge performance has been verified on standard NLP models.

However, because torch.nn.LayerNorm is only [fast for bf16 in torch versions>=1.13](https://github.com/pytorch/pytorch/commit/ef782e730dff7bef078fc95d7fb5b78bafcc5284), Apex's Fused LayerNorm is much faster for bf16 workflows right now. I have added FusedLayerNorm as a fallback inside LowPrecisionLayerNorm in this case, so the separate FusedLayerNorm algorithm can be removed shortly. We can remove the fallback later when Composer no longer supports torch<1.13.

FusedLayerNorm will be removed shortly in follow-up PR.


mvpatel2000:
1. Do you know why AMP doesn't downcast in layernorm? 
2. Instead of adding a new algorithm, can we just update FusedLayerNorm to triage to the right version (torch nn except if bf16 on old torch) under the hood? Seems much less confusing than adding another algorithm?

abhi-mosaic:
> 1. Do you know why AMP doesn't downcast in layernorm?
> 2. Instead of adding a new algorithm, can we just update FusedLayerNorm to triage to the right version (torch nn except if bf16 on old torch) under the hood? Seems much less confusing than adding another algorithm?

1.) I believe PyTorch AMP does not downcast layernorm because in their internal experiments (which we need to re-verify at large scale), they probably found that it was not numerically stable / led to worse convergence. So they hardcoded the ['autocast-to-fp32' ](https://pytorch.org/docs/stable/amp.html#cuda-ops-that-can-autocast-to-float32) inside PyTorch AMP.

2.) I am ok with renaming `FusedLayerNorm` -> `LowPrecisionLayerNorm`, but maybe that is equally accomplished by adding this algorithm and deleting the former in a followup PR. Leaving it to @mvpatel2000 to decide :) The **most important thing to me** is that we drop APEX from Composer -- it's a very large dependency and not really recommended anymore, and this seems to be the only place that we use it.

Also re. the switch statement on torch.bfloat16, I think that's totally fine... so basically if using (bfloat 16, torch<1.13) we should emit a warning saying "hey, this is a NO-OP, because if we switched to low precision bfloat16 layernorm it would be slower"

mvpatel2000:
> > 1. Do you know why AMP doesn't downcast in layernorm?
>
> 1.) I believe PyTorch AMP does not downcast layernorm because in their internal experiments (which we need to test at large scale), they probably found that it was not numerically stable / led to worse convergence. So they hardcoded the ['autocast-to-fp32' ](https://pytorch.org/docs/stable/amp.html#cuda-ops-that-can-autocast-to-float32) inside PyTorch AMP.

Seems a little sus that we do it if they saw issues... but *shrug* ok if research is fine with it fine with me I guess.

> > 2. Instead of adding a new algorithm, can we just update FusedLayerNorm to triage to the right version (torch nn except if bf16 on old torch) under the hood? Seems much less confusing than adding another algorithm?
> 
> 2.) I am ok with renaming `FusedLayerNorm` -> `LowPrecisionLayerNorm`, but maybe that is equally accomplished by adding this algorithm and deleting the former in a followup PR. Leaving it to @mvpatel2000 to decide :) The **most important thing to me** is that we drop APEX from Composer -- it's a very large dependency and not really recommended anymore, and this seems to be the only place that we use it.
> 
> Also re. the switch statement on torch.bfloat16, I think that's totally fine... so basically if using (bfloat 16, torch<1.13) we should emit a warning saying "hey, this is a NO-OP, because if we switched to low precision bfloat16 layernorm it would be slower". If anyone encounters this message, it will be an incentive to upgrade to torch 1.13 😉 )

I'm fine with renaming and removing support for bf16 until torch1.13. IIRC that path is not used right now, and if release is imminent, fine by me.

Defer to @nik-mosaic since its his PR

composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
```diff
@@ -0,0 +1,99 @@
...
+        apply_at (Event, optional): Event where algorithm is applied.
+    """
+
+    def __init__(self, apply_at: Optional[Event] = None):
+        self.apply_at = Event.INIT if apply_at is None else apply_at
+        if self.apply_at not in {Event.INIT, Event.AFTER_LOAD}:
+            raise ValueError('LowPrecisionLayerNorm only supports application on Event.INIT and Event.AFTER_LOAD.')
+
+    @property
+    def find_unused_parameters(self) -> bool:
```
dskhudia, line 90: Remove this

tests/algorithms/test_low_precision_layernorm
```diff
@@ -0,0 +1,60 @@
+# Copyright 2022 MosaicML Composer authors
```
dskhudia, line 1: The filename doesn't end in .py?

dskhudia (approved):

dskhudia:
IMO It's ok to leave bf16 as low performing if the PyTorch version is not right. There are many operators in PyTorch that don't perform well with bf16.

dskhudia:
Are you planning to remove FusedLayerNorm in followup PR? Another way is aliasing FusedLayerNorm with LPLayerNorm to make sure we don't break existing code.

composer/algorithms/low_precision_layernorm/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "low_precision_layernorm": {
+        "name": "Low Precision LayerNorm",
+        "class_name": "LowPrecisionLayerNorm",
+        "functional": "apply_low_precision_layernorm",
+        "tldr": "Forces LayerNorm to fp16 or bf16",
+        "attribution": "",
+        "link": "",
+        "domains": [
+            "nlp"
```
alextrott16, line 10: I'm not sure I'd call LayerNorm an NLP op. I think it's more general.

alextrott16 (approved):

composer/algorithms/low_precision_layernorm/metadata.json
nik-mosaic: Added CV tag as well.

nik-mosaic:
For backwards compatibility reasons (since many on the NLP team use bf16) I have updated LowPrecisionLayerNorm to call FusedLayerNorm internally if in bf16 mode and torch<1.13, so users don't lose speed. When Composer stops supporting torch<1.13 (or when we have a stable release post-1.13), I can remove this fallback.

> Are you planning to remove FusedLayerNorm in followup PR? Another way is aliasing FusedLayerNorm with LPLayerNorm to make sure we don't break existing code.

I will remove the FusedLayerNorm algorithm in a follow-up PR shortly after this one is merged.

mvpatel2000 (approved):

PR #1526, opened 2022-09-13
Fix compression file remote download exception handling.

knighton:
This fixes issue CO-1071

The problem is that we were expecting the wrong exception (for when the dataset is on localdisk), but actually each cloud backend will throw its own exception for the missing file.


karan6181 (approved):
LGTM. Thanks!

bandish-shah (changes requested):
Please add some more details on what this PR does in the description.  Also you can note the Jira issue number and it will link in the issue automatically.

bandish-shah (approved):
LGTM

PR #1527, opened 2022-09-14
Checkpointing Interface Refactor Part I: Renaming Arguments

eracah:
1.  Rename arguments
      1. `Trainer.save_folder` -> `Trainer.checkpoint_save_path`
      2. `Trainer.save_interval`  -> `Trainer.checkpoint_save_interval`
      3. `CheckpointSaver.folder` -> `CheckpointSaver.checkpoint_save_path`
      4. `Trainer.save_num_checkpoints_to_keep` -> `Trainer.num_checkpoints_to_keep`
      5. `CheckpointSaver.filename` -> `CheckpointSaver.checkpoint_filename`
      6. `CheckpointSaver.latest_filename` -> `CheckpointSaver.latest_checkpoint_filename`

2. Along for the ride
    *  Add skip tests for uninstalled modules (wandb and libcloud)

[PR Part 2](https://github.com/eracah/evan-composer/pull/1)

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -109,7 +109,7 @@ class CheckpointSaver(Callback):  # noqa: D101
         ...     CheckpointSaver(
         ...         checkpoint_save_path='{{run_name}}/checkpoints',
         ...         checkpoint_filename="ep{{epoch}}-ba{{batch}}-rank{{rank}}",
-        ...         latest_filename="latest-rank{{rank}}",
+        ...         latest_checkpoint_filename="latest-rank{{rank}}",
         ...         checkpoint_save_interval="1ep",
         ...         weights_only=False,
         ...     )
@@ -188,10 +188,10 @@ class CheckpointSaver(Callback):  # noqa: D101
             Leading slashes (``'/'``) will be stripped.
 
             To disable logging trace files as file artifacts, set this parameter to ``None``.
-        latest_filename (str, optional): A format string for a symlink which points to the last saved checkpoint.
+        latest_checkpoint_filename (str, optional): A format string for a symlink which points to the last saved checkpoint.
             Default: ``'latest-rank{{rank}}.pt'``.
 
-            Symlinks will be created approximately at ``{{checkpoint_save_path}}/{{latest_filename.format(...)}}``.
+            Symlinks will be created approximately at ``{{checkpoint_save_path}}/{{latest_checkpoint_filename.format(...)}}``.
 
             The same format variables as for ``name`` are available.
 
@@ -202,7 +202,7 @@ class CheckpointSaver(Callback):  # noqa: D101
             *   The :attr:`~.State.run_name` is 'awesome-training-run'
             *   The default ``checkpoint_save_path='{{run_name}}/checkpoints'`` is used.
             *   The default ``name='ep{{epoch}}-ba{{batch}}-rank{{rank}}'`` is used.
-            *   The default ``latest_filename='latest-rank{{rank}}'`` is used.
+            *   The default ``latest_checkpoint_filename='latest-rank{{rank}}'`` is used.
             *   The current epoch count is ``1``.
             *   The current batch count is ``42``.
 
@@ -228,7 +228,7 @@ class CheckpointSaver(Callback):  # noqa: D101
             Default: ``'{{run_name}}/checkpoints/latest-rank{{rank}}"``.
 
             Whenever a new checkpoint is saved, a symlink artifact is created or updated to point to the latest checkpoint's ``artifact_name``.
-            The artifact name will be determined by this format string. This parameter has no effect if ``latest_filename`` or ``artifact_name`` is ``None``.
+            The artifact name will be determined by this format string. This parameter has no effect if ``latest_checkpoint_filename`` or ``artifact_name`` is ``None``.
 
             .. seealso:: :doc:`Artifact Logging</trainer/artifact_logging>` for notes for file artifact logging.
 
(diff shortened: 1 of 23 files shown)


composer/callbacks/checkpoint_saver.py
```diff
@@ -289,25 +289,26 @@ class CheckpointSaver(Callback):  # noqa: D101
...
+            checkpoint_save_interval = checkpoint_periodically(checkpoint_save_interval)
+        self.checkpoint_save_interval = checkpoint_save_interval
 
-        self.folder = folder
+        self.checkpoint_save_path = checkpoint_save_path
 
-        self.filename = PartialFilePath(filename.lstrip('/'), folder)
-        self.latest_filename = PartialFilePath(latest_filename.lstrip('/'), folder) if latest_filename else None
+        self.checkpoint_filename = PartialFilePath(checkpoint_filename.lstrip('/'), checkpoint_save_path)
+        self.latest_checkpoint_filename = PartialFilePath(latest_checkpoint_filename.lstrip('/'),
```
mvpatel2000, line 310: minor nit: can we rename this to `checkpoint_path` and `latest_checkpoint_path`? Otherwise, the `self.checkpoint_filename.filename` is a bit weird

mvpatel2000 (approved):
LGTM

hanlint (changes requested):
Do we plan on merging all of the checkpoint refactor before the release? If not, then we may want to hold this for after the release, otherwise the signature change would be odd -- many of the other arguments that we intend to remove would still be there with a different naming convention (e.g. `save_`)

Also:
* should this be `checkpoint_save_folder` instead of `checkpoint_save_path`? to clarify that its a folder?
* should all checkpoint args be prefaced with checkpoint for clarity?

composer/callbacks/checkpoint_saver.py
```diff
@@ -289,25 +289,26 @@ class CheckpointSaver(Callback):  # noqa: D101
 
     def __init__(
         self,
-        folder: str = '{run_name}/checkpoints',
-        filename: str = 'ep{epoch}-ba{batch}-rank{rank}.pt',
+        checkpoint_save_path: str = '{run_name}/checkpoints',
```
hanlint, line 292: These are already inside of `CheckpointSaver`, so prefer to keep the names as before. Seems redundant to preface everything with `checkpoint` in the context of this object.

eracah:
> Do we plan on merging all of the checkpoint refactor before the release? If not, then we may want to hold this for after the release, otherwise the signature change would be odd -- many of the other arguments that we intend to remove would still be there with a different naming convention (e.g. `save_`)
> 
Yes we do plan to do it all. I am just splitting it up into a few smaller PR's. 


> Also:
> 
> * should this be `checkpoint_save_folder` instead of `checkpoint_save_path`? to clarify that its a folder?

I think it's fine as path.

> * should all checkpoint args be prefaced with checkpoint for clarity?\

yes, this prefixing was part of the design doc that was approved by researchers and engineers.

composer/callbacks/checkpoint_saver.py
eracah: I like the prefixing and everyone seemed to think it was fine who looked at the design doc

mvpatel2000:
> Do we plan on merging all of the checkpoint refactor before the release? If not, then we may want to hold this for after the release, otherwise the signature change would be odd -- many of the other arguments that we intend to remove would still be there with a different naming convention (e.g. `save_`)

I think it's fine to merge this in. We've talked a bit about being ok with dev having some changes that aren't in a release ready state to speedup dev velocity, and I think this PR is the right size / step in that direction. Otherwise, the entire refactor is one giant PR, and that's a pain to review. We're aiming to have everything in before the next release, so imo we should go ahead with this and continue bite sized.

> should all checkpoint args be prefaced with checkpoint for clarity?

I think everything not prefixed in this PR will be deleted in the next PR

composer/callbacks/checkpoint_saver.py
mvpatel2000: Personally, I'd rather have consistent names here with trainer. This isn't purely internal since users can make custom checkpoint savers, so it's useful to have the same names imo. I believe this was the intention in the original design doc

hanlint: Fundamentally the issue here is we need to avoid verbose argument names as much as possible, e.g. see the names in [PTL](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.ModelCheckpoint.html), HF (called it output_dir), and [Keras](https://keras.io/api/callbacks/model_checkpoint/). So our naming is an anti-pattern to the rest of the community.

hanlint (approved):
Unblocking this PR given the move to merge everything before the release.

For naming, I know this was discussed already, but consider removing `checkpoint_` from all our arguments, as the brevity is closer to the pattern of most frameworks out there (e.g. PTL, HF, Keras, etc. -- see comment for examples).

However, also OK with proceeding as already designed, and revisiting later. Leave that up to you @eracah .

eracah:
Okay, so @hanlint and @mvpatel2000 raised some good points on renaming some things. For sake of velocity, I will keep things as they are, but I will add all these name change proposals to the plan for a future find-replace PR that we will do after the rest of these checkpoint PR's are in.

eracah:
> Unblocking this PR given the move to merge everything before the release.
> 
> For naming, I know this was discussed already, but consider removing `checkpoint_` from all our arguments, as the brevity is closer to the pattern of most frameworks out there (e.g. PTL, HF, Keras, etc. -- see comment for examples).
> 
> However, also OK with proceeding as already designed, and revisiting later. Leave that up to you @eracah .

Thanks, Hanlin. Will try to get some of those name changes in at the end in the last PR of this series

composer/callbacks/checkpoint_saver.py
eracah: Yeah that's true. Thanks for raising this! Everyone else has more brevity and ours is an anti-pattern. I rather have extreme clarity even if the arguments are a bit redundant and clunky and insult the intelligence of some of the users, but I see why the other libraries went for brevity. As I mentioned below, let's consider removing the prefixes at the end of this PR series, but I'll keep as is for now.

composer/callbacks/checkpoint_saver.py
eracah: yeah this does get confusing. it is even more confusing in part 2 of the PR series. will consider it for the last PR in the series

PR #1507, opened 2022-09-07
Add v2 datasets behind a version flag

knighton:

PR #1528, opened 2022-09-15
Add ADE20K and COCO v2 dataset behind a version flag

karan6181:
## Description 
- Add a ADE20K and COCO v2 dataset behind a version flag
- Old PR with a similar change for reference https://github.com/mosaicml/composer/pull/1507

### DIFF
--- a/composer/datasets/ade20k_hparams.py
+++ b/composer/datasets/ade20k_hparams.py
@@ -24,6 +24,7 @@ from composer.datasets.synthetic import SyntheticBatchPairDataset
 from composer.datasets.synthetic_hparams import SyntheticHparamsMixin
 from composer.datasets.utils import NormalizationFn, pil_image_collate
 from composer.utils import dist
+from composer.utils.import_helpers import MissingConditionalImportError
 
 __all__ = ['ADE20kDatasetHparams', 'StreamingADE20kHparams']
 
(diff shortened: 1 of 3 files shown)


composer/datasets/ade20k_hparams.py
```diff
@@ -176,15 +179,61 @@ class StreamingADE20kHparams(DatasetHparams):
...
+                                      batch_size=batch_size)
+        elif self.version == 2:
+            try:
+                from streaming.vision import ADE20K
+            except ImportError as e:
+                raise MissingConditionalImportError(extra_deps_group='streaming',
+                                                    conda_package='mosaicml-streaming') from e
+            # Define data transformations based on data split
+            if self.split == 'train':
+                both_transforms = torch.nn.Sequential(
```
knighton, line 200: should this be `torchvision.transforms.Compose`

composer/datasets/ade20k_hparams.py
```diff
@@ -176,15 +179,61 @@ class StreamingADE20kHparams(DatasetHparams):
...
+                        class_max_percent=0.75,
+                        num_retry=10,
+                    ),
+                    RandomHFlipPair(),
+                )
+
+                # Photometric distoration values come from mmsegmentation:
+                # https://github.com/open-mmlab/mmsegmentation/blob/aa50358c71fe9c4cccdd2abe42433bdf702e757b/mmseg/datasets/pipelines/transforms.py#L861
+                r_mean, g_mean, b_mean = IMAGENET_CHANNEL_MEAN
+                image_transforms = torch.nn.Sequential(
```
knighton, line 215: should this be `torchvision.transforms.Compose`

knighton (approved):

composer/datasets/ade20k_hparams.py
karan6181: Generally we don’t use `torch.nn.Sequential()` for non-scriptable transformation and `transforms.ToTensor()` is not a scriptable transformation and we don’t have that in ADE20K. So, it's fine to use  `torch.nn.Sequential()` here.  Also, the code structure is similar to existing [ade20k.py](https://github.com/mosaicml/composer/blob/dev/composer/datasets/ade20k.py#L339).

composer/datasets/ade20k_hparams.py
karan6181: Generally we don’t use `torch.nn.Sequential()` for non-scriptable transformation and `transforms.ToTensor()` is not a scriptable transformation and we don’t have that in ADE20K. So, it's fine to use  `torch.nn.Sequential()` here.  Also, the code structure is similar to existing [ade20k.py](https://github.com/mosaicml/composer/blob/dev/composer/datasets/ade20k.py#L339).

PR #1530, opened 2022-09-15
Checkpointing Interface Refactor Part II: Removing Some Checkpoint Saving Trainer Args

eracah:
1. remove from `Trainer`:
  * save_filename
  * save_artifact_name
  * save_latest_filename
  * save_latest_artifact_name
  * save_overwrite
  * checkpoint_save_interval
  * save_weights_only 

2. Change some names in `CheckpointSaver` to be clearer (but clunkier! @hanlint)
* `weights_only` -> `save_weights_only`
*  `overwrite` -> `overwrite_checkpoints`
We can change these to be more concise at the end (I added note to [PR Plan](https://www.notion.so/Checkpointing-Interface-Refactor-f6286d3d25c04911815ad440c2c6228f#4ddf0a91020c48b1b5d3904814ab16f4) )

3. Adjust the following files to be compatible with the change:
    * `test_checkpoint.py`
    * `checkpoint_with_wandb.py`
    * `run_glue_trainer.py`
    * `test_trainer.py`
    * `trainer_hparams.py`

Along for the ride:
* add a `make test-all` option to makefile to make it easier to run all  tests including gpu, dist, etc.

Todo:
- [x] Fix `resumption.rst` doctests
- [ ] Optional: Put `checkpoint_filename` back in `Trainer`
- [x] Fix pyright issues in `run_glue_trainer.py`

[PR Part 3](https://github.com/eracah/evan-composer/pull/2)
[PR Part 4](https://github.com/eracah/evan-composer/pull/3)


eracah:
@hanlint, @mvpatel2000, One issue I am running into was removing `checkpoint_filename` from `Trainer` seemed like a good way to simplify things for users because the defaults seemed good, but looking a lot of docs, examples, tests, etc. it seems like more of a beginner argument than an intermediate one. Thoughts on putting it back in `Trainer`?

composer/callbacks/checkpoint_saver.py
```diff
@@ -296,9 +296,9 @@ def __init__(
         latest_artifact_name: Optional[str] = '{run_name}/checkpoints/latest-rank{rank}',
         checkpoint_save_interval: Union[Time, str, int, Callable[[State, Event], bool]] = '1ep',
         *,
-        overwrite: bool = False,
+        overwrite_checkpoints: bool = False,
         num_checkpoints_to_keep: int = -1,
-        weights_only: bool = False,
+        save_model_weights_only: bool = False,
```
mvpatel2000, line 301: Add to list of arguments we want to discuss length of variable name on? 

@hanlint I think the right solution here is to revisit our style guide next design meeting and ensure we all are consistent and agree on the vision.

composer/trainer/trainer.py
```diff
@@ -945,21 +908,23 @@ def __init__(
...
-                overwrite=save_overwrite,
-                weights_only=save_weights_only,
                 checkpoint_save_interval=checkpoint_save_interval,
                 num_checkpoints_to_keep=num_checkpoints_to_keep,
             )
             self.state.callbacks.append(self._checkpoint_saver)
+        # If the user specified a checkpoint callback, then use that as the _checkpoint_saver callback.
+        # TODO(eracah): deal with multiple CheckpointSaver callbacks.
+        elif checkpoint_callbacks:
+            self._checkpoint_saver = checkpoint_callbacks[0]
```
mvpatel2000, line 927: For now, let's assume autoresume is happening with the default CheckpointSaver created from Trainer args, and then you can file a JIRA and assign it to me about fixing autoresume. I don't want to overcomplicate this PR for now, and autoresume isn't something we really support with cloud rn.

Basically, this fixes a scenario we don't address right now, and I'm fine leaving it unaddressed and fixing in a separate PR. Also, I can handle it because I imagine we'll rework parts of autoresume once we have the cloud support anyways

mvpatel2000 (commented):
I'm fine keeping only one argument. We can revisit once more after based on feedback on new format, but I think lets stick with what we've designed so far given thats what people seem to like the most. Still, defer to you

composer/callbacks/checkpoint_saver.py
```diff
@@ -256,7 +256,7 @@ class CheckpointSaver(Callback):  # noqa: D101
             progress). It should return ``True`` if a checkpoint should be saved given the current state and
             event.
 
-        weights_only (bool): If ``True``, save only the model weights instead of the entire training state.
+        save_model_weights_only (bool): If ``True``, save only the model weights instead of the entire training state.
```
hanlint, line 259: I know the suggestion was to defer these to later, but would it be easier to just revert these here?

`save_weights_only` or `weights_only` is used in [Keras]((https://keras.io/api/callbacks/model_checkpoint/), [PTL](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.ModelCheckpoint.html), etc. Any deviation from that standard should be the exception, not the default.

composer/trainer/trainer.py
```diff
@@ -654,10 +623,10 @@ class Trainer:
             artifact stores.
         autoresume (bool, optional): Whether or not to enable autoresume, which allows for stopping and resuming
             training. This allows use of spot instances, as the training run is now fault tolerant.  This parameter
-            requires ``checkpoint_save_path`` and ``run_name`` to be specified and ``save_overwrite`` to be ``False``.
+            requires ``checkpoint_save_path`` and ``run_name`` to be specified.
             (default: ``False``)
 
-            When enabled, the checkpoint_save_path is checked for checkpoints of the format ``"{checkpoint_save_path}/{save_latest_filename}"``,
+            When enabled, the checkpoint_save_path is checked for checkpoints of the format ``"{checkpoint_save_path}/{latest_checkpoint_filename}"``,
```
hanlint, line 629: nit: `latest_checkpoint_filename` doesn't look like an argument to the trainer, so this seems confusing to the user as part of the docs?

tests/trainer/test_checkpoint.py
```diff
@@ -141,10 +142,19 @@ def _assert_weights_equivalent(self, m1: torch.nn.Module, m2: torch.nn.Module):
         for p1, p2 in zip(m1.parameters(), m2.parameters()):
             torch.testing.assert_close(p1, p2)
 
-    def get_trainer(self, **kwargs):
+    def get_trainer(self, checkpoint_save_path=None, overwrite_checkpoints=None, **kwargs):
```
hanlint, line 145: Explicit over implicit: instead, just have the tests calling this to create the checkpointsaver themselves, and pass it in as `callbacks` to `get_trainer`.

Then `get_trainer` just does something like:
```
if 'calbacks' in kwargs:
    callbacks = kwargs.pop('callbacks')
    callbacks.append([DummyStatefulCallback()])
else:
    callbacks = [DummyStatefulCallback()]
```
    
```

File: tests/trainer/test_checkpoint.py [hunk 6]
```diff
@@ -373,10 +381,19 @@ def test_surgery_resumption(self, tmp_path: pathlib.Path):
...
 
+        checkpoint_saver_kwargs = {}
+        if checkpoint_filename:
+            checkpoint_saver_kwargs.update({'checkpoint_filename': checkpoint_filename})
+        if checkpoint_save_path:
+            checkpoint_saver_kwargs.update({'checkpoint_save_path': checkpoint_save_path})
+        if checkpoint_save_interval:
+            checkpoint_saver_kwargs.update({'checkpoint_save_interval': checkpoint_save_interval})
+
+        checkpoint_saver = CheckpointSaver(**checkpoint_saver_kwargs)
```
hanlint, line 396: same comment here as above, this is important for readability of the tests, that the checkpointsaver creation is actually being done by the test, not the helper function.

composer/trainer/trainer.py
```diff
@@ -608,39 +606,10 @@ class Trainer:
                 For fine-grained control on checkpoint saving (e.g. to save different types of checkpoints
                 at different intervals), leave this parameter as ``None``, and instead pass
                 instance(s) of :class:`~.CheckpointSaver` directly as ``callbacks``.
-        save_filename (str, optional): A format string describing how to name checkpoints.
```
hanlint, line 611: Can the `checkpoint_save_path` docstrings mention something about additional options using the `CheckpointSaver` callback? Otherwise, users may think we have a pretty limited interface.

hanlint (commented):
Looking good! didn't dive into the details yet, but made some initial comments. Will make a second pass when ready.

tests/trainer/test_checkpoint.py
eracah: Oh nice. Good idea! Dependency injection. Done.

tests/trainer/test_checkpoint.py
eracah: done

composer/trainer/trainer.py
eracah: they do right?

eracah: this isn't a subtraction. It's added back in below.

composer/callbacks/checkpoint_saver.py
eracah: changed to `save_weights_only`, like Keras does

composer/trainer/trainer.py

composer/trainer/trainer.py
eracah: Ok that's fine. I need this logic for the next PR, so I will keep it.

eracah: assigned you a JIRA!

eracah:
Ok @hanlint, @mvpatel2000, take another look. I think everything should be resolved.

composer/trainer/trainer.py
```diff
@@ -945,21 +911,23 @@ def __init__(
...
-                latest_checkpoint_filename=save_latest_filename,
-                latest_artifact_name=save_latest_artifact_name,
-                overwrite=save_overwrite,
-                weights_only=save_weights_only,
                 checkpoint_save_interval=checkpoint_save_interval,
                 num_checkpoints_to_keep=num_checkpoints_to_keep,
             )
             self.state.callbacks.append(self._checkpoint_saver)
+        # If the user specified a checkpoint callback, then use that as the _checkpoint_saver callback.
+        # TODO(eracah): deal with multiple CheckpointSaver callbacks.
```
mvpatel2000, line 928: Will this be fixed in current series of PRs? Or do we need to move to JIRA?

mvpatel2000 (commented):
This LGTM, but I'd probably want @hanlint to also do a pass since this is very user-facing, and it will be good to have 2 reviews.

composer/trainer/trainer.py
eracah: oh yeah let me move to JIRA

eracah: done

bandish-shah: Thanks for filing the Jira to track!  Can we remove the TODO and print a warning here notifying the user that we're just using the first callback they specified in the list?

bandish-shah (approved):
LGTM!  Approving to unblock but please address my comment about printing a warning to the user if more than one `CheckpointSaver` callback is specified.

composer/trainer/trainer.py
eracah: done!

PR #1531, opened 2022-09-16
Add a log every N batches feature with console_log_interval flag

eracah:
CO-1089
Adds an argument to the Trainer and ProgressBarLogger to log to the console every `console_log_interval` batches.


eracah:
@abhi-mosaic, is this too similar to https://github.com/mosaicml/composer/pull/1357? Should I rename the argument to `log_every_n_batches`

hanlint (commented):
This looks good, thanks, some minimal comments

abhi-mosaic (commented):
Ditto to @hanlint 's comments, but this looks great to me!

PR #1533, opened 2022-09-16
Changing 'stable' to 'latest' in notebooks in examples

bcui19:
Changing 'stable' to 'latest' in notebooks in examples. 

CO-965


bandish-shah:
We actually want to continue using 'stable'.  What you need to do is hunt down the the API Reference URL of the object being referenced in the notebook on `latest` and then change 'latest' to 'stable' in that URL.

dblalock:
So [the issue](https://mosaicml.atlassian.net/browse/CO-965) is mostly that the layout of the API reference changed. E.g., [on this page](https://docs.mosaicml.com/en/latest/examples/training_without_local_storage.html) if you scroll down and click on `Logger` in the first set of bullet points, it links to:
```https://docs.mosaicml.com/en/stable/api_reference/composer.loggers.logger.html#composer.loggers.logger.Logger```

The problem with this isn't the "stable". It's that Logger is now listed in the API reference under a different path as 
```https://docs.mosaicml.com/en/latest/api_reference/generated/composer.loggers.Logger.html```. 

I.e., it moved from `composer.loggers.logger.Logger` to `composer.loggers.Logger` (since the latter is how we expect users to actually reference it).

Except in this example, we apparently don't even include `Logger` in the `composer.loggers` `__init__.py`'s `__all__`, so there isn't even a page for it in the new API reference. This might happen in other cases as well, although usually the name will have just moved.

So this sadly isn't a `sed` job, although greping for something like `://docs.mosaicml.com/en/stable/api_reference` might help you find hardcoded URLs that need to manually be updated.

@bcui19 Does that make sense?

PR #1535, opened 2022-09-17
Less strict name formatting

### LINKED ISSUE #1511: Remove `local_world_size` and `node_rank` requirement from file name formatting
## 🚀 Feature Request
It would be more flexible to not have the `local_world_size` and `node_rank` requirement in file name formatting when running composer trainers.

## Motivation
When using other plug-ins, PyTorch does not necessarily set those variables and it is difficult/unnecessary to set those variables. Moreover, the default file format string does not even use those, so having to set those environmental variables seems like an extra hassle.

hanlint:
Previously, if `NODE_RANK` was not set, formatting of files would fail, even though `{node_rank}` was not even in the format string.

This PR modifies the formatting in less strict mode, and will only error if requested variable is not set and also not in the format string.

Fixes #1511

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -68,6 +68,10 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
+class MissingEnvironmentError(Exception):
+    pass
+
+
 def _get_distributed_config_var(
     env_var: str,
     human_name: str,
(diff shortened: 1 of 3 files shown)


bandish-shah (approved):
LGTM

PR #1536, opened 2022-09-17
Pinned setuptools version to fix distutils version error

karan6181:
## Description
- Composer CI unittest is failing `AttributeError: module 'distutils' has no attribute 'version'`. This is because distutils Version classes are deprecated and recommendation is use `packaging.version` instead but the changes is not controlled by us as it is getting originated from [PyTorch Tensorboard code](https://github.com/pytorch/pytorch/blob/master/torch/utils/tensorboard/__init__.py#L2).
- Recommendation step mentioned in `https://github.com/pytorch/pytorch/issues/69894` is to pinned `setuptools` to `59.5.0`.
- There is already a tracking issue in PyTorch to fix this issue `https://github.com/pytorch/pytorch/issues/84712`.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -17,7 +17,7 @@ build:
 requirements:
   host:
     - python >=3.7
-    - setuptools
+    - setuptools <=59.5.0
     - git
   run:
     - python >=3.7,<4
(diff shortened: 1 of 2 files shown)


hanlint (approved):
Thanks!

PR #1552, opened 2022-09-21
Object Store Logger Race Condition + EMA Fix

mvpatel2000:
Fixes [CO-1097](https://mosaicml.atlassian.net/browse/CO-1097). Object logger calls join in `post_close`. At this point, the interpreter has shutdown so new futures cannot be scheduled. So, if the worker for the last checkpoint save has not scheduled the future yet, the upload will error out. To avoid this, we need to wait for all workers to finish after each Trainer function call to ensure all futures are scheduled. 

This is a partial fix because anything logged in close by another callback won't (potentially) be logged. I don't think this is possible to support though since once you're in the close phase, you can't create futures

Note that we can't really add a test for this since it's a race condition.


mvpatel2000:
Update: It looks like there were actually two (!!) bugs going on here. The previous implementation of `checkpoint_save_interval` was not idempotent, so when EMA was modified last week to move the EMA weights into the default weight slot on a checkpoint event, it broke checkpointing for end of training.

composer/callbacks/checkpoint_saver.py
```diff
@@ -55,20 +55,13 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
...
 
+        # Always checkpoint at end of training
         if elapsed_duration >= 1.0:
-            # if doing batch-wise checkpointing, and we saved a checkpoint at the batch_checkpoint event
-            # right before the epoch_checkpoint event, do not save another checkpoint at the epoch_checkpoint
-            # event if the batch count didn't increase.
-            if state.timestamp.batch != last_checkpoint_batch:
-                last_checkpoint_batch = state.timestamp.batch
-                return True
+            return True
```
hanlint, line 64: just so i understand, this is a stylistic change, not functional, correct?

hanlint (approved):
LGTM, thanks for tracing down this nasty race condition!

composer/callbacks/checkpoint_saver.py
mvpatel2000: Discussed offline, this is a functional change and not a stylistic one. See:
> Update: It looks like there were actually two (!!) bugs going on here. The previous implementation of checkpoint_save_interval was not idempotent, so when EMA was modified last week to move the EMA weights into the default weight slot on a checkpoint event, it broke checkpointing for end of training. Fixed checkpoint_save_interval to be idempotent.

mvpatel2000:
Red button merging because Evan is on vacation.

PR #1554, opened 2022-09-22
torch_xla multi-device support

dskhudia:
Adds support for multi-device training using torch_xla. 

The resnet9 on cifar10 trains fine on 2 GPUs using torch_xla.  

Some of the issues I ran into while adding this support:
- with torch_xla on GPUs, dist.is_initialized()returns false
    - Everything is handled by ENV vars
- composer.utils.dist.initialize_dist is called twice in YAHP flow.
    - Which make torch_xla thing we have 2x the number of GPUS
- There are some env vars that we need to setup before forking
- xm.* functions cannot be called until torch_xla env. is init properly.
    - We call them before init in constructing device object.
- broadcast doesn't exist in current release of torch_xla
  - Implemented with all_reduce. 

Test: 

`composer -n 2 examples/run_composer_trainer.py -f composer/yamls/models/resnet9_cifar10.yaml --train_dataset.cifar10.datadir /tmp/cifar --train_dataset.cifar10.download true --val_dataset.cifar10.datadir /tmp/cifar --val_dataset.cifar10.download true --max_duration 2ep --train_subset_num_batches 2 --device 'tpu'`

### DIFF
--- a/composer/trainer/ddp.py
+++ b/composer/trainer/ddp.py
@@ -4,6 +4,7 @@
 """Helpers for running distributed data parallel training."""
 
 import logging
+import os
 from contextlib import contextmanager, nullcontext
 from typing import Callable, ContextManager, Union, cast
 
@@ -11,8 +12,11 @@ import torch.nn
 from torch.nn.parallel import DistributedDataParallel
 
 from composer.core.state import State
-from composer.utils import dist
+from composer.utils import dist, is_torch_xla_installed
 from composer.utils.string_enum import StringEnum
+
+if is_torch_xla_installed():
+    import torch_xla.core.xla_env_vars as xenv
 
 __all__ = ['DDPSyncStrategy', 'ddp_sync_context', 'prepare_ddp_module']
 
@@ -104,6 +108,10 @@ def prepare_ddp_module(module: torch.nn.Module, find_unused_parameters: bool) ->
             to find parameters to not expect gradients for. This is useful if there are some
             parameters in the model that are not being trained.
     """
+    # No ddp wrapping for torch_xla devices
+    if is_torch_xla_installed() and xenv.LOCAL_WORKER in os.environ:
+        return module
+
     if dist.is_available() and dist.is_initialized():
         if any((p.requires_grad for p in module.parameters())):
             log.debug('Wrapping model with DistributedDataParallel')
--- a/composer/trainer/devices/device_tpu.py
+++ b/composer/trainer/devices/device_tpu.py
@@ -27,9 +27,12 @@ class DeviceTPU(Device):
     """
 
     def __init__(self):
-        import torch_xla.core.xla_model as xm
-
-        self._device = xm.xla_device()
+        # Can't use xm.xla_device() api here!
+        # Device is created before distributed is initialized.
+        # Initializing distributed after making a call to xm.xla_device
+        # results in the following error at dist init
+        # RuntimeError: Cannot replicate if number of devices (1) is different from 8
+        self._device = torch.device(f'xla:0')
 
     def module_to_device(self, module: T_nnModule) -> T_nnModule:
 
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -42,8 +42,15 @@ import torch
 import torch.distributed as dist
 import torch.utils.data
 
+from composer.utils.misc import is_torch_xla_installed
+
 if TYPE_CHECKING:
     from composer.trainer.devices import Device
+
+if is_torch_xla_installed():
+    import torch_xla.core.xla_env_vars as xenv
+    import torch_xla.core.xla_model as xm
+    import torch_xla.distributed.xla_multiprocessing as xmp
 
 TObj = TypeVar('TObj')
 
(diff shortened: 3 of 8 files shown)


dskhudia:
@hanlint : device_tpu should really be renamed to device_xla.

mvpatel2000:
> composer.utils.dist.initialize_dist is called twice in YAHP flow.

is this function not idempotent? why does calling it more than once matter? (asking bc we'll hit this soon if people start calling it outside trainer for entrypoints)

composer/utils/misc.py
```diff
@@ -62,3 +65,14 @@ def model_eval_mode(model: torch.nn.Module):
...
+
+def is_torch_xla_installed() -> bool:
+    """Whether torch_xla is installed in the current environment."""
+    try:
+        import torch_xla.core.xla_model as xm
+        del xm
+    except ImportError:
+        return False
+    else:
+        return True
```
mvpatel2000, line 78:
```suggestion
def is_torch_xla_installed() -> bool:
    """Whether torch_xla is installed in the current environment."""
    try:
        import torch_xla.core.xla_model as xm
        del xm
    except ModuleNotFoundError:
        return False
    else:
        return True
```
nit: if there's an error where module is found but it fails at some other step of importing, do we want to raise this to user? my instinct is yes

composer/trainer/ddp.py
```diff
@@ -104,7 +109,11 @@ def prepare_ddp_module(module: torch.nn.Module, find_unused_parameters: bool) ->
             to find parameters to not expect gradients for. This is useful if there are some
             parameters in the model that are not being trained.
     """
-    if dist.is_available() and dist.is_initialized():
+    # No ddp wrapping for torch_xla devices
+    if is_torch_xla_installed() and xenv.LOCAL_WORKER in os.environ:
+        return module
+
+    if (dist.is_available() and dist.is_initialized()):
```
mvpatel2000, line 116: minor nit
```suggestion
    if dist.is_available() and dist.is_initialized():
```

composer/trainer/devices/device_tpu.py
```diff
@@ -27,9 +28,8 @@ class DeviceTPU(Device):
     """
 
     def __init__(self):
-        import torch_xla.core.xla_model as xm
-
-        self._device = xm.xla_device()
+        # Can't use xm.xla_device() api here!
+        self._device = torch.device(f'xla:0')
```
mvpatel2000, line 32: Why can't we use it here?

composer/utils/dist.py
```diff
@@ -256,6 +294,9 @@ def broadcast_object_list(object_list: List[Any], src: int = 0) -> None:
     Returns:
         None:  ``object_list`` will be modified in-place and set to values of ``object_list`` from the ``src`` rank.
     """
+    if is_torch_xla_installed():
+        raise RuntimeError(f'broadcast_object_list is not supported with torch_xla!')
```
mvpatel2000, line 298:
```suggestion
        raise RuntimeError(f'broadcast_object_list is not supported with torch_xla')
```
(very) minor nit, feel free to ignore. Prefer no `!` to be consistent

composer/utils/dist.py
```diff
@@ -305,6 +346,9 @@ def all_gather_object(obj: TObj) -> List[TObj]:
     Returns:
         List[TObj]: A list of objects indexed by rank.
     """
+    if is_torch_xla_installed():
+        raise RuntimeError(f'all_gather_object is not supported with torch_xla!')
```
mvpatel2000, line 350:
```suggestion
        raise RuntimeError(f'all_gather_object is not supported with torch_xla')
```
same as above

composer/utils/dist.py
```diff
@@ -375,6 +419,9 @@ def initialize_dist(device: Device, timeout: datetime.timedelta):
                                f'of the current process group ({dist.get_backend()}). If you '
                                'wish to change backends, please restart the python process.')
         return
+    elif is_torch_xla_installed():
+        if xenv.LOCAL_WORKER in os.environ:
+            return
```
mvpatel2000, line 424:
```suggestion
    elif is_torch_xla_installed() and xenv.LOCAL_WORKER in os.environ:
            return
```

mvpatel2000 (commented):
Minor nits and questions. Defer to others for approval because less familiar with this. Can test and read through more closely if no one is familiar with it lol

dskhudia:
> > composer.utils.dist.initialize_dist is called twice in YAHP flow.
> 
> is this function not idempotent? why does calling it more than once matter? (asking bc we'll hit this soon if people start calling it outside trainer for entrypoints)

The way it's written, calling it second time doesn't have any effect. However, for torch_xla (torch.distributed.is_initialized) always returns False for GPUs so it was going through init again and torch_xla was assuming 2x the number of devices available. It's no longer a issue as I have added a check for torch_xla as well for repeated calls.

composer/trainer/devices/device_tpu.py
dskhudia: Updated comment. Let me know if it's still not clear.

composer/utils/dist.py
```diff
@@ -227,6 +256,15 @@ def broadcast(tensor: torch.Tensor, src: int) -> None:
             and tensor to be used to save received data otherwise.
         src (int): Source rank
     """
+    if is_torch_xla_installed() and xm.is_xla_tensor(tensor):
```
hanlint, line 259: Cleaner to break this out into its own function, e.g `broadcast_xla()`

hanlint (commented):
Looks good, but general comment is to break out the xla specific functionality into its own method (e.g. xla_all_reduce, xla_broadcast) for for modularity.

dskhudia:
> Looks good, but general comment is to break out the xla specific functionality into its own method (e.g. xla_all_reduce, xla_broadcast) for for modularity.

will do

hanlint:
> In most cases except broadcast it will be thin wrapper. Did you mean I should do it for every function or just where we have some extra logic?

Yeah, I think it would be cleaner if all functions were like:
```
def broadcast():
    if xla:
        _xla_broadcast(..)
    else:
         _dist_broadcast(..)  # or dist.broadcast if it really is just a one-liner
```

it's more verbose, but easier to read. The helper functions can be private, so no docstrings needed.

bandish-shah:
Closing out, we'll re-visit in the future.

PR #1557, opened 2022-09-23
Fix Onnx Export for Composer HuggingFaceModels

nik-mosaic:
- Due to [how torch.onnx.export](https://github.com/pytorch/pytorch/blob/108b25db25ef180bfbfaed2347c9b99103aa68ef/torch/onnx/utils.py#L250) processes arguments that are dictionaries (and since our HuggingFaceModel inputs are generally dictionaries), Onnx export for HuggingFaceModels is currently broken. This PR should fix this.

We fix this by writing 
`export_for_inference(..., sample_input=(input_batch, {})` 
instead of 
`export_for_inference(..., sample_input=(input_batch, )`. 

This works for both tensor inputs (ResNets, etc.) and dictionary inputs. 

In addition to the above change, we move the model and sample_input to CPU, since ONNX export models and inputs cannot be on the GPU. We add a test to verify this works.

We also manually specify the ONNX export opset_version to 13. This is currently the default version for PyTorch 1.12.1, but not for older versions of PyTorch (opset_version 9 was the previous default), which was causing an error. Ideally we would not specify this manually, but I cannot find a way to get the latest supported opset_version. Once we stop supporting PyTorch < 1.12.1, we should stop manually specifying this.

Finally, we also add an optional dynamic_axes input, which is required for older (1.10.2) PyTorch CPU versions when exporting HuggingFaceModels. If users in this scenario do not provide dynamic_axes for their inputs, they may see the following UserWarning and their exported model may not be correct.
`UserWarning: Type cannot be inferred, which might cause exported graph to produce incorrect results.`


dskhudia:
@nik-mosaic Thank you for hunting this down. I think a better solution would be if the `sample_input` is a dict we extract the names and pass those to `input_names` argument at export time. 

See this: https://github.com/pytorch/pytorch/blob/108b25db25ef180bfbfaed2347c9b99103aa68ef/torch/onnx/utils.py#L288

dskhudia:
Also while we are at this, we may want to assign the output names as well. HF models, depending on a config value, return a dict of outputs. https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/modeling_bert.py#L1373-L1378

dskhudia:
See the implementation of this function as well. https://github.com/huggingface/transformers/blob/02b176c4ce14340d26d42825523f406959c6c202/src/transformers/onnx/convert.py#L84

dskhudia:
Could you also check if the output type (dict or tuple of tensor) is the same between eager model output vs onnx exported model?

composer/utils/inference.py
```diff
@@ -180,11 +180,17 @@ def export_for_inference(
                     raise ValueError(f'sample_input argument is required for onnx export')
                 sample_input = ensure_tuple(sample_input)
 
+                input_names = ['input']
+
+                # Extract input names from sample_input if it is a dict
+                if isinstance(sample_input[0], dict):
```
dskhudia, line 186: Not sure if HF (or others) Models have a combination of keyword vs non-keyword arguments. For just keyword or just non-keyword, this would work.

tests/utils/test_inference.py
```diff
@@ -292,7 +354,6 @@ def test_export_with_other_logger(model_cls, sample_input):
                 model=model,
                 save_format=save_format,
                 save_path=save_path,
-                sample_input=(sample_input,),
```
dskhudia, line 295: This is not needed?

composer/utils/inference.py
nik-mosaic: Is there any way to make this robust?

tests/utils/test_inference.py
nik-mosaic: No, since this is a torchscript export.

nik-mosaic:
> Also while we are at this, we may want to assign the output names as well. HF models, depending on a config value, return a dict of outputs. https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/modeling_bert.py#L1373-L1378

Wouldn't this require writing Hugging Face-specific code for the output names? Can we do this and remain agnostic to the model type?

nik-mosaic:
> Could you also check if the output type (dict or tuple of tensor) is the same between eager model output vs onnx exported model?

They are not, the onnx output is just a list of arrays, the eager output is a tuple of tensors.

composer/utils/inference.py
dskhudia: Iterating over `sample_input` and extracting the first dict arg except when it's last?

dskhudia:
> > Could you also check if the output type (dict or tuple of tensor) is the same between eager model output vs onnx exported model?
> 
> They are not, the onnx output is just a list of arrays, the eager output is a dict of tensors.

I am ok keeping this as is (i.e., without output names) for now. However, someone directly using onnx output downstream where they were using eager may run into issues if the output of eager is a dict. Could you create a task for this problem?

dskhudia:
Also side comment for a future PR: Are we better off using [`export_pytorch`](https://github.com/huggingface/transformers/blob/02b176c4ce14340d26d42825523f406959c6c202/src/transformers/onnx/convert.py#L84) function from HF for HF models? They might have handled a few more edge cases that we might be missing.

composer/utils/inference.py
nik-mosaic: Done. I extract all dictionary arguments for named inputs.

nik-mosaic:
> Also side comment for a future PR: Are we better off using [`export_pytorch`](https://github.com/huggingface/transformers/blob/02b176c4ce14340d26d42825523f406959c6c202/src/transformers/onnx/convert.py#L84) function from HF for HF models? They might have handled a few more edge cases that we might be missing.

Yes, for HuggingFace models, I think we are better off calling this function. They do a better job with their OnnxConfig class, and have utilities to convert a HuggingFace config to an OnnxConfig so the inputs and outputs are named well for many classes of models.

dskhudia (approved):

composer/utils/inference.py
```diff
@@ -118,12 +119,24 @@ def export_for_inference(
...
 
-    # make a copy of the sample input so that we don't modify the original sample input
+    # Make a copy of the sample input so that we don't modify the original sample input
     sample_input = copy.deepcopy(sample_input)
 
+    # Move model and sample input to CPU for export
+    cpu = torch.device('cpu')
+    model.to(device=cpu)
+    if sample_input is not None:
+        for i in range(len(sample_input)):
```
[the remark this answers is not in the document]
nik-mosaic: Good catch.

hanlint (approved):

PR #1558, opened 2022-09-26
Fix MIoU metric when `self.total_union==0`

abhi-mosaic:
@mvpatel2000 Could you try this branch out with DeepLabV3 and let me know if it fixes the issue you were seeing?

### DIFF
--- a/composer/metrics/metrics.py
+++ b/composer/metrics/metrics.py
@@ -55,8 +55,9 @@ class MIoU(Metric):
 
     def compute(self):
         """Aggregate state across all processes and compute final metric."""
-        per_class_miou = self.total_intersect / self.total_union  # type: ignore (third-party)
-        per_class_miou[self.total_union == 0] = 0
+        total_intersect = self.total_intersect[self.total_union != 0]  # type: ignore (third-party)
+        total_union = self.total_union[self.total_union != 0]  # type: ignore (third-party)
+        per_class_miou = total_intersect / total_union
         return 100 * per_class_miou.mean()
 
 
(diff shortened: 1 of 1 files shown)


composer/metrics/metrics.py
```diff
@@ -55,7 +55,9 @@ def update(self, logits: Tensor, targets: Tensor):
 
     def compute(self):
         """Aggregate state across all processes and compute final metric."""
-        return 100 * (self.total_intersect / self.total_union).mean()  #type: ignore (third-party)
+        per_class_miou = self.total_intersect / self.total_union  # type: ignore (third-party)
+        per_class_miou[self.total_union == 0] = 0
+        return 100 * per_class_miou.mean()
```
mvpatel2000, line 60: Should we be excluding classes that aren't present instead of 0ing it out? CC: @Landanjs @coryMosaicML

mvpatel2000:
LGTM, fixes NaNs ([runs](https://wandb.ai/mosaic-ml/mihir-dice?workspace=user-mihirmosaic)). Will defer to Cory or Landan for approval

composer/metrics/metrics.py
coryMosaicML: Better to exclude, I think. What do you think @Landanjs?

coryMosaicML (approved):

composer/metrics/metrics.py
Landanjs: I think excluding makes more sense as well. Otherwise, the mIoU will be very small depending on the number of classes and how many are present.

Landanjs:
I will try running this, thank you for fixing my bug!

Landanjs (approved):
LGTM!

mvpatel2000 (approved):

PR #1561, opened 2022-09-26
Update context to raise errror

mvpatel2000:
Context now raises error if dist is not initialized instead of silently being skipped. Resolves https://mosaicml.atlassian.net/browse/CO-1107


hanlint (approved):
LGTM, should we add a test for this?

PR #1563, opened 2022-09-27
Add distributed sampler warning

mvpatel2000:
Adds warning on dataspec creation if no distributed sampler is provided. 

Note: Still a draft to get design right, might be buggy / need check to be moved around


mvpatel2000:
@hanlint I see a few options here:
1. Only emit a warning
2. Only emit an error (would anyone want to use something else? this seems too restrictive)
3. Emit warning and add a bool flag to let user skip warning (seems like weird design)
4. Autowrap distributed sampler and give user option to skip with bool flag
Let me know if you have any thoughts? Or we can move to design meeting

mvpatel2000:
Conclusion from poll w research: option 2 for now, with long term plan to do 4

composer/core/data_spec.py
```diff
@@ -174,14 +175,27 @@ def __init__(
...
+                     'or if the dataloader is mid-epoch. It is required that the training dataloader '
+                     'does not have an active iterator, so CPU dataset augmentations can be '
+                     'correctly inserted. To fix, please do not iterate over the dataloader before passing it into '
+                     'the Trainer.'))
+            world_size = dist.get_world_size()
+            if world_size > 1 and not (
+                (dataloader.sampler and isinstance(dataloader.sampler, DistributedSampler)) or
+                (dataloader.batch_sampler and isinstance(dataloader.batch_sampler, DistributedSampler))):
+                raise ValueError(f'The world_size({world_size}) > 1 but dataloader does not use '
+                                 'DistributedSampler. This will cause all ranks to train on the same '
```
hanlint, line 192: can you link to the torch docs for this?

hanlint (approved):
LGTM

composer/core/data_spec.py
```diff
@@ -174,14 +175,27 @@ def __init__(
...
+                                 'data, removing any benefit from multi-GPU training. To resolve this, '
+                                 'create a Dataloader with with DistributedSampler. For example, '
```
dakinggg, line 194: "with with" -> "with"

composer/core/data_spec.py
```diff
@@ -174,14 +175,30 @@ def __init__(
...
+                raise ValueError(
+                    ('The dataloader has an active iterator. This could occur '
+                     'if `persistent_workers=True` and the dataloader has already been iterated, '
+                     'or if the dataloader is mid-epoch. It is required that the training dataloader '
+                     'does not have an active iterator, so CPU dataset augmentations can be '
+                     'correctly inserted. To fix, please do not iterate over the dataloader before passing it into '
+                     'the Trainer.'))
+            world_size = dist.get_world_size()
+            if world_size > 1 and not (
+                (dataloader.sampler and isinstance(dataloader.sampler, DistributedSampler)) or
```
hanlint, line 189: So I just realized that (maybe?) with our streaming dataloader, a `DistributedSampler` should not be set: https://github.com/mosaicml/composer/blob/dev/composer/datasets/imagenet_hparams.py#L190

mvpatel2000: Per offline discussion, looks like we need a check to make sure its not an IterableDataset.

mvpatel2000:
> Mismatched elements: 1 / 5 (20.0%)
Greatest absolute difference: 3.3387914299964905e-07 at index (0, 0) (up to 1e-08 allowed)
Greatest relative difference: 8.56401274043212e-05 at index (0, 0) (up to 1e-05 allowed)

@hanlint  I think this is just bad thresholds. Can you do a closer pass just to double check on this though?

abhi-mosaic (approved):
The logic looks good to me, but could we also include a test that is missing the `DistributedSampler`, and checks that Composer throws a `ValueError` ?

tests/common/datasets.py
```diff
@@ -31,6 +33,22 @@ def __getitem__(self, index: int):
         return self.x[index], self.y[index]
 
 
+def get_random_classification_dataloader(
```
hanlint, line 36: On second thought, I would resist the urge to refactor the dataloader creation to some common file, but rather add dist.get_sampler below everywhere. It's more verbose, but also more readable and communicates the minimal set of argumenst needed to create the dataloader for the test; these rarely change so non DRY is what I would suggest.

tests/common/datasets.py
```diff
@@ -31,6 +33,22 @@ def __getitem__(self, index: int):
...
+    batch_size: int = 2,
```
hanlint, line 37: default batch size for dataloader is 1

hanlint (commented):
🤔 that test has been quite stable.. possible that some dataloader setting was changed incorrectly. 

See my note also on _not_ refactoring into a helper functions.

mvpatel2000:
Migrating this to https://github.com/mosaicml/composer/pull/1598 -- its easier to start from scratch than revert changes to all the tests

PR #1564, opened 2022-09-27
GLU Fixes

mvpatel2000:
Some quality of life improvements:
- does not error out if GLU is applied more than once, and instead a warning is emitted
- improvements to error messages to be more clear


alextrott16 (approved):
LGTM

dskhudia (approved):

PR #1567, opened 2022-09-28
Update comet links to include query params and point to main site

dakinggg:
- Change main comet link to point to the main site instead of the docs
- Add query params to all comet links

Items 1 and 5 in CO-1104


eracah (approved):
LGTM!
Thanks, @dakinggg !

PR #1570, opened 2022-09-29
Comet example

dakinggg:
Adds CometML in a couple places in the documentation, including the main logging example.

@eracah let me know if we should add a more dedicated example somewhere too. I don't think it would contain anything additional except setting the api key.

Item 4 from CO-1104


eracah (approved):
Perfect! LGTM. Thanks for doing these, @dakinggg!

PR #1571, opened 2022-09-29
Add missing _enabled flag, post_close, and clean up comet ml tests

dakinggg:
- Call experiment.end in post_close
- Log a Created from value to CometML
- Refactor test to be more complete

Items 2 and 3 from CO-1104
Also fixes CO-1160

Run from v0.10.0 branch: https://www.comet.com/mosaic-ml/daniel-test/92f7bd9ec1a948408fccfd59eb47f0bc?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step
Run from the commit in this PR on top of v0.10.0 (not on top of dev, because MNIST example is broken on dev right now): https://www.comet.com/mosaic-ml/daniel-test/dca086a1a5b1453f9640d9c165d38bb4?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step

Can see that the run on v0.10.0 branch (1:27pm) has an extra ghost experiment, while the run with this PR (1:22pm) does not 
Screen Shot 2022-09-29 at 1 30 12 PM

BERT run from this PR: https://www.comet.com/mosaic-ml/daniel-test/a80735743905431da12661c664f7e52f?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step

### DIFF
--- a/tests/loggers/test_cometml_logger.py
+++ b/tests/loggers/test_cometml_logger.py
@@ -1,15 +1,122 @@
 # Copyright 2022 MosaicML Composer authors
 # SPDX-License-Identifier: Apache-2.0
-
 import os
 import zipfile
+from collections import defaultdict
 from json import JSONDecoder
 from pathlib import Path
 
 import pytest
+from torch.utils.data import DataLoader
+
+from composer.trainer import Trainer
+from tests.common import RandomClassificationDataset, SimpleModel
 
 
-def test_comet_ml_logging(monkeypatch, tmp_path):
+def test_comet_ml_logging_train_loop(monkeypatch, tmp_path):
+    pytest.importorskip('comet_ml', reason='comet_ml is optional')
+    import comet_ml
+
+    monkeypatch.setattr(comet_ml, 'Experiment', comet_ml.OfflineExperiment)
+    from composer.loggers import CometMLLogger
+
+    # Set offline directory.
+    offline_directory = str(tmp_path / Path('.my_cometml_runs'))
+    os.environ['COMET_OFFLINE_DIRECTORY'] = offline_directory
+
+    comet_logger = CometMLLogger()
+
+    trainer = Trainer(
+        model=SimpleModel(),
+        train_dataloader=DataLoader(RandomClassificationDataset()),
+        train_subset_num_batches=2,
+        max_duration='2ep',
+        loggers=comet_logger,
+    )
+    trainer.fit()
+
+    del trainer
+
+    assert comet_logger.experiment is not None
+
+    # Open, decompress, decode, and extract offline dump of metrics.
+    comet_exp_dump_filepath = str(Path(offline_directory) / Path(comet_logger.experiment.id).with_suffix('.zip'))
+    zf = zipfile.ZipFile(comet_exp_dump_filepath)
+    comet_logs_path = zf.extract('messages.json', path=offline_directory)
+    jd = JSONDecoder()
+    msg_type_to_msgs = defaultdict(list)
+
+    with open(comet_logs_path) as f:
+        for line in f.readlines():
+            parsed_line = jd.decode(line)
+            msg_type_to_msgs[parsed_line['type']].append(parsed_line['payload'])
+
+    # Check that init set the run name
+    assert comet_logger.name is not None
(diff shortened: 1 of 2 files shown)


tests/loggers/test_cometml_logger.py
```diff
@@ -1,15 +1,122 @@
...
+    msg_type_to_msgs = defaultdict(list)
+
+    with open(comet_logs_path) as f:
+        for line in f.readlines():
+            parsed_line = jd.decode(line)
+            msg_type_to_msgs[parsed_line['type']].append(parsed_line['payload'])
+
+    # Check that init set the run name
+    assert comet_logger.name is not None
+    assert comet_logger.experiment.name is not None
```
eracah, line 56: The name won't be None if it's not set. It will actually be a randomly chosen name, so the best way to check if it is set is to check if it's value is correct.
```suggestion
    assert comet_logger.experiment.name == trainer.state.run_name
```
You'd have to remove the `del trainer` line here also.

tests/loggers/test_cometml_logger.py
```diff
@@ -1,15 +1,122 @@
...
+
+    # Check that basic metrics appear in the comet logs
+    assert len([
+        metric_msg for metric_msg in msg_type_to_msgs['metric_msg'] if metric_msg['metric']['metricName'] == 'epoch'
+    ]) == 2
+
+    # Check that basic params appear in the comet logs
+    assert len([
+        metric_msg for metric_msg in msg_type_to_msgs['metric_msg'] if metric_msg['metric']['metricName'] == 'epoch'
+    ]) > 0
```
eracah, line 66: wait shouldn't this be
```suggestion
    assert len([
        param_msg for param_msg in msg_type_to_msgs['parameter_msg'] if param_msg['param']['paramName'] == 'rank_zero_seed'
    ]) > 0
```
or something like this?

tests/loggers/test_cometml_logger.py
```diff
@@ -1,15 +1,122 @@
...
+    comet_exp_dump_filepath = str(Path(offline_directory) / Path(comet_logger.experiment.id).with_suffix('.zip'))
+    zf = zipfile.ZipFile(comet_exp_dump_filepath)
+    comet_logs_path = zf.extract('messages.json', path=offline_directory)
+    jd = JSONDecoder()
+    msg_type_to_msgs = defaultdict(list)
+
+    with open(comet_logs_path) as f:
+        for line in f.readlines():
+            parsed_line = jd.decode(line)
+            msg_type_to_msgs[parsed_line['type']].append(parsed_line['payload'])
```
eracah, line 52: Nice! A much cleaner implementation than `test_comet_ml_log_metrics_and_hyperparameters`

tests/loggers/test_cometml_logger.py
```diff
@@ -1,15 +1,122 @@
...
+        train_subset_num_batches=2,
+        max_duration='2ep',
+        loggers=comet_logger,
+    )
+    trainer.fit()
+
+    del trainer
+
+    assert comet_logger.experiment is not None
+
```
eracah, line 41: could you check for experiment ended here?
```suggestion
assert comet_logger.experiment.ended
```

eracah (changes requested):
Thanks for doing this! looks pretty good! Just a couple concerns regarding checking for name setting and checking for hyparams/params.

tests/loggers/test_cometml_logger.py
dakinggg: ah, this none check was to satisfy the type checker, but yes, i should add a check for ended as well

tests/loggers/test_cometml_logger.py
dakinggg: ah good point. I have to `del trainer` so that `experiment.end()` is called, but I'll just save the correct name before that.

tests/loggers/test_cometml_logger.py
dakinggg: Indeed it should

eracah (approved):
LGTM! Thanks, @dakinggg !

PR #1576, opened 2022-09-30
Remove eval batch split

mvpatel2000:
We originally added `eval_batch_split` as a feature similar to `grad_accum` to be run when evaluating metrics. This is currently used in eval mode and in some algorithms.

However, with the metrics refactor, we now run train metrics at a microbatch level. This uses the cached output from the forward call in that microbatch level. Now, `eval_batch_split` is no longer needed in train metrics. Further, if it is triggered, it will actually cause an error as composer still returns the cached outputs. Instead, we should let any OOMs here raise all the way up to auto grad accum level.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1721,7 +1721,9 @@ class Trainer:
         assert self._train_data_spec is not None, 'The train data spec should be set on __init__ or fit()'
         assert self.state.train_metrics is not None, 'The train metrics should be set on __init__ or fit()'
 
-        with torch.no_grad(), model_eval_mode(self.state.model), get_precision_context(self.state.precision):
+        with torch.no_grad(),\
+                model_eval_mode(self.state.model),\
+                get_precision_context(self.state.precision):
             if hasattr(self._original_model, 'validate'):  # backwards compatibility check
                 warnings.warn(
                     'Using validate() is no longer supported and will be removed in a future version. Please use eval_forward() instead.'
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -1721,48 +1721,24 @@ def _eval_train_metrics(self, device_batch):
...
-                                    metric.update(eval_outputs, target)
-                            else:
-                                eval_outputs = self._original_model.eval_forward(eval_microbatch, self.state.outputs)
-                                for _, metric in self.state.train_metrics.items():
-                                    self._original_model.update_metric(
-                                        eval_microbatch,
-                                        eval_outputs,
-                                        metric,
-                                    )
+        with torch.no_grad(), model_eval_mode(self.state.model), get_precision_context(self.state.precision):
```
hanlint, line 1724: any chance we can break this up into three lines?
```
with X, \
        Y, \
        C:
```

hanlint (approved):

composer/trainer/trainer.py
mvpatel2000: Can you elaborate why? I kinda like it on one line -- looks a little cleaner to me.

hanlint: This suggestion is more a style nit, but generally we want to stay away from super long lines, unless necessary. Also each LOC should ideally be doing one thing -- it's easy to miss there there are three contexts happening here. This would be a lot cleaner in python 3.10 as one could do:
```
with (
    torch.no_grad(),
    model_eval_mode(self.state.model),
    get_precision_context(self.state.precision),
):
```

mvpatel2000: Got it -- agree.

PR #1578, opened 2022-10-01
Enable saving and loading of checkpoints without creating an ObjectStoreLogger or ObjectStore

eracah:
CO-1143
CO-1144
CO-1023

Todo:

- [x] Implement saving
- [x] implement loading
- [x] fix autoresume
- [ ] unit tests for new functions in trainer.py
- [ ] unit tests for checkpoint_saver
- [ ] unit tests for load_checkpoint
- [x] manual e2e tests
- [ ] update docstrings accordingly
- [ ] update docs accordingly
- [ ] proofread/clean up
- [ ] add comments
- [x] fix wandb loading issue?
- [ ] fix pyright issues


composer/callbacks/checkpoint_saver.py
```diff
@@ -297,29 +271,44 @@ def __init__(
...
 
-        self.folder = folder
-
-        self.filename = PartialFilePath(filename.lstrip('/'), folder)
-        self.latest_filename = PartialFilePath(latest_filename.lstrip('/'), folder) if latest_filename else None
-
-        self.artifact_name = PartialFilePath(artifact_name) if artifact_name else None
-        self.latest_artifact_name = PartialFilePath(latest_artifact_name) if latest_artifact_name else None
+        
+        self.folder = urlparse(folder).path
```
dakinggg, line 275: no need to duplicate the urlparse call here, can make a variable `parsed_url`

composer/callbacks/checkpoint_saver.py
```diff
@@ -297,29 +271,44 @@ def __init__(
...
+        
 
+        self.filename = filename
+        self.latest_filename = latest_filename
         self.overwrite = overwrite
         self.saved_checkpoints: List[str] = []
         self.num_checkpoints_to_keep = num_checkpoints_to_keep
         self.weights_only = weights_only
+        self.checkpoint_file_path = str(Path(self.folder) / Path(self.filename))
+        if latest_filename:
```
dakinggg, line 288: prefer being explicit about which version of truthiness your relying on here. so `if latest_filename is None or len(latest_filename) == 0:`

composer/callbacks/checkpoint_saver.py
```diff
@@ -358,60 +347,50 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
...
-        if is_deepspeed and '{rank}' not in self.filename.filename:
-            raise ValueError(f'Save filename {self.filename.filename} must have {{rank}} for deepspeed.')
-
-        # save the checkpoint to the filename
-        filename = self.filename.format(state, is_deepspeed)
-
         saved_path = checkpoint.save_checkpoint(
             state=state,
-            filename=filename,
+            filename=self.file_path,
```
dakinggg, line 352: I think `self.file_path` is undefined?

composer/callbacks/checkpoint_saver.py
```diff
@@ -358,60 +347,50 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
...
+        checkpoint_file_path += ('.tar' if is_deepspeed and not is_tar(checkpoint_file_path) else '')
+
         if not saved_path:  # not all ranks save
             return
 
         if self.latest_filename is not None:
-            symlink = self.latest_filename.format(state, is_deepspeed)
-            os.makedirs(os.path.dirname(symlink), exist_ok=True)
+            symlink_name = format_name_with_dist_and_time(self.latest_filename, state.run_name, state.timestamp)
+            symlink_name += ('.tar' if is_deepspeed and not is_tar(self.latest_filename) else '')
```
dakinggg, line 364: prefer a function for this deepspeed bit since it is called multiple times

composer/callbacks/checkpoint_saver.py
```diff
@@ -358,60 +347,50 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
...
-                state,
-                is_deepspeed,
-            ).lstrip('/')
 
+      
+        if self.backend is not None:
             logger.file_artifact(log_level=log_level,
-                                 artifact_name=artifact_name,
-                                 file_path=filename,
+                                 artifact_name=checkpoint_file_path,
```
dakinggg, line 377: Does this remove the ability to name the checkpoint file differently locally and on the cloud? I think we should probably preserve that. Also the old default for the artifact name is probably a better default than the current default for the filename

composer/trainer/trainer.py
```diff
@@ -304,6 +307,34 @@ def _is_tpu_installed() -> bool:
...
+    backend, bucket = parse_result.scheme, parse_result.netloc
+    if backend == 's3':
+        return ObjectStoreLogger(object_store_cls=S3ObjectStore,
+                                 object_store_kwargs = {'bucket': bucket})
+    elif backend == 'wandb':
+        return WandBLogger(log_artifacts=True)
+    elif backend == '': # Local path
+        return None
+    else:
+        raise NotImplementedError(f'There is not implementation for the cloud backend {backend}. Please use either'
```
dakinggg, line 321: "There is not implementation" -> "There is not an implementation"

composer/trainer/trainer.py
```diff
@@ -304,6 +307,34 @@ def _is_tpu_installed() -> bool:
...
+    parse_result = urlparse(uri)
+    backend, bucket = parse_result.scheme, parse_result.netloc
+    if backend == 's3':
+        return S3ObjectStore(bucket=bucket)
+    elif backend == 'wandb':
+        return WandBLogger(log_artifacts=True)
+    elif backend == '': # Local path
+        return None
+    else:
+        raise NotImplementedError(f'There is not implementation for the cloud backend {backend}. Please use either'
```
dakinggg, line 334: same as above

composer/trainer/trainer.py
```diff
@@ -782,9 +812,7 @@ def __init__(
         # Save Checkpoint
         save_folder: Optional[str] = None,
         save_filename: str = 'ep{epoch}-ba{batch}-rank{rank}.pt',
-        save_artifact_name: str = '{run_name}/checkpoints/ep{epoch}-ba{batch}-rank{rank}',
```
dakinggg, line 785: same comment as above about losing the ability to name local/remote differently. Maybe I'm just forgetting what we decided, but I thought we were just going to name these args differently like `save_remote_filename` or whatever

composer/trainer/trainer.py
```diff
@@ -946,6 +974,17 @@ def __init__(
                     stream=console_stream,
                 ))
 
+        if checkpoint_save_path is not None:
+            object_store_logger = _create_object_store_logger_from_uri(uri=checkpoint_save_path)
+            if object_store_logger:
+                loggers.append(object_store_logger)
```
dakinggg, line 980: So what happens if I specify a `WandBLogger` and use W&B for checkpointing? Do I get two `WandBLoggers` in this list and is that a problem?

composer/trainer/trainer.py
```diff
@@ -946,6 +974,17 @@ def __init__(
                     stream=console_stream,
                 ))
 
+        if checkpoint_save_path is not None:
```
dakinggg, line 977: I think `checkpoint_save_path` is undefined?

composer/trainer/trainer.py
```diff
@@ -946,6 +974,17 @@ def __init__(
...
+            object_store_logger = _create_object_store_logger_from_uri(uri=checkpoint_save_path)
+            if object_store_logger:
+                loggers.append(object_store_logger)
+
+        
+        load_object_store = _create_object_store_from_uri(uri=load_path) if load_path is not None else None
+        if isinstance(load_object_store, LoggerDestination):
```
dakinggg, line 984: Why does this need to get added to the loggers?

composer/callbacks/checkpoint_saver.py
eracah: fixed

composer/callbacks/checkpoint_saver.py
eracah: ok reverted back to using `PartialFilePath`

composer/trainer/trainer.py
eracah: For these ones, we decided that if you specify `s3://{bucket}{path}`, then the artifact name will be {path} {filename} because the actual filename is the same for both; artifact name just has the parent path prefixed to it, but we can extract the parent path from `save_folder`

composer/trainer/trainer.py
eracah: It doesn't seem to cause a problem because wandb ignores calls to `wandb.log` with the exact same key and step, but we could add some code to de-duplicate loggers.

composer/trainer/trainer.py
eracah: fixed?

composer/trainer/trainer.py
eracah: if it's a `WandBLogger`

dakinggg: What I mean is, if the wandblogger is just an object store for loading a checkpoint, it doesn't actually need to be added to the loggers right? If I want a wandblogger for logging metrics and hyperparams I would specify that separately right?

composer/trainer/trainer.py
dakinggg: Hm, perhaps if the user specifies a wandblogger separately, we just use that logger for checkpoint things as well rather than creating a new one and potentially duplicating calls to wandb. Is that possible/reasonable?

composer/trainer/trainer.py
dakinggg: Ah I see. You're assuming that if a user passes a wandb:// path to load a checkpoint from, they also want wandb logging for their run.

composer/trainer/trainer.py
eracah: fixed

composer/trainer/trainer.py
eracah: fixed

composer/trainer/trainer.py

composer/callbacks/checkpoint_saver.py
eracah: fixed

composer/utils/checkpoint.py
```diff
@@ -67,6 +68,14 @@ def _get_write_mode(name: str) -> str:
     raise ValueError(f'{name} does not end with a valid tarfile extension.')
 
 
+def _parse_load_path(load_path: str):
+    parse_result = urlparse(load_path)
+    if parse_result.scheme == 'wandb':
+        return ''.join([parse_result.netloc, parse_result.path]).lstrip('/').replace('/', '.')
```
dakinggg, line 74: why is this joined with the empty string? I would expect it to be joined with a /. And also why are / replaced with .? I guess I'm missing something about what a load path is expected to be for wandb.

composer/callbacks/checkpoint_saver.py
```diff
@@ -78,6 +81,30 @@ def save_interval(state: State, event: Event):
     return save_interval
 
 
+def _extract_path_from_uri(uri: str):
```
dakinggg, line 84: kind of the same question as below, but why is this different logic from the load path parsing?

dakinggg, line 84: missing return type on these three helpers

composer/callbacks/checkpoint_saver.py
```diff
@@ -358,48 +391,54 @@ def _save_checkpoint(self, state: State, logger: Logger, log_level: LogLevel):
...
-                state,
-                is_deepspeed,
-            ).lstrip('/')
+        # If upload_file is True, then user specified a folder argument like so:
+        # scheme://{bucket}/{path}, which means the artifact name and the file_path are the same.
+        if self.upload_file:
+            # Upload checkpoint file
+            artifact_name = file_path
+            if self.using_wandb:
+                artifact_name = artifact_name.replace('/', '.')
```
dakinggg, line 425: same question as other places, but why does this replace happen? Maybe add a comment if it is something specific to what wandb expects

composer/utils/checkpoint.py
```diff
@@ -67,6 +68,14 @@ def _get_write_mode(name: str) -> str:
     raise ValueError(f'{name} does not end with a valid tarfile extension.')
 
 
+def _parse_load_path(load_path: str):
```
dakinggg, line 71: missing return type

composer/loggers/wandb_logger.py
```diff
@@ -101,6 +101,14 @@ def __init__(
         self.entity = entity
         self.project = project
 
+    @property
+    def log_artifacts(self):
```
dakinggg, line 105: missing return type

PR #1592, opened 2022-10-04
Image viz test

dakinggg:
# What does this PR do?
Unxfails an `ImageVisualizer` test and adds a test to make sure that `ImageVisualizer` + W&B doesn't cause problems.

Fixes CO-1181
Tests for the fix to the bug in CO-1169

# Before submitting

### DIFF
--- a/tests/callbacks/test_image_visualizer.py
+++ b/tests/callbacks/test_image_visualizer.py
@@ -35,7 +35,7 @@ def test_image_visualizer(interval: str):
         eval_dataloader=DataLoader(RandomImageDataset()),
         max_duration='1ep',
     )
-    pytest.xfail('This test segfaults. See https://mosaicml.atlassian.net/browse/CO-776')
+
     trainer.fit()
     num_train_steps = int(trainer.state.timestamp.batch)
     num_train_tables = len(in_memory_logger.data['Images/Train'])
--- a/tests/callbacks/test_image_visualizer.py
+++ b/tests/callbacks/test_image_visualizer.py
@@ -55,7 +55,7 @@ def test_wandb_and_image_visualizer():
 
     import wandb
 
-    image_interval = 4
+    image_interval = 2
     image_visualizer = ImageVisualizer(interval=f'{image_interval}ba')
     wandb_logger = WandBLogger()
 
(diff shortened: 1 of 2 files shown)


tests/callbacks/test_image_visualizer.py
```diff
@@ -44,3 +47,64 @@ def test_image_visualizer(interval: str):
...
+
+@pytest.mark.skipif(not _WANDB_INSTALLED, reason='Wandb is optional')
+def test_wandb_and_image_visualizer():
+    os.environ['WANDB_MODE'] = 'dryrun'
+
+    import wandb
+
+    image_interval = 2
+    image_visualizer = ImageVisualizer(interval=f'{image_interval}ba')
+    wandb_logger = WandBLogger()
```
hanlint, line 60: Would it be cleaner to provide `mode='offline'` to the `wandb.init`? (e.g. https://docs.wandb.ai/ref/python/init)

or alternatively, maybe `os.environ[WANDB_MODE]='offline'` should be a fixture that runs across the entire codebase so we are never dependant on wandb uptime?

hanlint: Also, anyway to set the file dir to a `tmpdir` fixture?

dakinggg: yes, I'll use a tmpdir and pass offline to init. I think we can leave it up to individual tests whether or not to use offline mode. currently this would be the only test running that actually uses the `WandBLogger` as far as i can tell.

dakinggg: @hanlint It turned out there was already a fixture that was setting mode to `disabled` for tests not market remote, so I just changed that to `offline` instead. Does that make sense? Or is that going to mess with other tests?

hanlint: Ah, that makes sense!

tests/fixtures/new_fixtures.py
```diff
@@ -45,7 +45,7 @@ def empty_logger(minimal_state: State) -> Logger:
 def disable_wandb(monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest):
     monkeypatch.setenv('WANDB_START_METHOD', 'thread')
     if request.node.get_closest_marker('remote') is None:
-        monkeypatch.setenv('WANDB_MODE', 'disabled')
+        monkeypatch.setenv('WANDB_MODE', 'offline')
```
eracah, line 48: Nice trick. Always set WANDB to offline mode

tests/callbacks/test_image_visualizer.py
```diff
@@ -44,3 +47,60 @@ def test_image_visualizer(interval: str):
...
+        all_run_text = _wandb_file.read()
+
+    train_metrics_accuracy_count = all_run_text.count('metrics/train/Accuracy')
+    eval_metrics_accuracy_count = all_run_text.count('metrics/eval/Accuracy')
+    eval_metrics_cross_entropy_count = all_run_text.count('metrics/eval/CrossEntropy')
+    train_loss_count = all_run_text.count('loss/train/total')
+
+    expected_number_train_loss_count = (dataset_size / batch_size) + 1  # wandb includes it in the file one extra time
+    expected_number_train_metrics_count = (dataset_size /
+                                           batch_size) + 2  # wandb includes it in the file two extra times
```
eracah, line 91: why the extra time(s)?

eracah (approved):
LGTM Really nice work, Daniel! We sorely needed this test. I like the counting of the keys as a workaround for serialized files. we should probably do the same thing for a Tensorboard unit test if you're game for that?

tests/callbacks/test_image_visualizer.py
dakinggg: I'm not 100% sure because the serialized file is a bit difficult to read, but it looks like some summary stuff gets logged at the end

dakinggg:
@eracah sure, I'll put a tensorboard test on the backlog

PR #1595, opened 2022-10-05
Rename randaugment to rand_augment

mvpatel2000:
# What does this PR do?

Currently, `randaugment` is inconsistently named compared to the rest of the methods. It should be `rand_augment`. This PR renames the files/imports/docs. Along for the ride, we fix the functional API docs which calls the wrong method.

N/A


mvpatel2000:
Jk I guess there's a lot of places without _. Closing

PR #1597, opened 2022-10-05
fix loss and eval_forward for HF models

dskhudia:
HuggingFace model output is a dict or a tuple based on `use_return_dict` config. For loss/eval_forward we assume that it's a dict. Fix this. 

TEST: 
Model with config that has `return_dict=True` works fine.

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -96,13 +96,19 @@ class HuggingFaceModel(ComposerModel):
         return output
 
     def loss(self, outputs, batch):
-        return outputs['loss']
+        if self.config.use_return_dict:
+            return outputs['loss']
+        else:
+            return outputs[0]
 
     def eval_forward(self, batch, outputs: Optional[Any] = None):
         output = outputs if outputs else self.forward(batch)
         if self.use_logits:
             self.labels = batch.pop('labels')
-            output = output['logits']
+            if self.config.use_return_dict:
+                output = output['logits']
+            else:
+                output = output[1]
 
             # if we are in the single class case, then remove the classes dimension
             if output.shape[1] == 1:
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -99,6 +99,7 @@ class HuggingFaceModel(ComposerModel):
         if self.config.use_return_dict:
             return outputs['loss']
         else:
+            # loss is at index 0 in the output tuple
             return outputs[0]
 
     def eval_forward(self, batch, outputs: Optional[Any] = None):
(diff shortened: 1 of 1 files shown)


composer/models/huggingface.py
```diff
@@ -96,13 +96,19 @@ def forward(self, batch):
         return output
 
     def loss(self, outputs, batch):
-        return outputs['loss']
+        if self.config.use_return_dict:
+            return outputs['loss']
```
karan6181, line 100: Question: Is `outputs` always a list and not a scalar int or float? wondering if there will be a index error or something like that? or do we need a check like `if isinstance(outputs, list)` ?

dskhudia: It's a tuple or dict.

mvpatel2000 (approved):
LGTM. Can we leave a comment indicating where `[0]` and `[1]` are coming from? Assuming you tested this and it works.

Would actually be nice to have a unit test, but not sure if thats easy to do

dskhudia:
> LGTM. Can we leave a comment indicating where `[0]` and `[1]` are coming from? Assuming you tested this and it works.
> 
> Would actually be nice to have a unit test, but not sure if thats easy to do

Added comments. Essentially loss is at index 0 in the tuple and logits are index 1 in the tuple. 

Some examples: https://github.com/huggingface/transformers/blob/a2a3afbc8d26d6170909365ffba6bd75e186255f/src/transformers/models/distilbert/modeling_distilbert.py#L787-L788

https://github.com/huggingface/transformers/blob/a2a3afbc8d26d6170909365ffba6bd75e186255f/src/transformers/models/electra/modeling_electra.py#L1317-L1318

composer/models/huggingface.py
```diff
@@ -96,13 +96,21 @@ def forward(self, batch):
...
     def eval_forward(self, batch, outputs: Optional[Any] = None):
         output = outputs if outputs else self.forward(batch)
         if self.use_logits:
             self.labels = batch.pop('labels')
-            output = output['logits']
+            if self.config.use_return_dict:
+                output = output['logits']
+            else:
+                # logits are at index 1 in the output tuple
+                output = output[1]
```
alextrott16, line 113: The assumption that loss and logits are at indices 0 and 1, respectively, depends on `labels` being in the batch. That seems like a safe assumption when extracting loss, but is there an edge case where the assumption doesn't hold for logits? I suppose `self.labels = batch.pop('labels')` acts like an `assert 'labels' in batch`, but if `outputs` is not `None` (i.e. it was precomputed and given as an input to `eval_forward`) can we be certain about the indexing?

I think this is probably an overly specific concern, so I'll give my approval. But it would be good to double check that logits at `output[1]` will always be a valid assumption.

alextrott16 (approved):

composer/models/huggingface.py
dskhudia: BTW indexing path won't be exercised for general HF models. It's only when someone specifically doesn't want dict outputs, this path will be exercised, i.e., by default it doesn't impact existing workflow at all. 

At least in the models I looked at (didn't look at all), this indexing works as expected.

PR #1598, opened 2022-10-06
Add distributed sampler error

mvpatel2000:
# What does this PR do?

Adds error if dataloader is used without distributed sampler.

[CO-1066](https://mosaicml.atlassian.net/browse/CO-1106)


mvpatel2000:
I'm very unsure why the grad accum check is failing. Would love fresh eyes

tests/trainer/test_trainer.py
```diff
@@ -573,11 +588,16 @@ def test_eval_is_excluded_from_wct_tracking(
...
+        dataset = RandomClassificationDataset()
         trainer = Trainer(
             model=model,
             train_dataloader=train_dataloader,
             train_subset_num_batches=2,  # make training fast
-            eval_dataloader=DataLoader(dataset=RandomClassificationDataset(), batch_size=2),
+            eval_dataloader=DataLoader(
+                dataset=dataset,
+                batch_size=2,
+                sampler=dist.get_sampler(RandomClassificationDataset()),
```
dakinggg, line 599: why not reusing the dataset you defined above here?

composer/core/data_spec.py
```diff
@@ -174,14 +176,33 @@ def __init__(
...
+                    ('The dataloader has an active iterator. This could occur '
+                     'if `persistent_workers=True` and the dataloader has already been iterated, '
+                     'or if the dataloader is mid-epoch. It is required that the training dataloader '
+                     'does not have an active iterator, so CPU dataset augmentations can be '
+                     'correctly inserted. To fix, please do not iterate over the dataloader before passing it into '
+                     'the Trainer.'))
+            world_size = dist.get_world_size()
+            # Check for Distributed Sampler if not using IterableDataset on more than 1 GPU
+            if world_size > 1 and not isinstance(dataloader.dataset, torch.utils.data.IterableDataset):
+                is_sampler_distributed = dataloader.sampler and isinstance(dataloader.sampler, DistributedSampler)
```
dakinggg, line 191: prefer explicitly checking `is None` rather than relying on truthiness?

composer/core/data_spec.py
```diff
@@ -174,14 +176,33 @@ def __init__(
...
+                    dataloader.batch_sampler, DistributedSampler)
+                if not is_sampler_distributed and not is_batch_sampler_distributed:
+                    raise ValueError(
+                        f'The world_size({world_size}) > 1 but dataloader does not use '
+                        'DistributedSampler. This will cause all ranks to train on the same '
+                        'data, removing any benefit from multi-GPU training. To resolve this, '
+                        'create a Dataloader with DistributedSampler. For example, '
+                        'DataLoader(..., sampler=composer.utils.dist.get_sampler(...)).'
+                        'Alternatively, the process group can be instantiated with '
+                        'composer.utils.dist.instantiate_dist(...) and DistributedSampler can '
```
dakinggg, line 202: is this supposed to `initialize_dist` rather than `instantiate_dist`?

mvpatel2000:
@eracah @dakinggg @bcui19 would love some help on this failing test.... I have no idea what's going on :(. Would love suggestions

eracah:
> @eracah @dakinggg @bcui19 would love some help on this failing test.... I have no idea what's going on :(. Would love suggestions

Make the tolerance higher?

mvpatel2000:
> > @eracah @dakinggg @bcui19 would love some help on this failing test.... I have no idea what's going on :(. Would love suggestions
> 
> Make the tolerance higher?

@hanlint thinks there might be a genuine failure. It's not clear to me how the PR changes would affect the values we see...

eracah:
> > > @eracah @dakinggg @bcui19 would love some help on this failing test.... I have no idea what's going on :(. Would love suggestions
> > 
> > 
> > Make the tolerance higher?
> 
> @hanlint thinks there might be a genuine failure. It's not clear to me how the PR changes would affect the values we see... so its weird to see failures starting now. Maybe just randomness change?

what sources of stochasticity do you need to fix? just the `rank_zero_seed`?

bcui19:
I checked this last night, I also wasn't sure why it was failing. Talked with @bandish-shah, mixed precision won't always produce the same results, so it could've been the random seed.

to repo it would be possible to re-run w/ the same seed that made the test fail.

Also now tests are passing, so it could've been the `rank_zero_seed`?

mvpatel2000:
on batch 12, we get different gradients because we get different forward pass outputs. The parameters are the same, and the grad_accum 2 value is closer to being correct (value you get with full precision) than the reference. So, I'm going to raise thresholds and call it a day

bcui19 (approved):
Based on offline discussion around the error, LGTM! Thanks for deep dive.

dakinggg (approved):

PR #1601, opened 2022-10-07
Object store logger refactor

dakinggg:
# What does this PR do?
Renames everything in `ObjectStoreLogger` to not use "object" or "artifact", and instead refer to uploading/downloading files. The commit messages are clear on the "root" of each change (e.g. renaming the `artifact_name` argument to `remote_file_name`, and then each change touches many other files as a result

Closes CO-1189
Closes CO-1208
Relates to CO-1167

# Related design
https://www.notion.so/ObjectStoreLogger-Artifact-Logging-Refactor-dec6545891e74bb0a40756d0177075c8#a1cc877e8256459784363bc8e141130c


eracah (commented):
Wow the code and docs are so much more readable now. Nicely done, especially the file_uploading docs. What do we think of `RemoteFileManager` insead of `UploaderDownloader`?

eracah (approved):
LGTM

eracah:
`object_name` is still used right? The PR description says it gets rid of all `Object` mentions?

composer/trainer/trainer.py
```diff
@@ -782,9 +782,9 @@ def __init__(
         # Save Checkpoint
         save_folder: Optional[str] = None,
         save_filename: str = 'ep{epoch}-ba{batch}-rank{rank}.pt',
-        save_artifact_name: str = '{run_name}/checkpoints/ep{epoch}-ba{batch}-rank{rank}',
+        save_remote_file_name: str = '{run_name}/checkpoints/ep{epoch}-ba{batch}-rank{rank}',
         save_latest_filename: Optional[str] = 'latest-rank{rank}.pt',
-        save_latest_artifact_name: Optional[str] = '{run_name}/checkpoints/latest-rank{rank}',
+        save_latest_remote_file_name: Optional[str] = '{run_name}/checkpoints/latest-rank{rank}',
```
mvpatel2000, line 787: do we want to backwards compatibility support this? CC: @bandish-shah 

I'm fine with hard break tbh because all this will go away anyways?

mvpatel2000 (approved):
LGTM. Agree with Evan on name choice. It's a pain to reupdate tho :(

composer/trainer/trainer.py
dakinggg: Evan is removing those args in his checkpointing refactor anyway

dakinggg:
@eracah oh yeah, that description is not quite accurate, will update. object_name is mostly going away in the next pr

dakinggg:
@mvpatel2000 I don't really think its a file manager because all it can do is upload and download? Also I am totally biased so I don't have to rebase my current work on a name change.

composer/trainer/trainer.py
mvpatel2000: Sounds good :)

mvpatel2000:
> @mvpatel2000 I don't really think its a file manager because all it can do is upload and download? Also I am totally biased so I don't have to rebase my current work on a name change.

I'm convinced. Good with leaving as is.

PR #1603, opened 2022-10-07
Remove log level and should_log_artifact

dakinggg:
# What does this PR do?
Removes the `should_log_artifact` which was unused as far as we know, and would require a user writing a custom filter function that takes in state and `LogLevel` to decide whether to log an artifact. Also, the `WandBLogger` did not support this argument. `should_log_artifact` was also the only use of `LogLevel`, so that has been removed everywhere.

Closes CO-1189
Relates to CO-1167

# Manual testing
Manually tested saving (loading) checkpoints to (from) WandB and S3.


eracah (approved):
Amazing. Feels so good to get rid of LogLevel. Like a big monkey off of our backs

PR #1608, opened 2022-10-08
Add more robust casting to int for fsdp min_params

dblalock:
One-line fix for [CO-1197](https://mosaicml.atlassian.net/browse/CO-1197).

Right now, if your `min_params` for fsdp is a string like '1e8', this line throws. I encountered this when using our public GPT benchmarks.


abhi-mosaic (approved):
LGTM!

PR #1613, opened 2022-10-11
Add LAMB Optimizer

hanlint:
This PR adds the LAMB Optimizer and makes some minor changes to the `Enwiki` streaming dataset. cc: @nik-mosaic


nik-mosaic (approved):
Looks good! Thanks.

composer/datasets/enwiki_hparams.py
```diff
@@ -27,8 +27,6 @@ class StreamingEnWikiHparams(DatasetHparams):
         local (str): Local filesystem directory where dataset is cached during operation.
             Default: ``'/tmp/mds-cache/mds-enwiki/'``
         split (str): What split of the dataset to use. Either ``'train'`` or ``'val'``. Default: ``'train'``.
-        mlm (bool): Whether or not to use masked language modeling. Default: ``False``.
```
mvpatel2000, line 30: Looks like a few extra changes came along for the ride here -- idt we should be deleting these?

hanlint: Defer to @nik-mosaic , but I think these should be deleted. This is for the wiki dataset.

PR #1614, opened 2022-10-12
S3 Checkpoint Saving By URI

eracah:
# What does this PR do?
This PR enables checkpoint saving to S3 using a URI by doing the following:
- [x] parses `save_folder` and if it's URI with `s3://` prefix, then creates an `RemoteUploaderDownloader` object
- [x] checks to make sure a `RemoteUploaderDownloader` object with an S3 `bucket_uri` is not already in `loggers` (if so then doesn't use the URI and throws a warning)
- [x] removes `save_remote_file_name` and `save_latest_remote_file_name` from `Trainer`
- [x] unit tests
- [x] manual tests

fixes CO-1143

# Before submitting


composer/trainer/trainer.py
```diff
@@ -289,6 +291,39 @@ def _generate_run_name() -> str:
...
+        uri: str, loggers: List[LoggerDestination]) -> Optional[RemoteUploaderDownloader]:
+    existing_remote_uds = [logger_dest for logger_dest in loggers if isinstance(logger_dest, RemoteUploaderDownloader)]
+    backend, bucket_name, _ = _parse_uri(uri)
+    if backend == '':
+        return None
+    for existing_remote_ud in existing_remote_uds:
+        if ((existing_remote_ud.remote_backend_name == backend) and
+            (existing_remote_ud.remote_bucket_name == bucket_name)):
+            warnings.warn(
+                f'There already exists a RemoteUploaderDownloader object to handle the uri: {uri} you specified')
```
dakinggg, line 304: I think this is `info` not `warning`? The user doesn't need to do anything, and this is totally fine, right?

dakinggg (commented):
EDIT: disregard

looks good to me, with a question about what should be in this pr vs a future pr. What do you think about including the removal of `remote_file_name` and `latest_remote_file_name` in this PR? Otherwise it is a confusing state because you have to pass in the s3 uri _and_ `remote_file_name` and `latest_remote_file_name`. I guess that would require not allowing the customization of `latest_remote_file_name`?

Will wait to approve until we answer that question and docs/examples are updated.

composer/trainer/trainer.py
```diff
@@ -289,6 +291,39 @@ def _generate_run_name() -> str:
...
+            return None
+    if backend == 's3':
+        return RemoteUploaderDownloader(bucket_uri=f'{backend}://{bucket_name}', backend_kwargs={'bucket': bucket_name})
```
dakinggg, line 307: The backend_kwargs here shouldn't be necessary because the RemoteUploaderDownloader parses the bucket name out of the uri.

composer/trainer/trainer.py
eracah: yeah, but they could remove the RemoteUploaderDownloader or remove the uri prefix?

dakinggg:
Oh I see, the `remote_file_name` and `latest_remote_file_name` are left in to support wandb/libcloud/sftp? Do you think we can get rid of those args by using `wandb://path`, `libcloud://path`, and `sftp://path` in this pr, and just require that the correct logger was created by the user in those cases (deferring automatic creation of those, but still removing the extra arguments)?

dakinggg:
Ok I thought about it some more and I'm ok leaving `remote_file_name` in for now to simplify things, but it looks like if I pass s3://bucket/path/to/checkpoints/ as the save folder, the path/to/checkpoints will only be used for the local path not the remote path. I think that path should be passed to s3 as the prefix?

Also, can you include abhi's request to add the PT file extension to the remote files?

composer/trainer/trainer.py
dakinggg: yeah, I guess for me its not quite warning level, but up to you, not a big deal.

composer/trainer/trainer.py
dakinggg, line 307: Can we raise an error here if the rest of the path is empty? Currently it crashes with `FileNotFoundError: [Errno 2] No such file or directory: ''`

eracah: fixed

eracah:
ok @dakinggg, I added support for empty path strings and I removed `save_remote_file_name` and `save_latest_remote_file_name`. PTAL

tests/trainer/test_checkpoint.py
```diff
@@ -134,6 +137,70 @@ def test_ignore_params(remove_field_paths: List[List[str]], filter_params: List[
...
+            self.get_trainer(save_folder=uri)
+
+    @pytest.mark.parametrize('local_path', ['foo/bar/baz'])
+    def test_local_paths_work(self, local_path: str):
+        self.get_trainer(save_folder=local_path)
+
+    @pytest.mark.parametrize('save_folder,expected_path',
+                             [('s3://bucket_name/{run_name}/my_checkpoints', '{run_name}/my_checkpoints'),
+                              ('{run_name}/my_checkpoints', '{run_name}/my_checkpoints'), ('s3://bucket_name/', ''),
+                              ('s3://bucket_name', '')])
```
dakinggg, line 181: I think this is a duplicate parameterization?

dakinggg (approved):
LGTM (few minor comments), assuming you have manually tested that you can save to wandb, s3 via uri, and s3 via libcloud.

examples/checkpoint_with_wandb.py
```diff
@@ -48,17 +48,17 @@
 
     # Checkpoint Saving Configuration
     loggers=wandb_logger,  # Log checkpoints via the WandB Logger
-    save_folder='/tmp/checkpoints',  # The trainer requires that checkpoints must be saved locally before being upload
+    save_folder='checkpoints',  # This the folder that checkpoints are saved to locally and remotely.
```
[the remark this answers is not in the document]
eracah: yeah that's part of another JIRA

PR #1615, opened 2022-10-12
Auto resumption

dakinggg:
# What does this PR do?
Add support for autoresume with distributed training.

Fixes [CO-1238](https://mosaicml.atlassian.net/browse/CO-1238?atlOrigin=eyJpIjoiNjdmNTdiZDYzODY3NDAyY2EwOGIyMjczMTc4OTg5ZWEiLCJwIjoiaiJ9)
While fixing this issue, we discovered a race condition that is related to this, that is [CO-1270](https://mosaicml.atlassian.net/browse/CO-1270)

# Manual testing
Manually tested a script that makes a trainer, trains it, makes another trainer with autoresume, and verifies that the params and run name are the same. This was tested on multiples gpus on one node.

Multi node does not currently work, but I also can't really test/debug it because of limited capacity on our cluster.

# Before submitting


composer/trainer/trainer.py
```diff
@@ -1209,31 +1230,61 @@ def _get_autoresume_checkpoint(
...
+                return latest_checkpoint_path
+        else:
+            # The checkpoint must at least exist for rank zero
+            latest_checkpoint_exists_on_rank_zero = self._device.tensor_to_device(
+                torch.tensor([os.path.exists(latest_checkpoint_path) and dist.get_global_rank() == 0],
+                             dtype=torch.uint8))
+            dist.all_reduce(latest_checkpoint_exists_on_rank_zero, reduce_operation='MAX')
+
+            if int(latest_checkpoint_exists_on_rank_zero.item()) == 0:
+                raise RuntimeError(
```
hanlint, line 1260: I think we should return `None` here (cc: @mvpatel2000 ?) so that a fresh run with `autoresume: True` still works.

composer/trainer/trainer.py
```diff
@@ -1187,6 +1187,27 @@ def saved_checkpoints(self) -> List[str]:
             return []
         return self._checkpoint_saver.saved_checkpoints
 
+    def _attempt_checkpoint_download(self, save_folder: str, latest_checkpoint_path: str,
```
hanlint, line 1190: nit: `_try_checkpoint_download` ? 

also, perhaps move the `os.makedirs` outside so this function does one thing.

composer/trainer/trainer.py
dakinggg: I did have log an info and return `None` at one point. That results in silent failure if the checkpoint doesn't exist/can't be found for some reason. I wasn't sure which was worse, silent failure in the case of it being missing? or erroring on the first run with autoresume true?

dakinggg: I think log a warning and return None makes sense

hanlint: Later on in the codepath, we already log a warning, so i think returning None makes sense.

THe idea here is that the same code/settings while automatically pick up on old checkpoints and resume.

dakinggg: Ah ok, makes sense.

composer/trainer/trainer.py
```diff
@@ -1209,31 +1228,61 @@ def _get_autoresume_checkpoint(
...
+        if self.deepspeed_enabled:
+            # Require all ranks to have their own local checkpoint if we wish to restore from it for deepspeed
+            latest_checkpoint_exists_on_all_ranks = self._device.tensor_to_device(
+                torch.tensor([os.path.exists(latest_checkpoint_path)], dtype=torch.uint8))
+            dist.all_reduce(latest_checkpoint_exists_on_all_ranks, reduce_operation='MIN')
+
+            if int(latest_checkpoint_exists_on_all_ranks.item()) == 0:
+                raise RuntimeError('DeepSpeed was enabled, but checkpoints were not found for all ranks')
+            else:
+                return latest_checkpoint_path
```
mvpatel2000, line 1250:
```suggestion
            return latest_checkpoint_path
```
minor nit: if it's raising an error, we don't need to `else` case this. Mostly for consistency with rest of style in codebase

composer/trainer/trainer.py
```diff
@@ -1209,31 +1228,61 @@ def _get_autoresume_checkpoint(
...
+        else:
+            # The checkpoint must at least exist for rank zero
+            latest_checkpoint_exists_on_rank_zero = self._device.tensor_to_device(
+                torch.tensor([os.path.exists(latest_checkpoint_path) and dist.get_global_rank() == 0],
+                             dtype=torch.uint8))
+            dist.all_reduce(latest_checkpoint_exists_on_rank_zero, reduce_operation='MAX')
+
+            if int(latest_checkpoint_exists_on_rank_zero.item()) == 0:
+                return None
+            else:
```
mvpatel2000, line 1260: same minor nit to unindent this else and remove it

composer/trainer/trainer.py
```diff
@@ -1209,31 +1228,61 @@ def _get_autoresume_checkpoint(
...
+
+                # At this point the rank 0 filepath should exist on all ranks
+                latest_checkpoint_exists_on_all_ranks = self._device.tensor_to_device(
+                    torch.tensor([os.path.exists(latest_checkpoint_path)], dtype=torch.uint8))
+                dist.all_reduce(latest_checkpoint_exists_on_all_ranks, reduce_operation='MIN')
+
+                if int(latest_checkpoint_exists_on_all_ranks.item()) == 0:
+                    raise RuntimeError('Downloading the checkpoint to all nodes failed')
+                else:
+                    return latest_checkpoint_path
```
mvpatel2000, line 1285:
```suggestion
                return latest_checkpoint_path
```
same minor nit

mvpatel2000 (approved):
Thanks for fixing this!! Approving to unblock, left some really minor nits for conventions we've used in repo. Actually, I'm not sure we're consistent with these... but it'd be nice to do so 🤷

composer/trainer/trainer.py
```diff
@@ -1204,35 +1226,73 @@ def _get_autoresume_checkpoint(
         Returns:
             Optional[str]: The path to the latest checkpoint, if found, otherwise None.
         """
+        log.debug(
+            f'Entering _get_autoresume_checkpoint with save_latest_remote_file_name: {save_latest_remote_file_name}')
```
mvpatel2000, line 1230: Delete since `log.info` prints this later on?

composer/trainer/trainer.py
```diff
@@ -1204,35 +1226,73 @@ def _get_autoresume_checkpoint(
...
 
+        log.info(
+            f'Looking for autoresume checkpoint: {save_latest_remote_file_name} (remote), {latest_checkpoint_path} (local)'
+        )
+
         # If latest checkpoint is not saved locally, try to fetch from loggers
-        if not os.path.exists(latest_checkpoint_path):
-            # Make save folder in case it doesn't exist so latest checkpoint can be downloaded
+        if not os.path.exists(latest_checkpoint_path) and (dist.get_global_rank() == 0 or self.deepspeed_enabled):
+            log.debug(f'Attempting to download the checkpoint on to rank {dist.get_global_rank()}')
```
mvpatel2000, line 1242: remove since we show a progress bar during download?

composer/trainer/trainer.py
```diff
@@ -1204,35 +1226,73 @@ def _get_autoresume_checkpoint(
...
+            dist.broadcast_object_list(save_latest_remote_file_name_list, src=0)
+            save_latest_remote_file_name = save_latest_remote_file_name_list[0]
+
+            # download the checkpoint on local rank 0 of all nodes
+            if dist.get_local_rank() == 0 and not os.path.exists(latest_checkpoint_path):
+                log.debug(f'Attempting to download the checkpoint {save_latest_remote_file_name} on to all nodes')
+                os.makedirs(save_folder, exist_ok=True)
+                self._try_checkpoint_download(latest_checkpoint_path, save_latest_remote_file_name, loggers,
+                                              load_progress_bar)
+            dist.barrier()
```
mvpatel2000, line 1283: I don't think this barrier is necessary because of the reduce two lines down

composer/trainer/trainer.py
```diff
@@ -1204,35 +1226,73 @@ def _get_autoresume_checkpoint(
...
+            # At this point the rank 0 filepath should exist on all ranks
+            latest_checkpoint_exists_on_all_ranks = self._device.tensor_to_device(
+                torch.tensor([os.path.exists(latest_checkpoint_path)], dtype=torch.uint8))
+            dist.all_reduce(latest_checkpoint_exists_on_all_ranks, reduce_operation='MIN')
+
+            log.debug(
+                f'Checkpoint {latest_checkpoint_path} exists on rank {dist.get_global_rank()}? {os.path.exists(latest_checkpoint_path)}'
+            )
```
mvpatel2000, line 1291: delete?

mvpatel2000 (commented):
I'll let Hanlin approve since it'd be nice to get two pairs of eyes on this. LGTM tho. Amazing job!

composer/trainer/trainer.py
```diff
@@ -1204,35 +1226,73 @@ def _get_autoresume_checkpoint(
...
-        # If latest checkpoint is saved locally, change load_path to it
-        if int(latest_checkpoint_exists.item()) == 1:
+            self._try_checkpoint_download(latest_checkpoint_path, save_latest_remote_file_name, loggers,
+                                          load_progress_bar)
+
+        if self.deepspeed_enabled:
+            # Require all ranks to have their own local checkpoint if we wish to restore from it for deepspeed
+            latest_checkpoint_exists_on_all_ranks = self._device.tensor_to_device(
+                torch.tensor([os.path.exists(latest_checkpoint_path)], dtype=torch.uint8))
+            dist.all_reduce(latest_checkpoint_exists_on_all_ranks, reduce_operation='MIN')
```
hanlint, line 1251: suggestion: I think both this check and the one below can be unified with `dist.all_gather_object()`, e.g.
```
# list by each rank
latest_checkpoint_exists = dist.all_gather_object(os.path.exists(latest_checkpoint_path))

if self.deepspeed_enabled:
    if not all(latest_checkpoint_exists):
        missing_ranks = [n for (n, exist) in enumerate(latest_checkpoint_exists) if not exist]
        raise RuntimeError(f'Deepspeed was enabled, but checkpoints missing on ranks: {missing_ranks}')
else:
    if not latest_checkpoint_exists[0]:
        return None
```

hanlint (approved):
LGTM, one suggestion to clean up the code a bit

composer/trainer/trainer.py
dakinggg: discussed offline, barrier is necessary so that `os.path.exists` is called after local rank 0 has finished downloading the checkpoint.

composer/trainer/trainer.py
dakinggg: I will leave this one because the progress bar doesn't show if the download can't be started for some reason,

composer/trainer/trainer.py
dakinggg: This was like the most important line for debugging so I'm going to leave it

composer/trainer/trainer.py
dakinggg: yes, this makes sense, thanks

dakinggg:
@hanlint I changed the `num_concurrent_uploads` default to 1. I experimented a bit, and at least on the demo example, if you checkpoint enough to want multiple upload workers, you end up bottlenecked on network communication with s3 (streaming data in and sending checkpoints out). With 1 worker, you should only end up with an invalid symlink if the job dies between finishing uploading the checkpoint and uploading the symlink (which is tiny). This would be quite unlikely, and even this issue could be resolved by having `save_overwrite=True` when using autoresume.

hanlint (approved):

eracah (approved):
Approving to unblock!

PR #1617, opened 2022-10-12
Add Weight Copying For O1 algorithms

nik-mosaic:
Similar to https://github.com/mosaicml/performance/pull/29, but for Fused LayerNorm. Note that this isn't strictly necessary, because one can apply Fused LayerNorm on Event.INIT, but we include weight copying anyway, since it makes it easier for users.

In addition, we create a new event, Event.POST_LOAD, which occurs after a checkpoint is loaded but before the model is DDP wrapped. This is where math-equivalent methods should be applied to checkpoints for models which have not been trained with these speedups.


mvpatel2000 (commented):
LGTM but defer to Daya

examples/run_composer_trainer.py
```diff
@@ -17,6 +17,8 @@
 import tempfile
 import warnings
 
+import performance
```
dskhudia, line 20: Remove these changes.

composer/algorithms/fused_layernorm/fused_layernorm.py
```diff
@@ -36,7 +36,11 @@ def from_LayerNorm(layer: torch.nn.Module, module_index: int) -> APEXFusedLayerN
     """Defines a replacement policy from a `torch.nn.LayerNorm` to a `apex.normalization.fused_layer_norm`"""
     assert isinstance(layer,
                       torch.nn.LayerNorm), 'The replacement policy will look for all instances of torch.nn.LayerNorm'
-    return APEXFusedLayerNorm(normalized_shape=layer.normalized_shape, eps=layer.eps)
+    fused_layernorm = APEXFusedLayerNorm(normalized_shape=layer.normalized_shape, eps=layer.eps)
```
dskhudia, line 39: This still runs at INIT event right?

dskhudia (approved):

composer/algorithms/fused_layernorm/fused_layernorm.py
nik-mosaic: By default, this runs at INIT. So, technically the weight copy (lines 40-42) is technically not needed, because the checkpoint will be loaded afterward and the weights will come from there. However, because other O1 speedups have to be run at Event.POST_LOAD, I added the weight copy so all O1 speedups have the same behavior at POST_LOAD.

examples/run_composer_trainer.py
nik-mosaic: Removed.

dskhudia (approved):

PR #1625, opened 2022-10-14
Blurpool idempotent

mvpatel2000:
# What does this PR do?

Previously, applying blurpool multiple times caused failures. This now makes it idempotent (useful for agent).

[CO-1213](https://mosaicml.atlassian.net/browse/CO-1213)

### DIFF
--- a/composer/algorithms/blurpool/blurpool_layers.py
+++ b/composer/algorithms/blurpool/blurpool_layers.py
@@ -220,16 +220,18 @@ class BlurConv2d(nn.Module):
 
     # based partially on https://pytorch.org/docs/stable/_modules/torch/nn/modules/conv.html#Conv2d
 
-    def __init__(self,
-                 in_channels: int,
-                 out_channels: int,
-                 kernel_size: _size_2_t,
-                 stride: _size_2_t = None,
-                 padding: _size_2_t = 0,
-                 dilation: _size_2_t = 1,
-                 groups: int = 1,
-                 bias: bool = True,
-                 blur_first: bool = True):
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: _size_2_t,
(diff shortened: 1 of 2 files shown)


dblalock (approved):
I'd rather see at least one unit test for the idempotence this is supposed to achieve, but I'll leave it up to you whether that goes in this PR or a subsequent one.

mvpatel2000:
Tested and it works :)

mvpatel2000:
> I'd rather see at least one unit test for the idempotence this is supposed to achieve, but I'll leave it up to you whether that goes in this PR or a subsequent one.

Will make unit tests for all surgery algos next week. Want fix in before v11 release EOD tmrw and I have other bugs I need to fix by then as well :(

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -220,16 +220,18 @@ class BlurConv2d(nn.Module):
...
-                 in_channels: int,
-                 out_channels: int,
-                 kernel_size: _size_2_t,
-                 stride: _size_2_t = None,
-                 padding: _size_2_t = 0,
-                 dilation: _size_2_t = 1,
-                 groups: int = 1,
-                 bias: bool = True,
-                 blur_first: bool = True):
+    def __init__(
```
dskhudia, line 223: Have we moved to this style of formatting? Is this done automatically?

dskhudia (approved):

composer/algorithms/blurpool/blurpool_layers.py
mvpatel2000: So... we're technically unopinionated on how we format these things. You can either have a trailing comma or not have it. I _personally_ prefer the the new format, and have some linters that do things like that for me. Must have snuck through, not removing since we're ok with both formats.

PR #1627, opened 2022-10-14
add stop method

hanlint:
# What does this PR do?

Adds a `state.stop_training()` method for algorithms and callbacks to gracefully stop training (finishing the rest of the events in the current batch and any scheduled evals and epoch ends). The EarlyStopping and MLPerf callback now uses this interface instead of setting the max duration itself.

# Before submitting

### DIFF
--- a/composer/callbacks/early_stopper.py
+++ b/composer/callbacks/early_stopper.py
@@ -134,10 +134,10 @@ class EarlyStopper(Callback):
         assert self.best_occurred is not None
         if self.patience.unit == TimeUnit.EPOCH:
             if state.timestamp.epoch - self.best_occurred.epoch > self.patience:
-                state.max_duration = state.timestamp.batch
+                state.stop_training()
         elif self.patience.unit == TimeUnit.BATCH:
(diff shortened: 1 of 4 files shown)


dakinggg (approved):
LGTM

mvpatel2000 (approved):
It feels a little weird for this to be a state method. I'd expect it to be more like a helper function that accepts a state object as an argument. 

I don't have strong opinions on this though (just a weird vibe lol) and I'm fine with it as is. I don't really have a concrete place I'd want to put it otherwise either, so lgtm.

hanlint:
I agree having it in state is a little weird. Its the indirection --> a method on state stops training but only when coupled with the trainer. `trainer.stop()` is most intutitive, but not accessible from our callbacks and algorithms, so at a bit of loss there. 

I could rename it from `state.stop_training()` to `state.set_duration_to_now` or something but that's pretty verbose.

mvpatel2000:
> I could rename it from `state.stop_training()` to `state.set_duration_to_now` or something but that's pretty verbose.
Let's leave as is. I don't love it, but its good enough imo. A bit weird, but I'd rather make it easy to use

PR #1632, opened 2022-10-14
Check CUDA Retries

mvpatel2000:
# What does this PR do?

Increases grad_accum if there are cuda alloc retries going on. This causes throughput degradation if we're seeing this.

[CO-1159](https://mosaicml.atlassian.net/browse/CO-1159)


mvpatel2000:
It’s really hard to get a batch size where we get multiple retries for malloc but no Cuda OOM. I was able to get a few runs where retry forces auto grad accum to 2x at batch 1 but it’s able to keep running without OOMs for 10-20 batches otherwise without OOMing. In this scenario, running the lower grad accum value up until the OOM resulted in higher throughput. 

I think this would probably work if we just barely lowered the batch size so that it would fit more cleanly into memory, but 2xing grad accum is worse – the lower microbatch size gives worse performance, which dominates any improvement from avoiding multiple malloc calls.

PR #1633, opened 2022-10-15
Downgrade dataset

mvpatel2000:
# What does this PR do?

Upgrade is breaking a few things. We need to revert https://github.com/mosaicml/composer/pull/1414


dskhudia (approved):
Approving to unblock

mvpatel2000:
Closing. This won't happen because streaming depends on 2.4.0

PR #1635, opened 2022-10-16
Remove c4 dataset

mvpatel2000:
# What does this PR do?

Removes C4 dataset. This is currently broken with `datasets` upgrade. We recommend using streaming datasets anyways, so we're just going to get rid of `datasets` dependency and the `C4Dataset` in Composer.

[CO-1271](https://mosaicml.atlassian.net/browse/CO-1271)


hanlint (approved):

mvpatel2000:
@abhi-mosaic  can you please verify dataloader changes to the gpt yamls

composer/datasets/c4_hparams.py
```diff
@@ -36,8 +35,12 @@ class StreamingC4Hparams(DatasetHparams):
         group_method (str): How to group text samples into token samples. Currently only `truncate` is supported.
         mlm (bool): Whether or not to use masked language modeling. Default: ``False``.
         mlm_probability (float): If ``mlm==True``, the probability that tokens are masked. Default: ``0.15``.
+        shuffle (bool): Whether to shuffle the samples in the dataset. Currently, shards are assigned and consumed with
+            deterministic per-device shard order, but shuffling affects the order of samples via (per-device) shuffle
+            buffers. Default: ``False``.
```
abhi-mosaic, line 40: This needs to be reworded as StreamingDataset is not deterministic yet. Could you ask @knighton for the best way to reword this description here?

composer/datasets/c4_hparams.py
```diff
@@ -53,8 +56,12 @@ class StreamingC4Hparams(DatasetHparams):
         'How to group text samples into token samples. Currently only `truncate` is supported.', default='truncate')
     mlm: bool = hp.optional('Whether or not to use masked language modeling.', default=False)
     mlm_probability: float = hp.optional('If `mlm==True`, the probability that tokens are masked.', default=0.15)
+    shuffle: bool = hp.optional(
+        'Whether to shuffle the samples in the dataset. Currently, shards are assigned and consumed with deterministic per-device shard order, but shuffling affects the order of samples via (per-device) shuffle buffers.',
+        default=True)
```
abhi-mosaic, line 61: Same thing here

composer/yamls/models/gpt3_125m.yaml
```diff
@@ -1,21 +1,23 @@
 train_dataset:
-  c4:
+  streaming_c4:
+    version: 2
+    remote: s3://mosaicml-internal-dataset-c4/mds/2/
+    local: /tmp/mds-cache/mds-c4/
     split: train
-    num_samples: 5120000 # 10e9 tok ~= 256[bs] * 20000[ba] * 2048[msl] = 5120000[sa] * 2048[msl]
```
abhi-mosaic, line 4: Confirming that you can remove this value here, but now you have to move this to `max_duration`, otherwise these training runs will be very, very long. Would recommend setting something like `max_duration: 20000ba`

abhi-mosaic (changes requested):
Added a few comments, once changes are in happy to re-review and merge!

composer/datasets/c4_hparams.py
knighton: "Whether to iterate over the samples in randomized order"?

composer/yamls/models/gpt3_125m.yaml
mvpatel2000: @knighton  is it possible to specify max_duration w streaming?

knighton: We don't currently support `max_duration` or such params, open to discussing offline? cc: @karan6181

abhi-mosaic: @mvpatel2000 sorry for the confusion, all I meant was that the **Trainer's max_duration** must now be updated from `1ep` to something like `20000ba` because there is no longer an epoch length.

Also, we will need to update the source code for `StreamingC4` to support `concat`, which I can help with.

knighton (approved):

abhi-mosaic (approved):
Looks good, thanks for taking care of this. Trusting @mvpatel2000 on launching the YAMLs to make sure that nothing breaks.

mvpatel2000:
Unit tests verify it parses correctly. I'm skipping tests checking that it actually runs since cluster is full, and these example YAMLs are all for a deprecated codepath for which we will be dropping support after the next release.

mvpatel2000:
Ah nevermind I just tested it because I felt bad. It works. woohoo.

PR #1644, opened 2022-10-18
Changed default logging to WARN for doctests

eracah:
# What does this PR do?

Changes to logging level to WARN for doctests. This reduces the verbosity of doctests and makes it easier to see your error.

CO-1285

# Before submitting

### DIFF
--- a/docs/source/doctest_fixtures.py
+++ b/docs/source/doctest_fixtures.py
@@ -10,6 +10,9 @@ The script is run before any doctests are executed,
 so all imports and variables are available in any doctest.
 The output of this setup script does not show up in the documentation.
 """
+import logging
+
+logging.basicConfig(level=logging.WARN)
 import os
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


karan6181:
Changes looks good to me. I have one question though, is it possible to customize logging based on the command-line parameter? for example, if user pass `-v` flag then the logging defaults to `DEBUG` and if user doesn't pass `-v` flag then it remain at `WARNING` level.

docs/source/doctest_fixtures.py
```diff
@@ -10,6 +10,9 @@
 so all imports and variables are available in any doctest.
 The output of this setup script does not show up in the documentation.
 """
+import logging
+
+logging.basicConfig(level=logging.WARN)
```
dakinggg, line 15: wait isn't the python default warn? is somewhere else setting it to info?

eracah: Not sure, but when you run doctests, it prints INFO statements, which clutter up stdout and stderr.

eracah:
> Changes looks good to me. I have one question though, is it possible to customize logging based on the command-line parameter? for example, if user pass `-v` flag then the logging defaults to `DEBUG` and if user doesn't pass `-v` flag then it remain at `WARNING` level.

I don't think so unless sphinx lets you add custom command line arguments.

PR #1645, opened 2022-10-18
Adjust speed monitor

mvpatel2000:
# What does this PR do?

Adjusts speed monitor to start tracking at `BEFORE_DATALOADER` instead of `BEFORE_BATCH`. This resolves issues with dataloader bottlenecks being potentially masked. Note that currently it should still work since data transfer is done async, but to be safe, we should start the clock on `BEFORE_DATALOADER`.


dskhudia (approved):

dskhudia:
Did you run a benchmark with it to see if it changed the numbers?

PR #1646, opened 2022-10-18
Fix many failing notebook tests

dakinggg:
# What does this PR do?
Fixes a bunch of failing notebook tests, specifically:
`huggingface_models.ipynb`: un xfails this test, it passes for me locally
`custom_speedup_methods.ipynb`: Fixes usage of %%time (has to be the first line of a cell)
`exporting_for_inference.ipynb`: The export for inference callback did not properly move all input types to cpu, and so was failing. Also adds a new gpu unittest that would've caught the issue.
`auto_grad_accum.ipynb`: Skip this on cpu. It fails when I run it via pytest with a CUDA error, but works when I run it as a notebook. So not sure whether it will pass in CI or not.
`migrate_from_ptl.ipynb`: Added some required functions for eval/metrics
`streaming_dataloader_facesynthetics.ipynb`: Pin version of `mmcv` and `mmcv-full`
`TPU_Training_with_composer.ipynb`: skip because it needs tpu
`training_without_local_storage.ipynb`: Lots of fixes related to recent checkpointing changes. Ran locally to ensure it works, but not entirely sure it will work in CI.

Note: The test whose tolerance I changed actually didn't exist before because I added the gpu device for that test. But let me know if that tolerance seems like an issue.

Fixes [CO-1278](https://mosaicml.atlassian.net/browse/CO-1278)

# Before submitting


composer/utils/inference.py
```diff
@@ -57,6 +57,29 @@ class ExportFormat(StringEnum):
     ONNX = 'onnx'
 
 
+def _move_sample_input_to_device(sample_input: Union[torch.Tensor, dict, list, Tuple],
```
hanlint, line 60: we do have `map_collection` if that helps (https://github.com/mosaicml/composer/blob/dev/composer/utils/iter_helpers.py#L13) here

hanlint (commented):
wow, thanks for fixing these!

composer/utils/inference.py
dakinggg: I think the recursion in this function actually matters, which makes it a bit trickier to use the helper, but definitely useful to know about.

composer/utils/inference.py
```diff
@@ -130,15 +153,11 @@ def export_for_inference(
...
         sample_input = ensure_tuple(sample_input)
-        for i in range(len(sample_input)):
-            if isinstance(sample_input[i], torch.Tensor):
-                sample_input[i] = sample_input[i].to(cpu)  # type: ignore
-            elif isinstance(sample_input[i], dict):
-                for key, value in sample_input[i].items():
-                    if isinstance(value, torch.Tensor):
-                        sample_input[i][key] = value.to(cpu)
+        sample_input = _move_sample_input_to_device(sample_input, cpu)
+        assert isinstance(sample_input, tuple)
```
nik-mosaic, line 160: Asserts are helpful for tests, less useful for code. Remove since this should always be true given the ensure_tuple above and code in _move_sample_to_device.

nik-mosaic (approved):
I only verified the inference code, which looks good to me. Thanks for fixing this.

composer/utils/inference.py
dakinggg: Ah sorry, I forgot to leave a comment saying this is for pyright. Will add

dakinggg: Ok, I failed at getting pyright to accept my use of generics to try to type this function better, but I moved the assert to where the pyright error comes from and added a comment.

dakinggg:
@mvpatel2000 bump to review when you get a chance

composer/utils/inference.py
```diff
@@ -57,6 +57,29 @@ class ExportFormat(StringEnum):
     ONNX = 'onnx'
 
 
+def _move_sample_input_to_device(sample_input: Optional[Union[torch.Tensor, dict, list, Tuple]],
+                                 device: torch.device) -> Optional[Union[torch.Tensor, dict, list, Tuple]]:
+    """Handle moving sample_input of various types to a device. If possible, avoids creating copies of the input."""
+    output = None
+    if isinstance(sample_input, torch.Tensor):
+        output = sample_input.to(device)
```
mvpatel2000, line 65: Nit: to be consistent with the rest of `Composer`, can we create `DeviceCPU` class and instead use `DeviceCPU.to(tensor_to_device)`?

composer/utils/inference.py
```diff
@@ -130,15 +153,10 @@ def export_for_inference(
...
     if sample_input is not None:
         sample_input = ensure_tuple(sample_input)
-        for i in range(len(sample_input)):
-            if isinstance(sample_input[i], torch.Tensor):
-                sample_input[i] = sample_input[i].to(cpu)  # type: ignore
-            elif isinstance(sample_input[i], dict):
-                for key, value in sample_input[i].items():
-                    if isinstance(value, torch.Tensor):
-                        sample_input[i][key] = value.to(cpu)
+        sample_input = _move_sample_input_to_device(sample_input, cpu)
```
mvpatel2000, line 159: same comment on constructing `DeviceCPU`

tests/utils/test_inference.py
```diff
@@ -254,6 +256,9 @@ def test_export_for_inference_onnx(model_cls, sample_input):
     model = model_cls()
     model.eval()
 
+    torch_device = 'cuda' if device == 'gpu' else 'cpu'
+    sample_input = (sample_input[0].to(torch_device), sample_input[1].to(torch_device))
+    model.to(torch_device)
```
mvpatel2000, line 261: same comment on `DeviceCPU`/`DeviceGPU`

tests/utils/test_inference.py
```diff
@@ -271,15 +276,15 @@ def test_export_for_inference_onnx(model_cls, sample_input):
         ort_session = ort.InferenceSession(save_path)
         loaded_model_out = ort_session.run(
             None,
-            {'input': sample_input[0].numpy()},
+            {'input': sample_input[0].cpu().numpy()},
```
mvpatel2000, line 279: same comment

mvpatel2000 (approved):

PR #1647, opened 2022-10-18
Attempt to fix flaky doctest

dakinggg:
# What does this PR do?
The doctests are extremely difficult to debug. The logs are different every time I run them locally, get swallowed when the doctest errors, and the offending doctest only fails some of the time. I couldn't figure it out, so I basically removed the offending doctest since this is blocking all CI. I can't be certain, but I believe the issue is limited to the doctests, because the normal tests pass, and I ran the doctest code in an ipython outside of the doctests and it worked consistently. If someone feels like they have a better handle on the doctests, they may want to try to figure this out.

Fixes [CO-1290](https://mosaicml.atlassian.net/browse/CO-1290)

### DIFF
--- a/docs/source/notes/resumption.rst
+++ b/docs/source/notes/resumption.rst
@@ -14,7 +14,7 @@ Resuming from checkpoints is commonly used to recover from hardware failures (e.
         model=model,
         train_dataloader=train_dataloader,
         max_duration="1ep",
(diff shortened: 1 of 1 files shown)


hanlint (approved):
I'm OK with this, the autoresume is more strictly tested in our unit testing, and the current testcode will make sure the API is correct.

PR #1652, opened 2022-10-19
Add Event.AFTER_LOAD

mvpatel2000:
# What does this PR do?

Adds a new event. This is required for weight-preserving surgery algorithms.

# What issue(s) does this change relate to?

[CO-1151](https://mosaicml.atlassian.net/browse/CO-1151)

PR #1653, opened 2022-10-19
Lazy Logging +  pretty print dict for hparams

eracah:
# What does this PR do?

Lazily logs hyperparameters to console and pretty prints them exactly like the yahp entrypoint does.

Before:

```
[hyperparameter]: seed: 42


[hyperparameter]: deterministic_mode: False


[hyperparameter]: dist_timeout: 300.0000
```

After:
```
Config:
algorithms: null
autoresume: false
callbacks:
  lr_monitor: {}
  speed_monitor:
    window_size: 100
console_stream: stderr
dataloader:
  num_workers: 8
  persistent_workers: true
  pin_memory: true
  prefetch_factor: 2
  timeout: 0.0
ddp_sync_strategy: null
deepspeed_config: null
deterministic_mode: false
device:
  gpu:
    allow_tf32: true
    device_id: null
dist_timeout: 300.0
eval_batch_size: auto
eval_interval: 1ep
eval_subset_num_batches: -1
evaluators: null
grad_accum: auto
grad_clip_norm: -1.0
load_ignore_keys: null
load_logger_destination: null
load_object_store: null
load_path: null
load_progress_bar: true
load_weights_only: false
log_to_console: null
loggers:
  object_store:
    bucket_uri: s3://mosaicml-internal-checkpoints-demo
    file_path_format_string: '{remote_file_name}'
    num_attempts: 3
    num_concurrent_uploads: 1
    object_store_hparams: null
    upload_staging_folder: null
    use_procs: true
max_duration: 1ep
model:
  resnet:
    groups: 1
    initializers:
    - KAIMING_NORMAL
    - BN_UNIFORM
    - LINEAR_LOG_CONSTANT_BIAS
    loss_name: binary_cross_entropy_with_logits
    model_name: resnet50
    num_classes: 1000
    pretrained: false
    weights: null
    width_per_group: 64
num_gpus_per_node: 8
num_nodes: 1
optimizers:
  decoupled_sgdw:
    dampening: 0.0
    lr: 2.048
    momentum: 0.875
    nesterov: false
    weight_decay: 0.0005
precision: AMP
profiler: null
progress_bar: true
python_log_level: INFO
rank_zero_seed: 42
run_name: resnet-50-demo-autoresume-27sm
save_filename: ep{epoch}-ba{batch}-rank{rank}.pt
save_folder: null
save_interval: 10ba
save_latest_filename: latest-rank{rank}.pt
save_num_checkpoints_to_keep: -1
save_overwrite: false
save_weights_only: false
scale_schedule_ratio: 1.0
schedulers:
  cosine_decay_with_warmup:
    alpha_f: 0.0
    scale_warmup: false
    t_max: 1dur
    t_warmup: 8ep
seed: 42
step_schedulers_every_batch: null
train_batch_size: 2048
train_dataloader_label: train
train_dataset:
  streaming_imagenet1k:
    crop_size: 176
    drop_last: true
    local: /tmp/mds-cache/mds-imagenet1k/
    remote: s3://mosaicml-internal-dataset-imagenet1k/mds/1/
    resize_size: -1
    shuffle: true
    split: train
    version: 1
train_subset_num_batches: -1
val_dataset:
  streaming_imagenet1k:
    crop_size: 224
    drop_last: false
    local: /tmp/mds-cache/mds-imagenet1k/
    remote: s3://mosaicml-internal-dataset-imagenet1k/mds/1/
    resize_size: 256
    shuffle: false
    split: val
    version: 1
```

fixes CO-1250

# Before submitting

### DIFF
--- a/composer/loggers/progress_bar_logger.py
+++ b/composer/loggers/progress_bar_logger.py
@@ -16,6 +16,7 @@ from composer.core.time import Timestamp, TimeUnit
 from composer.loggers.logger import Logger, format_log_data_value
 from composer.loggers.logger_destination import LoggerDestination
 from composer.utils import dist, is_notebook
+import yaml
 
 __all__ = ['ProgressBarLogger']
 
@@ -163,6 +164,8 @@ class ProgressBarLogger(LoggerDestination):
 
         self.stream = stream
         self.state: Optional[State] = None
+        self.hparams: Dict[str, Any] = {}
+        self.hparams_already_logged_to_console: bool = False
 
     @property
     def show_pbar(self) -> bool:
@@ -175,11 +178,22 @@ class ProgressBarLogger(LoggerDestination):
                 self._log_to_console(f'[trace]: {trace_name}:' + trace_str + '\n')
 
     def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
+        # Lazy logging of hyperparameters
+        self.hparams.update(hyperparameters)
+        # self.hparams.update(hyperparameters)
+        # for hparam_name, hparam in hyperparameters.items():
+        #     hparam_str = format_log_data_value(hparam)
+        #     log_str = f'[hyperparameter]: {hparam_name}: {hparam_str}'
+        #     self._log_to_console(log_str)
+
+    def _log_hparams_to_console(self):
         if self.should_log_to_console or self._show_pbar:
-            for hparam_name, hparam in hyperparameters.items():
-                hparam_str = format_log_data_value(hparam)
-                log_str = f'[hyperparameter]: {hparam_name}: {hparam_str}'
-                self._log_to_console(log_str)
+            if dist.get_local_rank() == 0:
+                self._log_to_console('*' * 30)
+                self._log_to_console('Config:')
+                self._log_to_console(yaml.dump(self.hparams))
+                self._log_to_console('*' * 30)
+
 
     def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
         for metric_name, metric_value in metrics.items():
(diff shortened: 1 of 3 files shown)


composer/loggers/progress_bar_logger.py
```diff
@@ -305,11 +313,19 @@ def init(self, state: State, logger: Logger) -> None:
...
+            self._log_hparams_to_console()
+
     def epoch_start(self, state: State, logger: Logger) -> None:
         if self.show_pbar and not self.train_pbar:
             self.train_pbar = self._build_pbar(state=state, is_train=True)
 
     def eval_start(self, state: State, logger: Logger) -> None:
+        if not self.hparams_already_logged_to_console:
+            self.hparams_already_logged_to_console = True
+            self._log_hparams_to_console()
```
mvpatel2000, line 328: Should this be on rank 0 only?

Also, can we do the same for `predict_start`

mvpatel2000 (commented):
Can you explain why we need to do this lazily at start of fit/eval/predict instead of end of init?

composer/loggers/progress_bar_logger.py
eracah: it is rank_zero only. See `_log_hparams_to_console` function.

mvpatel2000: > > Can you explain why we need to do this lazily at start of fit/eval/predict instead of end of init?
> 
> Because we want them all printed out in a block all at once and the hyperparameters come in at different times (i.e. `log_hyperparameters` is called multiple times). Also, there is no INIT_END event (there should be!)

What about `Event.AFTER_LOAD`? https://github.com/mosaicml/composer/pull/1652

eracah: ok added `predict_start` support

mvpatel2000: Can we move to `Event.AFTER_LOAD` instead? or do you not like that. Seems cleaner to actually give the hyperparameters after calling init and less duplicated code

eracah: > What about `Event.AFTER_LOAD`? #1652

Wow timely PR merge. The issue is that people may do this:
```
args = parser.parse_args()
trainer = Trainer(model=model, ... )
trainer.logger.log_hyperparameters(**vars(args))
trainer.fit()
```
So logging of the hyperparameters will happen after `__init__` returns

eracah: also, see https://github.com/mosaicml/composer/blob/dev/examples/run_composer_trainer.py#L47

mvpatel2000: Hm... ok. Feels a little hacky, but agree with you.

mvpatel2000 (approved):

composer/loggers/progress_bar_logger.py
eracah: yeah well very soon we will have auto-hparam logging, so we won't have that pattern.

PR #1658, opened 2022-10-21
Autoload algorithms

mvpatel2000:
# What does this PR do?

Surgery algorithms are complicated, messy, and easy to screw up. Now, we
1. do a best effort attempt to autocreate and apply algorithms which have `required_on_load=True`
2. in the case of failure, give users helpful warnings to give better guidance for fixing their runs
3. give an option to disable autoloading if they want to forcibly remove a surgery algo

Correspondingly, removes old tests that raise ValueError and adds new tests for the following scenarios

[CO-1151](https://mosaicml.atlassian.net/browse/CO-1151)


composer/trainer/trainer.py
```diff
@@ -622,6 +622,17 @@ class Trainer:
             If a callable, it should take one argument which is the state_dict. The callable is free to arbitrarily modify
             the state_dict before it is loaded.
 
+            (default: ``None``)
+        load_exclude_algorithms (List[str], optional): A list of algorithm names to exclude from loading.
```
dskhudia, line 626: `exclude_algorithms_on_load`? or is it too long?

composer/core/state.py
```diff
@@ -540,14 +543,91 @@ def state_dict(self) -> Dict[str, Any]:
...
+            except AttributeError:
+                log.warning(
+                    f'Found unknown algorithm {algo_name}. Skipping check for if it is required when loading checkpoint.'
+                )
+
+        try:
+            for algo, algo_repr in missing_algorithms:
+                if algo.match(Event.INIT, self):
+                    algo.apply(Event.INIT, self, logger)
+                self.algorithms.append(algo)
```
dskhudia, line 588: Do we need care about order here? How does engine interact/reorder now?

composer/trainer/trainer.py
mvpatel2000: would prefer to prefix with `load_` since all loading relating arguments are prefixed similarly

tests/algorithms/test_required_on_load.py
```diff
@@ -97,3 +100,63 @@ def test_idempotent(algo_name: str):
...
+        checkpoint_saver = [cb for cb in trainer1.state.callbacks if isinstance(cb, CheckpointSaver)][0]
+        checkpoint_saver._save_checkpoint(trainer1.state, trainer1.logger)
+
+        context = contextlib.nullcontext()
+        # Emit warning when autoloading
+        if not already_added and not exclude:
+            context = pytest.warns(UserWarning, match='Automatically adding required_on_load algorithm*')
+        # Excluding some algorithms leads to errors when loading
+        elif exclude:
+            if algo_name in ['Factorize', 'SqueezeExcite']:
```
dakinggg, line 139: Is it possible to get nicer errors for these two cases? or its not possible to know something is wrong until you hit the error?

composer/core/state.py
```diff
@@ -540,14 +546,87 @@ def state_dict(self) -> Dict[str, Any]:
...
+        for algo_name, serialized_value in state['algorithms']:
+            if hasattr(algorithms, algo_name) and getattr(algorithms, algo_name).required_on_load():
+                algo = eval(f"algorithms.{serialized_value['repr']}")
+                if exclude_algorithms is None or type(algo).__qualname__ not in exclude_algorithms:
+                    if type(algo) in state_algos and not serialized_value['repr'] in state_algos[type(algo)]:
+                        warnings.warn(
+                            textwrap.dedent(
+                                f"required_on_load algorithm {serialized_value['repr']} was enabled when training the "
+                                f"loaded checkpoint but is now specified in the following forms: {', '.join(state_algos[type(algo)])}."
+                                'Potential paarameter discrepencies for this required_on_load algorithm may lead to '
```
dakinggg, line 582: "paarameter" -> "parameter"

dakinggg, line 582: "discrepencies" -> "discrepancies"

composer/core/state.py
```diff
@@ -540,14 +546,87 @@ def state_dict(self) -> Dict[str, Any]:
...
+                if type(algo) not in state_algos:
+                    state_algos[type(algo)] = []
+                state_algos[type(algo)].append(algo.__repr__())
+
+        missing_algorithms = []
+        for algo_name, serialized_value in state['algorithms']:
+            if hasattr(algorithms, algo_name) and getattr(algorithms, algo_name).required_on_load():
+                algo = eval(f"algorithms.{serialized_value['repr']}")
+                if exclude_algorithms is None or type(algo).__qualname__ not in exclude_algorithms:
+                    if type(algo) in state_algos and not serialized_value['repr'] in state_algos[type(algo)]:
```
dakinggg, line 577: Could you explain when this warning would get hit?

mvpatel2000: For example, I add `BlurPool` to replace convs and pooling. In a second run, I add BlurPool but only replace convs. It would then hit this warning

tests/algorithms/test_required_on_load.py
mvpatel2000: Not really, you hit some weird stuff like optimizer mismatches params bc the surgery is missing

mvpatel2000: If you explicitly exclude the algorithm and then hit the error, I feel like thats kinda on you

dakinggg: yeah, thats fair :+1:

composer/core/state.py
mvpatel2000: @dskhudia I've added
1. reordering based on registered engine passes
2. error if this leads to an autoapply algorithm being sorted before an already applied algorithm

composer/core/state.py
```diff
@@ -579,26 +586,46 @@ def _apply_required_algorithms(
...
-                        missing_algorithms.append((algo, serialized_value['repr']))
+                        missing_algos.add(algo)
+                        missing_algo_names.append(algo_name)
+                        missing_algo_reprs.append(serialized_value['repr'])
+                        self.algorithms.append(algo)
+
+        # Reorder algorithms based on algorithm_passes from engine
+        algorithms = self.algorithms
+        if algorithm_passes is not None:
+            for passes in algorithm_passes:
```
dakinggg, line 600: nit: "passes" -> "pass"

dskhudia (approved):

composer/core/state.py
```diff
@@ -579,26 +586,46 @@ def _apply_required_algorithms(
...
+                algorithms = passes(algorithms, Event.INIT)
+        # Raise ValueError if algorithm_passes order any checkpoint algorithm is ordered before an
+        # already applied user specified algorithm
+        encountered_ckpt_algo = False
+        for algo in algorithms:
+            if algo in missing_algos:
+                encountered_ckpt_algo = True
+            elif encountered_ckpt_algo:
+                raise ValueError(
+                    textwrap.dedent(
```
dakinggg, line 610: This error should be something else I think? no exception was encountered here

composer/core/state.py
```diff
@@ -579,26 +586,46 @@ def _apply_required_algorithms(
...
+                        missing_algo_names.append(algo_name)
+                        missing_algo_reprs.append(serialized_value['repr'])
+                        self.algorithms.append(algo)
+
+        # Reorder algorithms based on algorithm_passes from engine
+        algorithms = self.algorithms
+        if algorithm_passes is not None:
+            for passes in algorithm_passes:
+                algorithms = passes(algorithms, Event.INIT)
+        # Raise ValueError if algorithm_passes order any checkpoint algorithm is ordered before an
```
dakinggg, line 602: "order any checkpoint algorithm is ordered before" -> "algorithm_passes order any checkpoint algorithm before"

dakinggg (approved):
LGTM

bcui19 (approved):
LGTM! As a question, from memory there are algorithms that need to be applied before others, that would then be on the user to put them in the correct order?

mvpatel2000:
> LGTM! As a question, from memory there are algorithms that need to be applied before others, that would then be on the user to put them in the correct order?

Yep, `FusedDenseGeluDense` comes before `FusedDropoutAddLayerNorm`. If you specify `FusedDropoutAddLayerNorm`, we can't autoapply `FusedDenseGeluDense`.

Note that the engine compilation passes handle reordering for the user, but this can only happen if the algorithms are all user specified (so we can sort and run on `Event.INIT`) or all non-user specified (so we can sort and run here when autoapplying).

composer/core/state.py
```diff
@@ -540,14 +547,117 @@ def state_dict(self) -> Dict[str, Any]:
 
         return state_dict
 
-    def load_model_state(self, state_dict: Dict[str, Any], strict: bool):
+    def _apply_required_algorithms(
+        self,
+        state: Dict[str, Any],
```
eracah, line 552: Any chance we could call this `state_dict`? I think that will add clarity because it is a bit more specific and `state` is already the name of this module and the name of the class

composer/core/state.py
```diff
@@ -540,14 +547,117 @@ def state_dict(self) -> Dict[str, Any]:
...
+                if type(algo) not in state_algos:
+                    state_algos[type(algo)] = []
+                state_algos[type(algo)].append(algo.__repr__())
+
+        # Gather algorithms to apply
+        missing_algos = set()
+        missing_algo_names = []
+        missing_algo_reprs = []
+        for algo_name, serialized_value in state['algorithms']:
+            if hasattr(algorithms, algo_name) and getattr(algorithms, algo_name).required_on_load():
```
eracah, line 581: any chance you can maybe add a small comment above some of these if blocks to clarify exactly what the case is and what you are doing in that case? I find the triple nested if statements to be a bit hard to follow.

composer/core/state.py
```diff
@@ -540,14 +547,117 @@ def state_dict(self) -> Dict[str, Any]:
...
+            if algo.required_on_load():
+                if type(algo) not in state_algos:
+                    state_algos[type(algo)] = []
+                state_algos[type(algo)].append(algo.__repr__())
+
+        # Gather algorithms to apply
+        missing_algos = set()
+        missing_algo_names = []
+        missing_algo_reprs = []
+        for algo_name, serialized_value in state['algorithms']:
```
eracah, line 580: I find some of this naming a bit confusing .`state_algos` is the algorithms passed into the `Trainer` and `state['algorithms']` are the algorithms in the  `state_dict` of the checkpoint? Any way to make this a bit clearer with variable naming or comments?

composer/core/state.py
```diff
@@ -540,14 +547,117 @@ def state_dict(self) -> Dict[str, Any]:
...
+                                'Potential parameter discrepancies for this required_on_load algorithm may lead to '
+                                'unexpected behavior, including failing to load weights for some layers.'))
+                    elif type(algo) not in state_algos:
+                        missing_algos.add(algo)
+                        missing_algo_names.append(algo_name)
+                        missing_algo_reprs.append(serialized_value['repr'])
+                        self.algorithms.append(algo)
+
+        # Reorder algorithms based on algorithm_passes from engine
+        algorithms = self.algorithms
```
eracah, line 598: The `algorithms` module defined above: `import composer.algorithms as algorithms` already has this name. Any chance you could rename that or this in order to disambiguate and avoid naming conflicts errors?

composer/core/state.py
```diff
@@ -540,14 +547,118 @@ def state_dict(self) -> Dict[str, Any]:
...
+        encountered_ckpt_algo = False
+        for algo in algorithms:
+            if algo in missing_algos:
+                encountered_ckpt_algo = True
+            elif encountered_ckpt_algo:
+                raise ValueError(
+                    textwrap.dedent('The following algorithms were enabled when training this checkpoint '
+                                    f'and are required to successfully load it: {missing_algo_reprs}. '
+                                    'Attempted to autocreate and apply required algorithms, but at least one '
+                                    'of the loaded algorithms was ordered before a user specified algorithm, '
```
eracah, line 613:
```suggestion
                                    'of the loaded algorithms was ordered before a user specified algorithm, which has already been applied, '
```

eracah (commented):
Nice work, Mihir! Well done navigating all the edge cases. Mostly looks good. Just a few nits about naming. The names `state` and `algorithms` are overused a bit in `state.py` and could cause confusion and even naming conflict errors.

PR #1663, opened 2022-10-23
Mid-epoch resumption [WIP]

knighton:
PR to accompany https://github.com/mosaicml/streaming/pull/37


composer/core/state.py
```diff
@@ -504,18 +546,38 @@ def fsdp_enabled(self):
...
+        dataset = self._dataset_of(self.train_dataloader)
+        if hasattr(dataset, 'state_dict'):
+            obj['train'] = dataset.state_dict()
+
+        for evaluator in self.evaluators:
+            dataset = self._dataset_of(evaluator.dataloader)
+            if hasattr(dataset, 'state_dict'):
+                st = dataset.state_dict()
+                st['sessions'] = []  # Don't save eval split's sessions, because we do not checkpoint during eval.
+                obj['eval'][evaluator.label] = st
```
mvpatel2000, line 564:
```suggestion
                obj['eval'][evaluator.label] = dataset.state_dict()
```
should be fine. When you call `dataset.state_dict()`, it'll never be in the middle of epoch

knighton: Something about switching to pre-incrementing epoch made this ugly

PR #1673, opened 2022-10-24
Add trainer arg for engine passes

mvpatel2000:
# What does this PR do?

As title. Also adds tests.

[CO-980](https://mosaicml.atlassian.net/browse/CO-980)


composer/trainer/trainer.py
```diff
@@ -433,6 +434,11 @@ class Trainer:
             no algorithms will be used. (default: ``None``)
 
             .. seealso:: :mod:`composer.algorithms` for the different algorithms built into Composer.
+        algorithm_passes ([AlgorithmPass | Tuple[AlgorithmPass, int] | Sequence[AlgorithmPass | Tuple[AlgorithmPass, int]], optional):
+            Optional list of passes to change order in which algorithms are applied. These passes are merged with already
+            specified passes in :class:`.Engine`. If ``None``, then no passes will be used. (default: ``None``)
```
dakinggg, line 439: Suggest being a little clearer here about what merging is happening.
"Optional list of passes to change the order in which algorithms are applied. These passes are merged with the default passes specific in :class:`.Engine`. If ``None``, no extra passes will be merged with the default passes.

composer/core/engine.py
```diff
@@ -159,14 +163,36 @@ class Engine():
         state (State): The initial :class:`.State` of the trainer. ``state`` will be modified in-place.
         logger (Logger): A :class:`.Logger` instance to be used for logging algorithm and callback
             specific metrics.
+        algorithm_passes ([AlgorithmPass | Tuple[AlgorithmPass, int] | Sequence[AlgorithmPass | Tuple[AlgorithmPass, int]], optional):
+            A sequence of algorithm_passes to be registered in Engine.
```
dakinggg, line 167: Could you make the documentation here more similar to what the trainer documentation says? That is, also mention here that the passes that get passed in are merged with the default passes.

dakinggg (approved):
Couple documentation requests, otherwise LGTM.

PR #1686, opened 2022-10-31
pin torch version

mvpatel2000:
# What does this PR do?

Pins torch version.


abhi-mosaic:
This might not be needed anymore based on Slack discussion

mvpatel2000:
> This might not be needed anymore based on Slack discussion

imo we should still pin something as core as torch to minor releases. Risk of things breaking is non-zero, and id rather check before we release. Open to changing though

PR #1689, opened 2022-11-02
Update glu check

mvpatel2000:
# What does this PR do?

GLU check should type, not id


composer/algorithms/gated_linear_units/gated_linear_units.py
```diff
@@ -99,7 +99,7 @@ def apply_gated_linear_units(model: torch.nn.Module,
             return
 
         # get the activation functions used
-        act_fns = {module.intermediate_act_fn for module in intermediate_modules}
+        act_fns = {type(module.intermediate_act_fn) for module in intermediate_modules}
```
eracah, line 102: if `act_fns` was of length zero before this change it will still be of length zero after right?

dskhudia (approved):

PR #1691, opened 2022-11-02
Update codeowners

mvpatel2000:
# What does this PR do?

- Add nik-mosaic to algo codeowners
- update CLI codeowners to eng team instead of just @jbloxham since he's busy with other stuff :) 
- adds karan to datasets (we shouldn't have single person owners in case of OOO)
- adds Daniel to logging (we shouldn't have single person owners in case of OOO)


bandish-shah (approved):
👍

PR #1692, opened 2022-11-02
Fix engine pass registration

mvpatel2000:
# What does this PR do?

Fixes edge case with types for registering passes

### DIFF
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -182,6 +182,11 @@ class Engine():
 
         self.algorithm_passes: List[passes.AlgorithmPass] = _get_default_passes()
         if algorithm_passes is not None:
+            # Wrap in list if not already a list or if it's a length 2 list specifying a single
+            # call to register_pass with type [AlgorithmPass, int]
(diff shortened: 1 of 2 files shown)


composer/core/engine.py
```diff
@@ -182,6 +182,11 @@ def __init__(
 
         self.algorithm_passes: List[passes.AlgorithmPass] = _get_default_passes()
         if algorithm_passes is not None:
+            # Wrap in list if not already a list or if it's a length 2 list specifying a single
```
dakinggg, line 185: Is there a particular reason you're changing it to allow a two element list?

dakinggg (approved):
LGTM, not sure I understand the reason for that change, but can't imagine it being harmful.

composer/core/engine.py
mvpatel2000: See extra test I added. If you pass args for a single call to register pass, it should wrap in a list. However, if you pass [algopass, idx]  instead of (algopass, idx), it will fail wrapping and error out. This just makes it slightly more robust

PR #1693, opened 2022-11-02
fix backwards compatability

mvpatel2000:
# What does this PR do?

Fixes backwards compatibility for checkpoints.

[CO-1350](https://mosaicml.atlassian.net/browse/CO-1350)


dakinggg (approved):
I guess we'll leave adding tests for this to the regression testing infra/plan

PR #1695, opened 2022-11-03
Add nccl env var

mvpatel2000:
# What does this PR do?

The timeout variable was being ignored before. We should actually use it by setting the appropriate env vars. From Pytorch Dist docs:

"""
– Timeout for operations executed against the process group. Default value equals 30 minutes. This is applicable for the gloo backend. For nccl, this is applicable only if the environment variable NCCL_BLOCKING_WAIT or NCCL_ASYNC_ERROR_HANDLING is set to 1.
"""

Also updates default dist_timeout to be same as what Pytorch has (30 min).

[CO-1356](https://mosaicml.atlassian.net/browse/CO-1356)


bandish-shah (approved):
👍

PR #1699, opened 2022-11-03
refactor devices

mvpatel2000:
# What does this PR do?

Moves Devices Folder

[CO-1340](https://mosaicml.atlassian.net/browse/CO-1340)


dakinggg (commented):
I think you forgot to actually add the `devices` folder?

mvpatel2000:
> I think you forgot to actually add the devices folder?

oops forgot to push

eracah (approved):
Siiick. Thanks, Mihir. LGTM. Cleans stuff up nicely

nik-mosaic (approved):

composer/devices/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 MosaicML Composer authors
```
hanlint, line 1: Can/should this just be a flat `composer/devices.py` file instead of a full blown module?

composer/utils/device.py
```diff
@@ -8,7 +8,7 @@
 import torch.cuda
```
hanlint, line 8: this is a short file.. should this be in `composer.devices` instead?

hanlint (approved):
LGTM, except consider simplifying the code by just havine a single `devices.py` file rather than a full blown module

composer/utils/device.py
mvpatel2000: Agree, will move

composer/devices/__init__.py
mvpatel2000: Might be worth it, though punting to different PR / discussion

composer/utils/device.py
mvpatel2000: Will do in separate PR -- it causes some messiness

PR #1702, opened 2022-11-04
Add docs for saving and loading checkpoints with GCS

eracah:
# What does this PR do?

add docs to `checkpointing.rst` for GCS saving and loading and add a docstring for `RemoteUploaderDownloader`

fixes CO-1386

# Before submitting
- [x] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.


docs/source/trainer/checkpointing.rst
```diff
@@ -379,11 +379,56 @@ There are a few additional trainer arguments which can be helpful to configure:
     will remove each checkpoint after it has been uploaded. Checkpoints are never deleted from object stores.
 
 
-Loading from Object Store
--------------------------
+Saving Checkpoints to Google Cloud Storage (GCS)
+------------------------------------------------
+
+To save checkpoints to GCS, make sure to create a :class:`.RemoteUploaderDownloader` instance
```
dakinggg, line 385: Maybe note that composer needs to be installed with `libcloud` extra deps?

docs/source/trainer/checkpointing.rst
```diff
@@ -379,11 +379,56 @@ There are a few additional trainer arguments which can be helpful to configure:
...
+    :skipif: not _LIBCLOUD_INSTALLED
+
+    from composer.loggers import RemoteUploaderDownloader
+    from composer.trainer import Trainer
+
+    remote_uploader_downloader = RemoteUploaderDownloader(
+        bucket_uri="libcloud://my-gcs-bucket",
+        backend_kwargs={
+            "provider": "google_storage",  # The Apache Libcloud provider name
+            "container": "my-gcs-bucket",  # The name of the cloud container (i.e. bucket) to use.
```
dakinggg, line 398: This arg shouldn't actually have to be provided, as it will be parsed from `bucket_uri`, but no harm leaving it in the example i think?

docs/source/trainer/checkpointing.rst
```diff
@@ -379,11 +379,56 @@ There are a few additional trainer arguments which can be helpful to configure:
...
+            "provider_kwargs": {  # The Apache Libcloud provider driver initialization arguments
+                'key': 'provider_key',  # The cloud provider key.
```
dakinggg, line 400: should we also have an example of how to provide the key and secret via environment variables? I believe that would be

```
RemoteUploaderDownloader(bucket_uri="libcloud://my-gcs-bucket",
    backend_kwargs={
        "provider": "google_storage",
        "container": "my-gcs-bucket",
        "key_environ": "KEY_ENV_VAR",
        "secret_environ": "SECRET_ENV_VAR"
    }
```

dakinggg (approved):
Left a few comments, but approving as is if you disagree with any.

bandish-shah (commented):
Should we also update the Logging and docstrings in `composer.loggers.RemoteUploaderDownloader` with info on GCP?  Or at least details on how to specify or know which providers are supported in the `backend_kwards` arg?

docs/source/trainer/checkpointing.rst
karan6181: I agree with Daniel's point. Is `key` and `secret` a mandatory argument? I think if it is a mandatory argument then user have to change their open source script to pass those key and secret since keeping it in open source world brings a security risk. Is there a way to fetch those keys during runtime by the  `libcloud` ? Like how it happens with AWS key. You add `aws_access_key_id` and `aws_secret_access_key` either in `~/.aws/credentials` file or add it to an instance via envrion variable and the code import those environ variable during runtime so that user doesn't have to provide key manually inside a code.

dakinggg: Sorry I wasn't totally clear, the code I gave above _should work today_ (but I haven't actually tried it). That is what is currently implemented for `LibcloudObjectStore`. The AWS stuff works even more smoothly because we have a specific `S3ObjectStore` that uses boto3, instead of the more generic libcloud.

dakinggg: also the user could just fetch the key/secret from env vars themselves, which would be essentially the same for them as passing the name of their env vars to us. e.g. `"key": os.environ["KEY_ENV_VAR"]` vs `"key_environ": "KEY_ENV_VAR"`. But both of those should be supported by our current implementation.

docs/source/trainer/checkpointing.rst
eracah: yeah?

docs/source/trainer/checkpointing.rst
eracah: added example with env variables

eracah:
> Should we also update the Logging and docstrings in `composer.loggers.RemoteUploaderDownloader` with info on GCP? Or at least details on how to specify or know which providers are supported in the `backend_kwards` arg?

done

PR #1707, opened 2022-11-04
Fix typing in EMA's _move_params_to_device()

coryMosaicML:
This PR cleans up the `_move_params_to_device()` function in EMA to account for `param_list` and `buffer_list` no longer existing after #1524. This was causing errors with Pyright 1.1.277. EMA functionality be identical after this change.


mvpatel2000 (approved):

bandish-shah (approved):
👍

PR #1716, opened 2022-11-08
Simple nlp tests

dakinggg:
# What does this PR do?
This PR does three things.
1. Adds a simple transformer classifier and a simple "text" (really just input ids, not using a tokenizer) classification dataset for testing purposes.
2. Adds a test for `HuggingfaceModel` using the above dataset to train, eval, and predict.
3. Removes the `model_inputs` from `HuggingfaceModel`

Fixes [CO-1396](https://mosaicml.atlassian.net/browse/CO-1396?atlOrigin=eyJpIjoiMmI0NmJiZjYzM2MxNDFlYTk0ZjA2NWVlM2E3NWE0MjIiLCJwIjoiaiJ9)
Relates to [CO-1110](https://mosaicml.atlassian.net/browse/CO-1110) and [CO-1358](https://mosaicml.atlassian.net/browse/CO-1358?atlOrigin=eyJpIjoiMTM3OTRiZTVkNzdhNDgxNzg1MjI3YjM4ZjU3ZDI0NjYiLCJwIjoiaiJ9)

# Before submitting


mvpatel2000 (commented):
I _think_ this looks good to me, but I would defer to NLP people to approve because I'm not really familiar with this. 

Only comment is it would be really nice to have follow on PRs that start adding more NLP tests into existing flows that only test vision. For example, export for inference should definitely tests this

dakinggg:
Agreed on both counts, waiting for other reviews, and we need a full NLP testing plan on top of this. This is just to set a base for that and help me more easily test huggingface stuff.

composer/models/huggingface.py
```diff
@@ -51,30 +52,20 @@ def __init__(self,
                  metrics: Optional[List[Metric]] = None) -> None:
         try:
             import transformers
+            del transformers  # unused
         except ImportError as e:
             raise MissingConditionalImportError(extra_deps_group='nlp', conda_package='transformers') from e
```
nik-mosaic, line 57: What is the point of this if we are not using the transformers import?

nik-mosaic (approved):
Looks like a good start for our NLP testing plan. Just one comment; other than that, looks good to merge!

alextrott16 (approved):
The changes to `huggingface.py` look good to me. As far as the tests go, the implementations seem reasonable but I'm not super clear on what the tests are designed to catch so it's hard to comment more.

composer/models/huggingface.py
dakinggg: Although we don't use the `transformers` import, you won't be able to use this class if `transformers` is not installed in your environment, because you have to pass in a model/tokenizer with a type from `transformers`. This error just makes it explicit, rather than crashing in some other less clear way. I'm not super attached to this pattern, but it follows a pattern from some other places in our code (e.g. https://github.com/mosaicml/composer/blob/658c4852880e53c0c26c3a5974d06a1532b3e6f1/composer/loggers/wandb_logger.py#L66-L73)

dakinggg: maybe in this case its better to just delete if/until this class actually needs access to the package

dakinggg: Ok, removed.

PR #1717, opened 2022-11-08
Remove coco and ssd

growlix:
# What does this PR do?

This PR removes COCO and SSD due to lack of use and maintenance. Deyahpification seemed like a good opportunity to pull the plug. From the [relevant ticket](https://mosaicml.atlassian.net/browse/RESEARCH-288):

>  I’m not sure if it would be worth the time to deyahpify COCO since it may not be easy to test. The only model that can use it is SSD, but no one has used this for sometime so it may not work. In the future, I think we want to move to pulling models from mmdetection which will likely require the COCO to be refactored to output a different data structure. Since we anticipate a refactor in the future, it may be worth delaying the deyahp until then. 

# Before submitting


A-Jacobson (approved):
I'm cool with it. 100% agree with what you quoted. Did i say that?

karan6181 (approved):
LGTM. Thanks!

Landanjs (approved):
LGTM! Another 1.3k LOC down

mvpatel2000 (approved):

PR #1730, opened 2022-11-15
Change `build_streaming_cifar10_dataloader()` to use v2 by default

growlix:
Fixes [1729](https://github.com/mosaicml/composer/pull/1729) to use v2 streaming dataloader by default instead of v1.


growlix:
Fixes https://github.com/mosaicml/composer/pull/1729 to use v2 streaming dataloader by default instead of v1.

karan6181 (approved):

PR #1735, opened 2022-11-16
Add python 3.10, pytorch 1.13, cuda 11.7

mvpatel2000:
# What does this PR do?

Add's PyTorch 1.13+CUDA11.7 images on Python 3.10

Removes build for legacy PyTorch 1.10 images.

[CO-1417](https://mosaicml.atlassian.net/browse/CO-1417)


docker/README.md
```diff
@@ -32,16 +32,18 @@ The base flavor contains PyTorch pre-installed; the vision flavor also includes
...
+| Ubuntu 20.04   | Base     | 1.12.1            | 11.6.2         | 3.9              | `mosaicml/pytorch:1.12.1_cu116-python3.9-ubuntu20.04`                                             |
+| Ubuntu 20.04   | Base     | 1.12.1            | cpu            | 3.9              | `mosaicml/pytorch:1.12.1_cpu-python3.9-ubuntu20.04`                                               |
+| Ubuntu 20.04   | Base     | 1.11.0            | 11.5.2         | 3.8              | `mosaicml/pytorch:1.11.0_cu115-python3.8-ubuntu20.04`                                             |
+| Ubuntu 20.04   | Base     | 1.11.0            | cpu            | 3.8              | `mosaicml/pytorch:1.11.0_cpu-python3.8-ubuntu20.04`                                               |
+| Ubuntu 20.04   | Base     | 1.10.2            | 11.3.1         | 3.7              | `mosaicml/pytorch:1.10.2_cu113-python3.7-ubuntu20.04`                                             |
+| Ubuntu 20.04   | Base     | 1.10.2            | cpu            | 3.7              | `mosaicml/pytorch:1.10.2_cpu-python3.7-ubuntu20.04`                                               |
```
bandish-shah, line 44: Can we take this opportunity to get rid of the Torch 1.10.2 images?  Would like to actively support 3 versions, these will still be around they just won't update anymore.

docker/build_matrix.yaml
```diff
@@ -43,97 +43,116 @@
   PYTORCH_VERSION: 1.12.1
   TAGS:
   - mosaicml/pytorch:1.12.1_cu116-python3.9-ubuntu20.04
-  - mosaicml/pytorch:latest
```
bandish-shah, line 46: I'm thinking we don't move the `latest` tag just yet since the focus of this PR is to just start building the Torch 1.13 images.  I think Docker convention is to typically point `latest` to the latest _stable_ image and even though we don't explicitly version these image just yet, the Torch 1.12 image is still technically the most stable.  We can update the `latest` pointer to Torch 1.13 after we have some soak time.  I also don't think we want to accidentally force customers to update to Torch 1.13 if they're using `latest`.

docker/build_matrix.yaml
```diff
@@ -43,97 +43,116 @@
...
+  TARGET: pytorch_stage
+  TORCHTEXT_VERSION: 0.14.0
+  TORCHVISION_VERSION: 0.14.0
+- BASE_IMAGE: nvidia/cuda:11.7.1-cudnn8-devel-ubuntu20.04
+  CUDA_VERSION: 11.7.1
+  PYTHON_VERSION: '3.10'
+  PYTORCH_VERSION: 1.13.0
+  TAGS:
+  - mosaicml/pytorch_vision:1.13.0_cu117-python3.10-ubuntu20.04
+  - mosaicml/pytorch_vision:latest
```
bandish-shah, line 75: Same regarding `latest` for the `mosaicml/pytorch_vision` image.

docker/build_matrix.yaml
```diff
@@ -43,97 +43,116 @@
...
+  TORCHTEXT_VERSION: 0.14.0
+  TORCHVISION_VERSION: 0.14.0
+- BASE_IMAGE: nvidia/cuda:11.7.1-cudnn8-devel-ubuntu20.04
   COMPOSER_INSTALL_COMMAND: mosaicml[all]
-  CUDA_VERSION: 11.6.2
-  PYTHON_VERSION: '3.9'
-  PYTORCH_VERSION: 1.12.1
+  CUDA_VERSION: 11.7.1
+  PYTHON_VERSION: '3.10'
+  PYTORCH_VERSION: 1.13.0
```
bandish-shah, line 105: Same thing regarding all the `mosaicml/composer` images, lets leave them to 1.12 for now.  For testing we can just pull the 1.13 images and install composer manually.

bandish-shah (changes requested):
Functionally looks good!  We're in this weird place right now where our Docker prod images are built from `dev` so we have to take care to not update `latest` tags prematurely.  Few comments inline about this.  Plan is to eventually bifurcate this into a dev and prod flow so we can be a bit more aggressive once that's done.

bandish-shah (approved):
LGTM!

mvpatel2000:
Looks like deepspeed had an unproperly wrapped torch import, so we have to go all the way to latest

mvpatel2000:
Looks like FSDP also uses deprecated code....

I've opened a JIRA [CO-1438](https://mosaicml.atlassian.net/browse/CO-1438) to track removing these warnings once Deepspeed and FSDP release fixes.

bcui19 (approved):
LGTM! Thanks for helping fix the FSDP tests. As a note (and I'll add a JIRA for this later). PyTorch 1.13 and FSDP had some breaking changes when it comes to the 'MixedPrecision' that is passed into FSDP, which affects what parameters can be run. I need to dig more into it, since it looks like some of it might be alleviated with nightly, but not 100% sure.

PR #1744, opened 2022-11-19
Add huggingface info to state dict

dakinggg:
# What does this PR do?
Add huggingface info to the state dict in the new `integrations` key. Later PRs will use this info. This PR is just for saving it, and a test to make sure it is enough information to reinstantiate an equivalent model/tokenizer.

Is the start to [CO-1109](https://mosaicml.atlassian.net/browse/CO-1109)

# Before submitting


A-Jacobson (approved):
hurray tests and this won't affect other models so I'm cool with it!

composer/core/state.py
```diff
@@ -511,7 +511,19 @@ def fsdp_enabled(self):
...
+        """Gets a dictionary of information about integrations to store in the state dict.
+
+        This metadata is used for loading things from state dict that need to be done outside
+        of the normal Composer load path (e.g. HuggingFace model/tokenizer).
+        """
+        from composer.models import HuggingFaceModel
+        integrations = {}
+        if isinstance(self.model, HuggingFaceModel):
+            integrations['huggingface'] = self.model.get_metadata()
+        return integrations
```
mvpatel2000, line 524: can we try catch instead of type checking? more pythonic and lets users override if they really want to. If we do that, we can rename `get_metadata` to `get_huggingface_metadata`

mvpatel2000 (approved):
LGTM. Left a minor suggestion

composer/core/state.py
dakinggg: Hmm, I'm not sure that I want to use this code path to support users easily packing arbitrary info into the state dict. I'd prefer we have "control" over what goes into the `huggingface` key in the state dict, and provide a different code path for users storing whatever they want in it. I think that would be a different design question, and we would probably create an interface like `MetadataStoringModel`, and then if you implement that interface we store whatever metadata you want in the checkpoint under the key that you provide. Perhaps we do want to support something like that at some point, and the hugging face one could be refactored to the common interface at that point, but I'd probably wait for at least one more use case to pop up before that. Thoughts?

composer/models/huggingface.py
```diff
@@ -114,3 +117,41 @@ def get_metrics(self, is_train: bool = False) -> Dict[str, Metric]:
...
         metric.update(outputs, self.labels)
+
+    def get_metadata(self):
+        model_output = {}
+        tokenizer_output = {}
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_dir = Path(tmp_dir)
+            model_dir = tmp_dir / 'model'
+            tokenizer_dir = tmp_dir / 'tokenizer'
+            self.model.config.save_pretrained(model_dir)
```
alextrott16, line 128: Since we're using this API, is it worth doing some type checking in `__init__` to make sure the model and tokenizer match the input signature?

composer/core/state.py
alextrott16: FWIW, Daniel's suggestion makes sense to me.

alextrott16 (approved):
See comment but overall LGTM

composer/core/state.py
alextrott16: FWIW, Daniel's suggestion makes sense to me.

mvpatel2000: Ok, I'm good with keeping it not user aligned. I would still suggest renaming to `get_huggingface_metadata` to be explicit. I also would prefer try catch over type checking because of the weird import. Iirc python is EAFP > LBYL. defer to you though

composer/models/huggingface.py
dakinggg: I'm going to leave this as is. We don't typically explicitly raise errors for wrong types, and instead rely on type checking tools + it will just crash if the user uses the wrong type.

composer/core/state.py
dakinggg: Going to leave as is, I prefer explicit, and in this case the "except" would probably be more common than the "try" because most models are not `HuggingFaceModel`. Open to revisiting in the future if the imports in this function are getting unwieldy.

PR #1745, opened 2022-11-19
inference mode

mvpatel2000:
# What does this PR do?

Inference mode is faster than `torch.no_grad`. It disables some metadata tracking to go faster. 

[CO-1431](https://mosaicml.atlassian.net/browse/CO-1431)


mvpatel2000:
Sad looks like we can't do this. :( In place operators are causing chaos

PR #1746, opened 2022-11-19
Global batch size

mvpatel2000:
# What does this PR do?

Previously, our helper functions to build dataloaders accepted a per device batch size. We've seen users repeatedly mess this up and instead pass global batch size. This PR switches our helper functions to use global batch size, which will make abstractions easier for users. 

As a follow-up, [CO-1428](https://mosaicml.atlassian.net/browse/CO-1428) will check to ensure the same global batch size is used when resuming or otherwise emit a warning.

[CO-1429](https://mosaicml.atlassian.net/browse/CO-1429)


mvpatel2000:
Requesting from Hanlin as the initial versions were constructed based on post-YAHP epics iirc

hanlint (commented):
LGTM -- cc: @codestar12 and the hydra repositories

Do any of the examples need to change?

codestar12:
looks good to me. This should simplify the hydra trainer as I won't need to manually apply this anymore.

codestar12 (commented):
LGTM. Do we still have plans to allow trainer to accept a factory function for building a dataset?

mvpatel2000:
> LGTM. Do we still have plans to allow trainer to accept a factory function for building a dataset?

Discussed offline, answer is no for now

knighton (approved):
Let's Get That Merged

PR #1749, opened 2022-11-21
Supporting `train_device_microbatch_size`

mvpatel2000:
# What does this PR do?

Supports `train_device_microbatch_size`, which we will eventually switch to after deprecating `grad_accum`. This has two nice properties:
1. users like to think about how big of a batch size fits in device. grad_accum is far more non intuitive mentally
2. this lets users scale arbitrarily without having to retune parameters for different # of GPUs.
3. some performance benefits -- unlocks better speedups with algorithms like selective sampling

In this PR, we 
- add `train_device_microbatch_size`
- migrate unit tests to use `train_device_microbatch_size`
- deprecate `grad_accum`

Note: We support everything from Trainer args in a backwards compatible way. We do a breaking change on `DataSpec` api, so it will be breaking for users who have custom `DataSpec`s and use `grad_accum`. This feels niche enough to be ok breaking, and backwards support for this gets really messy.

Note: We remove one test case in `test_split_batch` because it was broken the whole time and did not do intended behavior.

In the next PR, we will switch
- switch `eval_batch_split` to `eval_device_microbatch_size`

In the PR after 0.12 release, we will
- remove `grad_accum` and complete migration

[CO-956](https://mosaicml.atlassian.net/browse/CO-956)


mvpatel2000:
Tagging a whole bunch of people for review to get enough perspective on API changes. I'll be in and out this week / probably out after tonight so not super urgent

dakinggg:
Ok, after thinking about it more I'm good with the API changes. I think we don't really want users to have to think about this param at all (whether its grad accum or device batch size) and just have it be auto, so I'm not sure about the necessity of this change, but it seems ok to me. Please rerequest when its ready for full review.

tests/test_split_batch.py
```diff
@@ -85,46 +77,84 @@ def dummy_batches(batch_size=12):
...
 def test_split_tuple_long(batch):
-    microbatches = _default_split_batch(batch, num_microbatches=3)
+    microbatches = _default_split_batch(batch, microbatch_size=4)
     assert len(microbatches[0]) == 4
 
 
+@pytest.mark.parametrize('batch', dummy_batches(5))
+def test_odd_batch_sizes(batch):
+    microbatches = _default_split_batch(batch, microbatch_size=2)
+    # should split into [len(2), len(2), len(2)]
```
dakinggg, line 108: typo on size of last batch

composer/trainer/trainer.py
```diff
@@ -908,18 +979,47 @@ def __init__(
...
-                             "a mini-run with `grad_accum='auto'` to identify the optimal grad_accum value and "
-                             'then manually specify that in a second run with profiler.')
-        grad_accum = _get_initial_grad_accum(grad_accum)
+        # Microbatching
+        # To support backwards compatability, we currently support both train_device_microbatch_size
+        # and grad_accum. If both are specified, we will use train_device_microbatch_size.
+        if train_device_microbatch_size:
+            using_device_microbatch_size = True
+            if grad_accum != 1:
+                raise ValueError(
```
eracah, line 988: "If both are specified, we will use train_device_microbatch_size." , but here if both are specified you error out

composer/trainer/trainer.py
```diff
@@ -908,18 +979,47 @@ def __init__(
...
+            auto_microbatching = _is_auto_microbatching(train_device_microbatch_size, device=self._device)
+            if auto_microbatching and profiler:
+                raise ValueError("`train_device_microbatch_size='auto'` is not compatible with the profiler. It is "
+                                 "recommended to run a mini-run with `train_device_microbatch_size='auto'` to identify "
+                                 'the optimal train_device_microbatch_size value and then manually specify that in a '
+                                 'second run with profiler.')
+            # If auto_microbatching is True, the microbatch size will be determined at when
+            # dataloader is specified.
+            train_device_microbatch_size = _get_initial_train_device_microbatch_size(train_device_microbatch_size, None)
+        elif grad_accum:
```
eracah, line 1001: is there any way you can just convert `grad_accum` to `train_device_microbatch_size` and then just use `train_device_microbatch_size` the rest of the way instead of having to check which one is set at multiple points in the code?

composer/trainer/trainer.py
```diff
@@ -946,8 +1046,10 @@ def __init__(
             model=model,
             callbacks=callbacks,
             grad_accum=grad_accum,
-            auto_grad_accum=auto_grad_accum,
             eval_batch_split=eval_batch_split,
+            train_device_microbatch_size=train_device_microbatch_size,
+            auto_microbatching=auto_microbatching,
+            using_device_microbatch_size=using_device_microbatch_size,
```
eracah, line 1052: do we need all three of these variables. wouldn't just `train_device_microbatch_size` tell you all you need to know?

composer/trainer/trainer.py
```diff
@@ -1612,14 +1720,27 @@ def fit(
...
+                train_device_microbatch_size, self.state.train_dataloader)
+            self.state.using_device_microbatch_size = True
+        elif grad_accum is not None:
+            self.state.auto_microbatching = _is_auto_grad_accum(grad_accum, device=self._device)
+            if self.state.auto_microbatching and self.state.profiler:
                 raise ValueError("`grad_accum='auto'` is not compatible with the profiler. It is recommended to run "
                                  "a mini-run with `grad_accum='auto'` to identify the optimal grad_accum value and "
                                  'then manually specify that in a second run with profiler.')
             self.state.grad_accum = _get_initial_grad_accum(grad_accum)
+            self.state.using_device_microbatch_size = False
```
eracah, line 1743: aren't these two if blocks basically doing the same thing? can't you just map grad_accum to to microbatch_size and then do all the microbatch processing? instead of handling both these cases?
can't you just do:
```python
if grad_accum is not None:

  if grad_accum == 'auto':
    microbatch_size = 'auto'
  else:
    microbatch_size = batch_size // grad_accum

 do_microbatch_size_stuff()

```
or whatever the conversion is

eracah (commented):
Looks good, but seems like there is a lot of code that checks whether `grad_accum` is set or `train_device_microbatch_size` is set and then executes a different block of code depending on which one is set. But the blocks of code do the same thing essentially. Maybe it'd be simpler and more maintainable if you just converted `grad_accum` to the corresponding `device_microbatch_size right` off the bat and then proceeded as if the user specified `train_device_microbatch_size`.

tests/test_split_batch.py
```diff
@@ -85,46 +77,84 @@ def dummy_batches(batch_size=12):
...
+        assert len(last_microbatch['image']) == 1
+        assert len(last_microbatch['target']) == 1
+    if isinstance(last_microbatch, tuple):
+        assert len(last_microbatch[0]) == 1
+    if isinstance(last_microbatch, list):
+        assert len(last_microbatch) == 1
+
+
+@pytest.mark.parametrize('batch', dummy_batches(1))
+def test_batch_size_less_than_num_microbatches(batch):
```
dakinggg, line 121: `test_batch_size_less_than_num_microbatches` -> `test_microbatch_size_greater_than_batch_size`

tests/test_split_batch.py
```diff
@@ -85,46 +77,84 @@ def dummy_batches(batch_size=12):
...
+    with pytest.raises(ValueError):
+        _default_split_batch(batch, microbatch_size=3)
+
+
 @pytest.mark.parametrize('batch', [dummy_maskrcnn_batch(12)])
-def test_split_maskrcnn(batch):
-    microbatches = _split_list(batch, num_microbatches=3)
+def test_num_microbatches_split_maskrcnn(batch):
```
dakinggg, line 127: I think this test should be below the comment about older tests?

composer/trainer/trainer.py
mvpatel2000: Updated to be more clear. If grad_accum=1, we ignore it if train_device_microbatch_size is specified

composer/trainer/trainer.py
mvpatel2000: Unfortunately, no since the batch size is not known a priori. A user could use an iterable which doesn't expose this information, so we can't convert eagerly.

composer/trainer/trainer.py
mvpatel2000: `train_device_microbatch_size` is converted to batch_size if auto is configured. `using_device_microbatch_size` will go away post deprecation

composer/trainer/trainer.py
mvpatel2000: Same comment as before -- batch_size isn't known a priori

composer/core/state.py
```diff
@@ -199,6 +201,7 @@ class State(Serializable):
             before the dataloader is evaluated. The :attr:`~Timestamp.epoch` attribute for this timestamp is always
             ``0``.
         grad_accum (int): The number of gradient accumulation steps per batch.
+        train_device_microbatch_size (int): The size of each train microbatch per device.
```
dakinggg, line 204: duplicated doc?

composer/trainer/trainer.py
eracah: so then when we deprecate grad_accum there is no option for not specifying how you want to scale your batch size on a device: you either specify microbatch_size = 'auto' or you specify a constant?

composer/core/data_spec.py
```diff
@@ -128,9 +190,9 @@ class DataSpec:
             the batch is not modified.
 
         split_batch ((Batch, int) -> Sequence[Batch], optional): Function called by the :class:`.Trainer` to
-            split a batch (the first parameter) into the number of microbatches specified (the second parameter). If the
-            ``dataloader`` yields batches not of type :class:`torch.Tensor`, Mapping, Tuple, or List, then this function must
-            be specified.
+            split a batch (the first parameter) into the microbatches of a given size specified (the second parameter).
```
dakinggg, line 193: "into the microbatches" -> "into microbatches"

dakinggg, line 193: "given size specified" -> "given size"

composer/trainer/trainer.py
mvpatel2000: Correct. Instead of specifying a scalar (grad_accum) you directly specify batch size (microbatch_size) which is more intuitive and easier to figure out

composer/trainer/trainer.py
```diff
@@ -695,6 +758,13 @@ class Trainer:
                 it into ``grad_accum`` sections. Each section is of size ``train_dataloader // grad_accum``.
                 If the batch size of the dataloader is not divisible by ``grad_accum``,
                 then the last section will be of size ``batch_size mod grad_accum``.
+        train_device_microbatch_size (Union[int, str), optional): The number of samples to process on each device per
+            microbatch. during training. Gradients are summed over the microbatches per device. If set to ``auto``,
```
dakinggg, line 762: "microbatch." -> "microbatch"

composer/trainer/trainer.py
```diff
@@ -908,18 +979,47 @@ def __init__(
...
-        auto_grad_accum = _is_auto_grad_accum(grad_accum, device=self._device)
-        if auto_grad_accum and profiler:
-            raise ValueError("`grad_accum='auto'` is not compatible with the profiler. It is recommended to run "
-                             "a mini-run with `grad_accum='auto'` to identify the optimal grad_accum value and "
-                             'then manually specify that in a second run with profiler.')
-        grad_accum = _get_initial_grad_accum(grad_accum)
+        # Microbatching
+        # To support backwards compatability, we currently support both train_device_microbatch_size
+        # and grad_accum. If both are specified, we will use train_device_microbatch_size.
+        if train_device_microbatch_size:
```
dakinggg, line 985: explicitly check `is not None`

composer/trainer/trainer.py
```diff
@@ -908,18 +979,47 @@ def __init__(
...
+                    'Cannot use both train_device_microbatch_size and grad_accum. grad_accum is deprecated '
+                    'so it is recommended to use train_device_microbatch_size.')
+            grad_accum = None
+            auto_microbatching = _is_auto_microbatching(train_device_microbatch_size, device=self._device)
+            if auto_microbatching and profiler:
+                raise ValueError("`train_device_microbatch_size='auto'` is not compatible with the profiler. It is "
+                                 "recommended to run a mini-run with `train_device_microbatch_size='auto'` to identify "
+                                 'the optimal train_device_microbatch_size value and then manually specify that in a '
+                                 'second run with profiler.')
+            # If auto_microbatching is True, the microbatch size will be determined at when
```
dakinggg, line 998: "at when" -> "when"

composer/trainer/trainer.py
eracah: ah ok makes sense. thanks!

dakinggg (approved):
LGTM

nik-mosaic (approved):

composer/core/state.py
mvpatel2000: State also lists attributes along with args

tests/test_split_batch.py
mvpatel2000: Since it calls an underlying function below grad accum / microbatch size abstraction, I'm going to keep it

dakinggg: oh woops I guess i just looked at the name not the implementation. so shouldn't it be called `test_microbatch_size_split_maskrcnn`?

mvpatel2000: oh good catch, forgot to rename

PR #1750, opened 2022-11-22
Adds auto object store creation to `get_file`

dakinggg:
# What does this PR do?
I have had to write code that autoparses the URI to create an `S3ObjectStore` and then calls `get_file` with that newly created object store a couple of times, and the util to pull huggingface tokenizer/model out of a composer checkpoint is going to need to do the same, so I added the same auto parsing support as in trainer to the `get_file` function. It shouldn't change any behavior of any existing code interacting with `get_file`.

Manual tested that I can download a file from S3 using this, and also that checkpointing/autoresume from S3 still works.

Part of [CO-1110](https://mosaicml.atlassian.net/browse/CO-1110)

# Before submitting

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -18,7 +18,6 @@ import warnings
 from copy import deepcopy
 from pathlib import Path
 from typing import Any, Callable, ContextManager, Dict, Iterable, List, Optional, Sequence, TextIO, Tuple, Union, cast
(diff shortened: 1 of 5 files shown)


eracah (approved):
Nice LGTM. Maybe at some point we can push ALL the auto-parsing logic to `get_file` and to `upload_file`

PR #1754, opened 2022-11-24
Autoload HuggingFace model/tokenizer

dakinggg:
# What does this PR do?
This PR adds support for loading a huggingface model and tokenizer from the information saved in the the composer checkpoint that is added by #1744.

Design here: https://www.notion.so/Autoload-Huggingface-models-for-finetuning-59dda720b7ae4bee90d3594a5ab731d9
Closes: [CO-1109](https://mosaicml.atlassian.net/browse/CO-1109) and [CO-1110](https://mosaicml.atlassian.net/browse/CO-1110)

# Before submitting

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -9,6 +9,7 @@ import json
 import logging
 import tempfile
 import textwrap
+import warnings
 from collections import UserDict
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
@@ -69,6 +70,10 @@ class HuggingFaceModel(ComposerModel):
         self.model = model
         self.config = model.config
         self.tokenizer = tokenizer
+
+        if self.tokenizer is None:
+            warnings.warn(
+                'The tokenizer was not provided. This means the tokenizer config will not be saved in the checkpoint.')
 
         if tokenizer is not None and self.config.vocab_size != len(tokenizer):
             # set model's word embedding matrix and final lm_head to vocab size according to tokenizer
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -115,6 +115,8 @@ class HuggingFaceModel(ComposerModel):
         Returns:
             Tuple[transformers.PreTrainedModel, Optional[transformers.PreTrainedTokenizer]]: The loaded huggingface model and (if present) tokenizer
         """
+        import transformers
+
         # default local path to a tempfile if path is not provided
         if local_checkpoint_save_location is None:
             tmp_dir = tempfile.TemporaryDirectory()
@@ -151,7 +153,7 @@ class HuggingFaceModel(ComposerModel):
         else:
             # If the instantiation class is not explicitly provided, attempt to import the saved class and use it
             try:
-                saved_class = import_object(hf_model_state['config']['content'])
+                saved_class = import_object(':'.join(hf_model_state['config']['class'].rsplit('.', maxsplit=1)))
             except (ModuleNotFoundError, AttributeError):
                 raise ValueError(
                     textwrap.dedent(
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -105,46 +105,40 @@ class HuggingFaceModel(ComposerModel):
     ) -> Tuple[transformers.PreTrainedModel, Optional[transformers.PreTrainedTokenizer]]:
         """Loads a HuggingFace model (and tokenizer if present) from a composer checkpoint.
 
-        .. testsetup:: composer.models.huggingface.HuggingFaceModel.hf_from_composer_checkpoint
+        Example:
+
+        .. testsetup::hf_from_composer_checkpoint
+
+            import torch
+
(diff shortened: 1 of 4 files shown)


composer/models/huggingface.py
```diff
@@ -31,6 +36,8 @@ class HuggingFaceModel(ComposerModel):
     Args:
         model (transformers.PreTrainedModel): A 🤗 Transformers model.
         tokenizer (transformers.PreTrainedTokenizer, optional): The tokenizer used to prepare the dataset. Default ``None``.
+                                                                Note: If the tokenizer is provided, its config will be saved in the composer checkpoint, and it can be reloaded
```
mvpatel2000, line 39: should be note syntax? `.. note::`, see Trainer for reference

composer/models/huggingface.py
```diff
@@ -74,6 +94,140 @@ def __init__(self,
...
+                                                                                   If the input ``checkpoint_path`` is already a local path, this will be a symlink.
+                                                                                   Defaults to None, which will use a temporary file.
+
+        Raises:
+            ValueError: If the ``model_instantiation_class``, or the model class saved in the checkpoint, is not able to be imported
+
+        Returns:
+            Tuple[transformers.PreTrainedModel, Optional[transformers.PreTrainedTokenizer]]: The loaded HuggingFace model and (if present) tokenizer
+        """
+        import transformers
```
mvpatel2000, line 159: wrap in MissingConditionalImportError?

docs/source/doctest_fixtures.py
```diff
@@ -85,6 +85,7 @@
     sys.path.insert(0, _repo_root)
 
 from tests.common import SimpleModel
+from tests.common.datasets import RandomTextClassificationDataset
```
mvpatel2000, line 88: unused line?

mvpatel2000 (approved):
LGTM. Would want a review from NLP person as well though before merging?

composer/models/huggingface.py
dakinggg: It seemed like we only did that in class inits, not everywhere the import is used. But i guess you can use this function without calling the class init? I can wrap this if you think that makes sense.

docs/source/doctest_fixtures.py
dakinggg: Not unused, need to import here so it is accessible in doctests

composer/models/huggingface.py
dakinggg: ah thanks will do

composer/models/huggingface.py
dakinggg: wrapped

dakinggg:
@alextrott16 @A-Jacobson This is ready for review now

composer/models/huggingface.py
```diff
@@ -74,6 +95,145 @@ def __init__(self,
 
         self.labels = None  # set in eval_forward() if exists
 
+    @staticmethod
+    def hf_from_composer_checkpoint(
+        checkpoint_path: str,
+        model_instantiation_class: Optional[Union[Type[transformers.PreTrainedModel],
+                                                  Type['_BaseAutoModelClass']]] = None,
+        model_init_kwargs: Optional[dict] = None,
```
alextrott16, line 103: The naming `model_init_kwargs` strikes me as a bit vague. I think we tend to use `model_config` as the name for the config default overrides. In any case, I suggest naming it something that mentions the config, since that is what it directly changes.

composer/models/huggingface.py
```diff
@@ -74,6 +95,145 @@ def __init__(self,
...
+            hf_config_dict['id2label'] = {int(k): v for k, v in hf_config_dict['id2label'].items()}
+
+        loaded_config = transformers.AutoConfig.from_pretrained(hf_config_dict['_name_or_path'], **hf_config_dict)
+        if model_instantiation_class is not None:
+            # If the instantiation class is explicitly provided, use it
+            # The AutoModel* classes have `from_config`, while the PreTrainedModel classes do not
+            if issubclass(model_instantiation_class, transformers.models.auto.auto_factory._BaseAutoModelClass):
+                hf_model = model_instantiation_class.from_config(loaded_config)  # type: ignore
+            else:
+                hf_model = model_instantiation_class(loaded_config)
```
alextrott16, line 199: Since we live and die by the YAML, is there a way to support string arguments for `model_instantiation_class` (i.e. if the input is a string, use that to import the class and then use the imported class)? Maybe this is already solved with Hydra, but we don't want to force that.

alextrott16 (approved):
One name suggestion and one feature request. I leave them to you to decide whether they are worth doing. So, LGTM!
Disclaimer: if there are issues with the tests, it's very possible I have overlooked them, but they seem OK at a glance.

composer/models/huggingface.py
dakinggg: :+1: changed to `model_config_kwargs`

composer/models/huggingface.py
dakinggg: yeah, should be pretty easy to add, will do.

PR #1762, opened 2022-11-29
Add more event tests

mvpatel2000:
# What does this PR do?

As title. Spotted missing when documenting tests.


tests/test_events.py
```diff
@@ -124,8 +124,10 @@ def _assert_expected_event_calls(self, trainer: Trainer, eval_interval: Time, nu
 
         expected_num_calls = {
             Event.INIT: 1,
+            Event.AFTER_LOAD: 1,
             Event.EPOCH_START: num_epochs,
             Event.BATCH_START: total_steps,
+            Event.BEFORE_DATALOADER: total_steps + num_epochs,  # extra call per epoch when dataloader is exhausted
```
karan6181, line 130: Should this be `total_steps * num_epochs` instead of `total_steps + num_epochs` ? Maybe I am missing something here !!

mvpatel2000: it should be +. Basically, `BEFORE_DATALOADER` runs one extra time at end of each epoch (since the subsequent call  yields no data, resulting in a `StopIteration` exception). Since we don't know a priori we will encounter that, we run it one time more than every other event per epoch. 

`total_steps` is number of steps across all epochs

karan6181 (approved):
LGTM. Thanks!

PR #1763, opened 2022-11-29
Autolog hparams

eracah:
# What does this PR do?

automatically extracts and recursively logs all hyperparameters passed to `Trainer`

Manual test:
```python
train_set = datasets.MNIST("data", train=True, download=True, transform=transform)
val_set = datasets.MNIST("data", train=False, download=True, transform=transform)
train_dataloader = DataLoader(train_set, batch_size=128)
eval_dataloader = DataLoader(val_set, batch_size=64)
model=mnist_model(num_classes=10)


trainer = Trainer(
    model=model,
    max_duration="3ep",
    train_subset_num_batches=4,
    optimizers=[Adam(model.parameters(), eps=1e-3)],
    progress_bar=False,
    log_traces=False,
    log_to_console=True,
    console_log_interval='1ep',
    auto_log_hparams=True,
    algorithms=[ChannelsLast(), ColOut(), BlurPool()],
    callbacks=[SpeedMonitor(), MemoryMonitor(), ImageVisualizer()],
    loggers=[WandBLogger()],
)
trainer.fit()
```

[manual test result](https://wandb.ai/mosaic-ml/evan-composer-manual_tests/runs/bi2g61av/overview)

# Before submitting


dakinggg (changes requested):
I think this may require a design discussion. As implemented, it is pretty fragile and (I think) likely to crash. I tried it on a random test, and it crashed with recursion error (I think) because model has logger which has state which has model, creating an infinite loop. 

A couple other things:
- In the manual test linked, there are many `key: value` pairs that are `null: None` or `null: null`. 
- Why do you need to add `self.local_hparams` to some stuff? A user wouldn't know they need to implement for their code, so I'd prefer a design that doesn't necessitate that.

What do you think/is there a way to maybe make this less aggressively recursive and more robust? Basically, my code should never crash because I turned on `auto_log_hparams`, and I think I would need a lot more tests to prove that the current implementation is robust to many situations.

eracah:
> I think this may require a design discussion. As implemented, it is pretty fragile and (I think) likely to crash. I tried it on a random test, and it crashed with recursion error (I think) because model has logger which has state which has model, creating an infinite loop.
> 
> A couple other things:
> 
> * In the manual test linked, there are many `key: value` pairs that are `null: None` or `null: null`.
> * Why do you need to add `self.local_hparams` to some stuff? A user wouldn't know they need to implement for their code, so I'd prefer a design that doesn't necessitate that.
> 
> What do you think/is there a way to maybe make this less aggressively recursive and more robust? Basically, my code should never crash because I turned on `auto_log_hparams`, and I think I would need a lot more tests to prove that the current implementation is robust to many situations.

Ok let's discuss this offline

PR #1764, opened 2022-11-29
Remove Fused LayerNorm

nik-mosaic:
Draft PR. On hold until v0.13 or later to allow time for users to switch.

 Still need to fix some wording/comments in a few places, and test.


nik-mosaic:
Closing without merging --- we're holding off on removing Fused LayerNorm until we reach parity with a replacement. See #1931 for details.

PR #1765, opened 2022-11-29
Add device to state

mvpatel2000:
# What does this PR do?

As title. Will help algorithms/callbacks move data around

[CO-1448](https://mosaicml.atlassian.net/browse/CO-1448)


mvpatel2000:
@alextrott16 tagging since it touches seq length warmup. Feel free to suggestion any other simplifications that arise from this.

composer/core/state.py
```diff
@@ -139,6 +140,7 @@ class State(Serializable):
         algorithms (Algorithm | Sequence[Algorithm], optional): The algorithms used for training.
         callbacks (Callback | Sequence[Callback], optional): The callbacks used for training.
         deepspeed_config (Dict[str, Any], optional): The configuration dictionary for deepspeed.
+        device (Device): The device used by this process.
```
dakinggg, line 143: I'm not sure whether the name should be changed, or just the docs more clear, but I think we need to be a bit more clear about what this device is. Technically you can use whatever device you want in your model code. I guess this is the device that we are going to move your model to, for this process?

dakinggg (approved):
LGTM

eracah (approved):
LGTM.

nik-mosaic (approved):

PR #1766, opened 2022-11-29
Switch flash attention to tag

mvpatel2000:
# What does this PR do?

Now that flash attention has tags, we should use a release tag instead of a frozen fork.

[CO-1455](https://mosaicml.atlassian.net/browse/CO-1455)


bandish-shah (approved):
👍

PR #1769, opened 2022-12-01
unregister

mvpatel2000:
# What does this PR do?

Fixes memory not being freed with object deletion. 

Experiment training resnet twice in same python process
Before:
image
After: 
image

[CO-1443](https://mosaicml.atlassian.net/browse/CO-1443)

### DIFF
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -445,6 +445,8 @@ class Engine():
             # Do not attempt to shutdown again, since close() already ran via __atexit__ or was already invoked
             return
         self.close()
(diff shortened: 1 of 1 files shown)


dakinggg (approved):
LGTM, might be a good idea to get someone with more context on why this was added in the first place to review?

mvpatel2000:
> LGTM, might be a good idea to get someone with more context on why this was added in the first place to review?

Working with Hanlin to figure out why we need atexit. I want to remove if possible, but this will patch for now

PR #1774, opened 2022-12-02
Add OCI ObjectStore

eracah:
# What does this PR do?

adds an OCIObjectStore backend
ran manual tests for checkpointing and added unit tests

fixes CO-1297

# Before submitting


composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,96 @@
...
+            raise FileNotFoundError(f'Unable to locate oci://{self.bucket}@{self.namespace}/{object_name}')
+
+    def upload_object(
+        self,
+        object_name: str,
+        filename: Union[str, pathlib.Path],
+        callback: Optional[Callable[[int, int], None]] = None,
+    ):
+        del callback
+        with open(filename, 'rb') as f:
```
karan6181, line 69: You can also use `UploadManager` rather than `put_object` which I guess allows parallel uploads as a chunk. Something like this

```
from oci.object_storage import UploadManager
upload_manager = UploadManager(object_storage_client, allow_parallel_uploads=True, parallel_process_count=3)
response = upload_manager.upload_file(namespace, bucket_name, object_name, filename)
```

eracah: good call. done

tests/utils/object_store/object_store_settings.py
```diff
@@ -55,6 +55,9 @@
 object_stores = [
     pytest.param(x, marks=_object_store_marks[x], id=x.__name__)
     for x in get_module_subclasses(composer.utils.object_store, ObjectStore)
+    # Decided to make a separate test suite for OCI instead of use this existing one
+    # because it is kinda complex to work with.
+    if not issubclass(x, OCIObjectStore)
```
mvpatel2000, line 60: Can we reword this to be a note

mvpatel2000 (commented):
LGTM. do we need integration tests

composer/utils/file_helpers.py
```diff
@@ -315,7 +315,8 @@ def parse_uri(uri: str) -> Tuple[str, str, str]:
                               Backend and bucket name will be empty string if the input is a local path
     """
     parse_result = urlparse(uri)
-    backend, bucket_name, path = parse_result.scheme, parse_result.netloc, parse_result.path
+    backend, net_loc, path = parse_result.scheme, parse_result.netloc, parse_result.path
```
dakinggg, line 318: Do you need this logic in `maybe_create_remote_uploader_downloader_from_uri` too?

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""S3-Compatible object store."""
```
dakinggg, line 4: should be OCI?

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+            import oci
+        except ImportError as e:
+            raise MissingConditionalImportError(conda_package='oci',
+                                                extra_deps_group='oci',
+                                                conda_channel='conda-forge') from e
+
+        # Format paths
+        self.bucket = bucket.strip('/')
+        self.prefix = prefix.strip('/')
+        if self.prefix:
```
dakinggg, line 42: explicitly check empty string (so this doesn't capture all other truthy values

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        self.client = oci.object_storage.ObjectStorageClient(config=config,
+                                                             retry_strategy=oci.retry.DEFAULT_RETRY_STRATEGY)
+        self.namespace = self.client.get_namespace().data
+        self.upload_manager = oci.object_storage.UploadManager(self.client)
+
+    def get_uri(self, object_name: str) -> str:
+        return f'oci://{self.bucket}/{object_name}'
+
+    def get_object_size(self, object_name: str) -> int:
+        objects = self.client.list_objects(self.namespace, self.bucket, prefix=object_name, fields='size').data.objects
```
dakinggg, line 59: Is there no direct get object api?

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+        relevant_objects = [obj for obj in objects if obj.name == object_name]
+        if relevant_objects:
```
dakinggg, line 61: explicitly check list length

composer/utils/object_store/oci_object_store.py
dakinggg: can you use the same api is in download object? and then you don't need to raise your own error, can just rely on oci's error if object isn't found

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+    def download_object(
+        self,
+        object_name: str,
+        filename: Union[str, pathlib.Path],
+        overwrite: bool = False,
+        callback: Optional[Callable[[int, int], None]] = None,
+    ):
+        del callback
+        if os.path.exists(filename) and not overwrite:
+            raise FileExistsError(f'The file at {filename} already exists')
```
dakinggg, line 89: "exists" -> "exists and overwrite is set to False"

composer/utils/object_store/oci_object_store.py
```diff
@@ -0,0 +1,104 @@
...
+                                        file_path=filename)
+
+    def download_object(
+        self,
+        object_name: str,
+        filename: Union[str, pathlib.Path],
+        overwrite: bool = False,
+        callback: Optional[Callable[[int, int], None]] = None,
+    ):
+        del callback
```
dakinggg, line 87: Is there a reason the logic in this function is not identical to the `S3ObjectStore` logic (modulo the exist API of boto vs oci)?

composer/utils/file_helpers.py
eracah: isn't it in both? they both use that function

dakinggg: ah sorry, i misread which function this was in

tests/utils/object_store/object_store_settings.py
eracah: done

composer/utils/object_store/oci_object_store.py
eracah: good catch. fixed

composer/utils/object_store/oci_object_store.py
eracah: oohh right. done

composer/utils/object_store/oci_object_store.py
eracah: there is, but UploadManager handles large files better. see @karan6181's [comment ](https://github.com/mosaicml/composer/pull/1774#discussion_r1037832500)

composer/utils/object_store/oci_object_store.py
eracah: nice. yeah I'm sensing a pattern here. fixed

composer/utils/object_store/oci_object_store.py
eracah: done

composer/utils/object_store/oci_object_store.py
eracah: I think `upload_file` for s3 takes a callback and that function oci doesn't

dakinggg: ah, i see. there is also some difference in the tmp file cleanup that im not sure if needs to be replicated?

composer/utils/object_store/oci_object_store.py
eracah: actually, one can use get_object. It was just more convoluted to get the size from get_object, but I figured it out. fixed

composer/utils/object_store/oci_object_store.py
eracah: right, but we'll keep it simple for now

dakinggg (approved):

eracah:
> LGTM. do we need integration tests

at some point yes, but maybe wait until all the integration test infra is all set

mvpatel2000 (approved):
Lgtm

setup.py
```diff
@@ -189,6 +189,10 @@ def package_files(prefix: str, directory: str, extension: str):
     'apache-libcloud>=3.3.1,<4',
 ]
 
+extra_deps['oci'] = [
+    'oci>=oci-2.88.2',
+]
```
mvpatel2000, line 194:
```suggestion
extra_deps['oci'] = [
    'oci>=oci-2.88.2,<oci-3.0.0',
]
```
@eracah  can you confirm this new formatting works? why is there version naming weird and prefixed with oci.... In either case, we should pin the version

eracah: oh good catch. no it shouldn't have oci in the version. fixed

PR #1777, opened 2022-12-02
move to pypi release for flash attn

mvpatel2000:
# What does this PR do?

As title

[CO-1489](https://mosaicml.atlassian.net/browse/CO-1489)


bandish-shah (approved):
👍 approving to unblock, should we add a Composer dependency as well in `setup.py`?

mvpatel2000:
> 👍 approving to unblock, should we add a Composer dependency as well in `setup.py`?

only used in performance/benchmarks repo, so let's not for now

dakinggg:
@mvpatel2000 you should be able to merge this now if you merge dev

PR #1779, opened 2022-12-03
Speeding up tests

dakinggg:
# What does this PR do?
Just trying to speed up the tests. They're still pretty slow, but slightly less so. Local improvements don't directly translate to overall improvements on Jenkins (unsurprisingly). I only looked at cpu tests, not gpu. The one big thing that I'm aware of that I don't know how to fix is that comet/wandb are pretty slow to start/close, and we run them many times (mostly for the test that does the cross product of loggers and callbacks). If anyone has any bright ideas for speeding them up, please share.

Running list of what is in here:
- group wandb/comet tests
- reduce calls to HF hub
- reduce training length in some tests
- removes a bunch of unnecessary fit calls in HF tests
- makes the tests that download mnist/cifar daily
- reduce the parametrization grid for the console logger tests
- reduce training length in readmes
- reduce data/model size in precision tests

N/A

# Before submitting


mvpatel2000 (commented):
This is great! Only question is how much flattening all the wandb/comet tests takes. It's a little eh to put them all in one....

mvpatel2000 (approved):

tests/conftest.py
```diff
@@ -159,3 +159,30 @@ def s3_bucket(request: pytest.FixtureRequest):
         return 'my-bucket'
     else:
         return _get_option(request.config, 's3_bucket')
+
+
+# WARNING: These fixtures should be copied if used in a test, as they are session scope to avoid
+# multiple calls to the HuggingFace hub
+@pytest.fixture(scope='session')
```
mvpatel2000, line 166: Can you explain this a bit more? I don't quite get what's going on

dakinggg:
I couldn't think of any clever way to only close the loggers once or something. You have to close them to do any testing of the output because they don't flush their logs before they're closed, and any time the trainer is closed the callback is closed. Each test with a wandb/comet logger takes about four seconds, so it is pretty costly to run all of these separately.

tests/conftest.py
dakinggg: When we use `scope="session"`, the fixture is only created once and the same object is reused for all tests. By doing this we only make one call to the hugging face hub, and then the tests that need this fixture just make a copy, which is cheaper than the call to the hub. And the tests have to make a copy because they modify the object.

eracah (approved):
The comet and wandb parts LGTM. Thanks fordoing this, @dakinggg

tests/conftest.py
dakinggg: haha def not dumb, I will make the comment more clear before this PR goes in

tests/conftest.py
```diff
@@ -159,3 +159,30 @@ def s3_bucket(request: pytest.FixtureRequest):
...
+    hf_tokenizer = transformers.AutoTokenizer.from_pretrained('prajjwal1/bert-tiny')
+    return hf_tokenizer
+
+
+@pytest.fixture(scope='session')
+def tiny_bert_config():
+    transformers = pytest.importorskip('transformers')
+
+    hf_config = transformers.AutoConfig.from_pretrained('prajjwal1/bert-tiny')
+    return hf_config
```
mvpatel2000, line 188: So, my main concern here is someone will forget to deepcopy and other things will break in confusing ways that are hard to debug. I wonder if we can memoize with`functools.lru_cache` again?

dakinggg: yeah it is a fair concern. Unfortunately lru cache also returns a single object, so also needs to be deep copied

```
In [3]: @functools.lru_cache
   ...: def list_function():
   ...:     return [1,2,3]
   ...: 

In [4]: a = list_function()

In [5]: b = list_function()

In [6]: a[0] = 5

In [7]: a
Out[7]: [5, 2, 3]

In [8]: b
Out[8]: [5, 2, 3]
```

dakinggg: resolving offline

dakinggg:
@mvpatel2000 @eracah rerequested review because I added a bunch of stuff. Down to 12m, I think about 5 of that is jenkins/building stuff, and about 7 of that is actually running the tests.

composer/algorithms/seq_length_warmup/README.md
```diff
@@ -60,8 +60,8 @@ from composer.algorithms import SeqLengthWarmup
 trainer = Trainer(model=model,
                   train_dataloader=train_dataloader,
                   eval_dataloader=eval_dataloader,
-                  max_duration='250ep',
-                  algorithms=[SeqLengthWarmup()])
+                  max_duration='25ep',
+                  algorithms=[SeqLengthWarmup(max_seq_length=64)])
```
mvpatel2000, line 64: Can you confirm this with @alextrott16 ? iirc there were some reasons why it was a large length, but not 100% sure

tests/algorithms/test_required_on_load.py
```diff
@@ -85,15 +85,17 @@ def test_idempotent(algo_name: str):
         if algo_name == 'StochasticDepth':
             original_model = composer_resnet(model_name='resnet50')
         elif algo_name in ['Alibi', 'GatedLinearUnits']:
-            pytest.importorskip('transformers')
-            original_model = create_bert_classification()
+            transformers = pytest.importorskip('transformers')
+            from composer.models import HuggingFaceModel
+            hf_model = transformers.AutoModelForSequenceClassification.from_config(tiny_bert_config)
+            original_model = HuggingFaceModel(hf_model, use_logits=True)
```
mvpatel2000, line 91: why this change?

mvpatel2000 (commented):
LGTM. Left minor comments (really questions), holding approval until everyone reviews because test changes are a little scary (will let others approve). 

Do you have any profile data we can dump into PR to look back on later on if we see slowdowns? Curious how much breakdown you have on the total time / where it comes from.

tests/algorithms/test_required_on_load.py
dakinggg: create_bert_classification (with no args) calls out to huggingface hub. This avoids that, same as the other HF tests

composer/algorithms/seq_length_warmup/README.md
dakinggg: There are some constraints around the gap between min and max, and the duration, so that the warmup schedule works out, but will wait for Alex if the example needed to be a really long sequence length for some reason.

dakinggg:
I can't really provide a breakdown because the time on my laptop/interactive instance doesn't translate directly to jenkins time. so i'm not sure beyond what i've already said in the PR description.

nik-mosaic (approved):
(thumbs up emoji) Approved. Thanks.

eracah (approved):
LPGTM

mvpatel2000 (approved):

PR #1785, opened 2022-12-05
Update Auto Microbatching

mvpatel2000:
# What does this PR do?

- Fixes bug in auto microbatching
- Adds unit test

[CO-1473](https://mosaicml.atlassian.net/browse/CO-1473)

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -156,7 +156,7 @@ def _is_auto_grad_accum(grad_accum: Union[int, str], device: Device):
     if grad_accum == 'auto':
         warnings.warn(("Setting `grad_accum='auto'` is an experimental feature which may cause "
                        'uncaught Cuda Out of Memory errors. In this case, please manually '
-                       'set grad_accum explicitly to an integer instead. '))
+                       'set grad_accum explicitly to an integer instead.'))
         if not isinstance(device, DeviceGPU):
             raise ValueError('Can only use adaptive grad_accum on GPU. Please set grad_accum >= 1')
         return True
@@ -168,7 +168,7 @@ def _is_auto_microbatching(device_train_microbatch_size: Union[int, str], device
     if device_train_microbatch_size == 'auto':
         warnings.warn(("Setting `device_train_microbatch_size='auto'` is an experimental feature which may cause "
                        'uncaught Cuda Out of Memory errors. In this case, please manually '
-                       'set device_train_microbatch_size explicitly to an integer instead. '))
+                       'set device_train_microbatch_size explicitly to an integer instead.'))
         if not isinstance(device, DeviceGPU):
             raise ValueError(
                 'Can only use adaptive device_train_microbatch_size on GPU. Please set device_train_microbatch_size >= 1.'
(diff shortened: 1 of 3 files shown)


tests/trainer/test_trainer.py
```diff
@@ -385,6 +385,48 @@ def test_microbatch(
...
+        # Copy the model so the fit_trainer can start with the same parameter values as the init_trainer
+        copied_model = copy.deepcopy(model)
+
+        # Train once with the device_train_microbatch_size=1
+        baseline_event_counter_callback = EventCounterCallback()  # track the number of times microbatches are trained
+        baseline_trainer = Trainer(
+            model=model,
+            max_duration=max_duration,
+            train_dataloader=train_dataloader if dataloader_in_init else None,
+            device_train_microbatch_size='auto',
```
dakinggg, line 409: this should be 1 not auto?

tests/trainer/test_trainer.py
```diff
@@ -385,6 +385,48 @@ def test_microbatch(
...
+            callbacks=[baseline_event_counter_callback],
+        )
+        baseline_trainer.fit(train_dataloader=train_dataloader if not dataloader_in_init else None)
```
dakinggg, line 412: and should be passed as 1 here also if dataloader not in init?

dakinggg (approved):

composer/trainer/trainer.py
```diff
@@ -187,9 +187,16 @@ def _get_initial_grad_accum(grad_accum: Union[int, str]):
         raise ValueError("grad_accum must be an int or ``'auto'``")
 
 
-def _get_initial_device_train_microbatch_size(device_train_microbatch_size: Union[int, str],
-                                              train_dataloader: Optional[Iterable]):
-    if device_train_microbatch_size == 'auto':
+def _get_initial_device_train_microbatch_size(device_train_microbatch_size: Optional[Union[int, str]],
+                                              auto_microbatching: bool,
+                                              train_dataloader: Optional[Iterable]) -> Optional[int]:
```
eracah, line 192: A little confusing that the function takes `device_train_microbatch_size` and `auto_microbatching`. seems a little redundant because `device_train_microbatch_size` can be set to `'auto'`. Perhaps `_get_initial_device_train_microbatch_size` takes in `device` and then it checks like this:
```suggestion
def _get_initial_device_train_microbatch_size(device_train_microbatch_size: Optional[Union[int, str]],
                                              device: Device,
                                              train_dataloader: Optional[Iterable]) -> Optional[int]:
                                              
  if isinstance(device_train_microbatch_size, int):
    return device_train_microbatch_size
  elif _is_auto_microbatching(device_train_microbatch_size, device):
      if train_dataloader is None:
            return None
        try:
            batch_size = getattr(train_dataloader, 'batch_size')
        except AttributeError as e:
            raise AttributeError(
                'device_train_microbatch_size requires the `state.train_dataloader` to have a `batch_size` attribute.'
            ) from e
        return batch_size
        
  else:
    raise ValueError("device_train_microbatch_size must be an int or ``'auto'``")
                                        
```

eracah (approved):
Looks pretty good, but a little confusing with train_microbatching size and the auto_microbatching boolean

tests/trainer/test_trainer.py
mvpatel2000: Resolved offline

PR #1787, opened 2022-12-06
Remove streaming v1 datasets.

knighton:


karan6181:
Should we also remove files under `composer/datasets/streaming` folder ?

A-Jacobson (approved):
signing off that the streaming COCO output is correct and the interface is consistent with the other datasets. Thanks James

mvpatel2000 (approved):
Do we want to keep the example notebooks with streaming v2?

knighton:
> Do we want to keep the example notebooks with streaming v2?

Was thinking of readding with fancy V3 afterward

karan6181 (approved):
LGTM. Thanks!

PR #1789, opened 2022-12-06
Add Deprecation warning for Fused LayerNorm

nik-mosaic:
Add Deprecation warning for Fused LayerNorm. It will be removed in 0.13 and replaced by Low Precision LayerNorm.


composer/algorithms/fused_layernorm/fused_layernorm.py
```diff
@@ -32,6 +32,13 @@ def check_if_apex_installed():
         )
 
 
+def warn_fln_deprecation():
+    warnings.warn(
+        DeprecationWarning(
+            'Fused LayerNorm will be deprecated and removed in the future. Please use Low Precision LayerNorm instead.')
```
mvpatel2000, line 38:
```suggestion
            'Fused LayerNorm will be deprecated and removed in v0.13. Please use Low Precision LayerNorm instead.')
```

mvpatel2000 (approved):
LGTM. Left comment to update warning to clearly specify deprecation warning

PR #1790, opened 2022-12-06
Moving Pytest CPU to GHA

mvpatel2000:
# What does this PR do?

- Moves to GHA
- removes install test (duplicated by CPU tests)
- cancels tests if a new commit is pushed

[CO-1498](https://mosaicml.atlassian.net/browse/CO-1498)


.github/workflows/pytest-cpu.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Setup
+        run: |
+          set -ex
+          python -m pip install --upgrade pip wheel
+          python -m pip install -e .[all]
+      - name: Run Tests
+        id: tests
+        run: |
+          set -ex
+          python3 -m pytest
```
dakinggg, line 29: this is actually what jenkins runs

```12:16:57  + ./.ci/test.sh all not daily and not vision and not gpu
12:16:57  + EXTRA_DEPS=all
12:16:57  + MARKERS='not daily and not vision and not gpu'
12:16:57  + export WANDB_ENTITY=mosaicml-public-integration-tests
12:16:57  + WANDB_ENTITY=mosaicml-public-integration-tests
12:16:57  ++ date +%s
12:16:57  + export WANDB_PROJECT=integration-tests-75143-1670357817
12:16:57  + WANDB_PROJECT=integration-tests-75143-1670357817
12:16:57  + S3_BUCKET=mosaicml-internal-integration-testing
12:16:57  + SFTP_URI=sftp://mosaicml-integration-testing@s-c07c6cb0dd1441dbb.server.transfer.us-west-2.amazonaws.com/mosaicml-internal-integration-testing```

File: tests/test_docs.py [hunk 2]
```diff
@@ -0,0 +1,102 @@
+# Copyright 2022 MosaicML Composer authors
```
dakinggg, line 1: Any reason not to delete the `.ci/test_docs.py` file?

.github/workflows/pytest-cpu.yaml
```diff
@@ -0,0 +1,80 @@
...
+  workflow_dispatch: {}
+jobs:
+  pytest:
+    timeout-minutes: 30
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - container: mosaicml/pytorch:1.11.0_cpu-python3.8-ubuntu20.04
+            markers: 'not daily and not gpu and not vision and not remote and not doctest'
```
dakinggg, line 13: I don't think we want to turn off remote tests, right?

.ci/Jenkinsfile
```diff
@@ -186,7 +186,7 @@ void scheduleJob(jobs, String image, buildArgs, originalTags) {
     // buildArgs: The build args matrix
     // originalTags: The original tags
 
-    String markers = 'not daily'
+    String markers = 'not daily and not doctest'
```
eracah, line 189: we don't wanna run doctests for each PR? what if your PR changes a docstring?

.github/workflows/pytest-cpu.yaml
```diff
@@ -0,0 +1,80 @@
...
+  pull_request: {}
+  workflow_dispatch: {}
+jobs:
+  pytest:
+    timeout-minutes: 30
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - container: mosaicml/pytorch:1.11.0_cpu-python3.8-ubuntu20.04
```
dakinggg, line 12: can you give all of the matrix elements a `name`? (e.g. `cpu-3.8-1.11`)

tests/test_docs.py
mvpatel2000: it's used in Jenkins. We won't nuke it until we're fully migrated off

.github/workflows/pytest-cpu.yaml
mvpatel2000: They should run only daily. Daily will come in a separate PR

.github/workflows/pytest-cpu.yaml
```diff
@@ -0,0 +1,80 @@
+name: Pytest
```
eracah, line 1: Nice! this is a lot clearer compared to the jenkins config file

eracah (commented):
Nice looks pretty good. thanks for doing this!

karan6181 (commented):
This is Great . Does the `pytest-cpu` runs on every commit in a PR? And do we have to run the same test for all 3 versions of pytorch such as `1.11.0`, `1.12.1` and `1.13.0`? It would be overkill to have this run on every commit I think. 

To brainstorm, we can run some of the pytorch version test nightly once to ensure it passes.

tests/test_docs.py
dakinggg: oh its still used for nightlies. fair enough

.ci/Jenkinsfile
bandish-shah: It's being migrated to the `tests` area, see `tests/test_docs.py` and there's a new doctest marker.

dakinggg: the doctests do run on every pr, this is the default args in the jenkins file

.github/workflows/pytest-cpu.yaml
mvpatel2000: hm... I feel like we shouldnt be hitting AWS every commit... open to suggestions tho

dakinggg: I suppose its fine if it runs daily. It doesn't really seem like a problem to run on every commit though.

dakinggg: up to you

.ci/Jenkinsfile
mvpatel2000: Yep, so I copied the tests into `/tests` folder and added a new tag. Jenkins does something bespoke for doctests, I removed that and made it normal tests. However, I don't want to muck with Jenkins build to switch it, so they don't do that run

mvpatel2000:
> This is Great . Does the `pytest-cpu` runs on every commit in a PR? And do we have to run the same test for all 3 versions of pytorch such as `1.11.0`, `1.12.1` and `1.13.0`? It would be overkill to have this run on every commit I think.

Turns out they're all free for public repos so no harm. It fast fails and terminates early if u push again though

> Brainstorming an idea here, we can even run some of the pytorch version test nightly once to ensure it passes.

yep, I'm going to move some of the nightly tests here in another PR. will figure out cadence for them later

.github/workflows/pytest-cpu.yaml
```diff
@@ -0,0 +1,52 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - python_version: '3.8'
+            container: mosaicml/pytorch:1.11.0_cpu-python3.8-ubuntu20.04
+            markers: 'not daily and not gpu and not vision and not doctest'
+            pytest_command: 'pytest'
+          - python_version: '3.9'
+            container: mosaicml/pytorch:1.12.1_cpu-python3.9-ubuntu20.04
```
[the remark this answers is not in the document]
mvpatel2000: it should only run the vision tests so it won't rerun anything

bandish-shah (approved):
:shipit: Looks good!

PR #1797, opened 2022-12-07
Use streaming v3

knighton:
* Upgrade streaming datasets to v3
* Upgrade trainer/state to do mid-epoch resumption


knighton:
This PR obsoletes https://github.com/mosaicml/composer/pull/1663

composer/core/state.py
```diff
@@ -501,6 +525,19 @@ def algorithms(self):
     def algorithms(self, algorithms: Sequence[Algorithm]):
         self._algorithms[:] = algorithms
 
+    def _dataset_of(self, obj: Optional[Union[Iterable, Evaluator, DataSpec, DataLoader]]) -> Optional[Dataset]:
+        from composer.core.evaluator import Evaluator
+        if obj is None:
+            return None
+        if isinstance(obj, Evaluator):
+            obj = obj.dataloader
```
karan6181, line 533: Should this be `obj.dataloader.dataset` ?

knighton: "obj" (the "dataloader") may be an Evaluator. if it is, it has a "dataloader" field

that "dataloader" may actually be a DataSpec (DataLoader + on-device transform). if it is, it has a "dataloader" field

that result will be the actual DataLoader (return obj.dataset), unless the user has passed some other kind of Iterable in which case it doesn't have a dataset

knighton: Evaluator = wrapper for a dataloader to include metrics that apply to a specific dataset

composer/core/state.py
```diff
@@ -501,6 +525,19 @@ def algorithms(self):
...
+        if isinstance(obj, DataSpec):
+            obj = obj.dataloader
```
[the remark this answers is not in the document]
knighton: see previous comment

composer/trainer/trainer.py
```diff
@@ -1833,22 +1833,24 @@ def _spin_dataloaders(self):
         """
         # spin the evaluator dataloaders once to initialize its sampler deterministically
         # so it does not affect any other RNG reads
-        log.debug('Spinning the dataloaders')
```
mvpatel2000, line 1836: can we keep this?

composer/core/state.py
```diff
@@ -501,6 +525,19 @@ def algorithms(self):
     def algorithms(self, algorithms: Sequence[Algorithm]):
         self._algorithms[:] = algorithms
 
+    def _dataset_of(self, obj: Optional[Union[Iterable, Evaluator, DataSpec, DataLoader]]) -> Optional[Dataset]:
```
mvpatel2000, line 528: can you add a comment on how this fn works? I was also confused (like Karan) on how this worked so it'd be helpful to explain the iterative unpacking

composer/core/state.py
```diff
@@ -292,6 +295,8 @@ def __init__(
         dataloader: Optional[Iterable] = None,
         dataloader_label: Optional[str] = None,
         dataloader_len: Union[int, Time[int]] = -1,
+        dataset_state: Optional[Dict[str, Any]] = None,
+        dataset_resumption: Optional[Dict[str, Any]] = None,
```
mvpatel2000, line 299: Needs to be added to doc strings above. We should definitely write a good bit on what each of these are doing because the logic here is subtle and somewhat hard to understand -- it took me a while to remember everything and I helped design the original version :P

composer/core/state.py
```diff
@@ -549,12 +593,35 @@ def _get_state_metadata(self) -> Dict[str, Any]:
...
+        if hasattr(dataset, 'state_dict'):
+            obj['train'] = dataset.state_dict()  # pyright: ignore
+
+        for evaluator in self.evaluators:
+            dataset = self._dataset_of(evaluator)
+            if hasattr(dataset, 'state_dict'):
+                state = dataset.state_dict()  # pyright: ignore
+                # Don't save eval progress because we do not checkpoint during eval.
+                if isinstance(state, dict):
+                    state['sample_in_epoch'] = 0
```
mvpatel2000, line 612: Why is this necessary?

composer/core/state.py
```diff
@@ -747,6 +814,27 @@ def load_optim_state(self, state_dict: Dict[str, Any]):
...
+            dataset = self._dataset_of(evaluator)
+            if hasattr(dataset, 'load_state_dict') and evaluator.label in obj['eval']:
+                dataset.load_state_dict(obj['eval'][evaluator.label])  # pyright: ignore
+                del obj['eval'][evaluator.label]
+                if 'eval' not in self.dataset_resumption:
+                    self.dataset_resumption['eval'] = {}
+                # Note: We currently disable setting dataset_resumption for eval datasets,
+                # which means they have one sample fetched in _spin_dataloaders before training
+                # starts. This avoids "CUDA error: initialization error" -- its not clear why.
+                # self.dataset_resumption['eval'][evaluator.label] = True
```
mvpatel2000, line 836: Can we make sure we're tracking this somewhere in jira?

composer/core/state.py
```diff
@@ -378,6 +386,22 @@ def __init__(
...
 
+    @property
+    def train_dataloader(self):
+        """The train dataloader."""
+        return self._train_dataloader
+
+    @train_dataloader.setter
+    def train_dataloader(self, train_dataloader: Optional[Union[Iterable, DataLoader]]):
+        self._train_dataloader = train_dataloader
+        if self.dataset_state:
```
mvpatel2000, line 397:
```suggestion
        self._train_dataloader = train_dataloader
        # Load dataset state from checkpoint when train_dataloader is set
        if self.dataset_state:
```

composer/core/state.py
```diff
@@ -509,6 +546,13 @@ def evaluators(self):
     @evaluators.setter
     def evaluators(self, evaluators: Union[Evaluator, Sequence[Evaluator]]):
         self._evaluators[:] = list(ensure_tuple(evaluators))
+        if self.dataset_state:
```
mvpatel2000, line 549:
```suggestion
        self._evaluators[:] = list(ensure_tuple(evaluators))
        # Load dataset state from checkpoint when evaluators are set
        if self.dataset_state:
```

mvpatel2000: same as above

composer/core/state.py
mvpatel2000: It would be nice to include a small comment -- it's a little hard to understand what's going on here without context of entire design

mvpatel2000 (approved):
LGTM! Left a few comments to improve readability

composer/core/state.py
```diff
@@ -501,6 +525,19 @@ def algorithms(self):
...
+    def _dataset_of(self, obj: Optional[Union[Iterable, Evaluator, DataSpec, DataLoader]]) -> Optional[Dataset]:
+        from composer.core.evaluator import Evaluator
+        if obj is None:
+            return None
+        if isinstance(obj, Evaluator):
+            obj = obj.dataloader
+        if isinstance(obj, DataSpec):
+            obj = obj.dataloader
+        if isinstance(obj, DataLoader):
+            return obj.dataset
```
karan6181, line 537: looks like this should be just `return obj` instead of `return obj.dataset` since `StreamingDataLoader` has a method `state_dict()` without any parameter. However, `StreamingDataset` has a method `state_dict()` with one parameter.

knighton: Fixed that bug, and updated the doc/comments here.

composer/core/state.py
```diff
@@ -618,8 +618,9 @@ def _dataset_state_dict(self) -> Dict[str, Any]:
 
         dataset = self._dataset_of(self.train_dataloader)
         if hasattr(dataset, 'state_dict'):
-            sample = int(self.timestamp.sample_in_epoch)
-            obj['train'] = dataset.state_dict(sample)  # pyright: ignore
+            num_samples = int(self.timestamp.sample_in_epoch)
+            from_beginning = True
+            obj['train'] = dataset.state_dict(num_samples, from_beginning)  # pyright: ignore
```
karan6181, line 623: Maybe have this `obj['train'] = dataset.state_dict(num_samples, True)` ? and add a comment that it used in `StreamingDataset` ??

karan6181 (approved):

composer/core/state.py
knighton: cleaned up

composer/core/state.py
knighton: done

composer/core/state.py
knighton: added docstrings throughout

bmosaicml (approved):

Landanjs:
ImageNet looks good!
- Streamingv2: 76.48% accuracy
- Streamingv3: 76.52% accuracy
v3 was ~3% slower than v2, but I had to use 40GB A100 instead of 80GB, so I think that accounts for most of the difference.

I did notice v3 ran for 90 more iterations, but I think this was the accurate number of iterations. Maybe v2 had too few samples?

PR #1805, opened 2022-12-10
Remove pip install all from coverage action

dakinggg:
# What does this PR do?
As title, also switches to explicit docker image tags for the vision images because it is faster.

N/A


mvpatel2000 (commented):
Good catch! 

It looks like something broke with coverage report though.... The coverage is 20% and trainer.py is 7%. I definitely saw higher numbers when first building this. I think something might be wrong in the code which is fetching the artifacts.... perhaps they're overwriting each other? Can you fix it in this PR?

mvpatel2000 (approved):

PR #1815, opened 2022-12-13
fix inf eval

mvpatel2000:
# What does this PR do?

As title


dakinggg (approved):
LGTM, please add a test though (I disabled auto merge so you can add a test)

PR #1818, opened 2022-12-14
Add SHARP support to docker

mvpatel2000:
# What does this PR do?

As title. With this, a user can use SHARP if underlying hardware supports it. To do this, we 
0. SHARP already comes packaged with MOFED
1. install https://github.com/Mellanox/nccl-rdma-sharp-plugins
2. run `libtool --finish /usr/local/lib` to link things up

Tested using https://github.com/NVIDIA/nccl-tests.

TODO:
- SHARP seems to be picked up on NCCL tests but not by Composer when running something... unclear why

[CO-1533](https://mosaicml.atlassian.net/browse/CO-1533)


bandish-shah (approved):
LGTM!  Let's hand this off and see if any other dependencies are required.  Assuming there will be some onion peeling with this.

PR #1822, opened 2022-12-15
Testing Infra Cleanup

mvpatel2000:
# What does this PR do?

Cleans up some test infra which will lay groundwork for migrating to nightly tests.

- Simplifies fixture files, consolidating into one file for autouse fixtures and one for the rest
- Removes sftp logging test for remote (expensive to run)
- Minor fixes throughout tests

Note: Tags reordered in workflow are so it makes it a little more readable with nightly tests. I removed that workflow and will merge separately because idt it will be completed before holidays.

[CO-1520](https://mosaicml.atlassian.net/browse/CO-1520)


composer/trainer/trainer.py
```diff
@@ -1675,6 +1675,7 @@ def fit(
 
         # Max Duration
         if duration is not None:
+            log.info(str(duration))
```
dakinggg, line 1678: Is this log intentional? If so it should probably provide some more info about what it is logging

tests/conftest.py
```diff
@@ -45,7 +41,7 @@ def _add_option(parser: pytest.Parser, name: str, help: str, choices: Optional[L
     )
 
 
-def _get_option(config: pytest.Config, name: str, default: Optional[str] = None) -> str:
+def _get_option(config: pytest.Config, name: str, default: Optional[str] = None) -> str:  # type: ignore
```
dakinggg, line 44: Why was this type ignore added?

nik-mosaic (approved):

tests/fixtures/autouse_fixtures.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@pytest.fixture(autouse=True)
+def set_loglevels():
+    """Ensures all log levels are set to DEBUG."""
+    logging.basicConfig()
+    logging.getLogger(composer.__name__).setLevel(logging.DEBUG)
+
+
+@pytest.fixture
```
dakinggg, line 78: Is this supposed to be in the other fixtures file?

dakinggg (approved):
LGTM

tests/conftest.py
mvpatel2000: Function not used in file (imported elsewhere)

bandish-shah (approved):
👍

PR #1824, opened 2022-12-15
Fix Export Docs

mvpatel2000:
# What does this PR do?

As title

[CO-1553](https://mosaicml.atlassian.net/browse/CO-1553)

### DIFF
--- a/examples/exporting_for_inference.ipynb
+++ b/examples/exporting_for_inference.ipynb
@@ -23,11 +23,11 @@
     "\n",
     "We'll touch on:\n",
     "\n",
-    "* [Our standalone export API](#torchscript-export-using-standalone-api)\n",
-    "* [Exporting from the trainer with callbacks](#export-using-a-callback)\n",
-    "* [Exporting from the trainer directly](#alternative-export)\n",
-    "* [Exporting from an existing checkpoint](#exporting-from-an-existing-checkpoint)\n",
-    "* [Supported Composer algorithms](#algorithm-compatibility)\n",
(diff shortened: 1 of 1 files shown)


examples/exporting_for_inference.ipynb
```diff
@@ -23,11 +23,11 @@
     "\n",
     "We'll touch on:\n",
     "\n",
-    "* [Our standalone export API](#torchscript-export-using-standalone-api)\n",
-    "* [Exporting from the trainer with callbacks](#export-using-a-callback)\n",
-    "* [Exporting from the trainer directly](#alternative-export)\n",
-    "* [Exporting from an existing checkpoint](#exporting-from-an-existing-checkpoint)\n",
-    "* [Supported Composer algorithms](#algorithm-compatibility)\n",
+    "* [Our standalone export API](#Torchscript-Export-Using-Standalone-API)\n",
```
eracah, line 26: This one still doesn't work because it automatically transforms it  from`#Torchscript-Export-Using-Standalone-API` to `#Torchscript-Export-Using-Standalone-Api` in the URL

eracah (commented):
Nice work! all of them work now except the #Torchscript-Export-Using-Standalone-API one. See my comment

examples/exporting_for_inference.ipynb
eracah: at least when I test it using
```bash
make html
make host
```

eracah: ok works now

eracah (approved):
LGTM

PR #1828, opened 2022-12-16
Nlp pipeline test

dakinggg:
# What does this PR do?
Adds a unit test that exercises the full NLP pipeline (pretrain, finetune, export) for both a huggingface model and a simple transformer model.

Notes:
- we add an algorithm (gated linear units) for the HF model, but no algorithms for the simple transformer model, because (I think?) we don't have any NLP model surgery algorithms that apply to non HF models.
- we use onnx export for HF model, but torchscript export for the simple transformer model, because torch's `TransformerEncoder` does not seem to be able to be traced.

Nikhil: I've made a note to update the fixtures to use whatever new thing you create in your upcoming PR

Along for the ride:
- the precision test is failing again, had to tweak the threshold. creating a jira to look into this and why it keeps happening.

Closes [CO-1480](https://mosaicml.atlassian.net/browse/CO-1480)

# Before submitting


tests/test_full_nlp.py
```diff
@@ -0,0 +1,198 @@
...
+        raise ValueError('Unsupported save format')
+
+    torch.testing.assert_close(
+        loaded_model_out[1] if isinstance(loaded_model_out, list) else loaded_model_out.detach().numpy(),
+        original_output.detach().numpy()
+        if isinstance(original_output, torch.Tensor) else original_output.logits.detach().numpy())
+
+
+@device('cpu', 'gpu')
+@pytest.mark.parametrize('model_type,algorithms,save_format', [('tinybert_hf', [GatedLinearUnits()], 'onnx'),
```
nik-mosaic, line 147: I would separate out parameters into different parametrize lines if you want to grid-test over models x algorithms x export types. I think we should grid test a bit; we don't have to do everything if its slow.

tests/test_full_nlp.py
```diff
@@ -0,0 +1,198 @@
...
+    else:
+        raise ValueError('Unsupported model type.')
+    finetuning_trainer, finetuning_dataloader = finetuning_test_helper(tiny_bert_tokenizer, finetuning_model,
+                                                                       algorithms, pretraining_output_path, tmp_path,
+                                                                       device)
+
+    # inference
+    batch = next(iter(finetuning_dataloader))
+    finetuning_trainer.state.model.to('cpu')
+    finetuning_trainer.state.model.eval()
```
nik-mosaic, line 196: If this is the full NLP pipeline, we want to try saving the checkpoint to a bucket etc. and then loading from the bucket back for inference?

tests/test_full_nlp.py
```diff
@@ -0,0 +1,198 @@
...
+    torch.testing.assert_close(
+        loaded_model_out[1] if isinstance(loaded_model_out, list) else loaded_model_out.detach().numpy(),
+        original_output.detach().numpy()
+        if isinstance(original_output, torch.Tensor) else original_output.logits.detach().numpy())
+
+
+@device('cpu', 'gpu')
+@pytest.mark.parametrize('model_type,algorithms,save_format', [('tinybert_hf', [GatedLinearUnits()], 'onnx'),
+                                                               ('simpletransformer', [], 'torchscript')])
+def test_full_nlp_pipeline(model_type, algorithms, save_format, tiny_bert_tokenizer, tmp_path, request, device):
```
nik-mosaic, line 149: Would add a comment at the top of this test explaining it a little more, since it is long & includes multiple helpers. Will be easier for others coming across this file later on.

tests/test_simple_nlp.py
```diff
@@ -55,3 +56,56 @@ def test_simple_nlp():
     num_predict_batches_expected = ((size - 1) // batch_size) + 1
     assert len(predictions) == num_predict_batches_expected
     assert predictions[0].shape == (batch_size, 2)
+
+
+def test_simple_nlp_lm(tiny_bert_tokenizer, tiny_bert_model):
```
nik-mosaic, line 61: test_simple_nlp_mlm or lm?

nik-mosaic (approved):
Great work! Left a couple of comments, mostly minor cosmetic ones. I think this hits basically all we want for NLP integration tests. Thanks!

nik-mosaic:
You can also use LowPrecisionLayerNorm (or Fused LayerNorm, but that is being deprecated) for non-HF NLP models.

tests/test_full_nlp.py
dakinggg: discussed offline; the simple transformer is not traceable (no onnx) and doesn't have bert intermediate (no gatedlinearunits), but will add another algorithm for it.

dakinggg:
Turns out `LowPrecisionLayerNorm` is not scriptable, so I didn't do that.

tests/test_full_nlp.py
dakinggg: :+1:

PR #1831, opened 2022-12-16
Implement zero-shot Lambada evaluator in Composer

bmosaicml:
# What does this PR do?
This diff implements the Lambada zero-shot evaluator natively in composer using a custom dataset and metric. This is a major advantage over the Eleuther evaluation harness because it uses parallelization to achieve linear scaling on multi GPU, and is much faster than the Eleuther eval code even on signal GPU due to improvements in code quality and efficiency.

# Before submitting


composer/models/huggingface.py
```diff
@@ -296,7 +303,10 @@ def get_metrics(self, is_train: bool = False) -> Dict[str, Metric]:
         return metrics if metrics else {}
 
     def update_metric(self, batch: Any, outputs: Any, metric: Metric) -> None:
-        metric.update(outputs, self.labels)
+        if 'update_metric_handle' in batch:
+            batch['update_metric_handle'](metric, batch, outputs, self.labels)
+        else:
+            metric.update(outputs, self.labels)
```
abhi-mosaic, line 309: I think the `update_metric_handle` residing in the dataset is slightly against the design principle... could we not just check on `isinstance(metric, InContextLearningLMAccuracy)` and if so, pass in the `(batch, outputs, self.labels)`? The input args are a property of the _metric_, not the _dataset_, whose batches could be fed into many different metrics.

Also, I'm slightly confused why `self.labels` is a thing...  Shouldn't the labels come directly from the batch? Or is this using some HF model thing? Bit of a weird thing to see a `self` property on the model that is changing with each batch O.o

composer/models/huggingface.py
```diff
@@ -272,20 +275,24 @@ def loss(self, outputs, batch):
             return outputs[0]
 
     def eval_forward(self, batch, outputs: Optional[Any] = None):
-        output = outputs if outputs else self.forward(batch)
-        if self.use_logits:
+        if 'eval_forward_handle' in batch:
```
abhi-mosaic, line 278: @dakinggg is this the canon way of doing this now? Or a temporary workaround until we have some way of getting the eval dataloader name to the `eval_forward` function?

The idea of datasets defining model forward functions, getting autocast right, getting distributed stuff right, makes me nervous. It also means the dataset's `eval_forward_handle` is now tied to HF models only right?

dakinggg: My suggested approach right now is to define a custom collate function that packs a string "mode" into the batch, and then the `eval_forward` function uses that "mode" to determine which code path to run (this is what Sam did for his similar issue). I prefer this to actually packing the function reference into the batch. This deserves some docs that I have not written yet. As of right now, we don't have a plan to provide access to the `eval_dataloader.label` in `eval_forward`. I don't like the idea of changing the `eval_forward` function signature, but we could revisit if my suggestion above seems untenable.

As for your concerns about autocast, distributed, etc, I think it is ok because `eval_forward` gets called by composer inside all the correct contexts. Am I missing something here?

composer/models/huggingface.py
dakinggg: Agree with the first suggestion to check the type here rather than stashing the function reference on the batch.

I was also confused by the `self.labels` thing when I looked at this code. It is on my todo list to look at it and try to figure out why it was originally written this way and if it is necessary.

PR #1845, opened 2022-12-22
Feature/lambada evaluator

bmosaicml:
# What does this PR do?

This PR implements a composer-native evaluator for Lambada.

Previously we were dependent on the Eleuther eval harness to run Eval, which was implemented inefficiently and couldn't support multi GPU. We now support LAMBADA with multi-gpu and enjoy linear scaling w/ num GPUs.

This is related to the following JIRA https://mosaicml.atlassian.net/browse/RESEARCH-345

# Before submitting

[RESEARCH-345]: https://mosaicml.atlassian.net/browse/RESEARCH-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+                (context_enc + continuation_enc)[-(self.max_seq_len + 1):],
+                dtype=torch.long,
+            )
+            (inplen,) = inp.shape
+
+            # pad length from seq to padding_length
+            inp = torch.cat(
+                [
+                    inp,  # [seq]
+                    torch.LongTensor((self.max_seq_len - inplen) * [self.eos_tok_id]),  # [padding_length - seq]
```
dakinggg, line 63:
```suggestion
                    torch.LongTensor((self.max_seq_len - inplen) * [self.eos_tok_id]),  # [max_seq_len - seq]
```

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+
+        return {
+            'input_ids': torch.stack(inputs),
+            'continuation_indices': continuation_indices,
+            'mode': 'lm_task',
+            'labels': torch.stack(inputs),
+            'eos_tok_id': self.eos_tok_id
+        }
+
+    def get_num_samples_in_batch(self, batch) -> int:
```
dakinggg, line 79: What is this needed for? It looks like it is exactly the same as the default function: https://github.com/mosaicml/composer/blob/1a3f210e8a714a702e0f14409a93d4141f0d171c/composer/core/data_spec.py#L277

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+                    Please use a DataSpec and specify `get_num_samples_in_batch`."""))
+
+    def update_metric(self, metric, batch, output_logits, labels):
+        metric.update(batch, output_logits, labels)
+
+
+def get_lm_task_dataloader(dataset_uri, tokenizer, batch_size, max_seq_len, eos_tok_id):
+    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id)
+    sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
+    print(f'Using microbatch size {batch_size}')
```
dakinggg, line 112: Delete?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+from datasets import load_dataset
+from torch.utils.data import DataLoader, Dataset
+from transformers import AutoTokenizer
+
+from composer.core import DataSpec
+from composer.utils import dist, ensure_tuple
+from composer.utils.file_helpers import get_file
+
+
+class InContextLearningLMTaskDataset(Dataset):
```
dakinggg, line 16: Please add a docstring

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+
+import torch
+from datasets import load_dataset
+from torch.utils.data import DataLoader, Dataset
+from transformers import AutoTokenizer
+
+from composer.core import DataSpec
+from composer.utils import dist, ensure_tuple
+from composer.utils.file_helpers import get_file
+
```
dakinggg, line 14: Please add an `__all__` to this file

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+                textwrap.dedent(f"""\
+                    Cannot determine the batch size, as multiple Tensors of
+                    different lengths were found in the batch: sizes in batch: {dim0_sizes}.
+                    Please use a DataSpec and specify `get_num_samples_in_batch`."""))
+
+    def update_metric(self, metric, batch, output_logits, labels):
+        metric.update(batch, output_logits, labels)
+
+
+def get_lm_task_dataloader(dataset_uri, tokenizer, batch_size, max_seq_len, eos_tok_id):
```
dakinggg, line 109: Please add a docstring and argument types

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id)
+    sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
+    print(f'Using microbatch size {batch_size}')
+    return DataSpec(DataLoader(
```
dakinggg, line 113: I think you don't need to wrap this in `DataSpec`. Did you find that you did need to for some reason?

dakinggg (commented):
Left a bunch of comments, mostly about integrating everything nicely with composer, and other clean code stuff like tests and docstrings. The implementation itself looks correct to me. Happy to help with any of the comments related to integrating it with composer if they don't make sense or are difficult!

composer/models/huggingface.py
```diff
@@ -272,20 +273,30 @@ def loss(self, outputs, batch):
...
-        if self.use_logits:
+        if 'mode' in batch and batch['mode'] == 'lm_task':
             self.labels = batch.pop('labels')
-            if self.config.use_return_dict:
-                output = output['logits']
-            else:
-                # logits are at index 1 in the output tuple
-                output = output[1]
+            args = {'input_ids': batch['input_ids']}
+            args['attention_mask'] = ~(batch['input_ids'] == batch['eos_tok_id'])
```
abhi-mosaic, line 279: I think it would be safer if `attention_mask` was part of the batch, to match the same spec that HF default data collators produce. So basically, this line would move to `in_context_learning_evaluation.py`, and inside here, we wouldn't need to construct `args`, and just call `self.forward(batch)` on a HF-standard batch.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(
+        self,
+        dataset_uri: str,
+        tokenizer: AutoTokenizer,
+        max_seq_len: int,
+        eos_tok_id: int,
+        destination_path: str = 'icl_lm_task.json',
+    ):
+        get_file(dataset_uri, destination_path, overwrite=True)
```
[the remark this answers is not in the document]
bmosaicml: I think it's probably safer to overwrite just in case the existing file contains data for a different eval task (even thought right now I only designed this w/ LAMBADA in mind)

composer/models/huggingface.py
bmosaicml: The batch contains a bunch of keys that wouldn't be used by forward but are necessary for computing the metric, including the continuation indices, so I will need to do something special before calling forward either way.

One potential alternative is to group the input ids and attention mask into their own value within the batch, maybe under the key 'forward_input'?

composer/datasets/in_context_learning_evaluation.py
bmosaicml: The only difference is this little section ` if isinstance(t, torch.Tensor)` to ensure we don't try taking the shape of any of the non-tensor values in the batch dict

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,120 @@
...
+        inputs = []
+        continuation_indices = []
+        for data_pair in data:
+            context, continuation = data_pair['context'], data_pair['continuation']
+
+            context_enc = context['input_ids']
+            continuation_enc = continuation['input_ids']
+            continuation_span = torch.tensor(range(len(context_enc), len(context_enc) + len(continuation_enc)))
+
+            inp = torch.tensor(
```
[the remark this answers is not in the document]
bmosaicml: Yeah that's right! I'll add a comment

composer/models/huggingface.py
abhi-mosaic: I think it makes sense if batches of type `mode='lm_task'` have a unique format, but it still feels like that format should be enforced by the dataset+collate_fn. Would different models need to create different attention_masks? I'm basically just trying to move this line: `args['attention_mask'] = ~(batch['input_ids'] == batch['eos_tok_id'])` so it doesn't have to be written in every model's `eval_forward()`.

Understood about using a subset of the batch keys, I think that's what we do in our [HF example](https://github.com/mosaicml/examples/blob/fd087710737ee56d0efed7671c6b7238961c0c0d/llm/src/hf_causal_lm.py#L66-L68) and it seems fine.

composer/core/data_spec.py
```diff
@@ -214,15 +214,15 @@ class DataSpec:
 
     def __init__(
         self,
-        dataloader: Iterable,
+        dataloader: Union[Iterable, torch.utils.data.DataLoader],
```
abhi-mosaic, line 217: is `torch.utils.data.DataLoader` not a subclass of `Iterable` 😮 ? Was this always broken but only now getting caught by PyRight?

composer/trainer/trainer.py
```diff
@@ -36,6 +36,7 @@
 from composer.devices import Device, DeviceCPU, DeviceGPU, DeviceMPS, DeviceTPU
 from composer.loggers import (ConsoleLogger, Logger, LoggerDestination, ProgressBarLogger, RemoteUploaderDownloader,
                               WandBLogger)
+from composer.metrics import METRIC_DEFAULT_CTORS
```
abhi-mosaic, line 39: I'll admit I'm a bit confused why we need to modify `trainer.py`... do default metrics make sense for Composer? I'm not sure why we would want 'InContextLearningLMAccuracy' if training diffusion models, for instance.

bmosaicml: My main concern was the awkwardness of having the user specify the metric in two places (in the eval data loader and in the model) and my concern about whether that would cause issues if it were always specified in the model but only optionally instantiated with the eval data loader. Though upon second glance at the code I'm not sure if that latter concern holds.

composer/core/data_spec.py
bmosaicml: I think so 😮

composer/trainer/trainer.py
dakinggg: yeah, without a pretty compelling reason, I would avoid putting in this default metric constructor thing. The intention is that models define the instantiation of their metrics, and evaluators specify which of those metrics to use. If there is an issue with this design, we should def revisit, but probably don't want to change the design as part of this PR.

tests/metrics/test_nlp_metrics.py
```diff
@@ -200,3 +201,22 @@ def test_perplexity():
...
+    start = len(tiny_gpt2_tokenizer('The dog is')['input_ids'])
+    end = len(tiny_gpt2_tokenizer('The dog is')['input_ids']) + len(tiny_gpt2_tokenizer(' furry')['input_ids'])
+    cont_idxs = torch.tensor(list(range(start, end)))
+
+    input = torch.tensor([input + [pad] * (2048 - len(input))])
+
+    batch = {'continuation_indices': [cont_idxs], 'labels': input, 'input_ids': input}
+    logits = torch.nn.functional.one_hot(torch.roll(input, shifts=-1), num_classes=pad + 1)
+
+    metric = InContextLearningLMAccuracy()
```
dakinggg, line 219: Maybe add another one that tests something other than 1? Like an artificial logits that get you 50% accuracy. and maybe have a batch size greater than 1. just to cover the scenarios a little bit more.

composer/metrics/nlp.py
```diff
@@ -231,3 +234,49 @@ def compute(self) -> Tensor:
         """Returns torch.exp() of the LanguageCrossEntropyLoss."""
         avg_loss = super().compute()
         return torch.exp(avg_loss)
+
+
+class InContextLearningLMAccuracy(Metric):
+    r"""Computes accuracy for in-context learning language modeling tasks.
```
dakinggg, line 240:
```suggestion
    r"""Computes accuracy for in-context learning (ICL) language modeling (LM) tasks.
```

composer/metrics/nlp.py
```diff
@@ -231,3 +234,49 @@ def compute(self) -> Tensor:
...
+
+    ICL LM tasks consist of some number of example language modeling tasks (referred to as the 'context'), followed by a test task where the model must correctly predict all the tokens
+    following tokens in some passage (referred to as the 'continuation').
+
+    For example, the model may be provided the context below and evaluated on its ability to correctly predict the continuation. Note it doesn't matter
```
dakinggg, line 245:
```suggestion
    For example, the model may be provided the context below and evaluated on its ability to correctly predict the continuation. Note: it doesn't matter
```

composer/datasets/in_context_learning_evaluation.py
dakinggg: Looks like you didn't quite finish this one :)

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,123 @@
...
+def get_lm_task_dataloader(dataset_uri: str, tokenizer: Union[transformers.PreTrainedTokenizer,
+                                                              transformers.PreTrainedTokenizerFast], batch_size: int,
+                           max_seq_len: int, eos_tok_id: int) -> DataSpec:
+    """This constructs a dataloader capable of evaluating LLMs on in-context learning language modeling tasks, for example LAMBADA. An example usage is below:
+
+    dl = get_lm_task_dataloader(dataset_uri, tokenizer, 2, max_seq_len=2048, eos_tok_id=tokenizer.eos_token_id)
+    evaluator = Evaluator(label='lambada', dataloader=dl, metric_names=['InContextLearningLMAccuracy'])
+
+    Args:
+        dataset_uri (str): Either a local path, or a remote path (S3 or GCS)
```
dakinggg, line 106:
```suggestion
        dataset_uri (str): Either a local path, or a remote path beginning with ``s3://``, or another backend
                supported by :meth:`composer.utils.maybe_create_object_store_from_uri`.
```

tests/metrics/test_nlp_metrics.py
```diff
@@ -200,3 +201,22 @@ def test_perplexity():
     perplexity = perplexity_metric.compute()
 
     assert torch.equal(torch.exp(ce), perplexity)
+
+
+def test_in_context_learning_lm_accuracy(tiny_gpt2_tokenizer):
+    pad = tiny_gpt2_tokenizer.eos_token_id
```
dakinggg, line 207: The test fixture actually has a pad token, so I think you should be able to use that: https://github.com/mosaicml/composer/blob/bb2b12bd0235318325186a542904669fd4616863/tests/fixtures/fixtures.py#L170

composer/models/huggingface.py
```diff
@@ -273,7 +276,7 @@ def loss(self, outputs, batch):
 
     def eval_forward(self, batch, outputs: Optional[Any] = None):
         output = outputs if outputs else self.forward(batch)
-        if self.use_logits:
+        if self.use_logits or ('mode' in batch and batch['mode'] == 'lm_task'):
```
dakinggg, line 279:
```suggestion
        if self.use_logits or batch.get('mode', None) == 'lm_task':
```

dakinggg (approved):

tests/datasets/local_data/lambada_small.jsonz
```diff
@@ -0,0 +1,16 @@
+{"context": "With Tristran's next step he was standing beside a lake, and the candlelight shone brightly on the water; and then he was walking through the mountains, through lonely crags, where the candlelight was reflected in the eyes of the creatures of the high snows; and then he was walking through the clouds, which, while not entirely substantial, still supported his weight in comfort; and then, holding tightly to his candle, he was underground, and the candlelight glinted back at him from the wet cave walls; now he was in the mountains once more; and then he was on a road through wild forest, and he glimpsed a chariot being pulled by two goats, being driven by a woman in a red dress who looked, for the glimpse he got of her, the way Boadicea was drawn in his history books; and another step and he was in a leafy glen, and he could hear the chuckle of water as it splashed and sang its way into a small brook.\n\nHe took another step, but he was still in the", "continuation": " glen"}
```
knighton, line 1: `jsonz`? would have called it `jsonl`, just curious

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -0,0 +1,141 @@
...
+                )[-(self.max_seq_len + 1):],  # trim from the left if context + continuation are larger than max_seq_len
+                dtype=torch.long,
+            )
+            (inplen,) = inp.shape
+
+            # pad length from seq to padding_length
+            inp = torch.cat(
+                [
+                    inp,  # [seq]
+                    torch.LongTensor((self.max_seq_len - inplen) * [self.eos_tok_id]),
```
knighton, line 78: `inp_len`?

knighton (approved):

PR #1854, opened 2022-12-30
Update Docker images to use ‘posix_prefix’ paths

mvpatel2000:
# What does this PR do?

For Debian distros, Python 3.10 sysconfig default install paths have changed to `posix_local` instead of `posix_prefix` per:
https://lists.debian.org/debian-python/2022/03/msg00039.html

This is an issue with Triton which specifically uses `sysconfig.get_paths()` to determine install paths.  This PR sets the `DEB_PYTHON_INSTALL_LAYOUT=deb` to force previous default behavior in the Python 3.10 image.

Fixes [CO-1605](https://mosaicml.atlassian.net/browse/CO-1605)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -159,7 +159,7 @@ RUN \
 ARG PYTHON_VERSION
 
 # Python 3.10 changes where packages are installed. Workaround until all packages support new format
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
LGTM

bandish-shah (approved):
LGTM

PR #1855, opened 2023-01-04
Simpler auto log hparams

eracah:
# What does this PR do?

logs just hparams passed directly to Trainer. Any non-builtin objects are just logged as `obj.__class__.__name__` instead of crawling them for haprams

fix CO-586

Manual Test:
```python
train_set = datasets.MNIST("data", train=True, download=True, transform=transform)
val_set = datasets.MNIST("data", train=False, download=True, transform=transform)
train_dataloader = DataLoader(train_set, batch_size=128)
eval_dataloader = DataLoader(val_set, batch_size=64)
model=mnist_model(num_classes=10)


trainer = Trainer(
    model=model,
    max_duration="3ep",
    train_subset_num_batches=4,
    optimizers=[Adam(model.parameters(), eps=1e-3)],
    progress_bar=False,
    log_traces=False,
    log_to_console=True,
    console_log_interval='1ep',
    auto_log_hparams=True,
    algorithms=[ChannelsLast(), ColOut(), BlurPool()],
    callbacks=[SpeedMonitor(), MemoryMonitor(), ImageVisualizer()],
    loggers=[WandBLogger()],
)
trainer.fit()
```

[Wandb Result](https://wandb.ai/mosaic-ml/evan-composer-manual_tests/runs/23qgcv6v)
Console result:
```
Config:
algorithm_passes: null
algorithms:
- ChannelsLast
- ColOut
- BlurPool
auto_log_hparams: true
auto_microbatching: false
autoresume: false
blurpool/num_blurconv_layers: 0
blurpool/num_blurpool_layers: 0
callbacks:
- SpeedMonitor
- MemoryMonitor
- ImageVisualizer
console_log_interval: 1ep
console_stream: stderr
ddp_sync_strategy: null
deepspeed_config: null
deterministic_mode: false
device: DeviceCPU
device_train_microbatch_size: null
dist_timeout: 1800.0
eval_batch_split: 1
eval_dataloader: DataLoader
eval_interval: 1
eval_subset_num_batches: -1
fsdp_config: null
grad_accum: 1
latest_remote_file_name: null
load_exclude_algorithms: null
load_ignore_keys: null
load_object_store: null
load_path: null
load_progress_bar: true
load_weights_only: false
log_to_console: true
log_traces: false
loggers:
- WandBLogger
- ConsoleLogger
max_duration: 3ep
model: ComposerClassifier
num_cpus_per_node: 1
num_nodes: 1
num_optimizers: 1
optimizers:
- Adam
precision: Precision
profiler: null
progress_bar: false
python_log_level: null
rank_zero_seed: 791585187
remote_ud_has_format_string:
- false
- false
run_name: 1672804048-impossible-viper
save_filename: ep{epoch}-ba{batch}-rank{rank}.pt
save_folder: null
save_interval: 1ep
save_latest_filename: latest-rank{rank}.pt
save_num_checkpoints_to_keep: -1
save_overwrite: false
save_weights_only: false
scale_schedule_ratio: 1.0
schedulers: null
seed: 791585187
step_schedulers_every_batch: null
train_dataloader: DataLoader
train_dataloader_label: train
train_subset_num_batches: 4
using_device_microbatch_size: false
```

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -44,8 +44,8 @@ from composer.trainer._scale_schedule import scale_pytorch_scheduler
 from composer.trainer._scaler import ClosureGradScaler
 from composer.trainer.dist_strategy import DDPSyncStrategy, ddp_sync_context, prepare_ddp_module, prepare_fsdp_module
 from composer.utils import (ExportFormat, MissingConditionalImportError, ObjectStore, Transform, checkpoint, dist,
-                            ensure_tuple, export_with_logger, format_name_with_dist, get_device, get_file,
-                            is_tpu_installed, map_collection, maybe_create_object_store_from_uri,
+                            ensure_tuple, export_with_logger, extract_hparams, format_name_with_dist, get_device,
+                            get_file, is_tpu_installed, map_collection, maybe_create_object_store_from_uri,
                             maybe_create_remote_uploader_downloader_from_uri, model_eval_mode, parse_uri,
                             reproducibility)
 
@@ -585,6 +585,7 @@ class Trainer:
 
             Set to ``0`` to disable metrics logging to console.
         log_traces (bool): Whether to log traces or not. (default: ``False``)
+        auto_log_hparams (bool): Whether to automatically extract hyperparameters. (default: ``False``)
         load_path (str, optional):  The path format string to an existing checkpoint file.
 
             It can be a path to a file on the local disk, a URL, or if ``load_object_store`` is set, the object name
@@ -882,6 +883,7 @@ class Trainer:
         console_stream: Union[str, TextIO] = 'stderr',
         console_log_interval: Union[int, str, Time] = '1ep',
         log_traces: bool = False,
+        auto_log_hparams: bool = False,
 
         # Load Checkpoint
         load_path: Optional[str] = None,
@@ -929,6 +931,7 @@ class Trainer:
         python_log_level: Optional[str] = None,
     ):
 
+        self.auto_log_hparams = auto_log_hparams
         self.python_log_level = python_log_level
         if self.python_log_level is not None:
             logging.basicConfig(
@@ -1175,6 +1178,11 @@ class Trainer:
 
         # Run Event.INIT
         self.engine.run_event(Event.INIT)
+
+        # Log hparams.
+        if self.auto_log_hparams:
+            self.local_hparams = extract_hparams(locals())
+            self.logger.log_hyperparameters(self.local_hparams)
 
         # Log gpus and nodes.
         device_name = self.state.device.__class__.__name__.lstrip('Device').lower()
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1891,9 +1891,6 @@ class Trainer:
         """Run training for the specified number of epochs and log results."""
         # print training start
         log.info('Using precision %s', self.state.precision)
-        self.logger.log_hyperparameters(
-            {'enabled_algorithms/' + algo.__class__.__name__: True for algo in self.state.algorithms})
-
         assert self.state.dataloader is not None, 'dataloader is set in __init__() or fit()'
         assert self._train_data_spec is not None, 'The train data spec is set in __init__() or fit()'
         assert self.state.scaler is not None, 'scaler should have been set in __init__()'
--- a/composer/utils/auto_log_hparams.py
+++ b/composer/utils/auto_log_hparams.py
@@ -1,0 +1,105 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Contains helper functions for auto-logging hparams."""
+
+from typing import Any, Dict, List, Tuple
+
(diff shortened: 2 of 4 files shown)


composer/trainer/trainer.py
```diff
@@ -1176,6 +1179,11 @@ def __init__(
         # Run Event.INIT
         self.engine.run_event(Event.INIT)
 
+        # Log hparams.
+        if self.auto_log_hparams:
+            self.local_hparams = extract_hparams(locals())
```
dakinggg, line 1184: I think we need to deepcopy locals here, or be very careful not to modify it, otherwise could have unexpected side effects. And we probably don't want to deepcopy it, as that could be a costly operation. Thoughts?

composer/utils/auto_log_hparams.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Returns:
+        Dict[str, Any]: A nested dictionary with every element of locals_dict mapped to its
+            value or to another sub_dict.
+    """
+    hparams = {}
+    if 'self' in locals_dict:
+        locals_dict.pop('self')
+
+    for k, v in locals_dict.items():
```
dakinggg, line 26: should we exclude the key `_`? It showed up for me when I ran it locally

tests/utils/test_autolog_hparams.py
```diff
@@ -0,0 +1,175 @@
...
+        model=model,
+        train_dataloader=train_dl,
+        optimizers=optimizer,
+        auto_log_hparams=True,
+        progress_bar=False,
+        log_to_console=False,
+        run_name='test',
+        seed=3,
+        algorithms=[EMA()],
+        loggers=[MLFlowLogger()],
```
dakinggg, line 79: Maybe do `InMemoryLogger`, otherwise this tests needs a conditional check if mlflow is installed.

tests/utils/test_autolog_hparams.py
```diff
@@ -0,0 +1,175 @@
...
+        # Distributed Training
+        'dist_timeout': 1800.0,
+        'ddp_sync_strategy': None,
+
+        # Profiling
+        'profiler': None,
+
+        # Python logging
+        'python_log_level': None,
+        'auto_microbatching': False,
```
dakinggg, line 166: It is a bit weird that these local variables from the trainer init function show up...but I guess its fine?

composer/trainer/trainer.py
```diff
@@ -1883,9 +1891,6 @@ def _train_loop(self) -> None:
         """Run training for the specified number of epochs and log results."""
         # print training start
         log.info('Using precision %s', self.state.precision)
-        self.logger.log_hyperparameters(
```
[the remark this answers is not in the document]
eracah: this currently isn't used anywhere though right?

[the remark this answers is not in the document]
dakinggg: What do you mean by "isn't used"? It always shows up in the logs right now, right?

composer/trainer/trainer.py
eracah: I think if we only log the original object if it is int, str, or float and otherwise log `obj.__class__.__name__`, then we are good because:

a. we don't log locals() directly, we make a new dictionary.
b. any obj we log directly from locals() is immutable (int, float, str).

tests/utils/test_autolog_hparams.py
eracah: I think it's fine. we could move the auto-logging to the top of __init__ before those variables are created? wdyt?

composer/trainer/trainer.py
dakinggg: The first thing that happens inside of `extract_params` is that `self` is popped from the input dict though. Maybe everything else is fine if that line is changed to not pop, but rather just skip that key.

tests/utils/test_autolog_hparams.py
dakinggg: It is probably fine either way. If it stays here than any modifications that the trainer init makes to inputs will be captured by the logging. That is probably a good thing.

tests/utils/test_autolog_hparams.py
eracah: fixed

composer/utils/auto_log_hparams.py
eracah: fixed

tests/utils/test_autolog_hparams.py
eracah: ok sg

composer/trainer/trainer.py
eracah: ooohhhh good call. fixed

composer/trainer/trainer.py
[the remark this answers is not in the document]
eracah: oh I see I was gonna remove it because it is obviated by the auto-logging of algorithms, but I guess the user could not have auto-logging on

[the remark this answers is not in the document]
eracah: added it back in

dakinggg (approved):
LGTM

eracah:
> LGTM
Thanks!

PR #1870, opened 2023-01-09
Daily tests

mvpatel2000:
# What does this PR do?

Migrates daily tests to GHA. Now, tests marked with `@pytest.mark.daily` will run every time a PR is merged into dev/main and every night at 2:30am (github recommends picking a randomish time as opposed to midnight to avoid load spikes).

Along for the ride, a few tweaks to notebook tests to speed them up and avoid OOMs in the smaller GHA machines.

To verify tests passed, I opened a PR in my fork and changed workflow to run on pull requests as well. https://github.com/mvpatel2000/composer/pull/26 Commit `change workflows` has all tests passing before I set it to only run on merges (push) and scheduled cron

Note: This is still only CPU tests. GPU tests will be added later

Note: We currently run all `pytest.mark.remote` tests only in the daily tests because hitting real servers is expensive. We maintain the `remote` marker in this PR though since this decision might be changed later.

[CO-1519](https://mosaicml.atlassian.net/browse/CO-1519)

[CO-1519]: https://mosaicml.atlassian.net/browse/CO-1519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dakinggg (approved):
LGTM, with one question. Should the daily be running remote tests too? I think yes, and it looks like you have it set up for that with all of the secrets youre passing.

bandish-shah (approved):
LGTM

PR #1875, opened 2023-01-10
Disable new notebook in CI

mvpatel2000:
# What does this PR do?

Disables new notebook test in CI flow

Also removes fast fail on daily tests since we want to see all faillures

Also bumps images to torch 1.13.1 now that we've switched over


dakinggg (approved):

bandish-shah (approved):
LGTM

PR #1876, opened 2023-01-10
Add few shot and multiple choice to ICL evaluation

bmosaicml:
# What does this PR do?
This PR extends the existing ICL-LM evaluator framework to support few shot and multiple choice (e.g. PIQA)

I create a custom multiple choice in-context learning data loader, a special NLP metric of ICL-multiple choice, and extend the existing ICL-LM data loader to support few shot prompting.

Manually evaluation GPT Neo confirms Lambada 57.23% and PIQA 71%

# Before submitting
- [x ] Did you update any related docs and document your change?
- [x ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [x ] Did you run the tests locally to make sure they pass?
- [ x] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -100,9 +142,168 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+        return self.encoded_dataset[index]
+
+    def __len__(self):
+        return len(self.encoded_dataset)
+
+    def collate_fn(self, data):
+        inputs = []
+        continuation_indices = []
+        gold_idxs = []
+        choice_groupings = []
```
dakinggg, line 217: could you comment explaining the `choice_groupings` list? That one in particular took me a bit to understand that it was indexing into `continuation_indices` because each choice is its own continuation.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -100,9 +142,168 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+        preamble_string: str,  # e.g. 'translate english to french:'
+        example_delimiter: str,  # e.g. '\n'
+        continuation_delimiter: str,  # e.g. ''
+) -> DataSpec:
+    dataset = InContextLearningMultipleChoiceTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id, num_fewshot,
+                                                         preamble_string, example_delimiter, continuation_delimiter)
+    sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
+    batch_size = max(dataset.num_choices, batch_size)
+    effective_batchsize = batch_size // dataset.num_choices
+    print(f'Using microbatch size {effective_batchsize}')
```
dakinggg, line 285: Please either use a logger or remove the print statement

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -130,12 +331,15 @@ def get_lm_task_dataloader(dataset_uri: str, tokenizer: Union[transformers.PreTr
     Returns:
         DataLoader: A dataloader used for performing in-context learning evaluation on the dataset provided.
     """
-    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id)
+    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id, num_fewshot,
+                                             preamble_string, example_delimiter, continuation_delimiter)
     sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
+    print(f'Using microbatch size {batch_size}')
```
dakinggg, line 337: Please either use a logger or remove the print statement

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -100,9 +142,168 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+        max_seq_len: int,
+        eos_tok_id: int,
+        num_fewshot: int,
+        preamble_string: str,  # e.g. 'translate english to french:'
+        example_delimiter: str,  # e.g. '\n'
+        continuation_delimiter: str,  # e.g. ''
+) -> DataSpec:
+    dataset = InContextLearningMultipleChoiceTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id, num_fewshot,
+                                                         preamble_string, example_delimiter, continuation_delimiter)
+    sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
```
dakinggg, line 282: since this is an eval dataset, should shuffle be false?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -130,12 +331,15 @@ def get_lm_task_dataloader(dataset_uri: str, tokenizer: Union[transformers.PreTr
     Returns:
         DataLoader: A dataloader used for performing in-context learning evaluation on the dataset provided.
     """
-    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id)
+    dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, eos_tok_id, num_fewshot,
+                                             preamble_string, example_delimiter, continuation_delimiter)
     sampler = dist.get_sampler(dataset, drop_last=False, shuffle=True)
```
dakinggg, line 336: same as above on shuffle

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -0,0 +1,118 @@
...
+                                tokenizer,
+                                2,
+                                max_seq_len=2048,
+                                eos_tok_id=tokenizer.eos_token_id,
+                                num_fewshot=1,
+                                preamble_string='',
+                                example_delimiter='\n',
+                                continuation_delimiter='')
+
+    assert isinstance(dl.dataloader, DataLoader)  # pyright
```
dakinggg, line 33: I think it may be worth it to actually write out the expected value of each of these keys for both of these dataloader tests, for the first example. otherwise they are just testing that code doesn't crash and that the dictionary returned is correctly formed. As the dataset got more complicated with fewshot, I think this became more necessary.

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -0,0 +1,118 @@
...
+    evaluator = Evaluator(label='lambada', dataloader=dl, metric_names=['InContextLearningLMAccuracy'])
+    model = create_gpt2(use_pretrained=False, pretrained_model_name='EleutherAI/gpt-neo-125M')
+    model.add_eval_metrics(evaluator)
+    trainer = Trainer(model=model, max_duration='1ep', loggers=in_memory_logger)
+    trainer.eval(eval_dataloader=evaluator)
+    assert 'metrics/lambada/InContextLearningLMAccuracy' in in_memory_logger.data.keys()
+    assert in_memory_logger.data['metrics/lambada/InContextLearningLMAccuracy'][0][1].item() == 0
+
+
+@pytest.mark.parametrize('dataset_uri', ['piqa_small.jsonl', 'hellaswag_small.jsonl'])
```
dakinggg, line 94: Is there a need to parametrize over the two datasets? They look of the same format to me.

composer/metrics/nlp.py
```diff
@@ -247,12 +253,42 @@ class InContextLearningLMAccuracy(Metric):
...
+    """Computes accuracy with support for masked indicies.
 
     Adds metric state variables:
-        correct (float): The number of examples where the model correctly predicted the whole continuation.
-        total (float): The number of total examples seen.
+        correct (float): The number of instances where the prediction masked the target.
+        total (float): The number of total instances that were predicted.
 
     Args:
+        ignore_index (int): The class index to ignore. Default: -100.
```
dakinggg, line 291: I think this was added to the docstring by mistake?

dakinggg (approved):
Approving to unblock since I will be gone for the next week, but left a bunch of comments.

mvpatel2000:
Same as https://github.com/mosaicml/composer/pull/1879, please ensure that new tests are not slower than 5s on CPU/GPU!

composer/metrics/nlp.py
```diff
@@ -247,12 +253,42 @@ class InContextLearningLMAccuracy(Metric):
...
+    def compute(self):
+        assert isinstance(self.correct, Tensor)
+        assert isinstance(self.total, Tensor)
+        return self.correct.float() / self.total
+
+
+class InContextLearningMultipleChoiceAccuracy(InContextLearningMetric):
+    """Computes accuracy with support for masked indicies.
 
     Adds metric state variables:
```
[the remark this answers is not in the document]
bmosaicml: Good catch, this was a merging mistake

tests/datasets/test_in_context_learning_datasets.py
bmosaicml: They test different number of answers per Q (PIQA has two choices, hellaswag 4)

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -109,12 +198,193 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+        # gold_indices indicates which of the [0, N-1] choices is the correct one for each question.
+        batch = {
+            'input_ids': torch.stack(inputs),
+            'continuation_indices': continuation_indices,
+            'mode': 'icl_task',
+            'labels': torch.stack(inputs),
+            'gold_indices': gold_idxs,
+            'choice_groupings': choice_groupings
+        }
+        batch['attention_mask'] = ~(batch['input_ids'] == self.eos_tok_id)
```
abhi-mosaic, line 352: I don't think this is quite right, like I think `attention_mask` is meant to be `1` where there are tokens and `0` where there is padding. I think the `eos_token_id` is an artifact of the GPT2Tokenizer but not a general thing.

I think the safest thing todo here is to copy what HF's default data collator does (maybe even import and use directly!): https://github.com/huggingface/transformers/blob/44caf4f6f47120a4aaca561c9cc8041455bef705/src/transformers/data/data_collator.py#L1152-L1193

abhi-mosaic, line 352: I don't think this is quite right, like I think `attention_mask` is meant to be `1` where there are tokens and `0` where there is padding. I think the `eos_token_id` is an artifact of the GPT2Tokenizer but not a general thing.

I think the safest thing todo here is to copy what HF's default data collator does (maybe even import and use directly!): https://github.com/huggingface/transformers/blob/44caf4f6f47120a4aaca561c9cc8041455bef705/src/transformers/data/data_collator.py#L1152-L1193

composer/trainer/trainer.py
```diff
@@ -2552,7 +2552,6 @@ def eval(
 
         """
         if eval_dataloader is not None:
```
abhi-mosaic, line 2554: not sure if this is the linter acting up but probably best to `git checkout dev -- trainer/trainer.py` if no changes intended.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -109,12 +198,193 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+
+            choice_start_idx = len(continuation_indices)
+            preamble, context, choices, gold_idx = (data_pair['preamble'], data_pair['query'], data_pair['choices'],
+                                                    data_pair['gold_idx'])
+
+            for choice in choices:
+                context_enc = preamble['input_ids'] + context['input_ids']
+                continuation_enc = choice['input_ids']
+                inp, continuation_span = _make_padded_input(context_enc, continuation_enc, self.max_seq_len,
+                                                            self.eos_tok_id)
```
abhi-mosaic, line 328: Related to below, is there a reason we are padding with the `eos_tok_id` rather than the `pad_tok_id`? I guess if it's consistent all the way through that's fine, but it might be good to add a comment about it.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -109,12 +198,193 @@ def update_metric(self, metric, batch, output_logits, labels):
...
+            'choice_groupings': choice_groupings
+        }
+        batch['attention_mask'] = ~(batch['input_ids'] == self.eos_tok_id)
+        return batch
+
+    def get_num_samples_in_batch(self, batch) -> int:
+        return batch['input_ids'].shape[0]
+
+    def update_metric(self, metric, batch, output_logits, labels):
+        metric.update(batch, output_logits, labels)
```
abhi-mosaic, line 359: Are these functions necessary? I thought Composer could handle `get_num_samples_in_batch` HF style batches automatically.

The `update_metric` doesn't seem to be used.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -135,16 +405,35 @@ def get_lm_task_dataloader(dataset_uri: str, tokenizer: Union[transformers.PreTr
...
+
     return DataSpec(DataLoader(
         dataset,
-        batch_size=batch_size,
+        batch_size=effective_batchsize,
         sampler=sampler,
         collate_fn=dataset.collate_fn,
     ),
+                    device_transforms=None,
                     get_num_samples_in_batch=dataset.get_num_samples_in_batch)
```
abhi-mosaic, line 439: Related to above, if we can get away with Composer's default `get_num_samples_in_batch` function, then we can skip the `DataSpec` wrapper entirely and just return a `DataLoader`. But ofc if the custom function is needed, then no changes here.

abhi-mosaic (commented):
Looking good! I added some comments about `eos` padding and some (maybe) superfluous functions. @bmosaicml I think it would also be good to get 1 more NLP team reveiwer who is more familiar with eval to look over the tests... I am not an expert here so I could have missed something.

composer/datasets/in_context_learning_evaluation.py
dakinggg: I'm guessing that the default ends up raising here https://github.com/mosaicml/composer/blob/c99230435cd23728e9e5d8cbdc62bc92cbbaa64d/composer/core/data_spec.py#L297 or maybe even crashing here (https://github.com/mosaicml/composer/blob/c99230435cd23728e9e5d8cbdc62bc92cbbaa64d/composer/core/data_spec.py#L292) because of the non tensor items in the dictionary. Possible the default should get slightly smarter (i.e. look for `input_ids`, or skip non tensor items in the batch). Made a JIRA to look into this https://mosaicml.atlassian.net/browse/CO-1660.

dakinggg: `update_metric` is what composer calls to update metrics. It isn't implemented on the base class (maybe it should be?), so any subclass of `ComposerModel` that wants to do eval needs to implement it.

composer/datasets/in_context_learning_evaluation.py
bmosaicml: I think their collator is more or less the same logic, where they set the attention mask to be equal to the places where they did/didn't insert padding tokens.

I inserted the eos_tok_id tokens as padding in my collation function. As long as the eos tok is never used to have real semantic meaning then I think we would be good.

composer/datasets/in_context_learning_evaluation.py
bmosaicml: yep you're right about the update metric function being superfluous and needing to delete it. Also Daniel is right about why the default doesn't work!

composer/datasets/in_context_learning_evaluation.py
bmosaicml: The actual token used by the data loader can be the pad token or the eos token, so for clarity I will change the internal naming to pad token. For the purposes of constructing the data loader, the GPTTokenizer's eos_tok_id is the last id in the GPT embedding, the pad token is the index after, and does not appear in the embedding, hence we can't actually pass in the pad_token_id in our tensors

dakinggg (approved):
LGTM, maybe would be good to have another NLP person have a glance. Also trusting that you and abhi worked out that attention mask thing, I didn't look at that.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -134,17 +398,36 @@ def get_lm_task_dataloader(dataset_uri: str, tokenizer: Union[transformers.PreTr
...
+    elif icl_task_type == 'language_modeling':
+        dataset = InContextLearningLMTaskDataset(dataset_uri, tokenizer, max_seq_len, pad_tok_id, num_fewshot,
+                                                 prompt_string, example_delimiter, continuation_delimiter)
+        effective_batchsize = batch_size
+    else:
+        raise Exception(f'Unrecognized ICL task type: {icl_task_type}')
+
+    sampler = dist.get_sampler(dataset, drop_last=False, shuffle=False)
+
     return DataSpec(DataLoader(
```
dakinggg, line 426: as discussed offline, for MC, suggest overriding the split batch function with something that raises an error, to make absolutely sure that batches don't get split in a way your code doesn't expect. Another option is to restructure the batch so that the second dimension is actually splittable, but we can defer that since its more complicated.

dakinggg: https://mosaicml.atlassian.net/browse/CO-1694

composer/models/huggingface.py
```diff
@@ -274,9 +274,9 @@ def loss(self, outputs, batch):
             return outputs[0]
 
     def eval_forward(self, batch, outputs: Optional[Any] = None):
+        self.labels = batch.pop('labels')
```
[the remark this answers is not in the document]
bmosaicml: yeah basically, if the labels are part of the batch when you call forward, then the HF model will calculate cross entropy loss, which is a waste of time for the ICL eval metrics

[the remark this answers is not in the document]
dakinggg: Oh, I see. I think this might break some other functionality though, because other uses might assume that the output will contain the loss if `self.use_logits` is `False`. do you think it would make sense to only pop the labels before calling `forward` `if self.use_logits or batch.get('mode', None) == 'icl_task':`?

composer/models/huggingface.py
```diff
@@ -42,10 +42,6 @@ class HuggingFaceModel(ComposerModel):
                 using :meth:`HuggingFaceModel.hf_from_composer_checkpoint`. If the tokenizer is not provided here, it will not be saved in the composer checkpoint.
         use_logits (bool, optional): If True, the model's output logits will be used to calculate validation metrics. Else, metrics will be inferred from the HuggingFaceModel directly. Default: ``False``
         metrics (list[Metric], optional): list of torchmetrics to apply to the output of `validate`. Default: ``None``.
-        shift_labels (bool, optional): If True, the batch's labels will be shifted before being used to calculate metrics. This should be set to true for CausalLM models and false otherwise. If not specified, `shift_labels` will be set automatically based on the model class name. Default: ``None``.
```
dakinggg, line 45: looks like a bunch of stuff got deleted in a bad merge

composer/metrics/nlp.py
```diff
@@ -274,7 +292,63 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
+
+    def __init__(self, dist_sync_on_step: bool = False):
+        # state from multiple processes
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.add_state('correct', default=torch.tensor(0.0), dist_reduce_fx='sum')
+        self.add_state('total', default=torch.tensor(0.0), dist_reduce_fx='sum')
+
+    def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+        targets = torch.roll(labels, shifts=-1)
+        targets[:, -1] = -100
```
alextrott16, line 335: Same as above

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -4,18 +4,71 @@
...
+import random
+from typing import TYPE_CHECKING, Any, Union
 
 import torch
+import transformers
 from torch.utils.data import DataLoader, Dataset
+from tqdm import tqdm
 
-from composer.core import DataSpec
+from composer.core import DataSpec, _default_split_batch
```
dakinggg: double import

composer/models/huggingface.py
```diff
@@ -302,6 +300,14 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
             # if we are in the single class case, then remove the classes dimension
             if output.shape[1] == 1:
                 output = output.squeeze(dim=1)
+        else:
+            output = outputs if outputs else self.forward(batch)
+            self.labels = batch.pop('labels')
```
dakinggg, line 305: I think we don't want to pop the labels in the else case here? to match the previous behavior

composer/models/huggingface.py
```diff
@@ -302,6 +300,14 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
             # if we are in the single class case, then remove the classes dimension
             if output.shape[1] == 1:
                 output = output.squeeze(dim=1)
+        else:
+            output = outputs if outputs else self.forward(batch)
+
+        if self.shift_labels or batch.get('mode', None) == 'lm_task':
```
dakinggg, line 306: lm_task -> icl_task?

composer/models/huggingface.py
```diff
@@ -286,13 +286,11 @@ def loss(self, outputs, batch):
...
-        output = outputs if outputs else self.forward(batch)
-        if self.use_logits or batch.get('mode', None) == 'lm_task':
+        if self.use_logits or batch.get('mode', None) == 'icl_task':
+            # pop labels first to avoid computing loss
             self.labels = batch.pop('labels')
-            if self.shift_labels:
-                # HF CausalLM models internally shift labels before computing loss, so we do the same here
-                self.labels[:, :-1] = self.labels[:, 1:].clone()
-                self.labels[:, -1] = -100
+            output = outputs if outputs else self.forward(batch)
```
dakinggg, line 292: I think this might actually cause a problem. If you pass the batch through without "labels", the logits are no longer at index one in the tuple. Still verifying, just want to block merging until we figure it out.

composer/models/huggingface.py
```diff
@@ -302,6 +300,14 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
             # if we are in the single class case, then remove the classes dimension
             if output.shape[1] == 1:
                 output = output.squeeze(dim=1)
+        else:
+            output = outputs if outputs else self.forward(batch)
+
+        if self.shift_labels or batch.get('mode', None) == 'icl_task':
```
dakinggg, line 306: can you just move this code back to where it was? you can still change the conditional to be this or statement with icl task

composer/models/huggingface.py
dakinggg: ok, so we need to change
```
            if self.config.use_return_dict:
                output = output['logits']
            else:
                # logits are at index 1 in the output tuple
                output = output[1]
```
to
```
            if self.config.use_return_dict:
                output = output['logits']
            else:
                # logits are at index 0 because loss is not present
                output = output[0]
```

composer/models/huggingface.py
```diff
@@ -286,22 +286,29 @@ def loss(self, outputs, batch):
...
-            if self.shift_labels:
-                # HF CausalLM models internally shift labels before computing loss, so we do the same here
-                self.labels[:, :-1] = self.labels[:, 1:].clone()
-                self.labels[:, -1] = -100
+            output = outputs if outputs else self.forward(batch)
+
             if self.config.use_return_dict:
                 output = output['logits']
             else:
                 # logits are at index 1 in the output tuple
```
dakinggg, line 297: Also need to update the comment to be true

composer/models/huggingface.py
dakinggg: I think you added this back in somehow?

dakinggg (approved):
LGTM once you resolve the last open conversations. Could you please also put in the PR description that numbers that you get from a manual test with a pretrained model + the full dataset?

bmosaicml (commented):
lgtm!

setup.py
```diff
@@ -169,10 +169,7 @@ def package_files(prefix: str, directory: str, extension: str):
     'pycocotools>=2.0.4,<3',
 ]
 
-extra_deps['nlp'] = [
-    'datasets>=2.4,<3',
-    'transformers>=4.11,<5',
-]
+extra_deps['nlp'] = ['transformers>=4.11,<5', 'datasets>=2.4,<3']
```
mvpatel2000, line 172:
```suggestion
extra_deps['nlp'] = [
    'transformers>=4.11,<5',
    'datasets>=2.4,<3',
]
```

mvpatel2000: nit: to be consistent with style elsewhere

composer/metrics/__init__.py
```diff
@@ -5,23 +5,17 @@
...
-    'LossMetric',
-    'Perplexity',
-    'BinaryF1Score',
-    'HFCrossEntropy',
-    'LanguageCrossEntropy',
-    'MaskedAccuracy',
-    'InContextLearningLMAccuracy',
+    'MAP', 'MIoU', 'Dice', 'CrossEntropy', 'LossMetric', 'Perplexity', 'BinaryF1Score', 'HFCrossEntropy',
+    'LanguageCrossEntropy', 'MaskedAccuracy', 'InContextLearningLMAccuracy', 'InContextLearningMultipleChoiceAccuracy',
+    'InContextLearningMetric'
```
mvpatel2000, line 15: nit: can we add a comma and go back to old format to be consistent please

mvpatel2000 (commented):
One more small nit: can we change from JSONL to JSON to clearly indicate file size limit (which applies bc we use Github)

A few small nits... i'll stamp once we figure out release process

setup.py
dakinggg: this was yapf

mvpatel2000: Probably because there was no trailing comma

composer/metrics/__init__.py
bmosaicml: what do you mean by "add a comma"?

mvpatel2000: add a trailing comma behind `'InContextLearningMetric'`

composer/metrics/nlp.py
```diff
@@ -263,18 +281,69 @@ class InContextLearningLMAccuracy(Metric):
...
+        return self.correct / self.total
+
+
+class InContextLearningMultipleChoiceAccuracy(InContextLearningMetric):
+    r"""Computes accuracy for In-context learning (ICL) multiple choice (MC) tasks.
+
+    ICL MC tasks consists of a series of questions with some number of possible choices (only one of which can be correct).
+    At inference time each possible choice is given to the model as a separate input and the one for which the model assigns
+    the lowest perplexity to the choice is considered the model's choice. The model is correct if it "chooses" the right answer.
+
```
knighton, line 307: should this be one blank line?

knighton (approved):

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -201,7 +202,7 @@ class InContextLearningMultipleChoiceTaskDataset(Dataset):
 
     If each question has N answer choices, we construct N distinct inputs per question. In order to ensure
     consistency across multi-GPU, we set the batch size to be `min(N, batch_size)` so that all N
-    inputs per question can stored in the same batch.
+    inputs per question can stored in the same batch.
```
dakinggg, line 205: this trailing whitespace is causing lint to fail

PR #1890, opened 2023-01-17
Ensure loggers run init event before callbacks in Engine

eracah:
# What does this PR do?

We want loggers to be initialized before callbacks. Currently, we enforce that by ordering loggers before callbacks when we store them in State. We also have a unit test that makes sure this is true.

This PR changes the way we enforce that by not enforcing any ordering of callbacks. Instead, we have `engine.run_event` do the following:
* process loggers first then callbacks when the event is `INIT`
* callbacks first then loggers in all other events. This is because some logging logic is done lazily (collects all logging calls and flushes them at once). In that case we want all the callbacks to execute all their logging calls before the loggers actually log everything.


mvpatel2000 (approved):
Can we add a test for this please?

eracah:
> Can we add a test for this please?

done

PR #1903, opened 2023-01-24
Runtime estimator

mvpatel2000:
# What does this PR do?

Extends speed monitor to provide an estimate of how much time a run will take to complete. To do this, we multiply out the throughput by remaining batches, and then adjust for evaluators throughput and frequency of eval calls. 

Resnet:
image
Note: Spikes are an artifact of shuffling from streaming as best I can tell

TODO:
- test on language model

[CO-1664](https://mosaicml.atlassian.net/browse/CO-1664)

[CO-1664]: https://mosaicml.atlassian.net/browse/CO-1664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvpatel2000:
Not sure why doctest is failing... error looks bizarre. Will debug later

composer/callbacks/speed_monitor.py
```diff
@@ -124,3 +179,15 @@ def batch_end(self, state: State, logger: Logger):
...
+        if state.dataloader_label not in self.eval_wct_per_label:
+            self.eval_wct_per_label[state.dataloader_label] = []
+        self.eval_wct_per_label[state.dataloader_label].append(state.eval_timestamp.total_wct.total_seconds())
+        max_dur = self.get_elapsed_duration(state)
+        if max_dur is None:
+            warnings.warn(
+                'Attempting to estimate remaining time but `max_duration` is not set. Skipping adjustment for evaluation time.'
+            )
+        else:
+            self.eval_rate_per_label[state.dataloader_label] = max_dur / len(
```
eracah, line 192: Wait I'm a bit confused as to how this is computed. So it's elapsed_duration, which is of units batch? and it's divided by `len(self.eval_wct_per_label[state.dataloader_label])`, which is the number of batches that had an eval? Any chance we could add another comment here or give `len(self.eval_wct_per_label[state.dataloader_label])` a descriptive variable name?

eracah (changes requested):
Thanks for jumping on this Mihir! I thiiiink it is correct, but it is a bit hard to read. Could you maybe add some comments and/or some descriptive variable names?

composer/callbacks/speed_monitor.py
```diff
@@ -41,19 +42,21 @@ class SpeedMonitor(Callback):
...
+    | ``throughput/samples_per_sec``    | batches) of the number of samples processed per second  |
+    |                                   |                                                         |
+    +-----------------------------------+---------------------------------------------------------+
+    | ``wall_clock/train``              | Total elapsed training time                             |
+    +-----------------------------------+---------------------------------------------------------+
+    | ``wall_clock/val``                | Total elapsed validation time                           |
+    +-----------------------------------+---------------------------------------------------------+
+    | ``wall_clock/total``              | Total elapsed time (wall_clock/train + wall_clock/val)  |
+    +-----------------------------------+---------------------------------------------------------+
+    | ``wall_clock/remaining_estimate`` | Estimated time to completion                            |
```
dakinggg, line 58: Given the hairiness of this estimate. Please spell out exactly how the estimate is computed.

composer/callbacks/speed_monitor.py
```diff
@@ -95,6 +102,31 @@ def load_state_dict(self, state: Dict[str, Any]) -> None:
         )
         self.total_eval_wct = state['total_eval_wct']
 
+        # Added in 0.13. Load only if present to support backwards compatibility
+        if 'eval_wct_per_label' in state:
```
dakinggg, line 106: can use `.get(<key>, <default>)` to compress lines of code a bit

composer/callbacks/speed_monitor.py
```diff
@@ -95,6 +102,31 @@ def load_state_dict(self, state: Dict[str, Any]) -> None:
...
+        if state.max_duration is None:
+            return None
+        elif state.max_duration.unit == TimeUnit('ep') and state.timestamp.epoch.value > 1:
+            batches_per_epoch = (state.timestamp.batch -
+                                 state.timestamp.batch_in_epoch).value / state.timestamp.epoch.value
+            return state.timestamp.get('ba').value / (state.max_duration.value * batches_per_epoch)
+        elapsed_dur = state.get_elapsed_duration()
+        if elapsed_dur:
+            return elapsed_dur.value
+        return None
```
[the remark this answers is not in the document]
mvpatel2000: The last `return None` is because I unpack `elapsed_dur` with `.value`, so I need to None check it. I think in the rest of the codebase we don't use `else:` case if it ends in a return, rather prefering:
```
if ...
  return ...
return ...
```
so I think it makes sense to keep it like this to be consistent? wdyt?

composer/callbacks/speed_monitor.py
```diff
@@ -108,11 +140,34 @@ def batch_end(self, state: State, logger: Logger):
...
 
+            # Estimate remaining time
+            batch_wct_avg = sum(self.batch_wct_buffer) / len(self.batch_wct_buffer)
+            elapsed_dur = self.get_elapsed_duration(state)
+            if elapsed_dur is None:
+                warnings.warn('`max_duration` is not set. Cannot estimate remaining time.')
+            elif elapsed_dur > 0 and elapsed_dur != self.last_elapsed_duration:
+                # Only update the estimate if the elapsed duration has changed
+                self.last_elapsed_duration = elapsed_dur
+                remaining_time = batch_wct_avg * int(state.timestamp.batch) / elapsed_dur * (1 - elapsed_dur)
```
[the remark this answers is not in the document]
mvpatel2000: Instead of a comment, i split out the middle term into a new var with a descriptive name (total_num_batches). Is it clear enough now?

composer/callbacks/speed_monitor.py
```diff
@@ -70,16 +73,20 @@ def __init__(self, window_size: int = 100):
...
         return {
             'batch_start_num_samples': self.batch_start_num_samples,
             'batch_start_wct': self.batch_start_wct,
             'batch_wct_buffer': self.batch_wct_buffer,
             'batch_num_samples_buffer': self.batch_num_samples_buffer,
-            # "window_wct": self.window_wct,
-            # "window_num_samples": self.window_num_samples,
             'total_eval_wct': self.total_eval_wct,
+            'eval_wct_per_label': self.eval_wct_per_label,
+            'eval_rate_per_label': self.eval_rate_per_label,
```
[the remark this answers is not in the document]
mvpatel2000: its the frequency of how often eval is called in duration units, e.g. if its called every 10 batches at 100 batch total train, it is 0.1. is there a better name you have in mind?

composer/callbacks/speed_monitor.py
mvpatel2000: So this is basically saying elapsed duration divided by number of calls. The goal is to compute frequency of eval calls in duration units.

composer/callbacks/speed_monitor.py
```diff
@@ -95,6 +102,31 @@ def load_state_dict(self, state: Dict[str, Any]) -> None:
...
+        provided at least 1 epoch has passed by recording how many batches were in each epoch.
+        """
+        if state.max_duration is None:
+            return None
+        elif state.max_duration.unit == TimeUnit('ep') and state.timestamp.epoch.value > 1:
+            batches_per_epoch = (state.timestamp.batch -
+                                 state.timestamp.batch_in_epoch).value / state.timestamp.epoch.value
+            return state.timestamp.get('ba').value / (state.max_duration.value * batches_per_epoch)
+        elapsed_dur = state.get_elapsed_duration()
+        if elapsed_dur:
```
dakinggg, line 126: you know what my comment is here :)

composer/callbacks/speed_monitor.py
```diff
@@ -108,11 +140,34 @@ def batch_end(self, state: State, logger: Logger):
...
+            if elapsed_dur is None:
+                warnings.warn('`max_duration` is not set. Cannot estimate remaining time.')
+            elif elapsed_dur > 0 and elapsed_dur != self.last_elapsed_duration:
+                # Only update the estimate if the elapsed duration has changed
+                self.last_elapsed_duration = elapsed_dur
+                remaining_time = batch_wct_avg * int(state.timestamp.batch) / elapsed_dur * (1 - elapsed_dur)
+                # Add remaining time from each evaluator
+                for dataloader_label, eval_wcts in self.eval_wct_per_label.items():
+                    # Discard first eval_wct if possible as it often slower due to dataset downloading
+                    eval_wct_avg = None
```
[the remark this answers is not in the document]
mvpatel2000: eh... I want to push back on this since shortcutting a len call seems a bit unnecessary and it adds another level of indirection when tracing code. open to discussion tho

composer/callbacks/speed_monitor.py
```diff
@@ -108,11 +140,36 @@ def batch_end(self, state: State, logger: Logger):
...
         if len(self.batch_num_samples_buffer) == self.window_size:
+            # Log the throughput
             throughput = sum(self.batch_num_samples_buffer) / sum(self.batch_wct_buffer)
             logger.log_metrics({'throughput/samples_per_sec': throughput})
 
+            # Estimate remaining time
+            batch_wct_avg = sum(self.batch_wct_buffer) / len(self.batch_wct_buffer)
+            elapsed_dur = self.get_elapsed_duration(state)
+            if elapsed_dur is None:
+                warnings.warn('`max_duration` is not set. Cannot estimate remaining time.')
```
dakinggg, line 152: if you don't set `max_duration`, I assume we just run one epoch of your dataset? And we don't know have any insight into how many samples are in your dataset? Is that correct?

composer/callbacks/speed_monitor.py
```diff
@@ -108,11 +140,36 @@ def batch_end(self, state: State, logger: Logger):
...
+            elif elapsed_dur > 0 and elapsed_dur != self.last_elapsed_duration:
+                # Only update the estimate if the elapsed duration has changed
+                self.last_elapsed_duration = elapsed_dur
+                total_num_batches = int(state.timestamp.batch) / elapsed_dur
+                remaining_time = batch_wct_avg * total_num_batches * (1 - elapsed_dur)
```
dakinggg, line 157: you can move the `batch_wct_avg` computation inside this `elif` I think?

composer/callbacks/speed_monitor.py
```diff
@@ -124,3 +181,15 @@ def batch_end(self, state: State, logger: Logger):
     def eval_end(self, state: State, logger: Logger):
         del logger  # unused
         self.total_eval_wct += state.eval_timestamp.total_wct.total_seconds()
+        assert state.dataloader_label is not None, 'evaluator label must not be None'
```
dakinggg, line 184: QQ: When can eval dataloader label be none? if it can only be none when you only have one evaluator, can you just use a default name in the none case?

dakinggg (commented):
broadly lgtm as a first pass at this estimate. i do request some tests though :)

composer/callbacks/speed_monitor.py
mvpatel2000: Correct. Its until dataloader is exhausted

composer/callbacks/speed_monitor.py
mvpatel2000: It can be `None` if user explicitly sets it to None when constructing it. This violates type hints and breaks things iirc, it really should always be set. However, `state.dataloader_label` can be `None` as it is possible to have no active dataloader, so we need this check for lint reasons (logically it should never happen)

dakinggg: got it, can you change the comment on the assert to say its for lint reasons?

composer/callbacks/speed_monitor.py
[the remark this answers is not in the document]
eracah: ah ok that's fine.  I guess then we can remove the elif then?

composer/callbacks/speed_monitor.py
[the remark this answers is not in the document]
eracah: `eval_frequency` perhaps? Also def add this comment to it: `"the frequency of how often eval is called in duration units, e.g. if its called every 10 batches at 100 batch total train, it is 0.1. "`

composer/callbacks/speed_monitor.py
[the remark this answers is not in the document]
eracah: If I was asking you to call this `len_eval_wcts = len(eval_wcts)`, then I think that would be unnecessary indirection, but in this case the variable is to make it more clear what this term in the equation means i.e. if you were to explain out exactly why the formula is as it is, would you refer to this as the length of the `eval_wcts` list or as the number of times eval has happened? A comment could work too if you don't want to do a variable name

composer/callbacks/speed_monitor.py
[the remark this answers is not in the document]
eracah: ohhhhh wow `elapsed_dur` is a fraction between 0 and 1? Oh wow I didn't realize that. That makes things much clearer. It def is clear enough now IF you rename `elapsed_dur` to `elapsed_fraction` or `elapsed_percentage` or something.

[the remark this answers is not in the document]
eracah: Sorry last nit!

composer/callbacks/speed_monitor.py
eracah: ok yeah this makes sense now that I know elapsed_duration is a fraction

composer/callbacks/speed_monitor.py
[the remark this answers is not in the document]
mvpatel2000: changed my mind, will go with ur original suggestion

dakinggg (approved):
LGTM

composer/callbacks/speed_monitor.py
```diff
@@ -108,11 +140,38 @@ def batch_end(self, state: State, logger: Logger):
...
 
-        # Log the throughput
         if len(self.batch_num_samples_buffer) == self.window_size:
+            # Log the throughput
             throughput = sum(self.batch_num_samples_buffer) / sum(self.batch_wct_buffer)
             logger.log_metrics({'throughput/samples_per_sec': throughput})
 
+            # Estimate remaining time
+            elapsed_fraction = self.get_elapsed_duration(state)
+            print(throughput, elapsed_fraction)
```
hanlint, line 150: remove print

hanlint (commented):
There's a lot of fancy logic here -- is this needed?

Seems like a simpler approach is a separate ETACallback that does:
```
class ETACallback:
    def init():
        self.start_time = time.time()
    
    def batch_end():
        eta = (time.time() - self.start_time) / self.get_elapsed_duration
```

The eval calls timing would be baked into the ETA -- there would be some inaccuracy early on as the time spent on training vs. eval would be off, but I'm trying to decide if for typical use cases that inaccuracy is worth the complexity of this callback + needing to integrate this with SpeedMonitor vs. being a simple separate add-on.


```

mvpatel2000:
> There's a lot of fancy logic here -- is this needed?

Here's the kind of whipsawing I was seeing without the eval correction which led me to this (ignore bottom green line). In either case, I'm going to move this to a design doc before merging so I think it's something we can definitely discuss

image

mvpatel2000:
Closing, we merged elsewhere

PR #1906, opened 2023-01-24
Add primitive list support

mvpatel2000:
# What does this PR do?

Add support to dataspec to split batches consisting of primitive types, e.g. a list of strings as prompts.

[CO-1686](https://mosaicml.atlassian.net/browse/CO-1686)

[CO-1690](https://mosaicml.atlassian.net/browse/CO-1690)

[CO-1686]: https://mosaicml.atlassian.net/browse/CO-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CO-1690]: https://mosaicml.atlassian.net/browse/CO-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -51,6 +51,7 @@ def _num_microbatches_split_mapping(m, num_microbatches: int):
         elif isinstance(v, (List, Tuple)):
             chunked[k] = _num_microbatches_split_list(v, num_microbatches)
         elif isinstance(v, (int, float, str, bool)):
+            # Broadcast primitives to all chunks
             chunked[k] = [v] * num_microbatches
         else:
             raise ValueError(f'Unsupported batch type: {type(v)}.')
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -48,8 +48,12 @@ def _num_microbatches_split_mapping(m, num_microbatches: int):
     for k, v in m.items():
         if isinstance(v, torch.Tensor):
             chunked[k] = _num_microbatches_split_tensor(v, num_microbatches)
-        if isinstance(v, (List, Tuple)):
+        elif isinstance(v, (List, Tuple)):
             chunked[k] = _num_microbatches_split_list(v, num_microbatches)
+        elif isinstance(v, (int, float, str, bool)):
+            chunked[k] = [v] * num_microbatches
+        else:
+            raise ValueError(f'Unsupported batch type: {type(v)}.')
     num_chunks = len(list(chunked.values())[0])
     return [{k: v[idx] for k, v in chunked.items()} for idx in range(num_chunks)]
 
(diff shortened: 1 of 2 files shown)


dakinggg:
while you're here, can you also add a test (and support if necessary) for having a single string in the batch?
e.g.
```
{'input_ids': <tensor with batch dimension>, 'labels': <tensor with batch dimension>, 'metadata': <single string>}
```
was on my list to check if that errors or not/add support

tests/test_split_batch.py
```diff
@@ -76,6 +80,7 @@ def generate_maskrcnn_sample(num_detections,
 def dummy_batches(batch_size=12):
```
dakinggg, line 80: If i'm reading the tests right, this only tests that it doesn't error. can you also add a test for getting the right shape out?

dakinggg (approved):
LGTM

dakinggg (approved):
LGTM

composer/core/data_spec.py
```diff
@@ -48,8 +48,12 @@ def _num_microbatches_split_mapping(m, num_microbatches: int):
     for k, v in m.items():
         if isinstance(v, torch.Tensor):
             chunked[k] = _num_microbatches_split_tensor(v, num_microbatches)
-        if isinstance(v, (List, Tuple)):
+        elif isinstance(v, (List, Tuple)):
             chunked[k] = _num_microbatches_split_list(v, num_microbatches)
+        elif isinstance(v, (int, float, str, bool)):
+            chunked[k] = [v] * num_microbatches
```
eracah, line 54: so we replicate the int/float/str/bool `num_microbatches` times? Why is that?

composer/core/data_spec.py
```diff
@@ -74,6 +78,9 @@ def _num_microbatches_split_batch(batch: Any, num_microbatches: int) -> Sequence
     if isinstance(batch, Mapping):  # check for dictionary (hf style)
         return _num_microbatches_split_mapping(batch, num_microbatches)
 
+    if isinstance(batch, (Tuple, list)) and _check_list_is_primitives(batch):  # check for list of primitives
+        return _num_microbatches_split_list(batch, num_microbatches)
+
```
eracah, line 83: can we just put 
```
if _check_list_is_primitives(batch): 
    return _num_microbatches_split_list(batch, num_microbatches)
```
underneath line 84?

composer/core/data_spec.py
```diff
@@ -117,12 +124,36 @@ def _split_mapping(m, microbatch_size: int):
...
             chunked[k] = _split_list(v, microbatch_size)
-    num_chunks = len(list(chunked.values())[0])
+        elif isinstance(v, (int, float, str, bool)):
+            # Defer broadcasting primitives until we know num_chunks
+            pass
+        else:
+            raise ValueError(f'Unsupported batch type: {type(v)}.')
+    num_chunks = 1  # Default to 1 chunks if there are no tensors or everything is primitive
+    if len(chunked.keys()) != 0:
+        num_chunks = len(list(chunked.values())[0])
```
eracah, line 136:
```suggestion
    if len(chunked.keys()) != 0:
        num_chunks = len(list(chunked.values())[0])
    else:
        num_chunks = 1  # Default to 1 chunks if there are no tensors or everything is primitive
```

eracah (commented):
Looks good. Just had a couple nits 😉

composer/core/data_spec.py
mvpatel2000: added a comment! Requested feature from Daniel

eracah:
> Looks good. Just had a couple nits 😉

I will defiNITely approve the PR once those are fixed

composer/core/data_spec.py
mvpatel2000: prefer as is, explicitly indicates scope of `num_chunks` as outside if/else

eracah: but the readability... ok fine

PR #1913, opened 2023-01-26
Add logic for shifting labels before computing metrics

alextrott16:
# What does this PR do?

Adds a `shift_labels` init argument to `HuggingFaceModel` class. This instructs the model whether to shift labels by one token before computing metrics, which mimics the way HF Causal LM classes handle labels when computing loss. This fixes the current implementation, which never does this shifting and produces incorrect metric results for Causal LMs.

If `shift_labels` is not specified, `HuggingFaceModel` will try to infer the correct behavior based on whether the model is an instance of a registered HF Causal LM class (or a subclass of one).

https://mosaicml.atlassian.net/browse/CO-1691

# Before submitting

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -291,7 +291,7 @@ class HuggingFaceModel(ComposerModel):
             self.labels = batch.pop('labels')
             if self.shift_labels:
                 # HF CausalLM models internally shift labels before computing loss, so we do the same here
-                self.labels[:, :-1] = self.labels[:, 1:]
+                self.labels[:, :-1] = self.labels[:, 1:].clone()
                 self.labels[:, -1] = -100
             if self.config.use_return_dict:
                 output = output['logits']
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -30,17 +30,6 @@ log = logging.getLogger(__name__)
 __all__ = ['HuggingFaceModel']
 
 
-def _is_registered_causal_lm(model: transformers.PreTrainedModel) -> bool:
-    """Return True if model class is either a registered 🤗 Causal LM or a subclass of one"""
-    try:
-        from transformers.models.auto.modeling_auto import MODEL_FOR_CAUSAL_LM_MAPPING
-    except ImportError as e:
-        raise MissingConditionalImportError(extra_deps_group='nlp',
-                                            conda_package='transformers',
-                                            conda_channel='conda-forge') from e
-    causal_lm_classes = list(MODEL_FOR_CAUSAL_LM_MAPPING.values())
-    return any([isinstance(model, causal_lm_class) for causal_lm_class in causal_lm_classes])
-
 class HuggingFaceModel(ComposerModel):
     """
     A wrapper class that converts 🤗 Transformers models to composer models.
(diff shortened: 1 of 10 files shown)


composer/models/huggingface.py
```diff
@@ -42,6 +42,10 @@ class HuggingFaceModel(ComposerModel):
                 using :meth:`HuggingFaceModel.hf_from_composer_checkpoint`. If the tokenizer is not provided here, it will not be saved in the composer checkpoint.
         use_logits (bool, optional): If True, the model's output logits will be used to calculate validation metrics. Else, metrics will be inferred from the HuggingFaceModel directly. Default: ``False``
         metrics (list[Metric], optional): list of torchmetrics to apply to the output of `validate`. Default: ``None``.
+        shift_labels (bool, optional): If True, the batch's labels will be shifted before being used to calculate metrics. This should be set to true for CausalLM models and false otherwise. If not specified, `shift_labels` will be set automatically based on the model class name. Default: ``None``.
+
+            .. note:: To ensure correct behavior, set `shift_labels` manually if using a custom model (i.e., if `model` is not a
```
dakinggg, line 47:
```suggestion
            .. note:: To ensure correct behavior, set `shift_labels` manually if using a custom model (i.e., if `model` is not
```

dakinggg (commented):
Can you add a test please? Otherwise LGTM. Will approve after test

tests/models/test_hf_model.py
```diff
@@ -457,3 +457,49 @@ def test_hf_loading_errors(tiny_bert_model, tiny_bert_tokenizer, model_class_nam
     with error_contexts[model_class_name]:
         _, _ = HuggingFaceModel.hf_from_composer_checkpoint(str(tmp_path / 'hf-checkpoint.pt'),
                                                             model_class_name_to_class[model_class_name])
+
+
+@pytest.mark.parametrize('model_type', ['gpt', 'bert'])
```
dakinggg, line 462: Ah, sorry i told you to use the wrong thing for fixtures. unfortunately pytest does not allow you to put fixtures in the parametrization as you may have discovered, so instead you can use `configure_tiny_bert_model` and `configure_tiny_gpt_model` (again pattern match to create the gpt one), found in https://github.com/mosaicml/composer/blob/08f43a489b5d6cd04506c91bb8cbc52b06ebb296/tests/common/models.py#L376. example of doing this here: https://github.com/mosaicml/composer/blob/7d400151dcf14b497741ec427bfbc2345439b598/tests/algorithms/test_gated_linear_units.py#L70-L78

dakinggg (approved):
LGTM, left one comment to cleanup the fixture usage a bit (sorry, i pointed you to the wrong place for the usage you wanted)

tests/common/models.py
```diff
@@ -396,3 +396,15 @@ def configure_tiny_bert_config():
 
 def configure_tiny_bert_hf_model(use_logits=True):
     return HuggingFaceModel(configure_tiny_bert_model(), configure_tiny_bert_tokenizer(), use_logits)
+
+
+def configure_tiny_gpt_model():
+    return copy.deepcopy(pytest.tiny_gpt_model)
```
dakinggg, line 402: oops sorry, can you add the `try/excepts` that the bert ones have? I missed that in my PR

alextrott16: Yep, added.

PR #1915, opened 2023-01-27
Add coreweave object store support.

eracah:
# What does this PR do?

Coreweave object store is compatible with boto3. Uploading objects to Coreweave object store is almost exactly like writing to S3 using S3Object Store, except an endpoint_url must be set: `https://object.<region>.coreweave.com`.

Unfortunately, the endpoint_url can only be set programmatically (AWS doesn't support an endpoint_url env var or config field), so we create our own env var called `AWS_ENDPOINT_URL`, that we check for in the `S3ObjectStore`. The Trainer interface for the user is exactly the same as normal S3, except they have to set this env var.

Tested manually for save and load checkpoint

fix CO-1471

### DIFF
--- a/composer/utils/object_store/s3_object_store.py
+++ b/composer/utils/object_store/s3_object_store.py
@@ -93,6 +93,9 @@ class S3ObjectStore(ObjectStore):
         if client_config is None:
             client_config = {}
         config = Config(**client_config)
+        if 'AWS_ENDPOINT_URL' in os.environ and endpoint_url is None:
+            endpoint_url = os.environ['AWS_ENDPOINT_URL']
+
         self.client = boto3.Session().client(
             's3',
             config=config,
--- a/composer/utils/object_store/s3_object_store.py
+++ b/composer/utils/object_store/s3_object_store.py
@@ -93,8 +93,8 @@ class S3ObjectStore(ObjectStore):
         if client_config is None:
             client_config = {}
         config = Config(**client_config)
-        if 'AWS_ENDPOINT_URL' in os.environ and endpoint_url is None:
-            endpoint_url = os.environ['AWS_ENDPOINT_URL']
+        if 'S3_ENDPOINT_URL' in os.environ and endpoint_url is None:
+            endpoint_url = os.environ['S3_ENDPOINT_URL']
 
         self.client = boto3.Session().client(
             's3',
--- a/docs/source/trainer/checkpointing.rst
+++ b/docs/source/trainer/checkpointing.rst
@@ -331,7 +331,7 @@ and delete the checkpoints from the local disk. The checkpoints will be located 
 
 For uploading checkpoints to [Coreweave's object store](https://docs.coreweave.com/storage/object-storage), the code is very similar to the
 above S3 uploading code. The only difference is you must set your Coreweave endpoint url.
-To do this you can just set the ``AWS_ENDPOINT_URL`` environment variable before creating the
+To do this you can just set the ``S3_ENDPOINT_URL`` environment variable before creating the
 :class:`.Trainer`, like so:
 
 .. testcode::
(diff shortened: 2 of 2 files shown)


composer/utils/object_store/s3_object_store.py
```diff
@@ -93,6 +93,9 @@ def __init__(
         if client_config is None:
             client_config = {}
         config = Config(**client_config)
+        if 'AWS_ENDPOINT_URL' in os.environ and endpoint_url is None:
```
dakinggg, line 96: is `AWS_ENDPOINT_URL` the right name? Or is `S3_ENDPOINT_URL` more accurate?

dakinggg (approved):
LGTM assuming you did a manual test

docs/source/trainer/checkpointing.rst
```diff
@@ -329,6 +329,34 @@ and delete the checkpoints from the local disk. The checkpoints will be located
 ``checkpoints/ep3.pt`` for third epoch's checkpoints, for example. The full URI in this case would be:
 ``s3://my_bucket/checkpoints/ep3.pt``.
 
+For uploading checkpoints to Coreweave's object store, the code is very similar to the
```
hanlint, line 332: can you link to some docs about Coreweave's object store?

docs/source/trainer/checkpointing.rst
```diff
@@ -329,6 +329,34 @@ and delete the checkpoints from the local disk. The checkpoints will be located
...
+
+    os.environ['AWS_ENDPOINT_URL'] = 'https://object.las1.coreweave.com'
+    from composer.trainer import Trainer
+
+    # Save checkpoints every epoch to s3://my_bucket/checkpoints
+    trainer = Trainer(
+        model=model,
+        train_dataloader=train_dataloader,
+        max_duration='10ep',
+        save_folder='s3://my_bucket/checkpoints',
```
hanlint, line 350: Would it make more sense for the user to set a URI for coreweave (not sure what they use, we can ask them) That we use boto3/AWS to access coreweave seems like an implementation detail that the user should not need to be troubled with?

mvpatel2000:
Is there a way to test this in a unit test? Do we want to start adding test coverage for things like this (non-S3 integration tests)? 

Not blocking, fine w merging, just food for thought

docs/source/trainer/checkpointing.rst
eracah: yeah that's the weird thing is that they still use an s3 uri. See [here](https://docs.coreweave.com/storage/object-storage)

composer/utils/object_store/s3_object_store.py
eracah: oh good call. updated!

docs/source/trainer/checkpointing.rst
eracah: done

eracah:
> LGTM assuming you did a manual test

yeah did manual test of saving and loading checkpoints

eracah:
> Is there a way to test this in a unit test? Do we want to start adding test coverage for things like this (non-S3 integration tests)?
> 
> Not blocking, fine w merging, just food for thought

Hmmm. Don't really think so because it is exactly like an S3 unit test, but the user sets the env variable, so not sure how we would make a new unit test

composer/utils/object_store/s3_object_store.py
eracah: also fixed in the mcli PR

docs/source/trainer/checkpointing.rst
```diff
@@ -329,6 +329,34 @@ and delete the checkpoints from the local disk. The checkpoints will be located
 ``checkpoints/ep3.pt`` for third epoch's checkpoints, for example. The full URI in this case would be:
 ``s3://my_bucket/checkpoints/ep3.pt``.
 
+For uploading checkpoints to [Coreweave's object store](https://docs.coreweave.com/storage/object-storage), the code is very similar to the
+above S3 uploading code. The only difference is you must set your Coreweave endpoint url.
+To do this you can just set the ``AWS_ENDPOINT_URL`` environment variable before creating the
```
dakinggg, line 334: env var needs to be changed here too

eracah: oh right good catch

PR #1919, opened 2023-01-28
Fixes mis specified dependency

dakinggg:
# What does this PR do?
Fixes a mis specified dependency that results in installation from source failing with `'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers`. Is likely related to the latest release of setuptools (https://setuptools.pypa.io/en/latest/history.html#v67-0-0), but I'm not sure what the dependency path is that leads us there, as this fails without changing anything about the environment it is running in (i.e. did not upgrade any packages, and setuptools is pinned to `0.59.5` in the environment i'm running in).

isort also broke so requires an upgrade: https://github.com/PyCQA/isort/pull/2078

This also requires moving the pip install out of the mcli integration, so that it can pick up changes to setup.py before installing.

Closes [CO-1701](https://mosaicml.atlassian.net/browse/CO-1701)

# Before submitting

[CO-1701]: https://mosaicml.atlassian.net/browse/CO-1701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/mcp/mcp_pytest.py
+++ b/.github/mcp/mcp_pytest.py
@@ -28,7 +28,6 @@ if __name__ == '__main__':
         'integration_type': 'git_repo',
         'git_repo': 'mosaicml/composer',
         'ssh_clone': 'False',
-        'pip_install': '--user -e .[all]',
     }
     if args.git_branch is not None and args.git_commit is None:
         git_integration['git_branch'] = args.git_branch
@@ -44,6 +43,8 @@ if __name__ == '__main__':
         git fetch origin pull/{args.pr_number}/head:pr_branch
 
         git checkout pr_branch
+
+        pip install --user -e .[all]
 
         '''
 
--- a/.github/mcp/mcp_pytest.py
+++ b/.github/mcp/mcp_pytest.py
@@ -44,11 +44,11 @@ if __name__ == '__main__':
 
         git checkout pr_branch
 
-        pip install --user .[all]
-
         '''
 
     command += f'''
+
+    pip install --user .[all]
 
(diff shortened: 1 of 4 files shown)


mvpatel2000 (approved):

.github/mcp/mcp_pytest.py
```diff
@@ -45,6 +44,8 @@
 
         git checkout pr_branch
 
+        pip install --user -e .[all]
+
```
mvpatel2000, line 48:
```suggestion
        pip install --user .[all]
```
doesn't need to be editable

dakinggg: done

.github/mcp/mcp_pytest.py
```diff
@@ -45,6 +44,8 @@
 
         git checkout pr_branch
 
+        pip install --user .[all]
+
         '''
 
     command += f'''
```
mvpatel2000, line 51:
```suggestion
        '''

    command += f'''
    pip install --user .[all]

```

mvpatel2000 (approved):
LGTM. Will wait for non-GPU tests to pass before merging

mvpatel2000:
Red button merging as GPU workflow will continue to fail. 

This PR updates workflow to pip install after checking out branch, which means it will correctly use the right setup.py in future. However, `pull_request_target` workflows are not updated in PRs from fork, so it will not be able to be patched to pass tests here. GPU workflow is new so some problems like this are expected

PR #1926, opened 2023-01-30
pin setuptools in build requirements

dakinggg:
# What does this PR do?
We ran into an issue where a setuptools release broke pip install from source. This is because pip uses an isolated build environment, which would always install the latest setuptools. In this PR we pin the setuptools version for building the package so that we have control over when to upgrade. Tested `pip install -e '.[all]'` locally.

Closes [CO-1702](https://mosaicml.atlassian.net/browse/CO-1702)

# Before submitting

[CO-1702]: https://mosaicml.atlassian.net/browse/CO-1702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+# build requirements
+[build-system]
+requires = ["setuptools <= 68.0.0"]
+build-backend = "setuptools.build_meta"
 
 # iSort
 [tool.isort]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 # build requirements
 [build-system]
-requires = ["setuptools <= 68.0.0"]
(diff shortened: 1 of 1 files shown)


bandish-shah (approved):
LGTM

mvpatel2000:
Do you know if dependabot would pick this up for upgrades like it does with `setup.py`?

dakinggg:
My guess is no. But we probably never need to upgrade it, unless there is some huge new feature. The build system works as is right now, and I think we're ok to only upgrade when strictly necessary for some reason.

PR #1930, opened 2023-01-31
Temporarily pin pip to <23

dakinggg:
# What does this PR do?
This is a temporary workaround to unblock CI while we figure out the root cause. Currently installing from source does not work with pip==23 outside of a virtual env.


bandish-shah (approved):
LGTM

PR #1932, opened 2023-01-31
Fix mcp script to avoid follow

mvpatel2000:
# What does this PR do?

`follow_run_logs` occasionally disconnects. Instead, we need to wait for the run to complete and then get logs. Additionally, we want to add a timeout to kill daily tests if they don't finish in time.

Along for ride, adds `datasets` to nlp install which breaks install rn

[CO-1722](https://mosaicml.atlassian.net/browse/CO-1722)

[CO-1722]: https://mosaicml.atlassian.net/browse/CO-1722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -391,23 +391,6 @@ stage('Build') {
                 }
             }]
         }
-        else if (!isComposerImage) {
-            // If not rebuilding the docker image, and it's not a composer image, run GPU tests only on the "latest" image
-            // Only run tests on the 'mosaicml/pytorch' or 'mosaicml/pytorch_vision' images -- not 'mosaicml/composer'
-            Boolean isCpuImage = buildArgs['CUDA_VERSION'] == ''
-            Boolean shouldRunGpuTests = false
-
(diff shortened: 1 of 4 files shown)


.ci/Jenkinsfile
```diff
@@ -391,23 +391,6 @@ stage('Build') {
                 }
             }]
         }
-        else if (!isComposerImage) {
```
bandish-shah, line 394: Would it make sense to do this in it's own PR?  Just thinking if anything goes wrong it would be easier to revert a specific commit.

nik-mosaic:
Why does adding datasets break install?

mvpatel2000:
> Why does adding datasets break install?

Not having datasets breaks install bc it is used in the lm eval code but not registered as an nlp dependency

nik-mosaic (approved):

bandish-shah (approved):
LGTM

PR #1941, opened 2023-02-03
Refactor EMA to improve memory efficiency

coryMosaicML:
# What does this PR do?

This PR refactors composer's EMA algorithm. Changes:

- Avoids an extra copy of the model parameters, reducing extra device memory from 2x -> 1x the size of the model parameters
- Avoids duplicating non-trainable parameters which reduces memory required for models with many non trainable parameters.
- Avoids a deepcopy of `state.model` after it has been DDP-wrapped

A downside: it is now a bit more annoying to access the training weights or ema weights directly from the algorithm object. 
 
[CO-1525](https://mosaicml.atlassian.net/browse/CO-1525)

[CO-1525]: https://mosaicml.atlassian.net/browse/CO-1525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvpatel2000 (commented):
@coryMosaicML lgtm. Before approval though, would like to see plots / some kind of actual run verifying it works and gives same results as before. Can you please run something like that?

mvpatel2000 (approved):
LGTM. Once lint is fixed, feel free to merge

PR #1950, opened 2023-02-07
add support for enc-dec batches without decoder_input_ids

dakinggg:
# What does this PR do?
When using and encoder decoder model like T5, the forward pass either needs to be passed `decoder_input_ids` or `labels`. If it is passed `labels`, it will internally create `decoder_input_ids` using the `prepare_decoder_input_ids_from_labels` function. Since we pop the `labels` in `eval_forward` to avoid HF internally recomputing the loss, we need to create `decoder_input_ids` in the same way inside of our `eval_forward`

Closes [CO-1745](https://mosaicml.atlassian.net/browse/CO-1745)

# Before submitting

[CO-1745]: https://mosaicml.atlassian.net/browse/CO-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvpatel2000 (commented):
Defer to @alextrott16, will stamp after he's done

composer/models/huggingface.py
```diff
@@ -290,6 +290,16 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
             # pop labels first to avoid computing loss
             self.labels = batch.pop('labels')
 
+            # HF encoder decoder models like T5 expect either decoder_input_ids or labels,
+            # so we add decoder_input_ids to the batch if it is missing
+            if self.model.config.is_encoder_decoder and 'decoder_input_ids' not in batch:
+                if hasattr(self.model, 'prepare_decoder_input_ids_from_labels'):
+                    batch['decoder_input_ids'] = self.model.prepare_decoder_input_ids_from_labels(labels=self.labels)
+                else:
+                    raise RuntimeError(
```
alextrott16, line 299: I'm on the fence about whether to raise an error here. The reason is that HF models tend to have pretty good checks on inputs and we could leave it up to the model to tell us if a batch without `labels` and without `decoder_input_ids` is a problem. I checked a handful of HF EncDec models and they all have `prepare_decoder_input_ids_from_labels`, so in the unlikely event that you make it to Line 299 here, maybe you actually have an OK batch and the error is unnecessary.

Of course, if that line of thinking is wrong, then debugging will be made harder by removing this error. So, I leave this comment here just as something to consider. If you prefer to keep the error, that's fine with me.

alextrott16 (approved):
LGTM. Thanks for taking care of this so quickly!

composer/models/huggingface.py
dakinggg: good points. I think I will leave the error because I'd rather error in a case that we haven't considered rather than (maybe) silently fail...

mvpatel2000 (approved):

PR #1951, opened 2023-02-07
Update MCP GHA Name

mvpatel2000:
# What does this PR do?

Rename mcp tests in mcloud to include more descriptive names of whats running

[CO-1737](https://mosaicml.atlassian.net/browse/CO-1737)

[CO-1737]: https://mosaicml.atlassian.net/browse/CO-1737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/mcp/mcp_pytest.py
+++ b/.github/mcp/mcp_pytest.py
@@ -52,6 +52,10 @@ if __name__ == '__main__':
 
         '''
 
+    # Shorten name if too long
+    if len(name) > 56:
+        name = name[:56]
+
     command += f'''
 
     pip install --upgrade --user .[all]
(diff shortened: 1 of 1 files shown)


.github/mcp/mcp_pytest.py
```diff
@@ -26,20 +26,24 @@
...
     args = parser.parse_args()
 
+    name = args.name
     git_integration = {
         'integration_type': 'git_repo',
         'git_repo': 'mosaicml/composer',
         'ssh_clone': 'False',
     }
     if args.git_branch is not None and args.git_commit is None:
+        name += f'-branch-{args.git_branch}'
```
bandish-shah, line 36: Do we need to do any re-formatting here so that the branch string is formatted correctly?  For example I don't think a branch with '/' in the name will work in the run name.

mvpatel2000: MCLOUD should reformat for us:
```
Invalid run name "sd/baseline-112": Run names must be less than 63 characters and contain only lower-case letters, numbers, or "-". Converting to a valid name: sd-baseline-112
```
I've added a cap for max length

PR #1953, opened 2023-02-07
Move ffcv up in test order

dskhudia:
# What does this PR do?

~~Randomize the order of pytest to catch errors that may work if the tests are run in the same order.~~

Move the test up in order so that it passes. Don't ask me why!!

### DIFF
--- a/tests/trainer/test_trainer.py
+++ b/tests/trainer/test_trainer.py
@@ -893,6 +893,67 @@ class TestTrainerInitOrFit:
                 assert event_counter_callback.event_to_num_calls[Event.EPOCH_CHECKPOINT] == 2
 
 
+@pytest.mark.vision
+class TestFFCVDataloaders:
+
+    train_file = None
+    val_file = None
+    tmp_path = None
+
+    @pytest.fixture(autouse=True)
+    def create_dataset(self, tmp_path_factory: pytest.TempPathFactory):
+        dataset_train = RandomImageDataset(size=16, is_PIL=True)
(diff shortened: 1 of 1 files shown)


mvpatel2000:
Can we fix the ones failing rn that are deterministically failing first?

dskhudia:
ffcv test is not deterministically failing for me. It passes on local run.

dakinggg:
So, I agree with the sentiment of this change, but I'm a little worried that we may have a number of non deterministic issues and will spend a significant amount of time debugging them on unrelated PRs. Is there any way we can de risk that? Like, should someone spend a few days just trying different random orders and fixing anything that pops up?

dskhudia:
@dakinggg we don't have to merge this now but eventually do something like this to make testing more robust. Currently I want to see if the test_ffcv failure goes away with this.

dakinggg:
Sounds good :+1:

dskhudia:
>  non deterministic issues and will spend a significant amount of time debugging them on unrelated PRs.

BTW this is more one time effort than ongoing effort.

dakinggg:
> > non deterministic issues and will spend a significant amount of time debugging them on unrelated PRs.
> 
> BTW this is more one time effort than ongoing effort.

Absolutely :)

dskhudia:
Ignore any updates to this PR. This is just to fix the failing test.

bandish-shah (approved):
Super bizarre but LGTM.

PR #1955, opened 2023-02-08
Change how we check for forwards args in models for HF models

bcui19:
# What does this PR do?
We cache the model forward args before we wrap the model with FSDP. Before we'd be checking a FSDP model, which would result in an empty batch.

ran with FSDP locally:
https://wandb.ai/mosaic-ml/examples_fork-examples_ul2/runs/3044levt/overview?workspace=
ran with DDP locally:
https://wandb.ai/mosaic-ml/examples_fork-examples_ul2/runs/2eo4hqfu/overview?workspace=user-bcui

[CO-1756](https://mosaicml.atlassian.net/browse/CO-1756)

# Before submitting

[CO-1756]: https://mosaicml.atlassian.net/browse/CO-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bcui19:
There could be a pitfall for a user if they pass in a wrapped FSDP model into the HF wrapper, which is unlikely, but it is possible.

The other way is to check for `inspect.getfullargspec(self.model.forward).args` if it's not a wrapped FSDP model and `inspect.getfullargspec((self.model._fsdp_wrapped_module._fpw_module.forward).args`... Lmk if this would be more desired.

alextrott16:
Brilliantly simple fix. I love it.

> The other way is to check for inspect.getfullargspec(self.model.forward).args if it's not a wrapped FSDP model and inspect.getfullargspec((self.model._fsdp_wrapped_module._fpw_module.forward).args... Lmk if this would be more desired.

@bcui19 I would advocate for doing this in `__init__()` much in the same way you have it set up. If the user wraps (or unwraps?) the model after the `HuggingFaceModel` instance is initialized, your solution should still work, as long as `.model_forward_args` is correctly inferred at initialization.

So, I'm in favor of being able to get `.model_forward_args` correctly if (a) the model hasn't been wrapped at all, (b) the model has already been FSDP wrapped, or (c) the model has already been DDP wrapped.

bcui19:
I agree if we were to support this it should be in `__init__()`... I'm not certain if we should support in passing in a pre-FSDP wrapped model into `HuggingFaceModel,` since I think all native functions from the HF model e.g. `resize_token_embeddings` would be uncallable (it'd try to call the wrapped module)? So for now it might just be best to leave it as is and if you want to wrap the model, wrap the model after `HuggingFaceModel.__init__()`?

alextrott16 (approved):
LGTM

dakinggg (approved):
LGTM, thanks for fixing this. I'd probably prefer we avoid the whole inspect thing entirely, but that can be for a later day. Could you add a simple test that exercises `HuggingFaceModel` + FSDP? (or open a JIRA and assign it to me and I can do it at some point)

bcui19:
re inspect: I agree, for now will merge. Added CO-1768

PR #1965, opened 2023-02-14
Add support for FP8 on H100 using NVidia's TransformerEngine

dskhudia:
# What does this PR do?

Adds amp_fp8 precision. This will allows us to train model faster using FP8 precision on H100 systems. It's a no op if amp_fp8 precision is not used. 

FP8 training support on H100

### DIFF
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -5,11 +5,18 @@
 
 import contextlib
 import os
+import textwrap
 from typing import Generator, Union
 
 import torch
 
 from composer.utils import StringEnum
+
+try:
+    import transformer_engine.pytorch as te
+    te_installed = True
+except ImportError:
+    te_installed = False
 
 __all__ = ['Precision', 'get_precision_context']
 
@@ -19,13 +26,15 @@ class Precision(StringEnum):
 
     Attributes:
         FP32: Use 32-bit floating-point precision. Compatible with CPUs and GPUs.
-        AMP_FP16: Use :mod:`torch.cuda.amp` wih 16-bit floating-point precision. Only compatible
+        AMP_FP16: Use :mod:`torch.cuda.amp` with 16-bit floating-point precision. Only compatible
             with GPUs.
-        AMP_BF16: Use :mod:`torch.cuda.amp` wih 16-bit BFloat precision.
+        AMP_BF16: Use :mod:`torch.cuda.amp` with 16-bit BFloat precision.
+        AMP_FP8: Use :mod:`transformer_engine.pytorch.fp8_autocast` with 8-bit FP8 precison.
     """
     FP32 = 'fp32'
     AMP_FP16 = 'amp_fp16'
     AMP_BF16 = 'amp_bf16'
+    AMP_FP8 = 'amp_fp8'
 
 
 @contextlib.contextmanager
@@ -54,5 +63,25 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
         else:
             os.environ['XLA_USE_BF16'] = '1'
             yield
+    elif precision == Precision.AMP_FP8:
+        if te_installed and torch.cuda.get_device_capability()[0] > 8:
+            from transformer_engine.common.recipe import DelayedScaling, Format
+
+            # These default values for fp8_recipe are taken from NVidia's docs. We may want to change
+            # these once we get a chance to do more convergence experiments.
+            # https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#id1
+            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
+            fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo='max')
+            with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+                yield
+        else:
+            if te_installed:
+                raise RuntimeError('AMP_FP8 precision is used but current device does not support it.')
+            else:
+                raise ImportError(
+                    textwrap.dedent("""\
+                        AMP_FP8 precision is used but TransformerEngine is not installed.
+                        After making sure torch is already installed, please install it using
+                        pip install --upgrade git+https://github.com/NVIDIA/TransformerEngine.git@stable"""))
     else:
         raise ValueError(f'Unsupported precision: {precision}')
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -132,6 +132,10 @@ def get_torch_dtype(dtype: Union[Precision, str]):
         return torch.float16
     elif dtype in ['bfloat16', 'bfloat', 'torch.bfloat16', 'bf16', 'amp_bf16']:
         return torch.bfloat16
+    elif dtype in ['amp_fp8']:
+        # We use torch.bfloat16 by default for amp_fp8 as there is no
+        # fp8 datatype in PyTorch yet.
+        return torch.bfloat16
     else:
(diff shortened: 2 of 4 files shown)


composer/core/precision.py
```diff
@@ -54,5 +61,17 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
...
+    elif precision == Precision.AMP_FP8:
+        if te_installed and torch.cuda.get_device_capability()[0] > 8:
+            from transformer_engine.common.recipe import DelayedScaling, Format
+            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
+            fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo='max')
+            with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+                yield
+        else:
+            if te_installed:
+                raise ValueError(f'AMP_FP8 precision is used but current device does not support it.')
```
dakinggg, line 73: suggest `RuntimeError`

composer/core/precision.py
```diff
@@ -54,5 +61,17 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
...
+            else:
+                raise ValueError(f'AMP_FP8 precision is used but TransformerEngine is not installed.')
```
dakinggg, line 75: suggest `ImportError`

composer/core/precision.py
```diff
@@ -54,5 +61,17 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
         else:
             os.environ['XLA_USE_BF16'] = '1'
             yield
+    elif precision == Precision.AMP_FP8:
+        if te_installed and torch.cuda.get_device_capability()[0] > 8:
+            from transformer_engine.common.recipe import DelayedScaling, Format
+            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
```
dakinggg, line 67: Could you leave a comment indicating how you chose the settings below?

dakinggg (commented):
Could you add some brief docs to the precision class about this new option? Also, is this something you want to add to `setup.py` as an optional dependency, or too soon? Lastly, is there any way to write simple tests for this?

composer/core/precision.py
mvpatel2000: I think `ValueError` is correct -- we generally use it for invalid parameters to Trainer. For example, see the else case here

mvpatel2000 (commented):
Is there a target for `transformer_engine` in `setup.py` (is it already a dependency somewhere)? if not, can we added it?

dskhudia:
> Lastly, is there any way to write simple tests for this? 

@dakinggg Since it's not really usable on current hardware I can add some negative tests where we check for the errors on current hardware. Something like this useful?

dskhudia:
> Is there a target for transformer_engine in setup.py

> Also, is this something you want to add to setup.py as an optional dependency, or too soon?

Added a target txengine but two issues: 1) there is no pypi package for it 2) Installation depends on torch so it fails due to pip build isolation :-(

dakinggg:
Yeah, a negative test that errors would be helpful for now. Just to exercise the code path. Thanks! And hm, installation depending on torch is unfortunate. In that case, could you put installation instructions somewhere? Probably in the documentation for it fp8?

composer/core/precision.py
```diff
@@ -54,5 +62,21 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
...
+            # https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#id1
+            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
+            fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo='max')
+            with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+                yield
+        else:
+            if te_installed:
+                raise RuntimeError(f'AMP_FP8 precision is used but current device does not support it.')
+            else:
+                raise ImportError(f'AMP_FP8 precision is used but TransformerEngine is not installed.')
```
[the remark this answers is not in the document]
dskhudia: Cool. Let me add it here.

dskhudia:
@dakinggg added tests
@mvpatel2000 added installation instructions

tests/test_precision.py
```diff
@@ -87,3 +93,13 @@ def test_predict_precision_memory(precision: Precision):
...
+
+
+def test_amp_fp8_path():
+    trainer = get_trainer(Precision.AMP_FP8)
+    if te_installed:
+        with pytest.raises(RuntimeError, match='AMP_FP8 precision is used but current device does not support it'):
+            trainer.fit()
+    else:
+        with pytest.raises(ImportError, match='AMP_FP8 precision is used but TransformerEngine is not installed'):
+            trainer.fit()
```
mvpatel2000, line 105: This assumes we're on A100? Is there a way to check device from torch to verify this for first case? 

Idt merge blocking, but would be nice

mvpatel2000 (approved):

dakinggg (approved):

tests/test_precision.py
dskhudia: Will add. We have the same check in the precision itself.

lukaemon:
New to `fp8` or `bf8`. Does this mean it's possible to do `fp8` training with consumer level ada GPU like rtx 4090?

dskhudia:
@lukaemon : [Two variants of FP8 exist](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html) and there is nothing called bf8 on NVidia cards. 

> training with consumer level ada GPU

I think so. Provided you have CUDA 12, TransformerEngine layers in your model and using amp_fp8 precision. 
[NVidia announcement does point to FP8 in 4090](https://www.nvidia.com/en-us/geforce/news/rtx-40-series-graphics-cards-announcements/). 

> Ada’s new 4th Generation Tensor Cores are unbelievably fast, with an all new 8-Bit Floating Point (FP8) Tensor Engine, increasing throughput by up to 5X, to 1.32 Tensor-petaFLOPS on the GeForce RTX 4090.

lukaemon:
@dskhudia Thanks for clarification.

vgoklani:
@dskhudia No!!! 

https://github.com/NVIDIA/TransformerEngine/issues/15

vgoklani:
@lukaemon @dskhudia No!!!

https://github.com/NVIDIA/TransformerEngine/issues/15

lukaemon:
@vgoklani Thanks. Ada fp8 support after 23q2 at least.

composer/trainer/dist_strategy.py
```diff
@@ -132,6 +132,10 @@ def get_torch_dtype(dtype: Union[Precision, str]):
         return torch.float16
     elif dtype in ['bfloat16', 'bfloat', 'torch.bfloat16', 'bf16', 'amp_bf16']:
         return torch.bfloat16
+    elif dtype in ['amp_fp8']:
+        # We use torch.bfloat16 by default for amp_fp8 as there is no
+        # fp8 datatype in PyTorch yet.
+        return torch.bfloat16
```
thomasw21, line 138: Hi! New to your repo! Fp8 integration sounds super nice! I'm just trying to understand what this line implies? Does it mean that everything is running in bf16 when we specify fp8?

thomasw21: Nvm I think I misunderstood how the fp8 system worked I was somehow expecting to play with a specific dtype. Not the context manager. My bad

dskhudia: Glad that your confusion is resolved.

PR #1968, opened 2023-02-14
GHA release workflow, refactor PR and Daily workflows

bandish-shah:
# What does this PR do?

* Adds a `release.yaml` pipeline which can be triggered by the creation of a properly formatted tag (beginning with a 'v') or manually by repo admins.  The pipeline does the following:
  1. Runs code quality checks
  1. Builds source and wheel distributions
  1. Publishes PyPi production packages if triggered by by the `push` of a tag.  Otherwise builds and pushes packages to Test PyPi.
  1. Builds and pushes production Docker images if triggered by the `push` of a tag, skipped otherwise.  Production Docker build and push can be performed independently by triggering the `release-docker` pipeline manually.

* Adds a `release-docker.yaml` pipeline which builds and pushes production Docker images.  This flow can be triggered manually by repo admins

* Adds `coverage`, `pytest-cpu`, `pytest-gpu` and `docker-configure-build-push` reusable workflows

* Adds/refactors PR `pr-docker`, `pr-cpu`, `pr-gpu` pipelines to use reusable workflows

* Refacxtors `pytest-daily` to `daily` pipeline and uses reusable workflows

## Future Work (addressed by future PRs)

* Update `daily` to build Docker images and run tests on staged images
* Update Composer Docker builds to be gated by PyTorch images

Closes [CO-1186](https://mosaicml.atlassian.net/browse/CO-1186)

# Testing

All workflows were tested manually by adding the `pull_request` trigger to all workflows and running in GHA.  This trigger has been removed from the `daily`, `release-docker` and `release` workflows.  The `pr-gpu` workflow has been changed to use the `pull_request_target` workflow since it needs to provide secrets to PR's from forked repos.  Links to passing GHA runs for these workflows can be found below.

`pr-gpu`: https://github.com/mosaicml/composer/actions/runs/4199439168

`daily`: https://github.com/mosaicml/composer/actions/runs/4208510415

`release-docker`: https://github.com/mosaicml/composer/actions/runs/4208510424

`release`: https://github.com/mosaicml/composer/actions/runs/4210173449

# Before submitting

[CO-1186]: https://mosaicml.atlassian.net/browse/CO-1186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/pr-docker.yaml
```diff
@@ -1,13 +1,14 @@
 name: PR Docker
 on:
-  pull_request:
+  pull_request_target:
```
dakinggg, line 3: I think this is frowned upon because it would allow someone to open a PR with a change to this file that dumps our secrets

.github/workflows/pr-gpu.yaml
```diff
@@ -0,0 +1,33 @@
+name: PR GPU tests
+on:
+  pull_request_target:
```
dakinggg, line 3: same concern about secrets

.github/workflows/pr-gpu.yaml
```diff
@@ -0,0 +1,33 @@
...
+            pytest_command: 'coverage run -m pytest'
+    name: ${{ matrix.name }}
+    if: github.repository_owner == 'mosaicml'
+    with:
+      container: ${{ matrix.container }}
+      mcloud-timeout: 1200
+      name: ${{ matrix.name }}
+      pytest-command: ${{ matrix.pytest_command }}
+      pytest-markers: ${{ matrix.markers }}
+      python-version: 3.9
```
dakinggg, line 31: Is this used anywhere?

.github/workflows/pr-cpu.yaml
```diff
@@ -0,0 +1,53 @@
...
+            markers: 'not daily and not remote and not gpu and not vision and doctest'
+            pytest_command: 'coverage run -m pytest tests/test_docs.py'
+    name: ${{ matrix.name }}
+    if: github.repository_owner == 'mosaicml'
+    with:
+      container: ${{ matrix.container }}
+      name: ${{ matrix.name }}
+      pytest-command: ${{ matrix.pytest_command }}
+      pytest-markers: ${{ matrix.markers }}
+      python-version: 3.9
```
dakinggg, line 46: Is this used anywhere?

.github/workflows/pytest-cpu.yaml
```diff
@@ -1,48 +1,47 @@
...
+      pytest-command:
+        required: true
+        type: string
+      pytest-markers:
+        required: true
+        type: string
+      pytest-s3-bucket:
+        required: false
+        type: string
+      python-version:
```
dakinggg, line 20: Is this used anywhere?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - code-quality
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.x"
```
dakinggg, line 25: maybe this should be 3.9 to be consistent with other places

.github/workflows/release-docker.yaml
```diff
@@ -0,0 +1,63 @@
...
+        CUDA_VERSION=${{ matrix.CUDA_VERSION }}
+        MOFED_VERSION=${{ matrix.MOFED_VERSION }}
+        PYTHON_VERSION=${{ matrix.PYTHON_VERSION }}
+        PYTORCH_VERSION=${{ matrix.PYTORCH_VERSION }}
+        TORCHTEXT_VERSION=${{ matrix.TORCHTEXT_VERSION }}
+        TORCHVISION_VERSION=${{ matrix.TORCHVISION_VERSION }}
+      context: ./docker
+      image-name: ${{ matrix.IMAGE_NAME }}
+      image-uuid: ${{ matrix.UUID }}
+      push: false
```
dakinggg, line 57: I didn't quite understand. Isn't the "release" pipeline supposed to push docker images?

dakinggg (changes requested):
Blocking because of my question about secrets, otherwise LGTM and we can ship and fix

.github/workflows/pr-docker.yaml
dakinggg: Just to clarify, could someone change this workflow to run a (new) script which they check in and dumps secrets?

mvpatel2000: If a PR changes this workflow, it won't run in that PR -- it has to be merged for it to run.

It's the same protection against someone adding a new workflow which uses `pull_request_target` that tries to dump secrets

dakinggg: got it, thanks

dakinggg: can we document what would be an unsafe thing to do directly in this yaml? I'd like to lower the likelihood of someone later changing this file and doing something unsafe

.github/workflows/release-docker.yaml
mvpatel2000: Also confused about this

.github/workflows/docker-configure-build-push.yaml
```diff
@@ -0,0 +1,84 @@
...
+            IMAGE_TAG=${STAGING_REPO}:${{ inputs.image-uuid }}
+            IMAGE_CACHE="${STAGING_REPO}:${{ inputs.image-name }}-buildcache"
+          else
+            IMAGE_TAG=${{ inputs.tags }}
+            IMAGE_CACHE="${IMAGE_TAG/,*/}-buildcache"
+          fi
+
+          echo "IMAGE_TAG=${IMAGE_TAG}" >> ${GITHUB_ENV}
+          echo "IMAGE_CACHE=${IMAGE_CACHE}" >> ${GITHUB_ENV}
+
```
mvpatel2000, line 74: Can we echo the full name we're using here a separate step called "[README] Docker Image Name" or something to that effect? Common use case is to open PR changing dockerfile, get staging image, and test, and it seems like it's currently a pain to actually dig through and find the right tag here

mvpatel2000 (commented):
Mostly LGTM. Left 1 minor comment, and also confused by Daniel's points on python 3.9 and the push: false

.github/workflows/pr-docker.yaml
mvpatel2000: Yea, we should definitely put a comment here on why we're using `pull_request_target` and why it's ok

bandish-shah: Yea it's not ideal, there are two safeguards:
1. CI runs for PRs from external contributors must be approved by a maintainer
2. `pull_request_target` runs in the context of the target repo, not the merge commit 

1) safeguards against a malicious external user running anything (assuming the reviewer is diligent) and 2) prevents the external user from running malicious code (if the reviewer isn't diligent) unless that malicious code has already been committed to the target repo first.  It's not fool proof but it does make the case better.  More details on how all this works can be found here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

We could also just assume that anyone making changes to workflows already has write access to the repo and change this to a `pull_request` trigger.  It would just mean that maintainers doing CI infra work can't work from forks which is probably ok.  Let me know what you think.

.github/workflows/pr-gpu.yaml
bandish-shah: This unfortunately does need `pull_request_target` since the MCloud API secret needs to be passed.  In addition to 1) and 2) previously, the other safeguard in this case is that malicious code is actually sandboxed within the CI cluster so the damage is contained.  We do not pass any GH permissions to the code running on the MCloud CI cluster so this is probably fine.

.github/workflows/pr-gpu.yaml
bandish-shah: Yea these are inputs to the `pytest-gpu.yaml` reusable workflow.  The syntax is a bit confusing but think about it as the `pr-gpu.yaml` instantiates the `pytest-gpu.yaml` workflow and these are input arguments.

.github/workflows/pr-cpu.yaml
bandish-shah: Inputs to `pytest-cpu.yaml` workflow.

.github/workflows/pytest-cpu.yaml
bandish-shah: Good catch, it's not but it probably should be.  Fix coming.

.github/workflows/release-docker.yaml
bandish-shah: Good catch, had that as false before bc I didn't want to accidentally push new images.

.github/workflows/pr-docker.yaml
bandish-shah: I'm actually just going to go ahead and change this to `pull_request`.

.github/workflows/pytest-cpu.yaml
```diff
@@ -1,48 +1,51 @@
...
+    runs-on: ubuntu-latest
+    container: ${{ inputs.container }}
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ inputs.python-version }}
```
mvpatel2000, line 48: This should be removed so it doesn't override the python version in CPU image

mvpatel2000 (approved):
LGTM. There's enough moving pieces though that I would not merge until @dakinggg  also approves

dakinggg (approved):
LGTM, lets see if it works!

PR #1978, opened 2023-02-17
Fix ICL race conditions

dakinggg:
# What does this PR do?
Fixes two issues with ICL
- only downloads the dataset file once per node
- requires `destination_path` arg (and makes it possible to pass through the dataloader helper) to avoid accidentally overwriting files and loading the wrong ones

[CO-1796](https://mosaicml.atlassian.net/browse/CO-1796)

# Before submitting

[CO-1796]: https://mosaicml.atlassian.net/browse/CO-1796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -462,7 +462,7 @@ def local_rank_zero_download_and_wait(expected_file_path: str):
     It is expected that the file will be created by local rank zero. This function is useful
     as an alternative to ``run_local_rank_zero_first`` when downloading a file, because it does
     not require dist to be initialized. It only requires that the ``LOCAL_RANK`` environment variable
-    is set.
+    is set. If dist is initialized, you probably want to use ``run_local_rank_zero_first`` instead.
 
     Args:
         expected_file_path (str): The file to wait for existence of
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -459,7 +459,10 @@ def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool = False, s
 def local_rank_zero_download_and_wait(expected_file_path: str):
     """Context manager to wait for a file to exist on all ranks except local rank zero.
 
-    It is expected that the file will be created by local rank zero.
+    It is expected that the file will be created by local rank zero. This function is useful
+    as an alternative to ``run_local_rank_zero_first`` when downloading a file, because it does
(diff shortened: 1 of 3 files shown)


composer/utils/dist.py
```diff
@@ -454,6 +455,23 @@ def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool = False, s
...
+
+    Args:
+        expected_file_path (str): The file to wait for existence of
+    """
+    local_rank = get_local_rank()
+    if local_rank != 0:
+        while not os.path.exists(expected_file_path):
+            time.sleep(0.1)
+
+    yield
```
mvpatel2000, line 472: Can we instead use `run_local_rank_zero_first`? It's the same functionality -- it will run the context manager for rank 0 and then only continue on other ranks until after the context manager ends aka download is finished and data exists locally.

dakinggg: We can't use that because it requires dist to be initialized, which we don't want to force a user to do before creating their dataset.

mvpatel2000: Can you specify in docstring the difference and when to use this one over the other one?

dakinggg: Yup, will do

composer/utils/dist.py
```diff
@@ -454,6 +455,26 @@ def get_sampler(dataset: torch.utils.data.Dataset, *, drop_last: bool = False, s
...
 
 
+@contextmanager
+def local_rank_zero_download_and_wait(expected_file_path: str):
+    """Context manager to wait for a file to exist on all ranks except local rank zero.
+
+    It is expected that the file will be created by local rank zero. This function is useful
+    as an alternative to ``run_local_rank_zero_first`` when downloading a file, because it does
+    not require dist to be initialized. It only requires that the ``LOCAL_RANK`` environment variable
+    is set. If dist is initialized, you probably want to use ``run_local_rank_zero_first`` instead.
```
mvpatel2000, line 465:
```suggestion
    is set. If dist is initialized, you should use ``run_local_rank_zero_first`` instead to avoid busy waiting.
```

mvpatel2000 (approved):

karan6181 (approved):

PR #1979, opened 2023-02-17
Fix Logger During Predict

mvpatel2000:
# What does this PR do?

During predict, we currently use `self._state.timestamp`. This always returns the same value, so all logged metrics are overridden by the latest version. Instead, we should be using the predict timestamp during predict (which we know because predict always sets a fixed dataloader label).

Before (crouching crab) and after (tunneling macaw)
image

[CO-1797](https://mosaicml.atlassian.net/browse/CO-1797)

[CO-1797]: https://mosaicml.atlassian.net/browse/CO-1797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/loggers/logger.py
+++ b/composer/loggers/logger.py
@@ -52,6 +52,11 @@ class Logger:
         self.destinations = ensure_tuple(destinations)
         self._state = state
 
+    def get_step(self):
+        if self._state.dataloader_label == 'predict':
+            return self._state.predict_timestamp.batch.value
+        return self._state.timestamp.batch.value
+
     def log_traces(self, traces: Dict[str, Any]):
         for destination in self.destinations:
             destination.log_traces(traces)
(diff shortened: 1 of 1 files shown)


mvpatel2000:
@eracah what are the implications of this during eval? I think it will also break similarly if I just called eval?

mvpatel2000:
While this is a very nice quality of life improvement, it technically breaks things if a user names their dataloader "predict" :( so can't merge

dakinggg:
Could you just add something to state that says when predict is active instead?

mvpatel2000:
> Could you just add something to state that says when predict is active instead?

i think we probably want a state var that tracks train / predict / eval state. Will add to backlog

PR #1987, opened 2023-02-22
Speed monitor refactor

mvpatel2000:
# What does this PR do?

Revamps Speed Monitor to track a whole bunch of metrics from per device to MFU. This is mostly upstreamed from examples repo

Changes:
- replace flops for forward with `flops_per_batch`, letting user specify assumptions of forward vs. backward
- minor refactoring, use try / except when checking for `state.dataloader.dataset.max_seq_len `
- some tests

Screenshot:
image
w diffusion models

[CO-1761](https://mosaicml.atlassian.net/browse/CO-1761)

[CO-1761]: https://mosaicml.atlassian.net/browse/CO-1761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvpatel2000:
Discussing offline with Vitaliy about if we want to include the cost estimate stuff. PR is being held until that is finalized. Would like review from examples repo folks that the changes ive made to speed monitor are fine in description

composer/callbacks/speed_monitor.py
```diff
@@ -41,84 +161,139 @@ class SpeedMonitor(Callback):
...
+                model_flops_per_batch = composer_model.flops_per_batch  # type: ignore
+                flops_per_batch = None
+                if isinstance(model_flops_per_batch, (int, float)):
+                    flops_per_batch = model_flops_per_batch
+                elif isinstance(model_flops_per_batch, Callable):
+                    flops_per_batch = model_flops_per_batch(state.batch)
+                else:
+                    raise TypeError(f'flops_per_batch must be int, float, or callable accepting a batch and '
+                                    'returning an int or float. Instead, got {type(flops_per_batch)}.')
+                flops_per_sec = flops_per_batch * samples_per_sec
```
dakinggg, line 283: I may be getting confused between samples and batches, but i think this should be `* batches_per_sec`. This also gives me pause on a design question. `flops_ber_batch` is hard thing to put a number on, because it depends on batch size, which the model does not know about.

composer/callbacks/speed_monitor.py
```diff
@@ -41,84 +161,139 @@ class SpeedMonitor(Callback):
...
         # Track the batch num samples and wct to compute throughput over a window of batches
-        self.batch_start_num_samples = 0
-        self.batch_start_wct = 0.0
-        self.batch_wct_buffer: Deque[float] = deque(maxlen=window_size)
-        self.batch_num_samples_buffer: Deque[int] = deque(maxlen=window_size)
-        self.window_size = window_size
+        self.history_samples: Deque[int] = deque(maxlen=window_size + 1)
+        self.history_wct: Deque[float] = deque(maxlen=window_size + 1)
+
+        self.set_gpu_flops_available = False
```
dakinggg, line 222: I think this is unused?

composer/trainer/trainer.py
```diff
@@ -2207,8 +2207,10 @@ def _train_microbatches(self,
                     except TypeError:
                         optimizer.zero_grad()
 
-            # tracker for gradient accumulation
+            # Tracker for gradient accumulation
             current_batch_size = sum([self._train_data_spec.get_num_samples_in_batch(batch) for batch in microbatches])
+            # Cache batch, which will be overwritten by microbatches. Restore after microbatches complete
```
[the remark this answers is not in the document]
mvpatel2000: Yea, unfortunately we have to cache here as well because we want to restore before `AFTER_TRAIN_BATCH` event. There is another cache elsewhere in `train_microbatch` which is unnecessary though... removing. I think this was a confusion from a merge conflict 6 mo ago...

composer/callbacks/speed_monitor.py
```diff
@@ -41,84 +161,139 @@ class SpeedMonitor(Callback):
...
+            elapsed_batches = len(self.history_samples) - 1
+            elapsed_samples = int(self.history_samples[-1]) - int(self.history_samples[0])
+            elapsed_wct = self.history_wct[-1] - self.history_wct[0]
+            batches_per_sec = elapsed_batches / elapsed_wct
+            samples_per_sec = elapsed_samples / elapsed_wct
+            dev_batches_per_sec = batches_per_sec / world_size
+            dev_samples_per_sec = samples_per_sec / world_size
+            logger.log_metrics({'throughput/batches_per_sec': batches_per_sec})
+            logger.log_metrics({'throughput/samples_per_sec': samples_per_sec})
+            logger.log_metrics({'throughput/device/batches_per_sec': dev_batches_per_sec})
```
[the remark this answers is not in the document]
mvpatel2000: imo let's always log to be consistent. I imagine someone wants to do 1, 2, 4, 8... and compare per device throughput in the same graph

composer/trainer/trainer.py
```diff
@@ -2207,8 +2207,10 @@ def _train_microbatches(self,
...
+            current_batch = self.state.batch
```
dakinggg, line 2213: does this need to be a deepcopy for this to work in case something modifies the batch during `_train_microbatch`?

mvpatel2000: I don't think anything should modify it inside `_train_microbatch`. Maybe an algorithm changes it? but then we probably want the changed version (eg if someone is using a subset of samples?)

I would like to avoid deepcopy for performance reasons -- already seeing overhead from keeping batch on device in diffusion work

dakinggg: yeah i understand the performance implications. and yeah, an algorithm could change the batch. it would certainly be possible to write an algorithm that messed up this batch caching...but i guess thats ok

PR #1988, opened 2023-02-22
Fix failing sharded_checkpoint tests that fail when pytorch 1.13 is not installed

eracah:
# What does this PR do?

Skip all the tests in `test_sharded_checkpoint.py` IF pytorch 1.13 is not installed. This is because these tests use FSDP and if Composer uses FSDP and pytorch 1.13 is not installed, the process will fail. 

These failing tests do not mean anything will break for user, it just means the tests fail when they shouldn't

TLDR: add a pytest.skip call (conditional on pytorch 1.13 is not being installed) to all tests in `test_sharded_checkpoint.py`


tests/trainer/test_sharded_checkpoint.py
```diff
@@ -91,6 +91,8 @@ def _compare_model_params_between_state_dicts(state_dict1, state_dict2):
 @pytest.mark.gpu
 @world_size(2)
 def test_fsdp_full_state_dict_save(world_size, tmp_path: pathlib.Path):
+    if version.parse(torch.__version__) < version.parse('1.13.0'):
+        pytest.skip()
```
bandish-shah, line 95: I think it would be better to use a pytest marker here, for example:
https://github.com/mosaicml/composer/blob/be406894265e6540c2a5d603a6b66d499432ce73/tests/trainer/test_fsdp.py#L19

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -177,6 +179,8 @@ def test_fsdp_full_state_dict_save(world_size, tmp_path: pathlib.Path):
 @pytest.mark.gpu
 @world_size(2)
 def test_fsdp_full_state_dict_load(world_size, tmp_path: pathlib.Path):
+    if version.parse(torch.__version__) < version.parse('1.13.0'):
+        pytest.skip()
```
bandish-shah, line 183: Should probably do a pytest marker, see above.

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -305,6 +309,8 @@ def test_fsdp_partitioned_state_dict_save(world_size, tmp_path: pathlib.Path, st
 @world_size(2)
 @pytest.mark.parametrize('state_dict_type', ['local', 'sharded'])
 def test_fsdp_partitioned_state_dict_load(world_size, tmp_path: pathlib.Path, state_dict_type: str):
+    if version.parse(torch.__version__) < version.parse('1.13.0'):
+        pytest.skip()
```
bandish-shah, line 313: Should probably do a pytest marker, see above.

bandish-shah (changes requested):
See comments, better to use `@pytest.mark.skipif()` decorator.  Also would be good to give a clear reason as to why this test is skipped vs just spitting out a PyTorch version.

bandish-shah (approved):
LGTM thanks for the changes!

PR #1990, opened 2023-02-22
Add some logic to pass the correct github ref to mcp script

bandish-shah:
# What does this PR do?

Adds logic to the `pytest-gpu` workflow to pass the correct Github reference to the `mcp` test runner.  If the workflow is triggered by an event in a pull request, it will use the PR number.  Otherwise the commit SHA will be used if the run is on a protected branch, else the branch name is used.

# Before submitting

### DIFF
--- a/.github/workflows/pytest-gpu.yaml
+++ b/.github/workflows/pytest-gpu.yaml
@@ -58,5 +58,19 @@ jobs:
         id: tests
         run: |
           set -ex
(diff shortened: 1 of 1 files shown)


mvpatel2000 (approved):
Approving to unblock -- I assume this will take a few iterations.... Assuming ur testing as much as possible before merging

bandish-shah:
I tested the `daily` by manually adding the `pull_request` trigger in the PR.  Just removed before merging.

PR #1991, opened 2023-02-22
Runtime estimator

mvpatel2000:
# What does this PR do?

Estimates remaining runtime. Uses a simple timer interpolation + correction for eval. Graph shows a "normal" run and a corrected run with eval.

image

Note that this doesn't always work super smoothly, e.g. resnet estimates are not great because first epoch is a lot slower due to dataloader issues.

image

We will probably hit more issues as well later on, so this will require corrections.

Import Questions:
- is there a different place to log this? `wall_clock/` is intuitive, but everything else is in seconds so we need to report in seconds (whereas we would prefer to probably report hours?)

[CO-1664](https://mosaicml.atlassian.net/browse/CO-1664)

[CO-1664]: https://mosaicml.atlassian.net/browse/CO-1664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvpatel2000:
Note: The weird spikes in the wandb graph are because wandb associates the last logged time for a given step, which makes graph weird when x-axis is time. I spent a while investigating -- happy to discuss offline since it's a little too complicated to writeup. As a side effect, @eracah and I discovered where timestamp is advanced for batches vs. epochs compared to when metrics are calculated is inconsistent -- along for the ride I reorder the epoch one to be consistent.

Glabred (changes requested):
say sorry to me

Glabred (commented):
jk

composer/callbacks/runtime_estimator.py
```diff
@@ -0,0 +1,178 @@
...
+                return state.timestamp.get('ba').value / (state.max_duration.value * batches_per_epoch)
+            elif state.dataloader_len is not None:
+                return state.timestamp.get('ba').value / (state.max_duration.value * state.dataloader_len.value)
+        elapsed_dur = state.get_elapsed_duration()
+        if elapsed_dur is not None:
+            return elapsed_dur.value
+        return None
+
+    def batch_start(self, state: State, logger: Logger) -> None:
+        print(f'Batch start: {state.timestamp.get("ba")}')
```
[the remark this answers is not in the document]
dakinggg: didn't comment on all of them, but there are a bunch of prints that should probably be logs?

composer/callbacks/runtime_estimator.py
```diff
@@ -0,0 +1,178 @@
...
+                f'Elapsed time: {elapsed_time}, elapsed duration: {elapsed_dur}, checkpoint duration: {self.start_dur}')
+            rate = elapsed_time / (elapsed_dur - self.start_dur)
+            remaining_time = rate * (1 - elapsed_dur)
+
+            print(f'Batch end. train remaining_time: {remaining_time}')
+
+            # Add remaining time from each evaluator using known frequencies. We explicitly compute
+            # frequency instead of using time interpolation to avoid saw tooth pattern in estimates
+            for dataloader_label, eval_wcts in self.eval_wct_per_label.items():
+                # Discard first eval_wct if possible as it often slower due to dataset downloading
```
dakinggg, line 139:
```suggestion
                # Discard first eval_wct if possible as it is often slower due to dataset downloading
```

composer/callbacks/runtime_estimator.py
```diff
@@ -0,0 +1,178 @@
...
+        self.total_eval_wct += state.eval_timestamp.total_wct.total_seconds()
+        # state.dataloader_label should always be non-None unless user explicitly sets evaluator
+        # label to None, ignoring type hints
+        assert state.dataloader_label is not None, 'evaluator label must not be None'
+        if state.dataloader_label not in self.eval_wct_per_label:
+            self.eval_wct_per_label[state.dataloader_label] = []
+        self.eval_wct_per_label[state.dataloader_label].append(state.eval_timestamp.total_wct.total_seconds())
+        elapsed_fraction = self.get_elapsed_duration(state)
+        if elapsed_fraction is None:
+            warnings.warn(
```
dakinggg, line 172: is this possible to hit? batch_start should set `self._enabled` to `False` if `max_duration` is `None`, so I think this would never be hit?

composer/callbacks/runtime_estimator.py
```diff
@@ -0,0 +1,178 @@
...
+                num_evals_finished = len(eval_wcts)
+                if num_evals_finished > 1:
+                    eval_wct_avg = sum(eval_wcts[1:]) / (num_evals_finished - 1)
+                else:
+                    eval_wct_avg = sum(eval_wcts) / num_evals_finished
+                eval_rate = self.eval_frequency_per_label[dataloader_label]
+                print(
+                    f'\tdataloader_label: {dataloader_label}, eval_wct_avg: {eval_wct_avg}, eval_rate: {eval_rate}, num_evals_finished: {num_evals_finished}'
+                )
+                if eval_rate > 0:
```
dakinggg, line 150: Is it possible for `eval_rate` to be zero? I would think not?

dakinggg (approved):
LGTM with some comments

composer/trainer/trainer.py
```diff
@@ -2034,6 +2032,8 @@ def _train_loop(self) -> None:
                         for scheduler in self.state.schedulers:
                             scheduler.step()
 
+                    self.state.timestamp = self.state.timestamp.to_next_epoch()
```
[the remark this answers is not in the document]
dakinggg: nvm, read your comment after reviewing.

dakinggg:
Also, I'm willing to merge this with just manual tests, but please do add some unit tests in a follow on pr

tests/common/models.py
```diff
@@ -4,7 +4,7 @@
 """Contains commonly used models that are shared across the test suite."""
 import copy
 from functools import partial
-from typing import Any, Dict, Tuple, Union
+from typing import Any, Dict, Optional, Tuple, Union
```
bcui19, line 7: Do you want to move these files to another PR -- they seem related to another task?

bcui19 (approved):
LGTM, after addressing comments. +1 to adding tests in a separate PR

tests/common/models.py
mvpatel2000: This is because the test is failing and blocking this PR :(

mvpatel2000:
Added a test

hanlint (approved):

Issue #1985, opened 2023-02-21
Use of dynamic flow control in blurpool prevents quantisation.


BrettRyland:
### Environment
```
$ composer_collect_env
Collecting system information...
---------------------------------
System Environment Report        
Created: 2023-02-21 11:21:52 CET
---------------------------------

PyTorch information
-------------------
PyTorch version: 2.0.0.dev20230127+cu118
Is debug build: False
CUDA used to build PyTorch: 11.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.1 LTS (x86_64)
GCC version: (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
Clang version: Could not collect
CMake version: version 3.25.0
Libc version: glibc-2.35

Python version: 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] (64-bit runtime)
Python platform: Linux-5.15.0-58-generic-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.0.140
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA GeForce GTX 1080 Ti
Nvidia driver version: 525.85.12
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.8.0
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.8.0
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.22.4
[pip3] numpy-quaternion==2022.4.2
[pip3] pytorch-ranger==0.1.1
[pip3] pytorch-triton==2.0.0+0d7e753227
[pip3] torch==2.0.0.dev20230127+cu118
[...]

PR #1995, opened 2023-02-23
Allow FX graph mode post-training dynamic quantisation of BlurConv2d operations.

BrettRyland:
# What does this PR do?

This PR extracts the dynamic control flow operations in `blur_2d` into static control flow operations in the `BlurConv2d` class while leaving the dynamic control flow operations in place for `BlurMaxPool2d`. This allows FX graph mode post-training dynamic quantisation to be applied to models where blurpool has been applied to convolutions (though not max pooling).

There still seems to be an issue with static quantisation however, which I'm unsure of how to fix:
```
RuntimeError: quantized::conv2d_prepack() is missing value for argument 'bias'. Declaration: quantized::conv2d_prepack(Tensor weight, Tensor? bias, int[] stride, int[] padding, int[] dilation, int groups) -> __torch__.torch.classes.quantized.Conv2dPackedParamsBase
```

This PR partially fixes issue #1985 in that blurpool applied to convolutions uses static control flow and no longer prevents FX graph mode post-training dynamic quantisation. Blurpool applied to max pool continues to use dynamic control flow as before.

# Before submitting
- [n] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [n/a] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))

### DIFF
--- a/composer/algorithms/blurpool/blurpool_layers.py
+++ b/composer/algorithms/blurpool/blurpool_layers.py
@@ -30,11 +30,15 @@ def _padding_for_filt_2d_same(filt: torch.Tensor):
     return int(torch.div(h, 2)), int(torch.div(w, 2))
 
 
-def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.Tensor] = None) -> torch.Tensor:
+def blur_2d(input: torch.Tensor, channels: int = 0, stride: _size_2_t = 1, filter: Optional[torch.Tensor] = None) -> torch.Tensor:
     """Applies a spatial low-pass filter.
 
     Args:
         input (torch.Tensor): A 4d tensor of shape NCHW
+        channels (int, optional): The number of channels in the input tensor.
+            If non-positive, then dynamic control flow is used to determine the number of channels.
+            If positive, then static control flow is used and the filter dimensions should be appropriate for
+            the input size (note: this condition is always satisfied for the default filter and non-zero input size).
         stride (int | tuple, optional): Stride(s) along H and W axes. If a single value is passed, this
             value is used for both dimensions.
         filter (torch.Tensor, optional): A 2d or 4d tensor to be cross-correlated with the input tensor
@@ -54,24 +58,29 @@ def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.T
     Returns:
         The blurred input
     """
-    _, c, h, w = input.shape
-    n_in_channels = c
 
     if filter is None:
         filter = _default_2d_filter()
-    if (filter.shape[0] == 1) and (n_in_channels > 1):
-        # filt is already a rank 4 tensor
-        filter = filter.repeat((n_in_channels, 1, 1, 1))
 
-    _, _, filter_h, filter_w = filter.shape
+    if channels < 1:  # Use Dynamic Control Flow
+        _, c, h, w = input.shape
+        n_in_channels = c
+        if (filter.shape[0] == 1) and (n_in_channels > 1):
+            # filt is already a rank 4 tensor
+            filter = filter.repeat((n_in_channels, 1, 1, 1))
+
     padding = _padding_for_filt_2d_same(filter)
 
-    if h + 2 * padding[0] < filter_h:
-        return input
-    if w + 2 * padding[1] < filter_w:
-        return input
+    if channels < 1:  # Use Dynamic Control Flow
+        _, _, filter_h, filter_w = filter.shape
+        if h + 2 * padding[0] < filter_h:
+            return input
+        if w + 2 * padding[1] < filter_w:
+            return input
 
-    return F.conv2d(input, filter, stride=stride, padding=padding, groups=n_in_channels, bias=None)
+        return F.conv2d(input, filter, stride=stride, padding=padding, groups=n_in_channels, bias=None)  # Use Dynamic Control Flow
+
+    return F.conv2d(input, filter, stride=stride, padding=padding, groups=channels, bias=None)  # Use Static Control Flow
 
 
 def blurmax_pool2d(input: torch.Tensor,
@@ -138,7 +147,7 @@ def blurmax_pool2d(input: torch.Tensor,
                         padding=padding,
                         dilation=dilation,
                         ceil_mode=ceil_mode)
-    return blur_2d(maxs, stride=stride, filter=filter)
+    return blur_2d(maxs, channels=0, stride=stride, filter=filter)
 
 
 class BlurMaxPool2d(nn.Module):
@@ -240,11 +249,11 @@ class BlurConv2d(nn.Module):
             assert stride is not None
             conv_stride = stride
             self.blur_stride = 1
-            blur_nchannels = in_channels
+            self.blur_nchannels = in_channels
         else:
             conv_stride = 1
             self.blur_stride = kernel_size if (stride is None) else stride
-            blur_nchannels = out_channels
+            self.blur_nchannels = out_channels
 
         self.conv = torch.nn.Conv2d(
             in_channels=in_channels,
@@ -261,7 +270,7 @@ class BlurConv2d(nn.Module):
         # this is the full 4d tensor we want; materialize it once, instead
         # of just-in-time during forward; we can do this in this class but
         # not the others because we know in_channels during __init__
-        filt = _default_2d_filter().repeat(blur_nchannels, 1, 1, 1)
+        filt = _default_2d_filter().repeat(self.blur_nchannels, 1, 1, 1)
         self.register_buffer('blur_filter', filt)
 
     def forward(self, input: torch.Tensor):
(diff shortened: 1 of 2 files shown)


composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -304,11 +318,14 @@ def from_conv2d(module: torch.nn.Conv2d, module_index: int = -1, blur_first: boo
...
 
-    def __init__(self, stride: _size_2_t = 2, padding: _size_2_t = 1) -> None:
+    def __init__(self, channels: int = 0, stride: _size_2_t = 2, padding: _size_2_t = 1) -> None:
         super(BlurPool2d, self).__init__()
+        self.channels = channels
         self.stride = stride
         self.padding = padding
         self.register_buffer('filt2d', _default_2d_filter())
+        if self.channels > 0:
+            self.filt2d = self.filt2d.repeat(channels, 1, 1, 1)
```
dblalock, line 328:
```suggestion
        self.register_buffer('blur_filter', _default_2d_filter())
        if self.channels > 0:
            self.filt2d = self.blur_filter.repeat(channels, 1, 1, 1)
```
since it might be 4d, rather than 2d, let's change the name

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -304,11 +318,14 @@ def from_conv2d(module: torch.nn.Conv2d, module_index: int = -1, blur_first: boo
...
 
     def forward(self, input: torch.Tensor):
-        return blur_2d(input, stride=self.stride, filter=self.filt2d)
+        return blur_2d(input, channels=self.channels, stride=self.stride, filter=self.filt2d)
```
dblalock, line 331:
```suggestion
        return blur_2d(input, channels=self.channels, stride=self.stride, filter=self.blur_filter)
```
(propagating name change)

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -54,24 +61,31 @@ def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.T
...
-        return input
-    if w + 2 * padding[1] < filter_w:
-        return input
+    if channels < 1:  # Use Dynamic Control Flow
+        _, c, h, w = input.shape
+        n_in_channels = c
+
+        if (filter.shape[0] == 1) and (n_in_channels > 1):
+            # filt is already a rank 4 tensor
+            filter = filter.repeat((n_in_channels, 1, 1, 1))
```
dblalock, line 76:
```suggestion
        _, channels, h, w = input.shape

        if (filter.shape[0] == 1) and (channels > 1):
            # assume filt is already a rank 4 tensor
            filter = filter.repeat((channels, 1, 1, 1))
```

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -138,7 +152,7 @@ def blurmax_pool2d(input: torch.Tensor,
                         padding=padding,
                         dilation=dilation,
                         ceil_mode=ceil_mode)
-    return blur_2d(maxs, stride=stride, filter=filter)
+    return blur_2d(maxs, channels=0, stride=stride, filter=filter)
```
dblalock, line 155:
```suggestion
    return blur_2d(maxs, channels=-1, stride=stride, filter=filter)
```
super minor, but would be nice to adhere to the reshape/view convention that -1 is what indicates "infer from the input"

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -54,24 +61,31 @@ def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.T
...
+
+        if (filter.shape[0] == 1) and (n_in_channels > 1):
+            # filt is already a rank 4 tensor
+            filter = filter.repeat((n_in_channels, 1, 1, 1))
+
+        _, _, filter_h, filter_w = filter.shape
+        if h + 2 * padding[0] < filter_h:
+            return input
+        if w + 2 * padding[1] < filter_w:
+            return input
```
dblalock, line 82:
```suggestion
        _, _, filter_h, filter_w = filter.shape
        if h + 2 * padding[0] < filter_h:
            return input
        if w + 2 * padding[1] < filter_w:
            return input
```
It's probably better to just have the conv2d fail than have inconsistent behavior depending on whether `channels` is inferred.

dskhudia:
@BrettRyland We will take a look at it and get back to you soon. Also will check the other quantization error.

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -54,24 +61,31 @@ def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.T
...
+
+        # Use Dynamic Control Flow
+        return F.conv2d(input, filter, stride=stride, padding=padding, groups=n_in_channels, bias=None)
 
-    return F.conv2d(input, filter, stride=stride, padding=padding, groups=n_in_channels, bias=None)
+    # Use Static Control Flow
```
dblalock, line 87:
```suggestion
```
(since we're now using the same conv2d call for static and dynamic cases)

composer/algorithms/blurpool/blurpool_layers.py
```diff
@@ -54,24 +61,31 @@ def blur_2d(input: torch.Tensor, stride: _size_2_t = 1, filter: Optional[torch.T
...
+            filter = filter.repeat((n_in_channels, 1, 1, 1))
+
+        _, _, filter_h, filter_w = filter.shape
+        if h + 2 * padding[0] < filter_h:
+            return input
+        if w + 2 * padding[1] < filter_w:
+            return input
+
+        # Use Dynamic Control Flow
+        return F.conv2d(input, filter, stride=stride, padding=padding, groups=n_in_channels, bias=None)
```
dblalock, line 85:
```suggestion
```
saves a few loc and slightly clearer if we just set `channels` dynamically and have a shared call to conv2d

dblalock (approved):
LGTM; basically just suggested a few local changes to polish it.

The one subtlety I found is that this would change the semantics when the filter spatial size is larger than the input + padding, depending on whether `channels` is supplied or inferred. So I suggested just removing the checks and having it always throw at the conv2d in this case.

The "suggestions" aren't super legible since Github won't let them span deletions, but the result is just this:

```python
    if channels < 1:  # Use dynamic control flow
        _, channels, h, w = input.shape

        if (filter.shape[0] == 1) and (channels > 1):
            # assume filt is already a rank 4 tensor
            filter = filter.repeat((channels, 1, 1, 1))

    return F.conv2d(input, filter, stride=stride, padding=padding, groups=channels, bias=None)
```

dskhudia:
@BrettRyland Your issue regarding `RuntimeError: quantized::conv2d_prepack() `seems like a bug in PyTorch fx tracing for quantization. A workaround for this is to not use keywords arguments for F.conv2d. 


Instead of `return F.conv2d(input, filter, stride=stride, padding=padding, groups=channels, bias=None)` the following works:
from torch.nn.modules.utils import _pair
`return F.conv2d(input, filter, None, _pair(stride), _pair(padding), _pair(1), channels)`

composer/algorithms/blurpool/blurpool_layers.py
BrettRyland: I'm not sure what change to the code is being suggested here.
Also, since the padding is generated from the filter in `_padding_for_filt_2d_same`, aren't these conditions only going to occur if either `h` or `w` is 0?

BrettRyland:
> @BrettRyland Your issue regarding `RuntimeError: quantized::conv2d_prepack() `seems like a bug in PyTorch fx tracing for quantization. A workaround for this is to not use keywords arguments for F.conv2d.
> 
> Instead of `return F.conv2d(input, filter, stride=stride, padding=padding, groups=channels, bias=None)` the following works: from torch.nn.modules.utils import _pair `return F.conv2d(input, filter, None, _pair(stride), _pair(padding), _pair(1), channels)`
> 
> Feel free to modify PR as suggested above.

Thanks for this. I've updated the PR with all of the above suggestions other than the one I commented on above.

dskhudia:
I created an issue for PyTorch with a minimal repro (https://github.com/pytorch/pytorch/issues/95492)

```
import torch
import torch.nn as nn

from torch.ao.quantization import get_default_qconfig, QConfigMapping
from torch.ao.quantization.quantize_fx import prepare_fx, convert_fx
from torch.nn import functional as F


class myConv2d_1(nn.Module):
    def __init__(self, in_channels, out_channels, kernel_size):
        super().__init__()
        self.in_channels = in_channels
        self.out_channels = out_channels
        self.kernel_size = kernel_size
        self.weight = nn.Parameter(torch.randn(out_channels, in_channels, kernel_size, kernel_size))

    def forward(self, input):
        return F.conv2d(input, self.weight, None, [1, 1], [0, 0], [1, 1], 1)

class myConv2d_2(nn.Module):
    def __init__(self, in_channels, out_channels, kernel_size):
        super().__init__()
        self.in_channels = in_channels
        self.out_channels = out_channels
        self.kernel_size = kernel_size
        self.weight = nn.Parameter(torch.randn(out_channels, in_channels, kernel_size, kernel_size))

    def forward(self, input):
        return F.conv2d(input, self.weight, bias=None, stride=[1, 1], padding=[0, 0], dilation=[1, 1], groups=1)

# This works
#model = myConv2d_1(8, 64, 3)

# This doesn't work
model = myConv2d_2(8, 64, 3)

model.eval()
qconfig = get_default_qconfig("fbgemm")
qconfig_mapping = QConfigMapping().set_global(qconfig)

example_inputs = torch.randn(1, 8, 224, 224)

prepared_model = prepare_fx(model, qconfig_mapping, example_inputs)
model(example_inputs)
quantized_model = convert_fx(prepared_model)
out_q = quantized_model(example_inputs)
quantized_model.print_readable()
```

dskhudia:
@dblalock will handle the suggestions in a followup PR. Merging this for the release.

dskhudia (approved):

dskhudia:
@dblalock : Just checked all your comments in detail and I don't think anything else is needed for this PR. Please let me know if you think otherwise.

PR #1998, opened 2023-02-24
Adjust auto microbatch size when hitting cuda alloc retries

mvpatel2000:
# What does this PR do?

If a run is repeatedly hitting cuda alloc retries, throughput will collapse due to memory thrashing. I finally found a reliable repro of this, which has let me update auto grad accum to catch this case and lower the microbatch size to fix it.

Along for the ride, correctly resets grad scaler which was missing in original implementation

Blue is before, green is after
image

[CO-1827](https://mosaicml.atlassian.net/browse/CO-1827)

[CO-1827]: https://mosaicml.atlassian.net/browse/CO-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/algorithms/seq_length_warmup/seq_length_warmup.py
+++ b/composer/algorithms/seq_length_warmup/seq_length_warmup.py
@@ -288,6 +288,7 @@ class SeqLengthWarmup(Algorithm):
             model_inputs = {k: v[:per_gpu_batch] for k, v in batch_clone.items()}
 
             found_cuda_oom = 0  # int since bool BOR not supported on all torch.distributed backends
+            num_alloc_retries = torch.cuda.memory_stats()['num_alloc_retries']
             try:
                 # start by running a forward and backward pass
                 # of the maximum sequence length to allocate cache.
@@ -304,6 +305,11 @@ class SeqLengthWarmup(Algorithm):
                 for optimizer in state.optimizers:
                     optimizer.zero_grad()
 
+                # Raise error if automicrobatching and num_alloc_retries increased, as thrashing
+                # often leads to throughput slowdown
+                if state.auto_microbatching and torch.cuda.memory_stats()['num_alloc_retries'] > num_alloc_retries:
+                    raise RuntimeError(
+                        'num_alloc_retries > 1, which leads to memory thrashing and throughput decrease.')
             # This error/state.grad_accum handling mimics the logic in trainer._train_batch().
             except RuntimeError as e:
                 if state.auto_microbatching and _is_cuda_oom(e):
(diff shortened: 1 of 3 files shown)


mvpatel2000:
Looking for help here... don't understand test failure

composer/algorithms/seq_length_warmup/seq_length_warmup.py
```diff
@@ -304,6 +306,10 @@ def _activate_model(self, state: State, logger: Logger) -> None:
                 for optimizer in state.optimizers:
                     optimizer.zero_grad()
 
+                # Raise error if automicrobatching and num_alloc_retries increased, as thrashing
+                # often leads to throughput slowdown
+                if state.auto_microbatching and _get_num_alloc_retries() > num_alloc_retries:
+                    raise RuntimeError('num_alloc_retries > 1 which causes memory thrashing and throughput decrease')
```
dskhudia, line 312: `num_alloc_retries > 1` doesn't match with `_get_num_alloc_retries() > num_alloc_retries:`

mvpatel2000: Good catch, should be > 0

composer/trainer/trainer.py
```diff
@@ -2142,6 +2158,10 @@ def _train_batch(self, use_grad_scaling: bool) -> Dict[str, torch.Tensor]:
                                     xm.optimizer_step(optimizer, barrier=True)
                                 else:
                                     optimizer.step()
+                # Raise error if automicrobatching and num_alloc_retries increased, as thrashing
```
dakinggg, line 2161: Is error correct here? I'm just wondering how sure we are that _any_ retries is _always_ an error.

dakinggg: if we're not sure, we could add a trainer arg `num_cuda_alloc_retries_allowed` or whatever, default it to zero. that way at least a user can escape this error if they want.

composer/trainer/trainer.py
```diff
@@ -257,6 +268,8 @@ def _adjust_grad_accum(state: State, device_batch_size: int):
         del state.loss
     for optimizer in state.optimizers:
         optimizer.zero_grad(set_to_none=True)
+    if state.scaler is not None:
```
dakinggg, line 271: what does this do/why is it related to this PR?

composer/trainer/trainer.py
bcui19: +1 to this, I think we might want some loose heuristic or something softer than just has an alloc_retry to raise an error. I've seen instances where alloc_retries just happens once during training and it's fine otherwise (so there's borderline no throughput impact).

bcui19: image
e.g. this model hit alloc_retry once, and never hit it again.

mvpatel2000:
Closing because of the graph Brandon showed -- this doesn't always work

PR #1999, opened 2023-02-24
Reset scaler state

mvpatel2000:
# What does this PR do?

We currently don't reset the scalar state when adjusting grad accum. This is a problem because if we OOM in the wrong spot, we'll hit an error about unscaling after step. This was caught, tested, and shown in graphs here https://github.com/mosaicml/composer/pull/1998

[CO-1830](https://mosaicml.atlassian.net/browse/CO-1830)

[CO-1830]: https://mosaicml.atlassian.net/browse/CO-1830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dakinggg (approved):
LGTM

PR #2013, opened 2023-02-28
Patch Speed Monitor MFU

mvpatel2000:
# What does this PR do?

As title. Fixes speed monitor bug which has a problem with flops/MFU. We now keep a record of flop utilization, as the values might be different depending on the level of padding in the batch. We also correctly sum the flop counts from all the devices

image

[CO-1838](https://mosaicml.atlassian.net/browse/CO-1838)

[CO-1838]: https://mosaicml.atlassian.net/browse/CO-1838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -294,7 +294,6 @@ class SpeedMonitor(Callback):
             if not isinstance(model_flops_per_batch, Callable):
                 raise TypeError('flops_per_batch must a callable accepting a batch and '
                                 f'returning an int or float. Instead, got {type(model_flops_per_batch)}.')
-            print(state.batch['input_ids'].shape)
             device_flops_per_batch = model_flops_per_batch(state.batch)
 
             # Sum flops across all ranks since each rank computes the flops for its own batch
@@ -303,7 +302,6 @@ class SpeedMonitor(Callback):
             dist.all_reduce(flops_per_batch_tensor, reduce_operation='SUM')
             flops_per_batch = flops_per_batch_tensor.item()
 
-            print(f'flops_per_batch: {flops_per_batch}, device_flops_per_batch: {device_flops_per_batch}')
             self.history_flops.append(flops_per_batch)
 
         # Log the flops throughput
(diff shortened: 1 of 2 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -284,22 +285,37 @@ def batch_end(self, state: State, logger: Logger):
...
+            model_flops_per_batch = composer_model.flops_per_batch  # type: ignore
+            if not isinstance(model_flops_per_batch, Callable):
+                raise TypeError('flops_per_batch must a callable accepting a batch and '
+                                f'returning an int or float. Instead, got {type(model_flops_per_batch)}.')
+            device_flops_per_batch = model_flops_per_batch(state.batch)
+
+            # Sum flops across all ranks since each rank computes the flops for its own batch
+            flops_per_batch_tensor = state.device.tensor_to_device(
+                torch.tensor(device_flops_per_batch, dtype=torch.float))
+            dist.all_reduce(flops_per_batch_tensor, reduce_operation='SUM')
```
abhi-mosaic, line 302: I thiiink there might be a `dist.all_reduce_object(..)` that could save you the `torch.tensor` stuff but don't worry about it unless you're confident that it's safe

mvpatel2000: Only have an `all_gather_object` :(

I'm going to leave off for now because I don't want to rerun the tests ahead of release deadline, but this is probably a reasonable update we'd want to do in dist since this tensor wrapping occurs in a few places

vchiley: `flops_per_batch_tensor  = sum(all_gather_object(device_flops_per_batch))`

composer/callbacks/speed_monitor.py
```diff
@@ -284,22 +285,37 @@ def batch_end(self, state: State, logger: Logger):
...
+            flops_per_batch = flops_per_batch_tensor.item()
```
vchiley, line 303:
```suggestion
            flops_per_batch  = sum(all_gather_object(device_flops_per_batch))
```

would that work?

vchiley (approved):
lgtm

(with one nit cmt)

PR #2015, opened 2023-02-28
Remove runtime estimator state dict

mvpatel2000:
# What does this PR do?

Removes any state information from runtime estimator. This simplifies checkpoints. Also fixes resumption, which was broken before. 

image

Example above. Note that the orange line starts below since it doesn't know about eval when resuming until first eval.

[CO-1851](https://mosaicml.atlassian.net/browse/CO-1851)

[CO-1851]: https://mosaicml.atlassian.net/browse/CO-1851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/algorithms/blurpool/blurpool_layers.py
+++ b/composer/algorithms/blurpool/blurpool_layers.py
@@ -9,6 +9,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.common_types import _size_2_t
+from torch.nn.modules.utils import _pair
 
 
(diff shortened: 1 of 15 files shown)


eracah (approved):
LGTM. Why did it have a state_dict to begin with?

mvpatel2000:
> LGTM. Why did it have a state_dict to begin with?

Bad design :(

PR #2016, opened 2023-02-28
Add multi-gpu testing to test_algorithm_resumption

eracah:
# What does this PR do?

Resuming some model-surgery algorithms using DDP in multi-gpu settings can cause issues due to module renaming.
There is a unit test (`test_algorithm_resumption.py::test_algorithm_resumption`) that checks for every algorithm whether you can train a model with that algorithm, save a checkpoint, then resume the training from that checkpoint with the algorithm. However, that unit test only checks when using one gpu. 

In this PR, we:
*  extend `test_algorithm_resumption.py::test_algorithm_resumption` to include a 2 gpu setting. 
* To enable that setting, we also modify `algorithm_settings.get_alg_dataloader` to include support for distributed samplers
* Add an xfail for the algorithm SWA for the 2-gpu setting since it has a known issue that we won't fix at the moment

fix CO-1848

### DIFF
--- a/composer/algorithms/gyro_dropout/gyro_dropout.py
+++ b/composer/algorithms/gyro_dropout/gyro_dropout.py
@@ -121,6 +121,8 @@ class GyroDropout(Algorithm):
         self.sigma = sigma
         self.tau = tau
 
+        log.warning('GyroDropout is not implemented in a way that allows correct resumption from checkpoint.')
+
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}()'
 
--- a/composer/algorithms/gyro_dropout/gyro_dropout.py
+++ b/composer/algorithms/gyro_dropout/gyro_dropout.py
@@ -121,7 +121,9 @@ class GyroDropout(Algorithm):
         self.sigma = sigma
         self.tau = tau
 
-        log.warning('GyroDropout is not implemented in a way that allows correct resumption from checkpoint, which may lead to incorrect behavior.')
+        log.warning(
+            'GyroDropout is not implemented in a way that allows correct resumption from checkpoint, which may lead to incorrect behavior.'
+        )
 
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}()'
--- a/composer/algorithms/gyro_dropout/gyro_dropout.py
+++ b/composer/algorithms/gyro_dropout/gyro_dropout.py
@@ -121,7 +121,7 @@ class GyroDropout(Algorithm):
         self.sigma = sigma
         self.tau = tau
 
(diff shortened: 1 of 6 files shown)


composer/algorithms/gyro_dropout/gyro_dropout.py
```diff
@@ -121,6 +121,8 @@ def __init__(self, p: float = 0.5, sigma: int = 256, tau: int = 16):
         self.sigma = sigma
         self.tau = tau
 
+        log.warning('GyroDropout is not implemented in a way that allows correct resumption from checkpoint.')
```
mvpatel2000, line 124:
```suggestion
        log.warning('GyroDropout is not implemented in a way that allows correct resumption from checkpoint, leading to potentially unanticipated behavior.')
```

composer/algorithms/sam/sam.py
```diff
@@ -147,6 +147,7 @@ def __init__(
         epsilon: float = 1.0e-12,
         interval: int = 1,
     ):
+        log.warning('SAM has known issues of weight mismatch when loading from a checkpoint')
```
mvpatel2000, line 150:
```suggestion
        log.warning('SAM has known issues of weight mismatch when loading from a checkpoint, which will cause an error when resuming without `load_weights_only=True`.')
```

composer/algorithms/stochastic_depth/stochastic_depth.py
```diff
@@ -139,6 +139,8 @@ def __init__(self,
                  drop_distribution: str = 'linear',
                  drop_warmup: Union[float, Time, str] = 0.0):
 
+        log.warning('Stochastic depth has known issues of weight mismatch when loading from a checkpoint')
```
mvpatel2000, line 142:
```suggestion
        log.warning('Stochastic depth has known issues of weight mismatch when loading from a checkpoint, which will cause an error when resuming without `load_weights_only=True`.')
```

composer/algorithms/swa/swa.py
```diff
@@ -114,6 +114,9 @@ def __init__(self,
                  anneal_strategy: str = 'linear',
                  anneal_steps: int = 10,
                  swa_lr: Optional[float] = None):
+
+        log.warning('Using SWA when saving then resuming from a checkpoint on multiple gpu'
+                    's has issues. To avoid this remove SWA on resumption or set load_weights_only to True')
```
mvpatel2000, line 119:
```suggestion
        log.warning('SWA has known issues when resuming from a checkpoint on multiple GPUs, which will cause an error when resuming without `load_weights_only=True`.')
```

mvpatel2000 (approved):
LGTM. Left a few nits on error messages to make them a little cleaner

PR #2018, opened 2023-03-01
Fix eval duplicate logging issue

mvpatel2000:
# What does this PR do?

Fix eval duplicate logging issue. Along for the ride, refactor and clean up console log testing.

image

fix CO-1769


eracah (approved):
LGTM. Thanks for doing this! Tests are much better now. Can you add a sample output in the description?

PR #2022, opened 2023-03-01
Use follow in mcp script

mvpatel2000:
# What does this PR do?

MCLI has upgraded to a new release which lets us safely follow logs when using msdk. This is a quality of life improvement which allows users to follow logs from GHA as a test runs.

[CO-1858](https://mosaicml.atlassian.net/browse/CO-1858)

[CO-1858]: https://mosaicml.atlassian.net/browse/CO-1858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bandish-shah:
What testing have you done on this?  It was pretty flaky so I have some concerns that we'll see failures after periodic use.  Also can you post a sample output of what users would see in GHA?

.github/mcp/mcp_pytest.py
```diff
@@ -83,28 +83,22 @@
...
-        run = wait_for_run_status(run, status='completed', timeout=args.timeout)
-    except TimeoutError:
-        print(f'Run timed out and did not complete in {args.timeout/60} minutes.')
-
-    # Get run status and stop run
-    success = run.status == RunStatus.COMPLETED
-    print(f'Run completed with status: {run.status} (success={success})')
-    if run.status == RunStatus.RUNNING:
-        stop_runs([run])
-        print('Run stopped.')
```
bandish-shah, line 103: Are we sure we want to relax this try/except?  Can we put it around `follow_run_logs` below and if that fails revert back to polling for status?

bandish-shah (approved):
LGTM, approving to unblock.  Please update PR details with example output and some information on how this was tested.

Also some food for thought, this seems like a significant relaxation for a feature that's been historically very flaky but I'm also open to putting safeguards in until it proves to be an issue.

.github/mcp/mcp_pytest.py
mvpatel2000: So the try except is only for the timeout error which I set -- it does not catch the 500 error MAPi was occasionally sending. I think it's fine to not include it

PR #2041, opened 2023-03-07
Checkpoints Simplified

mvpatel2000:
# What does this PR do?

Refactors checkpoint workflow to be simpler and cleaner. Fixes bug resulting in timeout with checkpoints.

Top lines are old way, bottom lines are corrected code:
image

Future work
- fix autoresume in same way

[CO-1881](https://mosaicml.atlassian.net/browse/CO-1881)
[CO-1883](https://mosaicml.atlassian.net/browse/CO-1883)

[CO-1881]: https://mosaicml.atlassian.net/browse/CO-1881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CO-1883]: https://mosaicml.atlassian.net/browse/CO-1883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/utils/checkpoint.py
```diff
@@ -248,88 +235,78 @@ def download_checkpoint(
...
-                    # or could not be downloaded
-                    raise RuntimeError(f'Checkpoint {path} does not exist')
-
-        if rank_zero_checkpoint_filepath != rank_n_checkpoint_filepath:
-            # every RANK needs ITS OWN checkpoint.
-            # But, the  global rank zero is a special case -- these files are the same!
-            assert dist.get_global_rank() != 0, 'invariant violation'
+    local_path = _format_path_with_current_rank(path)
+    local_rank_zero_path = _get_local_rank_zero_path(local_path)
+    logging.log(logging.WARN, f'local_path={local_path}, local_rank_zero_path={local_rank_zero_path}')
```
dakinggg, line 272:
```suggestion
    log.info(f'local_path={local_path}, local_rank_zero_path={local_rank_zero_path}')
```

composer/utils/checkpoint.py
```diff
@@ -248,88 +235,78 @@ def download_checkpoint(
     object_store: Optional[Union[ObjectStore, LoggerDestination]],
     progress_bar: bool,
     fsdp_sharded_state_dict_enabled: bool = False,
-) -> Tuple[str, Optional[str], bool]:
+    deepspeed_checkpoint: bool = False,
+) -> Tuple[str, Optional[str]]:
     """Download the checkpoint stored at ``path``, potentially in ``object_store``, to ``node_checkpoint_folder``.
 
     Returns a tuple of  (``composer_states_filepath``, ``extracted_checkpoint_folder``, ``extracted_rank_n``).
```
dakinggg, line 242: return type changed

composer/utils/checkpoint.py
```diff
@@ -248,88 +235,78 @@ def download_checkpoint(
...
-            # every RANK needs ITS OWN checkpoint.
-            # But, the  global rank zero is a special case -- these files are the same!
-            assert dist.get_global_rank() != 0, 'invariant violation'
+    local_path = _format_path_with_current_rank(path)
+    local_rank_zero_path = _get_local_rank_zero_path(local_path)
+    logging.log(logging.WARN, f'local_path={local_path}, local_rank_zero_path={local_rank_zero_path}')
 
+    try:
+        # Download on local rank 0 or if path is different from local rank 0
+        if dist.get_local_rank() == 0 or local_path != local_rank_zero_path:
```
dakinggg, line 276: shouldn't only deepspeed and sharded checkpoints try get file on all ranks?

dakinggg (approved):
I think this looks right to me

composer/utils/checkpoint.py
mvpatel2000: If you format your path as `...rank{rank}.pt` and it's different from rank 0, we try to download anyways. A user could be using a weird checkpoint which is different per rank for some reason, and if they pass that in, we try it. We don't raise an error on other ranks if the file doesn't exist (as we expect it to not exist)

eracah:
Nicely done! Sorry I didn't see it time, but LGTM since I was there when you made the fix

PR #2044, opened 2023-03-08
Allow `LPLayerNorm` to support `self.bias=None`

abhi-mosaic:
We are experimenting with removing all biases from our MosaicGPT models. When we do so with `torch.nn.LayerNorm` it works, but with `LPLayerNorm` it fails. This just adds a check on `if self.bias is None`.


abhi-mosaic:
@bandish-shah there is a moderate likelihood that our MosaicGPT in `examples` will need this feature, so until this patch is released in Composer, we likely cannot tag a release of Examples against it (or we will need to maintain a second branch or something like that)

Basically this might be worth an 0.13.2, if you have other stuff you want to get in there.

mvpatel2000:
@abhi-mosaic Can you please add a unit test ensuring this works? Similarly, can you please add the same fix and unit test to `LowPrecisionGroupNorm`? it's probably something simple like extending existing models they use in unit tests to have a flag for bias or not (with default on).

> Basically this might be worth an 0.13.2, if you have other stuff you want to get in there.

We do have a lot of checkpointing fixes coming in this week....

bandish-shah:
> Basically this might be worth an 0.13.2, if you have other stuff you want to get in there.

@abhi-mosaic let's hold off, there are other fixes in progress.  We can target the 0.13.2 hot patch for late next week.

composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
```diff
@@ -99,7 +99,7 @@ def forward(self, x):
         module_device = x.device
         downcast_x = _cast_if_autocast_enabled(x)
         downcast_weight = _cast_if_autocast_enabled(self.weight)
```
vchiley, line 101:
```suggestion
        downcast_weight = _cast_if_autocast_enabled(self.weight) if self.weight is not None else self.weight
```
needed if `elementwise_affine=False`

mvpatel2000 (approved):
LGTM! Thanks for finishing this

abhi-mosaic:
Thanks @nik-mosaic !!!

PR #2045, opened 2023-03-08
Add support for ICL QA tasks and generation during evaluation with `HuggingFaceModel`

dakinggg:
# What does this PR do?
This PR does a few things. The main purpose of the PR is to add support for In Context Learning evaluation of generative tasks (i.e. open ended QA). In service of this,
- adds a `generate` capability to `HuggingFaceModel`, and calls into `generate` during `eval_forward` on batches with `mode=generate`
- adds a `InContextLearningQADataset`, which supports a dataset format with a question, answer, and answer aliases
- adds a `InContextLearningQAAccuracy` metric which takes text as input, normalizes it, and checks if the predicted text is a prefix for any of the answer aliases
- fixes a bug with all ICL datasets where if the dataset file is attempted to be retrieved on every rank instead of just local rank 0

TODOs:
- [x] Add unit tests for the new `HuggingFaceModel` capabilities
- [x] update all the relevant documentation
- [x] One final GPT neo run computing multiple ICL metrics, just to be sure. https://wandb.ai/mosaic-ml/daniel-debug/runs/ha2p5ih2?workspace=user-danielking

Points of comparison that the metric computation is reasonable (the eleuther eval harness doesn't appear to do the text normalization):
- eleuther eval harness with neo-2.7b limit 1000, 5-shot: 0.115
- ours neo 2.7b full, 5-shot: 0.17
- llama 7b after ~300b tokens, 64 shot?: 0.45 *evaluation procedure not open source
- eleuther eval harness with opt-6.7b limit 1000, 5-shot: 0.259
- ours opt-6.7b full, 5-shot: 0.3

Limitations:
- all of the ICL tasks are not really set up for encoder-decoder. This one is not different. We can add support to all of them at the same time in the future.

[CO-1831](https://mosaicml.atlassian.net/browse/CO-1831)

# Before submitting

[CO-1831]: https://mosaicml.atlassian.net/browse/CO-1831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bmosaicml (approved):
Great work Daniel! I know this alias thing is sued for TriviaQA. Are there any other ICL datasets you expect this to be useful for?

tests/models/test_hf_model.py
```diff
@@ -649,3 +650,99 @@ def test_embedding_resizing(tiny_bert_model, tiny_bert_tokenizer, embedding_resi
...
+    if device == 'cpu' and use_fsdp:
+        pytest.skip('FSDP is not supported on CPU.')
+    if world_size == 1 and use_fsdp:
+        pytest.xfail((
+            'Generation with world size 1 and FSDP fails with '
+            '`RuntimeError: The tensor has a non-zero number of elements, '
+            'but its data is not allocated yet. Caffe2 uses a lazy allocation, '
+            'so you will need to call mutable_data() or raw_mutable_data() to actually allocate memory.` '
+            'This issue is resolved with world size > 1 by a dummy call to forward (see HuggingFaceModel.dummy_forward_called), '
+            'but for some reason fails with world size 1.'))
```
samhavens, line 668: ooof

composer/models/huggingface.py
```diff
@@ -403,6 +421,40 @@ def get_metadata(self):
                     }
         return {'model': model_output, 'tokenizer': tokenizer_output}
 
+    def generate(self,
+                 input_ids: torch.Tensor,
+                 num_beams: int = 1,
+                 do_sample: bool = False,
+                 max_new_tokens: int = 20,
```
samhavens, line 428: why these but not e.g. top_p?

samhavens: I assume we can pass that with kwargs? I am just curious why these are explicitly not in kwargs

samhavens (approved):
🚀

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -41,13 +41,24 @@ def _make_padded_input(context_enc, continuation_enc, max_seq_len, pad_tok_id):
...
+    elif padding_side == 'left':
+        inp = torch.cat(
+            [
+                torch.LongTensor((max_seq_len - inp_len) * [pad_tok_id]),
+                inp,  # [seq]
+            ],
+            dim=0,
+        )
+    else:
+        raise ValueError(f'Unknown padding_side {padding_side}')
```
mvpatel2000, line 61:
```suggestion
        raise ValueError(f"Unknown padding_side {padding_side}. padding_side must be either 'left' or 'right'")
```

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -70,6 +81,155 @@ def _get_fewshot_sample_idxs(dataset_size, num_fewshot, sample_idx):
...
+                 continuation_delimiter: str, destination_path: str, question_prelimiter: str):
+        try:
+            from datasets import load_dataset  # pyright: ignore [reportGeneralTypeIssues]
+        except ImportError as e:
+            raise MissingConditionalImportError(extra_deps_group='nlp',
+                                                conda_package='datasets',
+                                                conda_channel='conda-forge') from e
+        with dist.local_rank_zero_download_and_wait(destination_path):
+            if dist.get_local_rank() == 0:
+                get_file(dataset_uri, destination_path, overwrite=True)
```
mvpatel2000, line 120: Currently, we use `dist.local_rank_zero_download_and_wait` to avoid dist barrier. However, if the download fails, this results in a deadlock -> NCCL timeout which is pretty nasty. What if we instead added a context that runs only local rank 0 and then uses a signal file (ensuring the path to signal file exists or use `/tmp/some_signal_file`)? This way we avoid nasty deadlocks with download failures

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -70,6 +81,155 @@ def _get_fewshot_sample_idxs(dataset_size, num_fewshot, sample_idx):
...
+                    if len(preamble) > 0:
+                        ctxt = f'{example_delimiter}{ctxt}'
+                    preamble += f'{ctxt}{continuation_delimiter}{cont}'
+
+            ctxt = self.samples[sample_idx]['context']
+            ctxt = f'{question_prelimiter}{ctxt}'
+            if len(preamble) > 0:
+                ctxt = f'{example_delimiter}{ctxt}'
+
+            # rstrip the continuation delimiter, because the prompt ending in a space results in degenerate output
```
mvpatel2000, line 173: Are you serious? that's absurd...

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -70,6 +81,155 @@ def _get_fewshot_sample_idxs(dataset_size, num_fewshot, sample_idx):
...
+            continuation_delimiter_stripped = continuation_delimiter.rstrip()
+            ctxt = f'{ctxt}{continuation_delimiter_stripped}'
+
+            # if the preamble is empty then this will be a 0-length list, unless the tokenizer adds special tokens to empty strings (e.g. OPT tokenizer)
```
mvpatel2000, line 177:
```suggestion
            # If the preamble is empty then this will be a 0-length list, unless the tokenizer adds special tokens to empty strings (e.g. OPT tokenizer)
```

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -70,6 +81,155 @@ def _get_fewshot_sample_idxs(dataset_size, num_fewshot, sample_idx):
...
+            encoded_example['preamble'] = self.tokenizer(preamble)
+            # if there is an EOS token added, we need to remove it so it is not in the middle of the prompt
```
mvpatel2000, line 179:
```suggestion
            # If there is an EOS token added, we need to remove it so it is not in the middle of the prompt
```

composer/metrics/nlp.py
```diff
@@ -287,6 +289,75 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
+            cleaned_sample_output = self.normalize_answer(sample_output)
+            cleaned_sample_labels = set(self.normalize_answer(label) for label in sample_labels)
+            if any(cleaned_sample_output.startswith(label) for label in cleaned_sample_labels):
+                self.correct += torch.tensor(1.0)
+            self.total += torch.tensor(1.0)
+
+    def compute(self):
+        assert isinstance(self.correct, Tensor)
+        assert isinstance(self.total, Tensor)
+        return self.correct / self.total
```
mvpatel2000, line 358: Can we add + epsilon (and make epsilon an argument) just to be safe? iirc other metrics implementations do this? 

I'm also fine erroring 🤷

composer/models/huggingface.py
```diff
@@ -403,6 +421,40 @@ def get_metadata(self):
...
+            max_new_tokens (int, optional): Maximum number of new tokens to generate. Defaults to 20.
+            **kwargs: Additional arguments passed to :meth:`transformers.GenerationMixin.generate` function.
+        """
+        # We need to call forward once in order for FSDP + generate to work
+        # See https://github.com/huggingface/accelerate/issues/570, https://github.com/huggingface/accelerate/issues/947,
+        # and https://github.com/pytorch/pytorch/issues/82461 for more info
+        if not self.dummy_forward_called:
+            with torch.no_grad():
+                self.model(input_ids=torch.tensor([[0]], dtype=torch.long, device=input_ids.device))
+            self.dummy_forward_called = True
```
mvpatel2000, line 448: "We have no plans to support this at the moment"  

F

I would recommend emitting a waring saying we are doing a dummy implementation, this will add some latency.

composer/models/huggingface.py
dakinggg: Yeah this is a fair question. I called out the really important ones explicitly for clarity...but there isn't any particular reason to do this. Open to putting everything in kwargs. Max length is extremely critical in order for this to work with fsdp, but the others are not.

composer/datasets/in_context_learning_evaluation.py
dakinggg: I'm pretty sure if the file doesn't exist it crashes, but will double check.

composer/datasets/in_context_learning_evaluation.py
dakinggg: 🤷‍♂️ not sure if it's a thing that is consistent across models, or if I was doing something else wrong. I should probably try to double check this before merging.

composer/models/huggingface.py
dakinggg: The latency is _extremely_ minor. This is one call to forward (and only once), on one token of input. Whereas generate is num_tokens calls to forward. Maybe worth adding a warning anyway. Also this reminded me that I need to check if this works for encoder decoders too.

composer/models/huggingface.py
samhavens: `num_beams` feels less important to me than `top_p`, since nucleus sampling is considered better lately. And if we expose `num_beams` I feel like we should include `no_repeat_ngram_size`.

I think I am in favor of explicitly exposing all the generation parameters. One benefit to this is that we can **check if the user sets `penalty_alpha` to a non-falsey value; if so I am pretty sure we need to error, since I don't think we support contrastive decoding** currently (we will need to conform our forward signature to huggingface standards; we need to support [output_hidden_states](https://github.com/huggingface/transformers/blob/v4.27.1/src/transformers/generation/utils.py#L1834) and [passing around the KV cache](https://github.com/huggingface/transformers/blob/v4.27.1/src/transformers/generation/utils.py#L1858-L1862)

dakinggg: There are kind of a lot of generate parameters... https://huggingface.co/docs/transformers/v4.27.1/en/main_classes/text_generation#transformers.GenerationConfig. I don't think _we_ (being `HuggingFaceModel`) need to error on any generation parameters, because the call to generate happens on the underlying HF model, and presumably HF errors if you do stuff that is not allowed. We (being `MosaicGPT`) may need some error handling for things that aren't supported yet in the `MosaicGPT` model. Also, if there are any args that we need to check here for some reason, we can just pop them off of. Ok, what if I just don't specially expose any args, and defer to the HF docs?

composer/datasets/in_context_learning_evaluation.py
dakinggg: Yeah, that is definitely the reason, I'm just not completely sure if the prompt _not_ ending in a space is a thing that all models prefer.

composer/datasets/in_context_learning_evaluation.py
dakinggg: confirmed that it just fails cleanly if the file does not exist.

dakinggg: adding the context you suggested is a good general idea, but going to skip for this pr

composer/metrics/nlp.py
dakinggg: This is what you get if you call compute before update (all of our metrics have this property)
```
In [2]: from composer.metrics import MaskedAccuracy

In [3]: acc = MaskedAccuracy()

In [4]: acc.compute()
/workdisk/danielking/miniconda3/envs/examples-llm/lib/python3.10/site-packages/torchmetrics/utilities/prints.py:36: UserWarning: The ``compute`` method of metric MaskedAccuracy was called before the ``update`` method which may lead to errors, as metric states have not yet been updated.
  warnings.warn(*args, **kwargs)
Out[4]: tensor(nan)
```

dakinggg: So I will leave it

composer/models/huggingface.py
dakinggg: fixed for t5. skipping the warning because this is so minor compared to the latency of generate.

composer/models/huggingface.py
dakinggg: ok i dropped all the explicit args and left a note in the docstring about where to find them in the HF docs.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -21,7 +21,7 @@
 __all__ = ['InContextLearningLMTaskDataset', 'InContextLearningMultipleChoiceTaskDataset', 'get_icl_task_dataloader']
 
 
-def _make_padded_input(context_enc, continuation_enc, max_seq_len, pad_tok_id):
+def _make_padded_input(context_enc, continuation_enc, max_seq_len, pad_tok_id, padding_side='right'):
```
dakinggg, line 24: padding side right does not work for decoder archs

dakinggg:
@mvpatel2000 requested rereview for the changes to setup.py and data_spec.py

bandish-shah (approved):
LGTM

PR #2047, opened 2023-03-08
update fsdp mixed precision

vchiley:
# What does this PR do?

Our MixedPrecision setting 'DEFAULT' should be updated to:
```
# If mixed_precision = 'default'
mixed_precision = MixedPrecision(
  param_dtype=autocast_precision,
  reduce_dtype=torch.float32,
  buffer_dtype=autocast_precision,
)
```
This will make 'DEFAULT' emulate AMP behavior.

use FULL if you need
use PURE if you want to live dangerously
standard AMP should be the 'DEFAULT' MixedPrecision setting (see above)

The current 'DEFAULT' shouldn't really be used.

Fixes https://mosaicml.atlassian.net/browse/CO-1896

# Before submitting
- [x] Was this change discussed/approved in an issue first? It is much more likely to be merged if so.

### DIFF
--- a/composer/trainer/mosaic_fsdp.py
+++ b/composer/trainer/mosaic_fsdp.py
@@ -84,8 +84,8 @@ def get_mixed_precision(precision, mixed_precision='DEFAULT', keep_low_precision
         if mixed_precision == 'FULL':
             pass
         elif mixed_precision == 'DEFAULT':
-            reduce_dtype = get_torch_dtype(precision)
-            buffer_dtype = torch.float32
+            param_dtype = get_torch_dtype(precision)
+            buffer_dtype = get_torch_dtype(precision)
         elif mixed_precision == 'PURE':
             param_dtype = get_torch_dtype(precision)
             reduce_dtype = get_torch_dtype(precision)
--- a/docs/source/notes/distributed_training.rst
+++ b/docs/source/notes/distributed_training.rst
@@ -202,7 +202,7 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as 
     )
     # If mixed_precision = 'default'; emulates automatic mixed precision training.
     mixed_precision = MixedPrecision(
-      param_dtype=autocast_precision,  # Master weigths stored in fp32 but are downcast to autocast_precision before the dist all_gather
+      param_dtype=autocast_precision,  # Master weights stored in fp32 but are downcast to autocast_precision before the dist all_gather
       reduce_dtype=torch.float32,  # gradient dist all_reduce in fp32
       buffer_dtype=autocast_precision,  # Buffers stored in fp32 but are downcast to autocast_precision before the dist all_gather
     )
--- a/docs/source/notes/distributed_training.rst
+++ b/docs/source/notes/distributed_training.rst
@@ -200,18 +200,17 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as 
       reduce_dtype=torch.float32,
       buffer_dtype=torch.float32,
     )
-    # If mixed_precision = 'default'
+    # If mixed_precision = 'default'; emulates automatic mixed precision training.
     mixed_precision = MixedPrecision(
-      param_dtype=torch.float32,
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=torch.float32,
+      param_dtype=autocast_precision,  # Master weigths stored in fp32 but are downcast to autocast_precision before the dist all_gather
+      reduce_dtype=torch.float32,  # gradient dist all_reduce in fp32
+      buffer_dtype=autocast_precision,  # Buffers stored in fp32 but are downcast to autocast_precision before the dist all_gather
     )
-
     # If mixed_precision = 'pure'
     mixed_precision = MixedPrecision(
-      param_dtype=autocast_precision, # Low precision master weights
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=autocast_precision, # Low precision buffers
+      param_dtype=autocast_precision,  # Master weigths stored in fp32 but are downcast to autocast_precision before the dist all_gather
+      reduce_dtype=autocast_precision,  # gradient dist all_reduce in autocast_precision
+      buffer_dtype=autocast_precision,  # Buffers stored in fp32 but are downcast to autocast_precision before the dist all_gather
     )
 
 An example code snippet for using FSDP with composer is provided below:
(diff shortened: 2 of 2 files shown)


docs/source/notes/distributed_training.rst
```diff
@@ -200,18 +200,17 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as
       reduce_dtype=torch.float32,
       buffer_dtype=torch.float32,
     )
-    # If mixed_precision = 'default'
+    # If mixed_precision = 'default'; emulates automatic mixed precision training.
     mixed_precision = MixedPrecision(
-      param_dtype=torch.float32,
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=torch.float32,
+      param_dtype=autocast_precision,  # Master weigths stored in fp32 but are downcast to autocast_precision before the dist all_gather
```
dakinggg, line 205:
```suggestion
      param_dtype=autocast_precision,  # Master weights stored in fp32 but are downcast to autocast_precision before the dist all_gather
```

dakinggg (commented):
LGTM, would like abhi to approve as well

docs/source/notes/distributed_training.rst
```diff
@@ -200,18 +200,17 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as
...
+      reduce_dtype=torch.float32,  # gradient dist all_reduce in fp32
+      buffer_dtype=autocast_precision,  # Buffers stored in fp32 but are downcast to autocast_precision before the dist all_gather
     )
-
     # If mixed_precision = 'pure'
     mixed_precision = MixedPrecision(
-      param_dtype=autocast_precision, # Low precision master weights
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=autocast_precision, # Low precision buffers
+      param_dtype=autocast_precision,  # Master weigths stored in fp32 but are downcast to autocast_precision before the dist all_gather
```
vchiley, line 211:
```suggestion
      param_dtype=autocast_precision,  # Master weights stored in fp32 but are downcast to autocast_precision before the dist all_gather
```

docs/source/notes/distributed_training.rst
```diff
@@ -200,18 +200,17 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as
...
       buffer_dtype=torch.float32,
     )
-    # If mixed_precision = 'default'
+    # If mixed_precision = 'default'; emulates automatic mixed precision training.
     mixed_precision = MixedPrecision(
-      param_dtype=torch.float32,
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=torch.float32,
+      param_dtype=autocast_precision,  # Master weights stored in fp32 but are downcast to autocast_precision before the dist all_gather
+      reduce_dtype=torch.float32,  # gradient dist all_reduce in fp32
```
mvpatel2000, line 206:
```suggestion
      reduce_dtype=torch.float32,  # Gradient dist all_reduce in fp32
```
Nit

docs/source/notes/distributed_training.rst
```diff
@@ -200,18 +200,17 @@ One Composer-specific pattern is that if :code:`mixed_precision` is provided as
...
+      buffer_dtype=autocast_precision,  # Buffers stored in fp32 but are downcast to autocast_precision before the dist all_gather
     )
-
     # If mixed_precision = 'pure'
     mixed_precision = MixedPrecision(
-      param_dtype=autocast_precision, # Low precision master weights
-      reduce_dtype=autocast_precision, # Low precision gradient communication
-      buffer_dtype=autocast_precision, # Low precision buffers
+      param_dtype=autocast_precision,  # Master weights stored in fp32 but are downcast to autocast_precision before the dist all_gather
+      reduce_dtype=autocast_precision,  # gradient dist all_reduce in autocast_precision
```
mvpatel2000, line 212:
```suggestion
      reduce_dtype=autocast_precision,  # Gradient dist all_reduce in autocast_precision
```
Nit

composer/trainer/mosaic_fsdp.py
```diff
@@ -84,8 +84,8 @@ def get_mixed_precision(precision, mixed_precision='DEFAULT', keep_low_precision
         if mixed_precision == 'FULL':
             pass
         elif mixed_precision == 'DEFAULT':
-            reduce_dtype = get_torch_dtype(precision)
-            buffer_dtype = torch.float32
+            param_dtype = get_torch_dtype(precision)
+            buffer_dtype = get_torch_dtype(precision)
```
mvpatel2000, line 88: Should we just make default PURE and get rid of DEFAULT? Because it seems like you should always use PURE?

mvpatel2000 (approved):

composer/trainer/mosaic_fsdp.py
vchiley: AMP is an industry standard and should be the `DEFAULT`.
PURE might work well with autocast_precision=bf16, but it might not work well with autocast_precision=fp16. These differences is not something we've tested so I wouldn't remove `DEFAULT`.

vchiley:
fix to test failing in https://github.com/mosaicml/composer/pull/2050

bcui19 (approved):
LGTM!

abhi-mosaic:
@vchiley can we get a side-by-side LLM run before and after this change? Aka can we do just a small 125M run with previous `DEFAULT` and new `DEFAULT` bheavior and post the screenshot here for posterity? Basically loss and MFU side by side

vchiley:
testing behavior change [here](https://wandb.ai/mosaic-ml/test_mp)

gpt125m-flash-mpf is using mp: `'FULL'`
gpt125m-flash-mpp is using mp: `'PURE'`
gpt125m-flash-mpd is using mp: `'DEFAULT'`

gpt125m-flash-mpp16b16 is using mp: `{'param_dtype': 'bf16', 'buffer_dtype': 'bf16'}` ie the proposed `'DEFAULT'`
([here](https://github.com/mosaicml/composer/blob/fbb4f339c2e2ee467e2e2ebe646a29a751b765df/composer/trainer/mosaic_fsdp.py#L76) reduce_dtype will default to `None` and use fp32).

At on 8 GPUs at MosaicGPT 125M to chinchilla point shows no diff between any of the runs. The speed is virtually the same for all of the configs, except 'FULL' is ~3% slower.
(the old DEFAULT is ~1% slower than the new DEFAULT)
(In this exact setup the new DEFAULT is ~0.2% faster than PURE??? this is either within measurement noise or the casting is more expensive than one would hope.

Screenshot 2023-03-09 at 12 21 00 PM


I tried running the 7B model with the same set of 4 diff mp configs on 32 GPUs to see performance:
Screenshot 2023-03-09 at 12 24 03 PM

It seems like PURE and the new DEFAULT run at the same speed and are faster than the old DEFAULT and PURE is slowest

PR #2050, opened 2023-03-08
Update FSDP meta weight tying tests to include precision testing

bcui19:
# What does this PR do?
This PR changes the weight-tying tests to test for all mixed-precision types. It also updates how we do the data checks for the weights. We need to do the data check in case both pointers point to NULL. We have to add `FSDP.summon_full_params(...)` because for lower precision weights the weights aren't instantiated yet.

https://github.com/mosaicml/composer/pull/2050

# Before submitting
- [ x] Did you run the tests locally to make sure they pass?
- [x ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))


vchiley (approved):
LGTM

nit: we could do:
`with trainer.state.model.module.summon_full_params(trainer.state.model.module):`
instead of needing to import fsdp
(if you want)

mvpatel2000 (approved):

PR #2051, opened 2023-03-09
Raise error if attempting to export FSDP model

hanlint:
# What does this PR do?

Our export for inference method requires a deepcopy in order to avoid side effects on the model, especially when exporting during training.

FSDP currently does not support deepcopies, so this PR throws an error to give a more informative error.


composer/utils/inference.py
```diff
@@ -142,6 +142,10 @@ def export_for_inference(
         raise ValueError(
             f'Directly exporting a DistributedDataParallel model is not supported. Export the module instead.')
 
+    if is_model_fsdp(model):
+        raise ValueError(
+            'Directly exporting a FSDP wrapped module is not supported. Load the model without FSDP wrapping.')
+
```
mvpatel2000, line 148:
```suggestion
    if is_model_fsdp(model):
        raise ValueError(
            'Directly exporting a FSDP wrapped module is not supported as the model is deepcopied to avoid '
            'side-effects, and FSDP does not support deepcopying. To export the model, load it without FSDP ' 
            'wrapping.')
```
Nit: We should explain why something is not supported -- otherwise, it's somewhat frustrating as a user.

mvpatel2000 (approved):

PR #2049, opened 2023-03-08
test and fix composer package name usage in composer_collect_env

dakinggg:
# What does this PR do?
We recently added the `composer` package in addition to `mosaicml` for the first time. We had one piece of code `get_composer_commit_hash` that relied on the `mosaicml` package name existing. This change makes the code robust to any package name (won't crash), and gives expected results for either `mosaicml` or `composer`.

Manually tested by doing `COMPOSER_PACKAGE_NAME=composer pip install .[all]` and verifying that `composer_collect_env` errors before this change, and does not error after this change. Also verified that the the tests suite fails when package is built as `composer`, so this new workflow would have caught this issue.

# What issue(s) does this change relate to?
Closes [CO-1901](https://mosaicml.atlassian.net/browse/CO-1901)
 https://github.com/mosaicml/composer/issues/2048 will be closed once re-released with this change

# Before submitting
- [x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [x] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [x] Did you update any related docs and document your change?
- [x] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [x] Did you run the tests locally to make sure they pass?
- [x] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

[CO-1901]: https://mosaicml.atlassian.net/browse/CO-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #2048, opened 2023-03-08
`pip install composer==0.13.1` leads to no metadata for composer env.


samuelstevens:
**Environment**

I can't run `composer_collect_env` because there is no package data found for `mosaicml`.
This causes bugs when checkpointing as well.

**To reproduce**

1. Make a new virtual env.
2. `pip install composer==0.13.1`
3. `composer_collect_env`

## Expected behavior

It works.

## Additional context

```
def get_composer_commit_hash() -> Optional[str]:
    # Use PEP-610 to get the commit hash                                                          
    # See https://packaging.python.org/en/latest/specifications/direct-url/
    files = importlib_metadata.files('mosaicml') # <= change this to 'composer'
    if files is None:
        return
    files = [f for f in files if str(f).endswith('direct_url.json')]
    if len(files) == 0:
        return
    f = files[0]
    direct_url = json.loads(f.read_text())
    vcs_info = direct_url.get('vcs_info', {})                                                     
    commit_id = vcs_info.get('commit_id')
    return commit_i
```

I can open a PR if you like.

dakinggg:
Hey @samuelstevens thanks for the report, and sorry about that! We'll get this fixed ASAP. In the meantime, you should still be able to use `pip install mosaicml`.

bandish-shah:
The `composer==0.13.1` PyPi package has been yanked, thanks for reporting this!

PR #2052, opened 2023-03-09
Add composer PyPI package tests to daily workflow

bandish-shah:
# What does this PR do?
 
Specifies the `composer_package_name` input for the `pytest-cpu.yaml` reusable workflow called by the `daily.yaml` workflow.  This field was added to add testing for the PyPi `composer` package as part of PR #2049, which broke the `daily.yaml` workflow.  It was not caught because the `daily.yaml` workflow doesn't get triggered until a merge to `dev`.  Since the workflow file is invalid, it doesn't run and we never saw a ❌ appear next to the status checks for each commit.

This change is tested by temporarily adding the `pull_request` event trigger to the workflow and making sure the workflow can run as part of a PR check.  
https://github.com/mosaicml/composer/actions/runs/4379724350

The `pull_request` trigger will be removed before merging.

Relates to [CO-1901](https://mosaicml.atlassian.net/browse/CO-1901)
Relates to #2048 

# Before submitting

[CO-1901]: https://mosaicml.atlassian.net/browse/CO-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -1,6 +1,5 @@
 name: Daily
 on:
-  pull_request:
   schedule:
     - cron: '30 2 * * *' # 2:30 every day
   push:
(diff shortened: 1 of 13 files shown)


.github/workflows/daily.yaml
```diff
@@ -1,5 +1,6 @@
 name: Daily
 on:
+  pull_request:
   schedule:
```
mvpatel2000, line 4:
```suggestion
on:
  schedule:
```

mvpatel2000 (approved):
Good catch!

Should we add a GPU job with `composer`?

bandish-shah:
> Should we add a GPU job with composer? We have bandwidth on test cluster for it rn, so it seems safe to do?
Just added!

PR #2053, opened 2023-03-10
Fix OCIObjectStore save_overwrite=False bug

eracah:
# What does this PR do?

## Background of the bug.
When saving a new object to `OCIObjectStore` with `save_overwrite=False`, the worker tasked with uploading the object will first check if the object is already in OCI first (by calling the `get_object` function of the oci sdk) before uploading it. The way it confirms that the object is not there is by catching a `FileNotFoundError`. The bug exists because the `OCIObjectStore` throws an `oci.exceptions.ServiceError` that goes uncaught instead of a .`FileNotFoundError`.

## Fix
We wrap all `oci.object_storage.ObjectStorageClient.get_object` calls in a try except block and reraise any `oci.exceptions.ServiceError` as `FileNotFoundErrors`. fixes CO-1874 and CO-1903

### Along for the ride
* Added the ability for RemoteUploaderDownloader workers to send their exception to the parent, so the user gets more informative errors instead of just: `'Upload worker crashed. Please check the logs.'` This fixes CO-1467

* Fixed a bug (CO-1811) where one of the workers has an error, but the last batch has already happened, so RemoteUploaderDownloader doesn't know that a worker has failed, but meanwhile there are still tasks in the queue, so the `wait_for_workers` function just runs forever. The fix is to check if the exceptions_queue is empty before entering the busy wait for workers to finish loop. Fixes CO-1811

### Manual Tests (local and remote with 2 gpu's)

- [x] Save with Non-existent bucket: **errors with proper error.**
- [x] Save with nonexistent object and save_overwrite=False: **works.**
- [x] Save with nonexistent object  with save_overwrite=True: **works.**
- [x] Save with existing object save_overwrite=False: **errors with proper error.**
- for short training runs on gpu's this triggers the bug from CO-1811 if this fix is not in place.
- will error with a very vague error due to CO-1467 if this fix is not in place.
- [x] Save with existing object save_overwrite=True: **works.**
- [x] Load with nonexistent object_name: **errors with proper error.**
- [x] Load with existent path: **works.**

### Autoresume Manual Tests
```
    trainer = Trainer(
        model=mnist_model(num_classes=10),
        max_duration="8ba", 
        save_interval="4ba",
        save_folder='oci://evan-test/auto_resume_test/{run_name}',
        save_overwrite=False,
        save_num_checkpoints_to_keep=0,
        log_to_console=True,
        console_log_interval='1ba',
        autoresume=True,
        run_name='evans-fun-run',   
    )
    trainer.fit()

    trainer = Trainer(
        model=mnist_model(num_classes=10),
        max_duration="20ba", 
        save_interval="4ba",
        save_folder='oci://evan-test/auto_resume_test/{run_name}',
        save_overwrite=False,
        save_num_checkpoints_to_keep=0,
        log_to_console=True,
        console_log_interval='1ba',
        autoresume=True,
        run_name='evans-fun-run',   
    )
    trainer.fit()
```
- [x] local autoresume (script above): **works**
- [x] remote 2 gpu autoresume (same script as above basically): **works**
   

### Manual Tests (training run)
- [x] Run examples/llm training and save checkpoints to OCI with save_overwrite: False: **works!!!**
- [x] Run examples/llm training and save checkpoints to OCI with save_overwrite: False pointed at the same folder: **errors as it should: `FileExistsError: Object oci://<foo>.pt already exists, but allow_overwrite was set to False.`

### Unit Tests
- [x] Unit test for RemoteUploaderDownloader exceptions queue
- [x] Unit tests for new exception handling in OCIObjectStore

fixes CO-1874
fixes CO-1467
fixes CO-1811
fixes CO-1903

### DIFF
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -463,7 +463,7 @@ class RemoteUploaderDownloader(LoggerDestination):
                 if len(self._logged_objects) == 0:
                     break
             time.sleep(0.2)  # Yield lock for enqueue thread
-        # Verify all tasks have been completed unless there are exceptions that have been thrown.
+        # Verify all tasks have been completed unless a worker threw an exception
         while not self._file_upload_queue.empty() and self._exception_queue.empty():
             time.sleep(0.2)
         if not self._exception_queue.empty():
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -457,8 +457,7 @@ class RemoteUploaderDownloader(LoggerDestination):
         post_close logic ensures existing uploads are completed, trying to schedule new uploads
         during this time will error.
         """
-        # Verify enqueue thread has processed all tasks
-        # But make sure to only enter this loop if a worker somewhere else hasn't thrown a fatal exception.
+        # Verify enqueue thread has processed all tasks unless a worker threw an exception
         while self._exception_queue.empty():
             with self._object_lock:
                 if len(self._logged_objects) == 0:
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -276,7 +276,8 @@ class RemoteUploaderDownloader(LoggerDestination):
                                            multiprocessing.JoinableQueue[Tuple[str, str,
                                                                                bool]],] = mp_ctx.JoinableQueue()
             self._completed_queue: Union[queue.Queue[str], multiprocessing.JoinableQueue[str],] = mp_ctx.JoinableQueue()
-            self._exception_queue: Union[queue.Queue[Exception], multiprocessing.JoinableQueue[Exception],] = mp_ctx.JoinableQueue()
+            self._exception_queue: Union[queue.Queue[Exception],
+                                         multiprocessing.JoinableQueue[Exception],] = mp_ctx.JoinableQueue()
             self._finished_cls: Union[Callable[[], multiprocessing._EventType], Type[threading.Event]] = mp_ctx.Event
             self._proc_class = mp_ctx.Process
         else:
@@ -602,7 +603,7 @@ def _upload_worker(
                     exception_queue.put_nowait(e)
                     raise e
             log.info('Uploading file %s to %s', file_path_to_upload, uri)
-            try: 
+            try:
                 remote_backend.upload_object(
                     object_name=remote_file_name,
                     filename=file_path_to_upload,
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -457,14 +457,18 @@ class RemoteUploaderDownloader(LoggerDestination):
         during this time will error.
         """
         # Verify enqueue thread has processed all tasks
-        while True:
+        # But make sure to only enter this loop if a worker somewhere else hasn't thrown a fatal exception.
+        while self._exception_queue.empty():
             with self._object_lock:
                 if len(self._logged_objects) == 0:
                     break
             time.sleep(0.2)  # Yield lock for enqueue thread
-        # Verify all tasks have been completed
-        while not self._file_upload_queue.empty():
+        # Verify all tasks have been completed unless there are exceptions that have been thrown.
+        while not self._file_upload_queue.empty() and self._exception_queue.empty():
             time.sleep(0.2)
+        if not self._exception_queue.empty():
+            e = self._exception_queue.get_nowait()
+            raise e
 
     def post_close(self):
         # Shutdown logic:
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -276,11 +276,13 @@ class RemoteUploaderDownloader(LoggerDestination):
                                            multiprocessing.JoinableQueue[Tuple[str, str,
                                                                                bool]],] = mp_ctx.JoinableQueue()
             self._completed_queue: Union[queue.Queue[str], multiprocessing.JoinableQueue[str],] = mp_ctx.JoinableQueue()
+            self._exception_queue: Union[queue.Queue[Exception], multiprocessing.JoinableQueue[Exception],] = mp_ctx.JoinableQueue()
             self._finished_cls: Union[Callable[[], multiprocessing._EventType], Type[threading.Event]] = mp_ctx.Event
             self._proc_class = mp_ctx.Process
         else:
             self._file_upload_queue = queue.Queue()
             self._completed_queue = queue.Queue()
+            self._exception_queue = queue.Queue()
             self._finished_cls = threading.Event
             self._proc_class = threading.Thread
         self._worker_flag: Optional[Union[multiprocessing._EventType, threading.Event]] = None
@@ -322,6 +324,7 @@ class RemoteUploaderDownloader(LoggerDestination):
                     'backend_kwargs': self.backend_kwargs,
                     'num_attempts': self.num_attempts,
                     'completed_queue': self._completed_queue,
+                    'exception_queue': self._exception_queue,
                 },
                 # The worker threads are joined in the shutdown procedure, so it is OK to set the daemon status
                 # Setting daemon status prevents the process from hanging if close was never called (e.g. in doctests)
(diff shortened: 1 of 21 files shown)


composer/utils/object_store/oci_object_store.py
```diff
@@ -16,6 +16,26 @@
 __all__ = ['OCIObjectStore']
 
 
+def _reraise_oci_errors(uri: str, e: Exception):
+    import oci
+    # If it's an oci service error with code: ObjectNotFound or status 404
+    if isinstance(e, oci.exceptions.ServiceError):
+        if e.status == 404: # type: ignore
+            if e.code == 'ObjectNotFound':  # type: ignore
+                raise FileNotFoundError(f'Object {uri} not found. {e.message}') from e
```
mvpatel2000, line 25:
```suggestion
            if e.code == 'ObjectNotFound':  # type: ignore
                raise FileNotFoundError(f'Object {uri} not found.') from e
```
The message should be shown because you're doing from e

composer/utils/object_store/oci_object_store.py
```diff
@@ -16,6 +16,26 @@
...
+            if e.code == 'BucketNotFound':
+                raise ValueError(f'Bucket specified in {uri} not found. {e.message}') from e
```
mvpatel2000, line 27:
```suggestion
            if e.code == 'BucketNotFound':
                raise ValueError(f'Bucket specified in {uri} not found.') from e
```
as above

composer/utils/object_store/oci_object_store.py
```diff
@@ -16,6 +16,26 @@
 __all__ = ['OCIObjectStore']
 
 
+def _reraise_oci_errors(uri: str, e: Exception):
+    import oci
```
mvpatel2000, line 20: does this need import gating

composer/utils/object_store/oci_object_store.py
```diff
@@ -42,29 +62,34 @@ def __init__(
...
+        try:
+            response = self.client.get_object(
+                namespace_name=self.namespace,
+                bucket_name=self.bucket,
+                object_name=object_name,
+            )
+        except Exception as e:
+            _reraise_oci_errors(self.get_uri(object_name), e)
+
+        return int(response.data.headers['Content-Length'])
```
mvpatel2000, line 92: missing response status 200 check? is that no longer necessary

composer/loggers/remote_uploader_downloader.py
```diff
@@ -450,14 +457,18 @@ def wait_for_workers(self):
         during this time will error.
         """
         # Verify enqueue thread has processed all tasks
-        while True:
+        # But make sure to only enter this loop if a worker somewhere else hasn't thrown a fatal exception.
```
mvpatel2000, line 460: Suggestion: # Verify enqueue thread has processed all tasks unless a worker threw an exception

composer/loggers/remote_uploader_downloader.py
```diff
@@ -450,14 +457,18 @@ def wait_for_workers(self):
...
+        while self._exception_queue.empty():
             with self._object_lock:
                 if len(self._logged_objects) == 0:
                     break
             time.sleep(0.2)  # Yield lock for enqueue thread
-        # Verify all tasks have been completed
-        while not self._file_upload_queue.empty():
+        # Verify all tasks have been completed unless there are exceptions that have been thrown.
```
mvpatel2000, line 466:
```suggestion
        # Verify all tasks have been completed unless a worker threw an exception
```

mvpatel2000 (commented):
Can we get a unit test for the exception stuff?

composer/utils/object_store/oci_object_store.py
eracah: Actually, for oci.exceptions.service_errors it's a lot prettier to extract the message because if you raise the error it prints a lot of other stuff. See [this code](https://github.com/oracle/oci-python-sdk/blob/master/src/oci/exceptions.py#L64). They actually pass the error_details to the Exception constructor not the message, so it's a lot more verbose if you raise the error without printing the message first

composer/utils/object_store/oci_object_store.py
eracah: same as above

composer/utils/object_store/oci_object_store.py
eracah: no because it's only called in functions called after the `__init__` is called. But I could add it anyway

composer/utils/object_store/oci_object_store.py
eracah: I could add it back in

dakinggg (commented):
Could you please also test autoresume? Both the initial run (with autoresume=True), and the subsequent resumption.

composer/loggers/remote_uploader_downloader.py
eracah: fixed

composer/utils/object_store/oci_object_store.py
eracah: done

composer/utils/object_store/oci_object_store.py
eracah: added it

eracah:
> Can we get a unit test for the exception stuff?

Done

eracah:
> Could you please also test autoresume? Both the initial run (with autoresume=True), and the subsequent resumption.

Done!

eracah:
Ok @mvpatel2000, @dakinggg, I addressed the comments and added unit tests for:
* RemoteUploaderDownloader exception queue
* OCIObjectStore

and manual tests for:
* autoresume local and remote
* LLM run with checkpoints existing

please give it another look-see!

tests/loggers/test_remote_uploader_downloader.py
```diff
@@ -287,3 +293,45 @@ def test_valid_backend_names():
         remote_uploader_downloader = RemoteUploaderDownloader(bucket_uri='magicloud://not-a-real-bucket')
         # Access the remote_backend property so that it is built
         _ = remote_uploader_downloader.remote_backend
+
+
+@pytest.mark.filterwarnings(r'ignore:Exception in thread:pytest.PytestUnhandledThreadExceptionWarning')
```
mvpatel2000, line 298: is this necessary

mvpatel2000 (approved):

tests/loggers/test_remote_uploader_downloader.py
```diff
@@ -119,15 +119,21 @@ def object_store_test_helper(
...
+                    # Then, crashes are detected on the next batch end, but
+                    # should be a FileExistsError not a runtime error because the parent will raise
+                    # the fatal exception that the worker throws.
+                    with pytest.raises(
+                            FileExistsError,
+                            match=
+                            f'Object local://{remote_file_name} already exists, but allow_overwrite was set to False.'):
                         remote_uploader_downloader.run_event(Event.BATCH_END, dummy_state, logger)
 
-                    with pytest.raises(RuntimeError):
```
dakinggg, line 126: this is no longer an error?

tests/utils/object_store/test_oci_object_store.py
```diff
@@ -34,59 +41,135 @@ def test_oci_obj_store(mock_bucket_name, monkeypatch):
     return oci_os
 
 
-def test_upload_object(test_oci_obj_store, monkeypatch, tmp_path, mock_bucket_name):
-    pytest.importorskip('oci')
+@pytest.mark.skipif(not _OCI_INSTALLED, reason='You must have oci installed to run this test')
```
dakinggg, line 44: we usually do this by just `oci = pytest.importorskip(oci)` in the test itself

dakinggg (approved):

tests/loggers/test_remote_uploader_downloader.py
eracah: yes. I added a comment explaining

tests/loggers/test_remote_uploader_downloader.py
eracah: it is, but the BATCH_END call would normally exit the process. It's only a technicality that pytest catches it and technically we could run another event for EPOCH_END. I updated it to separately test BATCH_END and EPOCH_END with a parametrization

tests/utils/object_store/test_oci_object_store.py
eracah: ok switched it back

PR #2054, opened 2023-03-10
Busy wait for local rank 0 download to avoid timeout on large file download

dakinggg:
# What does this PR do?
This PR adds a busy wait before the blocking all gather during autoresumption. This prevents a NCCL timeout when downloading the file on rank 0 takes a long time. Leaving refactoring this code [CO-1891](https://mosaicml.atlassian.net/browse/CO-1891) to another PR, to do one thing at a time.

Manual testing:
- Tried resuming the 7b run that originally hit this issue
- Successfully resumed on r12z1 (made it past the issue point, to the start of eval), two nodes
- Successfully started a fresh 7b run with autoresume=True on two nodes
- Failed to resume on r1z1 (CPU OOM - [CO-1909](https://mosaicml.atlassian.net/browse/CO-1909))

Closes [CO-1880](https://mosaicml.atlassian.net/browse/CO-1880)

# Before submitting

[CO-1891]: https://mosaicml.atlassian.net/browse/CO-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CO-1909]: https://mosaicml.atlassian.net/browse/CO-1909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CO-1909]: https://mosaicml.atlassian.net/browse/CO-1909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CO-1880]: https://mosaicml.atlassian.net/browse/CO-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/core/state.py
```diff
@@ -136,6 +136,8 @@ def get_fsdp_sharded_optim_state_dict(full_optim_state_dict: Dict[str, Any], mod
     if version.parse(torch.__version__) < version.parse('1.13.0'):
         raise RuntimeError('To use FSDP with Composer, you must use torch>=1.13.0.')
     from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+    log.debug(
```
eracah, line 139: do you mean to keep all these log.debug statements?

composer/trainer/trainer.py
```diff
@@ -1488,23 +1484,36 @@ def _get_autoresume_checkpoint(
...
+            # At this point the rank 0 filepath should exist on all ranks if the download succeeded
+            # list of whether the checkpoint exists on each rank
+            latest_checkpoint_exists = dist.all_gather_object(os.path.exists(latest_checkpoint_path))
 
             log.debug(
                 f'Checkpoint {latest_checkpoint_path} exists on rank {dist.get_global_rank()}? {os.path.exists(latest_checkpoint_path)}'
             )
 
-            if int(latest_checkpoint_exists_on_all_ranks.item()) == 0:
+            if not latest_checkpoint_exists[0]:
```
eracah, line 1513: so if `self._try_checkpoint_download` fails then rank 0's latest checkpoint path will be None then?

composer/trainer/trainer.py
```diff
@@ -1488,23 +1484,36 @@ def _get_autoresume_checkpoint(
...
+                # If the checkpoint doesn't exist on rank 0, don't crash, so the initial autoresume run can succeed
+                return None
+            elif not all(latest_checkpoint_exists):
```
eracah, line 1516: this is for local rank 0 successfully downloading, but rank 0 on another node failing?

composer/trainer/trainer.py
```diff
@@ -1488,23 +1484,36 @@ def _get_autoresume_checkpoint(
...
+            signal_file_path = os.path.join(os.path.dirname(latest_checkpoint_path),
+                                            '.local_rank0_completed_autoresume')
+            if dist.get_local_rank() == 0:
+                with open(signal_file_path, 'wb') as f:
+                    f.write(b'local_rank0_completed_autoresume')
+
+            # avoid the collective call until the local rank zero has finished trying to download the checkpoint
+            # so that we don't timeout for large downloads
+            with dist.local_rank_zero_download_and_wait(signal_file_path):
+                dist.barrier()
```
eracah, line 1503: ok so this is similar to the solution for normal  non-autoresume, so if there is a really large variance in load times between nodes then we still might timeout?

eracah (approved):
LGTM. I just had some questions for my own edification. Really well done. The context manager `dist.local_rank_zero_download_and_wait` is really clever.

composer/core/state.py
mvpatel2000: Let's keep them imo

composer/trainer/trainer.py
mvpatel2000, line 1503: The order here is a bit confusing because there's a context and a barrier. Can we split them out and make it more explicit what happens in what order?

composer/core/state.py
dakinggg: yeah, i try to differentiate between prints (added as one offs and dont intend to keep) and `log.debug` (added both because they are useful for me, and potentially useful for someone else in the future)

composer/trainer/trainer.py
dakinggg: that is correct. the old code tried to download on global rank 0 first, and used that is the indicator of whether autoresumption should happen, and then tried to download on all local rank 0s. this is to replicate that behavior.

composer/trainer/trainer.py
dakinggg: so you mean
```
with dist.local_rank_zero_download_and_wait(signal_file_path):
    pass
dist.barrier()
```
?

mvpatel2000: Yea maybe? I think that might help it be more explicit?

dakinggg: given that I can't test multinode again right now, id like to merge it as is. i still have a jira to clean this section up after your refactor gets in. is that ok?

mvpatel2000: sure, lets go! 🚀

mvpatel2000 (approved):

composer/trainer/trainer.py
dakinggg: `latest_checkpoint_path` won't but `None`, but there won't be any file at that path (`latest_checkpoint_exists = dist.all_gather_object(os.path.exists(latest_checkpoint_path))`)

dakinggg: and that path is just the path on rank 0 (i.e. this code does _not_ handle sharded checkpoints yet)
```
            # broadcast the local checkpoint path to all ranks
            latest_checkpoint_path_list = [os.path.abspath(latest_checkpoint_path)]
            dist.broadcast_object_list(latest_checkpoint_path_list, src=0)
            latest_checkpoint_path = latest_checkpoint_path_list[0]
            ```

PR #2057, opened 2023-03-10
Add flag to Composer Launcher that will kill all python processes at the end of the launcher

bcui19:
# What does this PR do?
Adds a flag to the composer launcher, which will run `pkill python` at the end of the launcher. This will kill off all python zombie processes at the end.

Running a test in MCloud we get:
```
composer-kill-zombies-CtNmn6  r1z1     a100_80gb  8        2023-03-10 04:19 PM  2023-03-10 04:20 PM  2023-03-10 04:22 PM  Failed (Exit code: 1. Reason: Error)
```

and here are the logs 
image

showing it exited cleanly.

# Before submitting

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -83,8 +83,10 @@ def _get_parser():
     parser.add_argument('-k',
                         '--kill_python_zombies',
                         action='store_true',
-                        help=('If set, at the end of composer, it will run ``kill python``'
-                              'which will kill all python processes.'))
+                        help=('If set, at the end of composer, it will run ``kill python`` '
+                              'which will kill all python processes. This can help in the case '
+                              'of NCCL timeouts where child processes become zombies and are '
+                              'not cleaned up.'))
 
     multinode_args = parser.add_argument_group(
         'multi-node arguments',
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -80,6 +80,10 @@ def _get_parser():
         help=('If set, run the training script as a command (i.e. without `python`). '
               'Cannot be used in conjunction with `module_mode`.'),
     )
+    parser.add_argument('-k',
+                        '--kill_python_zombies',
+                        action='store_true',
+                        help=('If set, at the end of composer, it will run '))
 
     multinode_args = parser.add_argument_group(
         'multi-node arguments',
@@ -494,11 +498,12 @@ def main():
         # what failed. No need to re-raise the exception, as `aggregate_process_returncode`
         # will return an appropriate error code, which will cause the script to exit.
         traceback.print_exc()
-        print('Killing training processes')
     finally:
         _cleanup_processes(processes)
         log_tmpdir.cleanup()
-        return _aggregate_process_returncode(processes)
+        return_code = _aggregate_process_returncode(processes)
+        os.system('pkill python')
+        return return_code
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


dskhudia:
This will happen at the end of training or when we encounter other issues. Is there any side effect of this? Once we reach this point, training is dead. right? No auto resumption etc.

composer/cli/launcher.py
```diff
@@ -80,6 +80,13 @@ def _get_parser():
         help=('If set, run the training script as a command (i.e. without `python`). '
               'Cannot be used in conjunction with `module_mode`.'),
     )
+    parser.add_argument('-k',
+                        '--kill_python_zombies',
+                        action='store_true',
+                        help=('If set, at the end of composer, it will run ``kill python`` '
+                              'which will kill all python processes. This can help in the case '
+                              'of NCCL timeouts where child processes become zombies and are '
+                              'not cleaned up.'))
```
mvpatel2000, line 89: Let's discuss offline... is it possible to find all children processes and kill only those? then we can turn on by default

bcui19: When I forced a NCCL timeout (and had extra child processes), this was the output:

```
ERROR:composer.cli.launcher:Global rank 0 (PID 66549) exited with code -15
before child processes
after child processes
```

if we run this:

```
if args.kill_python_zombies:
    current_process = psutil.Process()
    children = current_process.children(recursive=True)
    print ("before child processes")
    for child in children:
        print('Child pid is {}'.format(child.pid))
    print ("after child processes")
```

so it feels as if we're unable to actually get children when there is a NCCL timeout. At least in this naive manner.

bcui19:
@dskhudia Yeah this is at clean up time, e.g. we had a NCCL timeout, which leaves zombie child processes, which is likely undesirable

mvpatel2000 (approved):

composer/cli/launcher.py
```diff
@@ -494,11 +501,13 @@ def main():
         # what failed. No need to re-raise the exception, as `aggregate_process_returncode`
         # will return an appropriate error code, which will cause the script to exit.
         traceback.print_exc()
-        print('Killing training processes')
     finally:
         _cleanup_processes(processes)
         log_tmpdir.cleanup()
-        return _aggregate_process_returncode(processes)
+        return_code = _aggregate_process_returncode(processes)
+        if args.kill_python_zombies:
```
dakinggg, line 508: It looks like, as written, this happens whether `composer` finished successfully or not. Should we only do it if there is an error? I'm assuming we can tell something went wrong from the `return_code`

composer/cli/launcher.py
```diff
@@ -80,6 +80,13 @@ def _get_parser():
         help=('If set, run the training script as a command (i.e. without `python`). '
               'Cannot be used in conjunction with `module_mode`.'),
     )
+    parser.add_argument('-k',
```
dakinggg, line 83: could we get some more DANGER ZONE docstring? since this will kill _all_ python processes. and maybe log a warning about it if it is set.

composer/cli/launcher.py
bandish-shah: FWIW even though this is optional I think it's a bit too extreme to live in the launcher.  This might make sense in the isolated context of the Mosaic Platform but I would be so annoyed if I enabled this locally and it killed all python processes on my system.

If we had a way of building the process tree under the `composer` root process and cleaning up or if we explicitly searched for processes without a parent (zombies) and only `pkill`'d those then this would be ok.

bandish-shah (changes requested):
Let's consider lighter handed alternatives first, a syscall to `pkill python` is a bit too extreme even if it's optional.

mvpatel2000:
@bcui19 are we closing this PR? or is it still WIP

bcui19:
From the above discussion, it seems too heavy handed to have a flag that runs `pkill python` in composer. As a result, closing this PR.

PR #2058, opened 2023-03-10
Update docs with non-rank zero logs instructions

hanlint:
# What does this PR do?

Answers the common question about how to access the stdout and stderr of non-rank zero processes.


bandish-shah (approved):
👍

PR #2066, opened 2023-03-13
Activation monitor

bcui19:
# What does this PR do?
Adds a callback that monitors activations in the network. Every `interval` batches it will attach a forwards hook and logs the `max`, `average`, `l2 norm`, and `kurtosis` for the input and output activations.

the output in wandb looks sth like:
image

Also as a sanity check the average kurtosis is ~3 at initialization:
image

And it doesn't impact throughput that heavily:
image

In particular every `interval` batches it looks like we take maybe a ~5% throughput hit. However, making `interval` say 25ba or 50ba should alleviate the issues.

[CO-1915](https://mosaicml.atlassian.net/browse/CO-1915)

# Before submitting
- [x ] Did you update any related docs and document your change?
- [ x] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [ x] Did you run the tests locally to make sure they pass?
- [ x] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

[CO-1915]: https://mosaicml.atlassian.net/browse/CO-1915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/callbacks/activation_monitor.py
+++ b/composer/callbacks/activation_monitor.py
@@ -199,8 +199,8 @@ class ActivationMonitor(Callback):
     def compute_kurtosis(self, value: torch.Tensor):
         mean = value.mean()
         diffs = value - mean
-        m_4 = torch.pow(diffs, 4).sum()
-        var = torch.pow(diffs, 2).sum()
+        m_4 = torch.pow(diffs, 4).mean()
+        var = torch.pow(diffs, 2).mean()
         return m_4 / (var**2)
 
     def create_module_names(self, model: torch.nn.Module):
--- a/composer/callbacks/activation_monitor.py
+++ b/composer/callbacks/activation_monitor.py
@@ -38,41 +38,42 @@ class ActivationMonitor(Callback):
             ...     callbacks=[ActivationMonitor()],
             ... )
 
-    The metrics are logged by the :class:`.Logger` to the following keys described below. For convenience we have included
-    example metrics logged below:
+    The metrics are logged by the :class:`.Logger` to the following keys described below. Over an input of shape
+    (batch, hid_dim), (batch, seq_len, hid_dim), etc. we compute statistics across `hid_dim` then take the average
+    of these statistics. For convenience we have included example metrics logged:
 
         +-------------------------------------------------------+-----------------------------------------------------+
         | Key                                                   | Logged data                                         |
         +=======================================================+=====================================================+
-        |                                                       | The average max value of the last dimension of the  |
+        |                                                       | The average max value of the `hid_dim` of the       |
         | ``activations/max/MODULE_NAME/input_{n}``             | nth input activations into the current module.      |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
-        |                                                       | The average value of the nth input activations      |
-        | ``activations/average/MODULE_NAME/input_{n}``         | into the current module.                            |
+        |                                                       | The average value of the `hid_dim` input            |
+        | ``activations/average/MODULE_NAME/input_{n}``         | activations into the current module.                |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
-        |                                                       | The average L2 Norm of the last dimension of the    |
+        |                                                       | The average L2 Norm of the `hid_dim` of the         |
         | ``activations/l2_norm/MODULE_NAME/input_{n}``         | nth input activations into the current module.      |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
-        |                                                       | The kurtosis of the last dimension of the nth input |
-        | ``activations/kurtosis/MODULE_NAME/input_{n}``        | activations into the current module.                |
+        |                                                       | The average kurtosis of the `hid_dim` of the nth    |
+        | ``activations/kurtosis/MODULE_NAME/input_{n}``        | input activations into the current module.          |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
-        |                                                       | The average max value of the last dimension of the  |
+        |                                                       | The average max value of the `hid_dim` of the       |
         | ``activations/max/MODULE_NAME/output_{n}``            | nth ouput activations of the current module.        |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
-        |                                                       | The average value of the nth output activations of  |
-        | ``activations/average/MODULE_NAME/output_{n}``        | the current module.                                 |
+        |                                                       | The average value of the `hid_dim` of the output    |
+        | ``activations/average/MODULE_NAME/output_{n}``        | activations of the current module.                  |
         |                                                       |                                                     |
         +-------------------------------------------------------+-----------------------------------------------------+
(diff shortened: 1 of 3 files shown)


mvpatel2000 (commented):
Can we please include a warning in the docstring indicating this has a throughput hit, which requires there to be a log interval independent of the composer log interval? We should be very explicit that this has performance implications and thats why we're throttling it.

I'll do a more detailed pass after we have sign off from research that this is important to include

bmosaicml:
I'm finding this plot of activations to be really really surprising....I would expect average activations to be close to 0...

@vchiley isn't that what you'd expect as well?

composer/callbacks/activation_monitor.py
```diff
@@ -0,0 +1,216 @@
...
+        | ``activations/average/MODULE_NAME/input_{n}``         | into the current module.                            |
+        |                                                       |                                                     |
+        +-------------------------------------------------------+-----------------------------------------------------+
+        |                                                       | The average L2 Norm of the last dimension of the    |
+        | ``activations/l2_norm/MODULE_NAME/input_{n}``         | nth input activations into the current module.      |
+        |                                                       |                                                     |
+        +-------------------------------------------------------+-----------------------------------------------------+
+        |                                                       | The kurtosis of the last dimension of the nth input |
+        | ``activations/kurtosis/MODULE_NAME/input_{n}``        | activations into the current module.                |
```
bmosaicml, line 60: I think we want the kurtosis measured across the model dimension, so I'm a bit confused about your wording here. Can you clarify by showing for a tensor of size (B, seq_len, d_model) what you are calculating?

For instance, I think we should first average over B and seq_len to get vectors of size d_model, then compute our statistics over those averaged d_model sized vectors. Let me know if that agrees with your intuition and lets find a way to be extra careful that that's what we are actually doing.

composer/callbacks/activation_monitor.py
```diff
@@ -0,0 +1,216 @@
...
+        for handle in self.handles:
+            handle.remove()
+        # Resetting handles we track
+        self.handles = []
+
+    def register_forward_hook(self, model: torch.nn.Module, logger: Logger, step: Optional[int]):
+        model.apply(partial(self._register_forward_hook, logger, step))
+
+    def _register_forward_hook(self, logger: Logger, step: Optional[int], module: torch.nn.Module):
+        self.handles.append(module.register_forward_hook(partial(self.forward_hook, logger, step)))
```
bmosaicml, line 151: I've never seen this forward hook thing in action before...very cool!

composer/callbacks/activation_monitor.py
```diff
@@ -0,0 +1,216 @@
...
+            # Because we call max with `dim=-1` we need to call .values to get the actual values
+            metrics[f'activations/max/{name}{suffix}'] = value.max(dim=-1).values.mean().item()
+
+    def compute_kurtosis(self, value: torch.Tensor):
+        # Computes the kurtosis over the last dimension
+        mean = torch.mean(value, dim=-1).unsqueeze(-1)
+        diffs = value - mean
+        m_4 = torch.mean(torch.pow(diffs, 4), dim=-1)
+        var = torch.mean(torch.pow(diffs, 2), dim=-1)
+        return (m_4 / (var**2)).mean()
```
bmosaicml, line 213: I think you _may_ not need this last `mean` call...


It may be nice to either test these calculations against this https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kurtosis.html

Or just use scipy directly.

tests/callbacks/test_activation_monitor.py
```diff
@@ -0,0 +1,41 @@
...
+
+from composer.callbacks import ActivationMonitor
+from composer.loggers import InMemoryLogger
+from composer.optim import DecoupledAdamW
+from composer.trainer import Trainer
+from tests.common.datasets import RandomClassificationDataset
+from tests.common.models import SimpleModel
+
+
+def test_activation_monitor():
```
bmosaicml, line 15: I'd like to test some of your helper functions as well, such as compute_kurtosis and add_metrics

bmosaicml (changes requested):
Overall I think it's really good but I want to make sure we a very precise about what it is we are trying calculating and be very careful to make sure we are in fact calculating that

composer/callbacks/activation_monitor.py
bmosaicml: Actually I'm not even sure.. I believe the below graph suggests we should compute statistics over the dimension of d_model, then average over B and seq_len? Let me know if that makes sense to you. At any rate we should have one number per layer.

image

bcui19: Yeah I'm doing statistics over d_model. Happy to clarify

bcui19:
re: mean activations @bmosaicml , maybe I changed sth, but here's the new wandb:

https://wandb.ai/mosaic-ml/examples_fork-examples_llm/runs/11cflizo?workspace=user-bcui

and it looks much more reasonable

composer/callbacks/activation_monitor.py
bcui19: I think because of how we're computing statistics, for a batch of size (batch_size, seq_len, d_model) basically the `(m_4 / (var**2))` ends up being (batch_size, seq_len) so we need the reduction.

tests/callbacks/test_activation_monitor.py
bcui19: What tests would you like to see for `add_metrics`? Since what it fundamentally does is compute statistics over `d_model` with torch functions

bmosaicml:
Approving for now with the understanding the we should modify this later to restrict it only to the relevant layers:

https://mosaicml.atlassian.net/browse/CO-1964

bmosaicml (approved):

composer/callbacks/activation_monitor.py
```diff
@@ -0,0 +1,228 @@
...
+        elif isinstance(interval, str):
+            self.interval = Time.from_timestring(interval)
+        elif isinstance(interval, Time):
+            self.interval = interval
+
+        if self.interval.unit == TimeUnit.BATCH and self.interval < Time.from_timestring('10ba'):
+            warnings.warn(f'Currently the ActivationMonitor`s interval is set to {self.interval} '
+                          f'this is below our recommended value of 10ba. We recommend you raise '
+                          f'the interval to at least 10ba, as the activation monitor adds extra overhead '
+                          f'and is currently adversely impacting throughput performance.')
```
mvpatel2000, line 112:
```suggestion
        if self.interval.unit == TimeUnit.BATCH and self.interval < Time.from_timestring('10ba'):
            warnings.warn(f'Currently the ActivationMonitor`s interval is set to {self.interval} '
                          f'which is below our recommended value of 10ba. We recommend you raise '
                          f'the interval to at least 10ba, as the activation monitor adds extra overhead '
                          f'and decreases throughput.')
```

mvpatel2000 (approved):
LGTM.

PR #2067, opened 2023-03-14
Add `device` and `dtype` back to `LPLayerNorm`

abhi-mosaic:
I think the init signature got modified somewhere in the previous PR #2044. 
Reverting here, thanks @growlix for catching!

### DIFF
--- a/composer/algorithms/low_precision_groupnorm/low_precision_groupnorm.py
+++ b/composer/algorithms/low_precision_groupnorm/low_precision_groupnorm.py
@@ -112,11 +112,11 @@ def _to_LPGroupNorm(layer: torch.nn.Module, module_index: int) -> LPGroupNorm:
 
     with torch.no_grad():
         if layer.weight is None:
-            lp_groupnorm.weight = None  # type: ignore
+            lp_groupnorm.register_parameter('weight', None)
         else:
             lp_groupnorm.weight.copy_(layer.weight)  # type: ignore
         if layer.bias is None:
-            lp_groupnorm.bias = None  # type: ignore
+            lp_groupnorm.register_parameter('bias', None)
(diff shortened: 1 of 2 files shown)


mvpatel2000 (commented):
Can you please also fix LowPrecisionGroupNorm

mvpatel2000 (approved):
Please fix LPGN case before merging

vchiley (approved):

composer/algorithms/low_precision_groupnorm/low_precision_groupnorm.py
```diff
@@ -113,6 +105,19 @@ def _cast_if_autocast_enabled(tensor):
...
 def _to_LPGroupNorm(layer: torch.nn.Module, module_index: int) -> LPGroupNorm:
+    """Defines a replacement policy from a `torch.nn.GroupNorm` to a `LPGroupNorm`"""
     if not isinstance(layer, torch.nn.GroupNorm):
         raise TypeError(f'Expected torch.nn.GroupNorm, got {type(layer)}')
-    return LPGroupNorm(layer)
+    lp_groupnorm = LPGroupNorm(layer.num_groups, layer.num_channels, layer.eps, layer.affine)
+
+    with torch.no_grad():
+        if layer.weight is None:
+            lp_groupnorm.weight = None  # type: ignore
```
vchiley, line 115: Should this be:
```suggestion
            lp_groupnorm.register_parameter('weight', None )  # type: ignore
```
Not sure if ` # type: ignore` is needed in this case.
see [eg here](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/linear.py#L100)

same for bias, and layernorm weight and bias.

PR #2069, opened 2023-03-15
Repoint signal file in download checkpoint

mvpatel2000:
# What does this PR do?

We currently use a local file as a file lock for local synchronization. Sometimes, this folder may not exist, e.g. when loading from a remote object storage. Instead, we should be adding the file next to our local temp dir.

Additionally, we are missing downloading global rank 0 on non-rank 0 nodes. This ensures all ranks download it.


dakinggg (approved):
LGTM

eracah (approved):
LGTM

dakinggg (changes requested):
blocking for the moment

PR #2078, opened 2023-03-18
make composer fsdp work with latest torch

dskhudia:
# What does this PR do?

torch 2.0 has some fsdp related name changes for params and some imports are different. We conditionally make composer work with torch 2.0. This doesn't break composer for pre 2.0 torch version. 

Support for torch 2.0 with fsdp

### DIFF
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -134,13 +134,17 @@ def prepare_fsdp_module(model: torch.nn.Module, optimizers: Optional[Union[torch
         fsdp_config (Dict[str, Any]): The FSDP config.
         precision: (Precision): The precision being used by the Trainer, used to fill in defaults for FSDP `mixed_precision` settings.
     """
+    is_torch_2_0 = False
+    if version.parse(torch.__version__) >= version.parse('2.0.0'):
+        is_torch_2_0 = True
     if version.parse(torch.__version__) < version.parse('1.13.0'):
         raise RuntimeError('To use FSDP with Composer, you must use torch>=1.13.0.')
     from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (CheckpointImpl,
                                                                              apply_activation_checkpointing,
                                                                              checkpoint_wrapper)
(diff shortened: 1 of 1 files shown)


composer/trainer/dist_strategy.py
```diff
@@ -275,18 +279,34 @@ def _param_init_fn(module: torch.nn.Module) -> None:
...
                     else:
-                        return is_large
+                        if hasattr(module, '_fsdp_wrap'):
+                            return bool(module._fsdp_wrap)
+
+                        is_large = unwrapped_params >= min_params
+                        if hasattr(obj, 'fsdp_wrap_fn') and isinstance(obj.fsdp_wrap_fn, Callable):
+                            return obj.fsdp_wrap_fn(module) or is_large
+                        else:
+                            return is_large
```
vchiley, line 309:
```
            def __auto_wrap_policy(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
                if recurse:
                    return True
                else:
                    if hasattr(module, '_fsdp_wrap'):
                        return bool(module._fsdp_wrap)

                    is_large = nonwrapped_numel >= min_params
                    if hasattr(obj, 'fsdp_wrap_fn') and isinstance(obj.fsdp_wrap_fn, Callable):
                        return obj.fsdp_wrap_fn(module) or is_large
                    else:
                        return is_large

            if is_torch_2_0:
                def _auto_wrap_policy(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
                    return __auto_wrap_policy(module, nonwrapped_numel)
            else:
                def _auto_wrap_policy(module: torch.nn.Module, recurse: bool, unwrapped_params: int) -> bool:
                    return __auto_wrap_policy(module, unwrapped_params)
```

2 stones, one bird
(or something)

dskhudia: Thanks for the suggestion. Accepted.

bcui19 (approved):
LGTM! Thank you for doing this!

PR #2086, opened 2023-03-22
Disable wrapping for fsdp if specified

mvpatel2000:
# What does this PR do?

Gives an option to not wrap a submodule with FSDP, which is necessary for non-trainable components

CO 1 9 4 3 (not linking because I don't want it to close)


dakinggg (approved):
LGTM

PR #2088, opened 2023-03-23
Fix EMA resumption issue with calling trainer.eval() before trainer.fit()

coryMosaicML:
This PR fixes a bug where if `trainer.eval()` is called before `trainer.fit()` when resuming training with EMA, the model parameters and buffers are not moved to GPU. The fix: ema parameters/buffers are now moved to device on `Event.EVAL_START` as well as `Event.FIT_START`.

### DIFF
--- a/composer/algorithms/ema/ema.py
+++ b/composer/algorithms/ema/ema.py
@@ -266,9 +266,8 @@ class EMA(Algorithm):
             # Verify that the ema params are on the correct device.
             # Needed to ensure doing eval before training can resume correctly.
             self.ema_model.move_params_to_device(destination_model=state.model)
-            if self.ema_weights_active is False:
-                # Swap out the training model for the ema model in state
-                self._ensure_ema_weights_active(state)
+            # Swap out the training model for the ema model in state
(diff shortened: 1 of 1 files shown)


composer/algorithms/ema/ema.py
```diff
@@ -262,7 +262,10 @@ def apply(self, event: Event, state: State, logger: Logger) -> None:
             # Update the ema model
             compute_ema(state.model, self.ema_model, smoothing=self.smoothing)
 
-        if event == Event.EVAL_START and self.ema_weights_active is False:
+        if event == Event.EVAL_START:
+            # Verify that the ema params are on the correct device.
+            # Needed to ensure doing eval before training can resume correctly.
+            self.ema_model.move_params_to_device(destination_model=state.model)
```
mvpatel2000, line 268: Can we please call this on PREDICT_START too?

mvpatel2000 (approved):

composer/algorithms/ema/ema.py
coryMosaicML: Done

PR #2091, opened 2023-03-24
Patch EMA with FSDP

mvpatel2000:
# What does this PR do?

Fixes EMA to work with FSDP. FSDP doesn't like directly accessing torch tensors -- instead we have to use copy. This has a small memory impact, but since it's not during a fwd/backward, it should be fine

Confirmed with manual test curves are same w or w/o FSDP:
```
Eval batch=1/1] Eval on eval data:
	 Eval metrics/eval/FrechetInceptionDistance-scale-0p0: 418.2500
	 Eval metrics/eval/MeanSquaredError: 1.1608
	 Eval metrics/eval/MeanSquaredError-bin-0-to-1: 1.1608
```
```
[Eval batch=1/1] Eval on eval data:
	 Eval metrics/eval/FrechetInceptionDistance-scale-0p0: 418.2500
	 Eval metrics/eval/MeanSquaredError: 1.1608
	 Eval metrics/eval/MeanSquaredError-bin-0-to-1: 1.1608
```

[CO-1943](https://mosaicml.atlassian.net/browse/CO-1943)

[CO-1943]: https://mosaicml.atlassian.net/browse/CO-1943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/algorithms/ema/ema.py
+++ b/composer/algorithms/ema/ema.py
@@ -402,22 +402,28 @@ class EMAParameters:
 
             for name, param in model.named_parameters():
                 if name in ema_params:
-                    param.data, ema_params[name] = ema_params[name], param.data
+                    # Use copy instead of raw data access (eg .data) doesn't work with FSDP
+                    dummy_param = param.clone()
+                    param.copy_(ema_params[name])
+                    ema_params[name].copy_(dummy_param)
 
             for name, buffer in model.named_buffers():
                 if name in ema_buffers:
-                    buffer.data, ema_buffers[name] = ema_buffers[name], buffer.data
+                    # Use copy instead of raw data access (eg .data) doesn't work with FSDP
+                    dummy_buffer = buffer.clone()
+                    buffer.copy_(ema_buffers[name])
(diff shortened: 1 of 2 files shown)


dskhudia (approved):

mvpatel2000:
> Any automated tests or does this fix existing failing tests?

This is very hard to unit test as it does not give an error, just fails to swap weights. I think it's fine without

dakinggg:
Can't you unit test by just comparing weights? id rather have an FSDP + EMA unit test added since it seems like this has been tricky

mvpatel2000:
> Can't you unit test by just comparing weights? id rather have an FSDP + EMA unit test added since it seems like this has been tricky

Ok I guess it is possible, it's just really narrow since it doesn't work yet with FULL_SHARD. I would prefer to wait until we rework surgery to interact with FSDP correctly and unit test them all correctly.

PR #2093, opened 2023-03-25
Add sentencepiece support to `HuggingFaceModel`

dakinggg:
# What does this PR do?
This PR adds support for saving sentencepiece tokenizers by serializing and loading the `spiece.model` file (e.g. https://huggingface.co/bigscience/T0pp/blob/main/spiece.model)

Closes [CO-1902](https://mosaicml.atlassian.net/browse/CO-1902)

# Before submitting

[CO-1902]: https://mosaicml.atlassian.net/browse/CO-1902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -179,6 +179,8 @@ extra_deps['nlp'] = [
     'datasets>=2.4,<3',
 ]
 
+extra_deps['sentencepiece'] = ['sentencepiece==0.1.97']
+
 extra_deps['mlperf'] = [
     # TODO: use pip when available: https://github.com/mlcommons/logging/issues/218
     # "mlperf_logging @ git+https://github.com/mlperf/logging.git",
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -276,13 +276,25 @@ class HuggingFaceModel(ComposerModel):
         if hf_tokenizer_state != {}:
             with tempfile.TemporaryDirectory() as _tmp_dir:
                 for filename, saved_content in hf_tokenizer_state.items():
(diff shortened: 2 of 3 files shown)


alextrott16 (approved):
LGTM! Thanks for taking care of this!

dakinggg:
@bandish-shah since mihir is out all week, can I get your admin review on this?

setup.py
```diff
@@ -179,6 +179,8 @@ def package_files(prefix: str, directory: str, extension: str):
     'datasets>=2.4,<3',
 ]
 
+extra_deps['sentencepiece'] = ['sentencepiece==0.1.97']
```
bandish-shah, line 182: [nit] Do we need to create another build target?  Can we just include this in the `nlp` dep?

bandish-shah (approved):
LGTM

setup.py
dakinggg: I'd like to keep it separate because I've had issues installing the `sentencepiece` dependency in the past, so I don't want to force all `nlp` users to install it. Sound ok?

bandish-shah: Could you elaborate on what type of issues a bit more?

I think it's ok to keep it as it's own dependency for now then in light of that.

dakinggg: I had to manually install protobuf when I was testing this out on windows subsystem for linux. I don't really know if this is a persistent issue, as I did not encounter it on my mac, or in CI.

PR #2101, opened 2023-03-28
Allow user to set ONNX opset version when Exporting for Inference

nik-mosaic:
Previously, our `torch.onnx.export` opset version (which controls which operators are supported for inference) was hardcoded to 13. This PR makes `onnx_opset_version` an optional argument which the user can pass into our Composer `export_for_inference` method, which is then passed into our call to `torch.onnx.export`. 

If the user doesn't provide an onnx_opset_version, we default to None, which means that our call to torch.onnx.export will use PyTorch's default ONNX opset version. 

The Torch default ONNX opset version changes each PyTorch version. It is [now 14](https://pytorch.org/docs/stable/onnx.html#torch.onnx.export), which should [fix](https://github.com/onnx/onnx/blob/main/docs/Operators.md) the attention [issues we are seeing](https://mosaicml.slack.com/archives/C046WPCT99U/p1679685345620879) (exporting tril, triu).

Closes https://mosaicml.atlassian.net/browse/CO-1969

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -19,7 +19,7 @@ from torchmetrics import Metric
 
 from composer.metrics import InContextLearningMetric
 from composer.models.base import ComposerModel
-from composer.utils import MissingConditionalImportError, dist, get_file, import_object, safe_torch_load
+from composer.utils import MissingConditionalImportError, dist, get_file, import_object, is_model_fsdp, safe_torch_load
 
 if TYPE_CHECKING:
     import transformers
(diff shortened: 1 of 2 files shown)


tests/utils/test_inference.py
```diff
@@ -130,6 +130,7 @@ def test_huggingface_export_for_inference_onnx(tiny_bert_config):
             save_path=save_path,
             sample_input=(sample_input, {}),
             dynamic_axes=dynamic_axes,
+            onnx_opset_version=13,
```
dakinggg, line 133: same

PR #2111, opened 2023-03-30
pin mcp

dakinggg:
# What does this PR do?
Pins MCLI due to an API change. Opened a JIRA to resolve the API change and assigned it to @mvpatel2000


bandish-shah (approved):
LGTM

PR #2116, opened 2023-04-04
Use state_dict Torchmetrics Serialization

nik-mosaic:
# What does this PR do?

Our metrics objects were previously serialized using pickle. The pickling includes many fields which are unnecessary and may change from version to version, causing a mismatch when we stop and start training with slightly different configurations or upgrade Torchmetrics.

This PR changes the saving and loading of Composer state objects so that state.train_metrics and state.eval_metrics use Torchmetrics' built in [`state_dict()`](https://github.com/Lightning-AI/torchmetrics/blob/df299dcc9fbb4ef210d0d81f7be4d63eb40ce195/src/torchmetrics/metric.py#L748) method instead of Pickle. The `state_dict()` method essentially returns a dictionary of `<metric_name, metric_value>` <key, value> pairs, without any of the other data.

When loading a state dict from its serialized version, we recreate the metric as follows:.
(1) Get the state.model
(2) Call model.get_metrics(), which creates a default version of the metrics
(3) Loop over the dictionary of  `<metric_name, metric_value>` pairs from the serialization, matching metric names to the metrics from model.get_metrics() and populating values with the saved `metric_value`. If a saved metric isn't present in the model, we skip it.

Other changes in this PR:
- Gate `_ensure_metrics_device_and_dtype`, which is only necessary for Deepspeed models, behind an is_model_deepspeed() check. This should fix [CO-1910](https://mosaicml.atlassian.net/browse/CO-1910).

Testing:
- We have modified the test_checkpoint.py and test_sharded_checkpoint.py files to include a metric equality check wherever we have a weight equality or an optimizer equality check.
- Added `test_load_remote_checkpoint`, a backwards compatibility checkpoint test. I have uploaded a checkpoint saved with Composer 0.13.5 and default dependenices (torchmetrics 0.11.3, etc.). This test downloads the checkpoint and ensures equivalency with a currently trained version. As we continue to push to Composer, this test will become more useful because our remote checkpoint will remain frozen while our local one will include all future trainer/model/metrics changes.
- Locally, all tests pass on torchmetrics 0.11.4, but we do not bump the version in this PR.
- @coryMosaicML has verified that [CO-1910](https://mosaicml.atlassian.net/browse/CO-1910) has been fixed with this PR.

CO-1918, CO-1907, CO-1853.

[CO-1910]: https://mosaicml.atlassian.net/browse/CO-1910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -770,9 +770,9 @@ class State(Serializable):
                 for k, v in attribute_value.items():
                     # No need to use __qualname__, we already know this corresponds to
                     # a metric object when we deserialize.
-                    # Along with the rest of a Composer checkpoint, the state_dict() and _computed attributes of 
+                    # Along with the rest of a Composer checkpoint, the state_dict() and _computed attributes of
                     # a Torchmetrics object are enough information to recreate it upon serialization. We only serialize
-                    # the minimum metric information to maximize backwards compatibility --- old checkpoints 
+                    # the minimum metric information to maximize backwards compatibility --- old checkpoints
                     # will continue to be compatible even if other Torchmetrics attributes have changed.
                     # metric._computed stores the cached value of the previous metric computation
                     # We need to serialize this because it cannot always be recomputed from the state dict.


tests/trainer/test_checkpoint.py
```diff
@@ -461,8 +507,68 @@ def test_load_weights(self, device, load_weights_only):
...
+        and should be equivalent after training continues.
+        Checkpoint saved using: Composer 0.13.5 with default dependencies.
+        """
+        import os
+
+        trainer_1 = self.get_trainer(save_folder='first', device=device)
+        trainer_1.fit()
+        trainer_1.close()
+
+        remote_checkpoint = 's3://mosaicml-internal-checkpoints-test/backwards_compatibility/trained_cpu_checkpoint_ep2.pt'
```
mvpatel2000, line 532: move to OCI?

mvpatel2000 (commented):
This adds a backwards compatibility test -- very nice! But is this change backwards compatible? The checkpoint appears to be off dev and not 0.13.5

composer/core/state.py
```diff
@@ -780,6 +765,20 @@ def state_dict(self) -> Dict[str, Any]:
                 serialized_value = [(type(obj).__qualname__, obj.state_dict()) for obj in ensure_tuple(attribute_value)]
             elif attribute_name in _STATE_DICT_SERIALIZED_ATTRIBUTES:
                 serialized_value = {type(obj).__qualname__: obj.state_dict() for obj in ensure_tuple(attribute_value)}
+            elif attribute_name == 'train_metrics':
+                serialized_value = {}
+                for k, v in attribute_value.items():
+                    v.persistent(mode=True)
```
dakinggg, line 771: could you leave behind a comment to the torchmetrics docs here please?

composer/core/state.py
```diff
@@ -780,6 +765,20 @@ def state_dict(self) -> Dict[str, Any]:
...
+                    # No need to use __qualname__, we already know this corresponds to
+                    # a metric object when we deserialize.
+                    serialized_value[k] = {'state_dict': v.state_dict(), '_computed': v._computed}
```
dakinggg, line 774: what is `_computed`?

composer/core/state.py
```diff
@@ -1058,16 +1057,51 @@ def load_state_dict(
...
+                # Get current metrics object and populate each metric present
+                # in serialization with serialized data via load_state_dict()
                 state_field_value = getattr(self, attribute_name)
-                for metric_name, metric in serialized_value.items():
-                    metric._device = self.device._device
-                    state_field_value[metric_name] = metric
+                for metric_name in state_field_value.keys():
+                    if metric_name not in serialized_value:
+                        continue
+                    state_field_value[metric_name]._update_count += 1
```
dakinggg, line 1066: also need a comment here please

composer/core/state.py
```diff
@@ -1058,16 +1057,51 @@ def load_state_dict(
...
+                    missing_keys, unexpected_keys = state_field_value[metric_name].load_state_dict(
```
dakinggg, line 1067: why strict false?

composer/core/state.py
```diff
@@ -1058,16 +1057,51 @@ def load_state_dict(
...
+                        serialized_value[metric_name]['state_dict'], strict=False)
+                    state_field_value[metric_name]._computed = serialized_value[metric_name]['_computed']
+                    state_field_value[metric_name].persistent(mode=True)
```
dakinggg, line 1070: do we need to/should we be settings persistent to `True` on load?

composer/trainer/trainer.py
```diff
@@ -1765,15 +1766,19 @@ def close(self):
...
             # Safety check to ensure the metric and data are on the same device. Normally not
             # needed because the metric is automatically on the same device as the model.
             # See https://torchmetrics.readthedocs.io/en/latest/pages/overview.html for details.
             metrics[name] = self.state.device.module_to_device(metric)
-            metric.set_dtype(torch.float32)  # type: ignore
-
+            if is_model_deepspeed(self.state.model):
+                # HACK: DeepSpeed somehow manages to convert metric internal states to its own dtype. When
+                # running with FP16, this tends to result in overflows. Let's assume FP32 is good enough.
+                for key in metric._defaults:
```
dakinggg, line 1777: how come this is a new for loop?

tests/trainer/test_checkpoint.py
```diff
@@ -461,8 +507,70 @@ def test_load_weights(self, device, load_weights_only):
...
```
dakinggg, line 532: lets parametrize this path? then we can easily add new checkpoints to it over time? better yet, we would ls a bucket and do all of them, but lets defer that for now :) also sorry, why doesn't this work with oci? it should.

dakinggg: oh jk, cause our pytest suite doesn't have an oci integration, sg

dakinggg (commented):
Thanks for digging through this @nik-mosaic! Left a few comments and questions, mostly around the usage of private attributes.

dakinggg (approved):

composer/core/state.py
```diff
@@ -1058,16 +1060,75 @@ def load_state_dict(
...
-                    for metric_name, metric in eval_metrics.items():
-                        metric._device = self.device._device
-                        state_field_value[eval_key][metric_name] = metric
+                for eval_key in state_field_value.keys():
+                    if eval_key not in serialized_value:
+                        continue
+                    for metric_name in state_field_value[eval_key].keys():
+                        if metric_name not in serialized_value[eval_key]:
+                            continue
+                        state_field_value[eval_key][metric_name]._update_count += 1
```
mvpatel2000, line 1105: why?

composer/core/state.py
```diff
@@ -1058,16 +1060,75 @@ def load_state_dict(
...
+                    if metric_name not in serialized_value:
+                        continue
+                    # Increment _update_count so it is non-zero, preventing Torchmetrics from warning us when we call metric.compute()
+                    state_field_value[metric_name]._update_count += 1
+                    if isinstance(serialized_value[metric_name], Metric):
+                        serialized_value[metric_name].persistent(mode=True)
+                        metric_state_dict = serialized_value[metric_name].state_dict()
+                        metric_computed_field = serialized_value[metric_name]._computed
+                    elif isinstance(serialized_value[metric_name], dict):
+                        metric_state_dict = serialized_value[metric_name]['state_dict']
```
mvpatel2000, line 1076: can you comment when these two different code paths would be hit?

mvpatel2000 (commented):
Why do we not need the backwards compatibiltiy hacks with this? because wont they still be the old format?

mvpatel2000 (approved):
LGTM, thank you for taking this on!

PR #2118, opened 2023-04-04
Add torch.compile support for torch 2.0

karan6181:
# What does this PR do?
- Support of `torch.compile` as part of Trainer argument.
- Add a new parameter `compile_config` in Trainer() which support model compilation based on the value of `compile_config` parameter.

CO-1956

# Testing
## End-to-end model training 

**Node:** 1
**GPUs:** 4

| Model    | Dataset  | With compile thoughput/samples_per_sec | Without compile thoughput/samples_per_sec | Performance % |
| -------- | -------- | -------------------------------------- | ----------------------------------------- | ------------- |
| HF BERT  | C4       | 4259                                   | 3360                                      | 26.75%        |
| ResNet50 | ImageNet | 7424                                   | 5557                                      | 33.60%        |

### FSDP
- Some known FSDP Torch 2.0 issues when doing eval (Training works fine): https://github.com/pytorch/pytorch/issues/93288

# Before submitting

### DIFF
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -223,6 +223,16 @@ def prepare_fsdp_module(
     state_dict_type = fsdp_config.get('state_dict_type', 'full')
     activation_checkpointing_reentrant = fsdp_config.get('activation_checkpointing_reentrant', True)
 
+    kwargs = {}
+    if is_torch_2_0:
+        # Support of new parameter `use_orig_params` in PyTorch 2.0 or higher.
+        # Setting this to `True` has FSDP use `module`'s original parameters via method
+        # `nn.Module.named_parameters` instead of FSDP's internal class `FlatParameter`. However,
+        # setting it to `False` exposes FSDP's internal class `FlatParameter` via method
+        # `nn.Module.named_parameters`.
+        # Setting it to `True` is mandatory when using `torch.compile()`.
+        kwargs['use_orig_params'] = fsdp_config.get('use_orig_params', True)
+
     # We choose to not wrap the ComposerModel directly, but instead wrap any submodules like `ComposerModel.model`
     # This makes it safer to call ComposerModel-specific functions like 'eval_forward' that
     # may make calls to sharded submodules. If we only wrap the submodules, then any call that ComposerModel makes
@@ -352,6 +362,7 @@ def prepare_fsdp_module(
                 sync_module_states=sync_module_states,
                 forward_prefetch=forward_prefetch,
                 limit_all_gathers=limit_all_gathers,
+                **kwargs,
             )
 
             # Activation Checkpointing
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1358,6 +1358,7 @@ class Trainer:
             compiled_model = torch.compile(self.state.model, **compile_config)  # pyright: ignore
             self.state.model = compiled_model._orig_mod
             self.state.model.forward = compiled_model.dynamo_ctx(self.state.model.forward)
+            is_model_compiled = True
         elif not is_torch_2_0 and compile_config is not None:
             log.warning(f'`torch.compile()` is supported for PyTorch 2.0 or higher.' +
                         f'Either update your PyTorch version or disable parameter by providing ' +
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -50,7 +50,7 @@ from composer.utils import (ExportFormat, MissingConditionalImportError, ObjectS
                             ensure_tuple, export_with_logger, extract_hparams, format_name_with_dist, get_device,
                             get_file, is_tpu_installed, map_collection, maybe_create_object_store_from_uri,
                             maybe_create_remote_uploader_downloader_from_uri, model_eval_mode, parse_uri,
-                            reproducibility)
+                            reproducibility, using_torch_2_0)
 
 if is_tpu_installed():
     import torch_xla.core.xla_model as xm
@@ -763,6 +763,9 @@ class Trainer:
             ``logging.basicConfig`` won't be called).
 
             .. seealso:: The :mod:`logging` module in Python.
+        use_torch_compile: (bool, optional):  Wraps a model and returns a compiled model. (default: ``False``)
+
+            .. note:: Only supported with PyTorch version >= 2.x
 
     Attributes:
         state (State): The :class:`.State` object used to store training state.
@@ -860,6 +863,9 @@ class Trainer:
 
         # Python logging
         python_log_level: Optional[str] = None,
+
+        # Compile model for PyTorch 2.x
+        use_torch_compile: Optional[bool] = False,
     ):
 
         self.auto_log_hparams = auto_log_hparams
@@ -908,6 +914,14 @@ class Trainer:
         self.fsdp_config = fsdp_config
         self.deepspeed_enabled = self.deepspeed_config is not None
         self.fsdp_enabled = self.fsdp_config is not None
+        # Determine whether torch compile is enabled
+        is_torch_2_0 = using_torch_2_0()
+        if is_torch_2_0 and use_torch_compile:
+            model = torch.compile(model)  # pyright: ignore
+        elif not is_torch_2_0 and use_torch_compile:
+            raise ValueError(f'torch.compile() is supported for PyTorch version >= 2.x.' +
+                             f'Either update your PyTorch version or disable parameter ' +
+                             f'`use_torch_compile` to `False`.')
 
         # Distributed
         if self.deepspeed_enabled or self.fsdp_enabled or dist.get_world_size() > 1:
@@ -1094,8 +1108,12 @@ class Trainer:
             'node_name': os.environ.get('NODENAME', 'unknown because NODENAME environment variable not set')
         })
 
-        if not isinstance(self.state.model, ComposerModel):
-            raise ValueError('Provided model should be a subclass of ComposerModel.')
+        if use_torch_compile:
+            model_type = self.state.model._orig_mod
+        else:
+            model_type = self.state.model
+        if not isinstance(model_type, ComposerModel):
+            raise ValueError('Provided model must be a subclass of ComposerModel.')
 
         # After running Event.INIT, then set the "optional" elements of state that could be passed in on FIT instead of INIT
         # Setting these attributes here ensures that algorithms do not depend on unavailable attributes during Event.INIT
@@ -1180,8 +1198,6 @@ class Trainer:
         # If using DDP or DeepSpeed, we need to wrap the ComposerModel
         # But store a reference to the original model for functions like `eval_forward`, `get_metrics`, etc.
         self._original_model = self.state.model
-        if not isinstance(self._original_model, ComposerModel):
-            raise ValueError('self.state.model must be a subclass of ComposerModel.')
 
         # Configure Deepspeed
         if self.state.deepspeed_config is not None:
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -763,9 +763,11 @@ class Trainer:
             ``logging.basicConfig`` won't be called).
 
             .. seealso:: The :mod:`logging` module in Python.
-        use_torch_compile: (bool, optional):  Wraps a model and returns a compiled model. (default: ``False``)
-
-            .. note:: Only supported with PyTorch version >= 2.x
+        compile_config (Dict[str, Any], optional): Configuration for torch compile. Only supported with PyTorch 2.0 or higher.
+            Checkout [`torch.compile`](https://pytorch.org/get-started/pytorch-2.0/) for more details.
+            To use torch compile with default values, set it to empty dictionary ``{}``.
+            To use torch compile with custom config, set to a dictionary such as ``{'mode': 'max-autotune'}``.
+            To disable torch compile, set to ``None``. (default: ``None``)
 
     Attributes:
         state (State): The :class:`.State` object used to store training state.
@@ -864,8 +866,8 @@ class Trainer:
         # Python logging
         python_log_level: Optional[str] = None,
 
-        # Compile model for PyTorch 2.x
-        use_torch_compile: Optional[bool] = False,
+        # compile config for PyTorch 2.0 or higher
+        compile_config: Optional[Dict[str, Any]] = None,
     ):
 
         self.auto_log_hparams = auto_log_hparams
@@ -893,6 +895,27 @@ class Trainer:
             precision = Precision(precision)
         _validate_precision(precision, device)
 
+        # check if provided model is compiled or not
+        is_torch_2_0 = using_torch_2_0()
+        is_model_compiled = False
+        if is_torch_2_0:
+            from torch._dynamo import OptimizedModule
+            if isinstance(model, OptimizedModule):
+                log.warning(f'Provided `model` is already compiled with `torch.compile(). Ignoring ' +
+                            f'parameter `compile_config` if provided. If you would like `Trainer()` ' +
+                            f'to takes care of model compilation, provide a un-compiled model and ' +
+                            f'`compile_config` parameter.')
+                # The `torch.compile` function returns an object of type `torch._dynamo.OptimizedModule`
+                # which wraps the original `nn.Module` object and later patches its forward method to
+                # optimized `self.forward` method.
+                is_model_compiled = True
+                compiled_model = model._orig_mod
+                if not isinstance(compiled_model, ComposerModel):
+                    raise ValueError(f'Provided `model` must be a subclass of ComposerModel. ' +
+                                     f'Instead found as type `{type(compiled_model)}`')
+                compiled_model.forward = model.dynamo_ctx(compiled_model.forward)  # pyright: ignore
+                model = compiled_model
+
         # Microbatching
         auto_microbatching = _is_auto_microbatching(device_train_microbatch_size, device=device)
         if auto_microbatching and profiler:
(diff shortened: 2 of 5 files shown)


vchiley:
Wouldn't this compile just the model? Wouldn't we want to instead compile the entire training graph (including the optimizer step)?
From the [torch compile tutorial](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html):

Screenshot 2023-04-04 at 3 58 58 PM

dskhudia:
@vchiley : This would take care of forward and backward. Optimizer, I recently learned, currently creates a lot  of graph breaks and [foreach=True](https://pytorch.org/docs/stable/generated/torch.optim.SGD.html#sgd) for optimizers is a better solution.

composer/trainer/trainer.py
```diff
@@ -759,6 +759,9 @@ class Trainer:
             ``logging.basicConfig`` won't be called).
 
             .. seealso:: The :mod:`logging` module in Python.
+        use_torch_compile: (bool, optional):  Wraps a model and returns a compiled model. (default: ``False``)
```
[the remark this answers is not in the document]
karan6181: Updated the doc in https://github.com/mosaicml/composer/pull/2118/commits/c4517821736fa5446ff1d2c415e97c70156b6ca2.

composer/trainer/trainer.py
```diff
@@ -888,6 +894,15 @@ def __init__(
             precision = Precision(precision)
         _validate_precision(precision, device)
 
+        # Determine whether torch compile is enabled
+        is_torch_2_0 = using_torch_2_0()
+        if is_torch_2_0 and use_torch_compile:
+            model = torch.compile(model)  # pyright: ignore
```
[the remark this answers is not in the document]
karan6181: Yes. `error: "compile" is not a known member of module`

composer/trainer/trainer.py
```diff
@@ -888,6 +894,15 @@ def __init__(
...
+        elif not is_torch_2_0 and use_torch_compile:
+            raise ValueError(f'torch.compile() is supported for PyTorch version >= 2.x.' +
```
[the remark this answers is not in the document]
karan6181: In the commit https://github.com/mosaicml/composer/pull/2118/commits/c4517821736fa5446ff1d2c415e97c70156b6ca2, I have converted to a warning message since we do not want to break the flow if user by mistakenly provided `compile_config` for torch < 2.0

composer/trainer/trainer.py
```diff
@@ -1175,8 +1205,6 @@ def __init__(
         # If using DDP or DeepSpeed, we need to wrap the ComposerModel
         # But store a reference to the original model for functions like `eval_forward`, `get_metrics`, etc.
         self._original_model = self.state.model
-        if not isinstance(self._original_model, ComposerModel):
```
karan6181, line 1178: Repetitive model check, hence removed.

dskhudia (approved):

dskhudia:
@karan6181 : This looks great. I have two more requests: 1) Can you add some of the benchmarking numbers in the PR description 2) Can you add a test where we run training on a small model with and without compile and assert that we produce close enough output? This test will help us capture compile related issue when we run the model.

karan6181:
> @karan6181 : This looks great. I have two more requests: 1) Can you add some of the benchmarking numbers in the PR description 2) Can you add a test where we run training on a small model with and without compile and assert that we produce close enough output? This test will help us capture compile related issue when we run the model.

Done.

composer/trainer/trainer.py
```diff
@@ -887,6 +895,28 @@ def __init__(
...
+                            f'to takes care of model compilation, provide a un-compiled model and ' +
+                            f'`compile_config` parameter.')
+                # The `torch.compile` function returns an object of type `torch._dynamo.OptimizedModule`
+                # which wraps the original `nn.Module` object and later patches its forward method to
+                # optimized `self.forward` method.
+                is_model_compiled = True
+                compiled_model = model._orig_mod
+                if not isinstance(compiled_model, ComposerModel):
+                    raise ValueError(f'Provided `model` must be a subclass of ComposerModel. ' +
+                                     f'Instead found as type `{type(compiled_model)}`')
```
mvpatel2000, line 915: Can this error say something also indicating model is already compiled which may be related? In case user doesn't have python logging on

composer/trainer/trainer.py
```diff
@@ -1325,6 +1353,23 @@ def __init__(
...
+            self.state.model.forward = compiled_model.dynamo_ctx(self.state.model.forward)
+            is_model_compiled = True
+            # update local_hparams to ensure the `is_model_compiled` is set correctly for
+            # debugging purpose and for unit test.
+            if self.auto_log_hparams:
+                self.local_hparams['is_model_compiled'] = is_model_compiled
+        elif not is_torch_2_0 and compile_config is not None:
+            log.warning(f'`torch.compile` is supported for PyTorch 2.0 or higher.' +
+                        f'Either update your PyTorch version or disable parameter by providing ' +
+                        f'`compile_config` to `None`.')
```
mvpatel2000, line 1371:
```suggestion
            warnings.warn(f'`torch.compile` is supported for PyTorch 2.0 or higher.' +
                        f'Either update your PyTorch version or disable parameter by providing ' +
                        f'`compile_config` to `None`.')
```

mvpatel2000: We use warnings.warn when it's a user parameter specification issue, eg there's an actionable step

tests/trainer/test_trainer.py
```diff
@@ -98,6 +99,52 @@ def test_optimizer_params_on_device(
         target_device = 'cuda' if device == 'gpu' else 'cpu'
         assert all(param.device.type == target_device for param in parameters)
 
+    @pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.0'),
+                        reason='requires PyTorch 2.0 or higher')
+    @pytest.mark.parametrize('compile_config', [(None, False), ({}, True), ({'mode': 'reduce-overhead'}, True)])
+    def test_torch_compile(self, model: ComposerModel, compile_config: Any):
```
mvpatel2000, line 105: can u post screenshot showing this was run locally?

mvpatel2000 (commented):
Can we simplify the utility function in `misc.py` to 
```return version.parse(torch.__version__) >= version.parse('2.0.0')```
in this PR

File: composer/trainer/dist_strategy.py [hunk 8]
```diff
@@ -223,6 +223,16 @@ def sync_hook(*args):
...
 
+    kwargs = {}
+    if is_torch_2_0:
+        # Support of new parameter `use_orig_params` in PyTorch 2.0 or higher.
+        # Setting this to `True` has FSDP use `module`'s original parameters via method
+        # `nn.Module.named_parameters` instead of FSDP's internal class `FlatParameter`. However,
+        # setting it to `False` exposes FSDP's internal class `FlatParameter` via method
+        # `nn.Module.named_parameters`.
+        # Setting it to `True` is mandatory when using `torch.compile()`.
+        kwargs['use_orig_params'] = fsdp_config.get('use_orig_params', True)
```
dakinggg, line 234: what error does the user face if using FSDP + torch2 and setting `use_orig_params` to `False`? Should we error here in that case with an informative error?

composer/trainer/trainer.py
```diff
@@ -887,6 +895,28 @@ def __init__(
...
 
+        # check if provided model is compiled or not
+        is_torch_2_0 = using_torch_2_0()
+        is_model_compiled = False
+        if is_torch_2_0:
+            from torch._dynamo import OptimizedModule
+            if isinstance(model, OptimizedModule):
+                log.warning(f'Provided `model` is already compiled with `torch.compile`. Ignoring ' +
+                            f'parameter `compile_config` if provided. If you would like `Trainer` ' +
+                            f'to takes care of model compilation, provide a un-compiled model and ' +
```
dakinggg, line 906:
```suggestion
                            f'to takes care of model compilation, provide a not-compiled model and ' +
```

composer/trainer/trainer.py
```diff
@@ -887,6 +895,28 @@ def __init__(
             precision = Precision(precision)
         _validate_precision(precision, device)
 
+        # check if provided model is compiled or not
+        is_torch_2_0 = using_torch_2_0()
+        is_model_compiled = False
+        if is_torch_2_0:
+            from torch._dynamo import OptimizedModule
+            if isinstance(model, OptimizedModule):
```
dakinggg, line 903: do we need to recursively check all module for being `OptimizedModule` like we do for FSDP?

tests/trainer/test_trainer.py
```diff
@@ -98,6 +99,52 @@ def test_optimizer_params_on_device(
         target_device = 'cuda' if device == 'gpu' else 'cpu'
         assert all(param.device.type == target_device for param in parameters)
 
+    @pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.0'),
```
dakinggg, line 102: can we also add an FSDP + compile test please?

karan6181:
> Can we simplify the utility function in `misc.py` to `return version.parse(torch.__version__) >= version.parse('2.0.0')` in this PR

Done in https://github.com/mosaicml/composer/pull/2118/commits/d0e8481b1a3de2bb20239802a9adfac217fb45cc.

composer/trainer/trainer.py
karan6181: If the user has already compiled the model, then `model._orig_mod` would return their original model. So if user does not inherit `ComposerModel` then `model._orig_mod` would catch it and we throw an error. `torch.compile` doesn't change the underlying model inheritance type.

composer/trainer/trainer.py
karan6181: Done in https://github.com/mosaicml/composer/pull/2118/commits/d0e8481b1a3de2bb20239802a9adfac217fb45cc.

composer/trainer/trainer.py
karan6181: Done in https://github.com/mosaicml/composer/pull/2118/commits/d0e8481b1a3de2bb20239802a9adfac217fb45cc.

tests/trainer/test_trainer.py
karan6181: Done in https://github.com/mosaicml/composer/pull/2118/commits/d0e8481b1a3de2bb20239802a9adfac217fb45cc. 

**Test output:**

```
collected 6 items

tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[None-Precision.FP32] PASSED                                                                          [ 16%]
tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[None-Precision.AMP_BF16] PASSED                                                                      [ 33%]
tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[None-Precision.AMP_FP16] PASSED                                                                      [ 50%]
tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[compile_config1-Precision.FP32] PASSED                                                               [ 66%]
tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[compile_config1-Precision.AMP_BF16] PASSED                                                           [ 83%]
tests/trainer/test_trainer.py::TestTrainerInitOrFit::test_fsdp[compile_config1-Precision.AMP_FP16] PASSED                                                           [100%]

============================================================================ 6 passed in 5.55s ============================================================================
```

tests/trainer/test_trainer.py
karan6181: Sure. 

**Test output:**
```
collected 3 items

tests/trainer/test_trainer.py::TestTrainerInit::test_torch_compile[compile_config0] PASSED                                                                          [ 33%]
tests/trainer/test_trainer.py::TestTrainerInit::test_torch_compile[compile_config1] PASSED                                                                          [ 66%]
tests/trainer/test_trainer.py::TestTrainerInit::test_torch_compile[compile_config2] PASSED                                                                          [100%]

============================================================================ 3 passed in 1.99s ============================================================================
```

composer/trainer/dist_strategy.py
karan6181: - `FSDP` + `torch 2.0` + `use_orig_params=False` = no error
- `FSDP` + `torch 2.0` + `use_orig_params=True` = no error
- `FSDP` + `torch 2.0` + `use_orig_params=True` + without `torch.compile` = no error
- `FSDP` + `torch 2.0` + `use_orig_params=False` + with `torch.compile` = `AssertionError: Dynamo only supports FSDP with use_orig_params=True`

There is no harm in keeping `use_orig_params` to `True` for `torch 2.0` irrespective of torch.compile or not. If we don't do that then it would require a little complex logic check here where we have to check if the provided model by the user is compiled or not or if composer compile the model then we have to pass some sort of variable in `dist_strategy.py` to let `prepare_fsdp_module()` know about this. Hence, we enables it by default for torch 2.0

composer/trainer/trainer.py
karan6181: I don't think so since `torch.compile` returns a `torch._dynamo.OptimizedModule` which wraps the ComposerModel passed in as an argument but does not actually inherit from it. So just checking the first layer should suffice.

composer/trainer/trainer.py
```diff
@@ -1325,6 +1353,23 @@ def __init__(
...
+            self.state.model.forward = compiled_model.dynamo_ctx(self.state.model.forward)
+            is_model_compiled = True
+            # update local_hparams to ensure the `is_model_compiled` is set correctly for
+            # debugging purpose and for unit test.
+            if self.auto_log_hparams:
+                self.local_hparams['is_model_compiled'] = is_model_compiled
+        elif not is_torch_2_0 and compile_config is not None:
+            warnings.warn(f'`torch.compile` is supported for PyTorch 2.0 or higher.' +
+                          f'Either update your PyTorch version or disable parameter by providing ' +
+                          f'`compile_config` to `None`.')
```
vchiley, line 1371: Should this be something like a ValueError and not just a warning?

composer/trainer/trainer.py
```diff
@@ -1325,6 +1353,23 @@ def __init__(
             # Only wrap the module if required
             self.state.model = prepare_ddp_module(self.state.model, self._find_unused_parameters)
 
+        # The model would need to be torch.compile()'d after being wrapped in a distributed strategy
+        # to take advantage of any graph breaks.
+        if is_torch_2_0 and not is_model_compiled and compile_config is not None:
+            compiled_model = torch.compile(  # pyright: ignore [reportGeneralTypeIssues]
+                self.state.model, **compile_config)
+            self.state.model = compiled_model._orig_mod
+            self.state.model.forward = compiled_model.dynamo_ctx(self.state.model.forward)
```
vchiley, line 1362: Should this block, be combined with the logic block @ line 900?

vchiley (commented):
no llm throughput testing?

composer/trainer/trainer.py
karan6181: I have kept warning message since we do not want to break the flow if user by mistakenly provided compile_config for torch < 2.0 and the training can still proceed even if we don't compile the model.

composer/trainer/trainer.py
karan6181: According to [this](https://dev-discuss.pytorch.org/t/torchdynamo-update-11-making-fsdp-and-dynamo-work-together/1037) pytorch document, DDP and FSDP wrappers have special optimizations that run with `torch.compile`. Hence, the recommendation is to compile the model after being wrapped in either FSDP or DDP.

karan6181:
> no llm throughput testing?

We have a plan to test LLM and stable diffusion models, however, it should not block this PR going in. Thanks!

mvpatel2000 (approved):

composer/trainer/trainer.py
vchiley: Would this have stopped us from moving logic block @ line 900 down here?

Alternatively, if a user passes in a compiled model should we throw a value error that tells the user to pass in a regular model + a compile config so they can get the FSDP optimizations?

msaroufim:
@vchiley @mvpatel2000 reaching out from Twitter

Would strongly suggest adding a config for `torch.set_float32_matmul_precision("high")` since enabling tensor cores will make a dramatic difference for perf and for experiments make sure to experiment with the mode because `reduce-overhead` will have dramatic speedups for smaller batch size and `max-autotune` will be worthwhile for longer training runs

Also since our official release optimizer support is stronger meaning you should be able to compile your entire training loop

mvpatel2000:
@msaroufim  thanks for the helpful tip! I'll try it out very soon. Do you have a sense of when this may actually affect numerical stability / is not worth using? or is this something safe to turn on by default?

For `reduce-overhead` vs. `max-autotune`, do you have a ballpark sense of what "smaller batch size" means?

mvpatel2000:
> Would strongly suggest adding a config for torch.set_float32_matmul_precision("high") 

@msaroufim I'm seeing slightly choppier performance but not much improvement about this. I do have:
```
torch.backends.cuda.matmul.allow_tf32 = True
torch.backends.cudnn.allow_tf32 = True
```
and torch autocast (AMP + fp16 since I'm seeing slowdown w bf16) enabled. Would these flags offset the flag you suggested?
image

msaroufim:
The choppy performance suggests a recompilation is happening, this could either be because your benchmark has dynamic shape inputs in which case set `dynamic=True` (try the main branch though) since stable support isn't that great for dynamic shapes

Or because you have lots of graph breaks in your model, try running `fullgraph=True` which should hard error and try out the recompilation profiler which will tell you exactly what's changing about your program to warrant a recompilation https://pytorch.org/docs/master/compile/troubleshooting.html#excessive-recompilation

If you have a self contained repro too please lmk 

EDIT: Actually yes using autocast is known to graphbreak so best to move your autocast context manager outside of the compilation https://github.com/pytorch/pytorch/issues/100241

mvpatel2000:
Sorry, to be clear, this is just with `torch.set_float32_matmul_precision("high")`. I have not enabled torch compile yet (will do so soon, likely later today or tomorrow).

The autocast bit is very helpful though... I might have an issue with our generate function. I will take a look

PR #2123, opened 2023-04-05
Update auto microbatching warning

mvpatel2000:
# What does this PR do?

Updates auto microbatching warning to remove experimental label and instead highlight failure cases.

Also, swaps to non-deprecated fn (new name) and explicitly turns off sync context for non-DDP

[CO-2032](https://mosaicml.atlassian.net/browse/CO-2032)

[CO-2032]: https://mosaicml.atlassian.net/browse/CO-2032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/trainer/trainer.py
```diff
@@ -2240,7 +2243,7 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int,
         device_batch = deepcopy(self.state.batch)
 
         microbatch_num_samples = self._train_data_spec.get_num_samples_in_batch(self.state.batch)
-        if self.deepspeed_enabled:
+        if self.deepspeed_enabled or not isinstance(self.state.model, DistributedDataParallel):
```
dakinggg, line 2246: were you intending to check if the model is FSDP here (this is what the pr comment says)?

mvpatel2000: Sorry, it should be non-DDP. PR comment is bad (I assuemd non-DDP = FSDP) but it should be non-DDP since it includes no parallelism

dakinggg (approved):

PR #2133, opened 2023-04-10
Add SlackLogger w/ custom formatting to composer/logger 

waiwuc:
# What does this PR do?
1. Allows training metrics to be output to a slack channel using [slack webhooks](https://api.slack.com/messaging/webhooks. Create your own slack app and add an incoming webhook url, then export the url as an envar `SLACK_WEBHOOK_URL`. 
2. Allows slack message to be formatted using slack [block kit builder ](https://app.slack.com/block-kit-builder/T01G9MS9XUG). Pass in a lambda function that formats your metrics according to block kit builder guidelines. 
3. If formatter function is not passed in, default to logging a header + text section where keys are bolded and values are unformatted. 
4. Since formatter function can be passed into `SlackLogger` as a first class function, more complex layouts can be created easily. 

NOTE: See example use case in `example/slack_log_metrics.py`. 

## Examples 
- With simple formatting (just text section where keys are bolded) 
Screenshot 2023-04-10 at 4 13 54 PM
- With header section (steps as headers) 
Screenshot 2023-04-10 at 4 14 17 PM
Screenshot 2023-04-10 at 4 14 22 PM


composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,67 @@
+from typing import Any, Callable, Dict, Optional
+
+from slack_sdk import WebhookClient
+
+from composer.loggers.logger_destination import LoggerDestination
+
+
+class SlackLogger(LoggerDestination):
```
dakinggg, line 13: Should we add some `filter_keys` option or something? Otherwise this is going to log _everything_ to slack right?

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,67 @@
...
+        blocks += [
+            {
+                "type": "section", "text": {"type": "mrkdwn", "text": f"*{k}:* {v}"}
+            }
+            for k, v in data.items()
+        ]
+        return blocks
+        
+    def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        self.logged_metrics.update(metrics)
```
dakinggg, line 57: what is `self.logged_metrics` used for?

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,67 @@
...
+        webhook_url: str,
+        log_metrics_formatter_func: Optional[Callable] = None,
+    ) -> None:
+        self.webhook_url = webhook_url
+        self.logged_metrics: Dict[str, float] = {}
+        self.log_metrics_formatter_func = log_metrics_formatter_func if log_metrics_formatter_func else self._default_log_bold_key_normal_value_pair_with_header
+
+    # Rich message layouts can be created using Slack Blocks Kit.
+    # See documentation here: https://api.slack.com/messaging/composing/layouts
+    def _log_to_slack(
```
dakinggg, line 37: I think this function should either not be on the class, and just accept all the args (i.e. not use `self.webhook_url`), or use everything via `self` (i.e. `self.client`, `self.log_metrics_formatter_func`, `self.webhook_url`).

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,67 @@
...
+        webhook_url (str): Slack Webhook URL
+        log_metrics_formatter_func ((...) -> Any | None): A formatter function that returns list of blocks to be sent to slack
+    """
+
+    def __init__(
+        self,
+        webhook_url: str,
+        log_metrics_formatter_func: Optional[Callable] = None,
+    ) -> None:
+        self.webhook_url = webhook_url
```
eracah, line 31: What do you think of webhook url being optional if the user sets the SLACK_WEBHOOK_URL env variable. Then it would be a lot easier for someone to write a script because they wouldn't have to go look for it every time or save it to their script

examples/slack_log_metrics.py
```diff
@@ -0,0 +1,41 @@
...
+from composer.loggers import WandBLogger, ConsoleLogger, SlackLogger
+
+import os
+
+transform = transforms.Compose([transforms.ToTensor()])
+dataset = datasets.MNIST("data", train=True, download=True, transform=transform)
+train_dataloader = DataLoader(dataset, batch_size=128)
+
+SLACK_WEBHOOK_URL=os.getenv('SLACK_WEBHOOK_URL')
+
```
eracah, line 21: should we also include an example for using the slack logger with a callback?

composer/loggers/__init__.py
```diff
@@ -21,22 +21,12 @@
...
-    'ProgressBarLogger',
-    'WandBLogger',
-    'RemoteUploaderDownloader',
-    'TensorboardLogger',
-    'CometMLLogger',
-    'ConsoleLogger',
-    'MLFlowLogger',
-    'SlackLogger'
+    'Logger', 'LoggerDestination', 'FileLogger', 'InMemoryLogger', 'ProgressBarLogger', 'WandBLogger',
+    'RemoteUploaderDownloader', 'TensorboardLogger', 'CometMLLogger', 'ConsoleLogger', 'MLFlowLogger', 'SlackLogger'
```
dakinggg, line 31: please add a trailing comma back here to keep the formatting

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Args:
+        webhook_url (str): Slack Webhook URL
+        log_metrics_formatter_func ((...) -> Any | None): A formatter function that returns list of blocks to be sent to slack
+    """
+
+    def __init__(
+        self,
+        webhook_url: str = '',
+        log_metrics_formatter_func: Optional[Callable] = None,
```
eracah, line 29: should we add a log_interval here, where the user can specify how often to log a metric to slack

eracah (commented):
Looks pretty good. Thanks, Wai. Just a few a's.

composer/loggers/slack_logger.py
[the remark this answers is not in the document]
waiwuc: `log_metrics_formatter_func` is actually variadic, since we pass in `data` as the only required positional argument and `**kwargs`

There should be a way to cast this Callable as variadic though, let me look into it

waiwuc: let me do that, I think I could add it in the formatter function as well. Ideally have one param to be one configuration with both formatting + interval

waiwuc: I will defer this to another iteration, interval is a little harder to code up. Instead ill have an include_keys config that allows us to log only essential keys, so we minimize the log spam

composer/loggers/slack_logger.py
waiwuc: excess, i just removed it

composer/loggers/slack_logger.py
```diff
@@ -1,64 +1,192 @@
...
-        self.log_metrics_formatter_func = log_metrics_formatter_func if log_metrics_formatter_func else self._default_log_bold_key_normal_value_pair_with_header
+        try:
+            import slack_sdk
+            del slack_sdk
+        except ImportError as e:
+            raise MissingConditionalImportError('health_checker', 'slack_sdk', None) from e
+
+        self.client = WebClient()
+        self.slack_logging_api_key = os.environ.get('SLACK_LOGGING_API_KEY', None)
+        self.channel_id = os.environ.get('SLACK_LOGGING_CHANNEL_ID', None)
```
waiwuc, line 97: Changed this to use the post message api instead of webhooks, because this allows user to send messages to a channel directly.

Webhooks with custom channels has been deprecated, see https://glitch.com/~slack-channel-webhook-blueprint

composer/loggers/slack_logger.py
```diff
@@ -1,64 +1,192 @@
...
+
+        self.log_metrics_config = log_metrics_config
+        self.log_traces_config = log_traces_config
+        self.log_hparams_config = log_hparams_config
+
+        # Maintain logger state
+        now = time.time()
+        self.logs_since_last_log, self.last_log_time = list(Any), now
+        self.traces_since_last_log, self.last_trace_time = list(Any), now
+        self.hparams_since_last_log, self.last_hparams_time = list(Any), now
```
waiwuc, line 111: Allows different time intervals and formats for sending logs / traces / hparams. In most cases, user would implement just one out of these 3

composer/loggers/slack_logger.py
waiwuc: done!

composer/loggers/slack_logger.py
waiwuc: I updated to use this function for metrics, traces and hparams, so I needed to parametrize it

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,67 @@
...
+    """Log metrics to slack, using `Slack Webhook URL <https://api.slack.com/messaging/webhooks>`
+    
+    .. note::
+        See example instantiating of SlackLogger in `examples/slack_log_metrics.py`
+        Webhook URL will be passed as an environment variable.
+        Formatter function can be defined at callsite or default formatter function will be applied if not provided.
+         
+    Args:
+        webhook_url (str): Slack Webhook URL
+        log_metrics_formatter_func ((...) -> Any | None): A formatter function that returns list of blocks to be sent to slack
```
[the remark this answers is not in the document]
waiwuc: Pulled up to doc string

examples/slack_log_metrics.py
```diff
@@ -0,0 +1,41 @@
+from torchvision import datasets, transforms
```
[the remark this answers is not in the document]
waiwuc: done

waiwuc:
@eracah , @dakinggg , thanks for the reviews! 

This should be ready for another review - the major things changed are
1. Use Slack's `postMessage` api instead of webhooks. This allows user to send messages to a channel directly. Webhooks with custom channels has been deprecated, see https://glitch.com/~slack-channel-webhook-blueprint
2. implement `log_traces` and `log_hparam` in addition to `log_metrics`. This design is now generic enough to log all 3 types with different configs. 
3. Update formatter functions to configs that consist of intervals, format function, and filter keys, which allow users more control over which keys to log in what format and how frequently. 

Appreciate your reviews a ton :)

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Log metrics to slack, using Slack postMessage api - https://api.slack.com/methods/chat.postMessage.
```
dakinggg, line 4: Could you please move all of this into the `SlackLogger` class docstring? I'm pretty sure it needs to be there for our docs to pick it up. (although if you build the docs locally and it shows up, that is also fine)

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+
+from composer.loggers.logger_destination import LoggerDestination
+from composer.utils import MissingConditionalImportError
+
+
+class SlackLogConfig:
+    """Log config for slack logger, including formatter function, keys to include, and interval in seconds.
+
+    formatter_func ((...) -> Any | None): A formatter function that returns list of blocks to be sent to slack
+    include_keys (Set[str]): A set of metric/logs/traces keys to include in the message. If None, no keys are included.
```
dakinggg, line 54: what do you think about defaulting to all keys if `include_keys=None`. I think that behavior might be more useful as a default? not completely sure though

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+    interval_in_seconds (int): Interval in seconds to send logs to slack. If None, message is sent after every log call.
+    Note that no more than 50 items are allowed to send in a single message.
```
dakinggg, line 56: Should we flush after 50 items are present regardless of time interval?

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+        data: Dict[str, Any],
+        last_time_flushed: float,
+        blocks_to_log: List[Dict[str, Any]],
+        config: SlackLogConfig,
+        **kwargs,  # can be used to pass additional arguments to the formatter function (eg for headers)
+    ) -> Tuple[float, List[Dict[str, Any]]]:
+        filtered_data = {k: v for k, v in data.items() if k in config.include_keys} if config.include_keys else data
+        blocks = config.formatter_func(
+            filtered_data, **
+            kwargs) if config.formatter_func else self._default_log_bold_key_normal_value_pair_with_header(
```
dakinggg, line 124:
```suggestion
            kwargs) if config.formatter_func is not None else self._default_log_bold_key_normal_value_pair_with_header(
```

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+                filtered_data, **kwargs)
+
+        updated_blocks_to_log = blocks + blocks_to_log
+
+        if len(updated_blocks_to_log) > 50:
+            raise RuntimeError('Cannot send more than 50 items in a single slack message')
+
+        if len(updated_blocks_to_log) > 0 and self.channel_id and (
```
dakinggg, line 132:
```suggestion
        if len(updated_blocks_to_log) > 0 and self.channel_id is not None and (
```

dakinggg (approved):
LGTM! I would like some unit tests (with the slack client mocked), but am also ok if you'd prefer to get it in so people can start playing with it.

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+    ) -> Tuple[float, List[Dict[str, Any]]]:
+        filtered_data = {k: v for k, v in data.items() if k in config.include_keys} if config.include_keys else data
+        blocks = config.formatter_func(
+            filtered_data, **
+            kwargs) if config.formatter_func else self._default_log_bold_key_normal_value_pair_with_header(
+                filtered_data, **kwargs)
+
+        updated_blocks_to_log = blocks + blocks_to_log
+
+        if len(updated_blocks_to_log) > 50:
```
dakinggg, line 129: should we send in groups of 50 rather than just erroring out? or at least truncate and warn rather than error. I would not be super happy if the slack logger taking in too many messages crashed my run :P

composer/loggers/slack_logger.py
waiwuc: yes that would be better behavior than crashing the run, let me add this change

composer/loggers/slack_logger.py
waiwuc: yeah good point, None and empty should probably have different behavior, even though there's no use case for an empty set. 

Will make the change and update the docs.

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+            del slack_sdk
+        except ImportError as e:
+            raise MissingConditionalImportError('health_checker', 'slack_sdk', None) from e
+
+        self.client = WebClient()
+        self.slack_logging_api_key = os.environ.get('SLACK_LOGGING_API_KEY', None)
+        self.channel_id = os.environ.get('SLACK_LOGGING_CHANNEL_ID', None)
+
+        if self.slack_logging_api_key is None or self.channel_id is None:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
waiwuc: good point

composer/loggers/slack_logger.py
waiwuc: Ill make the change to flush the logs when exceeding 50. 

Was probably also a mistake to pick `RuntimeError` for these types of errors. Havent looked into the error boundaries in trainer, but wondering if we've thought of having a `LoggerError` class that could be caught and printed to std out rather than throwing?

dakinggg: we could probably be better with our errors. I think we mostly just abuse the standard errors and don't do much special try/catch like that

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,190 @@
...
+        log_metrics_config (SlackLogConfig): Configuration for logging metrics to slack.
+        log_traces_config (SlackLogConfig): Configuration for logging traces to slack.
+        log_hparams_config (SlackLogConfig): Configuration for logging hyperparameters to slack.
+    """
+
+    def __init__(
+        self,
+        log_metrics_config: Optional[SlackLogConfig] = None,
+        log_traces_config: Optional[SlackLogConfig] = None,
+        log_hparams_config: Optional[SlackLogConfig] = None,
```
eracah, line 85:
```suggestion
        formatter_func: Optional[Callable[..., List[Dict[str, Any]]]] = None,,
        include_keys: Optional[Set[str]] = None,
        interval_in_seconds: Optional[int] = None,
```

I think having the SlackLogConfig for each category logged is a bit of overkill:

1. It is kinda an unnecessary lift for a user learn about this new class and instantiate a new object for every type of filtering they want to do.
2. It is probably overkill to have a different config for traces, hyperparameters, and metrics. (Traces are only really used as debugging and it is easy for a user to mistake certain things as hyperparameters, which are actually metrics, so the user would be playing a game of guess and check to set the right config.
3. Seems like we want just an MVP. This seems like a premature abstraction

We could also a do a dictionary instead of flat arguments

eracah (changes requested):
Looks pretty good, but I really think we should simplify the logging config part. It would be simpler and cover almost every situation just to have one set of `formatter_func`, `include_keys`, and `interval_in_seconds`. 

Also this would set an unfair precedent for other loggers to use this config structure, which I think would overcomplicate and boiler-platify the rest of the Logger codebase

composer/loggers/slack_logger.py
waiwuc: moved

composer/loggers/slack_logger.py
waiwuc: yes that makes sense, updated!

waiwuc:
removed configs for hparams/metrics/traces, we can configure the `SlackLogger` directly, and use one config for all 3 types of logs. 

Also added a bunch of unit tests

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+        except ImportError as e:
+            raise MissingConditionalImportError('slack_logger', 'slack_sdk', None) from e
+
+        self.slack_logging_api_key = os.environ.get('SLACK_LOGGING_API_KEY', None)
+        self.channel_id = os.environ.get('SLACK_LOGGING_CHANNEL_ID', None)
+
+        if self.slack_logging_api_key is None:
+            raise RuntimeError('SLACK_LOGGING_API_KEY must be set as environment variable')
+        if self.channel_id is None:
+            raise RuntimeError('SLACK_LOGGING_CHANNEL_ID must be set as environment variable')
```
eracah, line 78: We could do something where we allow the user to optionally pass these into the constructor and if they are not passed in then we look at the env variables. Wdyt?

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+                    interval_in_seconds=1
+                ),
+            ],
+        )
+
+        trainer.fit()
+
+    Args:
+        formatter_func ((...) -> Any | None): A formatter function that returns list of blocks to be sent to slack.
+        include_keys (Set[str]): A set of metric/logs/traces keys to include in the message. If None, all keys are included.
```
eracah, line 52: I think we might want to tweak the defaults here. I don't think by default all keys should be included. It might be too easy for a user to just pollute their slack channel accidentally. We might want to make this a required argument. If the user passes in a sequence, then those are the keys logged and if they pass in 'all' or '*' or something then all keys are logged

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+        interval_in_seconds (int): Interval in seconds to send logs to slack. If None, message is sent after every log call.
+        max_logs_per_message (int)(default:50): Maximum number of logs to send in a single message. Note that no more than 50 items are allowed to send in a single message.
+        If more than 50 items are stored in buffer, the message flushed without waiting the full time interval.
+    """
+
+    def __init__(
+        self,
+        formatter_func: Optional[Callable[..., List[Dict[str, Any]]]] = None,
+        include_keys: Optional[Set[str]] = None,
```
eracah, line 61: I don't think this should be a set. Let's make it a `Sequence` and then we can cast it to set a inside the function

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+            **kwargs,  # can be used to pass additional arguments to the formatter function (eg for headers)
+    ):
+        filtered_data = {k: v for k, v in data.items() if k in self.include_keys
+                        } if self.include_keys is not None else data
+        blocks = self.formatter_func(
+            filtered_data, **
+            kwargs) if self.formatter_func is not None else self._default_log_bold_key_normal_value_pair_with_header(
+                filtered_data, **kwargs)
+
+        self.logs += blocks
```
eracah, line 101: hmmm I'm not sure appending to a list is the best approach here because if the user sets an infrequent interval, they will see pretty stale data and if they set a high max_logs_per_message they will get overwhelmed with a full history for every key. 
I think we may want to either have a queue or remove values from the list based on a separate dictionary of the keys that are "pending" to be logged and if a new value for that key comes in before that key, value pair gets posted, then we overwrite it?

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+        now = time.time()
+
+        if len(self.logs) > 0 and self.channel_id is not None and (
+                len(self.logs) >= self.max_logs_per_message or self.interval_in_seconds is None or
+                now - self.last_log_time >= self.interval_in_seconds):
```
eracah, line 106: this is a pretty complicated boolean. can we maybe split it up into a few boolean variables?

I think `self.channel_id is not None` is unnecessary since the code errors out if it is None right?
```suggestion
        interval_has_elapsed = (self.interval_in_seconds is None or (now - self.last_log_time) >= self.interval_in_seconds)
        if len(self.logs) >= self.max_logs_per_message or (interval_has_elapsed and len(self.logs) > 0): 
```
or you could even set self.interval_in_seconds to 0 if it is passed in as None

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+            self.client.chat_postMessage(token=f'{self.slack_logging_api_key}',
+                                         channel=self.channel_id,
+                                         blocks=self.logs[:self.max_logs_per_message],
+                                         text=str(self.logs[:self.max_logs_per_message]))
```
eracah, line 110: the `text` variable is supposed to be the formatted text in either plaintext or markdown right? it's not supposed be just a stringification of the block right? https://api.slack.com/methods/chat.postMessage#blocks_and_attachments

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,143 @@
...
+        interval_in_seconds (int): Interval in seconds to send logs to slack. If None, message is sent after every log call.
+        max_logs_per_message (int)(default:50): Maximum number of logs to send in a single message. Note that no more than 50 items are allowed to send in a single message.
+        If more than 50 items are stored in buffer, the message flushed without waiting the full time interval.
+    """
+
+    def __init__(
+        self,
+        formatter_func: Optional[Callable[..., List[Dict[str, Any]]]] = None,
+        include_keys: Optional[Set[str]] = None,
+        interval_in_seconds: Optional[int] = None,
```
eracah, line 62: Should this be seconds? That's not the most intuitive thing for the user. We may want to use the Time abstraction. See how it is done in ConsoleLogger for reference: https://github.com/mosaicml/composer/blob/dev/composer/loggers/console_logger.py#L42

eracah (changes requested):
Hey Wai, looks pretty good! Thanks for addressing my feedback from before, but a couple more bits of feedback for you.
Also, can you post the code that output those example images (you can redact the API key of course)? Also, thanks for adding those unit tests!

composer/loggers/slack_logger.py
waiwuc: I can make that change, but not convinced about the use case for having these as parameters in the init function. These slack keys should be treated as secrets and if so, they are best set as env variables.

composer/loggers/slack_logger.py
waiwuc: Ill make the update, but curious why not a `Set`?

composer/loggers/slack_logger.py
waiwuc: if they set a high `max_logs_per_message`, we will still flush the logs when it hits 50 as that's the maximum that we can send to slack client. 

I am conflicted about having a full history for every key. Conceptually, if we wanted to track something like loss spikes, would we not need to have a full history of total loss at every step? On the other hand, we have also decided against version history for run metadata (although maybe that's more because of technical limitation vs desired product behavior) - cc @mvpatel2000 for thoughts here too. 

Given slack is high noise channel, maybe we'd optimize for small number of keys but larger number of values logged over time, rather than large number of keys and snapshots being logged?

composer/loggers/slack_logger.py
waiwuc: yes let me split up into a few variables. I needed to add `self.channel_id is not None` for typing reasons

composer/loggers/slack_logger.py
waiwuc: `When you're using blocks in your message payload, the top-level text field becomes a fallback message displayed in notifications. Blocks should define everything else about the desired visible message.` 

text is there for notifications actually. I didn't test this behavior manually, but I definitely should, and probably just updated to a shorter string like f"Value for {len(self.logs)} metrics updated"

composer/loggers/slack_logger.py
waiwuc: great call out, I like that better, let me update this

composer/loggers/slack_logger.py
waiwuc: Synced quickly with @eracah , we can think about adding `log_full_history: Optional[bool]=False` for future, just launch this first without version history.

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,219 @@
...
+        if channel_id is None:
+            raise TypeError
+
+        inx = 0
+        while inx < len(self.log_dict.keys()):
+            max_log_entries_dict = dict(itertools.islice(self.log_dict.items(), inx, inx + self.max_logs_per_message))
+            self._format_and_send_blocks_to_slack(max_log_entries_dict, **kwargs)
+            inx += self.max_logs_per_message
+
+        # max_log_entries_dict = dict(itertools.islice(self.log_dict.items(), 0, 1))
```
eracah, line 196: remove old comments

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Format slack messages through rich message layouts created using Slack Blocks Kit.
+        See documentation here: https://api.slack.com/messaging/composing/layouts.
+        """
+        channel_id = self.channel_id
+        if channel_id is None:
+            raise TypeError
+
+        inx = 0
+        while inx < len(self.log_dict.keys()):
```
eracah, line 191: this just sends max_logs_per_message blocks at a time, but it still sends all the messages?

waiwuc: if there's 1000 metrics in between batches, this will flush all 1000 metrics 50 at a time

tests/loggers/test_slack_logger.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+from torch.utils.data import DataLoader
+
+from composer.loggers.slack_logger import SlackLogger
+from composer.trainer import Trainer
+from tests.common import RandomClassificationDataset, SimpleModel
+
+
+@pytest.mark.parametrize('include_keys', [['loss*']])
+def test_slack_logger_metrics(include_keys: Set[str]):
```
eracah, line 18:
```suggestion
@pytest.mark.parametrize('include_keys', [['loss*']])
@pytest.mark.parametrize('interval', ['1ba', '1ep'])
def test_slack_logger_metrics(include_keys: Set[str], interval: str):
```

tests/loggers/test_slack_logger.py
```diff
@@ -0,0 +1,57 @@
...
+    slack_logger = SlackLogger(
+        formatter_func=(lambda data, **kwargs: [{
+            'type': 'section',
+            'text': {
+                'type': 'mrkdwn',
+                'text': f'*{k}:* {v}'
+            }
+        } for k, v in data.items()]),
+        include_keys=include_keys,
+        log_interval='1ep',
```
eracah, line 32:
```suggestion
        log_interval=interval,
```

tests/loggers/test_slack_logger.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    with patch('slack_sdk.WebClient.chat_postMessage') as mock_slack:
+        trainer = Trainer(
+            model=SimpleModel(),
+            train_dataloader=DataLoader(RandomClassificationDataset()),
+            train_subset_num_batches=2,
+            max_duration='1ep',
```
eracah, line 40:
```suggestion
            max_duration=interval,
```

tests/loggers/test_slack_logger.py
```diff
@@ -0,0 +1,57 @@
...
+            loggers=[slack_logger],
+        )
+
+    # Log interval is 1 epoch, so no logs should be sent before fit
```
eracah, line 44:
```suggestion
    # Log interval is 1 epoch or 1 batch, so no logs should be sent before fit
```

eracah (changes requested):

composer/loggers/slack_logger.py
```diff
@@ -0,0 +1,216 @@
...
+        """Default formatter function if no formatter func is specified.
+
+        This function will:
+        1. Log the key-value pairs in bold (key) and normal (value) text.
+        2. When logging metrics, set the step number as the header of the section.
+
+        Args:
+            data (Dict[str, Any]): Data to be logged.
+            **kwargs: Additional arguments to be passed to the formatter function
+            (Only support "header" argument now)
```
eracah, line 144:
```suggestion
            (Only supports "header" argument now)
```

eracah (approved):
LGTM! Thanks, @waiwuc great work!

mvpatel2000 (approved):

PR #2135, opened 2023-04-11
Ece icl

bmosaicml:
# What does this PR do?
This diff adds expected calibration error for MC and LM tasks (cite: https://arxiv.org/pdf/1706.04599.pdf)

ECE breaks each prediction into a bucket based on the model's confidence* then calculates the accuracy within each bucket, and sums over the average gap between accuracy and confidence across all buckets (weighted by number of samples per bucket).

*confidence is not straightforward to define and we had to make some design decisions. For LM tasks we define confidence as the minimum probability assigned to the target token amongst all tokens in the continuation. For multiple choice we followed the scheme in https://arxiv.org/abs/2012.00955 in which the model confidence is defined as the softmax of average per-token probability assigned to the top question choice.

Ran Lambada with our 7b hero run and got the following numbers:

```
metrics/lambada/0-shot/InContextLearningLMAccuracy: 0.5577519536018372
metrics/lambada/0-shot/InContextLearningLMExpectedCalibrationError: 0.1612596958875656
```

# Before submitting


bmosaicml:
@dakinggg  any thoughts on why after checking out dev's version of the files my commit is still bad?

dakinggg:
Just had to merge in dev :)

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+            correct_update[bucket_idx] = torch.tensor(1.0)
+            self.bucket_totals += correct_update
+
+    def compute(self):
+        assert isinstance(self.bucket_correct, Tensor)
+        assert isinstance(self.bucket_totals, Tensor)
+
+        result = torch.tensor(0.0)
+        total_obs = torch.sum(self.bucket_totals)
+        for i in range(self.n_buckets):
```
dakinggg, line 520: suggest breaking this down a bit to make it easier to read and clearly lined up with the formula from the paper
```
result = 0
for i in range(self.n_buckets):
    accuracy = self.bucket_correct[i].item() / self.bucket_totals[i].item()
    lower_bound = (i/self.n_buckets))
    upper_bound = ((i+1)/self.n_buckets))
    confidence = (lower_bound + upper_bound) / 2
    result += (self.bucket_totals[i]/total_obs)*abs(accuracy - confidence)
return torch.tensor(result)
    
```

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+        n_buckets (int): Number of distinct buckets to split the confidence distribution into
+    """
+
+    # Make torchmetrics call update only once
+    full_state_update = False
+
+    def __init__(self, dist_sync_on_step: bool = False, n_buckets=10):
+        # state from multiple processes
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.n_buckets = n_buckets
```
dakinggg, line 554: same

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+    Args:
+        dist_sync_on_step (bool, optional): Synchronize metric state across processes at
+            each forward() before returning the value at the step. Default: ``False``.
+        n_buckets (int): Number of distinct buckets to split the confidence distribution into
+    """
+
+    # Make torchmetrics call update only once
+    full_state_update = False
+
+    def __init__(self, dist_sync_on_step: bool = False, n_buckets=10):
```
dakinggg, line 551:
```suggestion
    def __init__(self, dist_sync_on_step: bool = False, n_buckets: int = 10):
```

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.n_buckets = n_buckets
+        self.add_state('bucket_totals', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
+        self.add_state('bucket_correct', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
+
+    def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+        output_logits = torch.softmax(output_logits, dim=2)
+        probabilites = []
+        for batch_idx, cont_idx in enumerate(batch['continuation_indices']):
+            cont_tok_logits = output_logits[batch_idx].index_select(dim=0, index=cont_idx - 1)
```
dakinggg, line 491: is there supposed to be a softmax in here to make the logits into probs?

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+        # state from multiple processes
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.n_buckets = n_buckets
+        self.add_state('bucket_totals', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
+        self.add_state('bucket_correct', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
+
+    def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+        output_logits = torch.softmax(output_logits, dim=2)
+        for batch_idx, cont_idx in enumerate(batch['continuation_indices']):
+            cont_tok_logits = output_logits[batch_idx].index_select(dim=0, index=cont_idx - 1)
```
dakinggg, line 561: is there supposed to be a softmax in here to make the logits into probs?

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+            cont_tok_pred = cont_tok_logits.argmax(dim=-1)
+            confidence = cont_tok_logits.max(dim=-1).values.min()
+            cont_tok_targ = labels[batch_idx].index_select(dim=0, index=cont_idx - 1)
+            assert confidence >= 0.0 and confidence <= 1.0
+            bucket_idx = int(confidence * self.n_buckets)
+            if bucket_idx == self.n_buckets:
+                bucket_idx -= 1
+
+            correct_update = torch.zeros(self.n_buckets)
```
dakinggg, line 570: same comments as above about simplifying this section

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+            correct_update[bucket_idx] = torch.tensor(1.0)
+            self.bucket_totals += correct_update
+
+    def compute(self):
+        assert isinstance(self.bucket_correct, Tensor)
+        assert isinstance(self.bucket_totals, Tensor)
+
+        result = 0.0
+        total_obs = torch.sum(self.bucket_totals)
+        for i in range(self.n_buckets):
```
dakinggg, line 584: same comments as above about simplifying this section. Also, since this computation is common to both implementations, can you pull it out into a helper?

dakinggg: and then please test that helper function separately

composer/metrics/nlp.py
```diff
@@ -431,6 +432,7 @@ def __init__(self, dist_sync_on_step: bool = False):
         self.add_state('total', default=torch.tensor(0.0), dist_reduce_fx='sum')
 
     def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+        output_logits = torch.softmax(output_logits, dim=2)
```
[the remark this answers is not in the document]
bmosaicml: I think HF forward is going to provide unnormalized logits so I need to use softmax in multiple choice in order to ensure we get real perplexities. The unnormalized logits of different MC answers won't be directly to comparable to one another b/c they could be scaled differently. I'm not 100% certain of the above logic, does it make sense to you?

composer/metrics/nlp.py
bmosaicml: ditto, at the beginning of the update function.

composer/metrics/nlp.py
bmosaicml: I softmax all the logits at once at the beginning of the function...lmk if that's what you had in mind.

tests/metrics/test_nlp_metrics.py
```diff
@@ -325,3 +354,49 @@ def test_in_context_learning_mc_accuracy(tiny_gpt2_tokenizer):
...
+
+    # for the last two, the correct answer is continuation 3
+    # make the answer incorrect by making continuation 2 more likely for both answers
+    start, end = cont_idxs[3].tolist()[0], cont_idxs[3].tolist()[-1]
+    logits[3][start:end] = logits[2][start:end].clone()
+
+    metric = InContextLearningMCExpectedCalibrationError()
+
+    metric.update(batch, logits, batch['labels'])
+    assert metric.compute().item() > 0
```
dakinggg, line 402: can we do the same thing as the other test to actually test getting a specific number out?

composer/metrics/nlp.py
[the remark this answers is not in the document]
dakinggg: i'm a bit confused because this line is in `InContextLearningMultipleChoiceAccuracy`, not the ECE class, so this doesn't have any impact on the ECE, right?

composer/metrics/nlp.py
dakinggg: ahh, I missed that line, thanks!

composer/metrics/nlp.py
dakinggg: I missed it twice :)

dakinggg (commented):
Could you please also do a real run with mosaicgpt and show some plots that look reasonable?

composer/metrics/nlp.py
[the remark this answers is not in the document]
dakinggg: oh, are you saying the old MC metric was actually wrong?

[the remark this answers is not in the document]
dakinggg: i believe that `F.cross_entropy` handles the normalization for you (https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html#torch.nn.functional.cross_entropy) "input ([Tensor](https://pytorch.org/docs/stable/tensors.html#torch.Tensor)) – Predicted unnormalized logits; see Shape section below for supported shapes."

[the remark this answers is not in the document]
bmosaicml: yeah I think you're right! good catch

composer/metrics/nlp.py
```diff
@@ -451,3 +453,137 @@ def compute(self):
...
+    """
+
+    # Make torchmetrics call update only once
+    full_state_update = False
+
+    def __init__(self, dist_sync_on_step: bool = False, n_buckets=10):
+        # state from multiple processes
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.n_buckets = n_buckets
+        self.add_state('bucket_totals', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
```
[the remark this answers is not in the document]
dakinggg: did you do this one?

composer/metrics/nlp.py
```diff
@@ -451,3 +462,118 @@ def compute(self):
...
         assert isinstance(self.total, Tensor)
         return self.correct.float() / self.total
+
+
+class InContextLearningExpectedCalibrationError(InContextLearningMetric):
+
+    def __init__(self, dist_sync_on_step: bool = False, n_buckets: int = 10):
+        # state from multiple processes
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.n_buckets = n_buckets
```
dakinggg, line 472: should we raise an error if `n_buckets < 1`?

composer/metrics/nlp.py
```diff
@@ -451,3 +462,118 @@ def compute(self):
         assert isinstance(self.correct, Tensor)
         assert isinstance(self.total, Tensor)
         return self.correct.float() / self.total
+
+
+class InContextLearningExpectedCalibrationError(InContextLearningMetric):
```
dakinggg, line 467: please add a test for this class. it can just be: 1) instantiate class 2) add some numbers to buckets manually 3) call compute and assert the correct number is gotten.

dakinggg (approved):
LGTM! Please paste in a plot from a real run so anyone looking back at this PR has something to compare to.

composer/metrics/nlp.py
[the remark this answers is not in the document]
bmosaicml: `dist_reduce_fx (Optional) – Function to reduce state across multiple processes in distributed mode. If value is "sum", "mean", "cat", "min" or "max" we will use torch.sum, torch.mean, torch.cat, torch.min and torch.max` respectively, each with argument dim=0. Note that the "cat" reduction only makes sense if the state is a list, and not a tensor. The user can also pass a custom function in this parameter.`

PR #2136, opened 2023-04-12
Mcloud logger dest

mvpatel2000:
# What does this PR do?

Logs metrics to MCLI
image

Along for ride:
- minor clean up in wandb test file

[CO-1985](https://mosaicml.atlassian.net/browse/CO-1985)

[CO-1985]: https://mosaicml.atlassian.net/browse/CO-1985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+                self.buffered_metadata[key] = format_log_data_value(val)
+            if time.time() - self.time_last_logged > self.log_interval:
+                from mcli.api.exceptions import MAPIException
+                from mcli.sdk import update_run_metadata
+                try:
+                    update_run_metadata(self.run_name, self.buffered_metadata)
+                except MAPIException as e:
+                    log.error(f'Failed to log metadata to Mosaic with error: {e}')
+
+                self.buffered_metadata = {}
```
aspfohl, line 76: to prevent data loss?
```suggestion
               else:
                 self.buffered_metadata = {}
```

composer/loggers/__init__.py
```diff
@@ -37,4 +38,5 @@
     'CometMLLogger',
     'ConsoleLogger',
     'MLFlowLogger',
+    'MosaicLogger',
```
dakinggg, line 41: `MosaicMLLogger`?

composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+
+log = logging.getLogger(__name__)
+
+__all__ = ['MosaicLogger']
+
+
+class MosaicLogger(LoggerDestination):
+    """Log to MPlatform.
+
+    Logs metrics to MPlatform. Logging only happens on rank 0 every ``log_interval`` seconds to
```
dakinggg, line 25: can logging only on rank 0 be optional? or is that an mcli limitation?

composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+from composer.loggers.logger import format_log_data_value
+from composer.loggers.logger_destination import LoggerDestination
+from composer.utils import MissingConditionalImportError, dist
+
+log = logging.getLogger(__name__)
+
+__all__ = ['MosaicLogger']
+
+
+class MosaicLogger(LoggerDestination):
```
dakinggg, line 22: can you add some flushes of the buffered metadata on like fit end or something? otherwise you'll end up with some data not logged probably

composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+            if time.time() - self.time_last_logged > self.log_interval:
+                from mcli.api.exceptions import MAPIException
+                from mcli.sdk import update_run_metadata
+                try:
+                    update_run_metadata(self.run_name, self.buffered_metadata)
+                except MAPIException as e:
+                    log.error(f'Failed to log metadata to Mosaic with error: {e}')
+
+                self.buffered_metadata = {}
+                self.time_last_logged = time.time()
```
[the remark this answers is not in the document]
eracah: wait we wouldn't want `time_last_logged` to be updated if it failed right?

composer/loggers/mosaic_logger.py
eracah, line 76: you might want to have the flush logic happen at `batch_end`, `epoch_end`, `fit_end`, etc and only have the adding to buffer happen inside `_log_metadata`. See `ConsoleLogger` code for inspo

composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+            self.buffered_metadata: Dict[str, Any] = {}
+
+            try:
+                import mcli
+                del mcli
+            except ImportError as e:
+                raise MissingConditionalImportError(extra_deps_group='mcli',
+                                                    conda_package='mcli',
+                                                    conda_channel='conda-forge') from e
+            self.run_name = os.environ.get('RUN_NAME')
```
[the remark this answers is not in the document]
mvpatel2000: I'll add it to the top. Going to skip constants file because we don't have many constants, and it's easier to read everything in a single self contained file, especially if users want to copy paste and add a variant callback

composer/loggers/mosaic_logger.py
mvpatel2000: MAPI limitation, otherwise we're going to start overwhelming it

composer/loggers/mosaic_logger.py
```diff
@@ -0,0 +1,77 @@
...
+    def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
+        self._log_metadata(hyperparameters)
+
+    def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        self._log_metadata(metrics)
+
+    def _log_metadata(self, metadata: Dict[str, Any]) -> None:
+        if self._enabled:
+            for key, val in metadata.items():
+                self.buffered_metadata[key] = format_log_data_value(val)
```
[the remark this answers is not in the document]
mvpatel2000: Tensors are not JSON serializable unfortunately. I can change it to keep in primitive types. For tensors which are not singletons, eg loss, we do 
>         return 'Tensor of shape ' + str(data.shape)

Does that work? any suggestions?

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,135 @@
...
+                            'is disabled as it is unable to identify which run to log to.')
+                self._enabled = False
+
+    def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
+        self._log_metadata(hyperparameters)
+
+    def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        self._log_metadata(metrics)
+
+    def batch_end(self, state: State, logger: Logger) -> None:
```
dakinggg, line 74: are you sure you want to flush on every batch? that doesn't limit the sending _that_ much

composer/loggers/mosaicml_logger.py
```diff
@@ -0,0 +1,135 @@
...
+    from composer.core import State
+
+log = logging.getLogger(__name__)
+
+__all__ = ['MosaicMLLogger']
+
+RUN_NAME_ENV_VAR = 'RUN_NAME'
+
+
+class MosaicMLLogger(LoggerDestination):
```
dakinggg, line 32: should we have a `filter_keys` or something (i.e. if you want to avoid sending optimizer monitor logs to mcli or something)

composer/loggers/mosaicml_logger.py
mvpatel2000: It calls the flush function, but it won't flush unless the interval has passed

composer/loggers/mosaicml_logger.py
mvpatel2000: Good call... yea let me add something + unit test it

composer/loggers/mosaicml_logger.py
dakinggg: oh right I kinda missed that. the intention of the flush is to _write even if the interval has not passed_. thats why  suggested running it at fit end (and maybe some other common checkpoints). Also would be good to flush on a crash if possible (i.e. in the close or post_close or whatever).

eracah (approved):
Nice. All my feedback was addressed. LGTM

eracah:
> Nice. All my feedback was addressed. LGTM

Oh wait. Tests?

composer/loggers/mosaic_logger.py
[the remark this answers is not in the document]
dakinggg: lets just stick with the current thing of just printing the shape?

tests/loggers/test_wandb_logger.py
```diff
@@ -15,12 +15,8 @@
 import torch
 from torch.utils.data import DataLoader
```
dakinggg, line 16: Were the changes in this file cleanup?

dakinggg (approved):
Approving to unblock, but would like some tests

tests/loggers/test_wandb_logger.py
mvpatel2000: Yes, just cleaning up some imports

composer/trainer/trainer.py
```diff
@@ -824,7 +824,7 @@ def __init__(
         # Save Checkpoint
         save_folder: Optional[str] = None,
         save_filename: str = 'ep{epoch}-ba{batch}-rank{rank}.pt',
-        save_latest_filename: Optional[str] = 'latest-rank{rank}.pt',
+        save_latest_filename: str = 'latest-rank{rank}.pt',
```
dakinggg, line 827: Wait, isn't this actually optional? If none, won't save latest, right?

mvpatel2000: oops... good save...

bandish-shah (approved):
LGTM

setup.py
```diff
@@ -206,7 +206,13 @@ def package_files(prefix: str, directory: str, extension: str):
     'onnxruntime>=1.12.1,<2',
 ]
 
-extra_deps['mlflow'] = ['mlflow>=2.0.1,<3.0']
+extra_deps['mlflow'] = [
+    'mlflow>=2.0.1,<3.0',
+]
+
+extra_deps['mcli'] = [
+    'mosaicml-cli>=0.3.6,<0.4',
```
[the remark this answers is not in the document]
mvpatel2000: I think there's a few semantic changes at 0.4.0, and I'd rather pin to be safe. I'll loosen after that change set and possibly make this a required dep since mcli will become a lot lighter.... I'd rather be safe for now

PR #2138, opened 2023-04-12
Adding sharded grad scaler

bcui19:
# What does this PR do?
Adds support for sharded grad scaler when using FSDP.

Manually testing it with a print statement:

image
Also to make sure we're not using closures. 

[CO-2009](https://mosaicml.atlassian.net/browse/CO-2009)

# Before submitting

[CO-2009]: https://mosaicml.atlassian.net/browse/CO-2009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1222,6 +1222,10 @@ class Trainer:
         warnings.filterwarnings(action='ignore', message='torch.cuda.amp.GradScaler')
         self.state.scaler = ClosureGradScaler() if self._use_closures() else GradScaler()
 
+        if self.fsdp_config is not None:
+            from torch.distributed.fsdp.sharded_grad_scaler import ShardedGradScaler
+            self.state.scaler = ShardedGradScaler()
+
         # suppressing FSDP warning when auto grad accum exits the forward pass before completing
         warnings.filterwarnings(action='ignore', message='Forward order differs from that of the first iteration')
 
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -25,6 +25,7 @@ import torch
 import torch.distributed
 import torch.nn as nn
 import torch.utils.data
+from packaging import version
 from torch.cuda.amp.grad_scaler import GradScaler, _refresh_per_optimizer_state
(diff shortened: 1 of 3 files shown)


composer/trainer/trainer.py
```diff
@@ -1222,6 +1222,10 @@ def __init__(
         warnings.filterwarnings(action='ignore', message='torch.cuda.amp.GradScaler')
         self.state.scaler = ClosureGradScaler() if self._use_closures() else GradScaler()
 
+        if self.fsdp_config is not None:
+            from torch.distributed.fsdp.sharded_grad_scaler import ShardedGradScaler
+            self.state.scaler = ShardedGradScaler()
+
```
mvpatel2000, line 1228: I believe the FSDP config needs to be gated. 

also can we confirm closures aren't used here (just raise ValueError if they are, though we should never reach this state)

composer/trainer/trainer.py
```diff
@@ -1222,6 +1223,17 @@ def __init__(
         warnings.filterwarnings(action='ignore', message='torch.cuda.amp.GradScaler')
         self.state.scaler = ClosureGradScaler() if self._use_closures() else GradScaler()
 
+        if self.fsdp_config is not None:
+            if version.parse(torch.__version__) < version.parse('1.13.0'):
+                raise RuntimeError('To use FSDP with Composer, you must use torch>=1.13.0.')
+            from torch.distributed.fsdp.sharded_grad_scaler import ShardedGradScaler
+
+            # This state should never really ever be hit, but just in case we should raise a ValueError
+            if self._use_closures() and self.state.precision == Precision.AMP_FP16:
```
mvpatel2000, line 1232:
```suggestion
            # This state should never be reached, but we raise a ValueError just in case
            if self._use_closures() and self.state.precision == Precision.AMP_FP16:
```

mvpatel2000 (approved):

mvpatel2000:
LGTM on hero.

PR #2146, opened 2023-04-17
Make sure the timestamps of the checkpoints are the same when loading

eracah:
# What does this PR do?
In the edge case where one is using sharded checkpointing + autoresume and one rank's save checkpoint operation fails, then that rank will resume from an older checkpoint than the other ranks and this will be a silent error. 

This PR fixes that by checking to make sure that all timestamps loaded in match. If they don't it errors out

### Manual Test
```image: mosaicml/pytorch:latest
integrations:
- git_branch: release/v0.0.4
  git_repo: mosaicml/examples
  integration_type: git_repo
  pip_install: .[llm]
- git_branch: ckpt-timestamp
  git_repo: eracah/evan-composer
  integration_type: git_repo
  pip_install: .[all]
parameters:
  load_path: s3://mosaicml-internal-checkpoints-test/evan-test/checkpoints/gpt_125m/sharded-ws-8/125m-ws-8-save-FYeuN0/ep0-ba1-rank{rank}.pt
  max_duration: 4ba
```
logs:
```
Starting training...
Config:
enabled_algorithms/GradientClipping: true
node_name: a100-80sxm-h10-03
num_gpus_per_node: 8
num_nodes: 1
rank_zero_seed: 17

[batch=2/4]:
         ...
[batch=3/4]:
         ....
[batch=4/4]:
         ...
```

fix CO-2017

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -492,15 +492,15 @@ def _restore_checkpoint(
         dist.all_reduce(max_step_to_resume_from, reduce_operation='MAX')
         dist.all_reduce(min_step_to_resume_from, reduce_operation='MIN')
         if max_step_to_resume_from.data != min_step_to_resume_from.data:
-            raise RuntimeError(textwrap.dedent(
-                f'Timestamp mismatch error: batch to resume from {step_to_resume_from} is not the same on all ranks. '
-                'This usually occurs when at least one rank fails to save the last checkpoint '
-                'while using sharded checkpointing + autoresume. '
-                'Please manually resume by disabling autoresume and explicitly setting load_path '
-                'to the most recent checkpoints that all ranks have saved. '
-                'E.g. for the 10th batch: trainer = Trainer(autoresume=False, load_path="/path/to/checkpoint/ba10-rank{rank}.pt", ...). '
-                'Remember to keep the {rank} placeholder!' 
-            ))
+            raise RuntimeError(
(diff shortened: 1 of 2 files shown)


composer/utils/checkpoint.py
```diff
@@ -484,6 +484,20 @@ def _restore_checkpoint(
             exclude_algorithms=exclude_algorithms,
             algorithm_passes=algorithm_passes,
         )
+        step_to_resume_from = state.timestamp.batch.value
+        avg_step_to_resume_from = state.device.tensor_to_device(
```
dakinggg, line 488: Any particular reason we're doing this with floats? the timestamp should be an int and then we don't have any risk of floating point error messing things up

eracah: oh I mainly did it to avoid rounding error when averaging.

eracah: ok changed to min-max comparison

dakinggg (approved):

dakinggg:
Could you do a manual test too please? Preferable multi node, since we don't have unit test coverage for that. Would like to see a normal resumption succeed.

eracah:
> Could you do a manual test too please? Preferable multi node, since we don't have unit test coverage for that. Would like to see a normal resumption succeed.

Ok did a standard resumption of a sharded checkpoint and added the setup and logs to the PR description

PR #2149, opened 2023-04-19
Add support for `eval_interval` and `save_interval` in tokens

dakinggg:
# What does this PR do?
Adds support for specifying `eval_interval` and `save_interval` in tokens, as well as the combination of `max_duration` in tokens and `eval_interval` in `dur`.

Manual test:

```
GLOBAL_BATCH_SIZE = 512
MAX_SEQ_LEN = 2048
DURATION_INT = 104_857_600
MAX_DURATION = f'{DURATION_INT}tok'
EVAL_INT = 12_582_912
EVAL_INTERVAL = f'{EVAL_INT}tok'
SAVE_INT = 41_943_041
SAVE_INTERVAL = f'{SAVE_INT}tok'
```

Screen Shot 2023-04-18 at 10 47 29 PM
Screen Shot 2023-04-18 at 10 47 41 PM

TODO:
- [x] update docs
- [x] manual test

Closes [CO-1877](https://mosaicml.atlassian.net/browse/CO-1877)

# Before submitting

[CO-1877]: https://mosaicml.atlassian.net/browse/CO-1877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -72,6 +72,21 @@ def checkpoint_periodically(interval: Union[str, int, Time],
             event: Event,
             will_actually_checkpoint: bool = True,  # determines whether to update the last_value_checkpointed state
     ):
+        """The save interval function that determines whether to save a checkpoint.
+
+        Args:
+            state (State): The current :class:`.State` object
+            event (Event): The :class:`.Event` that has is being checked for checkpointing
+            will_actually_checkpoint (bool, optional): Whether or not the call to this function will actually write a checkpoint.
+                This determines whether or not to update the `last_value_checkpointed` state that keeps track of when the last checkpointing occurred.
+                Defaults to True.
+
+        Raises:
+            NotImplementedError: _description_
+
+        Returns:
+            _type_: _description_
+        """
         nonlocal last_value_checkpointed
 
         elapsed_duration = state.get_elapsed_duration()
@@ -350,6 +365,7 @@ class CheckpointSaver(Callback):  # noqa: D101
         os.makedirs(folder, exist_ok=True)
 
     def after_load(self, state: State, logger: Logger) -> None:
+        # This is done here to set the starting timestamp when we are resuming
         if not callable(self.save_interval):
             self.save_interval = checkpoint_periodically(self.save_interval, state.timestamp)
 
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -51,7 +51,7 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
         raise NotImplementedError(
             f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH or TimeUnit.BATCH.')
 
-    last_value_checkpointed = -1
+    last_value_checkpointed = 0
 
     def save_interval(state: State, event: Event):
         elapsed_duration = state.get_elapsed_duration()
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -6,13 +6,14 @@
 from __future__ import annotations
 
 import logging
+import math
 import os
 import pathlib
 import tempfile
 import textwrap
-from typing import List, Optional, Protocol, Union
+from typing import Callable, List, Optional, Union
 
-from composer.core import Callback, Event, State, Time, Timestamp, TimeUnit
+from composer.core import Callback, Event, State, Time, TimeUnit
 from composer.loggers import Logger
 from composer.utils import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, FORMAT_NAME_WITH_DIST_TABLE, PartialFilePath,
                             checkpoint, create_symlink_file, dist, ensure_folder_has_no_conflicting_files,
@@ -23,14 +24,7 @@ log = logging.getLogger(__name__)
 __all__ = ['CheckpointSaver', 'checkpoint_periodically']
 
 
-class SaveIntervalCallable(Protocol):
-
-    def __call__(self, state: State, event: Event, will_actually_checkpoint: bool) -> bool:
-        ...
-
-
-def checkpoint_periodically(interval: Union[str, int, Time],
-                            starting_timestamp: Optional[Timestamp] = None) -> SaveIntervalCallable:
+def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State, Event], bool]:
     r"""Helper function to create a checkpoint scheduler according to a specified interval.
 
     Args:
@@ -41,9 +35,6 @@ def checkpoint_periodically(interval: Union[str, int, Time],
 
             Checkpoints will be saved every ``n`` batches or epochs (depending on the unit),
             and at the end of training.
-        starting_timestamp (optional, :class:`.Timestamp`): The timestamp at which to initialize the save interval.
-            This will just be a timestamp with value zero unless resuming, in which case it will be the timestamp
-            that resumption is occurring from. If ``None``, will default to value zero. (default: ``None``)
 
     Returns:
         Callable[[State, Event], bool]: A function that can be passed as the ``save_interval``
@@ -63,32 +54,7 @@ def checkpoint_periodically(interval: Union[str, int, Time],
             f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
         )
 
-    last_value_checkpointed = 0
-    if starting_timestamp is not None:
-        last_value_checkpointed = starting_timestamp.get(interval.unit)
-
-    def save_interval(
-            state: State,
-            event: Event,
-            will_actually_checkpoint: bool = True,  # determines whether to update the last_value_checkpointed state
-    ):
-        """The save interval function that determines whether to save a checkpoint.
-
-        Args:
-            state (State): The current :class:`.State` object
-            event (Event): The :class:`.Event` that has is being checked for checkpointing
-            will_actually_checkpoint (bool, optional): Whether or not the call to this function will actually write a checkpoint.
-                This determines whether or not to update the `last_value_checkpointed` state that keeps track of when the last checkpointing occurred.
-                Defaults to True.
-
-        Raises:
-            NotImplementedError: _description_
-
-        Returns:
-            _type_: _description_
-        """
-        nonlocal last_value_checkpointed
-
+    def save_interval(state: State, event: Event):
         elapsed_duration = state.get_elapsed_duration()
         assert elapsed_duration is not None, 'elapsed_duration is set on the BATCH_CHECKPOINT and EPOCH_CHECKPOINT'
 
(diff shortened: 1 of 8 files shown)


composer/callbacks/checkpoint_saver.py
```diff
@@ -319,6 +349,10 @@ def init(self, state: State, logger: Logger) -> None:
         folder = format_name_with_dist(self.folder, state.run_name)
         os.makedirs(folder, exist_ok=True)
 
+    def after_load(self, state: State, logger: Logger) -> None:
+        if not callable(self.save_interval):
+            self.save_interval = checkpoint_periodically(self.save_interval, state.timestamp)
+
```
mvpatel2000, line 355: ~Why are we moving this?~ Is this to ensure the starting timestamp is set? Can we add a comment if so?

composer/callbacks/checkpoint_saver.py
```diff
@@ -45,28 +56,47 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
...
         save_event = Event.BATCH_CHECKPOINT
     else:
         raise NotImplementedError(
-            f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH or TimeUnit.BATCH.')
+            f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
+        )
+
+    last_value_checkpointed = 0
+    if starting_timestamp is not None:
+        last_value_checkpointed = starting_timestamp.get(interval.unit)
```
mvpatel2000, line 68: Is there a simpler way we can propagate this information besides adding something to state and the logic in checkpointing? What if we just call `checkpoint_periodically` but don't `will_actually_checkpoint=False` on after_load (and use timestamp instead?)

composer/callbacks/checkpoint_saver.py
```diff
@@ -45,28 +56,47 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
...
+
+    def save_interval(
+            state: State,
+            event: Event,
+            will_actually_checkpoint: bool = True,  # determines whether to update the last_value_checkpointed state
```
mvpatel2000, line 73: can we use doc strings instead of in-line comments

mvpatel2000 (commented):
I don't really like the timestamp hack and how resumption affects eval interval. Can we instead store the last_x attributes in state dict for checkpoint_saver?

composer/callbacks/checkpoint_saver.py
dakinggg: hmm i think that works, thanks, let me test it.

dakinggg: wait no, the addition to state was for eval, checkpointing doesn't need it. `will_actually_checkpoint` was added because EMA calls this funciton but then doesn't write a checkpoint

composer/callbacks/checkpoint_saver.py
dakinggg: yeah, will add a comment

tests/trainer/test_trainer_eval.py
```diff
@@ -256,6 +397,7 @@ def test_eval_params_init(
     assert event_counter_callback.event_to_num_calls[Event.EVAL_BATCH_START] == 1
 
 
+# TODO: token test
```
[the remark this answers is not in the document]
dakinggg: oops, removed.

dakinggg (commented):
@mvpatel2000 rereview please. went with the previous timestamp approach

composer/callbacks/checkpoint_saver.py
```diff
@@ -59,14 +62,17 @@ def save_interval(state: State, event: Event):
...
+            raise NotImplementedError(
+                f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
+            )
 
-        if event == save_event and int(count) % int(interval) == 0:
+        threshold_passed = math.floor(previous_count / interval.value) != math.floor(count / interval.value)
+        if event == save_event and threshold_passed:
             return True
 
         return False
```
mvpatel2000, line 78:
```suggestion
        return event == save_event and threshold_passed:
```

composer/callbacks/checkpoint_saver.py
```diff
@@ -45,11 +47,12 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
...
     if interval.unit == TimeUnit.EPOCH:
         save_event = Event.EPOCH_CHECKPOINT
-    elif interval.unit == TimeUnit.BATCH:
+    elif interval.unit in {TimeUnit.BATCH, TimeUnit.TOKEN, TimeUnit.SAMPLE}:
         save_event = Event.BATCH_CHECKPOINT
     else:
         raise NotImplementedError(
-            f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH or TimeUnit.BATCH.')
+            f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
+        )
```
mvpatel2000, line 55: Why not duration?

composer/callbacks/checkpoint_saver.py
```diff
@@ -59,14 +62,17 @@ def save_interval(state: State, event: Event):
...
             return True
 
-        if save_event == Event.EPOCH_CHECKPOINT:
-            count = state.timestamp.epoch
-        elif save_event == Event.BATCH_CHECKPOINT:
-            count = state.timestamp.batch
+        assert state.previous_timestamp is not None
+        if interval.unit in {TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, TimeUnit.SAMPLE}:
+            previous_count = state.previous_timestamp.get(interval.unit)
+            count = state.timestamp.get(interval.unit)
```
mvpatel2000, line 68: Same as above, why not duration?

composer/callbacks/checkpoint_saver.py
```diff
@@ -59,14 +62,17 @@ def save_interval(state: State, event: Event):
         if elapsed_duration >= 1.0:
             return True
 
-        if save_event == Event.EPOCH_CHECKPOINT:
-            count = state.timestamp.epoch
-        elif save_event == Event.BATCH_CHECKPOINT:
-            count = state.timestamp.batch
+        assert state.previous_timestamp is not None
```
mvpatel2000, line 65: Return false if None? Otherwise, if algo calls before it's set, you get nasty trace

composer/core/evaluator.py
```diff
@@ -51,17 +48,25 @@ def should_eval(state: State, event: Event):
         if eval_at_fit_end and event == Event.FIT_END and state.timestamp.batch != last_batch_seen:
             return True
 
-        if eval_interval.unit == TimeUnit.EPOCH and int(
-                state.timestamp.epoch) % int(eval_interval) == 0 and event == Event.EPOCH_END:
+        assert state.previous_timestamp is not None
```
mvpatel2000, line 51: Return false if None? Otherwise, if algo calls before it's set, you get nasty trace

composer/core/evaluator.py
```diff
@@ -51,17 +48,25 @@ def should_eval(state: State, event: Event):
...
+        if eval_interval.unit in {TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, TimeUnit.SAMPLE}:
+            previous_count = state.previous_timestamp.get(eval_interval.unit)
+            count = state.timestamp.get(eval_interval.unit)
+        else:
+            assert state.max_duration is not None
+            previous_count = state.previous_timestamp.get(state.max_duration.unit)
+            count = state.timestamp.get(state.max_duration.unit)
```
mvpatel2000, line 58: What is this else doing? I'ma bit confused

composer/core/evaluator.py
```diff
@@ -51,17 +48,25 @@ def should_eval(state: State, event: Event):
...
+
+        threshold_passed = math.floor(previous_count / eval_interval.value) != math.floor(count / eval_interval.value)
+
+        if eval_interval.unit == TimeUnit.EPOCH and event == Event.EPOCH_END and threshold_passed:
```
mvpatel2000, line 62: delete last_batch_seen? because of prev timestamp

composer/core/evaluator.py
```diff
@@ -85,7 +90,12 @@ def should_eval(state: State, event: Event):
                         last_batch_seen = state.timestamp.batch
                         return True
             elif state.max_duration.unit == TimeUnit.TOKEN and event == Event.BATCH_END:
-                raise ValueError(f'Evaluation interval of type `dur` is not supported yet for max_duration as `tok`')
+                tokens_per_interval = math.ceil(state.max_duration.value * eval_interval)
```
mvpatel2000, line 93: can we simplify the sample logic above to also be like this?

composer/core/state.py
```diff
@@ -416,6 +416,7 @@ def __init__(
         self._train_dataloader = train_dataloader
         self._evaluators = list(ensure_tuple(evaluators))
 
+        self.previous_timestamp: Optional[Timestamp] = None
```
mvpatel2000, line 419: Thoughts on instead starting it as `timestamp` instead of `None`? Does this make anything easier?

mvpatel2000 (commented):
Design looks good. Mostly cleanup comments and a few missing edge cases?

composer/callbacks/checkpoint_saver.py
dakinggg: well, it wasn't implemented before. We could use the same logic as for eval, but that seems like a separate PR?

composer/callbacks/checkpoint_saver.py
dakinggg: same answer

composer/core/evaluator.py
dakinggg: same answer, will do though.

composer/core/evaluator.py
dakinggg: if your eval interval unit is one of the incrementing ones, you track progress in that unit. If your eval interval unit is duration, you track progress with the unit of the max duration instead.

dakinggg: can add comment explaining

composer/core/evaluator.py
dakinggg: this logic was here before and is slightly different. It basically says "Checkpoint at fit end if you didn't already checkpoint that batch". but the prev timestamp doesn't track when you last checkpointed, so doesn't cover this. i'm sure there is a nicer way to do this, but i think out of scope of this pr.

composer/core/evaluator.py
dakinggg: we _can_, but the logic is slightly different so it would be a behavior change. wdyt?

composer/core/evaluator.py
dakinggg: added a comment, please unresolve if you think the comment is unclear or the logic is wrong.

composer/core/evaluator.py
mvpatel2000: I'd personally rather have consistent behavior

composer/callbacks/checkpoint_saver.py
mvpatel2000: Hm... I feel like we should use the same logic. I'm fine doing in separate PR

dakinggg:
@mvpatel2000 ready for rereview

mvpatel2000 (approved):

PR #2155, opened 2023-04-21
Log Time Attrs

mvpatel2000:
# What does this PR do?

Logs time attributes so they can easily be used as x-axis. Also groups all time things.

Currently, if the max duration is specified in batches and does not cleanly divide the number of epochs, the unlogged metrics are not discarded (unlike the case where it does divide cleanly, in which case the `epoch_end` event discards the data). This PR also updates console logger to discard any unlogged metrics at `fit_end` to handle this case.

image

# Issue:

[CO-1659](https://mosaicml.atlassian.net/browse/CO-1569)

[CO-1659]: https://mosaicml.atlassian.net/browse/CO-1659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/trainer/trainer.py
```diff
@@ -1912,10 +1912,17 @@ def _train_loop(self) -> None:
...
+                    # Log time values
                     self.logger.log_metrics({
-                        'trainer/global_step': int(self.state.timestamp.batch),
-                        'trainer/batch_idx': self.state.timestamp.batch_in_epoch.value,
+                        'time/batch': self.state.timestamp.batch.value,
+                        'time/sample': self.state.timestamp.sample.value,
+                        'time/batch_in_epoch': self.state.timestamp.batch_in_epoch.value,
+                        'time/sample_in_epoch': self.state.timestamp.sample_in_epoch.value,
                     })
+                    if rank_num_tokens > 0:
```
dakinggg, line 1923: did you want this condition to also be based on previous timestamp?

composer/trainer/trainer.py
```diff
@@ -1912,10 +1912,17 @@ def _train_loop(self) -> None:
...
 
                     self.engine.run_event(Event.BATCH_START)
+
+                    # Log time values
                     self.logger.log_metrics({
-                        'trainer/global_step': int(self.state.timestamp.batch),
-                        'trainer/batch_idx': self.state.timestamp.batch_in_epoch.value,
+                        'time/batch': self.state.timestamp.batch.value,
+                        'time/sample': self.state.timestamp.sample.value,
+                        'time/batch_in_epoch': self.state.timestamp.batch_in_epoch.value,
```
dakinggg, line 1920: decided to keep the "in epoch" ones?

composer/trainer/trainer.py
mvpatel2000: It messes up on first batch otherwise, so i'd prefer to use rank_num_tokens

composer/trainer/trainer.py
mvpatel2000: I realized batch_in_epoch is pretty important for vision runs

dakinggg (approved):

PR #2178, opened 2023-04-28
Fix ICL eval for sentencepiece tokenizers

dakinggg:
# What does this PR do?
This PR fixes the setting of a prefix space for the continuation for ICL eval. some tokenizers don't want this and some do. I am not aware of a property to check directly, so instead we run a simple string through ` a` and see what happens.

### DIFF
--- a/composer/datasets/in_context_learning_evaluation.py
+++ b/composer/datasets/in_context_learning_evaluation.py
@@ -296,8 +296,6 @@ class InContextLearningLMTaskDataset(Dataset):
         self.max_seq_len = max_seq_len
         self.pad_tok_id = pad_tok_id
         fewshot_rng = random.Random(fewshot_random_seed)
-        self.encoded_dataset = self.prep_examples(num_fewshot, prompt_string, example_delimiter, continuation_delimiter,
-                                                  fewshot_rng)
 
         # Test for whether a prefix space is needed before the continuation
         # sentencepiece tokenization should not have a prefix space, but gpt2 style BPE should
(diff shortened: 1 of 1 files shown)


bmosaicml (approved):

bmosaicml:
LGTM, I think the logic makes sense. Can you confirm that the perf on e.g. Lambada is reasonable?

samhavens:
some reference numbers:

opt-6.7b:
lambada_openai: 0.6771
arc_easy: 0.5926
hellaswag: 0.6699

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -296,6 +296,14 @@ def __init__(
         self.max_seq_len = max_seq_len
         self.pad_tok_id = pad_tok_id
         fewshot_rng = random.Random(fewshot_random_seed)
+
+        # Test for whether a prefix space is needed before the continuation
+        # sentencepiece tokenization should not have a prefix space, but gpt2 style BPE should
+        self.prefix_space = True
+        test = self.tokenizer(' a', add_special_tokens=False)['input_ids']
+        if len(test) > 1:
+            self.prefix_space = False
```
samhavens, line 305: should this be a `tokenizer_needs_prefix_space` function or something? I could see us needing this more, and it already shows up twice. Not a hard requirement

dakinggg: yeah, very reasonable, will change

samhavens (approved):

bmosaicml:
> 

Here's what I got without this change:

arc easy: opt-6.7b (0.578), cerebras-6.7b (0.564)
boolqmc: opt-6.7b (0.664), cerebras-6.7b (0.624)
lambada_openai: opt-6.7b (0.6767), cerebras-6.7b (0.6352)
hellaswag: opt-6.7b (0.676), cerebras-6.7b (0.5817)

samhavens:
Important context from @dakinggg on Slack:

> i was getting horrible results from llama before this change, including the 0s on lambada
> the behavior here is that LM tasks go to 0 because of the extra token, and MC tasks just get worse because there is an extra token in the middle that the model doesn’t expect
> and after this change llama gets reasonable scores for both

mvpatel2000 (approved):

PR #2181, opened 2023-04-29
Another attempt to fix the daily tests

dakinggg:
# What does this PR do?
1. Adds the `s3_bucket` config option to the GPU tests
2. Adjusts the precision for the old torchmetrics checkpoint test
3. Adds an attr to the loaded metric that needs to be present for torch 2
4. Special cases an equivalence tests for fused adam, because the loaded checkpoint has `False` and the CPU trained checkpoint has `None`
5. Ignores a few types so that precommit passes locally

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -8,6 +8,7 @@ import collections.abc
 import logging
 import textwrap
 import warnings
+from collections import OrderedDict
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Sequence, Union, cast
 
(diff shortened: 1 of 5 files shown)


composer/core/state.py
```diff
@@ -1118,6 +1118,8 @@ def load_state_dict(
                     if isinstance(serialized_value[metric_name], Metric):
                         # For checkpoints saved using Composer <= 0.13.5
                         serialized_value[metric_name].persistent(mode=True)
+                        # Add new attr in torch2
+                        serialized_value[metric_name]._state_dict_pre_hooks = OrderedDict()
```
mvpatel2000, line 1122: You can still use torch 2 and hit this on newer Composer. We should check if this exists or not on both sides and add conditionally?

dakinggg: I don't think you can actually? Prior to 0.14 (also meaning on torch 1.13.1 or earlier), we save torchmetrics full objects in the state dict. This code path can only be hit if a torchmetrics full object is in the state dict. The issue only occurs if you have loaded a full object, rather than just using the state dict. From 0.14 onwards, we will only save torchmetrics state dict, so it doesn't matter what torch version you are using.

dakinggg: lmk if that makes sense

mvpatel2000 (approved):

PR #2183, opened 2023-04-30
Fix space continuation issue for few shot ICL

dakinggg:
# What does this PR do?
This PR fixes an issue with few shot ICL eval, where the continuation delimiter was accidentally being rstripped both for the few shot examples, and for the test examples, when it was supposed to just be for the test example.

Wandb link with evals off of this branch (still populating at time of writing, but enough is there to verify expected outcome): https://wandb.ai/mosaic-ml/daniel-debug/groups/cont-space-evals-2/workspace?workspace=user-danielking

Expected results are that:
1. all 0 shot scores are unaffected from before this PR
2. MC few shot scores are mildly affected (probably made slightly better)
3. LM few shot scores are dramatically affected (going from 0 to non zero, should only be jeopardy)


bmosaicml (approved):
Do we need to modify all our datasets to move the space into the continuation like you did here?

dakinggg:
discussed with @bmosaicml offline, no need to change our datasets.

mvpatel2000 (approved):
trusting this is fine

PR #2184, opened 2023-04-30
Fix filelock in checkpoint download

mvpatel2000:
# What does this PR do?

Fix filelock in checkpoint download. Context manager was not used. Additionally, reword comment. 

Also changes trainer default to match `dist.initialize_dist` default

Faillure before:

image

After, successfully passes download (errors later on because keys mismatch since JF was downloading an unrelated ckpt, but whatever)
image


composer/utils/checkpoint.py
```diff
@@ -321,23 +321,20 @@ def download_checkpoint(path: str,
...
         signal_file_path = os.path.join(node_checkpoint_folder, '.local_rank0_completed')
         if dist.get_local_rank() == 0:
             with open(signal_file_path, 'wb') as f:
                 f.write(b'local_rank0_completed')
-        dist.local_rank_zero_download_and_wait(signal_file_path)
+
+        # Avoid the collective call until the local rank zero has finished trying to download the
+        # checkpoint so that we don't timeout for large downloads. This syncs all processes on the
+        # node
+        with dist.local_rank_zero_download_and_wait(signal_file_path):
```
dakinggg, line 332: slacked you as well, but basically, when using sharded checkpointing, `node_checkpoint_folder` is different for every rank, so they just wait here forever. you'll need to sync the local rank 0 path, either as an arg to this function, or at the top of this function.

composer/utils/checkpoint.py
```diff
@@ -321,23 +319,20 @@ def download_checkpoint(path: str,
                     pass
 
     finally:
-        # Wait for all checkpoints on the node to finish downloading
```
eracah, line 324: why is this comment removed? can you add a shorter one describing what we're doing here

composer/utils/checkpoint.py
eracah: If it uses sharded checkpointing, it really doesn't need to go down this codepath anyway because a non-zero rank is equally as likely to be the slow one as rank 0

eracah: Maybe we should move this logic up to the `if not checkpoint_is_sharded and dist.get_local_rank() == 0:` if block

eracah (commented):
I don't think this PR requires the checkpoint_folder change. See comments

composer/utils/checkpoint.py
mvpatel2000: Yep, I moved it closer to the logic

eracah (approved):
Manual shard test passed

PR #2197, opened 2023-05-04
Update FFCV

mvpatel2000:
# What does this PR do?

We use an old version of FFCV. With their latest versions, they loosen pins (eg for numpy/numba) which lets us avoid some nasty issues. Otherwise, if we upgrade numpy on vision / composer images, everything breaks. We've seen a few customers hit issues related to this upgrade.

Note: We don't add ffcv as a target because then [all] fails on CPU instances (eg no opencv) and it's not worth the headache

[CO-2069](https://mosaicml.atlassian.net/browse/CO-2069)

[CO-2069]: https://mosaicml.atlassian.net/browse/CO-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dakinggg (approved):
Leaving it up to you whether this is worth it or not cause im not familiar with ffcv, but is there anything that you would like to manually test?

mvpatel2000:
> Leaving it up to you whether this is worth it or not cause im not familiar with ffcv, but is there anything that you would like to manually test?

I'm happy with unit tests. We don't have a manual test for this since we don't use it actievly.

Issue #2168, opened 2023-04-25
Broken cod


eldarkurtic:
Training loop image is missing from the docs page https://docs.mosaicml.com/projects/composer/en/latest/examples/custom_speedup_methods.html: ![image](https://user-images.githubusercontent.com/8884008/234281217-d99debdc-5e3b-410e-ae99-5a4e65afd570.png)

(it is a very important reference point when implementing custom algorithms in composer)

eracah:
Should be fixed now, @eldarkurtic !

PR #2199, opened 2023-05-04
Fix Broken Training Loop Image Link

eracah:
# What does this PR do?
Fixes broken image link with composer training loop
Also, fixes broken markdown headings

Fix CO-2071
Fix https://github.com/mosaicml/composer/issues/2168

### DIFF
--- a/examples/custom_speedup_methods.ipynb
+++ b/examples/custom_speedup_methods.ipynb
@@ -366,20 +366,23 @@
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


mvpatel2000 (approved):
Header for "Implementing a Method with PyTorch" still needs to be fixed as well

eracah:
> Header for "Implementing a Method with PyTorch" still needs to be fixed as well

fixed!

PR #2203, opened 2023-05-05
Pin protobuf

mvpatel2000:
# What does this PR do?

Pins protobuf, which is required by sentencepiece, or it will break on some tokenizers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -182,7 +182,10 @@ extra_deps['nlp'] = [
     'datasets>=2.4,<3',
 ]
 
-extra_deps['sentencepiece'] = ['sentencepiece==0.1.98']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -182,7 +182,10 @@ def package_files(prefix: str, directory: str, extension: str):
     'datasets>=2.4,<3',
 ]
 
-extra_deps['sentencepiece'] = ['sentencepiece==0.1.98']
+extra_deps['sentencepiece'] = [
+    'protobuf<=3.20.0',
```
dakinggg, line 186:
```suggestion
    'protobuf<=3.21',
```

dakinggg (approved):

dakinggg (approved):

PR #2205, opened 2023-05-05
Move sharded checkpoints into their own intermediate prefix folder

eracah:
# What does this PR do?

fix CO-1990


mvpatel2000 (commented):
First pass.

composer/utils/checkpoint.py
```diff
@@ -75,9 +75,20 @@ def __init__(self, filename: str, folder: Optional[str] = None):
...
+                self.filename,
+            ) + extra_suffix
+        else:
+            return self.filename + extra_suffix
+
+    def format(self, state: State, is_deepspeed: bool = False, keep_placeholders: bool = False) -> str:
         # if filename already has a suffix (e.g. file.pt), this would append to be file.pt.tar
         extra_suffix = '.tar' if is_deepspeed and not is_tar(self.filename) else ''
+        if keep_placeholders:
+            return self._format_with_placeholders(extra_suffix=extra_suffix)
```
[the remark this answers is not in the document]
dakinggg: agree

eracah:
Ok, @dakinggg, @mvpatel2000, I changed the PR so the user configures what they want the prefix folder to be. PTAL

eracah:
> Ok, @dakinggg, @mvpatel2000, I changed the PR so the user configures what they want the prefix folder to be. PTAL

Also, updated FSDP docs to reflect this change

composer/callbacks/checkpoint_saver.py
```diff
@@ -318,11 +319,13 @@ def __init__(
         self.overwrite = overwrite
         self.saved_checkpoints: List[str] = []
         self.num_checkpoints_to_keep = num_checkpoints_to_keep
+        self.fsdp_sharded_state_dict_enabled: Optional[bool] = None
```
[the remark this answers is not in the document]
mvpatel2000: Actually, why is this necessary? Can't we always read from state?

composer/callbacks/checkpoint_saver.py
```diff
@@ -318,11 +318,13 @@ def __init__(
...
         self.weights_only = weights_only
 
         self.start_batch = None
 
     def init(self, state: State, logger: Logger) -> None:
+        self.fsdp_sharded_state_dict_enabled = state.fsdp_sharded_state_dict_enabled
```
mvpatel2000, line 327: Why do we need this on self instead of reading from state

composer/callbacks/checkpoint_saver.py
```diff
@@ -379,11 +381,11 @@ def _save_checkpoint(self, state: State, logger: Logger):
             raise ValueError(f'Save filename {self.filename.filename} must have {{rank}} for deepspeed.')
 
         # save the checkpoint to the filename
-        filename = self.filename.format(state, is_deepspeed)
+        filename_with_placeholders = self.filename.format(state, is_deepspeed, keep_placeholders=True)
```
mvpatel2000, line 384: Why does the vanilla case also keep placeholders?

composer/trainer/dist_strategy.py
```diff
@@ -424,6 +425,7 @@ def _check_fn(module: torch.nn.Module) -> bool:
         print(f'FSDP: Using forward_prefetch={forward_prefetch}')
         print(f'FSDP: Using limit_all_gathers={limit_all_gathers}')
         print(f'FSDP: Using state_dict_type={state_dict_type}')
+        print(f'FSDP: Using sharded_ckpt_prefix_dir={sharded_ckpt_prefix_dir}')
```
mvpatel2000, line 428: Can we change all these to `log.info`?

composer/utils/checkpoint.py
```diff
@@ -75,20 +75,25 @@ def __init__(self, filename: str, folder: Optional[str] = None):
...
+            folder += extra_suffix
         else:
-            return format_name_with_dist_and_time(
+            folder = format_name_with_dist_and_time(
                 self.filename,
                 state.run_name,
                 state.timestamp,
-            ) + extra_suffix
+            ) + extra_suffix if not keep_placeholders else self.filename + extra_suffix
+        return folder
```
mvpatel2000, line 96: Can we rework to do:

```
if self.folder
  if keep_placeholders:
    return ...
  else:
    return ..
else:
  if keep_placeholders:
    return ..
  else:
    return ...
```
imo multi-line ternaries are nasty to ready.

composer/callbacks/checkpoint_saver.py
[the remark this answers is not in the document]
eracah: oh you're right. it must have been an artifact from the big PR that I am chunking up

composer/trainer/dist_strategy.py
eracah: ooooh unfortunately that is outside the scope of this PR

composer/callbacks/checkpoint_saver.py
eracah: because that's what it should do anyway. `checkpoint.save_checkpoint` expects placeholders

eracah:
alright @mvpatel2000, I thiiiink I have addressed all the feedback. PTA(nother)L

eracah:
@dakinggg, you interested in squinting at the tests to make sure the remote tests are sufficient and I don't have to launch any manny tests?

composer/callbacks/checkpoint_saver.py
```diff
@@ -397,17 +397,32 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-            os.symlink(os.path.relpath(filename, os.path.dirname(symlink)), symlink)
+            os.symlink(os.path.relpath(saved_path, os.path.dirname(symlink)), symlink)
 
         # if remote file name provided, upload the checkpoint
         if self.remote_file_name is not None:
-            remote_file_name = self.remote_file_name.format(
-                state,
-                is_deepspeed,
-            ).lstrip('/')
+            # Use the stem of remote_filename (with rank info stripped) as a prefix folder.
```
dakinggg, line 404: this comment is out of date with new the new implementation, right?

composer/callbacks/checkpoint_saver.py
```diff
@@ -397,17 +397,32 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+            # e.g. s3://my-bucket/path/to/checkpoints/ep1-ba2-rank3.pt is now s3://my-bucket/path/to/checkpoints/ep1-ba2/ep1-ba2-rank3.pt
+            if state.fsdp_sharded_state_dict_enabled and state.sharded_ckpt_prefix_dir is not None:
```
dakinggg, line 406: What happens if `state.fsdp_sharded_state_dict_enabled` and `state.sharded_ckpt_prefix_dir is None`? Is that possible?

composer/utils/checkpoint.py
```diff
@@ -521,7 +530,18 @@ def save_checkpoint(
     if weights_only and not is_deepspeed:
         state_dict['state'] = {'model': state_dict['state']['model']}
 
-    save_filename = PartialFilePath(filename).format(state, is_deepspeed)
+    # Sharded checkpoints get their own little folder.
+    if state.fsdp_sharded_state_dict_enabled and state.sharded_ckpt_prefix_dir is not None:
```
dakinggg, line 534: same question

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -348,3 +352,43 @@ def test_mismatch_timestamp_error(world_size, tmp_path: pathlib.Path, state_dict
...
+@world_size(2)
+@pytest.mark.parametrize('use_remote', [pytest.param(True, marks=pytest.mark.remote), False])
+@pytest.mark.parametrize('state_dict_type', ['local', 'sharded'])
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('1.13.0'),
+                    reason='requires PyTorch 1.13 or higher')
+def test_sharded_folder(world_size,
+                        use_remote,
+                        tmp_path: pathlib.Path,
+                        state_dict_type: str,
+                        s3_bucket: str = 'mosaicml-internal-checkpoints-test/ckpt_folder_test/'):
```
dakinggg, line 367: this needs to point to the official integration test bucket (which it will if you don't give it a value).

dakinggg (approved):
I'd appreciate one manual resumption test, although the unit test coverage seems good to me.

composer/callbacks/checkpoint_saver.py
eracah: nope just trying to please pyright

eracah: well unless the user explicitly sets it to none

dakinggg: could you do an assert for the invariant rather than adding it to the if statement? I think that would be more clear.

eracah:
Ok manual tests look good

PR #2226, opened 2023-05-15
Upgrade to mcli0.4, smaller mcli improvements

aspfohl:
# What does this PR do?

Upgrades to mcli 0.4.0

Small changes:
* Uses root import instead of `mcli.sdk` (This is not required, but feels nicer 🙂)
* Renames mcp to mcli
* Removes `mcli init --mcloud` for CI testing - this is only user facing, not actually needed if api key env variable is set

Dependabot wants to update since we recently pushed 0.4.0 https://github.com/mosaicml/composer/pull/2220

### DIFF
--- a/composer/loggers/mosaicml_logger.py
+++ b/composer/loggers/mosaicml_logger.py
@@ -113,8 +113,8 @@ class MosaicMLLogger(LoggerDestination):
     def _flush_metadata(self, force_flush: bool = False) -> None:
         """Flush buffered metadata to MosaicML if enough time has passed since last flush."""
         if self._enabled and (time.time() - self.time_last_logged > self.log_interval or force_flush):
+            from mcli import update_run_metadata
             from mcli.api.exceptions import MAPIException
-            from mcli.sdk import update_run_metadata
             try:
                 update_run_metadata(self.run_name, self.buffered_metadata)
                 self.buffered_metadata = {}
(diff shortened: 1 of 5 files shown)


composer/loggers/mosaicml_logger.py
```diff
@@ -113,8 +113,8 @@ def _log_metadata(self, metadata: Dict[str, Any]) -> None:
     def _flush_metadata(self, force_flush: bool = False) -> None:
         """Flush buffered metadata to MosaicML if enough time has passed since last flush."""
         if self._enabled and (time.time() - self.time_last_logged > self.log_interval or force_flush):
+            from mcli import update_run_metadata
             from mcli.api.exceptions import MAPIException
```
mvpatel2000, line 117: Can we direct import the exception too?

aspfohl: Done.

mvpatel2000:
@aspfohl can you please fix unit tests and then we'll merge?

mvpatel2000:
Note: In the logger, we now do `mcli.update_run_metadata` so the unit test can actually monkeypatch it. Otherwise, it gets imported and we can't actually change it without something a lot more complicated (monkeypatching the body of the fn without changing the fn pointer, which is nasty)

mvpatel2000 (approved):

dakinggg (approved):
approving to unblock

PR #2228, opened 2023-05-17
Add cmake to the docker images

dakinggg:
# What does this PR do?
This is for llm-foundry, which needs to depend on a fork of triton that @vchiley has created until torch2 and triton work out their issues that prevent us from fully using the new triton.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -292,6 +292,11 @@ RUN if [ -n "$CUDA_VERSION" ] ; then \
         pip${PYTHON_VERSION} install --no-cache-dir flash-attn==1.0.3.post0; \
     fi
 
+###############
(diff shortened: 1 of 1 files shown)


vchiley (approved):
lgtm

dakinggg:
Closing because I opened this from a fork.

dskhudia:
@dakinggg : Is this to make triton work with torch2.0?

dakinggg:
Sort of. @vchiley's workaround to allow us to move to torch2 is to have two triton versions in the dependencies. In order to do this, he had to clone the pre-mlir triton, change the package name, and add it as a directly installed dependency in foundry. It requires cmake to build, so we're baking cmake into the docker images.

PR #2229, opened 2023-05-17
add cmake

dakinggg:
# What does this PR do?
This is for llm-foundry, which needs to depend on a fork of triton that @vchiley has created until torch2 and triton work out their issues that prevent us from fully using the new triton.


vchiley (approved):
lgtm

mvpatel2000 (approved):

PR #2234, opened 2023-05-18
Remove deprecated code / torch 1.11 / torch 1.12

mvpatel2000:
# What does this PR do?

We've deprecated several things that will be removed in next release. This purges deprecated code. 

We also no longer support torch 1.11 and torch 1.12. This removes these tests.

Since we tie python 3.8 to 1.11 and 3.9 to 1.12, this removes testing for those python versions. ~Do we want to add docker builds and CI/CD for those? or is lint sufficient?~ Per offline discussion, this is fine (and we'll leave in lint + smoketest for these versions)

[CO-2087](https://mosaicml.atlassian.net/browse/CO-2087)

[CO-2087]: https://mosaicml.atlassian.net/browse/CO-2087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dakinggg (commented):
Can you please also search for all mentions of `validate` or `metrics` instead of `get_metrics`? That was the original impetus for the JIRA about removing mentions of deprecated stuff.

mvpatel2000:
> Can you please also search for all mentions of `validate` or `metrics` instead of `get_metrics`? That was the original impetus for the JIRA about removing mentions of deprecated stuff.

Done. Don't see anything that showed up

dakinggg (approved):

Issue #2231, opened 2023-05-18
Error while saving checkpoint


singhalshikha518:
** Environment **

Collecting system information...
---------------------------------
System Environment Report        
Created: 2023-05-18 13:38:50 GMT
---------------------------------

PyTorch information
-------------------
PyTorch version: 2.0.1+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

OS: Linux Server 7.9 (x86_64)
GCC version: (GCC) 8.3.1 20190311 (Red Hat 8.3.1-3.2.0.7)
Clang version: Could not collect
CMake version: version 3.26.3
Libc version: glibc-2.17

Python version: 3.8.9 (default, Nov  1 2022, 18:17:19)  [GCC 4.8.5 20150623 (Red Hat 4.8.5-44.0.3)] (64-bit runtime)
Python platform: Linux-5.4.17-2136.310.7.el7uek.x86_64-x86_64-with-glibc2.17
Is CUDA available: True
CUDA runtime version: 11.6.55
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: 
GPU 0: NVIDIA A100-SXM4-40GB
GPU 1: NVIDIA A100-SXM4-40GB
GPU 2: NVIDIA A100-SXM4-40GB
GPU 3: NVIDIA A100-SXM4-40GB
GPU 4: NVIDIA A100-SXM4-40GB
GPU 5: NVIDIA A100-SXM4-40GB
GPU 6: NVIDIA A100-SXM4-40GB
GPU 7: NVIDIA A100-SXM4-40GB

Nvidia driver version: 470.141.03
cuDNN version: Probably one of the following:
/usr/lib64/libcudnn.so.8.5.0
/usr/lib64/libcudnn_adv_infer.so.8.5.0
/usr/lib64/libcudnn_adv_train.so.8.5.0
/usr/lib64/libcudnn_cnn_infer.so.8.5.0
/usr/lib64/libcudnn_cnn_train.so.8.5.0
/usr/lib64/libcudnn_ops_infer.so.8.5.0
/usr/lib64/libcudnn_ops_train.so.8.5.0
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                128
On-line CPU(s) list:   0-127
[...]

PR #2237, opened 2023-05-19
Change AdamW step to a tensor instead of an int

eracah:
# What does this PR do?
changes the step in AdamW to a tensor

this avoids this issue: https://github.com/pytorch/pytorch/issues/99079 and this one: https://github.com/mosaicml/composer/issues/2231 because by setting step to a tensor, there are no more non-tensors in the optimizer state, so we can't get to the buggy part of the torch code.

fix CO-2107


dakinggg (commented):
Can you add a test that would've caught this? I guess fsdp checkpoint with Adam?

eracah:
> Can you add a test that would've caught this? I guess fsdp checkpoint with Adam?

Ok added it. didn't run it, so will let the CI determine if it passes

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -81,8 +86,11 @@ def get_trainer(
 
 def _compare_optims_between_state_dicts(state_dict1, state_dict2):
     # Check that optim params are equal between checkpoint and in memory optimizer
-    state_dict1_optim_params = state_dict1['optimizers']['Adam']['state']
-    state_dict2_optim_params = state_dict2['optimizers']['Adam']['state']
+    optim_key1 = list(state_dict1['optimizers'].keys()).pop()
```
dakinggg, line 89: maybe assert these lists are of length 1, just to make the assumption more clear

dakinggg (approved):

tests/trainer/test_sharded_checkpoint.py
eracah: haha we've lost years of our lives checking if the number of optimizers is 1

shivashankarrs:
> Can you add a test that would've caught this? I guess fsdp checkpoint with Adam?

Thanks Daniel :) Hi here again.

dakinggg:
Hey @shivashankarrs :)

PR #2249, opened 2023-05-23
Fix daily tests by fixing test_fsdp_load_old_checkpoint

eracah:
# What does this PR do?
* adds prefix that won't be purged
* adds prefix that will be purged
* changes all ephemeral tests to use ephemeral rpefix
* changes loading checkpoints tests to use permanent prefix
* Also changes tests to run by composer version and loads checkpoints from a specific prefix that won't be purged
* Filters ShardedGradScaler warning for composer v0.13.5

fix CO-2099


mvpatel2000 (commented):
Can we please move these tests into a new file under `trainer` (`test_backwards_compatibility`)? 

Did we verify these tests pass now? 

Is there a way to have 1 fixture instead of merging in each test? Eg the fixture takes in the s3 bucket arg and returns the updated s3 path?

mvpatel2000 (approved):

PR #1428, opened 2022-08-19
Refactor CheckpointSaver

hanlint:
This is in a series of PRs refactoring the checkpointing code for readability and modularity without changing the public-facing API.

This PR cleans up the checkpoint saver code. Touching the `checkpoint.save_checkpoints` function will be deferred to another PR.

PR #2262, opened 2023-05-31
Enable Elastic Sharded Checkpointing

eracah:
# What does this PR do?
Enables elastic resumption behavior for sharded checkpointing.
From a user standpoint, sharded checkpointing is almost the same interface as before, but now you can load sharded checkpoints on a different world size than you saved.

This is enabled by integrating the `torch.distributed.checkpoint` package into Composer.
Namely we integrate:
1. `torch.distributed.checkpoint.save_state_dict` using `FileSystemWriter` for local saving
2.  `torch.distributed.checkpoint.load_state_dict` using `FileSystemReader` for local loading
3. Subclass `FileSystemReader` as `DistCPObjectStoreReader` for reading from object store

Along for the ride:
* fix torchmetrics casts tensors to numpy arrays at save and then converts them back to tensors upon load to avoid the tensors being sharded by FSDP.

Other changes:
* optimizer states are pushed to the root of the state dict
* symlinks point to parent directory of checkpoints instead of invidual files
* we backwards support legacy sharded checkpoints (pre-torch 2.0, pre-elastic)
* add a lot of tests to test_fsdp_checkpoint to make sure everything works

# Before submitting

## Save Local
- [x] implemented
- [x] approved by Daniel or Mihir

## Save Remote
- [x] implemented
- [x] approved by Daniel or Mihir

## Load Local
- [x] implemented
- [x] approved by Daniel or Mihir

## Load Remote
- [x] implemented
- [x] approved by Daniel or Mihir

## Double Check Torch 1.13 Works
- [x] done

## Add Elasticity Test
- [x] implemented
- [x] approved by Daniel or Mihir or Nikhil

## Save/Load local symlinks
- [x] implemented
- [x] approved by Daniel or Mihir

## Single Node Manual Tests
- [x] save 8 -> load 8
- [x] save 8 -> load 4
- [x] save 8 -> load 7
- [x] save 8 -> load 3
- [x] save 2 -> load 8
- [x] save 2 -> load 3
- [x] save 2 -> load 7

## Multinode Manual Tests
- [x] save 16 -> load 16
- [x] save 16 -> load 8
- [x] save 8 -> load 16
- [x] save 24 -> load 16
- [x] save 16 -> load 32
- [x] save 8 -> load 16 (pytorch 2.0.0, which fails as expected)

## Save/Load Remote Symlinks
- [x] implemented
- [x] approved by Daniel or Mihir

## Clean Up checkpoints
- [x] implemented
- [x] approved by Daniel or Mihir

## Support for legacy sharded files
- [x] implemented
- [x] approved by Daniel or Mihir

- [x] fix torchmetrics load issue
- [x] write updated docs
- [x] change version check to 2.01 for load

fix CO-1953


composer/utils/checkpoint.py
```diff
@@ -36,7 +35,7 @@
 
 _COMPOSER_STATES_FILENAME = 'composer_states.pt'
 _DEEPSPEED_TAG = 'deepspeed'  # always tag with the same, deterministic name. We'll rename the tarball to the appropriate name.
-
+_TORCH_DISTRIBUTED_CHECKPOINTS_FILENAME = f"__{dist.get_global_rank()}_0.distcp"
```
dakinggg, line 38: why is there a zero at the end of the filename?

composer/utils/checkpoint.py
```diff
@@ -36,7 +35,7 @@
...
```
mvpatel2000, line 39: putting `{dist.get_global_rank()}` will make it fail if dist is not initialized, which is a problem for "normal" non-distributed usage

composer/utils/checkpoint.py
```diff
@@ -532,32 +531,55 @@ def save_checkpoint(
...
-    # only rank 0 saves the state_dict unless state.fsdp_sharded_state_dict_enabled=True.
-    if dist.get_global_rank() == 0 or state.fsdp_sharded_state_dict_enabled:
+    # All ranks save for deepspeed.
+    if is_deepspeed:
+        log.debug('Saving deepspeed checkpoints to %s...', save_filename)
+        _save_deepspeed_model(state.deepspeed_model, save_filename)
+
+    # Sharded checkpointing for torch >=2.0 uses the torch.distributed.checkpoint module.
+    elif state.fsdp_sharded_state_dict_enabled and using_torch_2():
+        import torch.distributed.checkpoint as dist_cp
```
dakinggg, line 566: this needs try except i think?

composer/utils/checkpoint.py
```diff
@@ -532,32 +531,55 @@ def save_checkpoint(
 
     # Sharded checkpoints get their own little folder.
     if state.fsdp_sharded_state_dict_enabled:
+        if using_torch_2():
+            # Make state dict.
+            if weights_only:
+                state_dict = {'model': state.state_dict()['model']}
```
mvpatel2000, line 537:
```suggestion
            if weights_only:
                state_dict['model'] = state_dict['state']['model']
                del state_dict['state']
```
Weights only should keep rng per logic above?

composer/utils/checkpoint.py
```diff
@@ -532,32 +531,55 @@ def save_checkpoint(
...
+            else:
+                # Dictionary must be flat to faciliate loading optimizer params.
+                state_dict = {
+                    **state.state_dict(),
+                    'rng': reproducibility.get_rng_state(),
+                }
```
mvpatel2000, line 543:
```suggestion
                state_dict = {
                    **state['state'],
                    'rng': reproducibility.get_rng_state(),
                }
```

composer/utils/checkpoint.py
```diff
@@ -532,32 +531,55 @@ def save_checkpoint(
...
             torch.save(state_dict, f)
 
         if is_tar(save_filename):
             _compress_file(save_filename, basename=_COMPOSER_STATES_FILENAME)
 
-    # all ranks save for deepspeed
-    if is_deepspeed:
-        _save_deepspeed_model(state.deepspeed_model, save_filename)
+    else:
+       log.debug(f'Rank 0 is saving a checkpoint right now, but this rank {dist.get_global_rank()} doesn''t need to...')
```
dakinggg, line 582:
```suggestion
       log.debug(f'Rank 0 is saving a checkpoint right now, but this rank (rank {dist.get_global_rank()}) doesn''t need to...') 
```

mvpatel2000, line 582:
```suggestion
       log.debug(f'Only rank 0 is saving a checkpoint, so rank {dist.get_global_rank()} skips checkpointing.') 
```

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -422,3 +424,41 @@ def test_sharded_folder(world_size, use_remote, tmp_path: pathlib.Path, state_di
...
     trainer2.close()
+
+@pytest.mark.gpu
+@world_size(2)
+@pytest.mark.parametrize('state_dict_type', ['local', 'sharded'])
+@pytest.mark.parametrize('weights_only', [True, False])
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.0'),
+                    reason='requires PyTorch 2.0.0 or higher')
+@pytest.mark.filterwarnings(r'ignore:TypedStorage is deprecated.:UserWarning')
+def test_new_sharded_save(world_size, tmp_path: pathlib.Path, state_dict_type: str, weights_only: bool):
```
dakinggg, line 435: more descriptive name than "new" please

composer/utils/checkpoint.py
```diff
@@ -532,32 +531,55 @@ def save_checkpoint(
 
     # Sharded checkpoints get their own little folder.
     if state.fsdp_sharded_state_dict_enabled:
+        if using_torch_2():
+            # Make state dict.
+            if weights_only:
```
dakinggg, line 536: check whether we can _only_ bump optimizers to the top level and otherwise not change this logic from before this pr

composer/utils/checkpoint.py
eracah: it's the file count per rank and we set `single_file_per_rank` to True, so it will always be zero
https://github.com/pytorch/pytorch/blob/9a2df0a5af010201f31c87c593227659ae1adc4e/torch/distributed/checkpoint/filesystem.py#L371-L383

composer/utils/checkpoint.py
eracah: pretty sure it just returns 0 if dist is not initialized

composer/utils/checkpoint.py
eracah: why?

tests/trainer/test_sharded_checkpoint.py
eracah: it's just a temporary test until we get load working

composer/utils/checkpoint.py
eracah: ok changed it to just bump optimizers

composer/utils/checkpoint.py
dakinggg: like pyright or tests or something? but if it passes, no worries.

composer/utils/checkpoint.py
eracah: ended up keeping the old logic except for when weights_only is False and torch >=2.0

composer/utils/checkpoint.py
eracah: N/A

eracah:
Ok, I changed it so only optimizer gets bumped to the top. Can either @dakinggg or @mvpatel2000 click the `approved by Daniel or Mihir` box?

eracah:
ok, @mvpatel2000, @dakinggg, I'm ready for the next partial PR review. Most of the changes are [here](https://github.com/mosaicml/composer/pull/2262/commits/ebd828ab564dbd30e711fac90b5ce180b32a2245)

Thanks y'all 🥰

composer/utils/checkpoint.py
```diff
@@ -532,33 +528,49 @@ def save_checkpoint(
 
     # Sharded checkpoints get their own little folder.
     if state.fsdp_sharded_state_dict_enabled:
+        if using_torch_2():
```
dakinggg, line 531: suggest adding a comment here explaining why this upleveling is necessary (and link to pytorch docs/issues if any, but fine if that doesn't exist)

composer/callbacks/checkpoint_saver.py
```diff
@@ -411,14 +415,28 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                if dist.get_global_rank() == 0:
+                    metadata_local_file_name = os.path.join(
+                        Path(saved_path).parent,
+                        _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME
+                        )
+                    metadata_local_file_name = format_name_with_dist_and_time(metadata_local_file_name, state.run_name, state.timestamp)
+                    metadata_remote_file_name = os.path.join(
+                        Path(remote_file_name).parent,
+                        _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME)
+                    metadata_remote_file_name = format_name_with_dist_and_time(metadata_remote_file_name, state.run_name, state.timestamp)
```
mvpatel2000, line 431: It's a little weird that the intermediate values have the same name as the final value... can we rename that?

Also, can you put a comment on what metadata actually is?

composer/callbacks/checkpoint_saver.py
```diff
@@ -411,14 +415,28 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                    logger.upload_file(remote_file_name=metadata_remote_file_name, 	file_path=metadata_local_file_name, overwrite=self.overwrite)
```
dakinggg, line 432: just checking my understanding. only rank 0 needs to upload the metadata file because it is putting into a remote folder which will ultimately contain one file for each ranks checkpoint + one metadata file, whereas locally every rank needs to save the metadata file because they all have different folders. is that correct?

eracah: so all ranks save to the same folder, so there is only one metadata file, but all ranks participate in writing it's content. The FileSystemManager does a big gather/reduce on the metadata from the other ranks and then dumps the result ot a file. But yeah we just need one of the ranks to upload the file to object store

composer/callbacks/checkpoint_saver.py
eracah: done

eracah:
## Save Remote
done. thanks, @dakinggg and @mvpatel2000

eracah:
alright @dakinggg, @mvpatel2000 , I got local load working. give it a look when you can! it's [here](https://github.com/mosaicml/composer/pull/2262/commits/65171676892053a87d4e43d9cc095c4a381dcc7f)

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+            ignore_keys=ignore_keys,
+            exclude_algorithms=exclude_algorithms,
+            algorithm_passes=algorithm_passes,
+        )
+    else:
+        # Download the checkpoint to the node-local folder
+        log.debug('Loading checkpoint at %s', path)
+        # Each node gets one unique folder to store checkpoints that is shared amongst all local ranks in that node.
+        # If fsdp sharded state_dicts is enabled then EVERY rank gets a unique checkpoint folder.
+        needs_unique_checkpoint_folder = state.fsdp_sharded_state_dict_enabled or dist.get_local_rank() == 0
```
dakinggg, line 222: seems like all the sharded bits in this else should be removed from this code path since it is handled in the if above?

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+    load_weights_only: bool = False,
+    strict_model_weights: bool = False,
+    progress_bar: bool = True,
+    ignore_keys: Optional[Union[List[str], Callable[[Dict], None]]] = None,
+    exclude_algorithms: Optional[List[str]] = None,
+    algorithm_passes: Optional[List[AlgorithmPass]] = None,
+) -> Dict:
+
+    if not using_torch_2():
+        raise ValueError(f'Sharded checkpoint loading requires torch version >= 2.0.0 Got {torch.__version__}')
```
dakinggg, line 275:
```suggestion
        raise ValueError(f'Sharded checkpoint loading requires torch version >= 2.0.0. You have torch version {torch.__version__}')
```

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+
+    from torch.distributed import checkpoint as dist_cp
+    from torch.distributed.checkpoint.metadata import Metadata
+    from torch.distributed.checkpoint.optimizer import load_sharded_optimizer_state_dict
+
+    # This function is used so we can figure out which ranks need to load saved rng's and which can just make their own.
```
dakinggg, line 281:
```suggestion
    # This function is used so we can figure out which ranks need to load saved rngs and which can just make their own.
```

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+    def _get_num_ranks_that_saved_rng(metadata: Metadata):
+        rng_inds = []
+        for field_name, field_value in metadata.planner_data.items():
+            if 'rng' in field_name:
```
dakinggg, line 285: what are the field names? is there a reason this check is `in` rather than `==`? I assume yes, just wondering if we can make the check more explicit/looking to understand what is in this `metadata.planner_data` dict

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+                _, rng_rank_index, _ = field_value
+                rng_inds.append(rng_rank_index)
+        rng_inds = set(rng_inds)
+        return max(rng_inds) + 1
```
dakinggg, line 289: why is this a max over index values rather than just a count of how many ranks had rng state?

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,158 @@ def load_checkpoint(
...
+        if not load_weights_only:
+            optim_state = load_sharded_optimizer_state_dict(model_state_dict=state.state_dict()['model'],
+                                                            optimizer_key='optimizers',
+                                                            storage_reader=storage_reader)
+            state.load_optim_state(optim_state)
+
+    # 3. Optionally, load RNG.
+    rng_state_dicts = reproducibility.get_rng_state()
+    if not load_weights_only:
+        # If we are resuming on more ranks than were used at save time we only want to load in rng's for those ranks
```
dakinggg, line 322:
```suggestion
        # If we are resuming on more ranks than were used at save time we only want to load in rngs for those ranks
```

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,50 +333,58 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
+    if state_dict_type == 'local':
+        pytest.xfail(
+            'Loading a state_dict_type="local" checkpoint with strict=True errors out. See https://github.com/pytorch/pytorch/issues/102667 for more info'
+        )
     if autoresume:
         run_name = 'my-autoresume-run'
     else:
         run_name = None
-    save_folder = tmp_path
+    save_folder = '/tmp/test_checkpoints'
```
dakinggg, line 354: how come this change to hardcode tmp path?

dakinggg (commented):
lgtm

composer/utils/checkpoint.py
eracah: maybe keep it if someone uses torch <2.0 to do sharding?

tests/trainer/test_sharded_checkpoint.py
eracah: all ranks have to share the same folder because they all need the .metadata file. If you use tmp_path then each rank gets a different path

dakinggg: yeah, i think it is better to sync the tmp path from rank 0, so that the dir is auto cleaned up

composer/utils/checkpoint.py
dakinggg: but the if above is `if state.fsdp_sharded_state_dict_enabled:`. so you can't end up in this code path?

dakinggg: is the if statement above supposed to gate on torch 2?

composer/utils/checkpoint.py
eracah: rng state is a list and field_names are FQN's and they [append the index value to each item in a list](https://github.com/pytorch/pytorch/blob/28f43c767c0de48885f8de285ca4af4424d28f29/torch/distributed/checkpoint/_traverse.py#L72-L74), so this means the each rank's rng will be an fqn with 'rng' somewhere in there along with the rank. Seems easier just to check 'rng' in `field_name`

composer/utils/checkpoint.py
eracah: oh hmm yeah prob can do length. not sure why I did it that way

composer/utils/checkpoint.py
eracah: oh good call yeah should change to gate by torch version as well

composer/utils/checkpoint.py
eracah: fixed

composer/utils/checkpoint.py
eracah: ok done

eracah:
## Load Local
Done. Thanks @mvpatel2000 and @dakinggg !

eracah:
Ok, @dakinggg , @mvpatel2000 , load from remote is live. PTAL. it's [here](https://github.com/mosaicml/composer/pull/2262/commits/8d1077d95050ee7ceaa7ac49dff4f2c61663abf8)

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,234 @@ def load_checkpoint(
...
-            assert node_checkpoint_folder is not None
-
-            composer_states_filepath, extracted_checkpoint_folder, extracted_rank_n = download_checkpoint(
-                path=path,
-                node_checkpoint_folder=node_checkpoint_folder,
-                object_store=object_store,
-                progress_bar=progress_bar,
-                fsdp_sharded_state_dict_enabled=state.fsdp_sharded_state_dict_enabled,
-                deepspeed_sharded_checkpoint=is_model_deepspeed(state.model),
+            with tempfile.TemporaryDirectory() as tempdir:
```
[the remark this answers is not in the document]
eracah: yeah super convoluted. removed because if it's not a directory the metadata download will fail and give the proper error

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,234 @@ def load_checkpoint(
...
+            # Instantiate FileSystemReader with destination path b/c that's where we will download files to.
+            # Because we already download the metadata file to the destination, everything will work swimmingly
+            super().__init__(destination_path)
+
+        def read_data(self, plan: LoadPlan, planner: LoadPlanner):
+            # 1. Download to the destination all files that this rank is responsible for.
+            for plan_item in plan.items:
+                relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                file_destination = str(Path(self.destination_path) / Path(relative_file_path))
+                if not os.path.exists(file_destination):
```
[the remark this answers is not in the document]
eracah: the reason for the check is the same file can show up multiple times in the plan and even multiple ranks can see the same file, so we're just avoiding downloading the same file multiple times. there is not risk for overwriting because this is a temp file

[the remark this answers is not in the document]
eracah: added a comment to explain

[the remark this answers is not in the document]
dakinggg: ah, didn't know that, thanks

composer/utils/checkpoint.py
```diff
@@ -200,46 +201,231 @@ def load_checkpoint(
...
+
+        def read_data(self, plan: LoadPlan, planner: LoadPlanner):
+            # 1. Download to the destination all files that this rank is responsible for.
+            for plan_item in plan.items:
+                # Each plan item has a storage index which points to the relative path of the shard file at save time.
+                relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                # Download the shard file to the relative path it's associated to and save that relative path
+                # to the root directory specified to the FileSystem reader constructor.
+                file_destination = str(Path(self.destination_path) / Path(relative_file_path))
+                # THe file could have already been downloaded as diffeent plan items can point to same file.
```
dakinggg, line 337:
```suggestion
                # The file could have already been downloaded as diffeent plan items can point to same file.
```

dakinggg (commented):
section lgtm

eracah:
## Load Remote
Ok this part is done!

eracah:
## Add Elasticity Test
Done

eracah:
## Local Symlink Support
Done

eracah:
## Remote Symlink/Autoresume Support
done

eracah:
## Legacy Sharding Support
done

eracah:
## Legacy Sharded Support
Done

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,53 +346,179 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
+
+    mono_load_path = os.path.join(base_path, 'mono.pt')
+    mono_trainer = get_trainer(
+        fsdp_state_dict_type='full',
+        load_path=mono_load_path,
+        num_features=32,  # This parameter setting is very important. Don't change or the test will fail.
+        num_classes=8,  # This parameter setting is very important. Don't change or the test will fail.
         precision=precision,
-        autoresume=autoresume,
+        autoresume=False,  # Hardcoded to false b/c mono checkpoints don't have symlinks to them.
```
dakinggg, line 471: whats up here?

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,53 +346,179 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
 
-    # Compare saved state and loaded state for both ranks.
-    _compare_model_params_between_state_dicts(state_dict_from_trainer1, state_dict_from_trainer2)
-    _compare_optims_between_state_dicts(state_dict_from_trainer1, state_dict_from_trainer2)
-    _compare_metrics_between_state_dicts(state_dict_from_trainer1, state_dict_from_trainer2)
+    sharded_trainer = get_trainer(
+        fsdp_state_dict_type=state_dict_type,
+        save_folder=save_folder,
+        load_path=sharded_load_path,
+        num_features=32,  # This parameter setting is very important. Don't change or the test will fail.
```
dakinggg, line 488: can you say why its important (cause has to match the saved checkpoint)

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -349,6 +528,7 @@ def test_fsdp_partitioned_state_dict_load(world_size, tmp_path: pathlib.Path, st
 @pytest.mark.skipif(version.parse(torch.__version__) < version.parse('1.13.0'),
                     reason='requires PyTorch 1.13 or higher')
 def test_mismatch_timestamp_error(world_size, tmp_path: pathlib.Path, state_dict_type: str, autoresume: bool):
+    pytest.xfail('Fails for now until we get symlink suppport working')
```
dakinggg, line 531: remove now?

composer/trainer/trainer.py
```diff
@@ -1320,12 +1321,44 @@ def __init__(
...
                     'for all `RemoteUploaderDownloader` instances.')
             assert latest_remote_file_name is not None
-            autoresume_checkpoint_path = self._get_autoresume_checkpoint(
-                save_folder=save_folder,
-                save_latest_filename=save_latest_filename,
-                save_latest_remote_file_name=latest_remote_file_name,
-                loggers=loggers,
-                load_progress_bar=load_progress_bar)
+            if self.state.fsdp_elastic_sharded_enabled:
+                ar_object_store = maybe_create_object_store_from_uri(save_folder)
```
dakinggg, line 1325:
```suggestion
                autoresume_object_store = maybe_create_object_store_from_uri(save_folder)
```

composer/trainer/trainer.py
```diff
@@ -1320,12 +1321,44 @@ def __init__(
...
+                            autoresume_checkpoint_path = ar_object_store.get_uri(real_path)
+                        except FileNotFoundError:
+                            autoresume_checkpoint_path = None
+                # Symlink is local.
+                else:
+                    save_latest_filename = format_name_with_dist(save_latest_filename, self.state.run_name)
+                    save_folder = format_name_with_dist(save_folder, self.state.run_name)
+                    latest_checkpoint_path = os.path.join(save_folder, save_latest_filename)
+                    if os.path.exists(latest_checkpoint_path):
+                        if os.path.islink(latest_checkpoint_path):
```
dakinggg, line 1346: doesn't this have to be a symlink? or am i missing something?

composer/trainer/trainer.py
```diff
@@ -1320,12 +1321,44 @@ def __init__(
...
-            autoresume_checkpoint_path = self._get_autoresume_checkpoint(
-                save_folder=save_folder,
-                save_latest_filename=save_latest_filename,
-                save_latest_remote_file_name=latest_remote_file_name,
-                loggers=loggers,
-                load_progress_bar=load_progress_bar)
+            if self.state.fsdp_elastic_sharded_enabled:
+                ar_object_store = maybe_create_object_store_from_uri(save_folder)
+                # Symlink is on object store.
+                if ar_object_store is not None:
```
dakinggg, line 1327: doesn't this have to do a loop over loggers and find all the RUDs too? or are we just saying no support to autoresume + torch2 sharded + from your own RUD

composer/utils/checkpoint.py
```diff
@@ -100,6 +101,20 @@ def format(self, state: State, is_deepspeed: bool = False, keep_placeholders: bo
                 ) + extra_suffix
 
 
+def is_checkpoint_legacy_sharded(object_store, source_path):
```
dakinggg, line 104: types and docstring

eracah:
## Clean up local checkpoints
done

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,53 +346,179 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
+@pytest.mark.gpu
+@pytest.mark.remote
+@world_size(2)
+@pytest.mark.parametrize('state_dict_type', ['sharded'])
+@pytest.mark.parametrize('precision', ['amp_bf16', 'amp_fp16'])
+@pytest.mark.parametrize('autoresume', [False])  #, True])  # True commented out for now
+@pytest.mark.parametrize('num_shards', [2, 4, 7])
+@pytest.mark.parametrize('sharding_strategy', ['FULL_SHARD'])
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.1'), reason='requires PyTorch 2.01 or higher')
+@pytest.mark.filterwarnings(r'ignore:TypedStorage is deprecated.:UserWarning')
```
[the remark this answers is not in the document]
eracah: ?

tests/trainer/test_sharded_checkpoint.py
eracah: checkpoints saved with monolithic checkpoint saver callback don't have symlinks

eracah: added comment clarifying

tests/trainer/test_sharded_checkpoint.py
eracah: added clarifying comment

composer/trainer/trainer.py
eracah: I mean this covers downloading from the save_folder right?

composer/trainer/trainer.py
eracah: yeah I think so

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,53 +346,179 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
+    trainer2.fit()
+    trainer2.close()
+
+
+@pytest.mark.gpu
+@pytest.mark.remote
+@world_size(2)
+@pytest.mark.parametrize('state_dict_type', ['sharded'])
+@pytest.mark.parametrize('precision', ['amp_bf16', 'amp_fp16'])
+@pytest.mark.parametrize('autoresume', [False])  #, True])  # True commented out for now
```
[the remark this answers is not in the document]
eracah: done

tests/trainer/test_sharded_checkpoint.py
```diff
@@ -293,53 +346,179 @@ def test_fsdp_full_state_dict_load_with_ema(world_size, tmp_path: pathlib.Path,
...
+@pytest.mark.parametrize('num_shards', [2, 4, 7])
+@pytest.mark.parametrize('sharding_strategy', ['FULL_SHARD'])
```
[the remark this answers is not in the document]
eracah: do we have to?

composer/callbacks/checkpoint_saver.py
```diff
@@ -431,19 +458,42 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+            assert metadata_local_file_path is not None
+            self.saved_checkpoints.append((saved_path, metadata_local_file_path))
+        else:
+            self.saved_checkpoints.append(saved_path)
 
         if self.num_checkpoints_to_keep >= 0:
-            self._rotate_checkpoints()
+            self._rotate_checkpoints(sharding_enabled=state.fsdp_sharded_state_dict_enabled)
+
+    def _rotate_checkpoints(self, sharding_enabled: bool = False):
```
dakinggg, line 483: Is this rotation stuff tested with sharded checkpoints anywhere?

composer/trainer/trainer.py
dakinggg: yeah, but autoresume previously would loop over all loggers. Are we just saying that doesn't work for elastic sharding?

tests/trainer/test_sharded_checkpoint.py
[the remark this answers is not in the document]
dakinggg: is there a reason not to? it should just be adding that parameter/if it doesn't work, we should probably fix it...

composer/callbacks/checkpoint_saver.py
eracah: yes test_cleanup_sharded_checkpoints

composer/trainer/trainer.py
eracah: correct

tests/trainer/test_sharded_checkpoint.py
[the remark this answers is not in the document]
eracah: done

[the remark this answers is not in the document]
eracah: done

composer/callbacks/checkpoint_saver.py
```diff
@@ -267,9 +272,9 @@ class CheckpointSaver(Callback):  # noqa: D101
             remote file systems.
 
     Attributes:
-        saved_checkpoints (List[Tuple[Timestamp, List[pathlib.Path]]]): The checkpoint timestamps and filepaths.
+        saved_checkpoints (List[str]): The checkpoint filepaths.
```
dakinggg, line 275: were these timestamps used anywhere? this is an api change, so we might need to not do it?

composer/utils/checkpoint.py
```diff
@@ -200,46 +215,247 @@ def load_checkpoint(
...
+    ignore_keys: Optional[Union[List[str], Callable[[Dict], None]]] = None,
+    exclude_algorithms: Optional[List[str]] = None,
+    algorithm_passes: Optional[List[AlgorithmPass]] = None,
+) -> List[Dict]:
+
+    if not using_torch_2():
+        raise ValueError(
+            f'Sharded checkpoint loading requires torch version >= 2.0.0. You have torch version {torch.__version__}')
+
+    using_multinode = dist.get_world_size != dist.get_local_world_size
```
dakinggg, line 315: aren't these function calls?

dakinggg (commented):
Please add a full PR description, and document the final manual tests you have run. Otherwise it looks good to me.

EDIT: sorry, just saw the manual tests in the pr description. Do we have any tests around the backwards compat/errors when it won't work?

composer/utils/checkpoint.py
eracah: ohh yeah good catch

composer/callbacks/checkpoint_saver.py
eracah: I don't think they are. I guess I was trying to get pyright to work

eracah:
> EDIT: sorry, just saw the manual tests in the pr description. Do we have any tests around the backwards compat/errors when it won't work?

What do you mean? Like loading old checkpoints? there is a load_old_checkpoint test

eracah:
> Please add a full PR description, 

Done

dakinggg:
I mostly meant, for that one thing where you forgot the parenthesis for the functions, did we have a test that would've caught that issue?

composer/callbacks/checkpoint_saver.py
dakinggg: I think we should probably put it back in case someone has code relying on the timestamps being present in this attribute

eracah:
> I mostly meant, for that one thing where you forgot the parenthesis for the functions, did we have a test that would've caught that issue?

Confirmed. Got this error when running multinode load on torch 2.0 once I added the parentheses:

```│ /usr/lib/python3/dist-packages/composer/utils/checkpoint.py:317 in           │
│ load_sharded_checkpoint                                                      │
│                                                                              │
│   314 │                                                                      │
│   315 │   using_multinode = dist.get_world_size() != dist.get_local_world_si │
│   316 │   if not using_torch_2_0_1() and using_multinode:                    │
│ ❱ 317 │   │   raise ValueError(                                              │
│   318 │   │   │   f'Sharded checkpoint loading on >1 node requires torch ver │
│   319 │   │   )                                                              │
│   320                                                                        │
╰──────────────────────────────────────────────────────────────────────────────╯
ValueError: Sharded checkpoint loading on >1 node requires torch version >= 
2.0.1. You have torch version 2.0.0+cu117
```

dakinggg (approved):
🚢🚢 (with the one remaining comment)

composer/callbacks/checkpoint_saver.py
eracah: It was actually removed [here](https://github.com/mosaicml/composer/pull/1428/files#diff-b3be1fd7d6bb329e7dfcb406b7996e3ef170416faf5524caf50a451333b5f29eL314-R316) and the docstring is just an artifact

eracah: but I reverted it anyway

dakinggg: Oh wait, we should make the doctoring correct. Sorry I thought you changed the implementation. the perils of reviewing on my phone

PR #2263, opened 2023-05-31
Bump FlashAttention version in Dockerfile

nik-mosaic:
llm-foundry requires an upgrade, so we also upgrade here.
https://github.com/mosaicml/llm-foundry/pull/261


dakinggg:
foundry PR goes first

PR #2265, opened 2023-06-01
Update torch device naming convention for h100 gpus

vchiley:
# What does this PR do?

Updates torch device naming convention for h100 gpus

# Before submitting

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -97,7 +97,7 @@ def get_gpu_flops_available(state: State):
     device_name = torch.cuda.get_device_name().lower()
     if 'h100' in device_name and 'hbm3' in device_name:
         device_name = 'h100-sxm'
-    elif 'h100' in device_name and 'pcie' in device_name:
+    elif 'h100' in device_name and ('pcie' in device_name or 'hbm2e' in device_name):
         device_name = 'h100-pcie'
     elif 'a100' in device_name:
         device_name = 'a100'
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -95,9 +95,9 @@ def get_gpu_flops_available(state: State):
 
     # torch.cuda.get_device_name() ex output: 'NVIDIA A100-SXM4-40GB'
     device_name = torch.cuda.get_device_name().lower()
-    if 'h100-sxm' in device_name:
+    if 'h100' in device_name and 'hbm3' in device_name:
         device_name = 'h100-sxm'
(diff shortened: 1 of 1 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -95,9 +95,9 @@ def get_gpu_flops_available(state: State):
 
     # torch.cuda.get_device_name() ex output: 'NVIDIA A100-SXM4-40GB'
     device_name = torch.cuda.get_device_name().lower()
-    if 'h100-sxm' in device_name:
+    if 'h100' in device_name and 'hbm3' in device_name:
```
dakinggg, line 98: where did the old name come from and how do we know the new name is consistently correct?

dakinggg (approved):

composer/callbacks/speed_monitor.py
vchiley: Name was assumed given the v100 names:
```
    elif 'v100-sxm' in device_name:
         ...
    elif 'v100-pcie' in device_name:
```
but now I'm not convinced those are correct / has anyone tried to use those?

I also updt the pcie version because Daya previously saw a diff name
`elif 'h100' in device_name and ('pcie' in device_name or 'hbm2e' in device_name):`
ie we included `hbm2e` as an option

dakinggg: any chance these are documented somewhere? or exist somewhere in pytorch code?

vchiley: in the code it comes from [here](https://pytorch.org/docs/stable/_modules/torch/cuda.html#get_device_properties)
which calls `_get_device_properties` which doesn't exist until 3 lines before `_lazy_init` defines the fn which seems to be calling `torch._C._cuda_init` from [here](https://github.com/pytorch/pytorch/blob/20ca994a3ec1b315292768b97a3c896c13b678af/torch/_C/__init__.pyi.in#L1512) which is defined [here](https://github.com/pytorch/pytorch/blob/20ca994a3ec1b315292768b97a3c896c13b678af/torch/csrc/cuda/Module.cpp#L1371)

that might be the extent of my sleuthing abilities

dakinggg: alright cool, lets just merge this and fix later if we run into issues.

dakinggg: thanks for looking

vchiley:
added
```
pip install mosaicml@git+https://github.com/vchiley/composer.git@vc/h100_torch_name
```
to my cmd and now I get MFU
Screenshot 2023-05-31 at 6 15 43 PM

PR #2267, opened 2023-06-01
Spin dataloader arg

mvpatel2000:
# What does this PR do?

Add `spin_dataloader` argument to Trainer. This lets users disable spinning for faster resumption, which users have occasionally requested when they don't want deterministic resumption.

It's a little annoying that we have a var `spin_datalaoders` but a fn `_spin_dataloaders`.... It's probably fine but open to other ideas? We could also just flatten `_spin_dataloaders` since its only used once.

[CO-2154
](https://mosaicml.atlassian.net/browse/CO-2154)

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1524,6 +1524,7 @@ class Trainer:
         train_dataloader: Optional[Union[Iterable, DataSpec, Dict[str, Any]]] = None,
         train_dataloader_label: str = 'train',
         train_subset_num_batches: Optional[int] = None,
+        spin_dataloaders: Optional[bool] = None,
 
         # Timing
         duration: Optional[Union[int, str, Time[int]]] = None,
@@ -1618,6 +1619,7 @@ class Trainer:
             train_dataloader (Iterable | DataSpec | Dict[str, Any], optional): See :class:`.Trainer`.
             train_dataloader_label (str, optional): See :class:`.Trainer`.
             train_subset_num_batches (int, optional): See :class:`.Trainer`.
+            spin_dataloaders (bool, optional): See :class:`.Trainer`.
             reset_time (bool): Whether to reset the :attr:`.State.timestamp` to zero values. Defaults to False.
 
                 If ``True``, the timestamp will be zeroed out, causing :class:`.ComposerScheduler` and
@@ -1663,6 +1665,8 @@ class Trainer:
             _raise_missing_argument_exception('train_dataloader')
         if train_subset_num_batches is not None:
             self.state.dataloader_len = train_subset_num_batches
+        if spin_dataloaders is not None:
+            self.spin_dataloaders = spin_dataloaders
 
         # Reset Time
         if reset_time:
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -419,6 +419,11 @@ class Trainer:
             This setting will end each epoch early to avoid additional training that will not be profiled.
 
             This parameter is ignored if ``train_dataloader`` is not specified.
+        spin_dataloaders (bool, optional): If ``True``, dataloaders will be spun up to the current timestamp.
+            This ensures dataloaders continue from the same batch when resuming training. (default: ``True``)
+
+            .. note:: Spinning dataloaders can be potentially very slow. If resuming from a checkpoint where
+                determinism is not required, it is possible to resume faster by setting ``spin_dataloaders=False``.
         max_duration (Time | str | int, optional): The maximum duration to train. Can be an integer, which will be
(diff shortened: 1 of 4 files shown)


composer/trainer/trainer.py
```diff
@@ -419,6 +419,11 @@ class Trainer:
             This setting will end each epoch early to avoid additional training that will not be profiled.
 
             This parameter is ignored if ``train_dataloader`` is not specified.
+        spin_dataloaders (bool, optional): If ``True``, dataloaders will be spun up to the current timestamp.
+            This ensures dataloaders continue from the same batch when resuming training. (default: ``True``)
+
+            .. note:: Spinning dataloaders can be potentially very slow. If resuming from a checkpoint where
```
dakinggg, line 425: This note is a bit unclear to me. As is, it says that if you don't care about determinism, you're ok setting spin to False. This is somewhat incomplete information. Because if you're data order _is_ deterministic, you're not just not getting determinism, you're just only training on the first X samples (and repeating them however many times you restart). I'd rewrite to explain this a bit more thoroughly.

tests/utils/test_autolog_hparams.py
```diff
@@ -103,6 +103,7 @@ def test_extract_hparams_trainer():
         'train_dataloader': 'DataLoader',
         'train_dataloader_label': 'train',
         'train_subset_num_batches': -1,
+        'spin_dataloaders': True,
```
dakinggg, line 106: can you write a real test please?

composer/trainer/trainer.py
```diff
@@ -419,6 +419,11 @@ class Trainer:
             This setting will end each epoch early to avoid additional training that will not be profiled.
 
             This parameter is ignored if ``train_dataloader`` is not specified.
+        spin_dataloaders (bool, optional): If ``True``, dataloaders will be spun up to the current timestamp.
```
dakinggg, line 422: Please explain here the distinction between "spinning" and "resuming"

dakinggg, line 422: suggest `should_spin_dataloaders`

composer/trainer/trainer.py
```diff
@@ -1883,7 +1890,8 @@ def _train_loop(self) -> None:
 
         use_grad_scaling = self._use_grad_scaling(self.state.precision, self.state.scaler)
 
-        self._spin_dataloaders()
+        if self.spin_dataloaders:
+            self._spin_dataloaders()
```
dakinggg, line 1894: suggest renaming this to `_forward_dataloaders_to_current_epoch`, `_set_dataloader_rng_to_current_epoch`, or something similar.

dakinggg (commented):
Will approve once there is a test. Implementation looks good. Left some comments on naming and docs clarity.

composer/trainer/trainer.py
```diff
@@ -419,6 +419,15 @@ class Trainer:
...
             This parameter is ignored if ``train_dataloader`` is not specified.
+        spin_dataloaders (bool, optional): If ``True``, dataloaders will be spun up to the current timestamp
+            by skipping samples which have already been trained on if a dataloader does not have a way to
+            resume from the current batch. This ensures dataloaders continue from the same batch when resuming
+            training. (default: ``True``)
+
+            .. note:: Spinning dataloaders can be potentially very slow but is required to skip samples which
+                have already been trained on. If it is acceptable to repeat samples when resuming training,
+                it is possible to resume faster by setting ``spin_dataloaders=False``. This may have severe
+                performance implications and is generally not recommended.
```
dakinggg, line 430:
```suggestion
                performance implications and is generally not recommended unless you confidently understand the implications.
```

composer/trainer/trainer.py
```diff
@@ -419,6 +419,15 @@ class Trainer:
             This setting will end each epoch early to avoid additional training that will not be profiled.
 
             This parameter is ignored if ``train_dataloader`` is not specified.
+        spin_dataloaders (bool, optional): If ``True``, dataloaders will be spun up to the current timestamp
+            by skipping samples which have already been trained on if a dataloader does not have a way to
```
dakinggg, line 423:
```suggestion
            by skipping samples. If the dataloader has a way to directly
```

composer/trainer/trainer.py
```diff
@@ -419,6 +419,15 @@ class Trainer:
...
+            resume from the current batch. This ensures dataloaders continue from the same batch when resuming
```
dakinggg, line 424:
```suggestion
            resume from the current batch, no spinning will occur in any case, and this argument will be a no-op. This ensures dataloaders continue from the same batch when resuming
```

dakinggg (approved):

PR #2284, opened 2023-06-05
fix get file(overwite=True) to properly handle pre-existing files

bmosaicml:
# What does this PR do?

Fixes a bug in which _get_file(overwrite=True) would fail if the destination path already exists. Ideally, we would instead delete anything at the existing destination path and overwrite it. This created bugs because ICL datasets rely on this utility and was failing if multiple runs in a row tried to write to the same destination. https://github.com/mosaicml/composer/blob/9b2d5ec34b6c26651c6bdae526dc849a652743cc/composer/datasets/in_context_learning_evaluation.py#LL138C8-L140C72

# Before submitting


mvpatel2000 (approved):
LGTM! Thanks for the bugfix

PR #2285, opened 2023-06-05
Delete symlinks when not saving checkpoints locally

rishab-partha:
# What does this PR do?

Add unit tests and functionality so that we delete symlinks when there are no checkpoints that should be saved locally.

[CO-2112](https://mosaicml.atlassian.net/browse/CO-2112)

[CO-2112]: https://mosaicml.atlassian.net/browse/CO-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -390,7 +390,7 @@ class CheckpointSaver(Callback):  # noqa: D101
         if not saved_path:  # not all ranks save
             return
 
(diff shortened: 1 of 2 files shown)


eracah (approved):
We pair programmed this one, so no surprises here. LGTM

PR #2288, opened 2023-06-06
Monolithic Loading

mvpatel2000:
# What does this PR do?

Adds support for loading monolithic ckpts on only rank 0, which decreases ram usage by 8x. 

[CO-2155](https://mosaicml.atlassian.net/browse/CO-2155)

[CO-2155]: https://mosaicml.atlassian.net/browse/CO-2155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer/core/state.py
```diff
@@ -983,11 +986,15 @@ def load_model_state(
             torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(state_dict['model'], 'module.')
 
         try:
-            if self.fsdp_enabled and self.fsdp_state_dict_type is not None:
-                with fsdp_state_dict_type_context(self.model, state_dict_type=self.fsdp_state_dict_type):
+            # Load model if it exists. For FSDP monolith checkpoints, the model does not exist on ranks > 0
+            if 'model' in state_dict:
+                if self.fsdp_enabled and self.fsdp_state_dict_type is not None:
+                    with fsdp_state_dict_type_context(self.model, state_dict_type=self.fsdp_state_dict_type):
+                        missing_keys, unexpected_keys = self.model.load_state_dict(state_dict['model'], strict=strict)
```
eracah, line 993: are we sure this works. I think FSDP would hang because all ranks have to synchronize for a `with fsdp_state_dict_type_context(self.model, state_dict_type=self.fsdp_state_dict_type)` call. I don't think you need it if the model has not been wrapped with fsdp

mvpatel2000:
Requires: use orig params false

mvpatel2000:
TODO: Test with torch 1.13

composer/core/state.py
```diff
@@ -1017,29 +1085,36 @@ def load_optim_state(self, state_dict: Dict[str, Any]):
...
                 if version.parse(torch.__version__) < version.parse('1.13.0'):
                     raise RuntimeError('To use FSDP with Composer, you must use torch>=1.13.0.')
                 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
                 log.debug(f'Loading FSDP optimizer with fsdp_state_dict_type={self.fsdp_state_dict_type}')
-                if not using_torch_2():
-                    optim_state_dict = _legacy_optim_state_dict_to_load(optim_state_dict=optim_state_dict,
-                                                                        model=self.model,
-                                                                        optim=optimizer,
-                                                                        state_dict_type=self.fsdp_state_dict_type)
+                # Loading FSDP monolith on rank 0 only requires FSDP.scatter_full_optim_state_dict
```
bcui19, line 1100: @mvpatel2000 do you want to add a TODO here for when pytorch fixes?

tests/trainer/test_checkpoint.py
```diff
@@ -915,6 +922,97 @@ def test_resumption(
...
+        model_2_init_device: str,
+        tmp_path: pathlib.Path,
+    ):
+        save_interval = '1ba'
+        save_filename = 'ba{batch}-rank{rank}.pt'
+        resume_file = 'ba1-rank{rank}.pt'
+        final_checkpoint = 'latest-rank{rank}.pt'
+        fsdp_config = {
+            'use_orig_params': use_orig_params,
+            'sync_module_states': sync_module_states,
```
bcui19, line 959: do you want to explicitly set `state_dict_type='full'` here? In case we change a default later on?

bcui19 (approved):
LGTM! Left vey minor comments. This is huge

composer/core/state.py
mvpatel2000: I have a JIRA tracking. I'll leave the todo out of the code

PR #2302, opened 2023-06-13
Fix tests for wandb and mlflow loggers

b-chu:
# What does this PR do?

This PR fixes several logger tests. test_mlflow_logging_works will use cpu when both cpu and gpu are available. test_wandb_artifacts is reenabled for offline mode and mocks the wandb artifact.

[CO-1021](https://mosaicml.atlassian.net/browse/CO-1021)

[CO-1021]: https://mosaicml.atlassian.net/browse/CO-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/loggers/test_mlflow_logger.py
+++ b/tests/loggers/test_mlflow_logger.py
@@ -12,6 +12,7 @@ from torch.utils.data import DataLoader
 from composer.loggers import MLFlowLogger
 from composer.trainer import Trainer
 from tests.common.datasets import RandomImageDataset
+from tests.common.markers import device
 from tests.common.models import SimpleConvModel
 
 
@@ -64,7 +65,8 @@ def test_mlflow_experiment_set_up_correctly(tmp_path):
     test_mlflow_logger.post_close()
 
 
-def test_mlflow_logging_works(tmp_path):
+@device('cpu')
+def test_mlflow_logging_works(tmp_path, device):
     mlflow = pytest.importorskip('mlflow')
     mlflow_uri = tmp_path / Path('my-test-mlflow-uri')
     test_mlflow_logger = MLFlowLogger(tracking_uri=mlflow_uri)
@@ -79,7 +81,8 @@ def test_mlflow_logging_works(tmp_path):
                       train_dataloader=DataLoader(RandomImageDataset(size=dataset_size), batch_size),
                       eval_dataloader=DataLoader(RandomImageDataset(size=dataset_size), batch_size),
                       max_duration=f'{num_batches}ba',
-                      eval_interval=eval_interval)
+                      eval_interval=eval_interval,
+                      device=device)
     trainer.fit()
 
     run_info = mlflow.active_run().info
--- a/tests/loggers/test_wandb_logger.py
+++ b/tests/loggers/test_wandb_logger.py
@@ -7,22 +7,34 @@ import json
 import os
 import pathlib
 import pickle
+import shutil
 import uuid
 from pathlib import Path
 from typing import Sequence, Type
 
 import pytest
 import torch
+from _pytest.monkeypatch import MonkeyPatch
 from torch.utils.data import DataLoader
 
 from composer.core import Callback, Engine, Event, State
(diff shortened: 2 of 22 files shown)


tests/loggers/test_mlflow_logger.py
```diff
@@ -64,6 +64,7 @@ def test_mlflow_experiment_set_up_correctly(tmp_path):
     test_mlflow_logger.post_close()
 
 
+@pytest.mark.cpu
```
dakinggg, line 67: what was the issue here/why do we only want to run this on cpu?

tests/loggers/test_wandb_logger.py
```diff
@@ -259,32 +269,35 @@ def test_wandb_is_pickleable_when_disabled(dummy_state: State):
...
 @pytest.mark.world_size(2)
 @pytest.mark.parametrize('rank_zero_only', [True, False])
-@pytest.mark.skip('This test needs to be refactored to use a Mock API interface.')
-def test_wandb_artifacts(rank_zero_only: bool, tmp_path: pathlib.Path, dummy_state: State):
-    """Test that wandb artifacts logged on rank zero are accessible by all ranks."""
-    pytest.importorskip('wandb', reason='wandb is optional')
+def test_wandb_artifacts(monkeypatch: MonkeyPatch, rank_zero_only: bool, tmp_path: pathlib.Path, dummy_state: State):
+    """Test that wandb artifacts logged on rank zero are accessible."""
+    from wandb import Api
+    os.environ['WANDB_MODE'] = 'offline'
```
dakinggg, line 275: I think at the end of this tests you want to set the env var back to whatever it was initially?

tests/loggers/test_wandb_logger.py
```diff
@@ -296,14 +309,18 @@ def test_wandb_artifacts(rank_zero_only: bool, tmp_path: pathlib.Path, dummy_sta
...
+    monkeypatch.setattr(Api, "artifact", mock_artifact)
+
     def _validate_wandb_artifact():
         wandb_logger.download_file(wandb_artifact_name, str(downloaded_wandb_artifact_path))
         with open(downloaded_wandb_artifact_path, 'r') as f:
             assert f.read() == 'hello!'
-
-    _validate_wandb_artifact()
+            
+    if dist.get_global_rank() == 0:
```
dakinggg, line 325: the test docstring currently says that this test tests that an artifact uploaded from rank 0 is accessible on all ranks. doesn't this check make the test not do that?

tests/loggers/test_mlflow_logger.py
b-chu: When fetching the params, you get `num_gpus_per_node` instead of `num_cpus_per_node` which causes an AssertionError.

dakinggg: ah ok, thats fine, thanks! Since its marked as cpu only now, can you please pass `device='cpu'` to the `Trainer` construction?

tests/loggers/test_wandb_logger.py
b-chu: The artifact can't be accessed in offline mode, so I'm mocking the download portion by copying the file from the upload location to the download location. The file system is local to rank 0 though (since it was uploaded from rank 0) so only rank 0 will have access to the download. I also changed the docstring, but let me know if there's another way to test this on all ranks.

tests/loggers/test_mlflow_logger.py
b-chu: Done.

tests/loggers/test_wandb_logger.py
dakinggg: yeah, id recommend syncing the artifact path between ranks, since they all have access to the same file system. they will just have different `tmp_path`s

b-chu: Done.

dakinggg (approved):
Thanks for sorting through this!

PR #2313, opened 2023-06-16
Monolithic loading improvements

mvpatel2000:
# What does this PR do?

1. Raises ValueError if user has mixed initialization without syncing module states
2. Patches error when loading monolith ckpt with optimizer differing from ckpt
3. Patches error where user specifies load rank 0 only but then doesn't actually pass a load path

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -523,7 +523,7 @@ class State(Serializable):
             dist.all_reduce(all_ranks_meta, reduce_operation='MIN')
             any_ranks_meta = self.device.tensor_to_device(torch.tensor([rank_on_meta], dtype=torch.uint8))
             dist.all_reduce(any_ranks_meta, reduce_operation='MAX')
-            if all_ranks_meta.item() == any_ranks_meta.item() == 1:
+            if all_ranks_meta.item() != any_ranks_meta.item():
                 raise ValueError('Detected mixed initialization where some ranks have model on cpu or gpu and '
                                  'some ranks are on meta. Either keep all ranks on the same device or set '
                                  "fsdp_config['sync_module_states'] = True. Otherwise, some weights may be randomly "
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -516,6 +516,19 @@ class State(Serializable):
             if error_message != '':
                 raise ValueError(error_message)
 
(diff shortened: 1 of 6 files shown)


composer/core/state.py
```diff
@@ -516,6 +516,19 @@ def __init__(
...
                 raise ValueError(error_message)
 
+        if self.fsdp_enabled and self.fsdp_config is not None and self.fsdp_config['sync_module_states'] == False:
+            # Broadcast rank 0 meta check to all ranks so error can be raised on all ranks
+            rank_on_meta = 1 if next(model.parameters()).device.type == 'meta' else 0
+            all_ranks_meta = self.device.tensor_to_device(torch.tensor([rank_on_meta], dtype=torch.uint8))
+            dist.all_reduce(all_ranks_meta, reduce_operation='MIN')
+            any_ranks_meta = self.device.tensor_to_device(torch.tensor([rank_on_meta], dtype=torch.uint8))
+            dist.all_reduce(any_ranks_meta, reduce_operation='MAX')
+            if all_ranks_meta.item() == any_ranks_meta.item() == 1:
```
dakinggg, line 526: isn't this supposed to be `if any_ranks_meta and not all_ranks_meta`?

composer/core/state.py
```diff
@@ -1076,8 +1089,10 @@ def load_model_state(
             assert self.fsdp_config is not None
             log.info('Wrapping model with FSDP after loading model_state.')
             from composer.trainer.dist_strategy import prepare_fsdp_module
+            log.info('Imported file')
```
dakinggg, line 1092: either delete this one or making it a bit more descriptive?

composer/trainer/dist_strategy.py
```diff
@@ -208,6 +208,7 @@ def prepare_fsdp_module(
         device (Device): The device being used by the Trainer.
         auto_microbatching (bool, optional): Whether or not auto microbatching is enabled.
     """
+    log.info('start prepare_fsdp_module')
```
dakinggg, line 211:
```suggestion
    log.info('Start prepare_fsdp_module')
```

composer/core/state.py
dakinggg: still wrong i think, both 0 should be fine

dakinggg (approved):
Lgtm

PR #2320, opened 2023-06-20
In-line group to avoid OOM

mvpatel2000:
# What does this PR do?

In-line group to avoid extra ref. This avoids spiking peak GPU memory usage during wrapping. @sashaDoubov  will test.

[CO-2195](https://mosaicml.atlassian.net/browse/CO-2195)

[CO-2195]: https://mosaicml.atlassian.net/browse/CO-2195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sashaDoubov:
LGTM, we get the same loss curve when using multiple param groups with and without this change, memory usage is hard to debug though. 
image
image

sashaDoubov (approved):

PR #2321, opened 2023-06-20
Update local rank 0 to be elastic

mvpatel2000:
# What does this PR do?

Compute local ranks from global - local instead of using local world size. This lets us do different nodes with variable sizes

We're not going to test full elastic since it's not a priority. Just patching the obvious bugs for now.

[CO-2151](https://mosaicml.atlassian.net/browse/CO-2151)

[CO-2151]: https://mosaicml.atlassian.net/browse/CO-2151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eracah (approved):
LGTMD

PR #2322, opened 2023-06-20
Pin Apex

mvpatel2000:
# What does this PR do?

We currently install APEX off dev. This is causing issues, especially as they now require no build isolation. Moving to a pinned version before they introduce pyproject toml.

This is breaking GPU tests currently

Staged image test works: 
image


bandish-shah (approved):
LGTM

PR #2334, opened 2023-06-28
[fix] Remove wandb from the `ImageVisualizer`

antoinebrl:
# What does this PR do?

The `ImageVisualizer` doesn't need `wanbd`. This is the responsibility of the logger.

We are trying to use a different logger and the  `ImageVisualizer` crashes because `wanbd` is not installed. 

# Before submitting

### DIFF
--- a/composer/callbacks/image_visualizer.py
+++ b/composer/callbacks/image_visualizer.py
@@ -9,7 +9,6 @@ import torch
 from composer.core import Callback, State, Time, TimeUnit
 from composer.loggers import Logger
 from composer.loss.utils import infer_target_type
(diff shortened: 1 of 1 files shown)


eracah (approved):
Good catch! LGTM. What logger are you planning to use? CometML?

eracah:
Thanks @antoinebrl ! I think you just need to remove line 12: `from composer.utils import MissingConditionalImportError` and then the code quality checks should pass

antoinebrl:
Thanks! I removed L12.

We use Neptune.ai for experiment tracking.

Any chance this get released as a patch?

PR #2335, opened 2023-06-29
Add precision config arg for FP8

julian-q:
# What does this PR do?

This PR introduces a new field to the trainer `State` that configures the FP8 scaling recipe used by Transformer Engine. Previously, these values were [hardcoded](https://github.com/mosaicml/composer/blob/b73c776725c05fbd5b7cf43b63a8ca85ed455af0/composer/core/precision.py#L73-L74), but now we can experiment with different parameters by supplying a `precision_config` dictionary to the `Trainer`. The dictionary should contain init parameters for the [`DelayedScaling`](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html?highlight=delayedscaling#transformer_engine.common.recipe.DelayedScaling) class from Transformer Engine.

Implements https://mosaicml.atlassian.net/browse/CO-2210

### DIFF
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -6,7 +6,7 @@
 import contextlib
 import os
 import textwrap
-from typing import Generator, Union, Dict, Any
+from typing import Any, Dict, Generator, Union
 
 import torch
 
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -44,7 +44,8 @@ def get_precision_context(precision: Union[str, Precision],
 
     Args:
         precision (str | Precision): Precision for the context
-        precision_config (Dict[str, Any]): Config for FP8 scaling strategy. See parameters for `DelayedScaling <https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html?highlight=delayedscaling#transformer_engine.common.recipe.DelayedScaling>`_.
+        precision_config (Dict[str, Any]): Config for FP8 scaling strategy. See parameters for
+            `DelayedScaling <https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html?highlight=delayedscaling#transformer_engine.common.recipe.DelayedScaling>`_.
     """
     precision = Precision(precision)
     if precision == Precision.FP32:
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -44,7 +44,7 @@ def get_precision_context(precision: Union[str, Precision],
 
     Args:
         precision (str | Precision): Precision for the context
-        precision_config (Dict[str, Any]): Config for FP8 scaling strategy. See parameters for
+        precision_config (Optional[Dict[str, Any]]): Config for FP8 scaling strategy. See parameters for
             `DelayedScaling <https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/common.html?highlight=delayedscaling#transformer_engine.common.recipe.DelayedScaling>`_.
     """
     precision = Precision(precision)
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -68,12 +68,10 @@ def get_precision_context(precision: Union[str, Precision],
             yield
     elif precision == Precision.AMP_FP8:
         if te_installed and torch.cuda.get_device_capability()[0] > 8:
-            from transformer_engine.common.recipe import DelayedScaling, Format
+            from transformer_engine.common.recipe import DelayedScaling
 
             if precision_config is None:
                 precision_config = {}
-            if 'fp8_format' in precision_config and isinstance(precision_config['fp8_format'], str):
-                precision_config['fp8_format'] = Format[precision_config['fp8_format']]
             fp8_recipe = DelayedScaling(**precision_config)
             with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
                 yield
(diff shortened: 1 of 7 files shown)


composer/core/precision.py
```diff
@@ -38,11 +38,12 @@ class Precision(StringEnum):
 
 
 @contextlib.contextmanager
-def get_precision_context(precision: Union[str, Precision]) -> Generator[None, None, None]:
+def get_precision_context(precision: Union[str, Precision], precision_config: Dict[str, Any]) -> Generator[None, None, None]:
```
dskhudia, line 41: Could you default this to None please? I don't want to break stuff which uses this API outside composer.

julian-q: done!

composer/core/precision.py
```diff
@@ -67,11 +69,11 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
...
             from transformer_engine.common.recipe import DelayedScaling, Format
 
-            # These default values for fp8_recipe are taken from NVidia's docs. We may want to change
-            # these once we get a chance to do more convergence experiments.
-            # https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#id1
-            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
-            fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo='max')
+            if precision_config is None:
+                precision_config = {}
+            if 'fp8_format' in precision_config and isinstance(precision_config['fp8_format'], str):
```
dskhudia, line 74: If it's none, let us init it to previous defaults. Update the docs for the same.

julian-q: Do you want to let the `DelayedScaling` class decide the defaults? Previously it looks like we defaulted the arguments to the default ones from NVIDIA's docs: https://github.com/mosaicml/composer/blob/b73c776725c05fbd5b7cf43b63a8ca85ed455af0/composer/core/precision.py#L70-L72

Right now it looks like the default parameters for the class match what we had before, except for `amax_history_len`, which is `1024` instead of `16`: https://github.com/NVIDIA/TransformerEngine/blob/6f7853dc9fb5cf4254760ec3994d5a09946aaa81/transformer_engine/common/recipe.py#L127-L129

So it depends on if we agree with the default values TE has already ^

composer/core/precision.py
```diff
@@ -65,13 +68,11 @@ def get_precision_context(precision: Union[str, Precision]) -> Generator[None, N
...
-            from transformer_engine.common.recipe import DelayedScaling, Format
+            from transformer_engine.common.recipe import DelayedScaling
 
-            # These default values for fp8_recipe are taken from NVidia's docs. We may want to change
-            # these once we get a chance to do more convergence experiments.
-            # https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html#id1
-            fp8_format = Format.HYBRID  # E4M3 during forward pass, E5M2 during backward pass
-            fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo='max')
+            if precision_config is None:
+                precision_config = {}
```
dskhudia, line 74: I think we should set it to default values if it's not set.

dskhudia:
> Congratulations on your first PR! Will defer to @dskhudia if this needs more tests.

In terms of tests, this looks great to me.

composer/core/precision.py
```diff
@@ -38,11 +38,14 @@ class Precision(StringEnum):
 
 
 @contextlib.contextmanager
-def get_precision_context(precision: Union[str, Precision]) -> Generator[None, None, None]:
+def get_precision_context(precision: Union[str, Precision],
+                          precision_config: Optional[Dict[str, Any]] = None) -> Generator[None, None, None]:
```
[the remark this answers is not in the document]
julian-q: I initially had `precision_config: Dict[str, Any] = {}` but pre-commit says we shouldn't init parameters to mutable objects

composer/core/state.py
```diff
@@ -50,14 +50,14 @@ def fsdp_state_dict_type_context(module: torch.nn.Module, state_dict_type: str =
     """Context manager for materializing or loading an fsdp module's state dict.
 
     Args:
-        module (torch.nn.Module): The torch module that you want to call `state_dict()`
-            or `load_state_dict()` on.
+        module (torch.nn.Module): The torch module that you want to call ``state_dict()``
```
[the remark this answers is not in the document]
julian-q: I was trying to fix doctest errors but yeah I think this isn't necessary anymore! Thank you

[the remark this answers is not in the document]
julian-q: reverted

composer/core/precision.py
dskhudia: This is fine. I misunderstood.

dskhudia (approved):

composer/core/precision.py
```diff
@@ -38,11 +38,14 @@ class Precision(StringEnum):
...
     """Returns a context manager to automatically cast to a specific precision.
 
     Args:
         precision (str | Precision): Precision for the context
+        precision_config (Dict[str, Any]): Config for FP8 scaling strategy. See parameters for
```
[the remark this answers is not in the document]
julian-q: thanks I added the `Optional` for all the docs on `precision_config`

julian-q:
I noticed some errors once I was able to run on an H100 and made some minor fixes to the test I added

PR #2338, opened 2023-06-30
Set offload_to_cpu True for state_dict_type=sharded

eracah:
# What does this PR do?

sets the default for sharded and local state dicts to offload_to_cpu=True. This helps avoid OOMs for large models when saving sharded checkpoints

## Testing
Ran manual test of saving 30B checkpoints

https://github.com/mosaicml/llm-foundry/issues/367

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -90,6 +90,7 @@ def fsdp_state_dict_type_context(module: torch.nn.Module, state_dict_type: str =
         fsdp_state_dict_type = StateDictType.SHARDED_STATE_DICT
         state_dict_config = ShardedStateDictConfig()
         if using_torch_2():
+            state_dict_config = ShardedStateDictConfig(offload_to_cpu=True)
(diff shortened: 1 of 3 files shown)


mvpatel2000 (approved):

composer/core/state.py
```diff
@@ -524,6 +525,11 @@ def __init__(
         if self.fsdp_config is not None:
             self.sharded_ckpt_prefix_dir = self.fsdp_config['sharded_ckpt_prefix_dir']
 
+        if using_torch_2() and self.fsdp_state_dict_type == 'local':
+            raise DeprecationWarning(
+                textwrap.dedent("FSDP state_dict_type='local' is deprecated in torch>=2.0.0"
+                                "Please set fsdp_config['state_dict_type']='sharded' instead."))
```
mvpatel2000, line 531:
```suggestion
                textwrap.dedent("FSDP state_dict_type='local' is deprecated in torch>=2.0.0. "
                                "Please set fsdp_config['state_dict_type']='sharded' instead and will be removed in v0.17"))
```

PR #2347, opened 2023-07-03
Fix test_fsdp_load_old_checkpoint test to fix daily tests

eracah:
# What does this PR do?
fixes `_compare_model_params_between_state_dicts`, so that both tensors that need to be compared are cast to cpu.
This fixes the failed `test_fsdp_load_old_checkpoint` test


hanlint (approved):
LGTM!

PR #2352, opened 2023-07-06
Avoid overwriting seed==0

tbenthompson:
Before this change, if I create a Trainer with `seed=0`, the seed is overwritten with a random seed.


dakinggg (approved):
Thanks!

eracah:
Good catch, @tbenthompson !

PR #2353, opened 2023-07-06
Fix wandb errror with autoresume issue

eracah:
# What does this PR do?
The new wandb release (0.15.5) changes it's error messages for artifact downloading failures. Our `WandBLogger` is designed to be EAFP instead of LBYL, so it relies on catching an exact error to prevent autoresume from trying to download checkpoints from wandb that aren't actually there.
This PR makes it so the check is not so specific (i.e. it catches all wandb CommErrors instead of just ones with specific messages). As a result, the EAFP code will then work with the new wandb install.

### DIFF
--- a/.github/secrets/exclude.yaml
+++ b/.github/secrets/exclude.yaml
@@ -1,2 +1,3 @@
 .git/.*
 docs/source/_static/js/posthog.js
+docs/_build/html/_static/js/posthog.js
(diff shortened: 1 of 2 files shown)


dakinggg (approved):

antoinebrl:
Unfortunately this didn't make it into v0.15.1. Will it be part of the next release?

PR #2362, opened 2023-07-13
raise min mcli

mvpatel2000:
# What does this PR do?

MosaicML logger auto-adding requires version 0.4.12 in order to autoread key and avoid running mcli init for adding API key. 

[CO-2228](https://mosaicml.atlassian.net/browse/CO-2228)


nancyhung (approved):
LGTM

PR #2366, opened 2023-07-14
Move pydantic pin to deepspeed

mvpatel2000:
# What does this PR do?

Only deepspeed has errors with pydantic 2. Moving the pin down to there as we don't actually use in composer normally


dakinggg (commented):
How do we know it was only deepspeed that has issues?

mvpatel2000:
The pydantic error comes from Deepseed. They failed to pin and error. We don't use pydantic
image

dakinggg (approved):

PR #2377, opened 2023-07-20
Respect MLFLow experiment environment variable

aspfohl:
This removes the default of naming all MLFLow experiments `my-mlflow-experiment`, and respects the environment variable MLFLOW_EXPERIMENT_NAME https://mlflow.org/docs/latest/python_api/mlflow.html?highlight=mlflow_experiment_name#mlflow.start_run

Slightly updates the docs about default behavior for runs and experiments


composer/loggers/mlflow_logger.py
```diff
@@ -62,7 +67,8 @@ def init(self, state: State, logger: Logger) -> None:
             self.run_name += f'-rank{dist.get_global_rank()}'
 
         if self.experiment_name is None:
-            self.experiment_name = 'my-mlflow-experiment'
+            env_var = mlflow.environment_variables.MLFLOW_EXPERIMENT_NAME.name
```
eracah, line 70: can we also check for `MLFLOW_EXPERIMENT_ID` as well here?

composer/loggers/mlflow_logger.py
```diff
@@ -62,7 +67,8 @@ def init(self, state: State, logger: Logger) -> None:
             self.run_name += f'-rank{dist.get_global_rank()}'
```
eracah, line 67: Can we check for the `MLFLOW_RUN_ID` [env variable](https://mlflow.org/docs/latest/python_api/mlflow.environment_variables.html?highlight=environment%20variables#mlflow.environment_variables.MLFLOW_RUN_ID) as well? Should a similar change to line 63

eracah (commented):
Good catch. Thanks for doing this, @aspfohl ! Can you also add some unit tests to https://github.com/mosaicml/composer/blob/dev/tests/loggers/test_mlflow_logger.py#L43-L48

Skylion007 (approved):

eracah (approved):
Nice LGTM

PR #2378, opened 2023-07-20
regex changed in time.py

megha95:
# What does this PR do?
Made changes to composer/core/time.py: Currently, regex used to parse timestring and create Time object (to decide "max_duration" param for training) cannot handle cases such as '3_000_000_000tok'. So, we modified the regex accordingly. Now, it will divide timestring into 2 parts: value and unit of time and then use float() fn of python to convert value to a float. This way python directly handles parsing first part of string.

Some example timestrings that now will be supported: '4_000tok', '3_00_0sp', '3_0e10tok', etc.

[CO-2238](https://mosaicml.atlassian.net/browse/CO-2238)

# Before submitting

[CO-2238]: https://mosaicml.atlassian.net/browse/CO-2238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer/core/time.py
+++ b/composer/core/time.py
@@ -44,7 +44,7 @@ class TimeUnit(StringEnum):
     DURATION = 'dur'
 
 
-# regex for parsing time string, matches timeunit and chars prior to unit is value
+# regex for parsing time string, matches timeunit and chars prior to unit as value
 _TIME_STR_REGEX = re.compile(r'^(.+)(' + r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
                              flags=re.IGNORECASE)
 
--- a/composer/core/time.py
+++ b/composer/core/time.py
@@ -45,11 +45,11 @@ class TimeUnit(StringEnum):
 
 
 # regex for parsing integers / decimals / scientific notation
-_NUM_REGEX = r'-?[\d.]+(?:e-?\d+)?'
+_NUM_REGEX = r'.+'
 
 # regex for parsing a time string.
-_TIME_STR_REGEX = re.compile(r'^(?:' + r'|'.join(fr'(?:({_NUM_REGEX})({time_unit.value}))' for time_unit in TimeUnit) +
-                             r')$',
+_TIME_STR_REGEX = re.compile(r'^(' + fr'{_NUM_REGEX}' + ')(' +
+                             r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
                              flags=re.IGNORECASE)
 
 TValue = TypeVar('TValue', int, float)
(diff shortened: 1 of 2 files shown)


composer/core/time.py
```diff
@@ -45,11 +45,11 @@ class TimeUnit(StringEnum):
...
 
 # regex for parsing integers / decimals / scientific notation
-_NUM_REGEX = r'-?[\d.]+(?:e-?\d+)?'
+_NUM_REGEX = r'.+'
 
 # regex for parsing a time string.
-_TIME_STR_REGEX = re.compile(r'^(?:' + r'|'.join(fr'(?:({_NUM_REGEX})({time_unit.value}))' for time_unit in TimeUnit) +
-                             r')$',
+_TIME_STR_REGEX = re.compile(r'^(' + fr'{_NUM_REGEX}' + ')(' +
+                             r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
```
mvpatel2000, line 52: Can you please update the comments to be a bit more clear here?

It looks like the `_NUM_REGEX` is adjusted to match everything besides the time_unit. In that case, can we just get rid of that value and in-line it into the `_TIME_STR_REGEX` since it's been simplified? 

Can we also update the comment to be clear on how it works? Something like we match the unit and assume the rest is the number?

megha95: Sure, corrected it.

composer/core/time.py
```diff
@@ -44,12 +44,8 @@ class TimeUnit(StringEnum):
...
 
 
-# regex for parsing integers / decimals / scientific notation
-_NUM_REGEX = r'-?[\d.]+(?:e-?\d+)?'
-
-# regex for parsing a time string.
-_TIME_STR_REGEX = re.compile(r'^(?:' + r'|'.join(fr'(?:({_NUM_REGEX})({time_unit.value}))' for time_unit in TimeUnit) +
-                             r')$',
+# regex for parsing time string, matches timeunit and chars prior to unit is value
+_TIME_STR_REGEX = re.compile(r'^(.+)(' + r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
```
mvpatel2000, line 48:
```suggestion
# regex for parsing time string, matches timeunit and chars prior to unit as value
_TIME_STR_REGEX = re.compile(r'^(.+)(' + r'|'.join([fr'{time_unit.value}' for time_unit in TimeUnit]) + r')$',
```

minor nit

mvpatel2000 (approved):
LGTM

mvpatel2000 (approved):

PR #2379, opened 2023-07-20
Isolate env var in unit tests

mvpatel2000:
# What does this PR do?

Currently, unit tests all read the environment variable and use the same run name. This breaks things. Instead, we should hide the env var.

Along for ride, a unit test


eracah (approved):
LGTM. it worked on an interactive session for me

PR #2380, opened 2023-07-21
Fix: Ensure device after metric reset

andre-bauer:
# What does this PR do?

Rests metrics before device check.
In a metric where one of the state variables is a `list[str]` this otherwise calls `.to` on the strings. Instead now the states are reseted before. 

# Before submitting


mvpatel2000 (commented):
Thanks for the PR! Can you please give a bit more detail on the scenario where this would break before this PR? Not sure I completely understand the description

PR #2383, opened 2023-07-21
Change transformers

rishab-partha:
# What does this PR do?

Upgrade transformers to 4.31.0

LLaMa-2-70B compatibility


dakinggg (approved):
All good, just needs to pass CI

Skylion007 (approved):

composer/models/huggingface.py
```diff
@@ -170,6 +170,9 @@ def load_huggingface_tokenizer_from_saved_state(
                             for line in saved_content['content']:
                                 _tmp_file.write(line)
                                 _tmp_file.write('\n')
+                    elif saved_content['file_extension'] == '.py':
```
dakinggg, line 173: This must have gotten accidentally deleted at some point, because I originally fixed in https://github.com/mosaicml/composer/pull/2279/files

composer/models/huggingface.py
```diff
@@ -180,6 +183,7 @@ def load_huggingface_tokenizer_from_saved_state(
                         s.load_from_serialized_proto(saved_content['content'])
                         with open(tokenizer_file_path, 'wb') as _tmp_file:
                             _tmp_file.write(s.serialized_model_proto())
+
                 hf_tokenizer = transformers.AutoTokenizer.from_pretrained(_tmp_dir, trust_remote_code=True)
```
dakinggg, line 187: the trust remote code here was also supposed to parameterized in that same pr (https://github.com/mosaicml/composer/pull/2279/files), and somehow got deleted as well. Could you fix that here please?

composer/models/huggingface.py
rishab-partha: ah makes sense, it must have still worked for some reason w/ transformers 4.30.2 but broken w/ transformers 4.31

dakinggg: Yeah, if you read the description on that pr, it only affected an older version of transformers as far as i can tell, but maybe they changed it back to the way it was before? not exactly sure, but either way, what you've added is the right fix.

composer/models/huggingface.py
rishab-partha: done

dakinggg (commented):
Thanks

PR #2384, opened 2023-07-21
Deprecate Fused LayerNorm

nik-mosaic:
[WIP]

# What does this PR do?

Deprecates Fused LayerNorm. Fused LayerNorm has been largely replaced with LowPrecision LayerNorm, which has equivalent speed and performance guarantees.

Please note that Apex is still used in our tests. It is a dependency for T5-small, which is used in multiple places. Thus, removing FLN will not alleviate Apex installation issues.


mvpatel2000 (changes requested):
Must be deprecated first before removal

PR #2386, opened 2023-07-23
[fix] Use Enum value when logging hyper-parameters

antoinebrl:
# What does this PR do?

When logging hyper-parameters only built-in types were supported. Otherwise the class name was used. This cause a loss of information when using enum. For example when logging precision `precision: Precision`.

This contribution takes the value field if the object is an Enum. Resulting in `precision: amp_fp16`

### DIFF
--- a/tests/utils/test_autolog_hparams.py
+++ b/tests/utils/test_autolog_hparams.py
@@ -8,8 +8,13 @@ from composer.algorithms import EMA
 from composer.callbacks import SpeedMonitor
 from composer.loggers import InMemoryLogger
 from composer.trainer import Trainer
-from composer.utils import (convert_flat_dict_to_nested_dict, convert_nested_dict_to_flat_dict, extract_hparams,
-                            StringEnum, using_torch_2)
+from composer.utils import (
+    StringEnum,
+    convert_flat_dict_to_nested_dict,
+    convert_nested_dict_to_flat_dict,
+    extract_hparams,
+    using_torch_2,
+)
 from tests.common.datasets import RandomClassificationDataset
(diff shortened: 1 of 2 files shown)


eracah (approved):
Nice! Thanks for the contribution, @antoinebrl ! LGTM

antoinebrl:
Sorry there was a formatting issue after I added an import. This should be solved now!

tests/utils/test_autolog_hparams.py
```diff
@@ -6,10 +6,16 @@
...
 from composer.trainer import Trainer
-from composer.utils import (convert_flat_dict_to_nested_dict, convert_nested_dict_to_flat_dict, extract_hparams,
-                            using_torch_2)
+from composer.utils import (
+    StringEnum,
+    convert_flat_dict_to_nested_dict,
+    convert_nested_dict_to_flat_dict,
+    extract_hparams,
+    using_torch_2,
+)
```
mvpatel2000, line 18:
```suggestion
from composer.utils import (StringEnum, convert_flat_dict_to_nested_dict, convert_nested_dict_to_flat_dict,
                            extract_hparams, using_torch_2)
```

tests/utils/test_autolog_hparams.py
```diff
@@ -42,6 +48,9 @@ class Bar:
         def __init__(self):
             self.local_hparams = {'m': 11}
 
+    class Baz(StringEnum):
+        A = "abc"
```
mvpatel2000, line 52:
```suggestion
        A = 'abc'
```

mvpatel2000 (approved):

PR #2395, opened 2023-07-26
Add ruff pre-commit hook and apply some sane autofixes

Skylion007:
# What does this PR do?

* Adds ruff to pre-commit and enable a couple of rules. The full rule list can be found here if you want to enable more. When some additional configuration, this could replace a bunch of pre-commit hooks like isort / pydocstyle etc. 
* Benefit is that this linter is fast, implements other linter, and most importantly applies autofixes.

# Before submitting
- [X] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-    "image": "mosaicml/pytorch_vision:latest",
-    "remoteUser": "mosaicml",
-    "postCreateCommand": "pip install -e .[all] && pre-commit install"
+ "image": "mosaicml/pytorch_vision:latest",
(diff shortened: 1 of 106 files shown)


mvpatel2000:
@Skylion007 I suggest we discuss this before adding -- possibly at a design meeting?

Skylion007:
I had some informal discusses with @eracah and some other folks in June about this. I also only enabled the least controversial linter code (C4, and PERF), which replaces certain comprehensions etc with their more efficient equivalent if they are semantically identical. This should functionality be a noop and should just make the code (very marginally) more efficient. We have those error codes enabled across the PyTorch repo currently for example

PR #2396, opened 2023-07-26
Save metadata and integration when save_weights_only is set

eracah:
# What does this PR do?
When `save_weights_only` is specified to the Trainer, only the weights are saved and no metadata or integrations are saved. However, when doing inference it is often necessary to have the metadata along with the weights. The only current way to get metadata and integrations is to save everything in the state_dict (including the very large optimizer state). This PR enables saving the weights and the metadata and integrations, so you can save an inference-friendly checkpoint without the extra heft of the optimizer or other (not necessary for inference) state fields.

fix CO-2226

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -11,6 +11,7 @@ import os
 import pathlib
 import tempfile
 import textwrap
+from pathlib import Path
 from typing import Callable, List, Optional, Union
 
 from composer.core import Callback, Event, State, Time, TimeUnit
@@ -18,10 +19,14 @@ from composer.loggers import Logger
 from composer.utils import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, FORMAT_NAME_WITH_DIST_TABLE, PartialFilePath,
                             checkpoint, create_symlink_file, dist, ensure_folder_has_no_conflicting_files,
                             format_name_with_dist, format_name_with_dist_and_time, is_model_deepspeed, reproducibility)
+from composer.utils.checkpoint import _TORCH_DISTRIBUTED_CHECKPOINTS_FILENAME
+from composer.utils.misc import using_torch_2
 
 log = logging.getLogger(__name__)
 
 __all__ = ['CheckpointSaver', 'checkpoint_periodically']
+
+_TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME = '.metadata'
 
 
 def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State, Event], bool]:
@@ -312,7 +317,6 @@ class CheckpointSaver(Callback):  # noqa: D101
 
         self.filename = PartialFilePath(filename.lstrip('/'), folder)
         self.latest_filename = PartialFilePath(latest_filename.lstrip('/'), folder) if latest_filename else None
-
         self.remote_file_name = PartialFilePath(remote_file_name) if remote_file_name else None
         self.latest_remote_file_name = PartialFilePath(latest_remote_file_name) if latest_remote_file_name else None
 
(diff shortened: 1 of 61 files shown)


mvpatel2000 (commented):
What if we always included these in the weights only option? I'm just a little hesitant to add more trainer args...

dakinggg:
@mvpatel2000 the issue with always including metadata is that its kind of an API change. If there is code out there relying on a weights only checkpoint only having one key. That being said, it may be worth making that change, not sure.

composer/trainer/trainer.py
```diff
@@ -1106,6 +1108,11 @@ def __init__(
         self._checkpoint_saver = None
         latest_remote_file_name = None
         if save_folder is not None:
+            if save_weights_only and not save_weights_and_metadata_only:
+                warnings.warn(
+                    'save_weights_only=True only saves weights for now, but will changed to also save metadata and integrations in the future.'
```
dakinggg, line 1113:
```suggestion
                    'save_weights_only=True only saves weights for now, but will be changed to also save metadata and integrations in the future.'
```

dakinggg (commented):
LGTM pending the resolution of Mihir's comment

eracah:
@mvpatel2000, @dakinggg, ok fixed to add metadata saving to the save_weights_only option

composer/trainer/trainer.py
```diff
@@ -1106,6 +1106,11 @@ def __init__(
         self._checkpoint_saver = None
         latest_remote_file_name = None
         if save_folder is not None:
+            if save_weights_only:
+                warnings.warn(
+                    'save_weights_only=True now also saves metadata and integrations! Please adjust your workflow accordingly.'
```
Skylion007, line 1111: Please specify a `stack_level` for warning.warn as is recommended (how far to unroll the traceback when displaying the warning)

composer/trainer/trainer.py
```diff
@@ -1106,6 +1106,11 @@ def __init__(
         self._checkpoint_saver = None
         latest_remote_file_name = None
         if save_folder is not None:
+            if save_weights_only:
+                warnings.warn(
```
mvpatel2000, line 1110: Imo fine to remove? In general warnings.warn means a user has an action they need to do.

Maybe a log.info saying as of version 0.16, save_weights_only=True includes metadata and integrations for easier export to inference. This may break workflows which rely on only one key being present in checkpoint.

docs/source/trainer/checkpointing.rst
```diff
@@ -174,8 +174,9 @@ Saving for Inference
 --------------------
 
 By default, the :class:`.Trainer` stores the entire training state in each checkpoint. If you would like to store
-only the model weights in a checkpoint, set ``save_weights_only=True``.
-
+only the model weights in a checkpoint (plus metadata and integrations), set ``save_weights_only=True``. The metadata includes the information about the enviornment used to train the model and any integrations, such as HuggingFace,
+and thsoe integrations' metadata.
```
mvpatel2000, line 178:
```suggestion
and those integrations' metadata.
```

mvpatel2000 (approved):
Lgtm! Minor comment that we should probably avoid a warnings.warn in this case.

composer/callbacks/checkpoint_saver.py
```diff
@@ -266,6 +266,7 @@ class CheckpointSaver(Callback):  # noqa: D101
             This parameter only controls how many checkpoints are kept locally; checkpoints are not deleted from
             remote file systems.
 
+
```
dakinggg, line 269: Remove

composer/trainer/trainer.py
dakinggg: Yeah I'd go with info since the user has no way to get rid of this warning.

tests/trainer/test_checkpoint.py
```diff
@@ -380,6 +380,37 @@ def test_checkpoint_save_sample_interval(self, tiny_bert_tokenizer, save_interva
...
+                          save_filename=save_filename,
+                          save_weights_only=save_weights_only,
+                          save_interval='1ba')
+        trainer.fit()
+        trainer.close()
+        checkpoint_filepath = os.path.join(save_folder, save_filename.format(batch=1))
+        composer_state_dict = torch.load(checkpoint_filepath, map_location='cpu')
+
+        if save_weights_only:
+            assert set(composer_state_dict['state'].keys()) == {'model', 'metadata', 'integrations'}
```
dakinggg, line 407: Could you expand this test just a little bit and make sure the metadata at least is actually present, not just the key?

dakinggg (approved):
Lgtm thanks for taking care of this

PR #2399, opened 2023-07-28
Add code eval secrets to workflows

rishab-partha:
# What does this PR do?

Add code eval secrets to workflows

Fix CPU daily tests


mvpatel2000 (approved):
Can u manually run to verify? Just run mcli command for gpu tests except w daily flag. See logs of GHA gpu action for mcli command

rishab-partha:
Won't this still break? It's an mcloud error for any partial node run

PR #2407, opened 2023-07-31
Fix LOCAL_WORLD_SIZE in pytest

rishab-partha:
# What does this PR do?

Fix CI/CD by changing LOCAL_WORLD_SIZE

Fixing CI/CD


mvpatel2000:
Should we just wait for cloud to fix? I think this failing is a good test for platform broken?

rishab-partha:
@mvpatel2000 Cloud has already fixed, it's an issue on our end w.r.t. to default dist values.

dakinggg:
@mvpatel2000 This is because the ci is trying to run `make test-gpu` (note: _not_ `make test-dist-gpu`) with `LOCAL_WORLD_SIZE=2`, which is invalid and causes `initialize_dist` to fail. @rishab-partha I think we should actually fix this in the makefile, not the `mcli_pytest.py` file. so `make test` and `make test-gpu` should set the `LOCAL_WORLD_SIZE` to 1. wdyt?

rishab-partha:
fixed!

dakinggg (approved):
@mvpatel2000 will give you a couple hours to see if you want to reject this for some reason, otherwise will merge to unblock CI

mvpatel2000:
Lgtm. Same w foundry

PR #2409, opened 2023-08-02
Fix bug with elastic sharded checkpointing + autoresume + contract-expand 

eracah:
# What does this PR do?
previously, there was a bug where if you for example using autoresume:
1. saved with 16 ranks
2. loaded on 8 ranks then saved
3. then loaded on 16 ranks
it would error out on step 3 because ranks 8-15 would try to read old symlik files from step 1

All symlink files point to the same directory, so this PR fixes that bug by having all ranks read from the rank 0 symlink
and only rank 0 write a symlink

# Manual tests
tested 16->8->16 autoresume and the bug disappeared

### DIFF
--- a/.github/mcli/mcli_pytest.py
+++ b/.github/mcli/mcli_pytest.py
@@ -82,6 +82,7 @@ if __name__ == '__main__':
         image=args.image,
         integrations=[git_integration],
         command=command,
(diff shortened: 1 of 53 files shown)


mvpatel2000 (approved):
~Has this been manually tested?~ jk I see

PR #2411, opened 2023-08-03
Adding extra condition to avoid running eval_train_metrics

furkanbiten:
# What does this PR do?

When you return empty dictionary in a model within `get_metrics`, at every batch `eval_forward` is being run.

According to my investigation, this is caused by `_eval_train_metrics` within `Trainer` class. Hence, I have added an extra condition to avoid running `_eval_train_metrics` if the `train_metrics` is returned as an empty dictionary.

Otherwise, you need to return None in `get_metrics` within a `ComposerModel` which causes the checkpointing to break.

# Before submitting

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -2462,7 +2462,7 @@ class Trainer:
             self.engine.run_event(Event.AFTER_BACKWARD)
 
             # Use microbatch outputs to update training metrics
-            if self.state.train_metrics is not None:
+            if self.state.train_metrics is not None and len(self.state.train_metrics) != 0:
                 self.state.train_metrics = self._ensure_metrics_device_and_dtype(self.state.train_metrics)
                 self._eval_train_metrics(device_batch)
 


antoinebrl:
Sorry there was a formatting issue after I added an import. This should be solved now!

priba:
@mvpatel2000 any idea if this can get some attention? We are experiencing a big slowdown in our training.

mvpatel2000:
> @mvpatel2000 any idea if this can get some attention? We are experiencing a big slowdown in our training.

Sorry for the delay! We missed this PR going up. Will review sometime today :)

mvpatel2000 (commented):
Thanks for this PR! A few questions:

> We are experiencing a big slowdown in our training. - @priba 

Could you please elaborate on this? By default `eval_forward` should just return the outputs from calling forward on train and not cause any additional passes. From the Composer Model base class:
https://github.com/mosaicml/composer/blob/dev/composer/models/base.py#L162
```
    def eval_forward(
        self,
        batch: Any,
        outputs: Optional[Any] = None,
    ) -> Any:
        """Run the evaluation forward pass.

		By default, it returns the ``outputs`` if they are not None. Otherwise,
		``self(batch)`` is returned.

		Override this method for models that require custom validation logic -- e.g. self-supervised learning.

		Args:
			batch: The dataloader batch.
			outputs (Any, optional): If training, the outputs from the forward pass. Otherwise, None.

		Returns:
			Any: The evaluation outputs.
		"""
        return outputs if outputs is not None else self.forward(batch)
```

> Otherwise, you need to return None in get_metrics within a ComposerModel which causes the checkpointing to break. - @furkanbiten 

Could you please explain how it breaks? It shouldn't break... so if it does, this is a problem! It would be very nice if you could add a unit test in `test_checkpoint.py` to show 1) it breaks before this PR and 2) it fixes after this PR. I am happy to help out here if this is not clear

We will be doing a release soon, and I'd love to get this in if it's a blocker for you all

priba:
@mvpatel2000 in my case, I am running a detection network and `eval_forward` needs to run NMS. This is quite slow and not required for training. 

In the `ComposerModel`, `get_metrics` expect a `Dict[str, Metric]` as a return. By default, it returns an empty dictionary `{}`. In my case, I am running something like this:

```python
        if is_train:
            return {}

        return {"mAP": composer.metrics.MAP()}
```

However, providing  an empty dict will not work with the following comparison: `self.state.train_metrics is not None` See: https://github.com/mosaicml/composer/blob/34e6711cb9bcd0ae9178bf43c4fcea9215c7cdf6/composer/trainer/trainer.py#L2465 

After this `if`, this code is called: https://github.com/mosaicml/composer/blob/34e6711cb9bcd0ae9178bf43c4fcea9215c7cdf6/composer/trainer/trainer.py#L2161

Returning `None` might be a solution, but then the documentation must change.

mvpatel2000 (approved):
LGTM! Thank you for fixing this. I will also update the documentation soon

PR #2412, opened 2023-08-03
Log system metrics on each event

prithvikannan:
# What does this PR do?

Calculate system metrics using NVML and `psutil`, and log on each event.

# Before submitting


eracah:
Make sure to add a `run_event` or a specific event function!

eracah:
@prithvikannan , can you add an import to `composer/callbacks/__init__.py` similar to how SpeedMonitor does it?

setup.py
```diff
@@ -88,6 +88,7 @@ def package_files(prefix: str, directory: str, extension: str):
     'packaging>=21.3.0,<23',
     'importlib-metadata>=5.0.0,<7',
     'mosaicml-cli>=0.4.12,<0.5',
+    'py3nvml>=0.2.7'
```
mvpatel2000, line 91: Should we move this to a separate target for this callback? I've seen some weird issues w nvml and docker images in past... but not 100% sure on this. @hanlint  do you remember what you were seeing?

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,74 @@
...
+        if event in [
+            Event.BEFORE_FORWARD,
+            Event.AFTER_FORWARD,
+            Event.BEFORE_LOSS,
+            Event.AFTER_LOSS,
+            Event.BEFORE_BACKWARD,
+            Event.AFTER_BACKWARD,
+        ]:
+            return
+        system_metrics = self.compute_system_metrics()
```
eracah, line 40: you could do something like this to gather all nodes' system metrics
```suggestion
        local_node_system_metrics = self.compute_system_metrics()
        all_system_metrics = dist.all_gather_object(local_node_system_metrics)
        system_metrics = {key: value for local_metrics  in all_system_metrics for key, value in local_metrics.items()}
```

eracah (commented):
Nice! I added some suggestions to make it work multi-rank if one is not using child runs in mlflow

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,83 @@
...
+"""System metrics monitor callback."""
+
+from __future__ import annotations
+
+import logging
+
+import psutil
+from py3nvml import py3nvml
+import os
+import dist
```
eracah, line 13:
```suggestion
from composer.utils import dist
```

eracah (commented):
whoops gave you the wrong import

setup.py
prithvikannan: Good point. I made it a separate target and used the same python nvml lib that `health_check.py` is using.

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+
+"""System metrics monitor callback."""
+
+from __future__ import annotations
+
+import logging
+import os
+
+import psutil
+import pynvml
```
eracah, line 12: We need to import gate this for situations where the user doesn't have pynvml installed because this module will get imported whether the user uses this callback or not
```suggestion
```

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+log = logging.getLogger(__name__)
+
+__all__ = ['SystemMetricsMonitor']
+
+
+class SystemMetricsMonitor(Callback):
+    """Track system metrics."""
+
+    def __init__(self) -> None:
+        super().__init__()
```
eracah, line 27: import it inside the class instead
```suggestion
        super().__init__()
         try:
            import pynvml
         except ImportError as e:
            raise MissingConditionalImportError(extra_deps_group='pynvml',
                                                conda_package='pynvml',
                                                conda_channel='conda-forge') from e
```

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+                Event.AFTER_BACKWARD,
+        ]:
+            return
+        local_node_system_metrics = self.compute_system_metrics()
+        all_system_metrics = dist.all_gather_object(local_node_system_metrics)
+        system_metrics = {key: value for local_metrics in all_system_metrics for key, value in local_metrics.items()}
+        logger.log_metrics(system_metrics)
+
+    def compute_system_metrics(self):
+        system_metrics = {}
```
eracah, line 47:
```suggestion
        import pynvml
        system_metrics = {}
```

eracah (commented):
@prithvikannan, You have some tests due to import errors. these fixes should fix the issue

eracah (approved):
LGTM

composer/callbacks/system_metrics_monitor.py
```diff
@@ -0,0 +1,83 @@
...
+        # skip the microbatch events
+        if event in [
+                Event.BEFORE_FORWARD,
+                Event.AFTER_FORWARD,
+                Event.BEFORE_LOSS,
+                Event.AFTER_LOSS,
+                Event.BEFORE_BACKWARD,
+                Event.AFTER_BACKWARD,
+        ]:
+            return
```
mvpatel2000, line 46: Instead of this, can we be explicit in which events to run on? Maybe `BATCH_START`/`EVAL_BATCH_START`/`PREDICT_BATCH_START` is sufficient? There's many events that run very fast otherwise, eg
```
self.engine.run_event(Event.AFTER_DATALOADER)

self.engine.run_event(Event.BATCH_START)
```
so this might log too much?

mvpatel2000 (commented):
Could we please also get a test file in `tests/callbacks.py` and run a small loop on both GPU and CPU? Feel free to message me if you're not sure how to do so.

In particular, I would want to verify this doesn't crash the run if there's no GPU (CPU only run) and that it actually does log everything correctly on a GPU run

mvpatel2000 (commented):
LGTM! Ty :). Was this manually tested? If not, @eracah  can help you run a job to test it before we merge

mvpatel2000 (approved):

PR #2416, opened 2023-08-05
Add log image to mlflow

eracah:
# What does this PR do?
add the `log_image` function to mlflow

# Manual Test
Ran:
```
from torchvision import datasets, transforms

from composer import Trainer
from composer.models import mnist_model
from composer.core import Callback, State
from composer.loggers import MLFlowLogger, Logger
import os

dataset = datasets.MNIST("data", train=True, download=True, transform=transform)
train_dataloader = DataLoader(dataset, batch_size=16)

class ImageLogger(Callback):
    def __init__(self, ):
        pass

    def before_forward(self, state: State, logger: Logger):
        inputs = state.batch_get_item(key=0)
        images = inputs.data.cpu().numpy()
        logger.log_images(images, step=state.timestamp.batch, use_table=True)


os.environ['DATABRICKS_HOST']=READACTED
os.environ['DATABRICKS_TOKEN']=REDACTED

trainer = Trainer(
    model=mnist_model(num_classes=10),
    max_duration="5ba",
    loggers=[MLFlowLogger(experiment_name=REDACTED,
                    tracking_uri="databricks")],
    callbacks=[ImageLogger()],
    run_name='log_image_test',
    log_to_console=True,
    progress_bar=False
)
trainer.fit()
```
And got
Screenshot 2023-09-14 at 4 58 22 PM


tests/loggers/test_mlflow_logger.py
```diff
@@ -226,3 +228,46 @@ def test_mlflow_logging_works(tmp_path, device):
 
     expected_params_list = ['num_cpus_per_node', 'node_name', 'num_nodes', 'rank_zero_seed']
     assert set(expected_params_list) == set(actual_params_list)
+
+
+@device('cpu')
```
mvpatel2000, line 233:
```suggestion
```
Not needed

tests/loggers/test_mlflow_logger.py
```diff
@@ -226,3 +228,46 @@ def test_mlflow_logging_works(tmp_path, device):
...
+def test_mlflow_log_image_works(tmp_path, device):
+
+    class ImageLogger(Callback):
+
+        def __init__(self,):
+            pass
```
mvpatel2000, line 239:
```suggestion
```
Not needed

mvpatel2000 (approved):

composer/loggers/mlflow_logger.py
```diff
@@ -145,6 +149,26 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        self,
+        images: Union[np.ndarray, torch.Tensor, Sequence[Union[np.ndarray, torch.Tensor]]],
+        name: str = 'image',
+        channels_last: bool = False,
+        step: Optional[int] = None,
+        masks: Optional[Dict[str, Union[np.ndarray, torch.Tensor, Sequence[Union[np.ndarray, torch.Tensor]]]]] = None,
+        mask_class_labels: Optional[Dict[int, str]] = None,
+        use_table: bool = True,
+    ):
+        del masks, mask_class_labels, use_table  # Unused (only for wandb)
```
dakinggg, line 162: should we raise if these are set? silently not doing segmentation logging seems not great.

tests/loggers/test_mlflow_logger.py
```diff
@@ -293,3 +295,48 @@ def test_mlflow_logging_works(tmp_path, device):
...
+
+@device('cpu')
+def test_mlflow_log_image_works(tmp_path, device):
+
+    class ImageLogger(Callback):
+
+        def before_forward(self, state: State, logger: Logger):
+            inputs = state.batch_get_item(key=0)
+            images = inputs.data.cpu().numpy()
+            logger.log_images(images, step=state.timestamp.batch.value, use_table=True)
```
dakinggg, line 308:
```suggestion
            logger.log_images(images, step=state.timestamp.batch.value, use_table=True)
```
shouldn't need `use_table=True` here because its default true? since that arg is only for wandb, would be good to have the test not specify it

dakinggg (approved):
LGTM, please include evidence of manual test in PR description

composer/loggers/mlflow_logger.py
eracah: added warning.

tests/loggers/test_mlflow_logger.py
eracah: done

PR #2445, opened 2023-08-16
Fix GCSObjectStore to match function signatures of other object stores

eracah:
# What does this PR do?
1. change `__init__` args from `gcs_root_dir` -> `bucket`, `prefix`
2. change `upload_object` args: `src` -> `filename`; `dest` -> `object_name` and reorder args
3. change `download_object` args: `src` -> `object_name`; `dest` -> `filename`

## Manual Test
Ran this to success:

```python
from composer.models import ComposerClassifier
import torch
from composer.trainer import Trainer
from torch.utils.data import Dataset
from composer.models import ComposerClassifier

class RandomClassificationDataset(Dataset):
    """Classification dataset drawn from a normal distribution.

    Args:
        shape (Sequence[int]): shape of features (default: (1, 1, 1))
        size (int): number of samples (default: 100)
        num_classes (int): number of classes (default: 2)
    """

    def __init__(self, shape = (1, 1, 1), size: int = 100, num_classes: int = 2):
        self.size = size
        self.shape = shape
        self.num_classes = num_classes
        self.x = None
        self.y = None

    def __len__(self):
        return self.size

    def __getitem__(self, index: int):
        # Note: lazily generate data so it runs after Composer seeds everything, giving the same
        # dataset across multiple calls when using the same seed.
        if self.x is None:
            self.x = torch.randn(self.size, *self.shape)
        if self.y is None:
            self.y = torch.randint(0, self.num_classes, size=(self.size,))
        return self.x[index], self.y[index]



class SimpleMLP(ComposerClassifier):
    def __init__(self, num_features: int, num_classes: int = 4):
        net = torch.nn.Sequential(
            torch.nn.Linear(num_features, num_features, bias=True),
            torch.nn.ReLU(),
            torch.nn.Linear(num_features, num_classes, bias=False),
        )
        super().__init__(module=net, num_classes=num_classes)

if __name__ == '__main__':
    num_features=32
    num_classes=8

    model = SimpleMLP(num_features=num_features, num_classes=num_classes)
    dataset = RandomClassificationDataset(shape=(num_features, ), size=128)
    dataloader = DataLoader(dataset, sampler=dist.get_sampler(dataset), batch_size=32)
    optim = torch.optim.Adam(params=model.parameters())
    trainer = Trainer(
        model=model,
        optimizers=optim,
        train_dataloader=dataloader,
        save_folder='gs://<redacted>/checkpoints/{run_name}',
        max_duration='2ba',
        save_interval='2ba',
        save_filename='ba{batch}_rank{rank}.pt',
        save_overwrite=False,
        progress_bar=False,
        log_to_console=True,
        autoresume=False,
        save_num_checkpoints_to_keep=0,
    )
    trainer.fit()
    save_run_name = trainer.state.run_name
    trainer.close()

    
    model = SimpleMLP(num_features=num_features, num_classes=num_classes)
    dataset = RandomClassificationDataset(shape=(num_features, ), size=128)
    dataloader = DataLoader(dataset, sampler=dist.get_sampler(dataset), batch_size=32)
    optim = torch.optim.Adam(params=model.parameters())
    trainer = Trainer(
        model=model,
        optimizers=optim,
        train_dataloader=dataloader,
        max_duration='4ba',
        save_interval='2ba',
        save_filename='ba{batch}_rank{rank}.pt',
        save_overwrite=False,
        load_path=f'gs://<redacted>/checkpoints/{save_run_name}/ba2_rank0.pt',
        progress_bar=False,
        log_to_console=True,
        autoresume=False,
        save_num_checkpoints_to_keep=0,
    )
    trainer.fit()
    trainer.close()
```

### DIFF
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.15.1'
+__version__ = '0.16.0'
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -9,6 +9,7 @@ import logging
 import math
 import os
 import pathlib
+import shutil
 import tempfile
 import textwrap
 from pathlib import Path
@@ -485,9 +486,8 @@ class CheckpointSaver(Callback):  # noqa: D101
             prefix_dir = None
             checkpoint = self.saved_checkpoints.pop(0)
             prefix_dir = str(Path(checkpoint).parent)
-            os.remove(checkpoint)
-            if sharding_enabled and dist.get_global_rank() == 0:
-                if using_torch_2():
-                    metadata_file = os.path.join(prefix_dir, _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME)
-                    os.remove(metadata_file)
-                os.removedirs(prefix_dir)
+            if not sharding_enabled:
+                os.remove(checkpoint)
+            else:
+                if dist.get_global_rank() == 0:
+                    shutil.rmtree(prefix_dir)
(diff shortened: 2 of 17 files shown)


mvpatel2000 (approved):
LGTM! Thank you for hunting this down 🙏

PR #2449, opened 2023-08-18
Upstream Generate Callback 

irenedea:
# What does this PR do?

Upstreams the callback that logs generations to wandb from foundry to composer. https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/callbacks/generate_callback.py 

The upstreaming should include a few things beyond just copy pasting the code into composer:
- support all logger destinations that implement a log_text method (or log_table?).
- use the HuggingFaceModel.generate function in composer rather than calling the underyling hf model directly
- support composer’s time abstraction instead of just batch_log_interval
- support a batch size argument so that the maximum number of prompts that can be used is greater
- tests

# Manual Test

Using llm foundry quickstart example with wandb logging
cli:
```
# From llm-foundry quickstart
cd llm-foundry/scripts
# Convert C4 dataset to StreamingDataset format
python data_prep/convert_dataset_hf.py \
  --dataset c4 --data_subset en \
  --out_root my-copy-c4 --splits train_small val_small \
  --concat_tokens 2048 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>'

# Train an MPT-125m model for 10 batches
composer train/train.py \
  train/yamls/pretrain/mpt-125m.yaml \
  data_local=my-copy-c4 \
  train_loader.dataset.split=train_small \
  eval_loader.dataset.split=val_small \
  max_duration=10ba \
  eval_interval=0 \
  save_folder=mpt-125m
```
yaml
```
# same as mpt-125m.yaml
callbacks:
  # ... other callbacks
  generate:
    prompts:
      - 'My name is'
      - 'The best day'
      - 'How do you do?'
    interval: 1ba
    batch_size: 2
    max_new_tokens: 20

loggers:
  wandb: {}

```
https://wandb.ai/mosaic-ml/llm-foundry-scripts_train/runs/uddzyuqg

The most recent set of prompt/generations is shown in the table panel. Earlier sets can be found as json artifacts:

https://wandb.ai/mosaic-ml/llm-foundry-scripts_train/artifacts/run_table/run-uddzyuqg-generations/96215e792abd8837ba0a/files/generations.table.json

https://mosaicml.atlassian.net/browse/GRT-2010
https://mosaicml.atlassian.net/browse/GRT-2332

# Before submitting

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -6,7 +6,6 @@
 from __future__ import annotations
 
 import logging
-import math
 import os
 import pathlib
 import shutil
@@ -15,7 +14,8 @@ import textwrap
 from pathlib import Path
 from typing import Callable, List, Optional, Union
 
-from composer.core import Callback, Event, State, Time, TimeUnit
+from composer.callbacks.utils import create_interval_scheduler
+from composer.core import Callback, Event, State, Time
 from composer.loggers import Logger
 from composer.utils import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, FORMAT_NAME_WITH_DIST_TABLE, PartialFilePath,
                             checkpoint, create_symlink_file, dist, ensure_folder_has_no_conflicting_files,
@@ -25,67 +25,9 @@ from composer.utils.misc import using_torch_2
 
 log = logging.getLogger(__name__)
 
-__all__ = ['CheckpointSaver', 'checkpoint_periodically']
+__all__ = ['CheckpointSaver']
 
 _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME = '.metadata'
-
-
-def checkpoint_periodically(interval: Union[str, int, Time],
-                            save_end_of_training=True) -> Callable[[State, Event], bool]:
-    r"""Helper function to create a checkpoint scheduler according to a specified interval.
-
-    Args:
-        interval (Union[str, int, :class:`.Time`]): The interval describing how often checkpoints should be
-            saved. If an integer, it will be assumed to be in :attr:`.TimeUnit.EPOCH`\s.
-            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH`, :attr:`.TimeUnit.BATCH`,
-            :attr:`.TimeUnit.TOKEN`, or :attr:`.TimeUnit.SAMPLE`.
-
-            Checkpoints will be saved every ``n`` batches or epochs (depending on the unit),
-            and at the end of training.
-
-    Returns:
-        Callable[[State, Event], bool]: A function that can be passed as the ``save_interval``
-            argument into the :class:`.CheckpointSaver`.
-    """
-    if isinstance(interval, str):
-        interval = Time.from_timestring(interval)
-    if isinstance(interval, int):
-        interval = Time(interval, TimeUnit.EPOCH)
-
-    if interval.unit == TimeUnit.EPOCH:
-        save_event = Event.EPOCH_CHECKPOINT
-    elif interval.unit in {TimeUnit.BATCH, TimeUnit.TOKEN, TimeUnit.SAMPLE}:
-        save_event = Event.BATCH_CHECKPOINT
-    else:
-        raise NotImplementedError(
-            f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
-        )
-
-    def save_interval(state: State, event: Event):
-        elapsed_duration = state.get_elapsed_duration()
-        assert elapsed_duration is not None, 'elapsed_duration is set on the BATCH_CHECKPOINT and EPOCH_CHECKPOINT'
-
-        # Always checkpoint at end of training
-        if save_end_of_training and elapsed_duration >= 1.0:
-            return True
-
-        # previous timestamp will only be None if training has not started, but we are returning False
-        # in this case, just to be safe
-        if state.previous_timestamp is None:
-            return False
-
-        if interval.unit in {TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, TimeUnit.SAMPLE}:
-            previous_count = state.previous_timestamp.get(interval.unit)
-            count = state.timestamp.get(interval.unit)
-        else:
-            raise NotImplementedError(
-                f'Unknown checkpointing interval: {interval.unit}. Must be TimeUnit.EPOCH, TimeUnit.BATCH, TimeUnit.TOKEN, or TimeUnit.SAMPLE.'
-            )
-
-        threshold_passed = math.floor(previous_count / interval.value) != math.floor(count / interval.value)
-        return event == save_event and threshold_passed
-
-    return save_interval
 
 
 class CheckpointSaver(Callback):  # noqa: D101
@@ -310,7 +252,7 @@ class CheckpointSaver(Callback):  # noqa: D101
         latest_remote_file_name = str(latest_remote_file_name) if latest_remote_file_name is not None else None
 
         if not callable(save_interval):
-            save_interval = checkpoint_periodically(save_interval)
+            save_interval = create_interval_scheduler(save_interval)
         self.save_interval = save_interval
         self.last_checkpoint_batch: Optional[Time] = None
 
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -30,7 +30,8 @@ __all__ = ['CheckpointSaver', 'checkpoint_periodically']
 _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME = '.metadata'
 
 
-def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State, Event], bool]:
+def checkpoint_periodically(interval: Union[str, int, Time],
+                            save_end_of_training=True) -> Callable[[State, Event], bool]:
     r"""Helper function to create a checkpoint scheduler according to a specified interval.
 
     Args:
@@ -65,7 +66,7 @@ def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State,
         assert elapsed_duration is not None, 'elapsed_duration is set on the BATCH_CHECKPOINT and EPOCH_CHECKPOINT'
 
         # Always checkpoint at end of training
-        if elapsed_duration >= 1.0:
+        if save_end_of_training and elapsed_duration >= 1.0:
             return True
 
         # previous timestamp will only be None if training has not started, but we are returning False
--- a/composer/callbacks/generate.py
+++ b/composer/callbacks/generate.py
@@ -19,7 +19,7 @@ from composer.utils.import_helpers import MissingConditionalImportError
 
 
 class Generate(Callback):
-    """Periodically log generations.
+    """Periodically log generations from a set of prompts.
 
     Args:
         prompts (List[str]): The list of prompts you would like to produce generations for
@@ -32,7 +32,6 @@ class Generate(Callback):
     """
 
     def __init__(self, prompts: List[str], interval: Union[str, int, Time], batch_size: int, **kwargs: Any):
-
         self.prompts = prompts
         self.generate_kwargs = kwargs
         self.batch_size = batch_size
@@ -111,7 +110,7 @@ class Generate(Callback):
                 output_text = tokenizer.decode(output_tokens, skip_special_tokens=True)
                 rows.append([prompt, output_text])
 
-            # TODO: LOG TABLE
+            logger.log_table(columns=['prompt', 'generation'], rows=rows, name='generations')
 
         tokenizer.padding_side = original_padding_side
         model.train(mode=original_mode)
--- a/composer/callbacks/generate.py
+++ b/composer/callbacks/generate.py
@@ -1,0 +1,117 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Periodically log generations to wandb from a set of prompts."""
+from typing import Any, List, Union, cast
+
+from torch.nn.parallel import DistributedDataParallel
+from torch.utils.data import DataLoader, TensorDataset
+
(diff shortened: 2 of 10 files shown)


composer/callbacks/generate.py
```diff
@@ -0,0 +1,109 @@
...
+
+from torch.utils.data import TensorDataset, DataLoader
+
+Tokenizer = Union[PreTrainedTokenizer, PreTrainedTokenizerFast]
+
+class Generate(Callback):
+
+    def __init__(self, prompts: List[str], interval: Union[str, int, Time], batch_size: int,
+                 **kwargs: Any):
+        """Periodically log generations to wandb from a set of prompts.
```
irenedea, line 22: TODO: Update this

tests/callbacks/test_generation.py
```diff
@@ -0,0 +1,18 @@
+from composer.callbacks import Generate
```
irenedea, line 1: todo: rename file to test_generate to match the callback file name

composer/callbacks/checkpoint_saver.py
```diff
@@ -29,7 +29,7 @@
 _TORCH_DISTRIBUTED_CHECKPOINTS_METADATA_FILENAME = '.metadata'
 
 
-def checkpoint_periodically(interval: Union[str, int, Time]) -> Callable[[State, Event], bool]:
+def checkpoint_periodically(interval: Union[str, int, Time], save_end_of_training=True) -> Callable[[State, Event], bool]:
```
irenedea, line 32: We could rename and move this to a callbacks util file because it is used both by the CheckpointSaver and Generate callbacks

composer/callbacks/generate.py
```diff
@@ -0,0 +1,99 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Periodically log generations to wandb from a set of prompts."""
+from typing import Any, List, Union
+
+from composer.core import Callback, Event, State, Time
+from composer.loggers import Logger
+from composer.models import HuggingFaceModel
+from composer.utils import dist
+from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
```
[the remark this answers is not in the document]
dakinggg: any optional dependencies should be conditionally imported when they are needed. See https://github.com/mosaicml/composer/blob/8ceedfc09cba64584f9dcd09e4d015c55fcea759/composer/models/huggingface.py#L74-L81 for example.

composer/callbacks/generate.py
```diff
@@ -0,0 +1,116 @@
...
+            tokenized_input[k] = device.tensor_to_device(v)
+
+        all_input_ids = tokenized_input['input_ids']
+        all_attn_masks = tokenized_input['attention_mask']
+
+        batches = DataLoader(TensorDataset(all_input_ids, all_attn_masks), self.batch_size)
+
+        output_token_ids = []
+        # dummy forward call needed for FSDP to work consistently
+        model.dummy_forward_called = False  #
```
irenedea, line 89: This is necessary because the code in model.generate sets `dummy_forward_called` to True after the first call to `generate`, which causes a cuda memory error due to the dummy forward pass not being called in subsequent train batches. Perhaps we should be modifying the `HuggingFaceModel.generate` function to address this? @dakinggg

tests/callbacks/callback_settings.py
```diff
@@ -191,3 +201,11 @@ def test_something(constructor: Callable, yaml_dict: Dict[str, Any]):
     implementations = []
     ans = [_to_pytest_param(impl) for impl in implementations]
     return ans
+
+
+def get_cb_model_and_datasets(cb: Callback) -> Tuple[ComposerModel, DataLoader, DataLoader]:
```
irenedea, line 206: Had to do this refactor because Generate callback requires a HuggingFaceModel

dakinggg (commented):
Looking good overall! A couple additional requests for testing:
- Please do a manual test (an actual short training run with this callback enabled and verify generations get written to wandb)
- Please test that this callback doesn't affect model training. That is a unit test that trains for a few batches and trains (with all the same seeded randomness) for a few batches with the callback and then assert you get the same model out of each training run. e.g. something like https://github.com/mosaicml/composer/blob/e7a092216bcc88caca284265c8a83cd052b9b0e4/composer/callbacks/export_for_inference.py#L79

tests/callbacks/test_generate.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+    fsdp_config = None
+    if use_fsdp:
+        fsdp_config = {
+            'sharding_strategy': 'FULL_SHARD',
+        }
+
+    model = configure_tiny_gpt2_hf_model()
+    model.generate = Mock(wraps=model.generate)
```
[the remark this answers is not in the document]
irenedea: it's wrapped, so it still calls model.generate, but the mock allows me to count the number of times the method was called.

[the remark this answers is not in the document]
dakinggg: ah got it, thanks!

composer/callbacks/generate.py
```diff
@@ -0,0 +1,116 @@
...
+        tokenizer = model.tokenizer
+
+        try:
+            from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
+        except ImportError as e:
+            raise MissingConditionalImportError(extra_deps_group='nlp',
+                                                conda_package='transformers',
+                                                conda_channel='conda-forge') from e
+        Tokenizer = Union[PreTrainedTokenizer, PreTrainedTokenizerFast]
+        tokenizer = cast(Tokenizer, tokenizer)
```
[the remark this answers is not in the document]
irenedea: I believe it is because it fails the pyright test otherwise. Originally I had `assert isinstance(tokenizer, Tokenizer)`, but pyright is unable to infer its type from that for some reason. 

(The pyright error was something along the lines of tokenizer being a NoneType, so the accesses to its attributes below will fail. )

composer/callbacks/generate.py
```diff
@@ -0,0 +1,116 @@
...
+        device = state.device
+
+        # Stash the original value of padding_side because generation requires left padding
+        original_padding_side = tokenizer.padding_side
+        tokenizer.padding_side = 'left'
+        if tokenizer.pad_token_id is None:
+            tokenizer.pad_token_id = tokenizer.eos_token_id
+        tokenized_input = tokenizer(self.prompts, return_tensors='pt', padding=True)
+
+        for k, v in tokenized_input.items():
```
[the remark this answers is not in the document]
irenedea: oops yes forgot to remove this while testing

composer/callbacks/generate.py
```diff
@@ -0,0 +1,116 @@
...
+        self.prompts = prompts
+        self.generate_kwargs = kwargs
+        self.batch_size = batch_size
+        self.check_interval = create_interval_scheduler(interval, include_end_of_training=False)
+
+    def run_event(self, event: Event, state: State, logger: Logger) -> None:
+        if state.get_elapsed_duration() is not None and self.check_interval(state, event):
+            self.generate(state, logger)
+        else:
+            super().run_event(event, state, logger)
```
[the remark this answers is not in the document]
irenedea: Hm probably not for our particular case, but I figured it would be good to default to the original behavior, such that the other Callback behaviors work. For example, if someone wanted to extend Generate by overriding `after_load`, it would work as expected. wdyt?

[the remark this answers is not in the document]
irenedea: to really preserve behavior, i think `super().run_event(...)` should actually run always, not just in an `else` block

composer/callbacks/generate.py
irenedea: Do you think it make sense to modify `HuggingFaceModel` to set `dummy_forward_called` to `False` after a forward call with gradients enabled? (like this: https://github.com/mosaicml/composer/commit/a3b12c2e02a5537cfcd4327897fbf2d23cfd976a#diff-e61f9af34de66238d99995ebdc5d2645cea3c7ca7c303ac006239a720c5b620cR361) That seems to work...

I _think_ this makes sense because the dummy would be needed to help free the parameters after a (training) forward call

irenedea: If a proper fix is more involved than this, it makes sense to not bother fixing bc of torch2 fix 😛

composer/callbacks/generate.py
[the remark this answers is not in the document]
dakinggg: I think the idea with callbacks is that you _either_ override `run_event` to do whatever you want, _or_ define the functions (e.g. `after_load`) and let `run_event` handle dispatching to them. So imo better to not mix the two methods and just override `run_event` here _or_ define the `batch_checkpoint` and `epoch_checkpoint` function calls like the checkpoint saver does. I don't have a strong opinion one way or the other, but I do think it is better not to mix the two methods.

composer/callbacks/generate.py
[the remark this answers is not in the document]
dakinggg: cool cool, if pyright needs its thats fine

composer/callbacks/generate.py
dakinggg: I vote to just not fix it in `HuggingFaceModel.generate`. The solution wasn't that robust to start with, and I'm not sure how robust that line you added is (although it seems totally reasonable).

composer/callbacks/generate.py
```diff
@@ -0,0 +1,116 @@
...
+            tokenizer.pad_token_id = tokenizer.eos_token_id
+        tokenized_input = tokenizer(self.prompts, return_tensors='pt', padding=True)
+
+        for k, v in tokenized_input.items():
+            tokenized_input[k] = device.tensor_to_device(v)
+
+        all_input_ids = tokenized_input['input_ids']
+        all_attn_masks = tokenized_input['attention_mask']
+
+        batches = DataLoader(TensorDataset(all_input_ids, all_attn_masks), self.batch_size)
```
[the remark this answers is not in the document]
irenedea: Yeah, on the fence also; just thought it could be a little cleaner this way + maybe there's under-the-hood benefits of using pytorch api (though there doesn't really seem to be after looking at the impl)

I can switch over to doing a for loop. I think creating dataset/dataloader objects can be a bit overkill 😵‍💫

composer/callbacks/generate.py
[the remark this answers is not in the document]
irenedea: gotcha, sgtm!

composer/callbacks/generate.py
irenedea: sgtm!

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright 2022 MosaicML LLM Foundry authors
```
dakinggg, line 4: Drop the foundry license

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
...
+    """Periodically log generations from a set of prompts.
+
+    Args:
+        prompts (List[str]): The list of prompts you would like to produce generations for
+        interval (Union[str, int, :class:`.Time`]): The interval describing how often checkpoints should be
+            saved. If an integer, it will be assumed to be in :attr:`.TimeUnit.EPOCH`.
+            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH`, :attr:`.TimeUnit.BATCH`,
+            :attr:`.TimeUnit.TOKEN`, or :attr:`.TimeUnit.SAMPLE`.
+        batch_size (int): Size of a prompt batch for generation.
+        kwargs: All kwargs well be passed along to the call to generate. This is for things like `do_sample`, `top_p`, etc
```
dakinggg, line 28: well -> will

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    def __init__(self, prompts: List[str], interval: Union[str, int, Time], batch_size: int, **kwargs: Any):
```
dakinggg, line 31: Should we make batch size optional and default to the number of prompts (i.e. one batch)?

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
...
+        try:
+            from transformers import PreTrainedTokenizerBase
```
dakinggg, line 33: I'd just
```
import transformers
del transformers
```

here. Looks a little cleaner

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
...
+        self.check_interval = create_interval_scheduler(interval, include_end_of_training=False)
+
+    def run_event(self, event: Event, state: State, logger: Logger) -> None:
+        if state.get_elapsed_duration() is not None and self.check_interval(state, event):
+            self.generate(state, logger)
+
+    def generate(self, state: State, logger: Logger):
+        model = state.model.module if state.is_model_ddp else state.model
+        assert isinstance(
+            model, HuggingFaceModel
```
dakinggg, line 51: This should raise a value error rather than assert

composer/callbacks/generate.py
```diff
@@ -0,0 +1,110 @@
...
+        ), f'Expected HuggingFaceModel, but got {model.__class__.__name__}'  # TODO: Extend to support any models that have a generate method.
+
+        if not hasattr(model, 'tokenizer') or model.tokenizer is None:
+            raise ValueError(f'Model {model.__class__.__name__} does not have a tokenizer.')
```
dakinggg, line 55: "...which is required for generation"

tests/callbacks/test_generate.py
```diff
@@ -0,0 +1,120 @@
...
+            }
+        return None
+
+    def _create_trainer(self, device, max_duration, use_fsdp, generate_cb: Optional[Generate] = None) -> Trainer:
+        return Trainer(model=configure_tiny_gpt2_hf_model(),
+                       train_dataloader=dummy_gpt_lm_dataloader(),
+                       device=device,
+                       max_duration=max_duration,
+                       callbacks=generate_cb,
+                       fsdp_config={'sharding_strategy': 'FULL_SHARD'} if use_fsdp else None)
```
dakinggg, line 56: Guessing you meant to use the fsdp config helper function here?

dakinggg (approved):
Great work!

tests/callbacks/test_generate.py
irenedea: ahh good catch, actually was going to just delete the helper! originally made the helper bc I was calling it multiple times  across tests before the _create_trainer helper

composer/callbacks/generate.py
irenedea: good idea!

PR #2450, opened 2023-08-18
Materialize only model state_dict in memory for  `save_weights_only`

eracah:
# What does this PR do?

# Manual Tests
ran a save_weights only for MPT 30B with:
* Lion (run_name: `test-ckpt-oom-prevent-lion-8-GzSIU3`)
* Adam (run_name: `test-ckpt-oom-prevent-adam-8-lqMUBL`)

using torch 2.01 (not nightly) on 500GB cpu memory 8xA100_80GB

Both successfully saved

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -844,6 +844,24 @@ class State(Serializable):
 
         return obj
 
+    def get_model_state_dict(self) -> Dict[str, Any]:
+        """Collect the state dict for the model.
+
+        Returns:
+            Dict[str, Any]: The state dict for the model.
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
lgtm, please add a pr description and evidence of a manual test with and without save weights only (although im pretty sure this pr is safe)

eracah:
ok @dakinggg, manual tests

> lgtm, please add a pr description and evidence of a manual test with and without save weights only (although im pretty sure this pr is safe)

ok added manual tests

PR #2465, opened 2023-08-22
Fix local eval

rishab-partha:
# What does this PR do?

Fixes local eval by saving to different paths by rank.

Issues from Dan running local human eval.


j316chuck (commented):
LGTM thanks, is there a test for this?

mvpatel2000 (approved):
LGTM. Same comment as Chuck, would a test have caught this? maybe we can tag a local code eval test with world_size(2)?

rishab-partha:
added the world size 2 test

PR #2466, opened 2023-08-23
Add cot eval

bmosaicml:
# What does this PR do?
This extends the existing QATask to enable chain of thought prompting. The only change necessary is to search for the model's final answer after a delimiter at the end of the model generation as opposed to searching for the model's answer at the beginning of the output. (In CoT, the beginning of the model output is an ungraded explanation intended only to encourage the model to use more FLOPs to solve the problem).

# Before submitting
- [ x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -248,10 +263,15 @@ def collate_fn(self, data):
             inputs.append(inp)
             answers.append(aliases)
 
+            # we will search for the answer within the portion of the model response
+            # beginning with `cot_delimiter`
+            cot_delimiter = sample['cot_delimiter']
```
samhavens, line 268: Is there always a delimiter? I feel like the options I have seen for CoT are:
* Delimiter
* Answer at end
* Need to evaluate code to get answer (when the model is encouraged to write code to solve the problem, as in Program of Thought in MAmmoTH)

I might be wrong here, maybe the data is presented as if there is no delimiter in the paper, but then they add on `### So the answer is` or something. But if not, then,  maybe we want to have some syntax to support answer-at-end?

bmosaicml: In Wei's introduction of CoT they use delimiters. I haven't seen any papers in particular that instead check for answer at the end. Our models tend to ramble on and on, so I suspect using delimiters and then checking that the answer is at the beginning of the substring following the delimiter will yield much higher accuracy.

I don't see a very strong need to support both at this time but I could be convinced...

samhavens: Yeah that sounds reasonable to me. Just wanted to talk it out, thank you.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -193,11 +201,15 @@ def prep_examples(self, num_fewshot: int, prompt_string: str, example_delimiter:
...
+                    ctxt, cot, cont = (self.samples[fewshot_idx]['context'],
+                                       self.samples[fewshot_idx].get('chain_of_thought',
+                                                                     ''), self.samples[fewshot_idx]['answer'])
+                    if len(cot) == 0:
+                        cot_delimiter = ''
                     ctxt = f'{question_prelimiter}{ctxt}'
                     if len(preamble) > 0:
                         ctxt = f'{example_delimiter}{ctxt}'
-                    preamble += f'{ctxt}{continuation_delimiter}{cont}'
+                    preamble += f'{ctxt}{continuation_delimiter}{cot}{cot_delimiter}{cont}'
```
b-chu, line 212: Separate this out in a helper function, including `preamble = prompt_string`

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -193,11 +201,15 @@ def prep_examples(self, num_fewshot: int, prompt_string: str, example_delimiter:
             if num_fewshot > 0:
                 fewshot_idxs = _get_fewshot_sample_idxs(len(self.samples), num_fewshot, sample_idx, fewshot_rng)
                 for fewshot_idx in fewshot_idxs:
-                    ctxt, cont = self.samples[fewshot_idx]['context'], self.samples[fewshot_idx]['answer']
+                    ctxt, cot, cont = (self.samples[fewshot_idx]['context'],
```
b-chu, line 204: Separate this into 3 lines and use more descriptive variable names, e.g. `context` `cot` and `answer`

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -163,10 +169,10 @@ def __init__(
         self.max_answer_length = 0
         fewshot_rng = random.Random(fewshot_random_seed)
         self.encoded_dataset = self.prep_examples(num_fewshot, prompt_string, example_delimiter, continuation_delimiter,
-                                                  question_prelimiter, fewshot_rng)
+                                                  question_prelimiter, cot_delimiter, fewshot_rng)
 
     def prep_examples(self, num_fewshot: int, prompt_string: str, example_delimiter: str, continuation_delimiter: str,
-                      question_prelimiter: str, fewshot_rng: random.Random):
+                      question_prelimiter: str, cot_delimiter: str, fewshot_rng: random.Random):
```
b-chu, line 175: Does it make sense to default `cot_delimiter: str = ''`?

b-chu: Or `cot_delimiter: Optional[str] = None`? Will there be a case where some people do CoT without a specific delimiter (they use '') and some people who don't do CoT, in which case the delimiter should be `None` to represent the difference?

composer/datasets/in_context_learning_evaluation.py
b-chu: And similarly for variables below.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -21,6 +21,9 @@
 if TYPE_CHECKING:
     import transformers
 
+# allow models slightly more tokens than were used in the most verbose CoT in the dataset
+MAX_ANSWER_BUFFER_LENGTH = 10
```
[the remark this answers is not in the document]
b-chu: Why is the buffer 10?

composer/metrics/nlp.py
```diff
@@ -265,11 +265,22 @@ def replace_underscore(text: str) -> str:
 
         return white_space_fix(remove_articles(handle_punc(lower(replace_underscore(answer))))).strip()
 
-    def update(self, outputs: List[str], labels: List[List[str]]):
+    def update(self, batch, outputs: List[str], labels: List[List[str]]):
```
b-chu, line 268: `List` and `Dict` are being deprecated in favor of `list` and `dict`. Python 3.9, PEP 585 https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

b-chu: Also fix elsewhere

composer/metrics/nlp.py
```diff
@@ -265,11 +265,22 @@ def replace_underscore(text: str) -> str:
...
+        cot_delimiter = batch['cot_delimiter']
         for sample_output, sample_labels in zip(outputs, labels):
-            cleaned_sample_output = self.normalize_answer(sample_output)
+            if cot_delimiter is not None and len(cot_delimiter) > 0:
+                # in chain of thought, the final answer comes after the
```
b-chu, line 272: Capitalize first letter of comments.
```suggestion
                # In chain of thought, the final answer comes after the
```

composer/metrics/nlp.py
```diff
@@ -265,11 +265,22 @@ def replace_underscore(text: str) -> str:
...
+                # explanation and is delimited by `cot_delimiter`
+                final_answer = ''
+                if cot_delimiter in sample_output:
+                    final_answer = sample_output.split(cot_delimiter)[-1]
```
b-chu, line 276:
```suggestion
            final_answer = sample_output
            if cot_delimiter is not None and len(cot_delimiter) > 0:
                final_answer = final_answer.split(cot_delimiter)[-1]
```

b-chu (changes requested):
Thanks for the PR! I haven't looked at the test files, but my main question is what do you think about cot_delimiter being None as default? In cases where there's no CoT it's set as None and when there is CoT, it can be a delimiter or empty string.

composer/metrics/nlp.py
```diff
@@ -265,11 +265,22 @@ def replace_underscore(text: str) -> str:
 
         return white_space_fix(remove_articles(handle_punc(lower(replace_underscore(answer))))).strip()
 
-    def update(self, outputs: List[str], labels: List[List[str]]):
+    def update(self, batch, outputs: List[str], labels: List[List[str]]):
+        cot_delimiter = batch['cot_delimiter']
```
dakinggg, line 269: I think this should be optional in the batch, so `batch.get('cot_delimiter')`

composer/metrics/nlp.py
dakinggg, line 268: this is an API breaking change, so we need to make it backward compatible. I think you can do this by making batch the third argument and making it optional with a default of None. wdyt?

composer/datasets/in_context_learning_evaluation.py
bmosaicml: yes I think this will be a common case!

composer/metrics/nlp.py
bmosaicml: I can do that but it's going to result in the HuggingFace model's update_metric method making little sense

dakinggg: why does it make that method not make sense?

composer/datasets/in_context_learning_evaluation.py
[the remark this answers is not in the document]
b-chu: Prepend private global constants with `_`.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -125,6 +128,17 @@ class InContextLearningQATaskDataset(Dataset):
         fewshot_random_seed (int): Random seed to use for fewshot sampling
     """
 
+    @classmethod
+    def read_dataset(cls, dataset):
```
b-chu, line 132: Is there a reason this needs to be a class method instead of a generic function? Also use a for loop instead of a lambda inside a map

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -138,6 +152,7 @@ def __init__(
         destination_path: str,
         question_prelimiter: str,
         fewshot_random_seed: int,
+        cot_delimiter: str = '',
```
b-chu, line 155: How do you differentiate between if a particular CoT setup is using their delimiter as `''` vs no CoT is being done? Is the first case even possible?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -163,10 +173,31 @@ def __init__(
         self.max_answer_length = 0
         fewshot_rng = random.Random(fewshot_random_seed)
         self.encoded_dataset = self.prep_examples(num_fewshot, prompt_string, example_delimiter, continuation_delimiter,
-                                                  question_prelimiter, fewshot_rng)
+                                                  question_prelimiter, cot_delimiter, fewshot_rng)
+
+    def _format_prompt_and_fewshot(self, num_fewshot: int, prompt_string: str, example_delimiter: str,
+                                   continuation_delimiter: str, question_prelimiter: str, cot_delimiter: str,
+                                   fewshot_rng: random.Random, sample_idx: int):
```
b-chu, line 180: Add a return type hint

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -163,10 +173,31 @@ def __init__(
         self.max_answer_length = 0
         fewshot_rng = random.Random(fewshot_random_seed)
         self.encoded_dataset = self.prep_examples(num_fewshot, prompt_string, example_delimiter, continuation_delimiter,
-                                                  question_prelimiter, fewshot_rng)
+                                                  question_prelimiter, cot_delimiter, fewshot_rng)
+
+    def _format_prompt_and_fewshot(self, num_fewshot: int, prompt_string: str, example_delimiter: str,
```
b-chu, line 178: Add a docstring for this function

composer/datasets/in_context_learning_evaluation.py
b-chu: This argument should have a default value though.

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -248,10 +277,15 @@ def collate_fn(self, data):
             inputs.append(inp)
             answers.append(aliases)
 
+            # we will search for the answer within the portion of the model response
```
b-chu, line 280:
```suggestion
            # We will search for the answer within the portion of the model response
```

b-chu (changes requested):

composer/datasets/in_context_learning_evaluation.py
bmosaicml: Yes it is possible but there's no need to differentiate since the non-emptiness of the chain_of_thought key defines whether chain of thought is in effect

composer/datasets/in_context_learning_evaluation.py
bmosaicml: I can make it a regular function

b-chu (changes requested):

b-chu (approved):
Thanks for adding this in @bmosaicml !!

PR #2468, opened 2023-08-24
Fix torchmetrics backwards compatibility issue

eracah:
# What does this PR do?
1. add a get(..., None) call to `load_state_dict`
2. update checkpoints in backwards compat test to use eval_metrics


tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -349,7 +380,8 @@ def test_fsdp_load_old_checkpoint(world_size, tmp_path: pathlib.Path, precision:
         _compare_optims_between_state_dicts(state_dict1, state_dict2)
 
     # Continue to fit to make sure we can continue training.
-    trainer.fit()
+    if train_metrics is None and val_metrics is None:
```
dakinggg, line 383: please fix

composer/core/state.py
```diff
@@ -907,13 +907,19 @@ def state_dict(self) -> Dict[str, Any]:
...
                     _computed = v._computed
-                    _computed_device = str(_computed.device) if _computed is not None else None
-                    _np_computed = _computed.cpu().numpy() if _computed is not None else None
-                    serialized_value[k] = {
-                        'state_dict': v.state_dict(),
-                        '_computed': _np_computed,
-                        '_computed_device': _computed_device
-                    }
+                    if isinstance(_computed, torch.Tensor):
+                        _computed_device = str(_computed.device) if _computed is not None else None
```
dakinggg, line 911: `_compute` is not `None` at this point, because its a tensor.

composer/core/state.py
```diff
@@ -922,13 +928,19 @@ def state_dict(self) -> Dict[str, Any]:
...
                         _computed = v._computed
-                        _computed_device = str(_computed.device) if _computed is not None else None
-                        _np_computed = _computed.cpu().numpy() if _computed is not None else None
-                        serialized_value[eval_key][k] = {
-                            'state_dict': v.state_dict(),
-                            '_computed': _np_computed,
-                            '_computed_device': _computed_device
-                        }
+                        if isinstance(_computed, torch.Tensor):
+                            _computed_device = str(_computed.device) if _computed is not None else None
```
dakinggg, line 932: same

composer/core/state.py
```diff
@@ -907,13 +907,19 @@ def state_dict(self) -> Dict[str, Any]:
...
+                    if isinstance(_computed, torch.Tensor):
+                        _computed_device = str(_computed.device) if _computed is not None else None
+                        _np_computed = _computed.cpu().numpy() if _computed is not None else None
+                        serialized_value[k] = {
+                            'state_dict': v.state_dict(),
+                            '_computed': _np_computed,
+                            '_computed_device': _computed_device
+                        }
+                    else:
+                        serialized_value[k] = {
```
dakinggg, line 919: what is it is not a tensor for the else case/are we confident the else path works?

mvpatel2000: @eracah has said they only do weird invasive things if it is a single tensor. I have added a MAP test which verifies this (it has a dict)

tests/common/compare.py
```diff
@@ -53,8 +53,16 @@ def _check_item(item1: Any, item2: Any, path: str, rtol: float = 0.0, atol: floa
...
         item2._update_count += 1
-        assert item1.compute().allclose(item2.compute(), atol=atol,
-                                        rtol=rtol), f'{path} differs: {item1.compute()} != {item2.compute()}'
+        item1_compute = item1.compute()
+        item2_compute = item2.compute()
+        if isinstance(item1_compute, torch.Tensor) and isinstance(item2_compute, torch.Tensor):
+            assert item1_compute.allclose(item2_compute, atol=atol, rtol=rtol,
+                                          equal_nan=True), f'{path} differs: {item1_compute} != {item2_compute}'
+        elif isinstance(item1_compute, dict):
+            assert isinstance(item2_compute, dict)
```
dakinggg, line 62: this seems superfluous?

composer/utils/checkpoint.py
```diff
@@ -400,11 +400,24 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
         # We need no_grad because we overwrite tensor values with set_() when we do elastic loading and we don't want the set_ op recorded in the computation graph.
         with torch.no_grad():
-            # 1. Load just model first.
-            model_state_dict = {'state': {'model': state.state_dict()['model']}}
+            # 1. Load model and metadata first
+            model_state_dict = None
+            if load_weights_only:
+                model_state_dict = {'state': {'model': state.get_model_state_dict()}}
+            else:
+                cur_state_dict = state.state_dict()
```
eracah, line 408: I wish there was a way to do it without materializing the optimizer state

composer/utils/checkpoint.py
```diff
@@ -419,7 +432,7 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
                                                                 storage_reader=storage_reader)
                 state.load_optim_state(optim_state)
 
-        # 3. Optionally, load RNG.
+        # 3. Optionally load RNG
```
eracah, line 435: this is okay to do because there is a tensor in it I guess?

eracah (commented):
Ok looks pretty good. just had a couple comments

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -311,6 +318,9 @@ def test_fsdp_mixed_with_sync(world_size, tmp_path: pathlib.Path, sync_module_st
     pytest.param('0.14.1',
                  marks=pytest.mark.filterwarnings(
                      r'ignore:MosaicMLLogger is not in the state_dict. Its state will not be restored.:UserWarning')),
+    pytest.param('0.15.1',
```
dakinggg, line 321: we should add 0.16 here, yeah?

tests/common/compare.py
mvpatel2000: the dict case? it's required for MAP

composer/utils/checkpoint.py
mvpatel2000: Is it slow? do I have to be concerned about it?

composer/utils/checkpoint.py
mvpatel2000: yes I think so. Should we move it into initial load?

tests/trainer/test_fsdp_checkpoint.py
mvpatel2000: I will just add a 0.16.1 after this. It's kinda broken right now anyways...?

dakinggg: but you added code to support checkpoints saved with 0.16.0

tests/common/compare.py
dakinggg: no, the isinstance and the assert. is that just pyright being dumb?

mvpatel2000: I am comparing 1 and 2. I am checking if 1 is a dict and requiring 2 to be a dict as well in that case

dakinggg: ah, my b

dakinggg (approved):

PR #2479, opened 2023-08-25
MCLI Code Eval

rishab-partha:
# What does this PR do?

Example PR that can be merged when Code Eval is added to MCLI

Streamlining code eval


mvpatel2000 (commented):
This probably requires bumping setup.py as well.

composer/utils/eval_client/mcli_eval_client.py
```diff
@@ -0,0 +1,51 @@
...
+import mcli
+
+from composer.utils.eval_client.eval_client import EvalClient
+
+__all__ = ['MCLIEvalClient']
+log = logging.getLogger(__name__)
+
+
+class MCLIEvalClient(EvalClient):
+    """Utility for creating a client for and invoking an AWS Lambda through MCLI."""
```
aspfohl, line 20: I would suggest a rebrand here, similar to MosaicMLLogger. Would also make this specific to lambdas - we may want a more generic MosaicMLEvalClient in the future
```suggestion
class MosaicMLLambdaEvalClient(EvalClient):
    """Utility for creating a client for and invoking an AWS Lambda hosted on the MosaicML platform"""
```

composer/utils/eval_client/mcli_eval_client.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""MCLI compatible eval client."""
```
aspfohl, line 4: similar comment on all docstrings, comments, and variables

composer/utils/eval_client/mcli_eval_client.py
```diff
@@ -0,0 +1,51 @@
...
+        num_beams = len(payload[0])
+        num_tests = [len(generation_payload[0]) for generation_payload in payload]
+        cum_tests = (np.cumsum([0] + num_tests[:-1])*num_beams).tolist()
+        test_cases = [test_case for generation_payload in payload for beam_payload in generation_payload for test_case in beam_payload]
+        ret_helper = [False] * len(test_cases)
+        for i in range(self.num_retries):
+            try:
+                ret_helper = mcli.get_code_eval_output(test_cases)
+                break
+            except mcli.MAPIException as e:
```
aspfohl, line 44: Would recommend only retrying on >500 errors, or not retrying at all. Any insight on why this was needed @mvpatel2000

composer/utils/eval_client/mcli_eval_client.py
```diff
@@ -0,0 +1,51 @@
...
+
+class MCLIEvalClient(EvalClient):
+    """Utility for creating a client for and invoking an AWS Lambda through MCLI."""
+
+    def __init__(self, backoff: int = 3, num_retries: int = 5) -> None:
+        """Checks that the requisite environment variables are in the EvalClient.
+
+        There must be MOSAICML_PLATFORM to be on MCLI.
+        """
+        if os.environ.get('MOSAICML_PLATFORM', False) == False:
```
aspfohl, line 27: Would share MOSAICML_PLATFORM_ENV_VAR from the logger

aspfohl, line 27: This isn't really true right? You just need to have an api key or access token configured (which the mosaicml platform does for you). You could, say, run this locally or manually on a remote machine with mcli configured

It was slightly different for the MosaicMLLogger, because we also relied on the run name environment variable, which the platform sets

composer/utils/eval_client/mcli_eval_client.py
mvpatel2000: Cloud seems to fail transiently, e.g. timeout. We've seen this with Mosaicml logger?

aspfohl: This would likely be a `TimeoutError` error type, and not `MAPIException`

What I was originally saying here was any `MAPIException` with error code 400-499 (eg bad request) should not do retries

mvpatel2000 (approved):

PR #2496, opened 2023-08-31
Fail count on mosaicml logger

mvpatel2000:
# What does this PR do?

Disables mosaicml logger after 3 failed attempts to log to avoid doom loop


eracah (approved):
LGTM

PR #2512, opened 2023-09-04
Annotate `torch_prof_remote_file_name` as Optional

srstevenson:
The `torch_prof_remote_file_name` argument of `Profiler` is passed as the `remote_file_name` argument of `TorchProfiler`, which supports passing `None` to disable uploading trace files. Prior to this commit, passing `None` to `Profiler` to do this whilst using a static type checker led to a type error.

### DIFF
--- a/composer/profiler/profiler.py
+++ b/composer/profiler/profiler.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import logging
 import pathlib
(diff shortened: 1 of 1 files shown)


mvpatel2000 (approved):
Good catch! FYI it seems github is having issues so CI/CD is down for now...

PR #2514, opened 2023-09-05
Remove metric saving

mvpatel2000:
# What does this PR do?

Adds a flag gating saving metrics, which is false by default. Sharded checkpoints now also raise a value error instead of a warning when disabling metric saving. 

[GRT-2378](https://databricks.atlassian.net/browse/GRT-2378)

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -539,7 +539,7 @@ class State(Serializable):
         if self.fsdp_sharded_state_dict_enabled and self.save_metrics:
             # Sharded state dict breaks in many different ways with torchmetrics, due to both sharding
             # metric tensors and only sometimes flattening path names in state dict and _computed, so
-            # we saving metrics is not allowed with sharded state dict.
+            # saving metrics is not allowed with sharded state dict.
             raise ValueError(
                 textwrap.dedent('Saving metrics is not allowed with sharded state dict as metric tensors will '
                                 'be sharded and break on load. If you wish to save metric state, set '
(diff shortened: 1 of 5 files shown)


composer/core/state.py
```diff
@@ -531,6 +536,14 @@ def __init__(
                 textwrap.dedent(
                     "FSDP state_dict_type='local' is deprecated in torch>=2.0.0. "
                     "Please set fsdp_config['state_dict_type']='sharded' instead and will be removed in v0.17"))
+        if self.fsdp_sharded_state_dict_enabled and self.save_metrics:
+            # Sharded state dict breaks in many different ways with torchmetrics, due to both sharding
+            # metric tensors and only sometimes flattening path names in state dict and _computed, so
+            # we saving metrics is not allowed with sharded state dict.
```
dakinggg, line 542:
```suggestion
            # saving metrics is not allowed with sharded state dict.
```

composer/trainer/trainer.py
```diff
@@ -2115,6 +2120,7 @@ def _train_loop(self) -> None:
                     # did (e.g. duration specified in samples/batches/tokens), but it is still
                     # the end of the dataloader (i.e. next(dataloader) would raise StopIteration)
                     if self.state.train_metrics is not None:
+                        self.state.train_metrics = self._ensure_metrics_device_and_dtype(self.state.train_metrics)
```
dakinggg, line 2123: why this change?

dakinggg (approved):

composer/trainer/trainer.py
mvpatel2000: 5 batches per epoch, i checkpoint after batch 5. 

When checkpoint loads, model starts epoch 0 and then hits the end of epoch section because dataloader is exhausted. It tries to compute metric and hits metric is not dense (because we normally do this in the update call, but there were no batches run for epoch 0). 

This was previously not an issue as resuming a checkpoint meant loading metrics which were subsequently moved to GPU in load logic. 

There's a unit test that verifies this behavior in `load_checkpoint.py::test_resumption`

PR #2515, opened 2023-09-06
Fix daily tests by removing gpu marker

j316chuck:
# What does this PR do?

Updates `mcp_pytest` logging to include full unbuffered output and fixes daily tests by removing gpu marker from docs. 

# Validation: 

Before PR failed daily test on mixup_test.md: 
- https://github.com/mosaicml/composer/actions/runs/6089103543/job/16521181092

After PR no more failed daily test: 
- https://github.com/mosaicml/composer/actions/runs/6114292229 

# Future Work:
- Need to do some investigation on why this test is failing on GPU markers. Might be an issue with how `pytest-codeblocks` is clearing memory. For now just make sure all doctests run on cpu and not gpu.

### DIFF
--- a/.github/mcli/mcli_pytest.py
+++ b/.github/mcli/mcli_pytest.py
@@ -66,6 +66,8 @@ if __name__ == '__main__':
 
     export COMMON_ARGS="-v --durations=20 -m '{args.pytest_markers}' {s3_bucket_flag}"
 
+    export PYTHONUNBUFFERED=1
+
     make test PYTEST='{args.pytest_command}' EXTRA_ARGS="$COMMON_ARGS --codeblocks"
 
     make test-dist PYTEST='{args.pytest_command}' EXTRA_ARGS="$COMMON_ARGS" WORLD_SIZE=2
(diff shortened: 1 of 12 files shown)


mvpatel2000:
Are we hitting timeouts?

j316chuck:
I was experimenting with timeouts as a potential reason but it seems like it has more to do with our downloading of files/starting the test. 

Seen on the last three daily tests: 
- https://github.com/mosaicml/composer/actions/runs/6089103543/job/16521181092 
- https://github.com/mosaicml/composer/actions/runs/6090920591/job/16526640112 
- https://github.com/mosaicml/composer/actions/runs/6093884904/job/16534343207


Screenshot 2023-09-06 at 10 49 10 AM

j316chuck:
Running this test locally and there seems to be no hanging. Something is wrong with the daily configuration wifi/download hmm. 

Screenshot 2023-09-06 at 11 36 34 AM

mvpatel2000 (approved):
It won't let me suggest, but there's another GPU tag farther down in same readme we should remove...

PR #2519, opened 2023-09-08
Fix GCSObjectStore bug where hmac keys auth doesn't work

eracah:
# What does this PR do?
* modified `GCSObjectStore` to enables hmac keys to work by creating an S3ObjectStore object under the hood
* added two integration tests to test both auth procedures: `test_gs_object_store_integration_json_auth` and `test_gs_object_store_integration_hmac_auth

# Manual Tests
Ran:
1. `pytest tests/utils/object_store/test_gs_object_store.py::test_gs_object_store_integration_json_auth -m remote`
2. `pytest tests/utils/object_store/test_gs_object_store.py::test_gs_object_store_integration_hmac_auth -m remote`

Manual test 2 failed before the fix.

fixes GRT-2391


composer/utils/object_store/gcs_object_store.py
```diff
@@ -115,6 +117,9 @@ def get_object_size(self, object_name: str) -> int:
             FileNotFoundError: If the specified object does not exist in the cloud storage bucket.
             Exception: If an error occurs while trying to retrieve the object's size.
         """
+        if not self.use_gcs_sdk and self.s3_object_store is not None:
```
dakinggg, line 120: why do you need the s3 none check? shouldn't it always be not none if use sdk is false?

dakinggg: (same comment in all places)

tests/utils/object_store/test_gs_object_store.py
```diff
@@ -7,8 +7,40 @@
...
 from botocore.exceptions import ClientError
+from torch.utils.data import DataLoader
 
+from composer.trainer import Trainer
 from composer.utils import GCSObjectStore
+from tests.common import RandomClassificationDataset, SimpleModel
+
+
+@pytest.mark.remote
+def test_gs_object_store_integration_json_auth():
```
dakinggg, line 18: can you assert in here that the object store that gets created uses the sdk version?

dakinggg: since you use this for both tests, probs need a flag for which version its supposed to use, but basically just check that you're actually testing both versions.

composer/utils/object_store/gcs_object_store.py
eracah: to please the pyright gods!

dakinggg: ah, can you assert instead? otherwise it makes me worried about what happens if s3 object store is none cause it seems like thats possible

tests/utils/object_store/test_gs_object_store.py
eracah: ok added a check

dakinggg (approved):

b-chu (changes requested):
Why don't GCS hmac keys work anymore? Can we support GCS in a similar manner to streaming with the official python sdk (which is recommended and allows for more flexibility in authentication)?

https://github.com/mosaicml/streaming/pull/376

eracah:
> Why don't GCS hmac keys work anymore? Can we support GCS in a similar manner to streaming with the official python sdk (which is recommended and allows for more flexibility in authentication)?
> 
> [mosaicml/streaming#376](https://github.com/mosaicml/streaming/pull/376)

they work with this fix!

PR #2529, opened 2023-09-12
Squelch Memory Monitor warnings if device=meta

hanlint:
Previously, the memory monitor would emit a warning if the model, at start, is on `meta` device. This PR removes that warning, since the memory monitor will still work later if the model is moved to cuda.


mvpatel2000 (commented):
Good catch!

Can you add a small unit test for this? Should just be copy paste an existing test but with a meta model?

mvpatel2000 (approved):

PR #2531, opened 2023-09-12
Fix metrics keys sort in DecoupledAdamW for OptimizerMonitor FSDP metric agreggation

### LINKED ISSUE #2513: `DecoupledAdamW`'s `all_reduce` operation for `OptimizerMonitor` returns wrong results
I've posted the issue in `llm-foundry` repo but after looking at it decided to post it here to as the problem is mostly with the `composer` work and code. The link to issue with the examples and environment output is below.

[`llm-foudnry` issue](https://github.com/mosaicml/llm-foundry/issues/578)

---

While using the `llm-foundry` code to pretrain the model the issue with the `OptimizerMonitor` logged values was found. As it turns out the issue is actually caused by the wrong results of `all_reduce` operation that is used to collect the results from all ranks.

We used `llm-foundry` to pretrain the model. To begin with and reproduce results I used provided MPT 1B config that was changed to use custom tokenizer and `SHARD_GRAD_OP` FSDP strategy with mixed precision (bf16 for forward pass, fp32 for others). Values logged by `OptimizerMonitor` had "jumps" after resuming training both using `autoresume` option and continuing from specific checkpoint. It turns out that the final value that is obtained by `pre_reduce_metrics` and `dist_reduce_metrics` provided by optimizer (using `DecoupledAdamW`, which is default in config) is different across ranks.

After narrowing it down it was found that the actual `all_reduce` operation to collect metric values from all ranks provides wrong results ([link to line](https://github.com/mosaicml/composer/blob/dev/composer/optim/decoupled_weight_decay.py#L353)). It is said that the result is "bitwise identical on all ranks" but it not. During exploration I used the L2 norm of  the weight of the specific attention matrix. Using 8 GPUs and `SHARD_GRAD_OP` FSDP strategy the param's values are present on ranks 0 and 1. After `all_reduce` operation their norms are almost always does not change, 0 values on other ranks can change a little and become very small number that is not 0, but can also become just some random number. The `all_gather` operation returns bad results too that correlate with the results obtained by `all_reduce` operation.

At the same time initializing test variable with the rank number as a value in the same code block and then applying `all_reduce` operation to it returns expected correct result every time. So it seems that the problem is with this particular variable. Also using regular DDP provides expected and correct results, i.e. values across all ranks are the same and correct.

**To reproduce**

1. Clone `composer` (dev branch, the default one) from the GitHub repo and install it using -e option to be able to modify it.
2. Clone `llm-foundry` repo.
3. Use MPT 1B config, C4 subset from the README, SHARD_GRAD_OP strategy for FSDP config, bf16 mixed precision for forward, fp32 for buffer and reduce.
4. Print values for block 12 attention matrix before and after `all_reduce` operation in `composer`'s optimizer to get logged L2 norms (NAME).
5. Run training with `scripts/train/train.py`.

## Expected behavior

Parameter's (and gradients) L2 value that is the same across all ranks and that is the same with the one that can be obtained from `state_dict`. To be specific, `all_reduce` operation for `reduce` variable in `DecoupledAdamW` optimizer in `dist_reduce_metric` function should result in one right value across all ranks. Instead each rank has its own value.

## Additional context

Layer name: `model._fsdp_wrapped_module.transformer.blocks.12._fsdp_wrapped_module.attn.Wqkv.weight`.

Code for optimizer class to print values: 

```python
        all_keys = sorted(all_keys, key=lambda metric: 0 if 'l2_norm' in metric else 1)
        for metric in all_keys:
            if metric.startswith('l2_norm'):
                dist.barrier()

                tensors_gather_list = [
                    torch.tensor(0.0, dtype=torch.float32, device=torch.cuda.current_device())
                    for _ in range(dist.get_world_size())
                ]
                # tensors_gather_list = None

                test_reduce = torch.tensor(rank, dtype=torch.float32, device=torch.cuda.current_device())

                if metric in optimizer_metrics:
                    reduced_l2 = torch.tensor(optimizer_metrics[metric].item(), dtype=torch.float32, device=torch.cuda.current_device())
                else:
                    reduced_l2 = torch.tensor(0.0, dtype=torch.float32, device=torch.cuda.current_device())

                # old way
                # reduced = optimizer_metrics.get(metric, torch.tensor(0.0, device=torch.cuda.current_device()))

                # ------------------- DEBUG PRINTS -------------------
                if metric == opt_metrics_name:
                    print(f'rank {rank} | dist_reduce_metrics initial reduced value: {reduced_l2}')
                    print(f'rank {rank} | dist_reduce_metrics initial reduced value device: {reduced_l2.device}')

                    print(f'rank {rank} | test_reduce value: {test_reduce}')
                    print(f'rank {rank} | test_reduce value device: {test_reduce.device}')

                    print('--')
                # ------------------- DEBUG PRINTS -------------------

                if dist.get_world_size() > 1:
                    dist.barrier()

                    # tensors_gather_list = dist.all_gather(test_reduce)
                    torch.distributed.all_gather(tensors_gather_list, reduced_l2)

                    # torch.distributed.reduce(reduced_l2, dst=0)
                    dist.all_reduce(reduced_l2, reduce_operation='SUM')


                    dist.all_reduce(test_reduce, reduce_operation='SUM')

                dist.barrier()

                # ------------------- DEBUG PRINTS -------------------
                if metric == opt_metrics_name:
                    print(f'rank {rank} | tensors_gather_list: {tensors_gather_list}')

                    print(f'rank {rank} | dist_reduce_metrics reduced value: {reduced_l2}')
                    print(f'rank {rank} | dist_reduce_metrics reduced value sq root: {math.sqrt(reduced_l2)}')

                    print(f'rank {rank} | test_reduce value reduced: {test_reduce}')
                    print(f'rank {rank} | test_reduce value reduced device: {test_reduce.device}')

                    print('--------------------------------------------------------')
                # ------------------- DEBUG PRINTS -------------------

                dist.barrier()

                optimizer_metrics[metric] = math.sqrt(reduced_l2)
```

Example output:

example_output

example_output_v2

m1kol:
# What does this PR do?

Fix `optimizer_metrics` keys sorting in `DecoupledAdamW` `pre_reduce_metrics` and `dist_reduce_metrics` functions that are used to aggregate values for `OptimizerMonitor` in case of FSDP use so that keys order is the same on every rank.

- Apply regular sort function (`sorted`) to gathered keys without `key` parameter tweaks.
- Remove code blocks that were intended for `cosine` metrics as they are no longer calculated.

Fixes #2513

# Before submitting
- [ ] ~Is this change a documentation change or typo fix? If so, skip the rest of this checklist.~
- [ ] ~Did you update any related docs and document your change?~
- [ ] ~Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))~
- [ ] ~Did you run the tests locally to make sure they pass?~
  - No, but the change is small and the correctness was checked during debug and training runs.
  - Via `pre-commit` install, they ran automatically on commit.

### DIFF
--- a/composer/optim/decoupled_weight_decay.py
+++ b/composer/optim/decoupled_weight_decay.py
@@ -345,7 +345,9 @@ class DecoupledAdamW(AdamW):
         for keys in all_gathered_keys:
             all_keys.update(keys)
 
-        all_keys = sorted(all_keys, key=lambda metric: 0 if 'l2_norm' in metric else 1)
+        # Sort keys to ensure every rank has the same keys order
+        # Only L2 norm metric keys are present, can apply regular sort
+        all_keys = sorted(all_keys)
         for metric in all_keys:
             if metric.startswith('l2_norm'):
                 reduced = optimizer_metrics.get(metric, torch.tensor(0.0, device=torch.cuda.current_device()))
@@ -353,18 +355,6 @@ class DecoupledAdamW(AdamW):
                     dist.all_reduce(reduced, reduce_operation='SUM')
 
                 optimizer_metrics[metric] = math.sqrt(reduced)
-            elif metric.startswith('cosine'):
-                reduced = optimizer_metrics.get(metric, torch.tensor(0.0, device=torch.cuda.current_device()))
-                if dist.get_world_size() > 1:
-                    dist.all_reduce(reduced, reduce_operation='SUM')
-
-                _, vectors, layer = tuple(metric.split('/'))
-
-                A, B = tuple(vectors.split('_'))
-
-                A_reduced_norm = optimizer_metrics[f'l2_norm/{A}/{layer}']
-                B_reduced_norm = optimizer_metrics[f'l2_norm/{B}/{layer}']
-                optimizer_metrics[metric] = reduced / (A_reduced_norm * B_reduced_norm)
             else:
                 reduced = optimizer_metrics.get(metric, torch.tensor(0.0, device=torch.cuda.current_device()))
                 if dist.get_world_size() > 1:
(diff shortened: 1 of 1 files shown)


mvpatel2000 (commented):
Thanks for the PR! This looks correct to me, but I would like @bmosaicml to confirm and approve since he wrote the original version

bmosaicml (approved):
Great work!

mvpatel2000 (approved):

PR #2535, opened 2023-09-13
Fix GCS tests

mvpatel2000:
# What does this PR do?

Fixes GCS tests, which had a bug in the test that caused failures


eracah (approved):
LGTM

PR #2538, opened 2023-09-13
Change cast for mosaicml logger

mvpatel2000:
# What does this PR do?

Change cast for mosaicml logger. Also fixes test which should have caught this but was also buggy.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1080,9 +1080,8 @@ class Trainer:
                 loggers.append(remote_ud)
 
         # MosaicML Logger
-        if bool(os.environ.get(MOSAICML_PLATFORM_ENV_VAR,
-                               False)) and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR) is not None and not any(
-                                   isinstance(x, MosaicMLLogger) for x in loggers):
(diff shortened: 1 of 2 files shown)


composer/trainer/trainer.py
```diff
@@ -1080,9 +1080,8 @@ def __init__(
                 loggers.append(remote_ud)
 
         # MosaicML Logger
-        if bool(os.environ.get(MOSAICML_PLATFORM_ENV_VAR,
-                               False)) and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR) is not None and not any(
-                                   isinstance(x, MosaicMLLogger) for x in loggers):
+        if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and os.environ.get(
```
dakinggg, line 1083: is this really set as `false` and `true`? I feel like it is more common to use 0/1 for env var flags like this

mvpatel2000: This is set on cloud end, so out of our control for now.... That is what they picked

dakinggg: oook

dakinggg (approved):

PR #2541, opened 2023-09-15
Add log_model to MLFlowLogger

dakinggg:
# What does this PR do?
Adds a `log_model` function to the `MLFlowLogger`. Currently only supports the `transformers` flavor of model logging, but it is easy to add others if we want. I don't think I can add a generic `log_model` to `LoggerDestination` because the args would be very different for any other `LoggerDestination`s that could handle logging models, but lmk what you think.

This PR also unifies the run that the client and module versions of mlflow point to by calling `mlflow.start_run(run_id=self._run_id)` and `mlflow.end_run()`. This was necessary for this PR because `transformers.log_model` is only available on the module version, but hopefully fixes some other issues as well.

Manual test: logged a model to databricks successfully
Screenshot 2023-09-15 at 12 22 42 AM
and registered
Screenshot 2023-09-15 at 1 18 33 AM

# Before submitting


dakinggg:
@dbczumar @prithvikannan mind taking a look as well?

composer/loggers/mlflow_logger.py
```diff
@@ -150,6 +152,23 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+            flavor (str): The MLFlow model flavor to use. Currently only ``'transformers'`` is supported.
+            **kwargs: Keyword arguments to pass to the MLFlow model logging function.
+
+        Raises:
+            NotImplementedError: If ``flavor`` is not ``'transformers'``.
+        """
+        if self._enabled:
+            import mlflow
+            if flavor == 'transformers':
+                mlflow.transformers.log_model(**kwargs,)
```
prithvikannan, line 168: i dont think we have a better approach, but worth noting this call will not be handled via the batching/async logging mechanism used in `log_metric`/`log_param`.

composer/loggers/mlflow_logger.py
```diff
@@ -112,6 +113,7 @@ def init(self, state: State, logger: Logger) -> None:
                     run_name=self.run_name,
                 )
                 self._run_id = new_run.info.run_id
+            mlflow.start_run(run_id=self._run_id)
```
prithvikannan, line 116: nice, yea this (along with setting the `tracking_uri` above) should fix the mlflow fluent api vs mlflow client issues

prithvikannan (approved):
LGTM from the mlflow side

composer/loggers/mlflow_logger.py
dakinggg: Yeah, will it upload the multiple files associated with a huggingface checkpoint concurrently? or is it really just sending the whole object up in one chunk?

prithvikannan: We have multipart upload logic so I think it should be fairly performant

dakinggg: great

dakinggg:
@eracah please review

eracah (approved):
Looks pretty good. Do we know what would happen if rank_zero_only was false. Seems like it might not tie the correct run to the module maybe?

dakinggg:
@eracah not sure, but I think that would've been an issue before this PR too right?

PR #2542, opened 2023-09-15
Log runtime estimator units

mvpatel2000:
# What does this PR do?

Log runtime estimator units

### DIFF
--- a/composer/callbacks/runtime_estimator.py
+++ b/composer/callbacks/runtime_estimator.py
@@ -48,6 +48,8 @@ class RuntimeEstimator(Callback):
     +===================================+=========================================================+
     | `time/remaining_estimate`         | Estimated time to completion                            |
     +-----------------------------------+---------------------------------------------------------+
+    | `time/remaining_estimate_unit`    | Unit of time specified by user                          |
+    +-----------------------------------+---------------------------------------------------------+
 
     Args:
         skip_batches (int, optional): Number of batches to skip before starting clock to estimate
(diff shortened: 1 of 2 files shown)


composer/callbacks/runtime_estimator.py
```diff
@@ -64,6 +66,7 @@ def __init__(self, skip_batches: int = 1, time_unit: str = 'hours') -> None:
         self.start_dur = None
         self.train_dataloader_len = None
 
+        self.time_unit = time_unit
```
aspfohl, line 69: Recommend this, or updating line 111 to only log when its not none
```suggestion
        self.time_unit = time_unit or 'hours'
```

aspfohl (approved):
`time_unit` will only be 'seconds', 'minutes', 'hours', or 'days' right? Not any of the values of `TimeUnit` in composer core?

composer/callbacks/runtime_estimator.py
aspfohl: Oh wait I'm dumb, `time_unit: str = 'hours'` should be good enough above 😛

composer/callbacks/runtime_estimator.py
```diff
@@ -48,6 +48,8 @@ class RuntimeEstimator(Callback):
     +===================================+=========================================================+
     | `time/remaining_estimate`         | Estimated time to completion                            |
     +-----------------------------------+---------------------------------------------------------+
+    | `time/remaining_estimate_unit`    | Unit of time specified by user                          |
```
dakinggg, line 51: list the options in the docs here

dakinggg (commented):
Unit test and manual test please

PR #2549, opened 2023-09-18
Remove system NCCL

mvpatel2000:
# What does this PR do?

Removes system NCCL so the torch version is always used

[GRT-2428](https://databricks.atlassian.net/browse/GRT-2428)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -149,6 +149,12 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+######################################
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -280,6 +280,20 @@ RUN if [ -n "$MOFED_VERSION" ] ; then \
...
 
+######################################
+# Purge system NCCL (comes with torch)
+######################################
+
+RUN ls /
+
+RUN ls /usr
+
+RUN ls / usr/lib
```
j316chuck, line 291: I think this command will error out but not sure if this is blocking the build

PR #2565, opened 2023-09-25
Add version number to composer metadata logs

j316chuck:
# What does this PR do?
Add version number to composer metadata logs

Users have difficulty debugging which version of composer was running. It should be more clear now after 0.16.2. 

# Testing
Added unit test


composer/loggers/mosaicml_logger.py
```diff
@@ -88,6 +89,8 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
         self._log_metadata(metrics)
 
     def after_load(self, state: State, logger: Logger) -> None:
+        # Log composer version
+        self._log_metadata({'composer_version': composer_version})
```
mvpatel2000, line 93: I don't think this should be here. It should probably just be directly in `composer.__init__`?

j316chuck: Thanks for the suggestion! Fixed 👍

composer/trainer/trainer.py
```diff
@@ -1159,6 +1160,9 @@ def __init__(
             self.local_hparams = extract_hparams(locals())
             self.logger.log_hyperparameters(self.local_hparams)
 
+        # Log composer version.
+        self.logger.log_hyperparameters({'composer_version': composer_version})
```
dakinggg, line 1164: should be able to just use `composer.__version__`?

dakinggg, line 1164: if we're going to log this, lets also log the git commit if it exists? there is a utility func for this already in `collect_env.py`

dakinggg: but also might as well use the `get_composer_version` utility func in `collect_env.py`

j316chuck: fixed

j316chuck: fixd

dakinggg (approved):

composer/trainer/trainer.py
```diff
@@ -1159,6 +1159,11 @@ def __init__(
             self.local_hparams = extract_hparams(locals())
             self.logger.log_hyperparameters(self.local_hparams)
 
+        # Log composer version.
+        composer_env_dict = get_composer_env_dict()
+        self.logger.log_hyperparameters({'composer_version': composer_env_dict['composer_version']})
+        self.logger.log_hyperparameters({'composer_commit_hash': str(composer_env_dict['composer_commit_hash'])})
```
mvpatel2000, line 1165:
```suggestion
        # Log composer version
        composer_env_dict = get_composer_env_dict()
        self.logger.log_hyperparameters({'composer_version': composer_env_dict['composer_version']})
        self.logger.log_hyperparameters({'composer_commit_hash': str(composer_env_dict['composer_commit_hash'])})
```

eracah (approved):
LGTM

PR #2577, opened 2023-09-28
Add a property for each LoggerDestination to Logger

eracah:
# What does this PR do?
adds a property decorator for each loggerdestination, so a user can easily get a loggerdestination by just calling `mlflow_logger = trainer.logger.mlflow` for example


composer/loggers/logger.py
```diff
@@ -147,6 +147,127 @@ def has_file_upload_destination(self) -> bool:
...
+    def _get_logger_destination(self, logger_type) -> Optional[Union[LoggerDestination, Sequence[LoggerDestination]]]:
+        """Returns instances of the specified logger destination, if it/they exist."""
+        # Check every time just in case user appended to destinations list later.
+        logger_destinations = [destination for destination in self.destinations if isinstance(destination, logger_type)]
+        if len(logger_destinations) == 1:
+            return logger_destinations[0]
+        elif len(logger_destinations) > 1:
+            return logger_destinations
+        else:  # length is 0
+            return None
```
mvpatel2000, line 159: Can we type all of the functions so it's easy for users?

Also, I'm not sure i love its sometimes list and sometimes not because end user then always has to check for both scenarios. Can we say it always returns a list? Because usually I just want to apply operation to all logger destinations of that type

PR #2579, opened 2023-09-29
Add save and register wrappers to mlflow logger

dakinggg:
# What does this PR do?
Adds simple wrappers around mlflow save and register model functions, similar to the log model function

# Before submitting

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -47,8 +47,8 @@ class MLFlowLogger(LoggerDestination):
         model_registry_prefix (str, optional): The prefix to use when registering models.
             If registering to Unity Catalog, must be in the format ``{catalog_name}.{schema_name}``.
             (default: empty string)
-        model_registry_uri (str, optional): The URI of the model registry to use. If not set, will
-            default to ``databricks-uc``.
+        model_registry_uri (str, optional): The URI of the model registry to use. To register models
+            to Unity Catalog, set to ``databricks-uc``. (default: None)
     """
 
     def __init__(
@@ -59,7 +59,7 @@ class MLFlowLogger(LoggerDestination):
         rank_zero_only: bool = True,
         flush_interval: int = 10,
         model_registry_prefix: str = '',
-        model_registry_uri: str = 'databricks-uc',
+        model_registry_uri: Optional[str] = None,
     ) -> None:
         try:
             import mlflow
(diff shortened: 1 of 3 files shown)


smurching (approved):
Looks good from an MLflow usage perspective!

composer/loggers/mlflow_logger.py
```diff
@@ -63,12 +73,22 @@ def __init__(
...
+            if len(self.model_registry_prefix.split('.')) != 2:
+                raise ValueError(f'When registering to Unity Catalog, model_registry_prefix must be in the format ' +
+                                 f'{{catalog_name}}.{{schema_name}}, but got {self.model_registry_prefix}')
+
         self._rank_zero_only = rank_zero_only
         self._last_flush_time = time.time()
         self._flush_interval = flush_interval
         if self._enabled:
             self.tracking_uri = str(tracking_uri or mlflow.get_tracking_uri())
             mlflow.set_tracking_uri(self.tracking_uri)
```
smurching, line 88: @dbczumar @harupy one thing I'm realizing - there's no `MlflowClient` APIs currently for doing things like logging models or achieving functionality like `mlflow.register_model`, etc, which means that library code like this needs to set the global registry/tracking URIs, which affects any user code that runs subsequently and calls MLflow APIs (or the logger would need to remember/restore the original tracking/registry URIs)

Not sure if it's actually possible for additional user code to run after this particular logger is used :) and doesn't seem like a huge issue for MLflow in general, but just something interesting to note

dakinggg: yeah, I had to setup the module version of mlflow to match the client version in order for things to go to the right places. i assume that while this logger is active the user would be restricted to the setup of this logger (or if they deviated, things would get weird)

eracah (approved):
Looks pretty good. Also, I see you snuck in a fix for the gs_object_store thing. Manual test?

dakinggg:
@eracah oh woops, i meant to remove that gs thing. chuck is fixing the disk space ci issue in a separate pr. although the change was just in the gs object store tests, so there wasn't really anything to manual test anyway. just if it made ci not run out of disk space (which it did not).

tests/utils/object_store/test_gs_object_store.py
```diff
@@ -114,7 +114,7 @@ def test_download_object(gs_object_store, monkeypatch, tmp_path, result: str):
 
     def generate_dummy_file(x):
         with open(x, 'wb') as fp:
-            fp.write(bytes('0' * (1024 * 1024 * 1024), 'utf-8'))
+            fp.write(bytes('0' * (100), 'utf-8'))
```
dakinggg, line 117:
```suggestion
            fp.write(bytes('0' * (1024 * 1024 * 1024), 'utf-8'))
```

eracah:
> @eracah oh woops, i meant to remove that gs thing. chuck is fixing the disk space ci issue in a separate pr. although the change was just in the gs object store tests, so there wasn't really anything to manual test anyway. just if it made ci not run out of disk space (which it did not).

Oh I meant manual test for this PR

dakinggg:
oh yeah, i've landed models using this foundry pr (https://github.com/mosaicml/llm-foundry/pull/618) + this pr.

PR #2583, opened 2023-09-29
Fix pytest disk space OOM issue by adding tmp_path_retention_policy=None

j316chuck:
# Description

Composer's pytests use the `tmp_path` fixture which continually appends to the `base_tmpdir` over the course of a pytest session. This causes disk OOM errors when we run our full `pytest` suite with`base_tmpdir` reaching over 20+ GB of tmp files.

This PR removes the `tmp_path` per test function by adding [tmp_path_retention_policy=None
](https://docs.pytest.org/en/7.3.x/reference/reference.html#confval-tmp_path_retention_policy)
# Along for the ride
- Some code clean ups where we make the file path usage smaller. This was our initial hypothesis for the which we thought was the original issue. 

# Tests
Before:
- [Failing torch 2.1 pytests ](https://github.com/mosaicml/composer/actions/runs/6374759750/job/17299755748) (due to OOM)
- [Disk Space: 0 GB](https://github.com/mosaicml/composer/actions/runs/6357809391)

After:
- [Passing torch 2.1 pytests](https://github.com/mosaicml/composer/actions/runs/6382974386/job/17322664050)
- [Disk space: 22GB](https://github.com/mosaicml/composer/actions/runs/6382974386) 
- [Passing all tests](https://github.com/mosaicml/composer/actions/runs/6383529732/job/17324883764)

Screenshot 2023-10-02 at 10 46 45 AM
Screenshot 2023-10-02 at 10 47 22 AM

### DIFF
--- a/.github/workflows/pr-cpu.yaml
+++ b/.github/workflows/pr-cpu.yaml
@@ -32,7 +32,7 @@ jobs:
             pytest_command: 'coverage run -m pytest'
             composer_package_name: 'mosaicml'
           - name: 'cpu-3.10-2.1'
-            container: mosaicml/pytorch:2.1.0_cu121-nightly20230903-python3.10-ubuntu20.04
+            container: mosaicml/pytorch:2.1.0_cu121-nightly20230827-python3.10-ubuntu20.04
             markers: 'not daily and not remote and not gpu and not vision and not doctest'
             pytest_command: 'coverage run -m pytest'
             composer_package_name: 'mosaicml'
--- a/.github/workflows/pr-cpu.yaml
+++ b/.github/workflows/pr-cpu.yaml
@@ -31,11 +31,11 @@ jobs:
             markers: 'not daily and not remote and not gpu and not vision and not doctest'
             pytest_command: 'coverage run -m pytest'
             composer_package_name: 'mosaicml'
-          - name: 'cpu-3.10-2.1'
-            container: mosaicml/pytorch:2.1.0_cu121-nightly20230827-python3.10-ubuntu20.04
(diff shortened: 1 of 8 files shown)


.github/workflows/pr-cpu.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
             pytest_command: 'coverage run -m pytest'
             composer_package_name: 'mosaicml'
           - name: 'cpu-3.10-2.1'
-            container: mosaicml/pytorch:2.1.0_cu121-nightly20230903-python3.10-ubuntu20.04
+            container: mosaicml/pytorch:2.1.0_cu121-nightly20230827-python3.10-ubuntu20.04
```
j316chuck, line 35: Switching to docker image we usually use

tests/loggers/test_wandb_logger.py
```diff
@@ -40,6 +40,8 @@ def download(self, root: pathlib.Path):
 def test_wandb_logger(tmp_path, dummy_state):
     pytest.importorskip('wandb', reason='wandb is optional')
     os.environ['WANDB_DIR'] = str(tmp_path)
+    if not tmp_path.exists():
+        tmp_path.mkdir()
```
mvpatel2000, line 44: Why does this need to happen? Shouldn't it always exist?

mvpatel2000 (commented):
Can we make the same changes in foundry?

tests/loggers/test_wandb_logger.py
j316chuck: ya I can  remove this, this solution was from our manual deletion of tmp_dir solution

mvpatel2000 (approved):

PR #2585, opened 2023-09-29
Use FSDP CustomPolicy to support custom kwargs passed to different wrapped modules

cli99:
This PR uses FSDP CustomPolicy (introduced in pytorch 2.1.0) to support custom kwargs (e.g. pre-defined process group) passed to different wrapped modules. 
This change gets rid of the need to monkey patch fsdp `__init__`.

### DIFF
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -25,6 +25,7 @@ __all__ = ['DDPSyncStrategy', 'ddp_sync_context', 'prepare_ddp_module', 'prepare
 
 log = logging.getLogger(__name__)
 
+process_group_cache = {}
 
 class DDPSyncStrategy(StringEnum):
     """How and when gradient synchronization should happen.
@@ -458,8 +459,23 @@ def prepare_fsdp_module(
 
                 def _auto_wrap_policy_new(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
                     return __auto_wrap_policy(module, recurse, nonwrapped_numel)
+                from torch.distributed.fsdp.wrap import CustomPolicy
 
-                _auto_wrap_policy = _auto_wrap_policy_new
+                def lambda_fn(module: torch.nn.Module) -> Union[bool, dict]:
+                    ret = False
+                    if hasattr(module, '_fsdp_wrap'):
+                        ret = bool(module._fsdp_wrap)
+                    elif hasattr(obj, 'fsdp_wrap_fn') and isinstance(obj.fsdp_wrap_fn, Callable):
+                        ret = obj.fsdp_wrap_fn(module)
+                        from composer.trainer.mosaic_fsdp_utils import _set_custom_fsdp_module_kwargs
+                        if isinstance(ret, dict):
+                            ret = _set_custom_fsdp_module_kwargs(ret, process_group_cache)
+                    if ret and auto_microbatching:
+                        module.register_forward_hook(sync_hook)
+                        module.register_full_backward_hook(sync_hook)
+                    return ret
+                policy = CustomPolicy(lambda_fn)
+                _auto_wrap_policy = policy
 
             else:
 
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -461,24 +461,28 @@ def prepare_fsdp_module(
                 def _auto_wrap_policy_new(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
                     return __auto_wrap_policy(module, recurse, nonwrapped_numel)
 
-                from torch.distributed.fsdp.wrap import CustomPolicy
+                if version.parse(torch.__version__) > version.parse('2.0.1-rc4'):
+                    # CustomPolicy is only supported in torch v2.1.0-rc1 or higher
+                    from torch.distributed.fsdp.wrap import CustomPolicy
 
-                def lambda_fn(module: torch.nn.Module) -> Union[bool, dict]:
-                    ret = False
-                    if hasattr(module, '_fsdp_wrap'):
-                        ret = bool(module._fsdp_wrap)
-                    elif hasattr(obj, 'fsdp_wrap_fn') and isinstance(obj.fsdp_wrap_fn, Callable):
-                        ret = obj.fsdp_wrap_fn(module)
-                        from composer.trainer.mosaic_fsdp_utils import _set_custom_fsdp_module_kwargs
-                        if isinstance(ret, dict):
-                            ret = _set_custom_fsdp_module_kwargs(ret, process_group_cache)
-                    if ret and auto_microbatching:
-                        module.register_forward_hook(sync_hook)
-                        module.register_full_backward_hook(sync_hook)
-                    return ret
+                    def lambda_fn(module: torch.nn.Module) -> Union[bool, dict]:
+                        ret = False
+                        if hasattr(module, '_fsdp_wrap'):
+                            ret = bool(module._fsdp_wrap)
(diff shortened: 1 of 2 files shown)


composer/trainer/dist_strategy.py
```diff
@@ -459,7 +461,28 @@ def __auto_wrap_policy(module: torch.nn.Module, recurse: bool, nonwrapped_numel:
                 def _auto_wrap_policy_new(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
                     return __auto_wrap_policy(module, recurse, nonwrapped_numel)
 
-                _auto_wrap_policy = _auto_wrap_policy_new
+                if version.parse(torch.__version__) > version.parse('2.0.1-rc4'):
```
mvpatel2000, line 464: Should this be 2.1.0?

composer/trainer/dist_strategy.py
```diff
@@ -459,7 +461,28 @@ def __auto_wrap_policy(module: torch.nn.Module, recurse: bool, nonwrapped_numel:
...
+                            from composer.trainer.mosaic_fsdp_utils import _set_custom_fsdp_module_kwargs
+                            if isinstance(ret, dict):
+                                ret = _set_custom_fsdp_module_kwargs(ret, process_group_cache)
+                        if ret and auto_microbatching:
+                            module.register_forward_hook(sync_hook)
+                            module.register_full_backward_hook(sync_hook)
+                        return ret
+
+                    policy = CustomPolicy(lambda_fn)
+                    _auto_wrap_policy = policy
```
mvpatel2000, line 483: Can we add this as another (first case) to the if statement one level above? Because this does not use `_auto_wrap_policy_new`

cli99: done. thanks

composer/trainer/mosaic_fsdp.py
```diff
@@ -39,16 +39,5 @@ def patch_pytorch():
...
-        # Monkey path for torch < 2.1.1 ie torch == 2.1.0
-
-        # Monkey patch __init__ where __init__ calls the custom _auto_wrap fn
-        from composer.trainer.mosaic_fsdp_utils import init_fn_t2p1p0
-        FullyShardedDataParallel.__init__ = init_fn_t2p1p0  # type: ignore
-
-        # Monkey patch sharding method
-        ChunkShardingSpec.build_metadata = build_metadata
-        ChunkShardingSpec.shard = shard
-
```
mvpatel2000, line 52: The `ChunkShardingSpec` monkeypatch needs to stay to fix checkpointing things. Only these two lines can be removed:
```
        # Monkey patch __init__ where __init__ calls the custom _auto_wrap fn
        from composer.trainer.mosaic_fsdp_utils import init_fn_t2p1p0
        FullyShardedDataParallel.__init__ = init_fn_t2p1p0  # type: ignore
```

composer/trainer/dist_strategy.py
cli99: `version` compares somewhat lexicographically, e.g. version.parse('2.1.0-rc1') <  version.parse('2.1.0') or version.parse('2.1.0-dev') <  version.parse('2.1.0')

mvpatel2000: @cli99  I think this is resulting in this code path being triggered even for `2.0.1`. It is currently `2.0.1-rc4` where I think it should be `2.1.0-rc4`

composer/trainer/mosaic_fsdp.py
cli99: got it. good catch. thanks.

composer/trainer/dist_strategy.py
cli99: @mvpatel2000, you are right. However, for torch.__version_ == '2.1.0.devXXX', the condition is still evaluated to false if we use `2.1.0-rc4`, maybe we shall use `if version.parse(torch.__version__) > version.parse('2.1.0.dev')`:

mvpatel2000: That seems reasonable to me! if it works let's use it

vchiley (changes requested):
This PR should also remove all monkey patch utils created for torch2.1 ie rm [`_custom_or_policy_t2p1p0`](https://github.com/mosaicml/composer/blob/8e3b0381ca5aab644a901cc58757984b9d1e1aef/composer/trainer/mosaic_fsdp_utils.py#L726), [`_custom_recursive_wrap_t2p1p0`](https://github.com/mosaicml/composer/blob/8e3b0381ca5aab644a901cc58757984b9d1e1aef/composer/trainer/mosaic_fsdp_utils.py#L744), and [`init_fn_t2p1p0`](https://github.com/mosaicml/composer/blob/8e3b0381ca5aab644a901cc58757984b9d1e1aef/composer/trainer/mosaic_fsdp_utils.py#L831C9-L831C23) from [`mosaic_fsdp_utils.py`](https://github.com/mosaicml/composer/blob/8e3b0381ca5aab644a901cc58757984b9d1e1aef/composer/trainer/mosaic_fsdp_utils.py)

composer/trainer/dist_strategy.py
```diff
@@ -437,36 +439,56 @@ def _param_init_fn(module: torch.nn.Module) -> None:
...
+                        module.register_forward_hook(sync_hook)
+                        module.register_full_backward_hook(sync_hook)
+                    return ret
+
+                _auto_wrap_policy = CustomPolicy(lambda_fn)
             else:
+                # Choose which modules to FSDP wrap according to the following priority:
+                # If module has attribute `module._fsdp_wrap = ...`, always respect it
+                # Otherwise wrap if root object `obj.fsdp_wrap_fn(module)` is true.
+                def __auto_wrap_policy(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
```
b-chu, line 465: Is there a better name for this function that doesn't differentiate by one vs two underscores?

composer/trainer/dist_strategy.py
```diff
@@ -437,36 +439,56 @@ def _param_init_fn(module: torch.nn.Module) -> None:
...
+                        should_be_wrapped = obj.fsdp_wrap_fn(module)
+
+                    if should_be_wrapped and auto_microbatching:
+                        module.register_forward_hook(sync_hook)
+                        module.register_full_backward_hook(sync_hook)
+                    return should_be_wrapped
+
+                if is_torch_2_0:
+
+                    def _auto_wrap_policy_new(module: torch.nn.Module, recurse: bool, nonwrapped_numel: int) -> bool:
```
b-chu, line 481: Use a more descriptive name for this function.

composer/trainer/dist_strategy.py
```diff
@@ -437,36 +439,56 @@ def _param_init_fn(module: torch.nn.Module) -> None:
...
+                        should_be_wrapped = bool(module._fsdp_wrap)
+                    elif hasattr(obj, 'fsdp_wrap_fn') and isinstance(obj.fsdp_wrap_fn, Callable):
+                        should_be_wrapped = obj.fsdp_wrap_fn(module)
+
+                    if should_be_wrapped and auto_microbatching:
+                        module.register_forward_hook(sync_hook)
+                        module.register_full_backward_hook(sync_hook)
+                    return should_be_wrapped
+
+                if is_torch_2_0:
```
b-chu, line 479: Make this `elif` instead of using another level of nesting

composer/trainer/dist_strategy.py
mvpatel2000: This is an inner function that is rewrapped based on torch version. I agree it's not great...

composer/trainer/dist_strategy.py
mvpatel2000: It shares the `__auto_wrap_policy` so it is inside

mvpatel2000 (approved):
LGTM

PR #2598, opened 2023-10-02
Free outputs callback

mvpatel2000:
# What does this PR do?

Adds callback to support freeing outputs for memory savings. When not using `train_metrics`, `self.state.outputs` are not needed. However, they may take up a non-trivial amount of memory (`seq_length*vocab_size*microbatch_size*bytes_per_param`). For certain long sequence models, this memory starts to matter (~1-2GB), so having an option to free the memory is useful.

Existing tests (eg `TestCallbackTrains`) should be sufficient for this PR.

image

[GRT-2464](https://databricks.atlassian.net/browse/GRT-2464)


composer/callbacks/free_outputs.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Free train metrics ."""
```
j316chuck, line 4:
```suggestion
"""Free train metrics."""
```

composer/callbacks/free_outputs.py
```diff
@@ -0,0 +1,16 @@
...
+
+import torch
+
+from composer.core import Callback, State
+from composer.loggers import Logger
+
+
+class FreeOutputs(Callback):
+    """Free train metrics on AFTER_LOSS to reduce peak memory usage if not using train metrics."""
```
j316chuck, line 13: QQ: where is the "not using train metrics implemented"?

composer/callbacks/generate.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """Periodically log generations from a set of prompts."""
+
```
j316chuck, line 5: Did you mean to modify this file?

tests/callbacks/callback_settings.py
```diff
@@ -223,5 +224,8 @@ def get_cb_model_and_datasets(cb: Callback,
             )
         return (configure_tiny_gpt2_hf_model(), dummy_gpt_lm_dataloader(size=dl_size),
                 dummy_gpt_lm_dataloader(size=dl_size))
-    return (SimpleModel(), DataLoader(RandomClassificationDataset(size=dl_size), **default_dl_kwargs),
+    model = SimpleModel()
+    if isinstance(cb, FreeOutputs):
+        model.get_metrics = lambda is_train=False: {}
```
b-chu, line 229:
```suggestion
        model.get_metrics = lambda x: {}
``` 
Can it just be this?

b-chu (approved):
Approving, but wait for someone more familiar with callbacks to also approve

composer/callbacks/generate.py
mvpatel2000: Yes to be consistent style wise

composer/callbacks/free_outputs.py
mvpatel2000: User is free to pass them in if they want. https://github.com/mosaicml/llm-foundry/pull/642 here is a foundry PR to add that option in our stack

tests/callbacks/callback_settings.py
mvpatel2000: Needs a default value as there are places in the codebase where `is_train` is not specified. Plus, just nice to match signature

PR #2425, opened 2023-08-11
Fix a bug when batch type is dict and one of the values is the list.

mhh0318:
# What does this PR do?

Fix a bug when batch type is `dict` and one of the values is the `list`.

# What issue(s) does this change relate to?

# Before submitting
- [x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [x] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [x] Did you update any related docs and document your change?
- [ ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [x] Did you run the tests locally to make sure they pass?
- [ ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

PR #2599, opened 2023-10-03
Fix a bug when batch type is dict and one of the values is the list

mvpatel2000:
# What does this PR do?

Extends sample count functionality to include dict of lists. Adds a unit test. 

Extends https://github.com/mosaicml/composer/pull/2425 with a unit test + lint


composer/core/data_spec.py
```diff
@@ -244,7 +244,16 @@ def _default_get_num_samples_in_batch(self, batch: Batch) -> int:
...
         elif isinstance(batch, dict):
-            dim0_sizes = [t.shape[0] for t in batch.values()]
+            for t in batch.values():
+                if isinstance(t, torch.Tensor):
+                    dim0_sizes.append(t.shape[0])
+                elif isinstance(t, list):
+                    dim0_sizes.append(len(t))
+                else:
+                    raise ValueError('Unable to determine the batch size, batch contains'
+                                     f'an element of type {type(t)}, which does not have a'
```
dakinggg, line 254: since the code above checks types, not whether `t` has a shape, can you update the error message to say what it is actually checking (is tensor or list)

dakinggg (approved):

PR #2600, opened 2023-10-03
Refactor save interval and eval interval to share code

dakinggg:
This PR refactors the interval scheduler helper functions used for checkpointing and eval to use the same function. We can now use this unified function anywhere we need an interval scheduler, and only need to add new features in one place. An immediate outcome of this is that we can use `dur` interval scheduling for checkpointing now.


composer/core/evaluator.py
```diff
@@ -195,10 +107,13 @@ def eval_interval(self):
 
     @eval_interval.setter
     def eval_interval(self, eval_interval: Optional[Union[int, str, Time, Callable[[State, Event], bool]]]):
+        from composer.utils import create_interval_scheduler
```
mvpatel2000, line 110: This does not need to be inlined since core depends on utils

tests/trainer/test_trainer_eval.py
```diff
@@ -462,10 +462,10 @@ def test_eval_at_fit_end(eval_interval: Union[str, Time, int], max_duration: str
...
     )
 
-    evaluator.eval_interval = evaluate_periodically(
-        eval_interval=eval_interval,
-        eval_at_fit_end=eval_at_fit_end,
-    )
+    evaluator.eval_interval = create_interval_scheduler(interval=eval_interval,
+                                                        include_end_of_training=eval_at_fit_end,
+                                                        checkpoint_events=False,
+                                                        final_events={Event.FIT_END})
```
mvpatel2000, line 468: can we restore trailing comma so it's cleaner?

mvpatel2000 (commented):
minor nits. mostly lgtm

composer/callbacks/utils.py
```diff
@@ -1,16 +1,19 @@
...
-from composer.core import Event, State, Time, TimeUnit
+from composer.core import Event, State, Time
+from composer.utils.misc import create_interval_scheduler as _create_interval_scheduler
 
 
 def create_interval_scheduler(interval: Union[str, int, Time],
-                              include_end_of_training=True) -> Callable[[State, Event], bool]:
+                              include_end_of_training: bool = True,
+                              checkpoint_events: bool = True,
+                              final_events: Optional[Set[Event]] = None) -> Callable[[State, Event], bool]:
```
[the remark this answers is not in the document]
dakinggg: but i need to get a deprecation warning if anyone actually tries to use it, right?

mvpatel2000 (approved):
Very nice 🎉

PR #2605, opened 2023-10-04
Deprecate many datasets and models

mvpatel2000:
# What does this PR do?

Deprecates many datasets and models which are unused. Long term we will not have models and datasets in Composer.

[GRT-2472](https://databricks.atlassian.net/browse/GRT-2472)


dakinggg (approved):
🚢 🚀

PR #2606, opened 2023-10-04
Switch mosaicml logger to all exceptions

mvpatel2000:
# What does this PR do?

Switch mosaicml logger to all exceptions


composer/loggers/mosaicml_logger.py
```diff
@@ -167,7 +167,7 @@ def format_data_to_json_serializable(data: Any):
 
         # Unknown format catch-all
         return str(data)
-    except RuntimeError as e:
+    except Exception as e:
```
dakinggg, line 170: what error were you getting in her? shouldn't the code in this part be pretty clear about the errors it raises? its just json serialization.

PR #2607, opened 2023-10-04
Call generate callback at end of training

aspfohl:
# What does this PR do?

Switches Generate callback to be included in the end of training

Interface was feeling a little weird if you max duration & generate frequency aligned or were the same

Thoughts? 

# Before submitting


composer/callbacks/generate.py
```diff
@@ -41,7 +41,7 @@ def __init__(self,
         self.prompts = prompts
         self.generate_kwargs = kwargs
         self.batch_size = batch_size if batch_size is not None else len(prompts)
-        self.check_interval = create_interval_scheduler(interval, include_end_of_training=False)
+        self.check_interval = create_interval_scheduler(interval, include_end_of_training=True)
```
dakinggg, line 44: this doesn't quite do what you want. Since this callback gets run on every event, any events that occur after the last batch is trained would end up triggering this. You need to add some extra logic to make sure it doesn't run more than once for the same batch or scope down which events the callback runs on. See the hf_checkpointer in foundry for the former solution.

aspfohl: 👍 Updated! And added tests

tests/callbacks/test_generate.py
```diff
@@ -17,6 +17,7 @@
 from tests.common.models import configure_tiny_gpt2_hf_model
 
 
+# foo
```
dakinggg, line 20: remove?

tests/callbacks/test_generate.py
```diff
@@ -111,3 +112,29 @@ def test_calls(self, device, world_size, use_fsdp):
...
+        trainer = self._create_trainer(device, f'{train_batches}ba', use_fsdp, generate_cb)
+
+        # Mock methods
+        state = trainer.state
+        model = state.model.module if state.is_model_ddp else state.model
+        model.generate = Mock(wraps=model.generate)  # type: ignore
+        generate_cb.generate = Mock(wraps=generate_cb.generate)
+        trainer.logger.log_table = Mock()
+
+        trainer.fit()
```
dakinggg, line 135: can you manually run an extra event here and make sure it doesn't result in an extra call to generate? so like `trainer.engine.run_event(Event.FIT_END)`

dakinggg (approved):

PR #2608, opened 2023-10-05
Change the tokenizer json file to read binary

dakinggg:
# What does this PR do?
There is a bug that I don't entirely know the root cause of. Somehow calling HF generate (which does not have access to the tokenizer) causes checkpoint to fail when trying to json.load the tokenizer.json file, with a unicodedecodeerror. It is extremely difficult to reproduce. It only happens the first time I run it on a new interactive instance, so I don't have a unit test that reproduces it here. However, I manually confirmed that this change fixes the issue when using llama with the generate callback in foundry.

Closes [GRT-2473](https://databricks.atlassian.net/browse/GRT-2473)

# Before submitting


irenedea (approved):
LGTM!

mvpatel2000 (approved):
It's binary file?

dakinggg:
The file in question i believe is (its written from the rust tokenizers code so im not 100% sure, but there is a call to save_bytes). There are other json files that are utf-8 encoded, like the config. I don't think reading those in as binary is a problem (tested loading a normal json file as binary and it worked fine), since json.load supports this and always outputs strings, but lmk if you would like more evidence or disagree.

PR #2612, opened 2023-10-05
Clean up gpu tests

mvpatel2000:
# What does this PR do?

Clean up GPU tests. R
- Removes FFCV example notebook
- Removes FFCV GPU tests (and all vision GPU tests)
- Upgrades things to torch 2.1


.github/workflows/pr-gpu.yaml
```diff
@@ -12,14 +12,9 @@ jobs:
...
         include:
-          - name: 'gpu-3.10-1.13-vision-doctest'
-            container: mosaicml/pytorch_vision:1.13.1_cu117-python3.10-ubuntu20.04
-            markers: 'not daily and not remote and gpu and (vision or not vision) and (doctest or not doctest)'
-            pytest_command: 'coverage run -m pytest'
-            composer_package_name: 'mosaicml'
-          - name: 'gpu-3.10-2.0-vision-doctest'
-            container: mosaicml/pytorch_vision:2.0.1_cu117-python3.10-ubuntu20.04
-            markers: 'not daily and not remote and gpu and (vision or not vision) and (doctest or not doctest)'
+          - name: 'gpu-3.10-2.1'
```
j316chuck, line 15: are we not testing 2.0 and 1.3 images on torch anymore? I guess no need since we are migrating to training with more recent torch versions?

j316chuck (approved):
🚢 🚀

.github/workflows/pr-gpu.yaml
dakinggg: +1 we shouldn't drop 1.13 and 2.0 tests yet, right?

mvpatel2000: This is on the PR tests. The daily tests will still test all versions.

.github/workflows/daily.yaml
```diff
@@ -53,16 +53,16 @@ jobs:
             markers: 'daily and (remote or not remote) and not gpu and not vision and not doctest'
             pytest_command: 'coverage run -m pytest'
             composer_package_name: 'mosaicml'
-          - name: 'daily-cpu-3.10-1.13-composer'
```
dakinggg, line 56: keep?

dakinggg (approved):

PR #2627, opened 2023-10-09
Bump version to 0.16.4

mvpatel2000:
# What does this PR do?

Bump version to 0.16.4


.github/workflows/pytest-cpu.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
           name: coverage-${{ github.sha }}-${{ inputs.name }}
           path: .coverage
       - name: Notify slack fail
-        if: failure() && !cancelled() && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') && (github.event_name != 'pull_request' || github.event_name != 'pull_request_target')
+        if: failure() && !cancelled() && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') && (github.event_name != 'pull_request' && github.event_name != 'pull_request_target')
```
dakinggg, line 92: was this intentional? i don't think the event name can be both pull_request and pull_request_target

.github/workflows/pytest-gpu.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
 
           python .github/mcli/mcli_pytest.py --image '${{ inputs.container }}' --pip_package_name '${{ inputs.composer_package_name }}' --pytest_markers '${{ inputs.pytest-markers }}' --pytest_command '${{ inputs.pytest-command }}' --timeout ${{ inputs.mcloud-timeout }} ${REF_ARGS}
       - name: Notify slack fail
-        if: failure() && !cancelled() && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') && (github.event_name != 'pull_request' || github.event_name != 'pull_request_target')
+        if: failure() && !cancelled() && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') && (github.event_name != 'pull_request' && github.event_name != 'pull_request_target')
```
dakinggg, line 88: same

pyproject.toml
```diff
@@ -149,6 +149,8 @@ filterwarnings = [
     '''ignore:'has_mkldnn' is deprecated, please use 'torch.backends.mkldnn.is_available():UserWarning''',
     # Ignore torch distributed deprecated warnings
     '''ignore:torch.distributed.reduce_op is deprecated, please use torch.distributed.ReduceOp instead:UserWarning''',
+    # Ignore torch sharded tensor deprecated warnings
```
dakinggg, line 152: Is this something we need to take a jira to fix? or is it coming from pytorch internal usage?

tests/cli/__init__.py
```diff
@@ -1,2 +0,0 @@
-# Copyright 2022 MosaicML Composer authors
```
dakinggg, line 1: Unintentional?

tests/test_cli.py
```diff
@@ -12,9 +11,9 @@
 
 @pytest.mark.parametrize('args', [
     ['composer', '--version'],
-    [sys.executable, '-m', 'composer', '--version'],
-    [sys.executable, '-m', 'composer.cli', '--version'],
-    [sys.executable, '-m', 'composer.cli.launcher', '--version'],
+    ['python', '-m', 'composer', '--version'],
```
dakinggg, line 14: unintentional?

dakinggg (approved):
Approved so you don't need review again, but please resolve/respond to all the comments

pyproject.toml
mvpatel2000: internal pytorch usage

tests/cli/__init__.py
mvpatel2000: moved cli test to top level since folder has one file to be consistent

tests/test_cli.py
mvpatel2000: clean up from chuck -- python is better than sys executable imo

.github/workflows/pytest-cpu.yaml
mvpatel2000: This was buggy before -- it should only run if its not pull request and not pull request target (otherwise, it always evaluates true)

.github/workflows/pytest-gpu.yaml
mvpatel2000: This was buggy before -- it should only run if its not pull request and not pull request target (otherwise, it always evaluates true)

PR #2628, opened 2023-10-10
Patch default precision

mvpatel2000:
# What does this PR do?

Fixes default precision to use fp32 reduce dtype

[GRT-2491](https://databricks.atlassian.net/browse/GRT-2491)

### DIFF
--- a/composer/trainer/mosaic_fsdp_utils.py
+++ b/composer/trainer/mosaic_fsdp_utils.py
@@ -86,6 +86,7 @@ def get_mixed_precision(precision, mixed_precision='DEFAULT', keep_low_precision
             pass
         elif mixed_precision == 'DEFAULT':
             param_dtype = _get_torch_dtype(precision)
+            reduce_dtype = torch.float32
             buffer_dtype = _get_torch_dtype(precision)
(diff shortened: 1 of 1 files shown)


composer/trainer/mosaic_fsdp_utils.py
```diff
@@ -86,6 +86,7 @@ def get_mixed_precision(precision, mixed_precision='DEFAULT', keep_low_precision
             pass
         elif mixed_precision == 'DEFAULT':
             param_dtype = _get_torch_dtype(precision)
+            reduce_dtype = torch.float32
```
vchiley, line 89:
```suggestion
            reduce_dtype = torch.get_default_dtype()
```

This might be more correct:
doc: https://pytorch.org/docs/stable/generated/torch.get_default_dtype.html

vchiley (changes requested):

composer/trainer/mosaic_fsdp_utils.py
vchiley: granted this warrants changing docs...

vchiley (approved):
👍

PR #2633, opened 2023-10-11
Adding Model Data Init and Training Progress to MosaicMLLogger

jjanezhang:
# Adding Model Data Init and Training Progress to MosaicMLLogger

1. Logs timestamp when model data is downloaded and initialized 
2. Logs training progress metrics whenever we log to metadata

Testing
Started a run and confirmed that training metrics and data initialized are in run's metadata
image


dakinggg (commented):
curious what @mvpatel2000 and @eracah think, but do we need a new callback for this? That is, do we actually want to log this to all logger destinations? or should we just have the `MosaicMLLogger` do this?

mvpatel2000:
I think it's fine to log to one or all logger destinations -- the others just won't really care since and the UX is clean on most of them. I've done one offs like this using the logger API and not had issues.

I'm not sure this belongs in Composer though. The callbacks in Composer should be foundational callbacks which are broadly applicable. This feels like a callback for a bespoke use case which probably better belongs in whatever repo is leveraging it? @jjanezhang feel free to start a conversation in the composer channel if you feel it's appropriate

composer/loggers/mosaicml_logger.py
```diff
@@ -88,13 +90,30 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
         # Log WandB run URL if it exists. Must run on after_load as WandB is setup on event init
         for callback in state.callbacks:
             if isinstance(callback, WandBLogger):
                 run_url = callback.run_url
                 if run_url is not None:
                     self._log_metadata({'wandb/run_url': run_url})
 
+    def fit_start(self, state: State, logger: Logger) -> None:
+        # Log max duration in either epochs or tokens for run events
+        if state.max_duration.unit == TimeUnit.TOKEN:
```
dakinggg, line 104: Can we just log `max_duration_value` and `max_duration_units`?

composer/loggers/mosaicml_logger.py
```diff
@@ -88,13 +90,30 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+            self._log_metadata({'total_num_tokens': state.max_duration.value})
+        else:
+            num_batches_per_epoch = math.ceil(len(state.train_dataloader.dataset) / state.train_dataloader.batch_size)
```
dakinggg, line 107: This requires more try/catching if left this way. datasets are not guaranteed to have length, and dataloaders are not guaranteed to have batch size. I suggest doing the above suggestion, does that still work for your event needs?

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +194,40 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
         assert logger_count == 0
+
+
+def test_run_events_logged(monkeypatch, tiny_bert_tokenizer):
+    mock_mapi = MockMAPI()
+    monkeypatch.setattr(mcli, 'update_run_metadata', mock_mapi.update_run_metadata)
+    run_name1 = 'test-run-name1'
+    monkeypatch.setenv('RUN_NAME', run_name1)
+    trainer = Trainer(model=SimpleModel(),
+                      train_dataloader=DataLoader(RandomClassificationDataset()),
```
dakinggg, line 205: set `train_subset_num_batches=1`

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +194,40 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+                      max_duration='1ep',
+                      loggers=[MosaicMLLogger()])
+    trainer.fit()
+    assert isinstance(mock_mapi.run_metadata[run_name1]['mosaicml/model_initialized_time'], float)
+    assert mock_mapi.run_metadata[run_name1]['mosaicml/total_num_epochs'] == 1
+    assert mock_mapi.run_metadata[run_name1]['mosaicml/num_batches_per_epoch'] == 100
+    assert 'total_num_tokens' not in mock_mapi.run_metadata[run_name1]
+
+    run_name2 = 'test-run-name2'
```
dakinggg, line 214: I think this should be a separate unit test

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +194,40 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+    monkeypatch.setenv('RUN_NAME', run_name2)
+    transformers = pytest.importorskip('transformers')
```
dakinggg, line 216: Move the importorskip to the top of the test

composer/loggers/mosaicml_logger.py
jjanezhang: We're planning to log [epoch x/xx] [batch x/xx] if the user submits training duration in ep or ba, so I think we want to calculate num batches per epoch in addition to the max duration.

composer/loggers/mosaicml_logger.py
jjanezhang: Hmm I see what you mean, do you have an idea how often the dataset length and batch size are missing? It would be great to report [epoch x/xx] [batch x/xx] instead of just [epoch x/xx] but if this can't be reported often, then I think we should standardize to max duration and unit.

composer/loggers/mosaicml_logger.py
dakinggg: I think you can draw inspiration from https://github.com/mosaicml/composer/blob/5dff73082ec5f7dbfe2c1ee7c711b6acf3207742/composer/loggers/console_logger.py#L185-L206. but basically you'll need to account for the different possibilities and do different things. for the finetuning api everything will always be defined, so its just to make this logger robust.

mvpatel2000: You can look at https://github.com/mosaicml/composer/blob/dev/composer/callbacks/runtime_estimator.py#L99-L104 for inspiration

This function computes elapsed duration more robustly -- you can rip out parts of it https://github.com/mosaicml/composer/blob/dev/composer/callbacks/runtime_estimator.py#L90

composer/loggers/mosaicml_logger.py
```diff
@@ -80,21 +83,69 @@ def __init__(
...
         # Log WandB run URL if it exists. Must run on after_load as WandB is setup on event init
         for callback in state.callbacks:
             if isinstance(callback, WandBLogger):
                 run_url = callback.run_url
                 if run_url is not None:
                     self._log_metadata({'wandb/run_url': run_url})
 
+    def fit_start(self, state: State, logger: Logger) -> None:
+        # adds state to calculate training progress on log metadata
+        self._state = state
```
irenedea, line 105: instead of using `self._state` to track `state` we should just pass it along on L95.

composer/loggers/mosaicml_logger.py
```diff
@@ -80,21 +83,69 @@ def __init__(
...
+
+        If user submits max duration:
+        - in tokens -> format: [token=x/xx]
+        - in batches -> format: [batch=x/xx]
+        - in epoch -> format: [epoch=x/xx] [batch=x/xx] (where batch refers to batches completed in current epoch)
+        If batches per epoch cannot be calculated, return [epoch=x/xx]
+
+        If no training duration given -> format: ''
+        """
+        if state.max_duration is None:
```
irenedea, line 118: in what case will max_duration will be None?

irenedea: how come we return `{'training_progress': ''}` if there is no max_duration, but we return an empty dictionary (instantiated on L132) if the unit is not token, epoch, or batch? would we want these to be the same?

composer/loggers/mosaicml_logger.py
```diff
@@ -72,6 +74,7 @@ def __init__(
             self.time_last_logged = 0
             self.time_failed_count_adjusted = 0
             self.buffered_metadata: Dict[str, Any] = {}
+            self._state = None
```
irenedea, line 77: see comment below on removing `self._state`

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +196,88 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+    monkeypatch.setenv('RUN_NAME', run_name)
+    # Because batch timestamp resets to 0 at the end of the train, we use in memory logger to check an intermediate epoch timestamp
+    in_memory_logger = InMemoryLogger()
+    trainer = Trainer(model=SimpleModel(),
+                      train_dataloader=DataLoader(RandomClassificationDataset(), batch_size=2),
+                      train_subset_num_batches=2,
+                      max_duration='2ep',
+                      loggers=[MosaicMLLogger(), in_memory_logger])
+    trainer.fit()
+    assert 'mosaicml/training_progress_unit' in mock_mapi.run_metadata[run_name]
```
irenedea, line 273: nit: use a variable for mock_mapi.run_metadata[run_name] because it's referenced multiple times

composer/loggers/mosaicml_logger.py
```diff
@@ -80,21 +83,69 @@ def __init__(
...
+
+        If no training duration given -> format: ''
+        """
+        if state.max_duration is None:
+            return {'training_progress': ''}
+
+        if state.max_duration.unit == TimeUnit.TOKEN:
+            return {
+                'training_progress': f'[token={state.timestamp.token.value}/{state.max_duration.value}]',
+                'training_progress_unit': TimeUnit.TOKEN
```
irenedea, line 124: is it necessary to save the units as well?

irenedea (commented):
Nice! Had some comments and questions

composer/loggers/mosaicml_logger.py
```diff
@@ -80,21 +83,69 @@ def __init__(
...
+            return {
+                'training_progress': f'[batch={state.timestamp.batch.value}/{state.max_duration.value}]',
+                'training_progress_unit': TimeUnit.BATCH
+            }
+        training_progress_metrics = {}
+        if state.max_duration.unit == TimeUnit.EPOCH:
+            batches_per_epoch = None
+            if state.timestamp.epoch.value >= 1:
+                batches_per_epoch = math.ceil(
+                    (state.timestamp.batch - state.timestamp.batch_in_epoch).value // state.timestamp.epoch.value)
```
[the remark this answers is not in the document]
irenedea: actually maybe you don't even need the `int` cast in my suggestion as per the code here: https://github.com/mosaicml/composer/blob/dev/composer/callbacks/runtime_estimator.py#L99-L104

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +196,88 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+                                                    use_keys=True)
+    collator = transformers.DataCollatorForLanguageModeling(tokenizer=tiny_bert_tokenizer, mlm_probability=0.15)
+    dataloader = DataLoader(pretraining_train_dataset,
+                            batch_size=4,
+                            sampler=dist.get_sampler(pretraining_train_dataset),
+                            collate_fn=collator)
+    trainer = Trainer(model=SimpleTransformerMaskedLM(vocab_size=tiny_bert_tokenizer.vocab_size),
+                      train_dataloader=dataloader,
+                      max_duration='64tok',
+                      loggers=[MosaicMLLogger()])
```
irenedea, line 233: why is the Trainer/model instantiation different from the other tests for batch and epoch? Can we make it simple like those ones?

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +196,88 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
     # Otherwise, no logger
     else:
         assert logger_count == 0
+
```
irenedea, line 199: for unit tests, instead of testing with a full `trainer.fit()` for each thing, I think we should test 1 full `trainer.fit()` on init and a specific unit and then for each unit, we can test `_get_training_progress_metrics` separately. This way of testing would be more modular and you can much more easily test the training sub progress metric this way. 

You could use magic mock for mocking the state passed to _get_training_progress_metrics:
```
def test_get_training_progress_metrics():
     # test tokens
     logger = MosaicMLLogger()
     state = MagicMock()
     state.max_duration.unit = TimeUnit.TOKEN
     state.timestamp.token.value = ...
     # set other magic mock attributes
     result = logger._get_training_progress_metrics(state)
     # assert stuff

```

composer/loggers/mosaicml_logger.py
jjanezhang: I'm pretty confident max duration will default to 1 epoch for finetuning, but not sure for other runs so put a catch to make sure it doesn't error

composer/loggers/mosaicml_logger.py
jjanezhang: going to remove these

composer/loggers/mosaicml_logger.py
mvpatel2000: Note that max duration has to be set at this point -- it is type Union None as it does not need to be set on `Trainer.__init__` and can instead be set on `Trainer.fit()`. 

You could just assert it is not None here for type checker as this is a valid logical assumption (see L1777 Trainer for validation logic)

composer/loggers/mosaicml_logger.py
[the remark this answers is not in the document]
mvpatel2000: All of these should be cleanly dividing ints so yep agree with Irene

composer/loggers/mosaicml_logger.py
```diff
@@ -80,21 +83,69 @@ def __init__(
...
+            elif state.dataloader_len is not None:
+                batches_per_epoch = state.dataloader_len.value
+            training_progress_metrics = {
+                'training_progress': f'[epoch={state.timestamp.epoch.value}/{state.max_duration.value}]',
+                'training_progress_unit': TimeUnit.EPOCH
```
mvpatel2000, line 142: This assumes max_duration is in epochs which is not necessarily the case

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +196,93 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+    trainer = Trainer(model=SimpleModel(),
+                      train_dataloader=DataLoader(RandomClassificationDataset()),
+                      train_subset_num_batches=1,
+                      max_duration='1ep',
+                      loggers=[MosaicMLLogger()])
+    trainer.fit()
+    assert isinstance(mock_mapi.run_metadata[run_name]['mosaicml/model_initialized_time'], float)
+
+
+def test_progress_logged(monkeypatch, tiny_bert_tokenizer):
```
irenedea, line 215: nit: combine this test with the one above

tests/loggers/test_mosaicml_logger.py
```diff
@@ -194,3 +196,93 @@ def test_auto_add_logger(monkeypatch, platform_env_var, access_token_env_var, lo
...
+                      max_duration='4ba',
+                      loggers=[MosaicMLLogger()])
+    trainer.fit()
+    metadata = mock_mapi.run_metadata[run_name]
+    assert 'mosaicml/training_progress' in metadata
+    assert metadata['mosaicml/training_progress'] == '[batch=4/4]'
+    assert 'mosaicml/training_sub_progress' not in metadata
+
+
+def test_token_training_progress_logged():
```
irenedea, line 232: nit: i'd just rename these tests to not end with `logged` because we aren't actually logging anything

maybe to `test_token_training_progress_metrics`

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +88,58 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+            if cur_epoch < state.max_duration.value:
+                cur_epoch += 1
+            training_progress_metrics = {
+                'training_progress': f'[epoch={cur_epoch}/{state.max_duration.value}]',
+            }
+            training_progress_metrics['training_sub_progress'] = curr_progress
+        return training_progress_metrics
+
     def batch_end(self, state: State, logger: Logger) -> None:
+        self._log_metadata(self._get_training_progress_metrics(state))
```
irenedea, line 142: @mvpatel2000 Just want to confirm that logging the training progress metrics at batch_end only is sufficient?

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +88,58 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+            cur_batch = int(state.timestamp.batch_in_epoch)
+            cur_epoch = int(state.timestamp.epoch)
+            if int(state.timestamp.epoch) >= 1:
+                batches_per_epoch = int(
+                    (state.timestamp.batch - state.timestamp.batch_in_epoch).value / state.timestamp.epoch.value)
+                curr_progress = f'[batch={cur_batch}/{batches_per_epoch}]'
+            elif state.dataloader_len is None:
+                curr_progress = f'[batch={cur_batch}]'
+            else:
+                total = int(state.dataloader_len)
```
irenedea, line 131: is it possible to remove all the int casts here?

irenedea (commented):
Looks good, awesome work! Just some nits and wanted to double check that moving to batch_end is the right call 😄

composer/loggers/mosaicml_logger.py
irenedea: https://github.com/mosaicml/composer/pull/2633#discussion_r1369374456 "All of these should be cleanly dividing ints"

composer/loggers/mosaicml_logger.py
mvpatel2000: Yep! since the logger flushes every minute, it doesn't matter a ton where its logged. and batch_end is definitely frequent enough

irenedea: Awesome, thanks!

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +88,58 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+        # Log model data downloaded and initialized for run events
+        self._log_metadata({'model_initialized_time': time.time()})
         # Log WandB run URL if it exists. Must run on after_load as WandB is setup on event init
         for callback in state.callbacks:
             if isinstance(callback, WandBLogger):
                 run_url = callback.run_url
                 if run_url is not None:
                     self._log_metadata({'wandb/run_url': run_url})
 
+    def _get_training_progress_metrics(self, state: State) -> Dict[str, Any]:
```
irenedea, line 100: nit: Move this function outside the class because it doesn't rely on `self` at all 😃

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +88,58 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+        training_progress_metrics = {}
+        if state.max_duration.unit == TimeUnit.EPOCH:
+            cur_batch = int(state.timestamp.batch_in_epoch)
+            cur_epoch = int(state.timestamp.epoch)
+            if int(state.timestamp.epoch) >= 1:
+                batches_per_epoch = int(
+                    (state.timestamp.batch - state.timestamp.batch_in_epoch).value / state.timestamp.epoch.value)
+                curr_progress = f'[batch={cur_batch}/{batches_per_epoch}]'
+            elif state.dataloader_len is None:
+                curr_progress = f'[batch={cur_batch}]'
```
mvpatel2000, line 129: You should copy the logic from runtime estimator and cache dataloader_len as it will change if eval runs

composer/loggers/mosaicml_logger.py
mvpatel2000: this does need `self` since dataloader len should be cached (see logic from runtime estimator). missed in review but this comment reminded me :)

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +89,61 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+
+        If user submits max duration:
+        - in tokens -> format: [token=x/xx]
+        - in batches -> format: [batch=x/xx]
+        - in epoch -> format: [epoch=x/xx] [batch=x/xx] (where batch refers to batches completed in current epoch)
+        If batches per epoch cannot be calculated, return [epoch=x/xx]
+
+        If no training duration given -> format: ''
+        """
+        assert state.max_duration is not None
```
dakinggg, line 112: I think this function should short circuit and just return empty dict if `not self._enabled`?

dakinggg (approved):

mvpatel2000 (changes requested):
Requesting sign off from me for progress logic before this lands. Can review fast whenever Jane requests :)

mvpatel2000:
@jjanezhang  will review when tests pass :)

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +89,64 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+            }
+            training_progress_metrics['training_sub_progress'] = curr_progress
+        return training_progress_metrics
+
+    def batch_start(self, state: State, logger: Logger) -> None:
+        if state.dataloader_len is not None and self._enabled:
+            self.train_dataloader_len = state.dataloader_len.value
+
     def batch_end(self, state: State, logger: Logger) -> None:
+        self._log_metadata(self._get_training_progress_metrics(state))
```
mvpatel2000, line 149: do we need to also do at `fit_end`?

composer/loggers/mosaicml_logger.py
```diff
@@ -88,14 +89,64 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
+            elif self.train_dataloader_len is not None:
+                curr_progress = f'[batch={cur_batch}/{self.train_dataloader_len}]'
+            else:
+                curr_progress = f'[batch={cur_batch}]'
+            if cur_epoch < state.max_duration.value:
+                cur_epoch += 1
+            training_progress_metrics = {
+                'training_progress': f'[epoch={cur_epoch}/{state.max_duration.value}]',
+            }
+            training_progress_metrics['training_sub_progress'] = curr_progress
```
mvpatel2000, line 141:
```suggestion
            training_progress_metrics = {
                'training_sub_progress': curr_progress,
                'training_progress': f'[epoch={cur_epoch}/{state.max_duration.value}]',
            }
```

mvpatel2000 (approved):
LGTM! Sorry this took a bit to review

PR #2652, opened 2023-10-17
Refactor UCI Profiler

j316chuck:
# What does this PR do?
Refactor UCI Profiler

None


mvpatel2000 (changes requested):
Can you please detail why this is necessary and what happens in detail?

PR #2656, opened 2023-10-18
Gate HSDP by torch 2.1.0

mvpatel2000:
# What does this PR do?

Gate torch import by 2.1.0. We are not confident checkpointing works on earlier versions per @eracah

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -15,14 +15,16 @@ import tempfile
 import textwrap
 import warnings
 from pathlib import Path
+from packaging import version
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
-from packaging import version
 
-from composer.utils import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, ObjectStore, dist, format_name_with_dist,
-                            format_name_with_dist_and_time, get_file, is_model_deepspeed, is_tar, reproducibility,
-                            using_torch_2)
+from composer.utils import dist, reproducibility
+from composer.utils.file_helpers import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, format_name_with_dist,
+                                         format_name_with_dist_and_time, get_file, is_tar)
+from composer.utils.misc import is_model_deepspeed, using_torch_2
+from composer.utils.object_store import ObjectStore
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


b-chu (commented):
Do we know that it breaks on 2.0.1 and works on 2.1.0 for certain?

mvpatel2000:
> Do we know that it breaks on 2.0.1 and works on 2.1.0 for certain?

We know for sure it works on 2.1.0. Community users reported issues with 2.0.1

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -556,7 +556,8 @@ def test_fsdp_partitioned_state_dict_load(world_size, tmp_path: pathlib.Path, st
 @pytest.mark.parametrize('autoresume', [False, True])  # True commented out for now
 @pytest.mark.parametrize('num_shards', [2, 4, 7])
 @pytest.mark.parametrize('sharding_strategy', ['FULL_SHARD', 'SHARD_GRAD_OP'])
-@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.1'), reason='requires PyTorch 2.01 or higher')
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.0.1'),
```
j316chuck, line 559: should this be <=?

composer/utils/checkpoint.py
```diff
@@ -313,7 +313,7 @@ def load_sharded_checkpoint(
             f'Sharded checkpoint loading requires torch version >= 2.0.0. You have torch version {torch.__version__}')
 
     using_multinode = dist.get_world_size() != dist.get_local_world_size()
-    if not using_torch_2_0_1() and using_multinode:
+    if not version.parse(torch.__version__) >= version.parse('2.0.1') and using_multinode:
```
j316chuck, line 316: should this be 2.1.0?

j316chuck (approved):
should we pin the version for 2.1.0?

tests/trainer/test_fsdp_checkpoint.py
mvpatel2000: Nope, it is for 2.0. I just fixed a typo

composer/utils/checkpoint.py
mvpatel2000: nope, just removing utility fn and only doing for 2.0 to be consistent

PR #2660, opened 2023-10-20
Fix FSDP arg default to match torch

mvpatel2000:
# What does this PR do?

Fix FSDP arg default to match torch for limit all gathers

image


j316chuck (approved):
Thanks!

PR #2665, opened 2023-10-24
Torch 2.1 Resumption Support

mvpatel2000:
# What does this PR do?

As of torch 2.1, torch has changed how RNG states work which is breaking our tests. This PR patches tests.

In order to address this issue, we catch RNG load issues and emit a warning. As this generally does not affect resumption too much, we choose to emit a warning instead of raising an error

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -823,6 +823,7 @@ class State(Serializable):
         """
         metadata_dict = {}
         metadata_dict['composer_env_info'] = get_composer_env_dict()
+        metadata_dict['torch_version'] = torch.__version__
         metadata_dict['device'] = self.device.name
         metadata_dict['precision'] = self.precision.value
         metadata_dict['world_size'] = dist.get_world_size()
--- a/composer/utils/reproducibility.py
+++ b/composer/utils/reproducibility.py
@@ -215,7 +215,16 @@ def load_rng_state(rng_state_dicts: List[Dict[str, Any]]):
         log.debug('Restoring the RNG state')
 
         if is_cuda_available and has_cuda_rng_state:
-            torch.cuda.set_rng_state(rng_state_dict['cuda'])
+            try:
+                torch.cuda.set_rng_state(rng_state_dict['cuda'])
+            except RuntimeError as e:
+                if 'RNG state is wrong size' in str(e):
+                    raise RuntimeError('The RNG state could not be loaded from the checkpoint, '
+                                       'likely because a different version of torch was used to '
+                                       'save the checkpoint. Please use the same version of torch '
+                                       "or specify `load_ignore_keys=['rng'] when constructing "
+                                       'Trainer.') from e
+                raise e
 
         if is_cuda_available and not has_cuda_rng_state:
             warnings.warn(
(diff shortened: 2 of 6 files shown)


composer/core/state.py
```diff
@@ -823,6 +823,7 @@ def _get_state_metadata(self) -> Dict[str, Any]:
         """
         metadata_dict = {}
         metadata_dict['composer_env_info'] = get_composer_env_dict()
+        metadata_dict['torch_version'] = torch.__version__
```
j316chuck, line 826: can we add a test for this?

j316chuck (commented):
LGTM. I think we should add some tests for loading rng state of different pytorch versions, but up to you if you think it's worth the eng time/ease of implementation to add these test.

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -338,6 +338,7 @@ def test_fsdp_mixed_with_sync(world_size, tmp_path: pathlib.Path, sync_module_st
                      r'ignore:MosaicMLLogger is not in the state_dict. Its state will not be restored.:UserWarning'))
 ])
 @pytest.mark.filterwarnings(r'ignore:.*metrics are not saved with sharded state dict.*:UserWarning')
+@pytest.mark.filterwarnings(r'ignore:.*The CUDA RNG state could not be loaded.*:UserWarning')
```
j316chuck, line 341: curious why need to filter this warning since this should only show up if we load and save checkpoints of different versions?

composer/utils/reproducibility.py
```diff
@@ -215,7 +215,15 @@ def load_rng_state(rng_state_dicts: List[Dict[str, Any]]):
...
-            torch.cuda.set_rng_state(rng_state_dict['cuda'])
+            try:
+                torch.cuda.set_rng_state(rng_state_dict['cuda'])
+            except RuntimeError as e:
+                if 'RNG state is wrong size' in str(e):
+                    warnings.warn('The CUDA RNG state could not be loaded from the checkpoint, '
+                                  'likely because a different version of torch was used to save the '
+                                  'checkpoint. Skipping loading the CUDA RNG state.')
+                else:
+                    raise e
```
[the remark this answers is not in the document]
mvpatel2000: The daily test covers this! The test catched it and now I'm fixing it

tests/trainer/test_fsdp_checkpoint.py
mvpatel2000: Yep, this test does that. It saves on 2.0.1 iirc and loads on 2.1

j316chuck (approved):

composer/utils/reproducibility.py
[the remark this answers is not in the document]
b-chu: There should probably be a test that explicity expects this behavior for different versions (xfail) and not for the same version instead of filtering it as a warning

composer/core/state.py
b-chu, line 826: Not sure if this is the best interface, especially if we're considering exposing other package versions in the future

PR #2667, opened 2023-10-24
Add script to validate remote object store paths

irenedea:
# What does this PR do?

Adds a script that validates remote object store paths using composer utility functions. 

This is helpful for debugging especially on remote runs. 

# Before submitting

### DIFF
--- a/composer/utils/file_helpers.py
+++ b/composer/utils/file_helpers.py
@@ -400,6 +400,28 @@ def maybe_create_remote_uploader_downloader_from_uri(
                                   'one of the supported RemoteUploaderDownloader object stores')
 
 
+def list_remote_objects(remote_path: str) -> List[str]:
+    """List objects at the remote path.
+
+    If the path is valid, returns the list of objects at the path.
+    Otherwise, raises an error.
+
+    Args:
+        remote_path (str): Remote object store path.
+
+    Returns:
+        A list of objects at the remote path.
+    """
+    object_store = maybe_create_object_store_from_uri(remote_path)
+    if object_store is None:
+        raise ValueError(f'Failed to create object store. The given path {remote_path} is a local path.')
+    _, _, prefix = parse_uri(remote_path)
+    objects = object_store.list_objects(prefix)
+    if len(objects) == 0:
+        raise ValueError(f'No objects at path {remote_path} found. Please check your path and your access credentials.')
+    return objects
+
+
 def get_file(path: str,
              destination: str,
              object_store: Optional[Union[ObjectStore, LoggerDestination]] = None,
--- a/composer/utils/file_helpers.py
+++ b/composer/utils/file_helpers.py
@@ -403,9 +403,6 @@ def maybe_create_remote_uploader_downloader_from_uri(
 def list_remote_objects(remote_path: str) -> List[str]:
     """List objects at the remote path.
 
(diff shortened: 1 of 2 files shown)


j316chuck:
qq: Is this script supposed to be used locally?

j316chuck (approved):

irenedea:
@j316chuck the mcli yaml that i was using only supports s3, so i figured this would be an easy way to check all backends easily and through composer as well. Though, if this script fails, it doesn't necessarily rule out composer issues. I'll write an mcli yaml for this as well.

dakinggg (changes requested):
Can we make this part of the library instead of a script? similar to `composer_collect_env`

irenedea:
@dakinggg good suggestion! so it would both in the library as well as a script, yeah?

dakinggg:
Yeah, exact same as composer_collect_env setup

irenedea:
@dakinggg I think the best move would be to create a `list_remote_objects` (that does exactly what the script does now, but returns a list of objects if objects are found). This would probably belong in `file_helpers.py`. If we add this function to `file_helpers.py`, do you have any thoughts on where to put the script?

dakinggg:
Can you define `console_scripts` that don't use `main`? It kinda looks like you can, but I've never tried it. If so, can just define a function in `file_helpers.py` that prints out what you want, and define the `console_scripts` command to point to that function

irenedea:
@dakinggg Tried it, but it won't work--it needs to a be a function that takes no arguments. So we could (1) keep a tiny script in `scripts/validate_remote_path.py` and have it call the `list_remote_objects` function in `file_helpers.py` (2) create a small function that takes no args that is defined in `file_helpers.py` that acts as  a console script

dakinggg:
I think Its possible to set it up with args parsed by argparse. because the `composer` launcher is set up the same way. See the main function in there that parses args

setup.py
```diff
@@ -264,8 +264,8 @@ def package_files(prefix: str, directory: str, extension: str):
       install_requires=install_requires,
       entry_points={
           'console_scripts': [
-              'composer = composer.cli.launcher:main',
-              'composer_collect_env = composer.utils.collect_env:main',
+              'composer = composer.cli.launcher:main', 'composer_collect_env = composer.utils.collect_env:main',
+              'composer_validate_remote_path = composer.utils.file_helpers:validate_remote_path'
```
irenedea, line 268: validated that this works
```
composer_validate_remote_path oci://<path>
```

dakinggg, line 268: add back trailing comma so this is formatted nicely

composer/utils/file_helpers.py
```diff
@@ -400,6 +400,44 @@ def maybe_create_remote_uploader_downloader_from_uri(
...
+    _, _, prefix = parse_uri(remote_path)
+    objects = object_store.list_objects(prefix)
+    return objects
+
+
+def validate_remote_path():
+    """Entry point to composer_validate_remote_path cli command.
+
+    Validates a remote path.
+    If a the remote path is valid, prints a list of objects at the path.
```
dakinggg, line 424:
```suggestion
    If the remote path is valid, prints a list of objects at the path.
```

dakinggg (approved):

PR #2679, opened 2023-10-30
Secure Code Eval changes

mvpatel2000:
# What does this PR do?

1. Raise error if unauthorized from MAPI
2. Do not default an implementation and instead require user to explicitly set eval method

### DIFF
--- a/composer/metrics/nlp.py
+++ b/composer/metrics/nlp.py
@@ -528,7 +528,6 @@ class InContextLearningCodeEvalAccuracy(InContextLearningMetric):
         self.add_state('correct', default=torch.tensor(0.), dist_reduce_fx='sum')
         self.add_state('total', default=torch.tensor(0.), dist_reduce_fx='sum')
 
-        self.eval_device = 'LAMBDA'
         if not 'CODE_EVAL_DEVICE' in os.environ:
             raise ValueError(
                 'Attempting to use InContextLearningCodeEvalAccuracy but environment '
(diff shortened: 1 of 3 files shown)


dakinggg (approved):

dakinggg:
@mvpatel2000 I approved, but can you add unit tests for these error cases please?

composer/metrics/nlp.py
```diff
@@ -530,14 +530,12 @@ def __init__(self, dist_sync_on_step: bool = False):
...
-                log.info(f"'CODE_EVAL_DEVICE' env var was not set, so defaulting to 'LAMBDA' as eval device")
-                os.environ['CODE_EVAL_DEVICE'] = 'LAMBDA'
-        else:
-            self.eval_device = os.environ['CODE_EVAL_DEVICE'].upper()
+            raise ValueError(
+                'Attempting to use InContextLearningCodeEvalAccuracy but environment '
+                'variable `CODE_EVAL_DEVICE` is not set. Please set it to `CODE_EVAL_DEVICE` '
+                'to one of `LOCAL` (for unsafe local eval), `LAMBDA` (for AWS lambda ',
+                'evaluation), or `MOSAICML` (for lambda eval through MAPI).')
+        self.eval_device = os.environ['CODE_EVAL_DEVICE'].upper()
```
aspfohl, line 538: Shouldn't we `CODE_EVAL_DEVICE` is set to a valid value?

composer/utils/eval_client/mosaicml_lambda_eval_client.py
```diff
@@ -53,6 +53,9 @@ def invoke(self, payload: List[List[List[Dict[str, str]]]]) -> List[List[List[bo
                         log.error(f'Failed to get code eval output after {self.num_retries} retries. Error: {e}')
                     log.warning(f'Failed to get code eval output, retrying in {self.backoff**i} seconds.')
                     time.sleep(self.backoff**i)
+                elif e.status == 401:
```
aspfohl, line 56: nit
```suggestion
                elif e.status == HTTPStatus.UNAUTHORIZED:
```
add `from http import HTTPStatus` above from the standard py library https://docs.python.org/3/library/http.html

composer/metrics/nlp.py
dakinggg: +1

PR #2682, opened 2023-10-31
Lower sequence generation length on code gen to be dependent on max canonical solution length 

bmosaicml:
# What does this PR do?

This PR lowers the max seq len for code gen tasks to be max canonical soln length + 10. Previously it had been max seq len - max prompt length, which was superfluously large on all of our code gen benchmarks leading to wasted time and wasted memory.

Test run on 16 40GB A100s: `mpt-30b-code-eval-TqwVGm` is possible but requires `num_beams=5`

# Before submitting


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -1101,7 +1107,7 @@ def collate_fn(self, data):
             'test_outputs': test_outputs,  # list of test outputs
             'languages': languages,  # list of languages
             'pass_at_k': self.pass_at_k,
-            'generation_length': self.max_seq_len - self.max_prompt_length,
+            'generation_length': min(self.max_answer_length, self.max_seq_len - self.max_prompt_length),
```
maxisawesome, line 1110: if max_answer_length is longer than max_seq_len - max_prompt_len, this would set the generation_length to max_seq_len - max_prompt_len. Is that the intended behavior?

dakinggg (commented):
Would like someone on eval team to approve that we are ok running code generation tasks this way. Functionally, LGTM

composer/datasets/in_context_learning_evaluation.py
bmosaicml: yeah, I think otherwise we will exceed max_seq_len, which should be a hard cutoff. Tradeoff is people shouldn't set max_seq_len too low. I can log an error/warning message if the condition you mentioned holds.

maxisawesome (approved):
LGTM

dakinggg (approved):

PR #2700, opened 2023-11-08
Properly handle empty metric_names passed to Trainer._filter_metrics

irenedea:
# What does this PR do?
Addresses error using `use_train_metrics: false`: `KeyError: 'continuation_indices'`
Manual test runs:
no fix: empty-eval-gFgK8Q 
with fix: empty-eval-with-fix-wpQHSK 

Changes
(1) Previously, if `metric_names` is an empty list, we return `metrics` without filtering. This is incorrect as an empty list should signify no metrics. We update this to handle an empty list.
(2) evaluator.metric_names should default to None not an empty list, so we can properly populate it with defaults later

# Before submitting

### DIFF
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -86,11 +86,10 @@ class Evaluator:
         self.label = label
         self.dataloader = ensure_data_spec(dataloader)
 
-        self.metric_names = []
         if metric_names is not None:
             if not isinstance(metric_names, list):
                 raise ValueError(f'``metric_names`` should be a list of strings, not a {type(metric_names)}')
-            self.metric_names = metric_names
+        self.metric_names = metric_names
 
         self.subset_num_batches = subset_num_batches
         self._eval_interval = None
@@ -132,8 +131,8 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
     """
     if isinstance(evaluator, Evaluator):
         # If no metric_names are specified, populate with the defaults.
-        if len(evaluator.metric_names) == 0:
-            evaluator.metric_names.extend(default_metric_names)
+        if evaluator.metric_names is None:
+            evaluator.metric_names = default_metric_names
         return evaluator
     else:
         return Evaluator(
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -130,9 +130,6 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
         Evaluator: An evaluator.
     """
     if isinstance(evaluator, Evaluator):
-        # If no metric_names are specified, populate with the defaults.
-        if evaluator.metric_names is None:
-            evaluator.metric_names = default_metric_names
         return evaluator
     else:
         return Evaluator(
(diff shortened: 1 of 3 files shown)


irenedea:
TODO: see why it's causing a test failure

composer/core/evaluator.py
```diff
@@ -86,11 +86,10 @@ def __init__(
         self.label = label
         self.dataloader = ensure_data_spec(dataloader)
 
-        self.metric_names = []
         if metric_names is not None:
             if not isinstance(metric_names, list):
                 raise ValueError(f'``metric_names`` should be a list of strings, not a {type(metric_names)}')
-            self.metric_names = metric_names
+        self.metric_names = metric_names
```
irenedea, line 92: this preserves `metric_names` as `None`

composer/trainer/trainer.py
```diff
@@ -105,15 +105,14 @@ def _get_default_scheduler_frequency(schedulers: Optional[Union[Scheduler, Seque
 
 def _filter_metrics(metrics: Dict[str, Metric], metric_names: Optional[List[str]]) -> Dict[str, Metric]:
     """Filter the metrics based on the given metric_names as regex strings (e.g. 'Accuracy', 'f1' for 'BinaryF1Score', 'Top-.' for 'Top-1 Accuracy' and 'Top-2 Accuracy', etc). If no metric_names are provided, all metrics will be returned."""
+    if metric_names is None:
+        raise ValueError('metric_names is None but should be List[str].')
     metrics = deepcopy(metrics)
-    if not metric_names:
-        return metrics
```
mvpatel2000, line 110: Issue is `not []` evaluates to `True`. Instead, this should be `if metric_names is None`.

@dakinggg  this is your favorite kind of bug

mvpatel2000 (changes requested):
1. Can you add a unit test for this?
2. The PR changes the effective public API by changing the behavior of what happens when you don't specify metric_names. I think that's not intended -- we should still default to all metrics if none are specified (putting aside the concerns of the current API design of deepcopying metrics etc...). What if we did a milder change -- keep behavior as-is for `None`, and instead filter all if user list is `[]`?

irenedea:
@mvpatel2000 Actually I think it does preserve the original intended behavior. What happens around the [callsites](https://github.com/mosaicml/composer/blob/3830a55fa4676701fe6c4ca6dbbe6ed447e10218/composer/trainer/trainer.py#L1204) for `_filter_metrics` is the following:
(1) Upon instantiation of an evaluator, evaluator.metric_names is None by default. (Previously, it was an empty list by default, so it was problematic that we weren't able to differentiate because default and empty list.)
(2) `ensure_evaluator(evaluator, default_metrics=model_metric_names)` is called for each evaluator. I made the change that this should populate the evaluator with the defaults (aka all metrics) if not set yet. It seemed appropriate given the call parameters.
(3) `evaluator.metric_names` is now populated with defaults and used in `_filter_metrics`

irenedea:
Hm, though on second thought, it could be nicer to just not populate the defaults in ensure_evaluator, so we only need to edit _filter_metrics and evaluator init. The biggest thing was that evaluator.metric_names was never None previously, so we never actually differentiated between emtpy list and defaults.

mvpatel2000 (approved):
LGTM! Thanks for hunting this down :)

PR #2702, opened 2023-11-08
Switch mosaicml logger to use futures to enable better error handling

j316chuck:
# What does this PR do?
Switch mosaicml logger to use futures to enable better error handling.

Helps close out this [PR](https://github.com/mosaicml/composer/pull/2666) from Tyler. 

Tests that if mcli retries fail (currently it'll retry 10x with an exponentail backoff), then the run fails. 

# Tests
Added unit tests


siriuslee (commented):
LGTM! Thanks for adding the test

composer/loggers/mosaicml_logger.py
```diff
@@ -177,20 +180,20 @@ def _flush_metadata(self, force_flush: bool = False) -> None:
         """Flush buffered metadata to MosaicML if enough time has passed since last flush."""
         if self._enabled and (time.time() - self.time_last_logged > self.log_interval or force_flush):
             try:
-                mcli.update_run_metadata(self.run_name, self.buffered_metadata)
+                f = mcli.update_run_metadata(self.run_name, self.buffered_metadata, future=True, protect=True)
```
dakinggg, line 183: is the retry logic now inside this call?

j316chuck: Yes I believe so but I couldn't find where that retry logic was. @siriuslee mind linking me to how the retry logic is implemented?

dakinggg, line 183: The reason we implemented the "turn off after 3 failures" is that internal users runs were getting killed due to mapi connection issues. Is there any way we can get the behavior we need (customer runs can't continue if this logger stop working), while allowing internal runs to continue passed this logger failing? e.g. some flag like "ignore failures" that we can turn on internally.

siriuslee: Yes, it's within the the `update_run_metadata` call now. We use:
```
MAX_RETRIES = 10
MAX_INTERVAL = 60
RETRY_EXCEPTIONS = {
    HTTPStatus.TOO_MANY_REQUESTS,
    HTTPStatus.INTERNAL_SERVER_ERROR,
    HTTPStatus.BAD_GATEWAY,
    HTTPStatus.SERVICE_UNAVAILABLE,
    HTTPStatus.GATEWAY_TIMEOUT,
}
backoff.on_exception(
        backoff.expo,
        MAPIException,
        max_tries=MAX_RETRIES,
        jitter=backoff.random_jitter,
        giveup=should_giveup,
        logger=logger,
        backoff_log_level=logging.DEBUG,
        giveup_log_level=logging.DEBUG,
        max_value=MAX_INTERVAL,
    )
```

siriuslee: That's reasonable. The error handling can be disabled with a try - except around the `f.exception` call. You could disable the logger if that throws and the user specified to ignore errors. It'd be great if we can default to throw the error, though, and only suggest folks set that flag if it becomes a problem

dakinggg: sounds good to me, just want to give people an easy out if it is a problem

mvpatel2000: If we raise the error, we should wrap it with instructions on how to disable logger

j316chuck: Fixed

composer/loggers/mosaicml_logger.py
```diff
@@ -140,20 +147,25 @@ def _flush_metadata(self, force_flush: bool = False) -> None:
...
-                log.error(f'Failed to log metadata to Mosaic with error: {e}')
-                self.allowed_fails_left -= 1
-                self.time_failed_count_adjusted = time.time()
-                if self.allowed_fails_left <= 0:
+                log.info(f'Failed to log metadata to Mosaic with error: {e}')
+                if self.ignore_exceptions:
+                    log.info('Ignoring exception and disabling MosaicMLLogger.')
                     self._enabled = False
+                else:
+                    log.info('Raising exception. To ignore exceptions, set ignore_exceptions=True.')
```
Skylion007, line 167: Any reason we don't use log.exception here?

j316chuck: changed

dakinggg (approved):

composer/loggers/mosaicml_logger.py
```diff
@@ -140,20 +144,25 @@ def _flush_metadata(self, force_flush: bool = False) -> None:
...
+                for f in done:
+                    if f.exception() is not None:
+                        raise f.exception()  # type: ignore
+                self._futures = list(incomplete)
             except Exception as e:
-                log.error(f'Failed to log metadata to Mosaic with error: {e}')
-                self.allowed_fails_left -= 1
-                self.time_failed_count_adjusted = time.time()
-                if self.allowed_fails_left <= 0:
+                log.exception(f'Failed to log metadata to Mosaic with error: {e}')
```
Skylion007, line 159: Do you actually need to pass int the exception in the string? Double check because I think it should print out exception logging info.

docs/source/doctest_fixtures.py
```diff
@@ -92,6 +93,11 @@
 # Disable wandb
 os.environ['WANDB_MODE'] = 'disabled'
 
+# Disable MosaicMLLogger
+os.environ['MOSAICML_PLATFORM'] = 'false'
+if 'MOSAICML_ACCESS_TOKEN_FILE' in os.environ:
+    del os.environ['MOSAICML_ACCESS_TOKEN_FILE']
```
Skylion007, line 99:
```suggestion
os.environ.pop('MOSAICML_ACCESS_TOKEN_FILE', None)
```

composer/loggers/mosaicml_logger.py
j316chuck: You don't, thanks for this nit!

tests/fixtures/autouse_fixtures.py
```diff
@@ -118,8 +120,12 @@ def seed_all(rank_zero_seed: int, monkeypatch: pytest.MonkeyPatch):
 @pytest.fixture(autouse=True)
 def mapi_fixture(monkeypatch):
     # Composer auto-adds mosaicml logger when running on platform. Disable logging for tests.
-    mock_update = lambda *args, **kwargs: None
+    future_obj = Future()
+    future_obj.set_result(None)
+    mock_update = lambda *args, **kwargs: future_obj
     monkeypatch.setattr(mcli, 'update_run_metadata', mock_update)
+    # Disable the MosaicMLLogger by default
+    monkeypatch.setenv(MOSAICML_PLATFORM_ENV_VAR, 'false')
```
mvpatel2000, line 128: Because of this, don't we need to explicitly turn it back on in `test_mosaicml_logger.py`?

docs/source/doctest_fixtures.py
```diff
@@ -92,6 +93,10 @@
 # Disable wandb
 os.environ['WANDB_MODE'] = 'disabled'
 
+# Disable MosaicMLLogger
+os.environ['MOSAICML_PLATFORM'] = 'false'
+os.environ.pop('MOSAICML_ACCESS_TOKEN_FILE', None)
```
mvpatel2000, line 98: This has side-effects....

but maybe its fine

j316chuck: we can remove I don't think we need this anymore.

tests/fixtures/autouse_fixtures.py
j316chuck: we can remove I don't think we need this either.

docs/source/doctest_fixtures.py
```diff
@@ -53,6 +53,7 @@
 from composer.loggers import InMemoryLogger as InMemoryLogger
 from composer.loggers import Logger as Logger
 from composer.loggers import RemoteUploaderDownloader
+from composer.loggers.mosaicml_logger import MOSAICML_PLATFORM_ENV_VAR
```
mvpatel2000, line 56:
```suggestion
```

mvpatel2000 (approved):
@j316chuck  can we run one final manual test because we've changed a lot

j316chuck:
test run name: `composer-mpt-125m-chinchilla-regression-LqM65R`

PR #2704, opened 2023-11-09
Train loss NaN checking callback

coryMosaicML:
Add a simple NaN catching callback that will raise an error if a NaN in the train loss is found.


mvpatel2000 (approved):
LGTM thanks!!

CC: @dakinggg  we should add this into foundry on next release

PR #2706, opened 2023-11-09
Fix MPS with dict loss

mvpatel2000:
# What does this PR do?

Fix MPS with dict loss. Previously, the move to CPU would error if the loss was a dict, which was a feature added after this code was written. I manually tested locally.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -20,7 +20,7 @@ import warnings
 from collections import defaultdict
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Callable, ContextManager, Dict, Iterable, List, Optional, Sequence, TextIO, Tuple, Union, cast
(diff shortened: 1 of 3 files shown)


composer/trainer/trainer.py
```diff
@@ -2911,7 +2912,15 @@ def _eval_loop(
                                 if isinstance(self.state.device, DeviceMPS):
                                     # torchmetrics math has numerical errors on M1 devices
                                     # running the compute on CPU instead
-                                    outputs = self.state.outputs.cpu()
+                                    if isinstance(self.state.outputs, Mapping):
```
dakinggg, line 2915: do we need to account for nesting? I know we've had this discussion wrt torchmetrics before....

mvpatel2000: Hm... I took a quick look and I don't see any issues. If it nests, we just won't move it and there's risk that it gives numerical issues until torchmetrics fixes it which I think is fine

dakinggg (approved):

PR #2710, opened 2023-11-13
Fetching arguments for FSDP

mvpatel2000:
# What does this PR do?

Adds arguments for FSDP fetching. Safe after discussion with torch folks

[GRT-2569](https://databricks.atlassian.net/browse/GRT-2569)

### DIFF
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -510,8 +510,11 @@ def prepare_fsdp_module(
                 **kwargs,
             )
 
-            fsdp_obj._exec_order_data._forward_prefetch_limit = fsdp_config['forward_prefetch_limit']
-            fsdp_obj._exec_order_data._backward_prefetch_limit = fsdp_config['backward_prefetch_limit']
+            if hasattr(fsdp_obj, '_exec_order_data'):
+                if hasattr(fsdp_obj._exec_order_data, '_forward_prefetch_limit'):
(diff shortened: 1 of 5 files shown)


vchiley:
Modifying torch private variables == ❤️‍🔥 😍 

Does this need to be gated by torch version?

mvpatel2000:
> Does this need to be gated by torch version?

Ah good point. The code is the same and static since torch 1.13 so it should be fine. I have added a unit test to verify it doesn't crash as well

composer/trainer/dist_strategy.py
```diff
@@ -508,6 +510,9 @@ def _auto_wrap_policy_old(module: torch.nn.Module, recurse: bool, unwrapped_para
                 **kwargs,
             )
 
+            fsdp_obj._exec_order_data._forward_prefetch_limit = fsdp_config['forward_prefetch_limit']
```
dakinggg, line 513: Should probably check for the existence of these attrs since these are private...

mvpatel2000: I would prefer to leave as is since I have verified on torch versions so far, and I would rather have test fail while we upgrade to new torch so it flags as requiring change

dakinggg: These can change in a patch release though, which will instantly break everyone's training jobs.

dakinggg: Should also probably warn, with a note saying to open a composer issue

dakinggg (approved):

PR #2714, opened 2023-11-14
Add train finished run event

jjanezhang:
# Add train finished run event
- Adds timestamp when model is finished training to the run metadata 
- Will create a new `TRAIN_FINISHED` event in MAPI

# Testing
1. Added an assert in the unit test to make sure train finished is logged.
2. Confirmed that train finished time is added to run metadata with same timestamp as the last batch finishes
image
image


eracah (commented):
The code looks good! But I am curious if we want to log when the actual training finished or when everything is finished and about to shutdown. The reason is that as the code stands now, the train_finish_time wouldn't get logged until the last checkpoint was sucessfully uploaded to the cloud, which can take an extra 30min-1hour after the last batch_end depending on the size of the checkpoint.

eracah:
> The code looks good! But I am curious if we want to log when the actual training finished or when everything is finished and about to shutdown. The reason is that as the code stands now, the train_finish_time wouldn't get logged until the last checkpoint was sucessfully uploaded to the cloud, which can take an extra 30min-1hour after the last batch_end depending on the size of the checkpoint.

Perhaps running on the last `batch_end` or making sure `MosaicMLLogger` comes before `RemoteUploaderDownloader` in the list of loggers. @dakinggg, wdyt?

composer/loggers/mosaicml_logger.py
```diff
@@ -113,6 +113,8 @@ def epoch_end(self, state: State, logger: Logger) -> None:
         self._flush_metadata()
 
     def fit_end(self, state: State, logger: Logger) -> None:
+        # Log model training finished time for run events
+        self._log_metadata({'train_finished_time': time.time()})
```
mvpatel2000, line 117: Should this subtract off start time? or do u want absolute

dakinggg:
Yeah @eracah I think we want to log this before the wait for checkpoint upload. Is it simplest to make sure the mosaic logger runs before the RUD?

composer/loggers/mosaicml_logger.py
jjanezhang: I am reporting the timestamp for when training finishes so we want absolute!

eracah:
> Yeah @eracah I think we want to log this before the wait for checkpoint upload. Is it simplest to make sure the mosaic logger runs before the RUD?

Yup I think so! And any other RUDs that callbacks create (but I think those will always be added to the end of the list of callbacks)

composer/trainer/trainer.py
```diff
@@ -1078,18 +1078,18 @@ def __init__(
...
         if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and os.environ.get(
                 MOSAICML_ACCESS_TOKEN_ENV_VAR) is not None and not any(isinstance(x, MosaicMLLogger) for x in loggers):
             log.info('Detected run on MosaicML platform. Adding MosaicMLLogger to loggers.')
             mosaicml_logger = MosaicMLLogger()
             loggers.append(mosaicml_logger)
 
+        if save_folder is not None:
+            remote_ud = maybe_create_remote_uploader_downloader_from_uri(save_folder, loggers)
+            if remote_ud is not None:
+                loggers.append(remote_ud)
```
mvpatel2000, line 1091: Why is this moved?

mvpatel2000 (approved):

composer/trainer/trainer.py
jjanezhang: Moved so that the callback to log Mosiac info comes before RUD

eracah: @mvpatel2000, see https://github.com/mosaicml/composer/pull/2714#issuecomment-1813117839

eracah (approved):
LGTM. Thanks, @jjanezhang !

eracah:
> LGTM. Thanks, @jjanezhang !

Maybe add a comment describing why RUD has to be last?

PR #2717, opened 2023-11-16
Bump torch to 2.1.1 version

j316chuck:
# What does this PR do?
Bump base torch to 2.1.1 

https://databricks.atlassian.net/browse/UWU-2134


mvpatel2000 (commented):
You need to also bump setup.py and the mosaic_fsdp wrap. I think we can just remove the `else` clause since we need no more monkeypatches CC: @b-chu

j316chuck:
Bumped up setup.py. Let's not remove else since we might need a docker image down the line for @eracah's DTensor work.

j316chuck:
Oh i thought the else was for nightly else.

composer/trainer/mosaic_fsdp.py
```diff
@@ -50,5 +50,6 @@ def patch_pytorch():
         # Monkey patch partial state dict handling
         _state_dict_utils._sharded_pre_load_state_dict_hook = (_sharded_pre_load_state_dict_hook)
 
-    elif version.parse(torch.__version__) >= version.parse('2.1.1'):
-        raise NotImplementedError(f'FullyShardedDataParallel is not supported for torch >= 2.2.0')
+    else:
+        # Nothing to patch yet for torch >= 2.1.1
+        pass
```
mvpatel2000, line 55:
```suggestion
```

mvpatel2000 (approved):
LGTM. Can we remove `else pass` though

PR #2724, opened 2023-11-16
Fix activation cpu offloading

cli99:
This PR fixes activation cpu offloading. The original implementation 1) breaks when offload_to_cpu is enabled and activation_checkpointing is disabled 2) does not offload to cpu when activation_checkpointing and  offload_to_cpu are both enabled. See https://github.com/pytorch/pytorch/pull/85459.


composer/trainer/dist_strategy.py
```diff
@@ -535,20 +535,19 @@ def _auto_wrap_policy_old(module: torch.nn.Module, recurse: bool, unwrapped_para
...
-                        lambda module: checkpoint_wrapper(
-                            first_wrap_fn(module),  # type: ignore reportGeneralTypeIssues
-                            offload_to_cpu=True)) if activation_cpu_offload else first_wrap_fn
+                        lambda module: offload_wrapper(first_wrap_fn(module))  # type: ignore reportGeneralTypeIssues
+                    ) if activation_cpu_offload else first_wrap_fn
 
                 # Choose which modules to activation checkpoint according to the following priority:
                 # If module has attribute `module._activation_checkpointing = ...`, always respect it
                 # Otherwise checkpoint if root object `obj.activation_checkpointing_fn(module)` is true
+
```
mvpatel2000, line 550:
```suggestion
```
was this accidental?

composer/trainer/dist_strategy.py
```diff
@@ -221,7 +221,7 @@ def prepare_fsdp_module(
     is_torch_2_0 = using_torch_2()
     from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (CheckpointImpl,
                                                                              apply_activation_checkpointing,
-                                                                             checkpoint_wrapper)
+                                                                             checkpoint_wrapper, offload_wrapper)
```
mvpatel2000, line 224: can you verify this fn exists in all torch versions (>=1.13?)

mvpatel2000 (approved):

composer/trainer/dist_strategy.py
cli99: it turned out the offload_wrapper was added after 2.0.0, for 1.13, we will use the old implementation. I will add a guard on versions

mvpatel2000: Should import be inlined

mvpatel2000 (approved):

mvpatel2000 (approved):
Can we add a unit test for this? Maybe just take a run in `test_fsdp` and add activation ckpting

cli99:
@mvpatel2000 added unit test and inline import

tests/trainer/test_fsdp_act_ckpt_offload.py
```diff
@@ -0,0 +1,69 @@
...
+@world_size(2)
+@pytest.mark.gpu
+@pytest.mark.parametrize('activation_checkpointing', [True, False])
+@pytest.mark.parametrize('activation_cpu_offload', [True, False])
+def test_fsdp_act_ckpt_offload(
+    activation_checkpointing: bool,
+    activation_cpu_offload: bool,
+    world_size: int,
+):
+    model = (SimpleModel())
```
mvpatel2000, line 40:
```suggestion
    model = SimpleModel()
```

mvpatel2000 (approved):
Can we move test to `test_fsdp`?

PR #2726, opened 2023-11-17
Extend checkpoint loading to accept a validation function

irenedea:
# What does this PR do?
If a validation function is specified by the environment variable `CHECKPOINT_VALIDATION_FUNCTION`, the function is called prior to loading checkpoints via `torch.load`. 

The validation function is expected to be type `Callable[[Union[str, Path], bool]`, accepting the path to the checkpoint file and returning a boolean denoting whether or not the checkpoint is valid.

# Before submitting


irenedea:
it's ready for review! @eracah @dakinggg

tests/trainer/test_checkpoint.py
```diff
@@ -1289,3 +1290,42 @@ def test_rotate_checkpoints(
...
 
     dist.barrier()  # all ranks finish before cleaning up tmpdir
+
+
+def simple_validate(filepath: str):
+    with open(filepath, 'r') as f:
+        return f.read() == 'good'
+
+
+def test_checkpoint_validation(tmp_path):
```
dakinggg, line 1300: would be good to test this integration with trainer too, in addition to these unit tests

dakinggg (approved):
This LGTM, can you please open a PR to foundry to make sure everywhere we call `torch.load` uses the `safe_torch_load` util from composer?

composer/utils/checkpoint.py
```diff
@@ -39,6 +40,62 @@
...
+    fn = getattr(module, fn_name)
+    log.debug(f'Checkpoint validation function {name} has been found.')
+    return fn
+
+
+def _ensure_valid_checkpoint(checkpoint_filepath: Union[Path, str]) -> Union[Path, str]:
+    """Ensures that the checkpoint at checkpoint_filepath is valid.
+
+    using the function specified by the CHECKPOINT_VALIDATION_FUNCTION environment variable.
+    If CHECKPOINT_VALIDATION_FUNCTION, we skip validation.
```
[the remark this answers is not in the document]
irenedea:
```suggestion
    If CHECKPOINT_VALIDATION_FUNCTION is not set, we skip validation.
```

composer/utils/checkpoint.py
```diff
@@ -716,7 +773,7 @@ def safe_torch_load(
 
             return state_dict
         else:
-            return torch.load(composer_states_filepath, map_location=map_location)
+            return _torch_load_with_validation(composer_states_filepath, map_location=map_location)
```
eracah, line 776: note that elastic sharded checkpoints don't use torch.load (they use `torch.distributed.checkpoint.load_state_dict`), so this validation function wouldn't be called for that

composer/utils/checkpoint.py
```diff
@@ -39,6 +40,62 @@
...
+def _ensure_valid_checkpoint(checkpoint_filepath: Union[Path, str]) -> Union[Path, str]:
+    """Ensures that the checkpoint at checkpoint_filepath is valid.
+
+    using the function specified by the CHECKPOINT_VALIDATION_FUNCTION environment variable.
+    If CHECKPOINT_VALIDATION_FUNCTION is not set, we skip validation.
+
+    Args:
+        checkpoint_filepath (Union[Path,str]): The path to the checkpoint file.
+    """
+    fn_name = os.environ.get('CHECKPOINT_VALIDATION_FUNCTION', None)
```
eracah, line 71: why is the function is set by env variable? that seems very strange. why? and is there a bank of validation functions somewhere or something. are we setting the env variable or a user theoretically would? why would this not be a Trainer arg?

eracah (commented):
Nice! Sleekly done! looks pretty good, but I just had a few questions and comments

composer/utils/checkpoint.py
dakinggg: where is the file load for sharded checkpoints? Doesn't it still go through `_restore_checkpoint` (which starts by loading the state dict into memory)

dakinggg: ah nvm, i see it is `load_sharded_checkpoint`

dakinggg: Do you know where we can inject this for sharded checkpoints? It needs to run before we load the file into memory. Maybe loop over folder before calling `load_state_dict`?

composer/utils/checkpoint.py
```diff
@@ -338,8 +399,35 @@ def _get_num_ranks_that_saved_rng(metadata: Metadata):
...
+        """FileSystemReader that validates checkpoint files prior to reading."""
+
+        def __init__(self, path: str):
+            super().__init__(path)
+
+        def read_data(self, plan: LoadPlan, planner: LoadPlanner):
+            """Reads metadata file.
+
+            Raises:
+                ValueError if the metadata file is invalid.
```
irenedea, line 412:
```suggestion
            """Reads data files.

            Raises:
                ValueError if the data files are not valid checkpoints.
```

composer/utils/checkpoint.py
```diff
@@ -338,8 +399,37 @@ def _get_num_ranks_that_saved_rng(metadata: Metadata):
...
+            """
+            metadata_file_path = self.path / '.metadata'
+            _ensure_valid_checkpoint(metadata_file_path)
+            return super().read_metadata()
+
     # A subclass of FileSystemReader that downloads files from the object store before reading them from the local filesystem.
```
irenedea, line 431:
```suggestion
    # A subclass of FileSystemReaderWithValidation that downloads files from the object store before reading them from the local filesystem.
```

eracah (approved):
well done! LGTM

PR #2728, opened 2023-11-17
Check serialization for JSON in mosaicml logger

mvpatel2000:
# What does this PR do?

Check serialization for JSON in mosaicml logger. I can't add a unit test because if we had a way to cause it to not serialize, we would patch it...

### DIFF
--- a/composer/loggers/mosaicml_logger.py
+++ b/composer/loggers/mosaicml_logger.py
@@ -7,11 +7,11 @@ from __future__ import annotations
 
 import collections.abc
 import fnmatch
+import json
 import logging
 import operator
 import os
 import time
-import json
 import warnings
 from concurrent.futures import wait
 from functools import reduce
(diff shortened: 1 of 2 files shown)


dakinggg:
To test, just make some object that raises an error when you try to json serialize it and log that object. also may want to merge my pr first

dakinggg (approved):

composer/loggers/mosaicml_logger.py
```diff
@@ -220,21 +221,23 @@ def format_data_to_json_serializable(data: Any):
...
+        elif isinstance(data, collections.abc.Mapping):
+            ret = {format_data_to_json_serializable(k): format_data_to_json_serializable(v) for k, v in data.items()}
+        elif isinstance(data, collections.abc.Iterable):
+            ret = [format_data_to_json_serializable(v) for v in data]
+        else:  # Unknown format catch-all
+            ret = str(data)
+        json.dumps(ret)  # Check if ret is JSON serializable
+        return ret
     except RuntimeError as e:
         warnings.warn('Encountered unexpected error while formatting data to be JSON serializable. '
```
dakinggg, line 242: should we add the type of `data` to this warning message?

PR #2729, opened 2023-11-17
Add support for canned ACL environment variable

nik-mosaic:
# What does this PR do?

Adds support for environment variable, S3_CANNED_ACL, which if set is added to all S3ObjectStore.upload_object requests. If an ACL string is passed into upload_object directly, then the environment variable does not apply.

## Testing
Added mock tests to ensure arguments are set as expected.

Also tested this locally. The behavior is expected:
Case 1: No ACL passed in to upload_object, no environment variable set
Result: Object uploaded without custom ACL

Case 2: ACL passed in, no environment variable set
Result: Uses ACL passed in

Case 3: No ACL passed in, environment variable set:
Result: Uses environment variable

Case 4: ACL passed in, environment variable set
Result: Uses ACL passed in.

# Before submitting
- [x ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [x ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites)

### DIFF
--- a/tests/utils/object_store/test_s3_object_store.py
+++ b/tests/utils/object_store/test_s3_object_store.py
@@ -4,6 +4,8 @@
 import os
 import pathlib
 import threading
+from unittest import mock
+from unittest.mock import ANY, MagicMock
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


dakinggg (commented):
lgtm, but can you add a couple simple tests using mocks that just make sure the args to upload_object get set as you expect?

dakinggg (approved):

PR #2722, opened 2023-11-16
Override nvidia env var for 11.8

dakinggg:
# What does this PR do?
NVIDIA has pushed new base images that seem to be missing forward compatibility in the `NVIDIA_REQUIRE_CUDA` env var. We are overriding it here specifically for the 11.8 images that seem to be mistaken.

TODO:
- [x] check 11.8 staging image for new env var
- [x] check 12.1 staging image for correct env var
- [x] test resulting 11.8 staging image can be run on r1z2
- [x] test resulting 12.1 staging image can be run on r1z2
- [ ] [after merge] release new docker images

# What issue(s) does this change relate to?
[GRT-2582](https://databricks.atlassian.net/browse/GRT-2582)

PR #2742, opened 2023-11-28
Override NVIDIA environment variable for CUDA 12.1 images

bandish-shah:
# What does this PR do?

NVIDIA has pushed new base images that seem to be missing forward compatibility in the NVIDIA_REQUIRE_CUDA env var. We are overriding it here specifically for the 11.1 images.

To Do:
- [x] check 11.8 staging image for correct env var
- [x] check 12.1 staging image for new env var
- [x] test resulting 11.8 staging image can be run on r1z2
- [x] test resulting 12.1 staging image can be run on r1z2
- [ ] [after merge] release new docker images

Related to #2722.


j316chuck (approved):
Thanks!

PR #2745, opened 2023-11-28
fix checkpointing with activation checkpointing

cli99:
This PR fixes loading checkpoints with a different activation checkpointing granularity.

FSDP activation checkpointing wraps the module with `_checkpoint_wrapped_module`, when loading a model checkpoint, besides `_fsdp_wrapped_module`, `_checkpoint_wrapped_module` also needs to be striped out of the module name so that the name mapping is correct when loading state dict.


j316chuck (commented):
Can we add a test for this if it takes < 1 hour, P2 but would be nice 😃 . 

If not, can you add a comment describing what the replace line enables for context.

j316chuck:
Also, curious why we are merging to abra?

cli99:
@j316chuck, this branch fixes the above mentioned issue in the abra branch. The current dev branch isn't actually saving/loading the correct moe model, and the abra branch is needed for moes.

j316chuck (approved):

Issue #2743, opened 2023-11-28
eval drops when i have multi_eval dataset


germanjke:
Hi, I changed eval dataset to multi_eval dataset with this [pull request](https://github.com/mosaicml/llm-foundry/pull/603/)

Before:

```
eval_loader:
  dataset:
    max_seq_len: ${max_seq_len}
    shuffle: true
    shuffle_seed: ${global_seed}
    streams:
      a:
        local: a
        remote: s3://a
        repeat: 1.0
        split: val
  drop_last: true
  name: text
  num_workers: 8
  ```

After:

```
eval_loader:
- label: general
  dataset:
    max_seq_len: ${max_seq_len}
    shuffle: true
    shuffle_seed: ${global_seed}
    streams:
      a:
        local: a
        remote: s3://a
        repeat: 1.0
        split: val
  drop_last: true
  name: text
  num_workers: 8
- label: multi_eval_train_subset
  dataset:
    max_seq_len: ${max_seq_len}
    shuffle: true
    shuffle_seed: ${global_seed}
    streams:
      cc2021_39:
        local: b
        remote: s3://b
        repeat: 1.0
        split: train
  drop_last: true
  name: text
  num_workers: 8
- label: multi_eval_val_subset
  dataset:
    max_seq_len: ${max_seq_len}
    shuffle: true
    shuffle_seed: ${global_seed}
    streams:
      c:
        local: c
        remote: s3://c
        repeat: 1.0
        split: val
  drop_last: true
  name: text
  num_workers: 8
  ```

My loaders build, proofed by this logs:

```
Building train loader...
Building eval loader...
Initializing model...
Building trainer...
```

But I have this error in ` self._train_loop() `

```
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /llm-foundry/scripts/train/train.py:665 in <module>                     │
│                                                                              │
│   662 │   cfg = om.merge(yaml_cfg, cli_cfg)                                  │
│   663 │   om.resolve(cfg)                                                    │
│   664 │   assert isinstance(cfg, DictConfig)                                 │
│ ❱ 665 │   main(cfg)                                                          │
│   666                                                                        │
│                                                                              │
│ /tgpt/llm-foundry/scripts/train/train.py:651 in main                         │
│                                                                              │
│   648 │   │   trainer.eval()                                                 │
│   649 │                                                                      │
│   650 │   print('Starting training...')                                      │
│ ❱ 651 │   trainer.fit()                                                      │
[...]

PR #2746, opened 2023-11-28
Fix console logger for small datasets.

mvpatel2000:
# What does this PR do?

Fix console logger for small datasets. Adds a check for to verify removed value exists.

https://github.com/mosaicml/composer/issues/2743

### DIFF
--- a/composer/loggers/console_logger.py
+++ b/composer/loggers/console_logger.py
@@ -162,7 +162,8 @@ class ConsoleLogger(LoggerDestination):
         # Remove index of last batch, so that we don't print progress at end of last batch and then
         # at eval end.
         last_batch_idx = total_eval_batches
-        self.eval_batch_idxs_to_log.remove(last_batch_idx)
+        if last_batch_idx in self.eval_batch_idxs_to_log:  # eval_batch_idxs_to_log may be empty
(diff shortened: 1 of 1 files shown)


composer/loggers/console_logger.py
```diff
@@ -162,7 +162,8 @@ def eval_start(self, state: State, logger: Logger) -> None:
         # Remove index of last batch, so that we don't print progress at end of last batch and then
         # at eval end.
         last_batch_idx = total_eval_batches
-        self.eval_batch_idxs_to_log.remove(last_batch_idx)
+        if last_batch_idx in self.eval_batch_idxs_to_log:  # eval_batch_idxs_to_log may be empty
```
dakinggg, line 165: tiny nit, but if the only scenario we're checking here is when its empty, lets just have the if statement check if its empty?

dakinggg (approved):

eracah (approved):
LGTM

composer/loggers/console_logger.py
mvpatel2000: I'm going to leave as is since it's a little stronger than known edge cases but I'd rather catch and avoid a scenario we haven't envisioned vs. raise an error

germanjke:
thanks!

PR #2747, opened 2023-11-29
Add metadata logging for wandb

jjanezhang:
# Add metadata logging for wandb

- We are noticing that wandb url specifically is being dropped from regression test runs inconsistently, adding logs to try to debug this moving forward.

See Jira: https://databricks.atlassian.net/browse/GRT-2588

### DIFF
--- a/composer/loggers/mosaicml_logger.py
+++ b/composer/loggers/mosaicml_logger.py
@@ -102,6 +102,7 @@ class MosaicMLLogger(LoggerDestination):
                 run_url = callback.run_url
                 if run_url is not None:
                     self._log_metadata({'wandb/run_url': run_url})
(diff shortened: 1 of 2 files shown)


dakinggg (approved):

composer/loggers/mosaicml_logger.py
```diff
@@ -102,6 +102,9 @@ def after_load(self, state: State, logger: Logger) -> None:
                 run_url = callback.run_url
                 if run_url is not None:
                     self._log_metadata({'wandb/run_url': run_url})
+                    log.debug(f'Logging WandB run URL to metadata: {run_url}')
+                else:
+                    log.debug(f'WandB run URL not found, not logging to metadata')
```
dakinggg, line 107: nit: this doesn't need to be a format string

PR #2748, opened 2023-11-29
Ignore load ignore keys

mvpatel2000:
# What does this PR do?

Ignore load ignore keys as that should only apply to initial autoresume


dakinggg (approved):
Can you add a simple test please? This is an edge case we hadn't considered, so I'd like to cover it.

mvpatel2000:
> Can you add a simple test please? This is an edge case we hadn't considered, so I'd like to cover it.

Added unit test. Also verified it fails before

PR #2750, opened 2023-12-01
Parallelize OCI multipart download

coryMosaicML:
This PR downloads objects from OCI in multiple parts (default `10` mimicing the CLI setting) in parallel.

A run that uses this PR to load a checkpoint to start training from, save new checkpoints, and resume from a saved checkpoint is here: `sdxl-download-test-fU6bBO`


composer/utils/object_store/oci_object_store.py
```diff
@@ -132,17 +142,31 @@ def download_object(
...
-                object_name=object_name,
-            )
+            # Download parts in parallel
+            args_list = []
+            for i in range(num_parts):
+                start_byte = i * part_size
+                end_byte = min(start_byte + part_size - 1, int(object_size) - 1)
+                args_list.append((object_name, filename, start_byte, end_byte, i))
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                parts = list(executor.map(lambda p: self._download_part(*p), args_list))
```
Skylion007, line 159: You should use concurrent.futures.as_completed(.as_completed to write the file parts so they don't all sit in memory unnecessarily and are written as the finish if possible? Unless we have a good reason for this?

composer/utils/object_store/oci_object_store.py
```diff
@@ -116,12 +117,21 @@ def upload_object(
         except Exception as e:
             _reraise_oci_errors(self.get_uri(object_name), e)
 
+    def _download_part(self, object_name, filename, start_byte, end_byte, part_number):
+        range_header = f'bytes={start_byte}-{end_byte}'
+        tmp_part_path = str(filename) + f'.part_{part_number}.{uuid.uuid4()}.tmp'
```
Skylion007, line 122: nit:
```suggestion
        tmp_part_path = f'{str(filename)}.part_{part_number}.{uuid.uuid4()}.tmp'
```

composer/utils/object_store/oci_object_store.py
```diff
@@ -132,17 +142,31 @@ def download_object(
...
+                args_list.append((object_name, filename, start_byte, end_byte, i))
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                parts = list(executor.map(lambda p: self._download_part(*p), args_list))
         except Exception as e:
             _reraise_oci_errors(self.get_uri(object_name), e)
 
-        with open(tmp_path, 'wb') as f:
-            f.write(response.data.content)
+        # Combine parts
+        tmp_path = str(filename) + f'.{uuid.uuid4()}.tmp'
```
Skylion007, line 164:
```suggestion
        tmp_path = f'{str(filename)}.{uuid.uuid4()}.tmp'

File: composer/utils/object_store/oci_object_store.py (hunk 1 above)
Comment (coryMosaicML) [reply composer/utils/object_store/oci_object_store.py:159]: The reason was to preserve order of the parts for merging. But I can find another solution that sorts them out after.

File: composer/utils/object_store/oci_object_store.py (hunk 3 above)
Comment (coryMosaicML) [reply composer/utils/object_store/oci_object_store.py]: Done!

File: composer/utils/object_store/oci_object_store.py (hunk 2 above)
Comment (coryMosaicML) [reply composer/utils/object_store/oci_object_store.py]: Done!

File: composer/utils/object_store/oci_object_store.py (hunk 1 above)
Comment (coryMosaicML) [reply composer/utils/object_store/oci_object_store.py]: Should be fixed to use `concurrent.futures.as_completed` now.

File: composer/utils/object_store/oci_object_store.py [hunk 4]
```diff
@@ -132,17 +145,41 @@ def download_object(
...
+        head_object_response = self.client.head_object(self.namespace, self.bucket, object_name)
+        object_size = head_object_response.headers['content-length']
+        # Calculate the part sizes
+        base_part_size = int(object_size) // num_parts
+        remainder = int(object_size) % num_parts
+        part_sizes = [base_part_size] * num_parts
+        for i in range(remainder):
+            part_sizes[i] += 1
+        part_sizes = [part_size for part_size in part_sizes if part_size > 0]
+
```
Skylion007, line 158: I'm worried about the temporary files not being cleaned up. The better way to make sure this is not the issue is to use a 
```with tempfile.TemporaryDirectory(dir=dirname) as temp_dir:
```
When ever this context manager is exited, all files in the folder will be deleted regardless of whether it exists because of an error etc.
`dir=dirname` ensures that the files are written to same device / location as the target file, but in a subfolder. You may also want to look at the prefix/suffix arg for additional customization.

composer/utils/object_store/oci_object_store.py
```diff
@@ -130,19 +144,45 @@ def download_object(
...
+        tmp_part_dir = os.path.join(dirname, f'{str(filename)}-parts-{uuid.uuid4()}')
+        os.makedirs(tmp_part_dir, exist_ok=True)
+        with TemporaryDirectory(dir=tmp_part_dir) as temp_dir:
```
Skylion007, line 161: Not quite: I am pretty sure TemporaryDirectory puts a temporary directory inside of `dir`. You don't need to put a temporary directory inside a temporary directory you manage. Just tell it to put a t
```suggestion
        with TemporaryDirectory(dir=dirname) as temp_dir:
```
If you want to specify a specific prefix, you can do:
```suggestion
        with TemporaryDirectory(prefix=f'{str(filename)}-parts-', dir=dirname) as temp_dir:
```

composer/utils/object_store/oci_object_store.py
```diff
@@ -130,19 +144,45 @@ def download_object(
...
+            try:
+                # Download parts in parallel
+                parts = []
+                with concurrent.futures.ThreadPoolExecutor() as executor:
+                    futures = []
+                    start_byte = 0
+                    for i, part_size in enumerate(part_sizes):
+                        end_byte = start_byte + part_size - 1
+                        futures.append(
+                            executor.submit(self._download_part, object_name, temp_dir, start_byte, end_byte, i))
```
Skylion007, line 171: Also not sure if this is what you want as it will save a weird filename with
f`{tempdir}/.part_{part_number}.{uuid.uuid4()}.tmp'. You may want to remove the `.` or add a prefix base on the actual filename?

composer/utils/object_store/oci_object_store.py
```diff
@@ -130,19 +144,45 @@ def download_object(
...
+
+            # Combine parts
+            tmp_path = f'{str(filename)}.{uuid.uuid4()}.tmp'
+            with open(tmp_path, 'wb') as outfile:
+                for i, part_file_name in parts:
+                    with open(part_file_name, 'rb') as infile:
+                        outfile.write(infile.read())
 
         if overwrite:
             os.replace(tmp_path, filename)
```
Skylion007, line 188:
```suggestion
            tmp_path = os.path.join(temp_dir, f'{str(filename)}.{uuid.uuid4()}.tmp')
            with open(tmp_path, 'wb') as outfile:
                for i, part_file_name in parts:
                    with open(part_file_name, 'rb') as infile:
                        outfile.write(infile.read())

            if overwrite:
                os.replace(tmp_path, filename)
```

Skylion007 (approved):
LGTM

composer/utils/object_store/oci_object_store.py
coryMosaicML: Fixed!

composer/utils/object_store/oci_object_store.py
coryMosaicML: Fixed!

composer/utils/object_store/oci_object_store.py
```diff
@@ -130,24 +144,48 @@ def download_object(
...
-        if overwrite:
-            os.replace(tmp_path, filename)
-        else:
-            os.rename(tmp_path, filename)
+        # Get the size of the object
+        head_object_response = self.client.head_object(self.namespace, self.bucket, object_name)
+        object_size = head_object_response.headers['content-length']
+        # Calculate the part sizes
+        base_part_size = int(object_size) // num_parts
+        remainder = int(object_size) % num_parts
```
dakinggg, line 153: these two lines can be simplified using `divmod` I believe

dakinggg (commented):
LGTM, will approve after results of manual train and resume test. Would also be good to get @eracah to look since this is such core logic to a lot of our training jobs.

dakinggg (approved):
Thanks so much @coryMosaicML @Skylion007 ! Feel free to merge whenever you are ready.

composer/utils/object_store/oci_object_store.py
coryMosaicML: Thanks! Changed it to use `divmod`.

PR #2760, opened 2023-12-05
Send all metadata keys to update run metadata

jjanezhang:
# Send all metadata keys to update run metadata

This update is to address missing run metadata due to concurrent update run metadata updates.
1. No longer clearing the buffered metadata on flush metadata, sending all historical + new keys on every flush
2. Flushing both wandb and model init time together on `after_load`


composer/loggers/mosaicml_logger.py
```diff
@@ -93,19 +93,21 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> No
...
                 run_url = callback.run_url
                 if run_url is not None:
-                    self._log_metadata({'wandb/run_url': run_url})
                     log.debug(f'Logging WandB run URL to metadata: {run_url}')
+                    after_load_metadata['wandb/run_url'] = run_url
                 else:
                     log.debug('WandB run URL not found, not logging to metadata')
-        self._flush_metadata(force_flush=True)
+        self._log_metadata(after_load_metadata)
+        self._flush_metadata()
```
siriuslee, line 110: I don't think this is required since it's called at the end of `self._log_metadata`

jjanezhang: Just updated, we are going to force flush at the end of after load!

dakinggg (approved):
Please update the PR description to reflect current state of the PR.

jjanezhang:
Waiting on merging this because there are likely to be issues with duplicated run events if we do not delete run event keys from update run metadata

dakinggg:
@jjanezhang can we close this PR?

PR #2761, opened 2023-12-05
Remove flatten params

mvpatel2000:
# What does this PR do?

Remove flatten params. Was removed from torch

Along for ride:
- Update codeowners to simplify
- Speed up ICL tests
- disable mosaicml logger as it causes flakey tests


vchiley (approved):

tests/fixtures/autouse_fixtures.py
```diff
@@ -119,10 +117,7 @@ def seed_all(rank_zero_seed: int, monkeypatch: pytest.MonkeyPatch):
 @pytest.fixture(autouse=True)
 def mapi_fixture(monkeypatch):
     # Composer auto-adds mosaicml logger when running on platform. Disable logging for tests.
-    future_obj = Future()
-    future_obj.set_result(None)
-    mock_update = lambda *args, **kwargs: future_obj
-    monkeypatch.setattr(mcli, 'update_run_metadata', mock_update)
+    monkeypatch.setenv('MOSAICML_PLATFORM', 'False')
```
dakinggg, line 120: can you confirm for me that the actual mosaic ml logger tests still function as expected?

dakinggg (approved):
Was there any logic to the new max seq lens you set in tests? They seem a bit random

tests/fixtures/autouse_fixtures.py
mvpatel2000: It still passes? The test looks to verify mosaicml logger logged certain items, eg
```assert 'mosaicml/training_progress' in metadata```

and this works fine

mvpatel2000:
> Was there any logic to the new max seq lens you set in tests? They seem a bit random

I picked 64 to start and then raised them as required by failures in test due to too small seq len. It should be principled and we should prune datasets to support a fixed size, but unblocking CI/CD is more important

dskhudia (approved):

dakinggg (approved):

PR #2765, opened 2023-12-07
Fix FSDP _param_init_fn to not reinit parameters multiple times

dakinggg:
# What does this PR do?
FSDP docs say that `_param_init_fn` should "only initialize the parameters/buffers of the module, not those of its submodules. This is to avoid re-initialization." Our `_param_init_fn` does not currently follow this, which means that parameters _do_ get reinitialized multiple times. This mostly works, unless you have FSDP nesting and you try to reinit an already FSDPed module _and_ you are using mixed device initialization. That setting results in a hang, because `FSDP.apply` attempts to gather the parameters.

### DIFF
--- a/composer/trainer/dist_strategy.py
+++ b/composer/trainer/dist_strategy.py
@@ -370,89 +370,152 @@ def prepare_fsdp_module(
             if hasattr(obj, '_fsdp_wrap') and not bool(obj._fsdp_wrap):
                 continue
 
+            # A dictionary of all tied parameter pointers to module names
+            tied_pointers = {}
+
+            # Goes through all modules finding which weights have the same pointers
+            for _, mod in obj.named_modules():
+                for attr in ['weight', 'bias']:
+                    if hasattr(mod, attr):
+                        mod_attr = getattr(mod, attr)
+                        if mod_attr is None:
+                            continue
+
+                        ptr = id(mod_attr)
+                        mod_attr_list = tied_pointers.get(ptr, [])
+                        mod_attr_list.append((mod, attr))
+                        tied_pointers[ptr] = mod_attr_list
+
+                        # ptr = id(mod_attr)
+                        # ptr_attr = (ptr, attr)
+                        # name_list = tied_pointers.get(ptr_attr, [])
+                        # name_list.append(name)
+                        # tied_pointers[ptr_attr] = name_list
+
+            source_mod_to_mod_attr = {}
+            for mod_attr_list in tied_pointers.values():
+                if len(mod_attr_list) == 1:
+                    continue
+                first_mod, first_attr = mod_attr_list[0]
+                source_mod_to_mod_attr[first_mod] = [(target_mod, first_attr, dest_attr) for target_mod, dest_attr in mod_attr_list[1:]]
+
+            # # Creates a dictionary of module names that should be tied together
+            # tied_mod_names = collections.defaultdict(list)
+            # # Creates a set of modules we should not initialize
+            # should_not_init_params = set()
+            # for ptr_attr_type, mod_names in tied_pointers.items():
+            #     # No modules for this pointer are tied
+            #     if len(mod_names) == 1:
+            #         continue
+            #     _, attr_type = ptr_attr_type
+            #     first = next(mod_names.__iter__())
+            #     for elem in mod_names:
+            #         should_not_init_params.add('.'.join([elem, attr_type]))
+            #         tied_mod_names[(first, attr_type)].append(elem)
+            #     # Make sure at least one of the tied parameters is initialized
+            #     should_not_init_params.remove('.'.join([first, attr_type]))
+
+            print(f'[{dist.get_global_rank()}] SHOULD NOT INIT {source_mod_to_mod_attr=}')
+
             def _param_init_fn(module: torch.nn.Module) -> None:
-                print(f'[{dist.get_global_rank()}] _param_init_fn for {type(module)}')
-                # module.to_empty(device=torch.cuda.current_device(), recurse=False)
-                # print(f'[{dist.get_global_rank()}] emptied {type(module)}')
-                # A dictionary of all tied parameter pointers to module names
-                tied_pointers = {}
+                print(f'[{dist.get_global_rank()}] _param_init_fn for {type(module)} {id(module)} {id(module.weight) if hasattr(module, "weight") else "no weight"}')
+                is_meta = any(param.is_meta for param in module.parameters())
+                print(f'[{dist.get_global_rank()}] 1 {is_meta=} current device {torch.cuda.current_device()}')
+                module.to_empty(device=torch.cuda.current_device(), recurse=False)
 
-                # Goes through all modules finding which weights have the same pointers
-                for name, mod in module.named_modules():
-                    # Since FSDP recursively wraps, at parent modules we can encounter already
-                    # wrapped weights, as a result we should skip any modules with `_fsdp_wrapped_module.`
-                    if '_fsdp_wrapped_module' in name:
-                        continue
-                    for attr in ['weight', 'bias']:
-                        if hasattr(mod, attr):
-                            mod_attr = getattr(mod, attr)
-                            if mod_attr is None:
-                                continue
-                            ptr = id(mod_attr)
-                            ptr_attr = (ptr, attr)
-                            name_list = tied_pointers.get(ptr_attr, [])
-                            name_list.append(name)
-                            tied_pointers[ptr_attr] = name_list
-
-                # Creates a dictionary of module names that should be tied together
-                tied_mod_names = collections.defaultdict(list)
-                # Creates a set of modules we should not initialize
-                should_not_init_params = set()
-                for ptr_attr_type, mod_names in tied_pointers.items():
-                    # No modules for this pointer are tied
-                    if len(mod_names) == 1:
-                        continue
-                    _, attr_type = ptr_attr_type
-                    first = next(mod_names.__iter__())
-                    for elem in mod_names:
-                        should_not_init_params.add('.'.join([elem, attr_type]))
-                        tied_mod_names[(first, attr_type)].append(elem)
-                    # Make sure at least one of the tied parameters is initialized
-                    should_not_init_params.remove('.'.join([first, attr_type]))
-
-                meta_safe_apply(module,
-                                lambda t: torch.empty_like(t, device=f'cuda:{torch.cuda.current_device()}'),
-                                should_not_init_params,
-                                module_name='')
-
-                if len(tied_mod_names) > 0:
-                    warnings.warn(('The passed in model appears to have tied weights. In order to '
-                                   'support effective weight tying, the tied modules need to be '
-                                   'in the same FSDP module. If the weights are not properly tied '
-                                   'it can lead to loss spikes. We have tried our best to ensure '
-                                   'the tied weights are in the same FSDP module.'))
-
-                # Redoes weight tying
-                for name_attr, tied_names in tied_mod_names.items():
-                    name, attr = name_attr
-                    src_mod = module.get_submodule(name)
-                    # We need to make sure the source and destination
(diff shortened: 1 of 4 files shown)


tests/trainer/test_fsdp.py
```diff
@@ -13,10 +15,13 @@
 from tests.common import (EmbeddedWeightTiedModel, RandomClassificationDataset, SimpleModel, SimpleWeightTiedModel,
                           world_size)
 
+_INIT_DEVICES = ['cpu', 'meta', 'mixed']
```
vchiley, line 18: Do we also want to do cuda (and mark with gpu)?

tests/trainer/test_fsdp.py
```diff
@@ -31,7 +36,14 @@ def test_fsdp_device_initialization(model: ComposerClassifier, mixed_precision:
...
     num_classes = 10
-    model = model(num_features=num_classes, device=device)
+
+    resolved_device = device
+    if device == 'mixed':
+        if dist.get_local_rank() == 0:
+            resolved_device = 'cpu'
+        else:
+            resolved_device = 'meta'
+    model = model(num_features=num_classes, device=resolved_device)
```
vchiley, line 46: comment
this + `sync_module_states` is how meta init works.
TIL 😄

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
...
+                    if hasattr(obj, 'param_init_fn') and isinstance(obj.param_init_fn, Callable):
+                        obj.param_init_fn(module)
+                    elif hasattr(module, 'reset_parameters') and isinstance(module.reset_parameters, Callable):
+                        module.reset_parameters()
+                    else:
+                        raise ValueError(
+                            f'Object `{obj_name}` does not have a ``param_init_fn`` or a ``reset_parameters`` function. '
+                            'This leaves parameters without initialization. Please add a ``param_init_fn`` or ``reset_parameters`` '
+                            f'to module `{obj_name}`.')
+            else:
```
vchiley, line 433: Just to verify
Even tho its showing up as new code, the entire else block is what we previously had 
ie it doesn't need to be checked?

As a meta note: this will change init if using torch2.1+ vs torch < 2.1, right?

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
             if hasattr(obj, '_fsdp_wrap') and not bool(obj._fsdp_wrap):
                 continue
 
-            def _param_init_fn(module: torch.nn.Module) -> None:
-                # A dictionary of all tied parameter pointers to module names
+            # Rather than verifying these changes with older PyTorch versions, we are fixing forward here
+            if version.parse(torch.__version__) > version.parse('2.1.0'):
+                # A dictionary of all tied parameter pointers to (module, attr) tuples
                 tied_pointers = {}
```
vchiley, line 375: looks like duplicate code. can weight_tying logic be made into a function?

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
...
-                    # Since FSDP recursively wraps, at parent modules we can encounter already
-                    # wrapped weights, as a result we should skip any modules with `_fsdp_wrapped_module.`
-                    if '_fsdp_wrapped_module' in name:
-                        continue
+                for _, mod in obj.named_modules():
                     for attr in ['weight', 'bias']:
                         if hasattr(mod, attr):
                             mod_attr = getattr(mod, attr)
                             if mod_attr is None:
                                 continue
```
vchiley, line 383:
```suggestion
                for mod in obj.modules():
                    for attr, mod_attr in mod.named_parameters(recurse=False):
```

@bcui19 (?; or whoever originally wrote this) can you verify that this is fine?

tests/trainer/test_fsdp.py
dakinggg: I can...do you think that is actually testing something new and interesting?

tests/trainer/test_fsdp.py
dakinggg: *mixed, but yeah haha

composer/trainer/dist_strategy.py
dakinggg: its actually not duplicate code. I'm saving a different thing in the dict for this version. It may be possible to unify to the new implementation, but I wanted to avoid touching the previous code.

composer/trainer/dist_strategy.py
dakinggg: I'm guessing there is a reason we only did weight/bias? @bcui19

composer/trainer/dist_strategy.py
dakinggg: yes, the entire block in the `else` is copied from before

yes, meta init will have different initialization.

tests/trainer/test_fsdp.py
vchiley: not really...
just breadth testing 🤷‍♂️

composer/trainer/dist_strategy.py
vchiley: > yes, the entire block in the else is copied from before

👍 

> yes, meta init will have different initialization.

😞 
ok thats fine.

composer/trainer/dist_strategy.py
bcui19: Unfortunately, I don't remember if there was a particular reason we limited it to 'weight', 'bias'.

dakinggg: Given that, I'd prefer to not to change it, since we know the current version works. okay with the @vchiley ?

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
...
                 continue
 
-            def _param_init_fn(module: torch.nn.Module) -> None:
-                # A dictionary of all tied parameter pointers to module names
+            # Rather than verifying these changes with older PyTorch versions, we are fixing forward here
+            if version.parse(torch.__version__) > version.parse('2.1.0'):
+                # A dictionary of all tied parameter pointers to (module, attr) tuples
                 tied_pointers = {}
 
                 # Goes through all modules finding which weights have the same pointers
```
irenedea, line 377: In what cases would different weights have the same pointers?

dakinggg: It is common in transformer lms to have the input embedding and the output unembedding layers use the same weights. that is the main use i'm aware of, although likely weight tying may be used in other contexts.

tests/trainer/test_fsdp.py
dakinggg: added

tests/trainer/test_fsdp.py
```diff
@@ -63,8 +76,72 @@ def test_fsdp_device_initialization(model: ComposerClassifier, mixed_precision:
...
+@pytest.mark.parametrize(
+    'model,expected_param_inits',
+    [
+        (SimpleModel, 2),  # One call for each of the Linear layers
+        (EmbeddedWeightTiedModel, 3),  # Two calls for each of the SimpleMLP modules, minus one for weight tying
+    ])
+@pytest.mark.parametrize('device', _INIT_DEVICES)
+@world_size(2)
+@pytest.mark.gpu
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('1.13.0'),
```
irenedea, line 88: Should this be skipped if we're using anything less than torch 2.1.0? I'd expect it to fail on earlier versions.
```suggestion
@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.1.0'),
```

tests/trainer/test_fsdp.py
```diff
@@ -42,7 +54,8 @@ def test_fsdp_device_initialization(model: ComposerClassifier, mixed_precision:
         train_dataloader=dataloader,
         fsdp_config={
             'activation_checkpointing_reentrant': reentrant,
-            'mixed_precision': mixed_precision
+            'mixed_precision': mixed_precision,
+            'sync_module_states': True if device == 'mixed' else False,
```
irenedea, line 58: why is this true for mixed?

bcui19 (commented):
Left a minor comment on the usage of `to_empty,` barring resolving of that looks good. Thank you!

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
...
+                    # If we do not have any parameters or buffers on meta device managed by this module directly, we do not need to call the parameter init function.
+                    # It is assumed that whatever process moved the parameters off of meta device initialized them.
+                    # We expect this to occur if we have tied weights, as the second module will already have the weights initialized.
+                    is_meta = any(param.is_meta for param in module.parameters(recurse=False)) or any(
+                        buffer.is_meta for buffer in module.buffers(recurse=False))
+                    if not is_meta:
+                        return
+
+                    # Move all parameters and buffers to the current device
+                    module.to_empty(device=f'cuda:{torch.cuda.current_device()}', recurse=False)
```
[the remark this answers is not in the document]
dakinggg: I don't think so. Here is my logic.

I call `to_empty` with `recurse=False` so that it does _not_ initialize submodules. Above this I check if the module still has params/buffers on `meta` (again with `recurse=False`) before calling `to_empty`, so `to_empty` should not get called on anything that has already been moved off of `meta` device. Let me know what you think.

tests/trainer/test_fsdp.py
dakinggg: for mixed init, the weights need to come from rank 0, rather than initialized on each rank. that is what the `sync_module_states` flag does.

tests/trainer/test_fsdp.py
dakinggg: I _think_ this test should still pass on other torch versions.

composer/trainer/dist_strategy.py
```diff
@@ -369,78 +369,141 @@ def sync_hook(*args):
...
+                        obj.param_init_fn(module)
+                    elif hasattr(module, 'reset_parameters') and isinstance(module.reset_parameters, Callable):
+                        module.reset_parameters()
+                    else:
+                        raise ValueError(
+                            f'Object `{obj_name}` does not have a ``param_init_fn`` or a ``reset_parameters`` function. '
+                            'This leaves parameters without initialization. Please add a ``param_init_fn`` or ``reset_parameters`` '
+                            f'to module `{obj_name}`.')
+            else:
+
```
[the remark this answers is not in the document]
dakinggg: any deprecation would only be at the level of deprecating older torch versions, which is out of scope of this PR I think.

composer/trainer/dist_strategy.py
vchiley: weird but 🤷‍♂️ 

Is it possible to include a warning or something that takes param names that are not 'weight' or 'bias' and say: "Warning: these cannot be tied".

dakinggg: maybe i should just change it...

tests/trainer/test_fsdp.py
dakinggg: thinking about making the change apply to torch 2.0.1 as well

dakinggg: alright, yeah, here is when it changed: https://github.com/pytorch/pytorch/pull/104189#discussion_r1248269410

composer/trainer/dist_strategy.py
dakinggg: I'm going to change it to your version for the new code, will leave the old code untouched.

composer/trainer/dist_strategy.py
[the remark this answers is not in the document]
bcui19: I think this makes sense. Thanks for clarifying!

vchiley (approved):

PR #2775, opened 2023-12-12
Enable system metrics in mosaic mlflow logger

chenmoneygithub:
# What does this PR do?

Enable system metrics in mosaic mlflow logger. The default mode is on.

Please see the screenshot:
Screenshot 2023-12-11 at 5 05 58 PM

The order is a bit strange when there are multiple GPUs, we should fix it on the mlflow side. But it does not affect this PR.

# Before submitting

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -53,7 +53,7 @@ class MLFlowLogger(LoggerDestination):
         synchronous (bool, optional): Whether to log synchronously. If ``True``, Mlflow will log
             synchronously to the MLflow backend. If ``False``, Mlflow will log asynchronously. (default: ``False``)
         log_system_metrics (bool, optional): Whether to log system metrics. If ``True``, Mlflow will
-            log system metrics (CPU/GPU/memory/network usage) during training.
+            log system metrics (CPU/GPU/memory/network usage) during training. (default: ``True``)
     """
 
     def __init__(
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -47,7 +47,7 @@ class MLFlowLogger(LoggerDestination):
             (default: ``10``).
         model_registry_prefix (str, optional): The prefix to use when registering models.
             If registering to Unity Catalog, must be in the format ``{catalog_name}.{schema_name}``.
-            (default: empty string)
+            (default: `''`)
         model_registry_uri (str, optional): The URI of the model registry to use. To register models
             to Unity Catalog, set to ``databricks-uc``. (default: None)
         synchronous (bool, optional): Whether to log synchronously. If ``True``, Mlflow will log
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -51,7 +51,7 @@ class MLFlowLogger(LoggerDestination):
         model_registry_uri (str, optional): The URI of the model registry to use. To register models
             to Unity Catalog, set to ``databricks-uc``. (default: None)
         synchronous (bool, optional): Whether to log synchronously. If ``True``, Mlflow will log
-            synchronously to the MLflow backend. If ``False``, Mlflow will log asynchronously.
+            synchronously to the MLflow backend. If ``False``, Mlflow will log asynchronously. (default: ``False``)
         log_system_metrics (bool, optional): Whether to log system metrics. If ``True``, Mlflow will
             log system metrics (CPU/GPU/memory/network usage) during training.
     """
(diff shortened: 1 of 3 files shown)


tests/loggers/test_mlflow_logger.py
```diff
@@ -22,6 +22,16 @@
 from tests.models.test_hf_model import check_hf_model_equivalence, check_hf_tokenizer_equivalence
 
 
+@pytest.fixture(autouse=True)
```
eracah, line 25: does this have to be autouse=True? we are moving away from autouse=True because it makes debugging really hard. @mvpatel2000 Can you just inject it into the other MLflow tests?

eracah (commented):
Does this work for multinode?

tests/loggers/test_mlflow_logger.py
chenmoneygithub: This should only take effect in this test, but I can inline the code if this is against mosaic style. Let me know which one you prefer, thanks!

chenmoneygithub:
@eracah As discussed offline, multi-node support is super tricky to add to MLflow as a native support, because different platforms have their own communication protocol. While we would be interested to add this support, I am expecting it to take weeks to build a partially-working solution, so we won't be there in the near future...

composer/loggers/mlflow_logger.py
```diff
@@ -50,6 +50,10 @@ class MLFlowLogger(LoggerDestination):
             (default: empty string)
```
mvpatel2000, line 50:
```suggestion
            (default: `''`)
```

composer/loggers/mlflow_logger.py
```diff
@@ -50,6 +50,10 @@ class MLFlowLogger(LoggerDestination):
...
         model_registry_uri (str, optional): The URI of the model registry to use. To register models
             to Unity Catalog, set to ``databricks-uc``. (default: None)
+        synchronous (bool, optional): Whether to log synchronously. If ``True``, Mlflow will log
+            synchronously to the MLflow backend. If ``False``, Mlflow will log asynchronously.
```
mvpatel2000, line 54:
```suggestion
            synchronously to the MLflow backend. If ``False``, Mlflow will log asynchronously. (default: ``False``)
```

composer/loggers/mlflow_logger.py
```diff
@@ -50,6 +50,10 @@ class MLFlowLogger(LoggerDestination):
...
+        log_system_metrics (bool, optional): Whether to log system metrics. If ``True``, Mlflow will
+            log system metrics (CPU/GPU/memory/network usage) during training.
```
mvpatel2000, line 56:
```suggestion
            log system metrics (CPU/GPU/memory/network usage) during training. (default: ``True``)
```

mvpatel2000 (commented):
Minor nits:

composer/loggers/mlflow_logger.py
```diff
@@ -47,9 +47,13 @@ class MLFlowLogger(LoggerDestination):
             (default: ``10``).
         model_registry_prefix (str, optional): The prefix to use when registering models.
             If registering to Unity Catalog, must be in the format ``{catalog_name}.{schema_name}``.
-            (default: empty string)
+            (default: `''`)
         model_registry_uri (str, optional): The URI of the model registry to use. To register models
             to Unity Catalog, set to ``databricks-uc``. (default: None)
+        synchronous (bool, optional): Whether to log synchronously. If ``True``, Mlflow will log
```
eracah, line 53: when would anyone ever want synchronous? I think we should remove this in a future PR

eracah (approved):
LGTM

composer/loggers/mlflow_logger.py
chenmoneygithub: yea it's more like a safety flag, I will make a PR to remove it once we are confident with the functionality.

tests/loggers/test_mlflow_logger.py
```diff
@@ -431,14 +431,17 @@ def test_mlflow_register_uc_error(tmp_path, monkeypatch):
...
 def test_mlflow_logging_works(tmp_path, device):
-    pytest.importorskip('mlflow')
+    mlflow = pytest.importorskip('mlflow')
 
     mlflow_uri = tmp_path / Path('my-test-mlflow-uri')
     experiment_name = 'mlflow_logging_test'
     test_mlflow_logger = MLFlowLogger(
         tracking_uri=mlflow_uri,
         experiment_name=experiment_name,
+        log_system_metrics=True,
```
dakinggg, line 441: this is the default so you don't need to set it, right?

dakinggg (commented):
Could you share a link to an experiment with system metrics I can look at? and/or a list of all the system metrics that are tracked?

tests/loggers/test_mlflow_logger.py
chenmoneygithub: yea, I would prefer to explicitly set the arg so that the test is robust against changes and easier to read.

dakinggg (approved):

PR #2777, opened 2023-12-13
Improve torch memory profiling arguments processing

cli99:
This PR makes the following changes to the memory profiling enablement:
1. unless `torch_prof_memory_filename` is explicitly set to None, memory timeline generation is enabled.
2. when `torch_prof_memory_filename` is enabled, checks all the three flags torch_prof_with_stack, torch_prof_record_shapes, and torch_prof_profile_memory to be true, otherwise raise a value error (unit test is added)


composer/profiler/profiler.py
```diff
@@ -143,6 +143,23 @@ def __init__(
...
+            )
+
+        if (torch_prof_with_stack and torch_prof_record_shapes and
+                torch_prof_profile_memory) and torch_prof_memory_filename is None:
+
+            log.info(
+                f'Memory profiling is enabled but torch_prof_memory_filename is not set. Using {"rank{rank}.{batch}.pt.memory_trace.html"} as the filename for the memory timeline graph.'
+            )
+            torch_prof_memory_filename = 'rank{rank}.{batch}.pt.memory_trace.html'
+
```
j316chuck, line 162: Can we remove this? I think we shouldn't generate the memory timeline graph by default if it's not specified. `torch_prof_profile_memory`

j316chuck (commented):
LGTM: left two nits.


If it's simple, can we also add a unit test that checks different configurations of `torch_prof_folder, torch_prof_filename, torch_prof_remote_file_name, torch_prof_memory` etc. and ensures the correct ValueErrors are raised/not raised

composer/profiler/profiler.py
cli99: done.

cli99:
@j316chuck, added unit test. Only one case triggers a ValueError when using the profiler.

j316chuck (approved):
Thanks!

PR #2788, opened 2023-12-19
add memory snapshot callback

cli99:
This PR adds a callback to capture memory snapshot. A html file with interactive visualization would be generated if enabled. https://github.com/pytorch/pytorch.github.io/blob/site/assets/images/understanding-gpu-memory-1/snapshot.html shows an example visualization.

To enable the callback, in the yaml config file. (https://github.com/mosaicml/llm-foundry/pull/810)

```
callbacks:
    memory_snapshot:
        {
            "skip_batches": 2,
            "interval": 3,
            "folder": "traces",
            "overwrite": true,
            "filename": "rank{rank}.{batch}.pt.trace.memory_snapshot.html",
            "remote_file_name": "oci://bucket_name/{run_name}/torch_memory_snapshots/rank{rank}.{batch}.pt.trace.memory_snapshot.html",
        }
````
Below is an example memory snapshot over three batches of MPT-7B with micro batch size 4, FSDP full shard, on 8XH100 
image

For more details on the memory snapshot, refer to 
- https://pytorch.org/blog/understanding-gpu-memory-1/#capturing-memory-snapshots
- https://pytorch.org/docs/main/torch_cuda_memory.html#generating-a-snapshot

### CODE REVIEW

File: composer/callbacks/memory_snapshot.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+
+log = logging.getLogger(__name__)
+
+__all__ = ['MemorySnapshot']
+
+
+class MemorySnapshot(Callback):
+    """Logs the memory snapshot of the model.
+
+    This callback calls the torch memory snapshot API (see :func:`torch.cuda.memory._snapshot`) to record tensors' memory allocation of the model over a user defined interval, and provides a fine-grained GPU memory visualization for debugging GPU OOMs. Captured memory snapshots will show memory events including allocations, frees and OOMs, along with their stack traces.
```
j316chuck, line 25:
```suggestion
    This callback calls the torch memory snapshot API (see :func:`torch.cuda.memory._snapshot`) to record a model's tensor memory allocation over a user defined interval. This provides a fine-grained GPU memory visualization for debugging GPU OOMs. Captured memory snapshots will show memory events including allocations, frees and OOMs, along with their stack traces.
```

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+        self.overwrite = overwrite
+        self._enabled = True
+        self._start_time = None
+
+    def init(self, state: State, logger: Logger) -> None:
+        # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
+        model_device = next(state.model.parameters()).device
+
+        if model_device.type not in ('cuda', 'meta'):
+            warnings.warn(f'The memory snapshot only works on CUDA devices, but the model is on {model_device.type}.')
```
j316chuck, line 97: I prefer we raise an exception so the user explicitly knows they won't get memory profiles.  Lmk what you  all think cc: @mvpatel2000

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+        if state.timestamp.get(self.interval.unit).value == (self._start_time + self.interval.value):
+            self.export_memory_snapshot(state, logger)
+            self.stop_record_memory_history()
+            self._start_time = None
+            self._enabled = False
+
+    def start_record_memory_history(self) -> None:
+        if not torch.cuda.is_available():
+            log.info('CUDA unavailable. Not recording memory history')
+            return
```
j316chuck, line 127: Can remove if we explicity check if torch cuda is available at init

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+
+        log.info('Stopping snapshot record_memory_history')
+        torch.cuda.memory._record_memory_history()
+
+    def export_memory_snapshot(self, state: State, logger: Logger) -> None:
+        if not torch.cuda.is_available():
+            log.info('CUDA unavailable. Not exporting memory snapshot')
+            return
+
+        if self.filename:
```
j316chuck, line 145: no need for if statement since this variable is specified by default?

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+        if self.batches_left_to_skip > 0:
+            self.batches_left_to_skip -= 1
+            return
+        assert self._start_time is not None
+
+        if state.timestamp.get(self.interval.unit).value == (self._start_time + self.interval.value):
+            self.export_memory_snapshot(state, logger)
+            self.stop_record_memory_history()
+            self._start_time = None
+            self._enabled = False
```
j316chuck, line 122: Does this mean we will only record the cuda memory profile once? If so, can we record this in the docstring. 

A design decision to be made is should we incorporate this logic as a callback or into the existing torch profiler?

I'm okay with having it as a callback because it takes less memory/time but it is a bit confusing to the user that some memory traces are stored as profile flags and others are stored as callbacks.

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+            return
+
+        if self.batches_left_to_skip > 0:
+            self.batches_left_to_skip -= 1
+            return
+        assert self._start_time is not None
+
+        if state.timestamp.get(self.interval.unit).value == (self._start_time + self.interval.value):
+            self.export_memory_snapshot(state, logger)
+            self.stop_record_memory_history()
```
j316chuck, line 120: Just checking we are expected to `export` -> `stop_record` and not the other way around?

composer/callbacks/memory_snapshot.py
```diff
@@ -0,0 +1,161 @@
...
+            log.info('CUDA unavailable. Not recording memory history')
+            return
+
+        log.info('Starting snapshot record_memory_history')
+        torch.cuda.memory._record_memory_history(max_entries=self.max_entries)
+
+    def stop_record_memory_history(self) -> None:
+        if not torch.cuda.is_available():
+            log.info('CUDA unavailable. Not recording memory history')
+            return
```
j316chuck, line 135: Can also remove if we explicity check if torch cuda is available at init

tests/callbacks/test_memory_snapshot.py
```diff
@@ -0,0 +1,46 @@
...
+class FileUploaderTracker(LoggerDestination):
+
+    def __init__(self) -> None:
+        self.uploaded_files = []
+
+    def upload_file(self, state: State, remote_file_name: str, file_path: pathlib.Path, *, overwrite: bool):
+        del state, overwrite  # unused
+        self.uploaded_files.append((remote_file_name, file_path))
+
+
```
j316chuck, line 25: Can we add a test that checks that on CPU the memory callback raises an Error

tests/callbacks/test_memory_snapshot.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.gpu
+@pytest.mark.parametrize('interval', ['1ba', '3ba'])
+def test_memory_snapshot(interval: str, tmp_path: pathlib.Path):
```
j316chuck, line 28: Do we need tmp_path here?

j316chuck (commented):
Thanks for adding this Cheng! This PR looks good to me barring some nits. 

Additionally, here are two higher level use cases to consider in this PR: 
- One, should we add this callback in the profiler instead so we support having N memory traces instead of just 1 memory trace?
- Two, is it possible to add this memory trace when a run oom's? 

Here's the ticket: 
https://databricks.atlassian.net/browse/GRT-2231

composer/callbacks/memory_snapshot.py
cli99: I saw the check here https://github.com/mosaicml/composer/blob/dev/composer/callbacks/memory_monitor.py#L89-L94 and took it.

composer/callbacks/memory_snapshot.py
cli99: yes, export -> stop

cli99:
> Thanks for adding this Cheng! Overall, this PR looks good to me barring some style suggestions.
> 
> Additionally, here are two higher level use cases to consider for this PR:
> 
> * First, should we add this callback in the profiler instead so we support having N memory traces instead of just 1 memory trace? I can see pros and cons for both so just wanted to make sure we are aligned on this decision cc: @mvpatel2000
> * Second, is it possible to add this memory trace when a run oom's?
> 
> Here's the original ticket that this PR addresses that has some ideas on the second point: https://databricks.atlassian.net/browse/GRT-2231

My thoughts on the two questions
1) Composer profiler aligns with what torch profiler supports and takes the schedule parameters; memory snapshot call back does not reply on any torch profiler apis (nor its schedule) and shall live by itself.
2) Generating memory snapshot right before OOM is a separate feature. It's going to generate a snapshot at a single timestamp (not over a time interval) and the best visualization i think would be a flame graph. Two options here: a) add to the current memory monitor callback(in its init function) b) implement as a separate callback. What do you think? @j316chuck @mvpatel2000

composer/callbacks/memory_snapshot.py
mvpatel2000: I think it's fine to emit warning -- there are weird scenarios where user manually moves around model and callback is still useful

mvpatel2000 (approved):
LGTM. Can you please provide a manual test and a screenshot just to show it works on a real run?

1. I agree this should be separate from profiler. It's convenient to have modularity, especially because profiler slows things down a lot
2. I think merging OOM capture into this callback is reasonable given it's the same torch system. I suggest follow-on PR since this is basically done

composer/callbacks/memory_snapshot.py
j316chuck: Sounds good w/me

j316chuck:
Agreed with 1 and 2 from Mihir. Up to you whether or not you want to create follow up PR or not.

j316chuck (approved):

PR #2790, opened 2023-12-20
default to no torch profiler memory timeline

cli99:
This PR changes the default behavior from enabling memorytimeline with a default name to not enabling  memory timeline generation if users do not specify torch_profiler_memory_filename. Thus users don't get a ValueError the first time they run the profiler without the three required flags on (torch_prof_with_stack, torch_prof_record_shapes, and torch_prof_profile_memory).


j316chuck (approved):
thanks for fixing this

PR #2800, opened 2023-12-26
Add custom stopping criteria to ICL generate tasks

bmosaicml:
# What does this PR do?
Modeled after EleutherAI implementation of generate tasks, we enable custom stopping criteria to be implemented.

This relies on five separate changes: 

1. Implement a multi token stopping criteria modeled after [this](https://github.com/EleutherAI/lm-evaluation-harness/blob/af74a93d1abd9a35fee9d628c2637ae922a82541/lm_eval/utils.py#L625)
2. Modify QA tasks' collate fn to add stopping criteria to generate kwargs as well as pass them along in the batch as raw strings
3. Modify the QA metric to post-hoc apply the stopping criteria to any outputs it wasn't applied to. This is necessary because stopping criteria don't stop until all outputs in a device batch reach the stopping criteria, so by definition all but one output per batch will need to have it manually applied post-hoc.
4. Modify the split_batch fn to properly handle list values that need to be broadcast (the string literal stopping criteria)
5. Make normalization optional, since GSM8K doesn't use the standard normalization.

# Before submitting


composer/datasets/utils.py
```diff
@@ -166,3 +170,63 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
         else:
             dataset.transform = transforms.Compose([dataset.transform, transform])
             log.warning(transform_added_logstring)
+
+
+try:
```
bmosaicml, line 175: @dakinggg  what is the proper way to handle subclassing a class from `transformers`? Is it possible to do a conditional import, or does code like this make transformers a required dependency?

bmosaicml: @abhi-mosaic  do you know how to handle this?

bmosaicml: @mosaicml/composer-team-eng any thoughts on how to handle subclassing from a class defined in a conditional import?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -339,7 +349,7 @@ def split_batch(self, batch: Any, microbatch_size: int):
                 raise ValueError(f'Unexpected key {k}')
         num_chunks = len(chunked['input_ids'])
         for k, v in batch.items():
-            if isinstance(v, (int, float, str, bool, dict)):
+            if k in no_split:
```
abhi-mosaic, line 352: Just double checking, we never expect to get other keys that should not be split?

composer/datasets/utils.py
```diff
@@ -15,6 +15,9 @@
 from torchvision.datasets import VisionDataset
 
 from composer.core import Batch
+import transformers
+if TYPE_CHECKING:
+    import transformers
```
abhi-mosaic, line 20: Hmm this looks strange, if we need to import `transformers` anyways why do we need to gate on `TYPE_CHECKING`?

composer/datasets/utils.py
```diff
@@ -166,3 +169,54 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
...
+        self.done_tracker = [False] * batch_size
+        self.sequence = sequence
+        self.sequence_ids = tokenizer.encode(sequence, add_special_tokens=False)
+        # we look back for 2 more tokens than it takes to encode our stop sequence
+        # because tokenizers suck, and a model might generate `['\n', '\n']` but our `sequence` is `['\n\n']`
+        # and we don't want to mistakenly not stop a generation because our
+        # (string) stop sequence was output in a different tokenization
+
+        # NOTE: there is a minor danger that this will end up looking back 2 tokens into the past, into the inputs to the model,
+        # and stopping generation immediately as a result. With only 2 extra tokens of lookback, this risk is minimized
```
abhi-mosaic, line 194: this basically translates to "make sure your inputs dont end with a stop sequence", right?

abhi-mosaic (commented):
Couple small questions, logic seems right to me. Would love to see some unit tests.

composer/datasets/utils.py
dakinggg: I think you'll need to conditionally define the superclass. that is, in the except case, set the class to `None`, and then error properly on construction if the dependency is missing.

composer/datasets/utils.py
bmosaicml: This comment is taken from the Eleuther implementation. I believe that yes this would cause errors if the input ends in a stop sequence. Ideally it would be good to know the input length, but I think that information is not available at the time of `__call__`

composer/datasets/utils.py
bmosaicml: Modified this to remove the import. The real solution is to import of type checking AND try to import it when constructing the class/function but default to setting the function to None if we can't import.

bmosaicml:
> Couple small questions, logic seems right to me. Would love to see some unit tests.

I'll try and mock the `__call__` function somehow in a unit test!

composer/datasets/in_context_learning_evaluation.py
bmosaicml: The other keys would be handled in the above logic where we call _split_list and _default_split_batch. It's necessary that the above code correctly categorize all the keys so that they are either list, default, or no split.

maxisawesome:
LGTM. Importing a function only to check if it exists is a little funny to me, but I understand the reasoning based on past discussions and think it's a fine solution!

maxisawesome (approved):

composer/datasets/utils.py
```diff
@@ -16,6 +16,9 @@
 
 from composer.core import Batch
 
+if TYPE_CHECKING:
```
dakinggg, line 19: I don't think you need this for anything?

composer/datasets/utils.py
```diff
@@ -166,3 +169,61 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
         else:
             dataset.transform = transforms.Compose([dataset.transform, transform])
             log.warning(transform_added_logstring)
+
+
+try:
+    import transformers  # pyright: ignore [reportGeneralTypeIssues]
```
dakinggg, line 175: pyright complains about an import?

composer/datasets/utils.py
```diff
@@ -166,3 +169,61 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
...
+            self.tokenizer = tokenizer
+
+        def __call__(self, input_ids, scores: Optional[torch.FloatTensor] = None, **kwargs) -> bool:
+            # For efficiency, we compare the last n tokens where n is the number of tokens in the stop_sequence
+            lookback_ids_batch = input_ids[:, :][:, -self.stop_sequence_id_len:]
+
+            lookback_tokens_batch = self.tokenizer.batch_decode(lookback_ids_batch)
+            for i, done in enumerate(self.done_tracker):
+                if i >= len(lookback_tokens_batch):
+                    # the last batch of a dataset may b smaller than `batch_size`
```
dakinggg, line 209:
```suggestion
                    # The last batch of a dataset may be smaller than `batch_size`
```

composer/datasets/utils.py
```diff
@@ -166,3 +169,61 @@ def add_vision_dataset_transform(dataset: VisionDataset, transform: Callable, is
...
+                    # automatically set those indices in the done_tracker to True
```
dakinggg, line 210:
```suggestion
                    # Automatically set those indices in the done_tracker to True
```

dakinggg (approved):

PR #2802, opened 2023-12-29
[MLFlowObjectStore] [1/2] Base implementation for MLFlowObjectStore

jerrychen109:
# What does this PR do?
`ObjectStore` implementation using Databricks-managed MLFlow artifact storage.

To keep the PR size manageable, this PR only includes the implementation for `MLFlowObjectStore` and unit tests. A follow-up will include the necessary changes to enable end-to-end use of the class with the `Trainer` and `MLFlowLogger`.

Follow-up PR: #2810 

### Initialization with placeholders:
With `MLFLOW_EXPERIMENT_NAME=whattheheck` (ignore the name lmao)
```
>>> from composer.utils import MLFlowObjectStore
>>> store = MLFlowObjectStore('databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/')
>>> print (store.experiment_id, store.run_id)
2932562896319786 f8b3663cfca445c4a734f1e4665973fa
```
Created new run under `whattheheck` experiment
image
image

### Initialization from existing run:
```
>>> store = MLFlowObjectStore('databricks/mlflow-tracking/2932562896319786/c0320ca5207d4e8398efd217b7f16ad2/artifacts/')
>>> print (store.experiment_id, store.run_id)
2932562896319786 c0320ca5207d4e8398efd217b7f16ad2
```

### Uploading
```
>>> store.upload_object('hello.txt', 'hello.txt')
>>> store.upload_object('folder/hello.txt', 'hello.txt')
>>> store.upload_object('folder/different-base-name.txt', 'hello.txt')
>>> store.list_objects()
['hello.txt', 'folder/different-base-name.txt', 'folder/hello.txt']
```

### Get size
```
>>> store.get_object_size('hello.txt')
6
```

### Using absolute DBFS path instead of relative artifact path:
```
>>> dbfs_path = store.get_dbfs_path('hello.txt')
>>> print(dbfs_path)
databricks/mlflow-tracking/2932562896319786/f8b3663cfca445c4a734f1e4665973fa/artifacts/hello.txt
>>> store.get_object_size(dbfs_path)
6
```

### Downloading:
```
>>> store.download_object('hello.txt', 'local-hello.txt')
Downloading artifacts: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  2.04it/s]
```
image

Support checkpointing and other object storage using Databricks MLFlow

# Before submitting


composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,371 @@
...
+
+__all__ = ['MLFlowObjectStore']
+
+MLFLOW_DATABRICKS_TRACKING_URI = 'databricks'
+MLFLOW_DBFS_PATH_PREFIX = 'databricks/mlflow-tracking/'
+
+DEFAULT_MLFLOW_EXPERIMENT_NAME = 'mlflow-object-store'
+
+PLACEHOLDER_EXPERIMENT_ID = 'EXPERIMENT_ID'
+PLACEHOLDER_RUN_ID = 'RUN_ID'
```
jerrychen109, line 33: @dakinggg I'm not particularly attached to what the placeholders should be, but the original plan to use format strings such as `{mlflow_experiment_id}` and `{mlflow_run_id}` doesn't quite work because `str.format(...)` will throw a `KeyError` if not all format arguments are provided.

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+
+__all__ = ['MLFlowObjectStore']
+
+MLFLOW_DATABRICKS_TRACKING_URI = 'databricks'
+MLFLOW_DBFS_PATH_PREFIX = 'databricks/mlflow-tracking/'
+
+DEFAULT_MLFLOW_EXPERIMENT_NAME = 'mlflow-object-store'
+
+PLACEHOLDER_EXPERIMENT_ID = 'MLFLOW_EXPERIMENT_ID'
+PLACEHOLDER_RUN_ID = 'MLFLOW_RUN_ID'
```
jerrychen109, line 26: @dakinggg I'll see if I can get the format strings (i.e., `{mlflow_experiment_id}` and `{mlflow_run_id}`) working in the follow-up PR, but I'm finding it pretty tricky to do without needing to rewrite  a lot of the formatting code in `Trainer` and `CheckpointSaver`

jerrychen109, line 26: These will be format strings in the follow-up PR

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+        mlflow.environment_variables.MLFLOW_MULTIPART_UPLOAD_CHUNK_SIZE.set(multipart_upload_chunk_size)
+
+        experiment_id, run_id, _ = MLFlowObjectStore.parse_dbfs_path(path)
+        if experiment_id == PLACEHOLDER_EXPERIMENT_ID:
+            experiment_id = None
+        if run_id == PLACEHOLDER_RUN_ID:
+            run_id = None
+
+        # Construct the `experiment_id` and `run_id` depending on whether format placeholders were provided.
+        if not dist.is_initialized() or dist.get_global_rank() == 0:
```
dakinggg, line 140: do you need the `dist.is_initialized` check for some reason?

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+            experiment_id, run_id = self._init_run_info(experiment_id, run_id)
+
+        if dist.is_initialized():
```
dakinggg, line 143: dont think you need the initialized check here either

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                # If no `run_id` is provided, create a new run in the provided experiment.
+                run = self._mlflow_client.create_run(experiment_id)
+                run_id = run.info.run_id
+
+                log.debug(f'MLFlowObjectStore using new MLFlow run (run_id={run_id}) '
+                          f'for provided experiment (experiment_id={experiment_id})')
+
+        if experiment_id is None or run_id is None:
+            raise ValueError('MLFlowObjectStore failed to initialize experiment and run ID.')
```
dakinggg, line 202: when would this happen? this is not a very actionable error as is

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+
+        if experiment_id is None:
+            if run_id is not None:
+                raise ValueError('A `run_id` cannot be provided without a valid `experiment_id`.')
+
+            active_run = mlflow.active_run()
+            if active_run is not None:
+                experiment_id = active_run.info.experiment_id
+                run_id = active_run.info.run_id
+                log.debug(f'MLFlowObjectStore using active MLFlow run (run_id={run_id})')
```
dakinggg, line 166:
```suggestion
                log.debug(f'MLFlowObjectStore using active MLFlow run {run_id=}')
```

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                # If no active run exists, create a new run for the default experiment.
+                experiment_name = os.getenv(mlflow.environment_variables.MLFLOW_EXPERIMENT_NAME.name,
```
dakinggg, line 169: im guessing there is some shared logic in the init with the mlflow logger init. anything we can pull out into a helper?

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+
+                experiment = self._mlflow_client.get_experiment_by_name(experiment_name)
+                if experiment is not None:
+                    experiment_id = experiment.experiment_id
+                else:
+                    experiment_id = self._mlflow_client.create_experiment(experiment_name)
+
+                run_id = self._mlflow_client.create_run(experiment_id).info.run_id
+
+                log.debug(f'MLFlowObjectStore using a new MLFlow run (run_id={run_id}) '
```
dakinggg, line 180:
```suggestion
                log.debug(f'MLFlowObjectStore using a new MLFlow run {run_id=} '
```

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+                          f'for new experiment "{experiment_name}" (experiment_id={experiment_id})')
```
dakinggg, line 181:
```suggestion
                          f'for new experiment "{experiment_name}" {experiment_id=}')
```

dakinggg: same comment elsewhere

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+                          f'for provided experiment (experiment_id={experiment_id})')
+
+        if experiment_id is None or run_id is None:
+            raise ValueError('MLFlowObjectStore failed to initialize experiment and run ID.')
+
+        return experiment_id, run_id
+
+    @staticmethod
+    def parse_dbfs_path(path: str) -> Tuple[str, str, str]:
+        """Parses a DBFS path to extract the MLFlow experiment ID, run ID, and relative artifact path.
```
dakinggg, line 208:
```suggestion
        """Parses a DBFS path to extract the MLflow experiment ID, run ID, and relative artifact path.
```

MLflow is supposed to have lower case f everywhere (except the class names cause that requires deprecation)

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+
+        # Extract relative path from DBFS path.
+        artifact_path = self.get_artifact_path(object_name)
+        artifact_base_name = os.path.basename(artifact_path)
+        artifact_dir = os.path.dirname(artifact_path)
+
+        # Since MLFlow doesn't support uploading artifacts with a different base name than the local file,
+        # create a temporary symlink to the local file with the same base name as the desired artifact name.
+        filename = os.path.abspath(filename)
+        tmp_dir = f'/tmp/{uuid.uuid4()}'
```
dakinggg, line 278: should use pythons built in functionality for creating temporary directories

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+        if artifact_info is None:
+            raise FileNotFoundError(f'Object {self.get_dbfs_path(artifact_path)} not found')
+
+        filename = os.path.abspath(filename)
+        if os.path.exists(filename) and not overwrite:
+            raise FileExistsError(f'The file at {filename} already exists and overwrite is set to False.')
+
+        # MLFLow doesn't support downloading artifacts directly to a specified filename, so instead
+        # download to a temporary directory and then move the file to the desired location.
+        tmp_dir = f'/tmp/{uuid.uuid4()}'
```
dakinggg, line 325: same

dakinggg (commented):
Could you add to the PR description whatever manual testing you have done?

composer/utils/object_store/mlflow_object_store.py
jerrychen109: It was necessary for manual testing (i.e. just initializing an `MLflowObjectStore` directly). Without it, it would error out on the `dist` functions. More generally, I don't think the object store should necessarily assume it's being used in a distributed context from a modular design perspective.

composer/utils/object_store/mlflow_object_store.py
jerrychen109: See note above

composer/utils/object_store/mlflow_object_store.py
dakinggg: the dist functions are all supposed to be usable in a non dist environment, so nothing should be erroring out. e.g. see the logic here https://github.com/mosaicml/composer/blob/2aa50e7741a077ff21f5743934fbcf4b755d441e/composer/utils/dist.py#L346-L376

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+            from mlflow import MlflowClient
+        except ImportError as e:
+            raise MissingConditionalImportError('mlflow', conda_package='mlflow>=2.9.2,<3.0') from e
+
+        try:
+            from databricks.sdk import WorkspaceClient
+        except ImportError as e:
+            raise MissingConditionalImportError('databricks', conda_package='databricks-sdk>=0.15.0,<1.0') from e
+
+        tracking_uri = os.getenv('MLFLOW_TRACKING_URI', MLFLOW_DATABRICKS_TRACKING_URI)
```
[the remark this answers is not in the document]
jerrychen109: only reason being I didn't realize I could. thanks for pointing it out!

composer/utils/object_store/mlflow_object_store.py
jerrychen109: Hmm let me try it again without the check. I was definitely running into a problem that adding the check seemed to solve, but maybe it was symptomatic of a different issue since this was when we were still playing around with all the different ways to broadcast the experiment/run IDs

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+from composer.utils import dist
+from composer.utils.import_helpers import MissingConditionalImportError
+from composer.utils.object_store.object_store import ObjectStore, ObjectStoreTransientError
+
+__all__ = ['MLFlowObjectStore']
+
+MLFLOW_DATABRICKS_TRACKING_URI = 'databricks'
+MLFLOW_DBFS_PATH_PREFIX = 'databricks/mlflow-tracking/'
+
+DEFAULT_MLFLOW_EXPERIMENT_NAME = 'mlflow-object-store'
```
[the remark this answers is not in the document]
jerrychen109: I was thinking that if the object store is being used in isolation, it might be useful to distinguish it under a default experiment with a name dedicated to just being an object store, but that was on a bit of my own whim. More than happy to keep it consistent with the logger

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+
+class MLFlowObjectStore(ObjectStore):
+    """Utility class for uploading and downloading artifacts from MLFlow.
+
+    It can be initializd for an existing run, a new run in an existing experiment, the active run used by the `mlflow`
+    module, or a new run in a new experiment. See the documentation for ``path`` for more details.
+
+    .. note::
+        At this time, only Databricks-managed MLFlow with a 'databricks' tracking URI is supported.
+        Using this object store requires setting `DATABRICKS_HOST` and `DATABRICKS_TOKEN`
```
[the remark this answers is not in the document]
jerrychen109: Yup - will just link to Databricks authentication docs https://databricks-sdk-py.readthedocs.io/en/latest/authentication.html#databricks-native-authentication

composer/utils/object_store/mlflow_object_store.py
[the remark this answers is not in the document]
dakinggg: i don't have a strong opinion...happy to go with whatever you think is best. most people will not be using the default experiment name anyway.

composer/utils/object_store/uc_object_store.py
```diff
@@ -167,7 +167,7 @@ def download_object(self,
         try:
             from databricks.sdk.core import DatabricksError
             try:
-                with self.client.files.download(self._get_object_path(object_name)).contents as resp:
+                with self.client.files.download(self._get_object_path(object_name)).contents as resp:  # pyright: ignore
```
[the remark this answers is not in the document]
jerrychen109: Not sure why pyright didn't complain before, but `contents` is typed as `BinaryIO | None`, which is invalid as a context manager:
```
/Users/jerry.chen/mosaicml/composer/composer/utils/object_store/uc_object_store.py:170:22 - error: Object of type "None" cannot be used with "with" (reportOptionalContextManager)
```

Spent some time trying to dig through the databricks sdk source code and theoretically this should never actually be `None`:
* `contents` is the return value of `ApiClient.do()`: https://github.com/databricks/databricks-sdk-py/blob/193c3fbc255262de506b9bb927c87392c2c26720/databricks/sdk/service/files.py#L534
* `ApiClient.do()` should either return a value or throw an exception: https://github.com/databricks/databricks-sdk-py/blob/193c3fbc255262de506b9bb927c87392c2c26720/databricks/sdk/core.py#L1130-L1138

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -0,0 +1,392 @@
...
+            INVALID_STATE,
+            TEMPORARILY_UNAVAILABLE,
+            DEADLINE_EXCEEDED,
+        ]
+    ]
+    retryable_client_codes = [ErrorCode.Name(code) for code in [ABORTED, REQUEST_LIMIT_EXCEEDED, RESOURCE_EXHAUSTED]]
+    not_found_codes = [ErrorCode.Name(code) for code in [RESOURCE_DOES_NOT_EXIST, NOT_FOUND, ENDPOINT_NOT_FOUND]]
+
+    if isinstance(e, MlflowException):
+        error_code = e.error_code  # pyright: ignore
```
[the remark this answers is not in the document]
jerrychen109: I think this might actually be a pyright bug since it should be able to infer the type correctly from `if isinstance(e, MlflowException)` (VSCode/pylance infers correctly) but instead I get `error: Cannot access member "error_code" for type "Exception"`

I tried both `assert isinstance(e, MlflowException)` and `assert type(e) is MlflowException`, but pyright won't recognize it 😢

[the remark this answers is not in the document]
dakinggg: what about `assert hasattr(e, 'error_code')`? if that doesn't work, oh well, thanks for trying

composer/utils/object_store/uc_object_store.py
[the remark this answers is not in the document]
dakinggg: yeah i think assert is not none is better.

composer/utils/object_store/mlflow_object_store.py
jerrychen109: It should never happen but wanted to catch it just in case since it also plays more nicely with the typing (otherwise pyright will complain about potential `None` values

composer/utils/object_store/mlflow_object_store.py
jerrychen109: Looks like it works without the check 🤔. Will remove and verify in the follow-up PR that everything looks good E2E

composer/utils/object_store/mlflow_object_store.py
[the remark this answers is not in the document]
jerrychen109: no luck...will keep the ignore I guess

composer/utils/object_store/uc_object_store.py
```diff
@@ -231,6 +235,6 @@ def list_objects(self, prefix: Optional[str]) -> List[str]:
                                              path=self._UC_VOLUME_LIST_API_ENDPOINT,
                                              data=data,
                                              headers={'Source': 'mosaicml/composer'})
-            return [f['path'] for f in resp.get('files', []) if not f['is_dir']]
+            return [f['path'] for f in resp.get('files', []) if not f['is_dir']]  # pyright: ignore
```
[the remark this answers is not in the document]
jerrychen109: `resp` is typed `dict | BinaryIO`. Type asserting as `dict` works here

composer/utils/object_store/mlflow_object_store.py
jerrychen109: Actually just remembered why I needed to add the checks - the object store initialization fails in the RUD workers without it since it's running in a different process:
```
Process SpawnProcess-1:
Traceback (most recent call last):6-ac8f-e18cce030076/ift-mpt-7b-vymnpr/checkpoints/ep4-ba204-rank0.pt:   0%|          | 0.00/49.5G [00:00<?, ?iB/s]
  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3/dist-packages/composer/loggers/remote_uploader_downloader.py", line 624, in _upload_worker
    remote_backend = _build_remote_backend(remote_backend_name, backend_kwargs)
  File "/usr/lib/python3/dist-packages/composer/loggers/remote_uploader_downloader.py", line 69, in _build_remote_backend
    return remote_backend_cls(**backend_kwargs)
  File "/usr/lib/python3/dist-packages/composer/utils/object_store/mlflow_object_store.py", line 147, in __init__
    dist.broadcast_object_list(mlflow_info, src=0)
  File "/usr/lib/python3/dist-packages/composer/utils/dist.py", line 371, in broadcast_object_list
    raise RuntimeError(f'The world_size({world_size}) > 1, but the distributed package is not '
RuntimeError: The world_size(8) > 1, but the distributed package is not available or has not been initialized. Please check you have initialized the distributed runtime and that PyTorch has been built with distributed support. If calling this function outside Trainer, please ensure that `composer.utils.dist.initialize_dist` has been called first.
```

dakinggg: ah right. i think that just means the init isn't happening the way you intend it to? That is, the object store constructor is called in the RUD subprocesses, so dist won't be initialized. Didn't we discuss putting the dist sync logic in RUD instead of in the object store constructor itself? i might be misremembering.

jerrychen109: Right - for the RUD workers, I broadcast the updated path in the RUD https://github.com/mosaicml/composer/blob/02a373343c7a0188f2f88b289320b31cf901f550/composer/loggers/remote_uploader_downloader.py#L344-L356

But we still need to do a sync in the object store constructor for cases where the object store is used outside a RUD (e.g., `maybe_create_object_store_from_uri()`). We need some way to gate which instances actually attempt to initialize, which would typically just be rank 0, but that will fail in the workers. I suppose I could pull out the case where both IDs are provided to avoid any sort of `dist` operations in that scenario, but that seems uglier/less readable than just checking for `dist` initialization.
```
if experiment_id is not None and run_id is not None:
  run = self._mlflow_client.get_run(run_id)
  if run.info.experiment_id != experiment_id:
      raise ValueError(
          f'Provided `run_id` {run_id} does not belong to provided experiment {experiment_id}. '
          f'Found experiment {run.info.experiment_id}.')
else:
  if dist.get_global_rank() == 0:
    experiment_id, run_id = init_run_info()

  mlflow_info = [experiment_id, run_id]
  dist.broadcast_object_list(mlflow_info, src=0)
  experiment_id, run_id = mlflow_info
  

```

It also restricts how the object store can be used in subprocesses by imposing the requirement that the IDs must be provided, while checking `dist.is_initialized()` doesn't.

composer/utils/object_store/mlflow_object_store.py
jerrychen109: It's a little awkward since the inits take inherently different inputs, so the logic is actually pretty different. The logger accepts init from run/experiment name and then may or may not create the experiment/run depending on whether ID environment variables are set, while the object store depends entirely on the URI with the experiment/run IDs.

composer/utils/object_store/mlflow_object_store.py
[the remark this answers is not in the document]
jerrychen109: I take it back - `mlflow.get_tracking_uri()` falls back to a local mlflow path in the env var isn't set, and it also makes testing a pain because the `tracking_uri` gets set globally by other tests for `MLFlowLogger`, and there's no easy way to just "unset" the active tracking uri. Going to revert back to just reading the env var and defaulting to Databricks directly

composer/utils/object_store/mlflow_object_store.py
dakinggg: Ah I see. I think I had been thinking we would force `MLFlowObjectStore` construction to always have the experiment and run id. For most of our cases, RUD would handle populating these. Do we have a use case for `MLFlowObjectStore` needing to create stuff on its own or can we just error with it being not supported? I'm also open to keeping your logic, just prefer to avoid extra distributed logic wherever possible.

jerrychen109: We need it for autoresume, which creates the object store with `maybe_create_object_store_from_uri` instead of `maybe_create_remote_uploader_downloader_from_uri`. In particular, the returned object store in the first resumption won't have the IDs populated in its path, so it needs to run the init logic that checks the active run.

dakinggg: Thanks for the explanation! Sorry I have forgotten some of our past conversations 😅 What do you think about adding the same sync logic to `maybe_create_object_store_from_uri` instead of putting it in the object store init?

dakinggg (approved):

PR #2803, opened 2023-12-30
Fix load_ignore_keys with rng

mvpatel2000:
# What does this PR do?

Fix load_ignore_keys with rng. Previously, if rng was ignored, this would result in a missing key error. Additionally adds unit tests for load_ignore_Keys


j316chuck (approved):
Thanks! Probably should have @eracah or @b-chu double check this as well

PR #2810, opened 2024-01-02
[MLFlowObjectStore] [2/2] Support checkpointing with MLFlow

jerrychen109:
# What does this PR do?
Follow-up PR to #2802 that integrates `MLFlowObjectStore` with the `Trainer`.
* Adds `MLFlowObjectStore` as a backend for `RemoteUploaderDownloader`
* Updates `MLFlowLogger` to tag MLFlow runs with the Composer run name so autoresume continues the same MLFlow run.
* Updates save folder and filenames after `Event.INIT` to avoid format placeholders in run logs
* Logs `mlflow_experiment_id` and `mlflow_run_id` as hyperparams

Tested manually with both unsharded and sharded checkpoints:
* Full run with `save_folder` set to `dbfs:/databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/'
* Manually stopping run after a few checkpoints and resuming to verify that autoresume begins from latest checkpoint
* Verifying HF checkpoints uploaded at end of run
* With `load_path` set to the checkpoints path of a previous MLFlow run, loads the weights from that checkpoint

# Before submitting


composer/utils/__init__.py
```diff
@@ -26,68 +26,20 @@
...
+    'GCSObjectStore', 'UCObjectStore', 'MLFlowObjectStore', 'MissingConditionalImportError', 'import_object',
+    'is_model_deepspeed', 'is_model_fsdp', 'is_notebook', 'StringEnum', 'load_checkpoint', 'save_checkpoint',
+    'safe_torch_load', 'ensure_folder_is_empty', 'ensure_folder_has_no_conflicting_files', 'export_for_inference',
+    'export_with_logger', 'quantize_dynamic', 'format_name_with_dist', 'format_name_with_dist_and_time', 'is_tar',
+    'maybe_create_object_store_from_uri', 'maybe_create_remote_uploader_downloader_from_uri', 'parse_uri', 'batch_get',
+    'batch_set', 'configure_excepthook', 'disable_env_report', 'enable_env_report', 'print_env',
+    'get_composer_env_dict', 'retry', 'model_eval_mode', 'get_device', 'is_tpu_installed', 'is_hpu_installed',
+    'ExportFormat', 'Transform', 'export_with_logger', 'extract_hparams', 'convert_nested_dict_to_flat_dict',
+    'convert_flat_dict_to_nested_dict', 'using_torch_2', 'create_interval_scheduler', 'EvalClient', 'LambdaEvalClient',
+    'LocalEvalClient', 'MosaicMLLambdaEvalClient', 'partial_format'
```
jerrychen109, line 44: Auto-formatted by pre-commit yapf

jerrychen109: Due to missing trailing comma - will fix

composer/utils/file_helpers.py
```diff
@@ -166,7 +169,8 @@ def ensure_folder_has_no_conflicting_files(folder_name: Union[str, pathlib.Path]
 
 
 def format_name_with_dist(format_str: str, run_name: str, **extra_format_kwargs: object):  # noqa: D103
-    formatted_str = format_str.format(
+    formatted_str = partial_format(
```
jerrychen109, line 172: Needed to fix the HF checkpointer, which is implemented and initialized in foundry using the original template string `save_folder`, and it formats its checkpoints using `format_name_with_dist` and `format_name_with_dist_and_time`.. There isn't really a clean way to go back and update its save folder since it's in the form of a generic callback once it reaches composer, but the object store supports placeholders in the upload/download paths.

composer/utils/string_helpers.py
```diff
@@ -0,0 +1,26 @@
...
+def partial_format(s, *args, **kwargs):
+    """Format a string with a partial set of arguments.
+
+    Since `str.format()` raises a `KeyError` if a format key is missing from the arguments, this
+    function allows for a partial set of arguments to be provided. Any missing arguments will be
+    left as-is in the string.
+    """
+    result = s
+    done = False
+    while not done:
```
dakinggg, line 16: I don't love the while loop here which in theory might be an infinite loop. is there any other way? maybe we just go back to hardcoded placeholder strings for mlflow instead of the proper {} string formatting?

dakinggg (commented):
The logic looks right to me. I'm only wondering if its better to stick with not {} placeholders for the mlflow object store to avoid all of the partial format changes. I know we've gone back and forth on this, what do you think now after having implemented it this way?

jerrychen109:
I'm personally in favor of keeping the {} placeholders now. The logic won't be that much simpler without them - we still need to replace the mlflow placeholders; each instance of `partial_format` would just turn into two `replace()` calls, which I feel is actually a little messier.

Discussed with @dakinggg over Slack, we decided that the implementation of `partial_format` should be safe and we're not too worried about infinite loop behavior. I restructured the code slightly to make it more clear that if the `try` block ever completes or throws an unexpected exception (not `KeyError` or `IndexError`), the function will return on propagate the exception.

Another benefit of keeping the curly braces + `partial_format` is that it should technically make `format_name_with_dist` and `format_name_with_dist` a bit more robust in the rare situation that someone actually wants to put curly braces in their path but not as one of the supported format args.

composer/utils/string_helpers.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def partial_format(s, *args, **kwargs) -> str:
+    """Format a string with a partial set of arguments.
+
+    Since `str.format()` raises a `KeyError` if a format key is missing from the arguments, this
+    function allows for a partial set of arguments to be provided. Any missing arguments will be
+    left as-is in the string.
+    """
+    while True:
```
dakinggg, line 14: can we just put a "max iters" here so it can't be infinite? can be something pretty large.

composer/utils/string_helpers.py
```diff
@@ -0,0 +1,22 @@
...
+
+def partial_format(s, *args, **kwargs) -> str:
+    """Format a string with a partial set of arguments.
+
+    Since `str.format()` raises a `KeyError` if a format key is missing from the arguments, this
+    function allows for a partial set of arguments to be provided. Any missing arguments will be
+    left as-is in the string.
+    """
+    max_iters = 10_000  # Just in case we get stuck in a loop somehow.
+    while max_iters:
```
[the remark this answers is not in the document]
jerrychen109: oops will fix

composer/trainer/trainer.py
```diff
@@ -1158,6 +1164,36 @@ def __init__(
...
+
+            if mlflow_logger is not None:
+                mlflow_experiment_id = mlflow_logger._experiment_id
+                mlflow_run_id = mlflow_logger._run_id
+
+                # Log the experiment and run IDs as hyperparameters.
+                self.logger.log_hyperparameters({
+                    'mlflow_experiment_id': mlflow_experiment_id,
+                    'mlflow_run_id': mlflow_run_id
+                })
```
mvpatel2000, line 1184: This should just be done in side the MLFlow logger (which is a callback) on Event.init

composer/trainer/trainer.py
```diff
@@ -1085,6 +1086,11 @@ def __init__(
             mosaicml_logger = MosaicMLLogger()
             loggers.append(mosaicml_logger)
 
+        # Remote Uploader Downloader
+        # Keep the ``RemoteUploaderDownloader`` below client-provided loggers so the loggers init callbacks run before
+        # the ``RemoteUploaderDownloader`` init. This is necessary to use an ``MLFlowObjectStore`` to log objects to a
+        # run managed by an ``MLFlowLogger``, as the ``MLFlowObjectStore`` relies on the ``MLFlowLogger`` to initialize
+        # the active MLFlow run.
```
mvpatel2000, line 1093: I see comment but no code changes?

composer/trainer/trainer.py
jerrychen109: If there are multiple loggers, wouldn't we want to wait until post INIT before logging hyperparameters to make sure all loggers are initialized? If we did it inside `MLFlowLogger.init()`, then e.g., the `MosaicMLLogger` might not be initialized, right?

mvpatel2000: hm... I see. Can you do `Event.AFTER_LOAD`? Composer pulling out specific callbacks and having them operate is a big anti-pattern

composer/trainer/trainer.py
jerrychen109: Wanted to document that there's an ugly ordering dependency that isn't obvious but needs to be maintained for MLFlowLogger/MLFlowObjectStore to work correctly

composer/utils/string_helpers.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    max_iters = 10_000  # Just in case we get stuck in a loop somehow.
+    for _ in range(max_iters):
+        try:
+            return s.format(*args, **kwargs)
+        except IndexError as e:  # Missing positional arg
+            args += ('{}',)
+        except KeyError as e:  # Missing keyword arg
+            key = e.args[0]
+            kwargs[key] = '{' + key + '}'
```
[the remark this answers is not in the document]
jerrychen109: Don't think so - @dakinggg also spent like 30 min looking into ways to do this lol

`str.format` just throws an exception on the first missing format arg it finds, but the exception only contains info about that individual arg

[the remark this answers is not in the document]
dakinggg: Yeah I tried for a while...couldn't come up with anything that didn't involve reimplementing str.format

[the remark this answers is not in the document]
dakinggg: Certainly open to any suggestions :)

composer/trainer/trainer.py
```diff
@@ -1158,6 +1164,36 @@ def __init__(
...
+                # populate them now.
+                mlflow_format_kwargs = {
+                    MLFLOW_EXPERIMENT_ID_FORMAT_KEY: mlflow_experiment_id,
+                    MLFLOW_RUN_ID_FORMAT_KEY: mlflow_run_id
+                }
+
+                save_folder = partial_format(save_folder, **mlflow_format_kwargs)
+                if latest_remote_file_name is not None:
+                    latest_remote_file_name = partial_format(latest_remote_file_name, **mlflow_format_kwargs)
+
```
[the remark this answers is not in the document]
jerrychen109: `save_folder` for MLflow will be something like `dbfs:/databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/`, where `{mlflow_experiment_id}` and `{mlflow_run_id}` will be replaced/formatted at runtime.

When the actual object stores are created through `maybe_create_object_store` and `maybe_create_remote_uploader_downloader`, the object stores will sync to the experiment/run initialized by the `MLFlowLogger`. However, from the perspectives of `Trainer` and `CheckpointSaver`, which is constructed pre INIT, `save_folder` is still the original value with the placeholders. The updates to `save_folder` both here and in `CheckpointSaver` are primarily UX improvements, as logging messages would otherwise continue to use the placeholders rather than the actual paths being used by the underlying object stores.

Trainer autoresume example:
```
2024-01-01 22:39:44,635: rank7[785][MainThread]: DEBUG: composer.trainer.trainer: Read path databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/checkpoints/ift-mpt-7b-zomexk/checkpoints/ep0-ba2 from symlink file
2024-01-01 22:39:44,636: rank7[785][MainThread]: INFO: composer.trainer.trainer: Autoresuming training from checkpoint
```

CheckpointSaver example:
```
024-01-01 22:42:42,820: rank7[785][MainThread]: DEBUG: composer.utils.checkpoint: Saving sharded checkpoints to databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/checkpoints/ift-mpt-7b-zomexk/checkpoints/ep0-ba4/__7_0.distcp...
2024-01-01 22:42:57,619: rank7[785][MainThread]: DEBUG: composer.callbacks.checkpoint_saver: Checkpoint locally saved to databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/checkpoints/ift-mpt-7b-zomexk/checkpoints/ep0-ba4/__7_0.distcp
2024-01-01 22:42:57,619: rank7[785][MainThread]: DEBUG: composer.callbacks.checkpoint_saver: Uploading checkpoint to databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/checkpoints/ift-mpt-7b-zomexk/checkpoints/ep0-ba4/__7_0.distcp
2024-01-01 22:43:05,797: rank7[785][MainThread]: DEBUG: composer.callbacks.checkpoint_saver: Creating symlink file /tmp/tmpyblyc0lg/latest.symlink -> databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/artifacts/checkpoints/ift-mpt-7b-zomexk/checkpoints/ep0-ba4
```

composer/trainer/trainer.py
jerrychen109: I think we can do the hyperparameter logging in `Event.AFTER_LOAD`. However, see my answer below for why we need to update the `save_folder` with the MLflow run information, so we still need some way to pull that info out inside `Trainer`

composer/trainer/trainer.py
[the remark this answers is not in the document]
mvpatel2000: Can we use `self._checkpoint_saver.save_folder`

composer/trainer/trainer.py
mvpatel2000: Let's still move this log line though?

composer/trainer/trainer.py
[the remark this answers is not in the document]
jerrychen109: Unfortunately cannot - `CheckpointSaver` is initialized with a different value (`parsed_save_folder`) that only contains the save folder path but not the rest of the remote URI information.

mvpatel2000 (approved):

PR #2812, opened 2024-01-02
Require sync module states with HSDP

mvpatel2000:
# What does this PR do?

Require sync module states with HSDP or otherwise different replicas will have different initialization points


j316chuck (approved):
LGTM! P2 but would be nice to add a test

mvpatel2000:
> LGTM! P2 but would be nice to add a test

since we do 2 GPUs, hard to unit test HSDP :(. On todo list to expand to 4 for a limited set of tests but its a big lift

PR #2814, opened 2024-01-04
Bump torch version -- DO NOT RELEASE

mvpatel2000:
# What does this PR do?

Bump torch version to use nightly. Do not release until torch 2.2 officially comes out.


dakinggg (approved):
So if we do another release before torch 2.2, we will manually change the torch version back on the release branch, right?

mvpatel2000:
> So if we do another release before torch 2.2, we will manually change the torch version back on the release branch, right?

Or cherry-pick without this

PR #2819, opened 2024-01-05
checkpoint saver tracks all checkpoints/intervals in state

aspfohl:
# What does this PR do?

Tracks all saved checkpoints (across resumptions) and intervals for tracking. Used in https://github.com/mosaicml/llm-foundry/pull/813

Note: still working on the broken `first` test 🙃 

# Before submitting


composer/callbacks/checkpoint_saver.py
```diff
@@ -303,11 +304,16 @@ def epoch_checkpoint(self, state: State, logger: Logger):
                 logger,
             )
 
-    def get_state_dict(self, state):
```
aspfohl, line 306: This looked orphaned, long since moved to `composer/utils/checkpoint.py`

tests/callbacks/test_checkpoint_saver.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
```
aspfohl, line 5:
```suggestion
```

composer/callbacks/checkpoint_saver.py
```diff
@@ -319,16 +333,19 @@ def _save_checkpoint(self, state: State, logger: Logger):
 
         # save the checkpoint to the filename
         filename_with_placeholders = self.filename.format(state, is_deepspeed, keep_placeholders=True)
+        save_filename = checkpoint.get_save_filename(state, filename_with_placeholders)
+        self.all_saved_checkpoints_to_timestamp[save_filename] = state.timestamp
 
-        saved_path = checkpoint.save_checkpoint(
+        saved_path = checkpoint._save_checkpoint(
```
aspfohl, line 339: Split this up in checkpoint.save_checkpoint so that the callback could inject the current checkpoint into the state otherwise it's always one-off. Looks like `checkpoint.save_checkpoint` is heavily documented so didn't want to make it backwards breaking

composer/callbacks/checkpoint_saver.py
```diff
@@ -423,10 +440,10 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
 
         while len(self.saved_checkpoints) > self.num_checkpoints_to_keep:
             prefix_dir = None
-            checkpoint = self.saved_checkpoints.pop(0)
-            prefix_dir = str(Path(checkpoint).parent)
+            checkpoint_to_delete = self.saved_checkpoints.pop(0)
```
aspfohl, line 443: nit: was raising pylint warnings that we could be overriding the `checkpoint` variable (import of `composer/utils/checkpoint.py`)

composer/utils/checkpoint.py
```diff
@@ -694,7 +694,8 @@ def _remove_paths(obj: Union[list, dict[str, Any]], exclude_paths: list[list[str
 
     # Recurse first, so in the case of a list, the indexing is consistent
     for key, paths_to_recurse in keys_to_recurse.items():
-        _remove_paths(obj[key], paths_to_recurse)
+        if key in obj:
+            _remove_paths(obj[key], paths_to_recurse)
```
aspfohl, line 698: Filed a ticket but I think this should a bug outside of the scope of this PR - the way this was written the following callback state would raise a key error because backslashes aren't respected:
```
'CheckpointSaver': {
   'all_saved_checkpoints_to_timestamp': {
      'first/ep1.pt': {'epoch': 1, 'batch': 5, 'sample': 10, 'token': 0, 'batch_in_epoch': 0, 'sample_in_epoch': 0, 'token_in_epoch': 0, 'total_wct': datetime.timedelta(microseconds=31455), 'epoch_wct': datetime.timedelta(0), 'batch_wct': datetime.timedelta(0)}
   }
}
```

Specifically our key of interest in flatterned to be
```
state/callbacks/CheckpointSaver/all_saved_checkpoints_to_timestamp/first/ep1.pt
```

So _remove_paths is look for a checkpoint state something like:
```
'CheckpointSaver': {
   'all_saved_checkpoints_to_timestamp': {
      'first': {
         'ep1.pt': {'epoch': 1, 'batch': 5, 'sample': 10, 'token': 0, 'batch_in_epoch': 0, 'sample_in_epoch': 0, 'token_in_epoch': 0, 'total_wct': datetime.timedelta(microseconds=31455), 'epoch_wct': datetime.timedelta(0), 'batch_wct': datetime.timedelta(0)}
      }
   }
}
```

dakinggg: ohh yeah, i don't think you can have `/` in state dict keys (not sure if a composer thing or a general thing). structure delineation is usually done with `.`.

tests/callbacks/test_checkpoint_saver.py
```diff
@@ -0,0 +1,43 @@
...
+    ts = new_state_dict['all_saved_checkpoints_to_timestamp']['example-checkpoint.pt']
+    assert isinstance(ts, dict)
+    assert ts['epoch'] == 1
+    assert ts['batch'] == 2
+    assert ts['sample'] == 0
+
+    checkpoint_saver.load_state_dict(new_state_dict)
+
+    assert checkpoint_saver.all_saved_checkpoints_to_timestamp
+    assert 'example-checkpoint.pt' in checkpoint_saver.all_saved_checkpoints_to_timestamp
```
dakinggg, line 38: add length 1 assert here too

composer/utils/checkpoint.py
dakinggg: Can think about this a bit more, but I don't think its great to silently skip keys that don't exist here. Can we revert this change and instead save the state in a different way? could do a `split('/')` before saving, or save a list instead of a dictionary, or something like that.

composer/callbacks/checkpoint_saver.py
dakinggg: Yeah, no idea what this is from. Don't think its used.

composer/callbacks/checkpoint_saver.py
```diff
@@ -303,11 +304,24 @@ def epoch_checkpoint(self, state: State, logger: Logger):
                 logger,
             )
 
-    def get_state_dict(self, state):
-        return {
-            'state': state.state_dict(),
-            'rng': reproducibility.get_rng_state(),
-        }
+    def state_dict(self) -> Dict[str, Any]:
+        state_dict = super().state_dict()
```
mvpatel2000, line 308: I think this can be removed -- we dont do it anywhere else and the parent has no state dict functionality

composer/callbacks/checkpoint_saver.py
```diff
@@ -303,11 +304,24 @@ def epoch_checkpoint(self, state: State, logger: Logger):
...
+        all_checkpoints = []
+        for save_filename, timestamp in self.all_saved_checkpoints_to_timestamp.items():
+            all_checkpoints.append((save_filename, timestamp.state_dict()))
+
+        # TODO: consider saving additional state for checkpoint rotation
+        state_dict['all_saved_checkpoints_to_timestamp'] = all_checkpoints
+        return state_dict
+
+    def load_state_dict(self, state: Dict[str, Any]):
+        super().load_state_dict(state)
```
mvpatel2000, line 319: same as above

composer/callbacks/checkpoint_saver.py
```diff
@@ -303,11 +304,24 @@ def epoch_checkpoint(self, state: State, logger: Logger):
...
+        if 'all_saved_checkpoints_to_timestamp' in state:
+            for (save_filename, timestamp_state) in state['all_saved_checkpoints_to_timestamp']:
+                new_timestamp = Timestamp()
+                new_timestamp.load_state_dict(timestamp_state)
+                self.all_saved_checkpoints_to_timestamp[save_filename] = new_timestamp
```
mvpatel2000, line 324:
```suggestion
                load_timestamp = Timestamp()
                load_timestamp.load_state_dict(timestamp_state)
                self.all_saved_checkpoints_to_timestamp[save_filename] = load_timestamp
```
so it's more clear it's not a "new" timestamp but rather loading the old one?

composer/callbacks/checkpoint_saver.py
```diff
@@ -319,16 +333,19 @@ def _save_checkpoint(self, state: State, logger: Logger):
 
         # save the checkpoint to the filename
         filename_with_placeholders = self.filename.format(state, is_deepspeed, keep_placeholders=True)
+        save_filename = checkpoint.get_save_filename(state, filename_with_placeholders)
```
mvpatel2000, line 336: do we need refactor? It looks like `save_checkpoint` returns name? We can change that behavior to return on all ranks if desired?

aspfohl: Yep - we need to know the filename and add it to state before the local checkpoint is saved - otherwise the checkpoint will be missing reference to itself when resumed

composer/callbacks/checkpoint_saver.py
aspfohl: I wanted something like `Timestamp.load_from_state(state) -> Timestamp` but realizing this could technically be `Timestamp(**timestamp_state)`? `load_state_dict` still feels a little safer for backwards/forwards compatible support

composer/callbacks/checkpoint_saver.py
aspfohl: I think I copied from EMA - but agreed 😛

composer/callbacks/checkpoint_saver.py
aspfohl: Added comment in 5770ee40a898dcc5bd571e9b4143da8407df5f51 since this didn't feel obvious

composer/callbacks/checkpoint_saver.py
mvpatel2000: yea I think load_state_dict is safer

composer/callbacks/checkpoint_saver.py
```diff
@@ -303,11 +304,23 @@ def epoch_checkpoint(self, state: State, logger: Logger):
...
-            'rng': reproducibility.get_rng_state(),
-        }
+    def state_dict(self) -> Dict[str, Any]:
+        state_dict = {}
+
+        all_checkpoints = []
+        for save_filename, timestamp in self.all_saved_checkpoints_to_timestamp.items():
+            all_checkpoints.append((save_filename, timestamp.state_dict()))
+
+        # TODO: consider saving additional state for checkpoint rotation
```
mvpatel2000, line 314:
```suggestion
```

composer/callbacks/checkpoint_saver.py
```diff
@@ -319,16 +332,21 @@ def _save_checkpoint(self, state: State, logger: Logger):
 
         # save the checkpoint to the filename
         filename_with_placeholders = self.filename.format(state, is_deepspeed, keep_placeholders=True)
+        save_filename = checkpoint.get_save_filename(state, filename_with_placeholders)
+        # It's important this is done before the checkpoint is saved, so that the state includes
+        # reference to the latest checkpoint (itself) when it is saved.
```
mvpatel2000, line 337:
```suggestion
        # Store before saving so state_dict in checkpoint has reference to latest checkpoint (itself)
```
Nit: more concise

mvpatel2000 (approved):
LGTM! :)

PR #2821, opened 2024-01-07
Adds DTensor Support

mvpatel2000:
# What does this PR do?
Adds Dtensor support.

- Updates to use new torch functions `get_state_dict` and `set_state_dict`, which simplifies checkpoint logic for all distributed settings
- Adds Dtensor/device mesh support
- Fixes a few bugs that were found along the way, e.g. simplifies how Alibi freezes params

Test failures for compile are from nightly issues on torch


tests/callbacks/test_memory_monitor.py
```diff
@@ -10,9 +10,7 @@
 from tests.common import RandomClassificationDataset, SimpleModel, device
 
 
-@device('cpu', 'gpu')
 def test_memory_monitor_warnings_on_cpu_models(device: str):
-    # Error if the user sets device=cpu even when cuda is available
```
mvpatel2000, line 15: Removed GPU test because it breaks things and doesn't really test anything meaningful.

All pytests share the same python process, and we've already initialized a NCCL backend. If we try running a unit test with device=cpu in the GPU test suite, we get an error trying to initialize a GLOO backend on a python process which already has a NCCL process initialized. This is only an issue because all tests run in the same process.

mvpatel2000:
Remaining changes seem mostly related to surgery and niche. mostly ready for review

composer/utils/checkpoint.py
```diff
@@ -165,11 +165,15 @@ def format(self, state: State, is_deepspeed: bool = False, keep_placeholders: bo
...
     if object_store is None:
+        if not os.path.exists(source_path):
+            raise FileNotFoundError(f"Couldn't find the directory {source_path}")
         return not os.path.exists(metadata_path)
     else:
         try:
             with tempfile.TemporaryDirectory() as temp_dir:
                 metadata_destination = os.path.join(str(temp_dir), '.metadata')
+                if len(object_store.list_objects(prefix=source_path)) == 0:
+                    raise FileNotFoundError(f"Couldn't find the prefix {object_store.get_uri(object_name=source_path)}")
```
mvpatel2000, line 176: @eracah are these changes meant to be included?

composer/utils/checkpoint.py
```diff
@@ -949,16 +967,40 @@ def save_checkpoint(
...
+        process_group = None
+        device_mesh = state.fsdp_device_mesh
+        if device_mesh is not None and device_mesh.ndim == 2:
+            expect_file = (device_mesh.get_local_rank(mesh_dim=0) == 0)
+            process_group = device_mesh.get_group(1)  # Only save on first replica
+            log.debug(
+                f'global_rank={dist.get_global_rank()}, {expect_file=}, process_group={get_process_group_ranks(process_group)}'
+            )
+        else:
+            expect_file = True
```
dakinggg, line 982: Is 3d device mesh allowed? I'm assuming this else is for ndim=1. If that's correct, let's error for ndim=3

composer/trainer/trainer.py
```diff
@@ -960,10 +960,7 @@ def __init__(
         assert not isinstance(device_train_microbatch_size, str)
 
         # Distributed
-        if deepspeed_config is not None or fsdp_config is not None or dist.get_world_size() > 1:
-            # Deepspeed and FSDP both require torch.distributed to be initialized, even if the world size is 1
-            # And torch.distributed is always required for multi-rank training
-            dist.initialize_dist(device, dist_timeout)
+        dist.initialize_dist(device, dist_timeout)
```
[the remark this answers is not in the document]
mvpatel2000: `get/set_state_dict` requires default PG is initialized, even if its size 1

composer/utils/checkpoint.py
mvpatel2000: It should still work. If we recognize the mesh structure and can save only replica 1, we do so (optimization), but if we don't, we save all replicas (which still works)

eracah: I don't think FSDP ever would see a 3D device mesh it would only see a 2D one or a 2D submesh with a 3D parent

eracah: it only sees 1D (fsdp) or 2D (hsdp)

eracah (approved):
LGTM

composer/trainer/dist_strategy.py
```diff
@@ -243,10 +243,11 @@ def prepare_fsdp_module(
                              'gpu and some ranks are on meta. Either keep all ranks on the same '
                              "device or set fsdp_config['sync_module_states'] = True. Otherwise, "
                              'some weights may be randomly initialized when loading a checkpoint.')
-        if fsdp_config['sharding_strategy'] in ('HYBRID_SHARD', '_HYBRID_SHARD_ZERO2'):
-            raise ValueError('HSDP (HYBRID_SHARD or _HYBRID_SHARD_ZERO2) requires '
-                             'fsdp_config["sync_module_states"] = True or different replicas will '
-                             'have different weights.')
+        # Comment out while we debug deadlock
```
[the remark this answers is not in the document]
mvpatel2000: Yes, given force sync modules

composer/trainer/trainer.py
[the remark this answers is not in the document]
antoinebrl: The latest release (`0.18.0`) breaks for us due to this modification. We see two errors related to this problem:
- we run our unit tests with `pytest` (no xdist). When we want to evaluate the behavior of some components for different workers we mock the environment variables: `mocker.patch.dict("os.environ", {"RANK": "1"})`. The error is : 
    ```
    .venv/lib/python3.10/site-packages/torch/distributed/rendezvous.py:236: in _env_rendezvous_handler
        world_size = int(_get_env_or_raise("WORLD_SIZE"))
        ValueError: Error initializing torch.distributed using env:// rendezvous: environment variable WORLD_SIZE expected, but not set
    ```
    I can mock the other environment variables but ultimately a valid `MASTER_ADDR` and `MASTER_PORT` are needed.
- We have integration tests where we launch small trainings in the CI. We trigger them from Python (single process) as we need to do a bit of hacking on the Hydra decorator. We see this error: `Torch distributed is initialized but environment variable LOCAL_WORLD_SIZE is not set.`

What would be a good way to use the `Trainer` without using the `composer` command?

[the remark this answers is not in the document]
mvpatel2000: I'm going to resolve here and move to issue so we can centralize to one place!

PR #2828, opened 2024-01-09
Remove duplicate checkpoint verifications

eracah:
# What does this PR do?
Right now if you load sharded checkpoints, your stdout gets polluted with:
`2024-01-09 00:30:15,783: rank4[280590][MainThread]: DEBUG: composer.utils.checkpoint: No validation function specified. Skipping checkpoint validation`

Upon further inspection it is the same file that it was inspecting multiple times.
This is because the multiple read_items can have the same storage data file. See [this comment](https://github.com/mosaicml/composer/blob/dev/composer/utils/checkpoint.py#L460)

This PR checks which files have already been verified before verifying


composer/utils/checkpoint.py
```diff
@@ -413,9 +413,13 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
             Raises:
                 ValueError if the data file is invalid.
             """
+            already_read = set()
```
irenedea, line 416: nit: rename to `validated_checkpoint_paths`

irenedea (approved):
thank you!!! lgtm!

composer/utils/checkpoint.py
eracah: done

PR #2829, opened 2024-01-09
Integrate PEFT LoRA with HuggingFaceModel

dakinggg:
This PR integrates PEFT LoRA with the `HuggingFaceModel` base class.

Manual test (requires a Foundry PR also), comparing the current lora implementation on foundry main (no fsdp), this implementation (no fsdp), and this implementation (with fsdp)
Screenshot 2024-01-18 at 12 16 23 AM

TODO:
- [x] rerun lora test (after refactor) `mpt-7b-lora-after-fsdp-14-vcRBaL`
- [x] rerun non lora test (without peft installed, after refactor) `no-lora-after-3-ObUYGR`
- [x] test that a small checkpoint shows up in object store `mpt-7b-lora-after-fsdp-14-vcRBaL`


composer/models/huggingface.py
```diff
@@ -642,6 +732,29 @@ def get_hf_config_from_composer_state_dict(state_dict: Dict[str, Any],
                 f'config has a valid `_name_or_path`.')
 
 
+def get_peft_config_from_composer_state_dict(state_dict: Dict[str, Any]) -> Optional['PeftConfig']:
+    """Get a PEFT config from a composer state dict
```
dakinggg, line 736: mention that it could return none

composer/models/huggingface.py
```diff
@@ -718,6 +831,46 @@ def write_huggingface_pretrained_from_composer_checkpoint(
...
+    # NOTE: This only works for default adapter name
+    if peft_config is not None:
+        weights_state_dict = filter_state_dict_peft(weights_state_dict, peft_config)
+
+        torch.save(weights_state_dict, Path(output_folder) / 'adapter_model.bin')
+    else:
+        torch.save(weights_state_dict, Path(output_folder) / 'pytorch_model.bin')
+
+
+def filter_state_dict_peft(state_dict: Dict[str, Any],
```
dakinggg, line 850: docstring

tests/models/test_hf_model.py
```diff
@@ -6,7 +6,7 @@
 import tempfile
```
dakinggg, line 6: missing tests:
- peft/task type conversion to upper case
- error for unsupported peft type

composer/models/huggingface.py
```diff
@@ -23,8 +24,15 @@
 from composer.models.base import ComposerModel
 from composer.utils import MissingConditionalImportError, dist, get_file, import_object, is_model_fsdp, safe_torch_load
 
+try:
+    from peft import PeftModel, get_peft_model
+    _peft_installed = True
+except:
+    _peft_installed = False
```
mvpatel2000, line 31: Is underscore notation consistent with rest of repo? I personally prefer `is_peft_installed` but whatever is consistent is best

composer/models/huggingface.py
```diff
@@ -66,14 +74,16 @@ class HuggingFaceModel(ComposerModel):
...
                  tokenizer: Optional[Union[transformers.PreTrainedTokenizer,
                                            transformers.PreTrainedTokenizerFast]] = None,
                  use_logits: Optional[bool] = False,
                  metrics: Optional[List[Metric]] = None,
                  eval_metrics: Optional[List[Metric]] = None,
                  shift_labels: Optional[bool] = None,
-                 allow_embedding_resizing: bool = False) -> None:
+                 allow_embedding_resizing: bool = False,
+                 peft_config: Optional['PeftConfig'] = None,
+                 peft_filter_state_dict_trainable: bool = True) -> None:
```
mvpatel2000, line 86: Can we add doc string?

composer/models/huggingface.py
```diff
@@ -84,10 +94,41 @@ def __init__(self,
...
-        self.config = model.config
+        self.config: PretrainedConfig = model.config
         self.model_forward_args = inspect.getfullargspec(self.model.forward).args
+
+        if _peft_installed and self.model_forward_args == ['self']:
+            from peft import PeftModel
+            if isinstance(self.model, PeftModel):
+                self.model_forward_args = inspect.getfullargspec(self.model.base_model.model.forward).args
+
+        # inspect.getfullargspec HuggingFace quantized model could not return args correctly
```
mvpatel2000, line 105: does this comment need to be here?

composer/models/huggingface.py
dakinggg: will check, I was thinking underscore because this isn't really intended for use outside of this file, but no strong opinion.

composer/models/huggingface.py
dakinggg: oops, yes, will do.

composer/models/huggingface.py
dakinggg: maybe? This was copy pasted over from foundry. I should check whether we can unify to just use this version of the arg getting instead of the other version.

composer/models/huggingface.py
```diff
@@ -84,10 +96,39 @@ def __init__(self,
 
         super().__init__()
         self.model = model
-        self.config = model.config
-        self.model_forward_args = inspect.getfullargspec(self.model.forward).args
+        self.config: PretrainedConfig = model.config
```
irenedea, line 99: This init function is getting quite long and it's hard to read. What do you think about doing a refactor here? Suggest a function per class property and a function for fast fails and checks

composer/datasets/utils.py
```diff
@@ -196,7 +196,7 @@ def __init__(
             self.stop_sequence_id_len = len(self.stop_sequence_ids) + 2
             self.tokenizer = tokenizer
 
-        def __call__(self, input_ids, scores: Optional[torch.FloatTensor] = None, **kwargs) -> bool:
+        def __call__(self, input_ids: torch.LongTensor, scores: Optional[torch.FloatTensor] = None, **kwargs) -> bool:
```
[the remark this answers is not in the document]
dakinggg: input ids are not floats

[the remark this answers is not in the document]
dakinggg: https://github.com/huggingface/transformers/blob/28751958874eccb155fa2ab10a79bf8068d9ae29/src/transformers/models/llama/modeling_llama.py#L977

composer/models/huggingface.py
```diff
@@ -142,6 +184,30 @@ def __init__(self,
...
 
+        if peft_config is not None:
+            from peft import PeftModel
+            if isinstance(self.model, PeftModel):
+                warnings.warn('PEFT model was passed in directly. Ignoring the provided PEFT config.')
+            else:
+                self.model = get_peft_model(self.model, peft_config)
+                log.info(f'PEFT model created. {self.model}')
+
+        self.model_is_peft = False
```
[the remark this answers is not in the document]
dakinggg: ooops, this was supposed to be `self.using_peft`. Would have crashed if run in an environment with peft installed.

composer/models/huggingface.py
```diff
@@ -433,7 +499,10 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
 
             # HF encoder decoder models like T5 expect either decoder_input_ids or labels,
             # so we add decoder_input_ids to the batch if it is missing
-            if self.model.config.is_encoder_decoder and 'decoder_input_ids' not in batch:
+            from transformers import PretrainedConfig
+            assert isinstance(self.model.config, PretrainedConfig)
+            model_config: PretrainedConfig = self.model.config
+            if model_config.is_encoder_decoder and 'decoder_input_ids' not in batch:
```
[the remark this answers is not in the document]
dakinggg: this is the only way I could get pyright to accept. open to another option.

composer/models/huggingface.py
```diff
@@ -588,6 +678,11 @@ def _is_registered_causal_lm(model: transformers.PreTrainedModel) -> bool:
                                             conda_package='transformers',
                                             conda_channel='conda-forge') from e
 
+    if peft_installed and isinstance(model, PeftModel):
```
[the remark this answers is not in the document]
dakinggg: no, it will have been imported if peft is installed. but there are some other places I don't need to reimport `PeftModel` that I should fix.

composer/models/huggingface.py
```diff
@@ -718,6 +836,59 @@ def write_huggingface_pretrained_from_composer_checkpoint(
...
+        remove_adapter_names (bool, optional): Whether to remove the adapter names from the state dict keys. Defaults to True.
+
+    Returns:
+        Dict[str, Any]: The filtered state dict
+    """
+
+    if peft_config.peft_type != 'LORA':
+        raise NotImplementedError(f'Only LoRA PEFT is supported. Got {peft_config.peft_type}')
+
+    # Filtering copied from https://github.com/huggingface/peft/blob/4186c9b104644fd247a4cc0dc2dfc1ede4665204/src/peft/utils/save_and_load.py#L68C1-L86C116
```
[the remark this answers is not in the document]
dakinggg: unfortunately that doesn't work because the function from the peft library takes in a `PeftModel`, which I don't have access to in all the places I need this function.

composer/models/huggingface.py
```diff
@@ -718,6 +836,59 @@ def write_huggingface_pretrained_from_composer_checkpoint(
...
 
+    peft_config = get_peft_config_from_composer_state_dict(composer_state_dict)
+    if peft_config is not None:
+        peft_config.save_pretrained(str(output_folder))
+
     weights_state_dict = composer_state_dict['state']['model']
     torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(weights_state_dict, prefix='model.')
-    torch.save(weights_state_dict, Path(output_folder) / 'pytorch_model.bin')
+
+    # NOTE: This only works for default adapter name
```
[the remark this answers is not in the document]
dakinggg: Technically peft models can have multiple named adapters (the default name is `default`). The way I wrote this function just hardcoded `default` because the way it figured out which adapter to use comes from the `PeftModel.active_adapter`, but I don't have a `PeftModel` everywhere I need this function. Since we're not trying to support all PEFT features at this time, its just hardcoded to default. But let me generalize the implementation a little bit to make it easier to add support in the future.

composer/models/huggingface.py
```diff
@@ -66,14 +76,16 @@ class HuggingFaceModel(ComposerModel):
...
                  tokenizer: Optional[Union[transformers.PreTrainedTokenizer,
                                            transformers.PreTrainedTokenizerFast]] = None,
                  use_logits: Optional[bool] = False,
                  metrics: Optional[List[Metric]] = None,
                  eval_metrics: Optional[List[Metric]] = None,
                  shift_labels: Optional[bool] = None,
-                 allow_embedding_resizing: bool = False) -> None:
+                 allow_embedding_resizing: bool = False,
+                 peft_config: Optional['PeftConfig'] = None,
+                 peft_filter_state_dict_trainable: bool = True) -> None:
```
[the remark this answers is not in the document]
dakinggg: sorry, what do you mean its not clear its a boolean when reading code below? or what name are you suggesting?

composer/models/huggingface.py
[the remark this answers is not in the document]
dakinggg: forgot I already have `self.config`. using that instead.

composer/models/huggingface.py
dakinggg: cleaned up

composer/models/huggingface.py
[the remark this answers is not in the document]
dakinggg: refactored to take in the argument name (even though this code will only set it to default), and updated the comment a bit

composer/models/huggingface.py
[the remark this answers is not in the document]
irenedea: oh sorry, yeah I just mean the name isn't a clear boolean name. Maybe `should_filter_state_dict_peft`?

irenedea (approved):
LGTM! thanks for adding the tiny mistral fixtures and better typing throughout!

composer/models/huggingface.py
```diff
@@ -23,22 +25,29 @@
 from composer.models.base import ComposerModel
 from composer.utils import MissingConditionalImportError, dist, get_file, import_object, is_model_fsdp, safe_torch_load
 
+try:
+    from peft import PeftModel, get_peft_model
+    peft_installed = True
```
irenedea, line 30: nit: perhaps `is_peft_installed`

dakinggg: Originally I had `_is_peft_installed`, but turns out everywhere else in composer we do it like `library_installed`, so I'm keeping this.

composer/models/huggingface.py
```diff
@@ -82,65 +95,111 @@ def __init__(self,
...
+
+        if is_causal_lm and not self.shift_labels:
+            log.warning('The shift_labels argument was set to False but the model is an instance of a'
+                        ' HuggingFace Causal LM. This may lead to incorrect behavior.')
+            # Note: No warning if shift_labels and not is_causal_lm, since the model may simply be a custom class.
+
+        if peft_config is not None:
+            self.model = _maybe_get_peft_model(peft_config, self.model)
 
+        self.using_peft = isinstance(self.model, PeftModel) if peft_installed else False
```
irenedea, line 142: nit
```suggestion
        self.is_peft_model = isinstance(self.model, PeftModel) if peft_installed else False
```

dakinggg: I went with the current naming because `self` is not really a peft model, it contains a peft model. Would prefer to leave it, lmk if you object strongly.

mvpatel2000 (approved):
Does the huggingface class actually need to be in Composer? I suppose it's a useful general utility...

1 minor nit. I think this makes sense otherwise

composer/models/huggingface.py
```diff
@@ -48,6 +57,8 @@ class HuggingFaceModel(ComposerModel):
         eval_metrics (list[Metric], optional): list of torchmetrics to compute on the eval_dataloader, or be accessible to :class:`Evaluator`s. Default: ``None``.
         shift_labels (bool, optional): If True, the batch's labels will be shifted before being used to calculate metrics. This should be set to true for CausalLM models and false otherwise. If not specified, `shift_labels` will be set automatically based on the model class name. Default: ``None``.
         allow_embedding_resizing (bool, optional): If True, the model's embeddings will be automatically resized when they are smaller than the tokenizer vocab size. Default: ``False``.
+        peft_config (PeftConfig, optional): Optional PEFT config to apply to the model. If provided, the model will be converted to a PEFT model. Only LoRA is currently supported.
+        should_filter_state_dict_peft (bool, optional): If True _and_ PEFT is active, the state dict will only contain the PEFT weights, not the frozen base model weights.
```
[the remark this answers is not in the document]
dakinggg: done

Issue #2632, opened 2023-10-11
AttributeError on DeviceMPS


agathedvlp:
** Environment **

PyTorch information
-------------------
PyTorch version: 2.0.1
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 13.6 (arm64)
GCC version: Could not collect
Clang version: 15.0.0 (clang-1500.0.40.1)
CMake version: Could not collect
Libc version: N/A

Python version: 3.10.13 (main, Sep  4 2023, 18:40:56) [Clang 14.0.3 (clang-1403.0.22.14.1)] (64-bit runtime)
Python platform: macOS-13.6-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M2 Pro

Versions of relevant libraries:
[pip3] mypy==1.5.1
[pip3] mypy-extensions==1.0.0
[pip3] numpy==1.25.2
[pip3] pytorch-lightning==2.0.9.post0
[pip3] pytorch-ranger==0.1.1
[pip3] torch==2.0.1
[pip3] torch-optimizer==0.3.0
[pip3] torchmetrics==1.0.3
[pip3] torchvision==0.15.2
[conda] Could not collect

Composer information
--------------------
Composer version: 0.16.3
Composer commit hash: None
Host processor model name: Apple M2 Pro
Host processor core count: 12
Number of nodes: 1
Accelerator model name: N/A
Accelerators per node: 0
CUDA Device Count: 0

## Context 
`self._eval_loop()` crashes line 2910 of train.py when `isinstance(self.state.device, DeviceMPS)` and `self.state.outputs` is a dictionary with error: `AttributeError: 'dict' object has no attribute 'cpu'` however manually setting the device to "cpu" runs without error.

## Expected behavior
We would expect to have the same behaviour regardless of the device.

mvpatel2000:
Thanks for flagging this! We will have a fix soon :)

mvpatel2000:
https://github.com/mosaicml/composer/pull/2706 Should be fixed! Please let me know if you hit any errors

PR #2834, opened 2024-01-10
Fix MPS with sequence loss 

JAEarly:
# What does this PR do?

**TLDR**
Fixes the trainer eval loop for MPS devices when the model output is a list of tensors rather than a `torch.Tensor` or `Mapping`.

**Related**
Similar issue but with dict: #2632 
PR for above issue: #2706

**Details**
In the trainer evaluation loop, model outputs (`self.state.outputs`) are moved to the CPU if they are on an MPS device (to avoid torchvision numerical errors on MPS devices). Currently this works fine if `self.state.outputs` is a Mapping or a torch.Tensor. However, it fails for Sequence types (e.g. a list of torch.Tensors).

According to the State class, `outputs` is expected to be of type `torch.Tensor | Sequence[torch.Tensor]`. So Sequence types should be supported in this process in the eval loop. Strictly, `outputs` should not be a `Mapping`, despite the eval operation supporting this. I have left the `Mapping` support in place for now but happy to revisit.

As it was a little difficult to debug this issue, I have added an error message which should make it clearer if an invalid output type is used (such that it cannot correctly be mapped from MPS to CPU). 

# Before submitting

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -2983,6 +2983,13 @@ class Trainer:
                                                 outputs[k] = v.cpu()
                                             else:
                                                 outputs[k] = v
+                                    elif isinstance(self.state.outputs, Sequence):
+                                        outputs = []
+                                        for v in self.state.outputs:
+                                            if isinstance(v, torch.Tensor):
+                                                outputs.append(v.cpu())
+                                            else:
+                                                outputs.append(v)
                                     else:
                                         outputs = self.state.outputs.cpu()
                                 else:
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -2990,10 +2990,8 @@ class Trainer:
                                                 outputs.append(v.cpu())
                                             else:
                                                 outputs.append(v)
-                                    elif isinstance(self.state.outputs, torch.Tensor):
-                                        outputs = self.state.outputs.cpu()
                                     else:
-                                        raise ValueError('Unsupported type for state.outputs in trainer eval loop.')
+                                        outputs = self.state.outputs.cpu()
                                 else:
                                     outputs = self.state.outputs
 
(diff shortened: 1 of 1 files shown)


JAEarly:
Upon further inspection, `eval_forward` can strictly return `Any` type, so keeping support for `Mapping` makes sense. Adding support for `self.state.options == None` in this PR may also be useful.

JAEarly:
@mvpatel2000 this is a relatively small change but is fairly annoying for running stuff on MPS, could it be merged in please? Thank you!

composer/trainer/trainer.py
```diff
@@ -2983,8 +2983,17 @@ def _eval_loop(
...
+                                        outputs = []
+                                        for v in self.state.outputs:
+                                            if isinstance(v, torch.Tensor):
+                                                outputs.append(v.cpu())
+                                            else:
+                                                outputs.append(v)
+                                    elif isinstance(self.state.outputs, torch.Tensor):
                                         outputs = self.state.outputs.cpu()
+                                    else:
+                                        raise ValueError('Unsupported type for state.outputs in trainer eval loop.')
```
dakinggg, line 2996: To be a bit safer, I think we don't want to error here. Since the output of `eval_forward` is `Any`, the user could actually be returning a custom object (that could in theory have `.cpu()` defined on it. We don't want to break any existing workflows doing something like this. So I would leave the fallback case of `outputs = self.state.outputs.cpu()`. Does this make sense to you?

dakinggg (commented):
Thanks for the fix! I left one small comment, and happy to approve if that makes sense to you!

composer/trainer/trainer.py
JAEarly: Made the change as you suggested, thanks!

dakinggg (approved):
Thanks!

mvpatel2000:
@JAEarly apologies for the delay in review! Next time please feel free to tag me as a reviewer.... I will see if we can setup auto-tagging as well. We didn't see this earlier :(

mvpatel2000 (approved):

PR #2835, opened 2024-01-11
Check ckpt dir exists before checking for metadata file in legacy sharding check

eracah:
# What does this PR do?
Bug where if user misspecified load_path and it did not exist, then `.metadata` file would not exist and then composer would think it was legacy sharded checkpoint. 

This fix checks existence of dir/prefix in object store first and then fails fast


composer/utils/checkpoint.py
```diff
@@ -165,13 +165,22 @@ def format(self, state: State, is_deepspeed: bool = False, keep_placeholders: bo
 def is_checkpoint_legacy_sharded(object_store: Optional[ObjectStore], source_path: str):
     metadata_path = str(Path(source_path) / Path('.metadata'))
     if object_store is None:
+        # If directory doesn't exist, then it's not any checkpoint.
+        if os.path.exists(source_path):
+            raise FileNotFoundError(f"Couldn't find the directory {source_path}")
```
mvpatel2000, line 170: Just to be clear, because this is in `load_checkpoint`, it should not affect autoresume related logic, right?

eracah: uh it shouldn't

mvpatel2000 (commented):
Add a test otherwise LGTM?

PR #2839, opened 2024-01-11
Log all gpu rank stdout/err to MosaicML platform

jjanezhang:
# Log all gpu rank stdout/err to MosaicML platform

We are no longer differentiating between stderr and stdout for gpu rank logs.

If run is sent by Mosaic platform, redirect GPU local rank 1-7 logs to `gpu_x.txt` in Mosaic logs directory for each node rank. 

image

On mosaic platform
image

When mosaic env var is false
image


siriuslee (commented):
Platform parts look good if the stdout / stderr change is acceptable

dakinggg (commented):
Does this change mean that runs on the mosaic platform will no longer show the run logs via `mcli logs`?

jjanezhang:
> Does this change mean that runs on the mosaic platform will no longer show the run logs via `mcli logs`?

We will still be able to see the gpu 0 rank logs (+ failed gpu rank logs when the run fails) using `mcli logs` for non finetuning runs. The only diff is that when your run fails, you will get the full chronological logs of all failed gpu ranks instead of separated stdout and stderr. Additionally, you will be able to get individual gpu rank logs regardless of whether the gpu rank failed or not.

composer/cli/launcher.py
```diff
@@ -466,12 +447,14 @@ def main():
     log.setLevel(logging.INFO if args.verbose else logging.WARN)
 
     processes = {}
-
     log_tmpdir = tempfile.TemporaryDirectory()
-    if args.stdout is None:
-        args.stdout = f'{log_tmpdir.name}/rank{{rank}}.stdout.txt'
-    if args.stderr is None:
-        args.stderr = f'{log_tmpdir.name}/rank{{rank}}.stderr.txt'
+    log_file_format = f'{log_tmpdir.name}/rank{{rank}}.txt'
```
dakinggg, line 451: I'm thinking we should preserve the option of the old behavior somehow since this is an api breaking change. @mvpatel2000 wdyt?

composer/cli/launcher.py
```diff
@@ -313,19 +314,20 @@ def _get_file(format: str):
                         local_world_size=nproc,
                         node_rank=node_rank,
                     )
-                    return open(filename, 'x+')
+                    return open(filename, 'a+')
```
dakinggg, line 317: curious why you needed to make this change? I think failing if the log file already exists is a reasonable behavior.

composer/loggers/mosaicml_logger.py
```diff
@@ -36,6 +36,7 @@
 RUN_NAME_ENV_VAR = 'RUN_NAME'
 MOSAICML_PLATFORM_ENV_VAR = 'MOSAICML_PLATFORM'
 MOSAICML_ACCESS_TOKEN_ENV_VAR = 'MOSAICML_ACCESS_TOKEN_FILE'
+MOSAICML_LOG_DIR = 'MOSAICML_LOG_DIR'
```
dakinggg, line 39:
```suggestion
MOSAICML_LOG_DIR_ENV_VAR = 'MOSAICML_LOG_DIR'
```

dakinggg (approved):

composer/cli/launcher.py
```diff
@@ -473,6 +479,14 @@ def main():
     if args.stderr is None:
         args.stderr = f'{log_tmpdir.name}/rank{{rank}}.stderr.txt'
 
+    # If running on the Mosaic platform, log all gpu ranks' stderr and stdout to Mosaic platform
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and os.environ.get(
+            MOSAICML_ACCESS_TOKEN_ENV_VAR) is not None and os.environ.get(MOSAICML_LOG_DIR) is not None:
```
mvpatel2000, line 484:
```suggestion
    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and os.environ.get(MOSAICML_LOG_DIR) is not None:
```
I don't think we need the access token, so we can skip that check

composer/cli/launcher.py
```diff
@@ -473,6 +479,14 @@ def main():
...
+        log.info('Logging all gpu ranks to Mosaic Platform')
+        log_file_format = f'{os.environ.get(MOSAICML_LOG_DIR)}/gpu_{{rank}}.txt'
+        args.stdout = log_file_format
```
mvpatel2000, line 487: if overriding a user specified argument, please raise a warning

composer/cli/launcher.py
```diff
@@ -260,7 +261,7 @@ def _launch_processes(
     command_mode: bool,
     training_script: str,
     stdout_file_format: str,
-    stderr_file_format: str,
+    stderr_file_format: Union[str, None],
```
mvpatel2000, line 264: Why is this needed? Is it to force it to stdout?

jjanezhang: Yea, if no stderr file format is provided then we default to piping stderr to stdout

composer/cli/launcher.py
jjanezhang: good q, I don't think we actually need a+ because none of the gpu log files should exist at this point.

composer/cli/launcher.py
```diff
@@ -473,6 +480,19 @@ def main():
...
 
+    # If running on the Mosaic platform, log all gpu ranks' stderr and stdout to Mosaic platform
+    if os.environ.get(
+            MOSAICML_PLATFORM_ENV_VAR,
+            'false').lower() == 'true' and str(os.environ.get(MOSAICML_LOG_DIR_ENV_VAR, 'false')).lower() != 'false':
+        log.info('Logging all GPU ranks to Mosaic Platform.')
+        log_file_format = f'{os.environ.get(MOSAICML_LOG_DIR_ENV_VAR)}/gpu_{{rank}}.txt'
+        if args.stderr is not None or args.stdout is not None:
+            warnings.warn(
+                'Logging to Mosaic Platform. Ignoring provided stdout and stderr args. To use provided stdout and stderr, set MOSAICML_LOG_DIR=false'
```
mvpatel2000, line 491:
```suggestion
                'Logging to Mosaic Platform. Ignoring provided stdout and stderr args. To use provided stdout and stderr, set MOSAICML_LOG_DIR=false.'
```
Nit

PR #2892, opened 2024-01-22
Version upgrade vincent

KuuCi:
Upgrading python 3.10 to 3.11


mvpatel2000 (commented):
@KuuCi ~what do you think about updating dockerfiles later once torch 2.2 comes out and tie 3.11 to that for images?~

Let's just replace torch 2.1.2 with 3.11 and keep rest at existing? WDYT. Also ok with bumping all if ur confident

docker/Dockerfile
```diff
@@ -77,55 +77,56 @@ ARG CUDA_VERSION
 # If this file is present after the first command, kaniko
 # won't be able to build the docker image.
 RUN if [ -n "$CUDA_VERSION" ]; then \
-        rm -f /usr/local/cuda-$(echo $CUDA_VERSION | cut -c -4)/cuda-$(echo $CUDA_VERSION | cut -c -4); \
```
hanlint, line 80: Are these dockerfile changes needed? Looks like unnecessary formatting changes, should we revert?

docker/generate_build_matrix.py
```diff
@@ -32,6 +32,14 @@ def _get_torchvision_version(pytorch_version: str):
     raise ValueError(f'Invalid pytorch_version: {pytorch_version}')
 
 
+def _get_pytorch_version(python_version: str):
```
hanlint, line 35: nit: since this can return multiple versions, should be `_get_pytorch_versions`?

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -73,13 +73,13 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
     seq1 = tiny_gpt2_tokenizer('Dogs are furry')['input_ids']
     seq2 = tiny_gpt2_tokenizer('Dogs are furry\n\n')['input_ids']
     seq1 = [50257] * (len(seq2) - len(seq1)) + seq1
-    input_ids = torch.tensor([seq1, seq2])
+    input_ids = torch.LongTensor([seq1, seq2])
```
hanlint, line 76: Curious to understand why this change is needed

KuuCi: This was something caught by the new lint tests that Brian pushed where it was required to be LongTensor

docker/generate_build_matrix.py
KuuCi: resolved

docker/Dockerfile
KuuCi: Only thing that should be changing here is manually installing snappy as it will not properly install normally

hanlint: Looks good, thanks!

docker/README.md
```diff
@@ -32,14 +32,14 @@ To install composer, once inside the image, run `pip install mosaicml`.
 | Linux Distro   | Flavor   | PyTorch Version   | CUDA Version        | Python Version   | Docker Tags                                                                              |
 |----------------|----------|-------------------|---------------------|------------------|------------------------------------------------------------------------------------------|
 | Ubuntu 20.04   | Base     | 2.3.0             | 12.1.0 (Infiniband) | 3.10             | `mosaicml/pytorch:2.3.0_cu121-nightly20240110-python3.10-ubuntu20.04`                    |
-| Ubuntu 20.04   | Base     | 2.1.2             | 12.1.0 (Infiniband) | 3.10             | `mosaicml/pytorch:latest`, `mosaicml/pytorch:2.1.2_cu121-python3.10-ubuntu20.04`         |
-| Ubuntu 20.04   | Base     | 2.1.2             | 12.1.0 (EFA)        | 3.10             | `mosaicml/pytorch:latest-aws`, `mosaicml/pytorch:2.1.2_cu121-python3.10-ubuntu20.04-aws` |
-| Ubuntu 20.04   | Base     | 2.1.2             | cpu                 | 3.10             | `mosaicml/pytorch:latest_cpu`, `mosaicml/pytorch:2.1.2_cpu-python3.10-ubuntu20.04`       |
```
j316chuck, line 37: There should be no red here for stability reasons. How about we only add a nightly python3.11 entry and only have one green.

docker/generate_build_matrix.py
```diff
@@ -32,6 +32,14 @@ def _get_torchvision_version(pytorch_version: str):
     raise ValueError(f'Invalid pytorch_version: {pytorch_version}')
 
 
+def _get_pytorch_versions(python_version: str):
+    if python_version == '3.10':
+        return ['1.13.1', '2.0.1', '2.1.2']
+    if python_version == '3.11':
+        return ['2.1.2']
```
j316chuck, line 39: let's just make this a nightly entry

docker/Dockerfile
```diff
@@ -356,7 +357,8 @@ RUN apt-get update && \
 RUN pip install --no-cache-dir --upgrade \
         certifi${CERTIFI_VERSION} \
         ipython${IPYTHON_VERSION} \
-        urllib3${URLLIB3_VERSION}
+        urllib3${URLLIB3_VERSION} \
+        python-snappy
```
j316chuck, line 361: Where are the FA2 changes?

PR #2893, opened 2024-01-22
Azure checkpointing support

mvpatel2000:
# What does this PR do?

Azure checkpointing support


dakinggg (approved):

eracah (changes requested):
Code looks good, but tests? Any manual tests to share? Ideally some unit tests too

PR #2897, opened 2024-01-23
Pass PG into checkpoint load and load rng with state_dict

mvpatel2000:
# What does this PR do?

1. Pass process group into load call (which should be faster)
2. Load RNG in one call to avoid duplicate reads

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -507,7 +507,10 @@ def load_sharded_checkpoint(
                 # For older versions of torch, we load optimizer separately.
                 if version.parse(torch.__version__) < version.parse('2.1.3'):
                     cur_state_dict.pop('optimizers')
-                state_dict = {'state': cur_state_dict}
+                state_dict = {
+                    'state': cur_state_dict,
+                    'rng': reproducibility.get_rng_state(),
+                }
 
             if ignore_keys:
                 # Filter provided list of key paths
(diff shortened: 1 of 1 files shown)


composer/utils/checkpoint.py
```diff
@@ -518,21 +510,35 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
                 # Ensure state exists
                 state_dict['state'] = state_dict.get('state', {})
 
+            # Only some ranks are meant to load checkpoint
+            expect_file = False
+            process_group = None
+            device_mesh = state.fsdp_device_mesh
+            if device_mesh is not None and device_mesh.ndim == 2:
```
eracah, line 517: Little comment addition here
```suggestion
            # device_mesh.ndim == 2 signifies hybrid sharding.
            if device_mesh is not None and device_mesh.ndim == 2:
```

composer/utils/checkpoint.py
```diff
@@ -518,21 +510,35 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+                expect_file = (device_mesh.get_local_rank(mesh_dim=0) == 0)
```
eracah, line 518: add another clarification comment
```suggestion
                # mesh_dim=0 is the replication dimension, so only the devices on one of the replicas (we pick 0) need to save.
                expect_file = (device_mesh.get_local_rank(mesh_dim=0) == 0)
```

composer/utils/checkpoint.py
```diff
@@ -518,21 +510,35 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+                if expect_file:
+                    process_group = device_mesh.get_group(1)  # Only save on first replica
```
eracah, line 520: Another comment
```suggestion
                    # Get the underlying process group for the sharding dim of the mesh that this rank is in.
                    process_group = device_mesh.get_group(1)  # Only save on first replica
```

eracah (approved):
LGTM just a few clarifying comments to add to make it easier to read

eracah:
Tests?

mvpatel2000:
> Tests?

I think existing sharded checkpoints cover this. We unfortunately cannot test HSDP with 2 GPUs, so I've queued manual tests I'll DM you.

eracah (approved):
Looks decent

PR #2899, opened 2024-01-23
Remove monkeypatch and new state dict APIs for torch 2.2

mvpatel2000:
# What does this PR do?

Remove monkeypatch and new state dict APIs (which also always require dist init) for torch 2.2. We will delay to 2.3.

### DIFF
--- a/composer/trainer/mosaic_fsdp.py
+++ b/composer/trainer/mosaic_fsdp.py
@@ -69,16 +69,6 @@ def patch_pytorch():
         from torch.distributed.fsdp import _runtime_utils
         _runtime_utils._validate_and_get_hybrid_shard_state = lambda *args, **kwargs: None
 
-        # Monkeypatch dtensor support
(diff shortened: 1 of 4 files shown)


composer/trainer/mosaic_fsdp.py
```diff
@@ -69,16 +69,6 @@ def patch_pytorch():
         from torch.distributed.fsdp import _runtime_utils
         _runtime_utils._validate_and_get_hybrid_shard_state = lambda *args, **kwargs: None
 
-        # Monkeypatch dtensor support
```
dskhudia, line 72: Is this already in torch 2.2?

dskhudia (approved):

composer/trainer/mosaic_fsdp.py
mvpatel2000: It's buggy so we wont support in torch 2.2

linden-li (approved):

PR #2901, opened 2024-01-24
Only load RNG keys that exist

mvpatel2000:
# What does this PR do?

Only load RNG keys that exist. This was broken in https://github.com/mosaicml/composer/pull/2897 but was not caught in manual tests for some reason 🤔


eracah (approved):
LDTM

PR #2902, opened 2024-01-24
Fix bug with CoT early stopping and LLama2 tokenizer

bmosaicml:
# What does this PR do?

LLaMa2 tokenizer didn't work with early stopping because it inserted some extra token ([28971](https://github.com/huggingface/transformers/issues/26273)) before \n\n. This caused the stop sequence to appear longer than it really was, which resulted in the stopping criteria looking back too far into the input and causing generation to stop as soon as it begins. With the new implementation we remove any superfluous tokens in the stopping criteria (i.e. those tokens that add nothing to the decoded result) I also reduced the amount of extra lookback we do to make it so we don't ever look back farther than we should and stop as soon as generation begins.

Testing: `mpt-eval-NzfmSy`

```
| Category   | Benchmark   | Subtask   |   Accuracy | Number few shot   | Model                    |
|:-----------|:------------|:----------|-----------:|:------------------|:-------------------------|
|            | gsm8k       |           |   0.141774 | 0-shot            | meta-llama/Llama-2-7b-hf |
```

Results replicate LLama2 paper (14.6 for 7b)

# Before submitting

### DIFF
--- a/composer/datasets/utils.py
+++ b/composer/datasets/utils.py
@@ -199,7 +199,7 @@ try:
             self.stop_sequence_id_len = len(self.stop_sequence_ids) + 1
             self.tokenizer = tokenizer
 
-        def __call__(self, input_ids, scores: Optional[torch.FloatTensor] = None, **kwargs) -> bool:
+        def __call__(self, input_ids: torch.LongTensor, scores: Optional[torch.FloatTensor] = None, **kwargs) -> bool:
             # For efficiency, we compare the last n tokens where n is the number of tokens in the stop_sequence
             lookback_ids_batch = input_ids[:, :][:, -self.stop_sequence_id_len:]
             lookback_tokens_batch = self.tokenizer.batch_decode(lookback_ids_batch)
@@ -215,7 +215,7 @@ try:
             return False not in self.done_tracker
 
     def stop_sequences_criteria(
-        tokenizer: transformers.PreTrainedTokenizer,
+        tokenizer: transformers.PreTrainedTokenizerBase,
         stop_sequences: List[str],
         batch_size: int,
     ) -> transformers.StoppingCriteriaList:
--- a/composer/datasets/utils.py
+++ b/composer/datasets/utils.py
@@ -179,7 +179,7 @@ try:
         def __init__(
             self,
             stop_sequence: str,
-            tokenizer: transformers.PreTrainedTokenizer,
+            tokenizer: transformers.PreTrainedTokenizerBase,
             batch_size: int,
         ) -> None:
             self.done_tracker = [False] * batch_size
--- a/composer/datasets/utils.py
+++ b/composer/datasets/utils.py
@@ -186,20 +186,22 @@ try:
             self.stop_sequence = stop_sequence
             self.stop_sequence_ids = tokenizer.encode(stop_sequence, add_special_tokens=False)
 
-            # we look back for 2 more tokens than it takes to encode our stop sequence
+            # sentence piece tokenizers add a superflous underline token before string-initial \n
+            # that throws off our calculation of the stop sequence length
+            # so we remove any token ids that produce empty strings
+            self.stop_sequence_ids = [id for id in self.stop_sequence_ids if tokenizer.decode(id) != '']
+
+            # we look back for 1 more token than it takes to encode our stop sequence
             # because tokenizers suck, and a model might generate `['\n', '\n']` but our `sequence` is `['\n\n']`
             # and we don't want to mistakenly not stop a generation because our
             # (string) stop sequence was output in a different tokenization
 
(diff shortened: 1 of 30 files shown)


tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -83,6 +83,24 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
     assert eos_criteria(input_ids, None)
 
 
+def test_stop_sequences_criteria_sentencepiece():
+    pytest.importorskip('datasets')
+
+    tokenizer = AutoTokenizer.from_pretrained('huggyllama/llama-7b', use_fast=False)
```
dakinggg, line 89: please use the llama tokenizer fixture not download this one

composer/datasets/utils.py
```diff
@@ -186,20 +186,22 @@ def __init__(
...
+
+            # we look back for 1 more token than it takes to encode our stop sequence
             # because tokenizers suck, and a model might generate `['\n', '\n']` but our `sequence` is `['\n\n']`
             # and we don't want to mistakenly not stop a generation because our
             # (string) stop sequence was output in a different tokenization
 
-            # NOTE: there is a minor danger that this will end up looking back 2 tokens into the past, into the inputs to the model,
-            # and stopping generation immediately as a result. With only 2 extra tokens of lookback, this risk is minimized
-            self.stop_sequence_id_len = len(self.stop_sequence_ids) + 2
+            self.stop_sequence_id_len = len(self.stop_sequence_ids) + 1
```
dakinggg, line 199: why did you change this to +1

dakinggg:
Can you please include some evidence this resolves the issue in the pr description?

composer/datasets/utils.py
bmosaicml: since the model generates 1 token before the first call to the stopping criteria, this ensures we will never stop too early for any of the stopping criteria we use in practice (since they're all multi-token).

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -220,13 +220,31 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
     seq2 = tiny_gpt2_tokenizer('Dogs are furry\n\n')['input_ids']
     seq1 = [50257] * (len(seq2) - len(seq1)) + seq1
     input_ids = torch.tensor([seq1, seq2])
-    assert not eos_criteria(input_ids, None)
+    assert not eos_criteria(input_ids.type(torch.long), None)  # pyright: ignore[reportGeneralTypeIssues]
```
dakinggg, line 223: instead of this, you just want to construct `LongTensor` on the line above

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -220,13 +220,31 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
...
 
     eos_criteria = MultiTokenEOSCriteria('\n\n', tiny_gpt2_tokenizer, 2)
     seq1 = tiny_gpt2_tokenizer('Dogs are furry\n\n')['input_ids']
     seq2 = tiny_gpt2_tokenizer('Dogs are furry\n\n')['input_ids']
     input_ids = torch.tensor([seq1, seq2])
-    assert eos_criteria(input_ids, None)
+    assert eos_criteria(input_ids.type(torch.long), None)  # pyright: ignore[reportGeneralTypeIssues]
```
dakinggg, line 229: same

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -220,13 +220,31 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
...
+
+
+def test_stop_sequences_criteria_sentencepiece(tiny_llama_tokenizer):
+    pytest.importorskip('datasets')
+
+    tokenizer = tiny_llama_tokenizer
+    eos_criteria = MultiTokenEOSCriteria('\n\n', tokenizer, 2)
+    seq1 = tokenizer('\n\nDogs')['input_ids']  # check to make sure starting with the start sequence doesnt break it
```
dakinggg, line 237:
```suggestion
    seq1 = tokenizer('\n\nDogs')['input_ids']  # check to make sure starting with the stop sequence doesnt break it
```
?

tests/datasets/test_in_context_learning_datasets.py
```diff
@@ -220,13 +220,31 @@ def test_stop_sequences_criteria(tiny_gpt2_tokenizer):
...
+    seq2 = tokenizer('Dogs are furry\n\n')['input_ids']
+    seq1 = [50257] * (len(seq2) - len(seq1)) + seq1
```
dakinggg, line 239: what is 50257 supposed to do here? It is outside the vocab of llama

dakinggg (approved):

PR #2903, opened 2024-01-25
File lock for sharded checkpoint download

mvpatel2000:
# What does this PR do?

Adds file lock to download for sharded checkpoints to minimize potential time spent at barrier to avoid requiring large timeouts.

Along for ride, removes some annoying frequent logs.


dakinggg (commented):
Lgtm, would like to see manual test

PR #2907, opened 2024-01-26
Fix FP8 checkpoint resumption with onnx export flag

j316chuck:
# What does this PR do?
Before this PR checkpoint resumption on fp8 MPT's did not work since checkpoints serialized extra fp8 buffers in bytes which is incompatible with torch's `load_sharded_optimizers`

After this PR checkpoint resumption from fp8 serializes out extra fp8 buffers in tensors in onnx mode. 

# Tests

**Torch 2.1**
- Failed run w/out this PR: `mpt-125m-sharded-regression-fp8-ignore-td0frt` 🔴  
- Working run w/this PR: `mpt-125m-sharded-regression-fp8-ignore-6DD62n` ✅

**Torch 2.3**
- Run w/out this PR: `mpt-125m-sharded-regression-fp8-ignore-tAGCla` ✅ 
- Working run w/this PR: `mpt-125m-sharded-regression-fp8-ignore-8bZirC` ✅

### DIFF
--- a/composer/core/precision.py
+++ b/composer/core/precision.py
@@ -78,6 +78,9 @@ def get_precision_context(precision: Union[str, Precision],
                 }
             fp8_recipe = DelayedScaling(**precision_config)
             with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+                # The te.onnx_export flag ensures that we save all fp8 buffers
+                # as tensors instead of bytes. This is necessary for proper
+                # saving and resumption of checkpoints.
                 with te.onnx_export(enabled=True):
                     yield
         else:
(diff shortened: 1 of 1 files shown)


composer/core/precision.py
```diff
@@ -78,7 +78,8 @@ def get_precision_context(precision: Union[str, Precision],
                 }
             fp8_recipe = DelayedScaling(**precision_config)
             with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
-                yield
+                with te.onnx_export(enabled=True):
```
dskhudia, line 81: @j316chuck : can you add a comment here why onnx_export is necessary? People reading this code in future will get confused.

dskhudia (approved):

dskhudia:
@j316chuck : Thinking a bit more about it, I think we should apply it when we are creating a checkpoint instead of the time when we are requesting precision.

composer/core/precision.py
j316chuck: Yes that is a good point. Let me do that

j316chuck (commented):
Added comment

dskhudia:
@j316chuck : sounds good. I looked at the defintion of this context manager and it doesn't do much besides setting a global variable. I was worried about this slowing down training but it probably doesn't. Feel free to merge.

PR #2914, opened 2024-01-28
Broadcast files for HSDP

mvpatel2000:
# What does this PR do?

When training with HSDP, we store multiple replicas of a model. Downloading all replicas can use an enormous amount of bandwidth.

Instead of having all ranks download the checkpoint with HSDP, we now only download on replica 0 and broadcast files to other ranks. This massively saves on download for large checkpoints.

Along for ride, I've removed the process group specification for dist_cp.load. The current implementation is incorrect since it passes in the PG only for replica 0 and `None` for other replicas, which is incorrect. Specifying PGs for all replicas leads to deadlock... will be sorted out in a second PR 

Also fixes a RNG issue in the refactor which prevents increasing number of ranks


composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+            if first_replica:
+                log.debug(f'Rank {dist.get_global_rank()} starting to download files.')
+                for plan_item in plan.items:
+                    # Each plan item has a storage index which points to the relative path of the shard file at save time.
+                    relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                    # Download the shard file to the relative path it's associated to and save that relative path
+                    # to the root directory specified to the FileSystem reader constructor.
+                    file_destination = str(Path(self.destination_path) / Path(relative_file_path))
+                    # The file could have already been downloaded as different plan items can point to same file.
+                    if not os.path.exists(file_destination):
```
dakinggg, line 469: jira for figuring out if this is a race condition. not blocking pr since its not any worse than before.

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                        log.debug(f'Downloading {relative_file_path} to {file_destination}.')
+                        self.object_store.download_object(object_name=str(
+                            Path(self.source_path) / Path(relative_file_path)),
+                                                          filename=file_destination)
+                        log.debug(f'Finished downloading {relative_file_path} to {file_destination}.')
 
             # 2. Wait for all ranks to finish.
+            log.debug(f'Finished downloading all files.')
```
dakinggg, line 477:
```suggestion
            log.debug(f'Rank {dist.get_global_rank()} finished downloading all files.')
```

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
             dist.barrier()
+            log.debug('Done waiting for all ranks to finish downloading files.')
+
+            # 3. Broadcast files to all other replicas if HSDP
+            if device_mesh is not None and device_mesh.ndim == 2:
+                # Broadcast file to all replicas. Assume replica size is at least 1 node
+                replicate_process_group = device_mesh.get_group(0)  # Replicate replicate_process_group
```
dakinggg, line 484: the comment on this line doesn't add any info. Does it say what you meant?

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                    if not os.path.exists(file_destination):
+                        log.debug(f'Downloading {relative_file_path} to {file_destination}.')
+                        self.object_store.download_object(object_name=str(
+                            Path(self.source_path) / Path(relative_file_path)),
+                                                          filename=file_destination)
+                        log.debug(f'Finished downloading {relative_file_path} to {file_destination}.')
 
             # 2. Wait for all ranks to finish.
+            log.debug(f'Finished downloading all files.')
             dist.barrier()
```
dakinggg, line 478: If the single replica download takes longer than `dist_timeout`, this will nccl out right?

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                replicate_process_group = device_mesh.get_group(0)  # Replicate replicate_process_group
+                shard_size = device_mesh.size(1)
+
+                # Send list of files to all ranks
+                file_list = [sorted(os.listdir(self.destination_path))]
+                dist.broadcast_object_list(file_list,
+                                           src=dist.get_global_rank() % shard_size,
+                                           group=replicate_process_group)
+                file_list = file_list[0]
+                log.debug(f'{file_list=}')
```
dakinggg, line 493: add some text to this debug statement

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                        continue
+                    if dist.get_local_rank() == 0:
+                        full_path = os.path.join(self.destination_path, file_name)
+                        log.debug(f'Transferring {full_path=}')
+                        file_object = [None]
+                        if dist.get_global_rank() % shard_size == dist.get_global_rank():
+                            # Process with rank 0 reads the file and prepares the object
+                            with open(full_path, 'rb') as f:
+                                file_object = [{'content': f.read()}]
+                            # log.debug(f'md5sum of {full_path=} is {file_object["content"][:10]}')
```
dakinggg, line 507: del

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                                           src=dist.get_global_rank() % shard_size,
+                                           group=replicate_process_group)
+                file_list = file_list[0]
+                log.debug(f'{file_list=}')
+
+                # Send each file to the appropriate rank
+                for file_name in file_list:
+                    if 'metadata' in file_name:  # All ranks already have the metadata file
+                        continue
+                    if dist.get_local_rank() == 0:
```
dakinggg, line 499: why is this local rank 0 gated? I would think that every rank in replica 0 needs to send its files to its peers?

composer/utils/checkpoint.py
```diff
@@ -443,23 +453,75 @@ def __init__(self, source_path: str, destination_path: str, object_store):
...
+                            # Process with rank 0 reads the file and prepares the object
+                            with open(full_path, 'rb') as f:
+                                file_object = [{'content': f.read()}]
+                            # log.debug(f'md5sum of {full_path=} is {file_object["content"][:10]}')
+                        dist.broadcast_object_list(file_object,
+                                                   src=dist.get_global_rank() % shard_size,
+                                                   group=replicate_process_group)
+                        received_file_object = file_object[0]
+                        assert received_file_object is not None
+                        if dist.get_global_rank() % shard_size != dist.get_global_rank():
```
dakinggg, line 513: you do `dist.get_global_rank() % shard_size` stuff in a bunch of places. maybe define some descriptive variables for these mods and checks

composer/utils/checkpoint.py
```diff
@@ -510,34 +573,33 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+                process_group = device_mesh.get_group(1)
+                shard_size = device_mesh.size(1)
+                coordinator_rank = dist.get_global_rank() % shard_size
 
             if version.parse(torch.__version__) > version.parse('2.2.9'):
                 dist_cp.load(  # type: ignore
                     state_dict=state_dict,
                     storage_reader=storage_reader,
                     planner=load_planner,
                     process_group=process_group,
```
dakinggg, line 590: your PR description says you removed the pg from load, but i see it here? could you explain more

composer/trainer/trainer.py
```diff
@@ -1469,7 +1469,6 @@ def __init__(
                 # Set load arguments to defaults as this applies only to the initial non-autoresume
                 # load. We do not reset `load_strict_model_weights` for models with frozen layers.
                 load_weights_only = False
-                load_ignore_keys = None
```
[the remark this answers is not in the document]
mvpatel2000: I fixed this morning -- resolving

composer/utils/checkpoint.py
mvpatel2000: log.debug automatically captures rank for you

composer/utils/checkpoint.py
mvpatel2000: It looks in the filesystem which is shared across node, so only local rank 0 needs to do it.

composer/utils/checkpoint.py
dakinggg: Yes, but I think the text of the message is deceptive because it reads like the "everyone is done" message

composer/utils/checkpoint.py
dakinggg: ahh i see. I think the variable names i requested below will make this more clear :)

composer/utils/checkpoint.py
mvpatel2000: Yes. It's pretty fast though (<2 min at 100B+ scale)

dakinggg: This depends on your model size vs replica size though, so definitely could happen...maybe just jira because we don't have a mechanism for cross node busy wait.

dakinggg (commented):
i think lgtm but would like evan to review

eracah (approved):
Nice LGTM. Looks like the distcpreader still works as intended. Great work! Clever workaround. What a lift to get this to work!

PR #2920, opened 2024-01-29
Fix daily tests for peft + fsdp

dakinggg:
On torch 2.0, fsdp wrapped modules can not have both frozen and unfrozen params. On 2.1+, if you have `use_orig_params=True`, they can. So we need a special case for the tests here.

### DIFF
--- a/tests/models/test_hf_model.py
+++ b/tests/models/test_hf_model.py
@@ -477,8 +477,6 @@ def get_lm_trainer(hf_model,
                    should_save_peft_only: bool = False):
     transformers = pytest.importorskip('transformers')
 
-    torch_less_than_2 = version.parse(torch.__version__) < version.parse('2.1.0')
-
     metrics: List[Metric] = [LanguageCrossEntropy(ignore_index=-100)]
     if not is_conditional_generation:
         metrics.append(MaskedAccuracy(ignore_index=-100))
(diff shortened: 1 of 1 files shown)


tests/models/test_hf_model.py
```diff
@@ -477,6 +477,8 @@ def get_lm_trainer(hf_model,
                    should_save_peft_only: bool = False):
     transformers = pytest.importorskip('transformers')
 
+    torch_less_than_2 = version.parse(torch.__version__) < version.parse('2.1.0')
```
mvpatel2000, line 480: can we roll up this var given used in 1 place

dakinggg: ok fine :P

tests/models/test_hf_model.py
```diff
@@ -490,6 +490,14 @@ def get_lm_trainer(hf_model,
         should_save_peft_only=should_save_peft_only,
     )
 
+    if version.parse(torch.__version__) < version.parse('2.1.0') and peft_config is not None:
+        for name, module in model.named_modules():
+            if 'lora' in name.lower() and 'default' in name.lower():
+                has_parameters = any(True for _ in module.parameters())
+                has_buffers = any(True for _ in module.buffers())
+                if has_parameters or has_buffers:
+                    module._fsdp_wrap = True  # type: ignore
```
mvpatel2000, line 499: Can you add a comment please?

mvpatel2000 (approved):

PR #2923, opened 2024-01-30
Fix daily tests for peft on 1.13

dakinggg:
# What does this PR do?
The dailys surfaced an error with peft + fsdp that I haven't seen before
```
    def __setstate__(self, state):
        super().__setstate__(state)
        for group in self.param_groups:
            group.setdefault('amsgrad', False)
            group.setdefault('maximize', False)
            group.setdefault('foreach', None)
            group.setdefault('capturable', False)
        state_values = list(self.state.values())
>       step_is_tensor = (len(state_values) != 0) and torch.is_tensor(state_values[0]['step'])
E       KeyError: 'step'
```

Given that (a) we will not support 1.13 for much longer anyway, as everyone should be using torch 2+, and (b) there is no clean way to _only_ prevent peft usage on 1.13 when used in conjunction with FSDP, I decided to just gate peft usage on torch 2+.

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -23,7 +23,8 @@ from torchmetrics import Metric
 
 from composer.metrics import InContextLearningMetric, InContextLearningQAAccuracy
 from composer.models.base import ComposerModel
-from composer.utils import MissingConditionalImportError, dist, get_file, import_object, is_model_fsdp, safe_torch_load, using_torch_2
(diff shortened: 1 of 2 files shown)


dakinggg:
@j316chuck another PR to fix the dailys

mvpatel2000 (approved):

irenedea (approved):

PR #2925, opened 2024-01-30
Debug OOM error when 403 is caught

KuuCi:
This PR aims to debug an Oom error that appears when a 403 error is caught.


dakinggg:
Closing as no longer needed

PR #2931, opened 2024-01-30
Deprecate python3.8

j316chuck:
# What does this PR do?
Deprecate python3.8

# Related Issues
https://databricks.atlassian.net/browse/GRT-2659


mvpatel2000 (approved):
Can you also update the chip in the readme please

j316chuck:
I think the chip will be automatically updated once we release composer 0.19.0 since it comes from: https://img.shields.io/pypi/pyversions/mosaicml

PR #2935, opened 2024-01-30
Remove deprecations for 0.18

mvpatel2000:
# What does this PR do?

Remove deprecations for 0.18. Mostly models/datasets folders


dakinggg (approved):

antoinebrl:
Hello 👋 Would it be possible to know the motivations behind removing support for well known datasets (in various formats) and some models, including the integration with `timm`?

antoinebrl:
Thanks @mvpatel2000! As you said the creating model with `timm` requires calling two function so no need to provide any abstraction around this. Removing some elements your the API makes a lot of sense. I think some of it would have made good documentation materials to present your support and integration into the larger ecosystem.

PR #2945, opened 2024-01-31
Fix dailies by tweaking GPU versions and removing torch 1.13

mvpatel2000:
# What does this PR do?

Fix dailies by tweaking GPU versions and removing torch 1.13


b-chu:
Is it possible to rename the commit? The commit history isn't great now

PR #2947, opened 2024-01-31
Re-adding traceback formatting

jjanezhang:
# Re-adding traceback formatting

Somehow lost our rich formatting for tracebacks. Added it back to trainer.

## Testing 
image

Also catches foundry exceptions
Screenshot 2024-01-31 at 11 05 21 AM

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,8 @@
 
 version: 2
 updates:
-  - package-ecosystem: "pip" # See documentation for possible values
-    directory: "/" # Location of package manifests
-    schedule:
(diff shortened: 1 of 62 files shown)


dakinggg (approved):
LGTM, but lets check with @mvpatel2000 as well just to make sure no objections

mvpatel2000 (commented):
Honestly, I'm not a huge fan... I feel like this should be a user-facing decision as opposed to something Composer does. Eg I think foundry should do this but Composer should stay unopinionated

mvpatel2000 (changes requested):
Holding per discussion

PR #2952, opened 2024-02-01
Fix the daily notebook tests

dakinggg:
Previously the notebook install of `pip install mosaicml` was overwriting the install of `composer` for the `composer` package tests. This meant it got the latest release instead of the intended source code install. This PR patches the install to install the right package name based on what is installed.

Screenshot 2024-01-31 at 5 01 13 PM


j316chuck (approved):

j316chuck:
Can we remove the last two pr's code

PR #2953, opened 2024-02-01
Fix daily FSDP checkpoint tests

j316chuck:
# What does this PR do?
Fixes fsdp load old ckpt tests and pass daily tests again

# Tests
Passing dailies: https://github.com/mosaicml/composer/actions/runs/7735817067


mvpatel2000 (approved):
Good catch :)

.github/workflows/daily.yaml
```diff
@@ -8,6 +8,7 @@ on:
     - main
     - release/**
   workflow_dispatch:
+  pull_request:
```
dakinggg, line 11: do not merge

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -461,13 +467,76 @@ def test_fsdp_load_old_checkpoint(
...
 
     if (dist.get_global_rank() == 0 and state_dict_type == 'full') or state_dict_type == 'sharded':
-        filled_load_path = load_path.format(rank=dist.get_global_rank())
-        destination = str(tmp_path / pathlib.Path(filled_load_path).name)
-        get_file(filled_load_path, destination=destination)
-        with open(destination, 'rb') as f:
-            state_dict1 = torch.load(f)['state']
-        _compare_model_params_between_state_dicts(state_dict1, state_dict2)
+        if state_dict_type == 'sharded' and composer_version >= '0.16.0':
+            trainer2 = get_trainer(
```
mvpatel2000, line 471: why >= 0.16.0? We need a comment here likely

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -461,13 +467,76 @@ def test_fsdp_load_old_checkpoint(
...
+            process_group = None
+            dist_cp.load_state_dict(
+                state_dict=state_dict,
+                storage_reader=storage_reader,
+                planner=None,
+                process_group=process_group,
+            )
+            if version.parse(torch.__version__) < version.parse('2.2.9'):
+                from torch.distributed.checkpoint.optimizer import load_sharded_optimizer_state_dict
+                from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
```
mvpatel2000, line 509: Can we just call `load_sharded_checkpoint instead of copying logic? Especially because @eracah  pointed out this code has some issues and will be fixed soon? otherwise this will code rot

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -11,7 +11,7 @@
 import uuid
 from contextlib import nullcontext as does_not_raise
 from functools import partial
-from typing import Any, Callable, Optional, Sequence, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Union
```
b-chu, line 14: We don't need to import Dict anymore with 3.9

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -411,6 +411,9 @@ def test_fsdp_load_old_checkpoint(
     s3_read_only_prefix: str,
     composer_version: str,
 ):
+    if composer_version == '0.18.1' and state_dict_type == 'full' and precision == 'amp_bf16' and sharding_strategy == 'FULL_SHARD':
+        pytest.skip('TODO: This checkpoint is missing')
```
b-chu, line 415: Is this TODO for this PR or subsequent PR?

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -461,13 +467,76 @@ def test_fsdp_load_old_checkpoint(
     state_dict2 = trainer.state.state_dict()
 
     if (dist.get_global_rank() == 0 and state_dict_type == 'full') or state_dict_type == 'sharded':
-        filled_load_path = load_path.format(rank=dist.get_global_rank())
-        destination = str(tmp_path / pathlib.Path(filled_load_path).name)
-        get_file(filled_load_path, destination=destination)
-        with open(destination, 'rb') as f:
-            state_dict1 = torch.load(f)['state']
-        _compare_model_params_between_state_dicts(state_dict1, state_dict2)
+        if state_dict_type == 'sharded' and composer_version >= '0.16.0':
```
b-chu, line 470: Should we use version package instead of strings? e.g. https://github.com/mosaicml/composer/blob/dev/composer/trainer/mosaic_fsdp.py

j316chuck: fixed

tests/trainer/test_fsdp_checkpoint.py
j316chuck: subsequent PR

tests/trainer/test_fsdp_checkpoint.py
j316chuck: removed

tests/trainer/test_fsdp_checkpoint.py
j316chuck: @eracah we made some fixes to fsdp_load_checkpoint but I am not the expert in this code path so feel free to redesign the fix/test in a follow up PR.

PR #2956, opened 2024-02-01
Fix daily tests post resnet-cifar deprecation

mvpatel2000:
# What does this PR do?

Fix daily tests post resnet-cifar deprecation. Updates notebooks to not use resnet-cifar


dakinggg (approved):
A bit silly that we now have the resnet copied into every notebook...but not gonna block it.

j316chuck (approved):

PR #2958, opened 2024-02-01
Add OOM observer with memory visualizations

cli99:
This PR add a callback to generate visualizations of the state of allocated memory during an OutOfMemory exception. This callback registers an observer with the allocator that will be called everytime it is about to raise an OutOfMemoryError before any memory has been release while unwinding the exception. OOMObserver is attached to the Trainer at init stage. 

To enable this, in the yaml config,  do below, also need this change in the foundry https://github.com/mosaicml/llm-foundry/pull/932
```
callbacks:
    oom_observer:
        {
            "folder": "traces",
            "overwrite": true,
            "filename": "rank{rank}_oom",
            "remote_file_name": "oci://bucket_name/{run_name}/oom_traces/rank{rank}_oom",
        }
```
The visualizations include 
- snapshot of the memory state, `filename_snapshot.pickle`, can be used later for analysis with https://github.com/pytorch/pytorch/blob/main/torch/cuda/_memory_viz.py
- trace plot, `filename_trace_plot.html`
trace_plot

- segment plot, `filename_segment_plot.html`
segment_plot

- segment flamegraph, `filename_segment_flamegraph.svg`
segment_flamegraph

- memory flamegrap, filename_memory_flamegraph.svg`
memory_flamegraph


composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+            warnings.warn(f'OOMObserver only works on CUDA devices, but the model is on {model_device.type}.')
+            self._enabled = False
+        else:
+            self.folder_name = format_name_with_dist(self.folder, state.run_name)
+            os.makedirs(self.folder_name, exist_ok=True)
+            if not self.overwrite:
+                ensure_folder_is_empty(self.folder_name)
+
+        def oom_observer(device: int, alloc: int, device_alloc: int, device_free: int):
+            # snapshot right after an OOM happened
```
mvpatel2000, line 112:
```suggestion
            # Snapshot right after an OOM happened
```
nit

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,48 @@
...
+    oom_observer = OOMObserver()
+
+    simple_model = SimpleModel()
+
+    # Construct the trainer and train
+    _ = Trainer(
+        model=simple_model,
+        callbacks=oom_observer,
+        train_dataloader=DataLoader(RandomClassificationDataset()),
+        max_duration='2ba',
```
mvpatel2000, line 47: doesn't actually call fit?

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,48 @@
...
+            train_dataloader=DataLoader(RandomClassificationDataset()),
+            max_duration='1ba',
+        )
+
+
+@pytest.mark.gpu
+def test_oom_observer():
+    if version.parse(torch.__version__) <= version.parse('2.1.0.dev'):
+        # memory snapshot is supported after PyTorch 2.1.0.
+        return
```
mvpatel2000, line 36: can we use pytest.skip

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,48 @@
...
+    # Construct the callbacks
+    oom_observer = OOMObserver()
+
+    simple_model = SimpleModel()
+
+    # Construct the trainer and train
+    _ = Trainer(
+        model=simple_model,
+        callbacks=oom_observer,
```
mvpatel2000, line 45: can we simulate an OOM?

cli99: I don't thinks so. it needs a real OOM to trigger the torch cuda oom observer

tests/callbacks/test_oom_observer.py
cli99: added trainer.fit(), though it won't trigger OOM

tests/callbacks/test_oom_observer.py
mvpatel2000: hm... how do they trigger it? there is no way to mock it?

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+            _, _, self.remote_path_in_bucket = parse_uri(remote_file_name)
+        else:
+            self.remote_path_in_bucket = None
+
+        if version.parse(torch.__version__) > version.parse('2.1.0.dev'):  # type: ignore
+            # OOMObserver  is only supported in torch v2.1.0-rc1 or higher
+            self._enabled = True
+        else:
+            self._enabled = False
+            log.warning('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
```
j316chuck, line 94:
```suggestion
            warnings.warn('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
```

j316chuck:
@cli99 can you add one more test to make sure we can add both a `MemorySnapshot` and a `OOMSnapshot`. Just want to test the scenario where a customer uses both to understand their model and it works out of the box!

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def init(self, state: State, logger: Logger) -> None:
+        if not self._enabled:
+            return
+        # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
+        model_device = next(state.model.parameters()).device
```
j316chuck, line 100:
```suggestion
        model_device = state.device
```

j316chuck: Please lmk if there's a reason why we aren't using `state.device`?

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+            ... )
+
+    .. note::
+        OOMObserver is only supported for GPU devices.
+
+    Args:
+        max_entries (int, optional): Maximum number of memory alloc/free events to record. Defaults to 100000.
+        folder (str, optional): A format string describing the folder containing the memory visualization files.
+            Defaults to ``'{{run_name}}/torch_traces'``.
+        filename (str, optional): A format string describing the prefix used to name the memory visualization files.
```
j316chuck, line 50: Slight preference for naming things as _oom instead of .oom but will let you make the call on that

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+                    fd.write(torch.cuda._memory_viz.memory(snapshot))  # type: ignore
+
+                log.info(f'Saved memory visualizations to local files with prefix = {filename} during OOM')
+
+                if self.remote_path_in_bucket is not None:
+                    for f in [
+                            snapshot_file, trace_plot_file, segment_plot_file, segment_flamegraph_file,
+                            memory_flamegraph_file
+                    ]:
+                        remote_file_name = (self.remote_path_in_bucket + os.path.split(f)[-1]).lstrip('/')
```
j316chuck, line 157:
```suggestion
                        remote_file_name = (self.remote_path_in_bucket + os.path.basename(f)
```

j316chuck (commented):
Looks great @cli99 barring a couple of nits and the test case  above 🎉 ! 

Also if you could figure out how to mock/add the unit test Mihir described that would be awesome but imo not blocking

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Args:
+        max_entries (int, optional): Maximum number of memory alloc/free events to record. Defaults to 100000.
+        folder (str, optional): A format string describing the folder containing the memory visualization files.
+            Defaults to ``'{{run_name}}/torch_traces'``.
+        filename (str, optional): A format string describing the prefix used to name the memory visualization files.
+            Defaults to ``'rank{{rank}}.oom'``.
+        remote_file_name (str, optional): A format string describing the prefix for the memory visualization remote file name.
+            Defaults to ``'{{run_name}}/oom_traces/rank{{rank}}.oom'``.
+
```
mvpatel2000, line 54: should this be `.oom`? is that the file extension that should be there...?

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,175 @@
...
+            _, _, self.remote_path_in_bucket = parse_uri(remote_file_name)
+        else:
+            self.remote_path_in_bucket = None
+
+        if version.parse(torch.__version__) > version.parse('2.1.0.dev'):  # type: ignore
+            # OOMObserver  is only supported in torch v2.1.0-rc1 or higher
+            self._enabled = True
+        else:
+            self._enabled = False
+            warnings.warn('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
```
mvpatel2000, line 94:
```suggestion
        if version.parse(torch.__version__.split('.dev')[0]) >= version.parse('2.1.0'):  # type: ignore
            # OOMObserver  is only supported in torch v2.1.0-rc1 or higher
            self._enabled = True
        else:
            self._enabled = False
            warnings.warn('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
```
thoughts? im thinking we redo repo to do this to avoid this scenario

tests/callbacks/test_oom_observer.py
mvpatel2000: unresolving since I don't think it was changed...?

mvpatel2000: actually the correct version is to put:
```
@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.1.0'),
                    reason='OOM Observer requires PyTorch 2.1 or higher')
```
on top of function as a mark

mvpatel2000 (commented):
LGTM, just some minor concerns :)

tests/callbacks/test_oom_observer.py
cli99: I created a actual OOM and see if it passes ci.

composer/callbacks/oom_observer.py
cli99: updated the names to use _oom

cli99:
> @cli99 mind adding a unit test to make sure we can add both a `MemorySnapshot` and an `OOMSnapshot` callback. Just want to test the scenario where someone calls `torch.cuda._memory.snapshot` twice.

added a test to show the two callbacks are compatible. Both `torch.cuda._memory.snapshot` and `torch.cuda.memory._record_memory_history` allows multiple calls. @j316chuck

tests/callbacks/test_oom_observer.py
cli99: forgot to push. : ) . just changed it as suggested

cli99:
@j316chuck, using warning.warn causes issues in a bunch cpu callback tests, where they don't not catch user warnings, so I switch to log.warning

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,177 @@
...
+        self.remote_file_name = remote_file_name
+        self.overwrite = overwrite
+        if remote_file_name:
+            self.remote_file_name = remote_file_name
+            _, _, self.remote_path_in_bucket = parse_uri(remote_file_name)
+        else:
+            self.remote_path_in_bucket = None
+
+        if version.parse(torch.__version__.split('.dev')[0]) >= version.parse('2.1.0'):  # type: ignore
+            # OOMObserver  is only supported in torch v2.1.0-rc1 or higher
```
mvpatel2000, line 90:
```suggestion
            # OOMObserver is only supported in torch v2.1.0 or higher
```

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,177 @@
...
+            self._enabled = True
+        else:
+            self._enabled = False
+            log.warning('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
```
mvpatel2000, line 94:
```suggestion
            warnings.warn('OOMObserver is supported after PyTorch 2.1.0. Disabling OOMObserver callback.')
```
I think this should be warnings.warn given it may require user action.

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,177 @@
...
+            self._enabled = False
+        else:
+            self.folder_name = format_name_with_dist(self.folder, state.run_name)
+            os.makedirs(self.folder_name, exist_ok=True)
+            if not self.overwrite:
+                ensure_folder_is_empty(self.folder_name)
+
+        def oom_observer(device: int, alloc: int, device_alloc: int, device_free: int):
+            # Snapshot right after an OOM happened
+            log.warn('Out Of Memory (OOM) observed')
```
mvpatel2000, line 115:
```suggestion
            log.warning('Out Of Memory (OOM) observed')
```
log.warn is deprecated

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,177 @@
...
+            log.warning('OOMObserver is supported after PyTorch 2.1.0. Skipping oom observer callback.')
+
+    def init(self, state: State, logger: Logger) -> None:
+        if not self._enabled:
+            return
+        # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
+        model_device = next(state.model.parameters()).device
+
+        if model_device.type not in ('cuda', 'meta'):
+            log.warning(
```
mvpatel2000, line 103:
```suggestion
            warnings.warn(
```

composer/callbacks/oom_observer.py
```diff
@@ -0,0 +1,177 @@
...
+                f'OOMObserver only works on CUDA devices, but the model is on {model_device.type}. OOMObserver is disabled'
```
mvpatel2000, line 104:
```suggestion
                f'OOMObserver only works on CUDA devices, but the model is on {model_device.type}. Disabling OOMObserver.'
```
Nit: period + consistent actionable verb like with other comments

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,97 @@
...
+from composer.callbacks import MemorySnapshot, OOMObserver
+from composer.loggers import LoggerDestination
+from composer.trainer import Trainer
+from tests.common import RandomClassificationDataset, SimpleModel, device
+
+
+@device('cpu', 'gpu')
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.1.0'),
+                    reason='OOM Observer requires PyTorch 2.1 or higher')
+def test_oom_observer_warnings_on_cpu_models(device: str):
```
mvpatel2000, line 21: I don't understand this test. Why are we also running on GPU but with device CPU?

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,97 @@
...
+    def upload_file(self, state: State, remote_file_name: str, file_path: pathlib.Path, *, overwrite: bool):
+        del state, overwrite  # unused
+        self.uploaded_files.append((remote_file_name, file_path))
+
+
+@pytest.mark.gpu
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.1.0'),
+                    reason='OOM Observer requires PyTorch 2.1 or higher')
+def test_oom_observer():
+
```
mvpatel2000, line 50:
```suggestion
```
nit

mvpatel2000 (commented):
The test is really nice!

I have some minor clean up comments but PR LGTM

composer/callbacks/oom_observer.py
cli99: warnings.warn will break cpu callback tests where user warnings are not catched. If training without oom observer support due to older pt version, the user shall just get notified in the log, no action shall be expected.

tests/callbacks/test_oom_observer.py
cli99: removed the gpu for this test

composer/callbacks/oom_observer.py
mvpatel2000: I believe the CPU tests let you attach marks to filter these warnings?

mvpatel2000: I don't have a super strong opinion here. But, I do think we should be consistent, and elsewhere we seem to prefer warnings.warn?

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,95 @@
...
+    del device  # unused. always using cpu
+    ob = OOMObserver()
+    Trainer(
+        model=SimpleModel(),
+        callbacks=ob,
+        device='cpu',
+        train_dataloader=DataLoader(RandomClassificationDataset()),
+        max_duration='1ba',
+    )
+    assert ob._enabled is False
```
mvpatel2000, line 32:
```suggestion
def test_oom_observer_warnings_on_cpu_models():
    ob = OOMObserver()
    Trainer(
        model=SimpleModel(),
        callbacks=ob,
        train_dataloader=DataLoader(RandomClassificationDataset()),
        max_duration='1ba',
    )
    assert ob._enabled is False
```
I think other tests instead add a pytest context and check that it raises the warning it expects?

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Construct the callbacks
+    oom_observer = OOMObserver()
+    memory_snapshot = MemorySnapshot(skip_batches=0, interval='1ba')
+
+    simple_model = SimpleModel()
+
+    file_tracker_destination = FileUploaderTracker()
+
+    trainer = Trainer(
```
mvpatel2000, line 86:
```suggestion
def test_oom_observer_with_memory_snapshot():
    # Construct the callbacks
    oom_observer = OOMObserver()
    memory_snapshot = MemorySnapshot(skip_batches=0, interval='1ba')
    simple_model = SimpleModel()
    file_tracker_destination = FileUploaderTracker()

    trainer = Trainer(
```
nit :P not strongly held opinion

composer/callbacks/oom_observer.py
dakinggg: warnings.warn is for when the user can do something, log.warning is when they can't. should use `pytest.mark.filterwarnings` to ignore

tests/callbacks/test_oom_observer.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.mark.gpu
+@pytest.mark.skipif(version.parse(torch.__version__) < version.parse('2.1.0'),
+                    reason='OOM Observer requires PyTorch 2.1 or higher')
+def test_oom_observer():
+    # Construct the callbacks
+    oom_observer = OOMObserver()
+
+    simple_model = SimpleModel()
+
+    file_tracker_destination = FileUploaderTracker()
```
mvpatel2000, line 54:
```suggestion
    oom_observer = OOMObserver()
    simple_model = SimpleModel()
    file_tracker_destination = FileUploaderTracker()
```
same minor nit for more concise code

composer/callbacks/oom_observer.py
dakinggg: see https://docs.python.org/3/howto/logging.html

cli99: do we expect users to take any actions? I feel pt version is not something users will switch just because they saw oom observer is not supported. More like they use some pt version and are told oom observer is not supported and disabled.

mvpatel2000: I think this is an actionable task. It is "avoidable" (from the logging docs)

dakinggg: Yeah, this includes anything that the user _could_ do something about, not taking into account the likelihood of them doing that (at least that is my interpretation). `log.warning` is when the library just wants to tell you something important, but youcan't do anything about it.

cli99: sounds good. will make the change

mvpatel2000 (approved):
LGTM!

Sorry for being a bit nit heavy on reviews 😂

PR #2959, opened 2024-02-02
Modify UCObjectStore.list_objects to lists all files recursively

irenedea:
# What does this PR do?
Modifies `UCObjectStore.list_objects` to list all files recursively. Previously, `UCObjectStore.list_objects` only listed top level files.

This is consistent with other `ObjectStore` behavior. 

# Tests

(1) Manually tested with:
```
from composer.utils.object_store import UCObjectStore

uc = UCObjectStore(path="Volumes/main/nancyhung/cpt/small")
res = uc.list_objects(prefix="Volumes/main/nancyhung/cpt/small")
print('\n'.join(res))
```
Output:
```
/Volumes/main/nancyhung/cpt/small/AAA-fake/foo/chat.jsonl
/Volumes/main/nancyhung/cpt/small/ABT/sec_0000001800_10-K_1993_metadata.json
/Volumes/main/nancyhung/cpt/small/ABT/sec_0000001800_10-K_1993_txt.txt
/Volumes/main/nancyhung/cpt/small/ABT/sec_0000001800_10-K_1994_metadata.json
/Volumes/main/nancyhung/cpt/small/ABT/sec_0000001800_10-K_1994_txt.txt
...
```
I checked that each subfolder was represented.

Before these changes, there was no output printed because there were no top level files.

(2) Manually tested that an exception is raised when the max recursive depth is reached.


composer/utils/object_store/uc_object_store.py
```diff
@@ -234,14 +230,30 @@ def list_objects(self, prefix: Optional[str]) -> List[str]:
...
-            assert isinstance(resp, dict)
-            return [f['path'] for f in resp.get('files', []) if not f['is_dir']]
+            logging.warn('UCObjectStore.list_objects is experimental.')
+            max_recursion_depth = 4
+
+            def get_uc_files(dir_path: str, recursion_depth: int = 0) -> list[str]:
+                if recursion_depth == max_recursion_depth:
+                    raise Exception(
+                        f'Objects at {dir_path} cannot be downloaded. Please reduce the' +
+                        ' level of folder nesting from {prefix} in UC Volumes to under {max_recursion_depth}.')
```
irenedea, line 241:
```suggestion

            def get_uc_files(dir_path: str, recursion_depth: int = 0) -> list[str]:
```

I don't feel very strongly about this recursive depth check. imo it's a weak proxy for too many calls to the api, and we don't really have any evidence that it's valuable or necessary. On the other hand, it could cause a surprising failure for users.

irenedea: Other thoughts or opinions here? @nancyhung @dakinggg @panchalhp-db

panchalhp-db, line 241:
```suggestion
                        f' level of folder nesting from {prefix} in UC Volumes to under {max_recursion_depth}.')
```

panchalhp-db: I agree that a folder nested at depth 1 could also result in a lot of API calls and just capping the depth does not solve that problem. That said having this unbounded makes me a little worried especially since we don't have metrics to track how many API calls / depths this goes to on average with a user. We could keep the max depth something high like 10 if it helps so practically no one should hit it (as  @dakinggg suggested) but it still guards us against absurd errors like StackOverflow or MaxRecursionDepth.

Also might be a good idea to run this by the UC volume folks.

irenedea: Messaged the UC volume folks about this. I can also rewrite this iteratively to avoid any stackoverflow or maxrecursiondepth issues.

irenedea: After discussing offline with @panchalhp-db and the UC volume folks, we will make sure to handle any 429 errors with retries and do not need bound the recursion depth. 

I made two changes:
(1) Rewrote this iteratively to avoid any StackOverflow or MaxRecursionDepth errors.
(2) The databricks SDK handles http errors and retries under the hood, so I just added a comment to document.

composer/utils/object_store/uc_object_store.py
```diff
@@ -234,14 +230,35 @@ def list_objects(self, prefix: Optional[str]) -> List[str]:
...
         try:
-            data = json.dumps({'path': self._get_object_path(prefix)})
             # NOTE: This API is in preview and should not be directly used outside of this instance
-            resp = self.client.api_client.do(method='GET',
-                                             path=self._UC_VOLUME_LIST_API_ENDPOINT,
-                                             data=data,
-                                             headers={'Source': 'mosaicml/composer'})
-            assert isinstance(resp, dict)
-            return [f['path'] for f in resp.get('files', []) if not f['is_dir']]
+            logging.warn('UCObjectStorelist_objects is experimental.')
```
dakinggg, line 234:
```suggestion
            logging.warn('UCObjectStore.list_objects is experimental.')
```

tests/utils/object_store/test_uc_object_store.py
```diff
@@ -160,6 +160,50 @@ def generate_dummy_file(_):
...
+    }]
+
+    prefix = 'Volumes/catalog/schema/volume/path/to/folder'
+
+    ws_client.api_client.do = MagicMock(side_effect=[uc_list_api_responses[0], uc_list_api_responses[1]])
+    actual_files = uc_object_store.list_objects(prefix=prefix)
+
+    assert actual_files == expected_files
+
+    print(ws_client.api_client.do.call_args_list)
```
dakinggg, line 198:
```suggestion
```

dakinggg (approved):

panchalhp-db (approved):
LGTM!

PR #2961, opened 2024-02-03
Adding the step argument to logger.log_table

ShashankMosaicML:
Currently, `logger.log_table` does not provide a `step` argument to the loggers. This creates a problem when using wandb logger, which increments the `step` while logging if it is not provided a `step`. Hence, if we want to log both metrics and tables, the metrics go out of sync with the actual steps. The fix is similar to [what we do for `logger.log_metrics`](https://github.com/mosaicml/composer/blob/99289adb14b70fcf1a3b51d15035e7c7f84eda59/composer/loggers/logger.py#L73), we start providing the step explicitly to the wandb logger for `log_table` as well.


mvpatel2000 (commented):
Please provide a PR description when it's ready for review! Eg why is this needed

ShashankMosaicML:
> Please provide a PR description when it's ready for review! Eg why is this needed

Done!

mvpatel2000 (approved):
LGTM thanks!

PR #2965, opened 2024-02-06
Refactor `update_metric`

maxisawesome:
# What does this PR do?

Refactors `huggingface.py`'s `update_metric` to not require checking the metric's type.

# Before submitting


bmosaicml (approved):
I approve of this design and think it would be the best way to remove the InContextLearning classes from composer. Thanks Max!

eracah (approved):
LGTM. So some metrics need the batch because some metric configs are stored in the batch? hmmm seems like there should be a cleaner way...

dakinggg (changes requested):
I suggested an alternate design on the doc, blocking until we've discussed a bit.

composer/models/huggingface.py
```diff
@@ -532,12 +532,8 @@ def get_metrics(self, is_train: bool = False) -> Dict[str, Metric]:
         return metrics if metrics else {}
 
     def update_metric(self, batch: Any, outputs: Any, metric: Metric) -> None:
-        if isinstance(metric, InContextLearningQAAccuracy):
-            assert self.labels is not None
+        if getattr(metric, 'requires_batch', False):
```
dakinggg, line 535: attr is named wrong, its `need_batch`

dakinggg (approved):

bmosaicml:
> LGTM. So some metrics need the batch because some metric configs are stored in the batch? hmmm seems like there should be a cleaner way...

It's not just global configs stored (though that is the case for some metrics, and def not good design), but also batch level info, e.g. which indices of the output correspond to the expected answer. The batch level info is probably harder to factor out.

composer/metrics/nlp.py
```diff
@@ -196,6 +196,10 @@ def compute(self) -> Tensor:
 
 class InContextLearningMetric(Metric):
 
+    def __init__(self, *args, **kwargs):
+        super.__init__(*args, **kwargs)
```
bmosaicml, line 200: Should it be `super().__init__` ?

composer/metrics/nlp.py
```diff
@@ -196,6 +196,10 @@ def compute(self) -> Tensor:
 
 class InContextLearningMetric(Metric):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
bmosaicml, line 200: Added missing parentheses

composer/models/huggingface.py
```diff
@@ -532,12 +531,8 @@ def get_metrics(self, is_train: bool = False) -> Dict[str, Metric]:
         return metrics if metrics else {}
 
     def update_metric(self, batch: Any, outputs: Any, metric: Metric) -> None:
-        if isinstance(metric, InContextLearningQAAccuracy):
-            assert self.labels is not None
+        if getattr(metric, 'needs_batch', False):
             metric.update(batch=batch, outputs=outputs, labels=self.labels)  # pyright: ignore [reportGeneralTypeIssues]
```
dakinggg, line 535: can you see if this type ignore can be removed now?

dakinggg (approved):

dakinggg (changes requested):
Blocking because we found a bug.

composer/metrics/nlp.py
```diff
@@ -328,9 +332,23 @@ def __init__(self, dist_sync_on_step: bool = False):
...
+               labels: Optional[torch.Tensor] = None,
+               outputs: Optional[torch.Tensor] = None):
+        if output_logits is not None:
+            warnings.warn(
+                ('`output_logits` has been deprecated and will be removed in a future release'),
+                DeprecationWarning,
+            )
+            outputs = output_logits
+
+        assert labels is not None
```
dakinggg, line 347: this should raise with clear error if not none, rather than just assert

composer/metrics/nlp.py
```diff
@@ -196,6 +196,10 @@ def compute(self) -> Tensor:
 
 class InContextLearningMetric(Metric):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.needs_batch = True
+
     def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
```
dakinggg, line 203: signature of the base class needs to change as well

composer/metrics/nlp.py
```diff
@@ -328,9 +332,23 @@ def __init__(self, dist_sync_on_step: bool = False):
         self.add_state('correct', default=torch.tensor(0.), dist_reduce_fx='sum')
         self.add_state('total', default=torch.tensor(0.), dist_reduce_fx='sum')
 
-    def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+    def update(self,
+               batch: dict,
+               output_logits: Optional[torch.Tensor] = None,
+               labels: Optional[torch.Tensor] = None,
+               outputs: Optional[torch.Tensor] = None):
+        if output_logits is not None:
```
dakinggg, line 340: please pull this arg checking logic out into a helper function that returns the correct arguments so that the logic only exists in one place

bmosaicml (approved):

composer/metrics/nlp.py
```diff
@@ -196,7 +196,15 @@ def compute(self) -> Tensor:
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.needs_batch = True
+
+    def update(self,
+               batch: dict,
+               output_logits: Optional[torch.Tensor] = None,
+               labels: Optional[torch.Tensor] = None,
+               outputs: Optional[torch.Tensor] = None):
         """Abstract interface for computing an in-context learning metrics.
```
dakinggg, line 208: Leave a comment in the docstring that the args are changing and `output_logits` will be removed in v0.21

composer/metrics/nlp.py
```diff
@@ -210,6 +218,27 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
         """
         raise NotImplementedError
 
+    @classmethod
+    def rename_args(
+            cls,
```
dakinggg, line 223: since you don't use `cls`, this should be `@staticmethod`, yeah?

composer/metrics/nlp.py
```diff
@@ -210,6 +218,27 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
+            batch: dict,
+            output_logits: Optional[torch.Tensor] = None,
+            labels: Optional[torch.Tensor] = None,
+            outputs: Optional[torch.Tensor] = None) -> Tuple[dict, Optional[torch.Tensor], torch.Tensor, torch.Tensor]:
+        if output_logits is not None:
+            warnings.warn(
+                ('`output_logits` has been renamed to `outputs   and will be removed in a future release'),
```
dakinggg, line 230:
```suggestion
                ('`output_logits` has been renamed to `outputs   and will be removed in v0.21'),
```

composer/metrics/nlp.py
```diff
@@ -210,6 +218,27 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
+                DeprecationWarning,
+            )
+            outputs = output_logits
+
+        if labels is None:
+            raise ValueError('`labels` cannot be None')
+        if outputs is None:
+            raise ValueError('`outputs` cannot be None')
+
+        return batch, output_logits, labels, outputs
```
dakinggg, line 240: this should just return the three args we actually need, not `output_logits`

composer/metrics/nlp.py
```diff
@@ -266,9 +299,8 @@ def replace_underscore(text: str) -> str:
 
         return white_space_fix(remove_articles(handle_punc(lower(replace_underscore(answer))))).strip()
 
-    def update(self, outputs: List[str], labels: List[List[str]], batch: Optional[Dict[str, Any]] = None):
-        if batch is None:
-            batch = {}
+    # TODO: rearrange these? why could batch be optional?
```
maxisawesome, line 302: @dakinggg what do you think about making this change now? everywhere else has batch first

dakinggg: wow thats annoying. hm...technically thats breaking, but maybe making the signature correct with the base class is more important and its unlikely to actually break anyone... @mvpatel2000 wdyt?

dakinggg: @maxisawesome I vote we leave as is for now, and when we actually remove `output_logits`, we make the breaking change here too so that it matches the base class signature. and then call it out in the release notes.

composer/metrics/nlp.py
```diff
@@ -210,6 +221,27 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
+    @staticmethod
+    def rename_args(batch: dict,
+                    output_logits: Optional[torch.Tensor] = None,
+                    labels: Optional[torch.Tensor] = None,
+                    outputs: Optional[torch.Tensor] = None) -> Tuple[dict, torch.Tensor, torch.Tensor]:
+        if outputs is not None and output_logits is not None:
+            raise ValueError('Cannot use both `outputs` and `output_logits`')
+        if output_logits is not None:
+            warnings.warn(
+                ('`output_logits` has been renamed to `outputs` and will be removed in a future release'),
```
dakinggg, line 233:
```suggestion
                ('`output_logits` has been renamed to `outputs` and will be removed in v0.21'),
```

composer/metrics/nlp.py
```diff
@@ -424,6 +473,7 @@ def __init__(self, dist_sync_on_step: bool = False, n_buckets: int = 10):
         self.add_state('bucket_correct', default=torch.zeros(n_buckets), dist_reduce_fx='sum')
 
     def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor):
+        # TODO: fix
```
dakinggg, line 476: what is going on here?

composer/metrics/nlp.py
maxisawesome: Okay, that works

composer/metrics/nlp.py
maxisawesome: removed

composer/metrics/nlp.py
```diff
@@ -266,9 +298,7 @@ def replace_underscore(text: str) -> str:
 
         return white_space_fix(remove_articles(handle_punc(lower(replace_underscore(answer))))).strip()
 
-    def update(self, outputs: List[str], labels: List[List[str]], batch: Optional[Dict[str, Any]] = None):
-        if batch is None:
-            batch = {}
+    def update(self, outputs: List[str], labels: List[List[str]], batch: Dict[str, Any]):
```
dakinggg, line 301: Is the batch not optional? it seems optional cause everything accessed on it is defaulted, and you don't have to specify any of those things

dakinggg (approved):
One comment, but otherwise LGTM

composer/metrics/nlp.py
maxisawesome: Only place it's called is in our original update_metric function which will always pass it a batch. Based on the code though, it's not optional (can't call .get on None)
I will note that this is a prime example jamming stuff in the batch that really is used in metric computation.

dakinggg: well it was optional cause it used to do
```
if batch is None:
    batch = {}
```

maxisawesome: we decided to fix it in a later release

PR #2966, opened 2024-02-06
Add torch 2.2 + python3.11 tests

j316chuck:
# What does this PR do?
Add torch 2.2 + python3.11 tests

# Tests
- Daily: https://github.com/mosaicml/composer/actions/runs/7809790673

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -33,6 +33,11 @@ jobs:
           markers: not daily and (remote or not remote) and not gpu and not doctest
           pytest_command: coverage run -m pytest
           composer_package_name: composer
+        - name: cpu-3.11-2.2
+          container: mosaicml/pytorch:2.2.0_cpu-python3.11-ubuntu20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -18,6 +18,11 @@ jobs:
           markers: not daily and not remote and gpu and (doctest or not doctest)
           pytest_command: coverage run -m pytest
           composer_package_name: mosaicml
+        - name: gpu-3.11-2.2
+          container: mosaicml/pytorch:2.2.0_cu121-python3.11-ubuntu20.04
+          markers: not daily and not remote and gpu and (doctest or not doctest)
+          pytest_command: coverage run -m pytest
+          composer_package_name: mosaicml
```
mvpatel2000, line 25: Let's keep only on daily for now to minimize load and only keep `-latest` tag (2.1) for per run?

mvpatel2000 (commented):
can u add a daily one too?

PR #2967, opened 2024-02-06
Add a register_model_with_run_id api to MLflowLogger

dakinggg:
This is a different way or registering the model that is necessary if you want to preserve the link to the mlflow run.

Manually tested that the link to the run is preserved.
Screenshot 2024-02-06 at 6 26 10 PM


composer/loggers/mlflow_logger.py
```diff
@@ -346,6 +346,55 @@ def log_model(self, flavor: Literal['transformers'], **kwargs):
...
                 raise NotImplementedError(f'flavor {flavor} not supported.')
 
+    def create_model_version(
+        self,
+        model_uri: str,
+        name: str,
+        await_creation_for: int = 300,
+        tags: Optional[Dict[str, Any]] = None,
+    ):
+        """Similar to MLflow's create_model_version, but will also create the initial model if it does not exist.
```
irenedea, line 356: Why do we need to create the initial model if it does not exist?

composer/loggers/mlflow_logger.py
```diff
@@ -346,6 +346,55 @@ def log_model(self, flavor: Literal['transformers'], **kwargs):
...
+            # https://github.com/mlflow/mlflow/blob/3ba1e50e90a38be19920cb9118593a43d7cfa90e/mlflow/tracking/_model_registry/fluent.py#L90-L103
+            try:
+                create_model_response = self._mlflow_client.create_registered_model(full_name)
+                log.info(f'Successfully registered model {name} with {create_model_response.name}')
+            except MlflowException as e:
+                if e.error_code in (
+                        ErrorCode.Name(RESOURCE_ALREADY_EXISTS),
+                        ErrorCode.Name(ALREADY_EXISTS),
+                ):
+                    log.info(f'Registered model {name} already exists. Creating a new version of this model...')
```
irenedea, line 382: So even if L375 (`create_model_response = self._mlflow_client.create_registered_model(full_name)`) raises an exception, a new model version is created anyway?

irenedea: (A little surprised to see that there isn't extra logic after this line to create a new model)

composer/loggers/mlflow_logger.py
```diff
@@ -346,6 +346,55 @@ def log_model(self, flavor: Literal['transformers'], **kwargs):
...
+    def create_model_version(
+        self,
+        model_uri: str,
+        name: str,
+        await_creation_for: int = 300,
+        tags: Optional[Dict[str, Any]] = None,
+    ):
+        """Similar to MLflow's create_model_version, but will also create the initial model if it does not exist.
+
+        This function differs from ``register_model`` in that it also uses the run id.
```
irenedea, line 358: Why happens [here](https://github.com/mlflow/mlflow/blob/3ba1e50e90a38be19920cb9118593a43d7cfa90e/mlflow/tracking/_model_registry/fluent.py#L109)? Is that run_id not valid or is the mdoel_uri not a runs uri?

composer/loggers/mlflow_logger.py
```diff
@@ -346,6 +346,55 @@ def log_model(self, flavor: Literal['transformers'], **kwargs):
             else:
                 raise NotImplementedError(f'flavor {flavor} not supported.')
 
+    def create_model_version(
```
irenedea, line 349: where do we plan to call this function?

composer/loggers/mlflow_logger.py
irenedea: Ah, I looked at the mlflow code more and it make sense to me. It looks like it's more similar to `register_model`. It seems like the only difference is that it uses the run_id directly from the logger instead of parsing it from model_uri?

Maybe rename the function to `register_model_with_run_id`?

irenedea (commented):
Looks pretty good! Just had some questions

composer/loggers/mlflow_logger.py
irenedea: Looking at the mlflow code this makes sense.

composer/loggers/mlflow_logger.py
dakinggg: https://github.com/mosaicml/llm-foundry/pull/953

composer/loggers/mlflow_logger.py
dakinggg: yeah, thats fair. I called it this cause originally it was just a wrapper around `create_model_version`, but then I discovered I had to do the try/except thing to create if not exists. Will update.

composer/loggers/mlflow_logger.py
dakinggg: Yeah because in our case the uri is a local path, there is not run to parse out of it.

composer/loggers/mlflow_logger.py
irenedea: Can you update the docstring to just say that it's similar to register model?

irenedea (approved):
LGTM!

eracah (approved):

PR #2968, opened 2024-02-07
Memory snapshot dump pickle

cli99:
This PR dump the raw pickle memory snapshot along with the post-processed html file, so that we can perform analysis on pickle later.


mvpatel2000 (approved):
What's the pickle file used for? Like what additional info does it have?

j316chuck (approved):

PR #2969, opened 2024-02-07
Make CodeEval respect device_eval_batch_size

josejg:
# What does this PR do?

Re-implementation of our CodeEval so that it respects the device eval batch size

Since the previous implementation silently overrode device_eval_batch_size to be generations_per_sample, there are two clear benefits of the rewrite:

- Better utilization when generations_per_sample < device_eval_batch_size 
- Prevents OOM when generations_per_sample >> device_eval_batch_size 

# Regression testing

These are some regression tests using the PR code and the previous implementation (plus a patch to use temperature 0.2) which was not possible before @maxisawesome foundry PR (which was not merged at the time I ran these experiments). This experiments are with generations_per_sample=20 so some variance is expected.

| Model                                |   This PR |   Before |
|:-------------------------------------|----------:|-----------:|
| meta-llama/Llama-2-7b-hf             |     12.47 |      13.69 |
| codellama/CodeLlama-7b-hf            |     27.99 |      27.68 |
| codellama/CodeLlama-13b-hf           |     30.73 |      30.37 |
| deepseek-ai/deepseek-coder-6.7b-base |     39.82 |      40.3  |
| codellama/CodeLlama-34b-hf           |     41.31 |      42.29 |
| deepseek-ai/deepseek-coder-33b-base  |     47.96 |      47.56 |
| Phind/Phind-CodeLlama-34B-v2         |     65.12 |      65.27 |

# Before submitting


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -1234,11 +1235,13 @@ class InContextLearningCodeEvalDataset(InContextLearningDataset):
     def __init__(
         self,
         generations_per_sample: int,
-        pass_at_k: int = 1,
+        pass_at_k: Union[int, list[int]] = 1,
```
dakinggg, line 1238: why is pass at k a list now?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -18,9 +18,10 @@
 from composer.datasets.utils import stop_sequences_criteria
 from composer.utils import MissingConditionalImportError, dist, get_file
 
+from datasets import Dataset as HFDataset  # pyright: ignore[reportGeneralTypeIssues]
```
dakinggg, line 21: this can't be at the top level because datasets is an option dependency

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -1288,15 +1294,26 @@ def __init__(
...
+            'sample_id': [],
+            'pass_at_k': list(pass_at_k),
+            'generations_per_sample': generations_per_sample,
+            'dataset_size': dataset_size,
         }
         if 'generation_kwargs' in kwargs:
             self.update_generation_kwargs(kwargs['generation_kwargs'])
 
+    def repeat_dataset(self, dataset: HFDataset, repetitions: int) -> HFDataset:
+        df = dataset.to_pandas()
```
dakinggg, line 1311: I think this is a little bit dangerous because it puts the whole dataset in memory, something that is otherwise avoided because everything else uses arrow datasets. Is there an alternative approach that doesn't required materializing N copies of the dataset in memory.

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
             labels (List[str]): A list of the correct code generations, for compatibility with existing HF generate
             functionalities. This is not used.
         """
+        if not self._initialized:
+            # we need to defer the accumulator initialization because it depends on dataset size
```
dakinggg, line 609:
```suggestion
            # We need to defer the accumulator initialization because it depends on dataset size
```

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+            dataset_size = batch['dataset_size']
+            device = batch['input_ids'].device
+            self.add_state('correct', default=torch.zeros(dataset_size, device=device), dist_reduce_fx='sum')
```
dakinggg, line 612: how come the tensors in the state are now on device? were they just on cpu before? is there any reason from them to be on gpu?

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
             labels (List[str]): A list of the correct code generations, for compatibility with existing HF generate
             functionalities. This is not used.
         """
+        if not self._initialized:
```
dakinggg, line 608: break out initializing into a helper function

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+            for test_input, test_output in zip(test_inputs, test_outputs):
+                payload = {
+                    'code': final_code,
+                    'input': test_input,
+                    'output': test_output,
+                    'entry_point': entry_point,
+                    'language': language,
+                }
+
+                # TODO: refactor client.invoke API
```
dakinggg, line 643: whats the todo here?

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+                result = client.invoke([[[payload]]])[0][0][0]
```
dakinggg, line 644: why the change of structure to invoke the client one payload at a time? was the other approach causing problems?

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+            if all(test_results):
+                self.correct[idx] += 1.0
 
         client.close()  # pyright: ignore [reportOptionalMemberAccess]
 
     def compute(self):
         assert isinstance(self.correct, Tensor)
         assert isinstance(self.total, Tensor)
-        return self.correct / self.total
+        assert (self.total == self.num_generations).all().item()
```
dakinggg, line 655: this should probably raise a nice error if it fails

composer/metrics/nlp.py
```diff
@@ -604,51 +605,62 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+        results = {}
+        dataset_size = len(self.correct)
```
dakinggg, line 657: lets just make a dataset size attribute on the class and use that?

composer/datasets/in_context_learning_evaluation.py
dakinggg: ah once i got to the end i now understand. it is to compute multiple pass at ks without having to run this multiple times. 👍

dakinggg (commented):
Overall lgtm, the metric thing is a little gross, but the current design does not afford any other options that i see, and i think its safe, so im good with it.

composer/datasets/in_context_learning_evaluation.py
dakinggg: https://github.com/mosaicml/composer/blob/375ea0cc3f8e0f5fadf77825f06f2909f87e9fe3/composer/datasets/in_context_learning_evaluation.py#L294-L302

dakinggg: and https://github.com/mosaicml/composer/blob/375ea0cc3f8e0f5fadf77825f06f2909f87e9fe3/composer/datasets/in_context_learning_evaluation.py#L21-L23

composer/metrics/nlp.py
josejg: Making the client be dict[str,str] -> bool instead of the triple nested list version

composer/metrics/nlp.py
josejg: The previous approach assumed that all generations corresponded to the same sample, now they can be heterogeneous, so the nesting gets in the way. Namely, I could construct the triple nested list struct, but I'd have to spend lines constructing it and then unwrapping the result to update the running counts. This way I believe it's "cleaner" (minus the whole f([[[x]]])[0][0][0] )

composer/metrics/nlp.py
josejg: Done

composer/metrics/nlp.py
josejg: good point, got carried away. Addressed

composer/metrics/nlp.py
josejg: sgtm

composer/metrics/nlp.py
josejg: ok

composer/datasets/in_context_learning_evaluation.py
josejg: Fixed the import

composer/datasets/in_context_learning_evaluation.py
josejg: Sounds good, refactored with Dataset.from_generator

composer/metrics/nlp.py
dakinggg: Could you just make a jira for this instead of putting a todo in the code?

dakinggg (commented):
Could you include in the PR description evidence that before and after this PR, results don't change (assuming you carefully set the hparams to be the same). Or if that isn't possible for some reason, at least a number on a popular model that we can match to something. Otherwise, LGTM and will approve once you have that

josejg:
> Could you include in the PR description evidence that before and after this PR, results don't change (assuming you carefully set the hparams to be the same). Or if that isn't possible for some reason, at least a number on a popular model that we can match to something. Otherwise, LGTM and will approve once you have that

Added experiments on public models to the PR description

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -1288,15 +1293,27 @@ def __init__(
             'generation_kwargs': {
                 'pad_token_id': self.pad_tok_id,
                 'num_beams': 1,  # single beam
-                'num_return_sequences': generations_per_sample,
```
eitanturok, line 1291: You should update the documentation for `InContextLearningCodeEvalDataset` to remove the line about putting `num_return_sequences` in `generations_kwargs`. Specifically, you should remove line 1225 of the file:
```
- num_return_sequences: Value passed for 'generations_per_sample', how many generations per prompt
```
This will make things a lot clearer

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -1288,15 +1293,27 @@ def __init__(
...
+            'pass_at_k': list(pass_at_k),
+            'generations_per_sample': generations_per_sample,
+            'dataset_size': dataset_size,
         }
         if 'generation_kwargs' in kwargs:
             self.update_generation_kwargs(kwargs['generation_kwargs'])
 
+    def repeat_dataset(self, dataset: HFDataset, repetitions: int) -> HFDataset:
+        def repeated_dataset():
+            for i, sample in enumerate(dataset):
```
eitanturok, line 1311: This is very similar to how `bigcode-evaluation-harness` implements a repeated_dataset. Not sure if you got it from them, but their code to do this can be found [here](https://github.com/bigcode-project/bigcode-evaluation-harness/blob/00967d12093ef614de7bdad0772aed8e4118f1fd/bigcode_eval/utils.py#L123).

josejg: I figured this one independently, but thanks for the pointer!

composer/datasets/in_context_learning_evaluation.py
josejg: good catch, thanks

josejg:
I think we have a couple of tests that have the right intention but are testing the wrong thing:

- https://github.com/mosaicml/composer/blob/dev/tests/datasets/test_in_context_learning_datasets.py#L1622
- https://github.com/mosaicml/composer/blob/dev/tests/datasets/test_in_context_learning_datasets.py#L1556

The idea is to make sure that inputs aren't overly left padded. The issue is that the ICL dataloader maps using `dataset.map` and decides the left padding based on the entire dataset. Before we were not catching that because we were only looking at all the examples.

To be fair, the test is testing for the ideal behaviour, the issue is that our ICL code doesn't do that and that contribution feels out of scope for this PR.

josejg:
Reworked some of the tests. However, I'd like to know whether we can remove unnecessary left padding in a per batch basis.

dakinggg (commented):
@josejg I think it looks good to me and you are right that the tests were simply incorrect before. Are you saying you want to trim the extra left padding from each individual batch (since padding is determined based on the full dataset as opposed to each individual batch)?

composer/metrics/nlp.py
```diff
@@ -670,51 +683,54 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
...
+            if all(test_results):
+                self.correct[idx] += 1.0
 
         client.close()  # pyright: ignore [reportOptionalMemberAccess]
 
     def compute(self):
         assert isinstance(self.correct, Tensor)
         assert isinstance(self.total, Tensor)
-        return self.correct / self.total
+        if not (self.total == self.num_generations).all().item():
```
dakinggg, line 723: Request adjusting this to permit subset_num_batches as discussed offline

dakinggg (approved):
🚢

PR #2970, opened 2024-02-07
Add Torch 2.2 tests

mvpatel2000:
# What does this PR do?

Add torch 2.2 tests


j316chuck:
This will break the daily tests see: https://github.com/mosaicml/composer/actions/runs/7818637121

j316chuck (commented):
Let's fix tests before merge? wdyt.

mvpatel2000:
> Let's fix tests before merge? wdyt

Discussed offline, merging to turn on CI/CD so new PRs don't add issues

PR #2971, opened 2024-02-07
Remove bespoke codeowners

mvpatel2000:
# What does this PR do?

We still require reviews from @dakinggg and @eracah for loggers. This doesn't seem necessary and for some reason composer-team-eng approval w existing codeowners file is not enough.

Given it's not that big of a project code-wise, let's just strip bespoke codeowners.


dakinggg (approved):
I'm good with this, please wait for evan's approval as well.

eracah (approved):
LQGTM

PR #2533, opened 2023-09-12
[WIP] Error logging callback

bmosaicml:
# What does this PR do?
This PR adds a callback that logs ICL outputs during eval. It modifies the custom metrics to keep track of incorrect model outputs. Each metric is responsible for specifying the table schema for logging cached responses as well as specifying how to format the cached responses using the tokenizer.

The EvalOutputLogging callback is then responsible for logging the cached results in table format after each evaluation.

Design [doc](https://www.notion.so/Eval-output-logging-callback-cdf475b61c4a4cf6b2b65d18910380af)

# What issue(s) does this change relate to?

# Before submitting
- [x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [ ] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [x] Did you update any related docs and document your change?
- [x] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [x] Did you run the tests locally to make sure they pass?
- [x] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

PR #2977, opened 2024-02-08
Output eval logging (batch level)

maxisawesome:
# What does this PR do?
Log eval outputs after each batch using `logger.log_table`. This is an alternate design to logging at the end of eval found [here](https://github.com/mosaicml/composer/pull/2533).

REQUIRES LLM-FOUNDRY BRANCH: https://github.com/mosaicml/llm-foundry/pull/961

Eval only run: [wandb](https://wandb.ai/mosaic-ml/maxisawesome/runs/66mhk5ql)
run name: `test-batch-logging-kuIoME`


composer/metrics/nlp.py
```diff
@@ -381,14 +446,31 @@ def update(self, batch: dict, output_logits: torch.Tensor, labels: torch.Tensor)
...
-
             if idx_min == gold_idx:
                 self.correct += torch.tensor(1.0)
+                metric_result_dict['result'].append(1)
+            else:
+                metric_result_dict['result'].append(0)
+
+            question = batch['input_ids'][start][:batch['continuation_indices'][start][0]]
+
+            # TODO: Seems broken for schema tasks
```
maxisawesome, line 461: We might not be setting our continuation indices correctly for schema tasks but needs more investigation

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,77 @@
...
+        # Depad and decode input_ids
+        input_tensor = state.batch['input_ids']
+        logged_input = []
+        for input_list in input_tensor:
+            depadded_input = [tok for tok in input_list if tok != state.dataloader.dataset.pad_tok_id]
+            logged_input.append(state.dataloader.dataset.tokenizer.decode(depadded_input))
+        logging_dict['input'] = logged_input
+
+        # Log anything from the batch that's specified in the yaml
+        for key in self.batch_keys_to_log:
```
maxisawesome, line 54: We might not need this specification. Idea was that if you wanted to log something extra that exists in the batch, you can specify them in the yaml, but not really sure what that would be looking back at it.

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,77 @@
...
+        #       This might not be true if, for example, logits are logged.
+        # detokenize data in rows
+        rows = [[state.dataloader.dataset.tokenizer.decode(x) if isinstance(x, torch.Tensor) else x
+                 for x in row]
+                for row in rows]
+
+        # TODO:
+        # wandb: WARNING Step only supports monotonically increasing values, use define_metric to set a custom x axis. For details see: https://wandb.me/define-metric
+        # wandb: WARNING (User provided step: 0 is less than current step: 164. Dropping entry: {'metrics/human_eval/0-shot/InContextLearningCodeEvalAccuracy': 0.0, '_timestamp': 1707370410.1504738}).
+        # TODO: How else to chose this?
```
maxisawesome, line 73: any suggestions? :^)

composer/metrics/nlp.py
```diff
@@ -327,15 +375,31 @@ def __init__(self, dist_sync_on_step: bool = False):
         super().__init__(dist_sync_on_step=dist_sync_on_step)
         self.add_state('correct', default=torch.tensor(0.), dist_reduce_fx='sum')
         self.add_state('total', default=torch.tensor(0.), dist_reduce_fx='sum')
+        self.metric_result_dict = {
+            'context_tok': [],
```
maxisawesome, line 379: lets ensure these keys are as homogenized as possible

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # TODO:
+        # wandb: WARNING Step only supports monotonically increasing values, use define_metric to set a custom x axis. For details see: https://wandb.me/define-metric
+        # wandb: WARNING (User provided step: 0 is less than current step: 164. Dropping entry: {'metrics/human_eval/0-shot/InContextLearningCodeEvalAccuracy': 0.0, '_timestamp': 1707370410.1504738}).
+        assert state.dataloader_label is not None
+        name = state.dataloader_label
+        # TODO: How else to chose this?
+        for dest_logger in logger.destinations:
+            if dest_logger.__class__.__name__ == 'WandBLogger' or dest_logger.__class__.__name__ == 'MLFlowLogger':
+                dest_logger.log_table(columns, rows, name=name, step=0)
```
eracah, line 66: maybe have the user select in the constructor which destination to write the table to. I don't think we should hardcode this callback to wandb and mlflow. This would be the best way to do it I think if you were hardcoding mlflow and wandb
```suggestion
        for dest_logger in logger.destinations:
            if isinstance(dest_logger, WandBLogger) or isinstance(dest_logger, MLFlowLogger):
                dest_logger.log_table(columns, rows, name=name, step=0)
```

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,68 @@
...
+
+        state.metric_outputs = {}
```
eracah, line 68: don't think it's a good idea for the callback to clear the state attribute because other callbacks may want to use it

eracah (commented):
ok added some comments on the callback itself. haven't looked at the loggers

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,68 @@
...
+        assert state.metric_outputs is not None
+
+        logging_dict = deepcopy(state.metric_outputs)
+        if state.batch['mode'] == 'generate':
+            # Outputs are already detokenized
+            logging_dict['outputs'] = state.outputs
+        logging_dict['metric_name'] = [state.metric_outputs['metric_name'] for _ in range(0, len(state.outputs))]
+
+        # Depad and decode input_ids
+        input_tensor = state.batch['input_ids']
```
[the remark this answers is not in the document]
maxisawesome: no good reason, changed to input_ids

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,68 @@
...
+        logged_input = []
+        assert state.dataloader is not None
+        assert hasattr(state.dataloader, 'dataset')
+        assert hasattr(state.dataloader.dataset, 'tokenizer')
+        # assert state.dataloader.dataset is not None
+        # assert state.dataloader.dataset.tokenizer is not None
+        for input_list in input_tensor:
```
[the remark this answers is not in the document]
maxisawesome: No it's just a batch tensor of size (batch_size X gen_length). Iterating through it like this just works. I changed it to `for input_list in input_tensor.tolist()` to make it clearer?

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def eval_batch_end(self, state: State, logger: Logger) -> None:
+        assert state.outputs is not None
+        assert state.metric_outputs is not None
+
+        logging_dict = deepcopy(state.metric_outputs)
+        if state.batch['mode'] == 'generate':
+            # Outputs are already detokenized
+            logging_dict['outputs'] = state.outputs
+        logging_dict['metric_name'] = [state.metric_outputs['metric_name'] for _ in range(0, len(state.outputs))]
```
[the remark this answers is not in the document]
maxisawesome: I add the name of the metric once here: https://github.com/mosaicml/composer/pull/2977/files#diff-e61f9af34de66238d99995ebdc5d2645cea3c7ca7c303ac006239a720c5b620cR546

And this line duplicates it so that there's one entry of the metric for each datapoint in the batch (and thus each row will have a copy of it). I could move this to the update_metric function.

composer/callbacks/eval_output_logging_callback.py
maxisawesome: changed this to log to every logger except the console logger

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(self, *args, **kwargs)
+        self.warn_batch_is_not_dict = True
+
+    def eval_batch_end(self, state: State, logger: Logger) -> None:
+        if not isinstance(state.batch, Dict):
+            if self.warn_batch_is_not_dict:
+                warnings.warn(f'''EvalOutputLogging only supports batchs that are dictionary. \
```
eracah, line 34:
```suggestion
                warnings.warn(f'''EvalOutputLogging only supports batches that are dictionary. \
```

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,81 @@
...
+
+    ICL metrics are required to support caching the model's responses including information on whether model was correct.
+    Metrics are responsible for returning the results of individual datapoints in a dictionary of lists.
+    The callback will log the metric name, the depadded and detokenized input, any data stored in state.metric_outputs, and
+    any keys from the batch pased into `batch_keys_to_log`. It will do so after every eval batch.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(self, *args, **kwargs)
+        self.warn_batch_is_not_dict = True
```
eracah, line 29: hmm seems like there would be a better way to toggle this?

composer/trainer/trainer.py
```diff
@@ -3031,11 +3031,12 @@ def _eval_loop(
                                     outputs = self.state.outputs
 
                                 for metric in metrics.values():
-                                    self._original_model.update_metric(
+                                    metric_outputs = self._original_model.update_metric(
```
mvpatel2000, line 3034: Model doesn't require returning anything per type hints?

composer/metrics/nlp.py
```diff
@@ -201,6 +203,31 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.needs_batch = True
 
+    def _wrap_update(self, update: Callable) -> Callable:
```
eracah, line 206: comment here describing why you're monkeypatching torchmetrics function

mvpatel2000 (commented):
Why do you need the trainer/state changes?

sorry for slow review :(

composer/trainer/trainer.py
maxisawesome: currently its only in hf model: https://github.com/mosaicml/composer/pull/2977/files/25f1872fc109d00138b8a46dfed2c06aa89160a9#diff-e61f9af34de66238d99995ebdc5d2645cea3c7ca7c303ac006239a720c5b620cR533

afaik we only do evals w/ huggingface models? Given the implementations here: https://github.com/mosaicml/composer/blob/25f1872fc109d00138b8a46dfed2c06aa89160a9/composer/models/huggingface.py#L488

maxisawesome:
> Why do you need the trainer/state changes?
> 
> sorry for slow review :(

because we need to save per-batch eval results somewhere for the callback to access them

composer/metrics/nlp.py
maxisawesome: Not broken - schema tasks' continuation_ids are actually constant. The 'context' changes and the 'continuation' stays the same, and we take the item with a lower perplexity and the model's 'choice'. Thus, continuation was always the same for these tasks and I was getting spooked. I added more logged content (the entirety of MC options, ie the context + answer combined for all possible answers) to make this clear.

composer/loggers/in_memory_logger.py
```diff
@@ -83,8 +83,10 @@ def log_table(self,
             raise MissingConditionalImportError(extra_deps_group='pandas',
                                                 conda_package='pandas',
                                                 conda_channel='conda-forge') from e
-        table = pd.DataFrame.from_records(data=rows, columns=columns).to_json(orient='split', index=False)
-        assert isinstance(table, str)
+        table = pd.DataFrame.from_records(data=rows, columns=columns).to_json(orient='split',
+                                                                              index=False,
+                                                                              force_ascii=False)
```
mvpatel2000, line 88: why is force ascii false now needed?

composer/metrics/nlp.py
```diff
@@ -303,8 +342,10 @@ def replace_underscore(text: str) -> str:
 
     def update(self, outputs: List[str], labels: List[List[str]], batch: Dict[str, Any]):
         cot_delimiter = batch.get('cot_delimiter', '')
+
```
mvpatel2000, line 345:
```suggestion
```
accident? or intentional

composer/metrics/nlp.py
```diff
@@ -372,13 +424,23 @@ def update(self,
                                                                      labels=labels,
                                                                      outputs=outputs)
 
+        metric_result_dict = copy.deepcopy(self.metric_result_dict)
         for batch_idx, cont_idx in enumerate(batch['continuation_indices']):
             cont_tok_pred = outputs[batch_idx].index_select(dim=0, index=cont_idx - 1).argmax(dim=-1)
             cont_tok_targ = labels[batch_idx].index_select(dim=0, index=cont_idx - 1)
+            # TODO: okay to do context_tok here? or do we wanna do that in the logger?
```
mvpatel2000, line 431: what does this mean?

composer/metrics/nlp.py
```diff
@@ -650,7 +750,7 @@ def estimator(self, n: int, c: int, k: int) -> float:
             return 1.0
         return 1.0 - float(np.prod(1.0 - k / np.arange(n - c + 1, n + 1)))
 
-    def _initialize_state(self, batch: dict[str, Any]):
+    def _initialize_state(self, batch: Dict[str, Any]):
```
mvpatel2000, line 753:
```suggestion
    def _initialize_state(self, batch: dict[str, Any]):
```
It should be dict

composer/metrics/nlp.py
```diff
@@ -692,15 +792,19 @@ def update(self, batch: Dict[str, Any], outputs: List[str], labels: List[str]):
         del labels  # never used
         client = self.get_client()
 
+        metric_result_dict = copy.deepcopy(self.metric_result_dict)
```
mvpatel2000, line 795: why deepcopy now? same as above

composer/trainer/trainer.py
mvpatel2000: Do we need to return this? Can we just pull the metric from state?

composer/loggers/in_memory_logger.py
maxisawesome: @bmosaicml added this in his original PR.

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -0,0 +1,120 @@
...
+        columns = list(logging_dict.keys())
+        rows = [list(item) for item in zip(*logging_dict.values())]
+
+        assert state.dataloader_label is not None
+        if not self.name:
+            # If only running eval, step will be 0
+            # If running training, step will be current training step
+            step = state.timestamp.batch.value
+            self.name = f'{state.dataloader_label}_step_{step}'
+            self.columns = columns
```
eracah, line 108: why do we only set columns `if not self.name`?

composer/trainer/trainer.py
maxisawesome: We are returning this in order to set the value in state.

eracah (approved):
LGTM. I mostly looked closely at the callbacks and loggers part

eracah:
Also I just want to say great work! This is a herculean PR requiring deep, bespoke knowledge while juggling several different parts of the composer codebase. Not an easy one to wrangle and seems like you managed to make it work!

mvpatel2000 (changes requested):
im holding until a few things are resolved. But this is my top priority to help u land :)

composer/metrics/nlp.py
maxisawesome: the class variable is a template for each new batch to construct what should be logged. It's the same thing we do here: https://github.com/mosaicml/composer/blob/dev/composer/datasets/in_context_learning_evaluation.py#L618

composer/metrics/nlp.py
maxisawesome: outdated comment

composer/callbacks/eval_output_logging_callback.py
maxisawesome: We only want to set these on the first batch. Steps increment in different ways during training vs during eval, but we only want to record the name/step at the first step to keep them consistent. During training, the eval step is static throughout a single set of evals (ie at training step 1000, eval step will always be the same). During Eval, the steps incremented with each batch. In both cases, we want to log to a table using the name/step at the first batch received.

mvpatel2000 (approved):

PR #2978, opened 2024-02-08
Format load_path with name

mvpatel2000:
# What does this PR do?

Format load_path with name. This way if I specify run_name as a format var in a load_path, like we frequently do with save_folder, it will automatically format it. Eg `oci://ckpts/mihir/{run_name}/ep0-ba50` will be formatted correctly.

Also adds unit test.

I like to test resumption by keeping a yaml that explicitly formats name into save/load path, but that doesn't work currently.

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -28,7 +28,7 @@ from torch.distributed.checkpoint.planner import LoadPlan, LoadPlanner
 from composer.utils import dist, reproducibility
 from composer.utils.file_helpers import (FORMAT_NAME_WITH_DIST_AND_TIME_TABLE, format_name_with_dist,
                                          format_name_with_dist_and_time, get_file, is_tar)
-from composer.utils.misc import is_model_deepspeed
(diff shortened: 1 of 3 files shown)


composer/utils/checkpoint.py
```diff
@@ -625,18 +626,18 @@ def download_checkpoint(path: str,
                     # or could not be downloaded
                     raise RuntimeError(f'Checkpoint {path} does not exist')
         elif checkpoint_is_sharded:
-            # if the checkpoint is sharded, then every rank needs to download its own checkpoint
+            # If the checkpoint is sharded, then every rank needs to download its own checkpoint
+            path = _format_path_with_current_rank(path)
```
dakinggg, line 630: just checking that I'm not misreading, this section is just a refactor with no logical changes, yeah?

dakinggg (approved):

composer/utils/checkpoint.py
mvpatel2000: Yep

PR #2985, opened 2024-02-09
Remove planner validation

mvpatel2000:
# What does this PR do?

Removes planner validation. This should be done by Torch.

The planners let an end-user arbitrarily modify state dict before loading/saving. It's an optional argument to customize things further.

@b-chu I didn't get where type hints are supposed to validate -- can you clarify the comment?

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -456,8 +456,6 @@ def load_sharded_checkpoint(
 
     if state.fsdp_config is None:
         raise ValueError('Loading a sharded checkpoint requires passing an FSDP config to Trainer.')
-    load_planner = state.fsdp_config['load_planner']
-    _validate_load_planner(load_planner)
 
     def _get_num_ranks_that_saved_rng(metadata: Metadata):
         rng_inds = []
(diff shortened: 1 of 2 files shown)


composer/utils/checkpoint.py
```diff
@@ -752,40 +750,6 @@ def filter_func(state_dict: dict) -> None:
     return filter_func
 
 
-def _validate_save_planner(save_planner: Optional[Any]) -> None:
```
snarayan21, line 755: mind adding a bit to the PR description about what save planner and load planner do / why torch needs them when loading sharded checkpoints? That would help the rest of the team get more familiar with checkpointing code

tests/utils/eval_client/test_local_eval_client.py
```diff
@@ -29,10 +29,11 @@
...
+    """Test invocation function for LocalEvalClient.
+
+    Code can succeed, fail compilation, time out, or be incorrect in C, C++, Python, JS.
     """
-    import os
-    os.makedirs(os.path.dirname(tmp_path), exist_ok=True)
+    dist.barrier()  # Ensure all processes are ready to run the test as invoke doesn't use dist
     eval_client = LocalEvalClient()
     input = '(1,)' if language == 'python' else '1'
     assert eval_client.invoke([[[{
```
snarayan21, line 39: doesnt `invoke` use multiprocessing through the `invoke_helper` and should wait on all processes to complete using `p.join(TIMEOUT)`? Instead of using `dist.barrier()`, why not increase `TIMEOUT`?

snarayan21 (commented):
just a couple comments

tests/utils/eval_client/test_local_eval_client.py
mvpatel2000: I want it to sync per test -- are you saying it already does this?

b-chu (approved):

PR #2986, opened 2024-02-09
fix skip_first for resumption

bigning:
# What does this PR do?

currently the profiler’s skip_first argument counts batches from the beginning of training and not from the start of the current resumption.  This PR make the skip_first count from the start of the last resumption. 

# test
pytest tests/profiler/test_profiler.py -k test_skip_first_after_resumption

image

### DIFF
--- a/composer/profiler/torch_profiler.py
+++ b/composer/profiler/torch_profiler.py
@@ -225,6 +225,9 @@ class TorchProfiler(Callback):  # noqa: D101
         self.num_traces_to_keep = num_traces_to_keep
         self.saved_traces = OrderedDict()
         self.profiler: Optional[torch.profiler.profile] = None
+        # This is used in schedule function to update the skip_first, so
+        # that we count the skip_first starting from the 1st batch after resumption
+        self.resumption_batch_idx: int = 0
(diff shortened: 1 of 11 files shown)


mvpatel2000 (approved):
LGTM! Nice test.

As a side note, the profiler is one of the earliest things we designed and is frankly poorly done. We really should have made it a callback directly (eg CheckpointSaver) and I don't really like the orchestrator profiler class...

mvpatel2000:
Btw, you can run lint with `pre-commit --all-files` locally

composer/profiler/torch_profiler.py
```diff
@@ -328,6 +328,11 @@ def handler_fn(prof: torch.profiler.profiler.profile):
         )
         self.profiler.__enter__()
 
+    def after_load(self, state: State, logger: Logger) -> None:
+        del logger
+        assert state.profiler is not None
+        state.profiler.resumption_batch_idx = int(state.timestamp.batch_in_epoch)
```
mvpatel2000, line 334: What if we make `profiler.py` a callback instead?

PR #2987, opened 2024-02-09
use flash attn fuse cross entropy loss to reduce metric memory usage

cli99:
This PR uses fused cross entropy loss from flash attention in the metric LanguageCrossEntropy (also LanguagePerplexity). 
The current torch.nn.CrossEntropyLoss call needs `6 * seq_len * vocab_size` GPU memory, and can be the bottleneck memory usage when sequence length is long (where act ckpt is probably used). Using  cross entropy loss from flash attn resolves this problem.

Example test model with long sequence and full act ckpt:
with torch loss fn:
image

with flash_attn loss fn
image

### DIFF
--- a/composer/metrics/nlp.py
+++ b/composer/metrics/nlp.py
@@ -97,7 +97,11 @@ class LanguageCrossEntropy(Metric):
         super().__init__(dist_sync_on_step=dist_sync_on_step)
 
         self.ignore_index = ignore_index
-        self.loss_fn = torch.nn.CrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
+        try:
+            from flash_attn.losses.cross_entropy import CrossEntropyLoss as FusedCrossEntropyLoss
+            self.loss_fn = FusedCrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
+        except ImportError:
+            self.loss_fn = torch.nn.CrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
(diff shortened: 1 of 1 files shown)


dakinggg:
@cli99 consider doing it as in https://github.com/mosaicml/llm-foundry/pull/575/files, to avoid introducing flash attn as a composer dependency.

composer/metrics/nlp.py
```diff
@@ -97,7 +97,11 @@ def __init__(self, dist_sync_on_step: bool = False, ignore_index: int = -100):
         super().__init__(dist_sync_on_step=dist_sync_on_step)
 
         self.ignore_index = ignore_index
-        self.loss_fn = torch.nn.CrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
+        try:
+            from flash_attn.losses.cross_entropy import CrossEntropyLoss as FusedCrossEntropyLoss
+            self.loss_fn = FusedCrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
+        except ImportError:
+            self.loss_fn = torch.nn.CrossEntropyLoss(ignore_index=ignore_index, reduction='sum')
```
j316chuck, line 104: Should this live in llm-foundry? CC: @dakinggg

dakinggg: see above

dakinggg: i have a very old pr that i never merged

mvpatel2000:
@cli99 should we close or is this still WIP

PR #2982, opened 2024-02-08
Fix UC object store bugfix

nancyhung:
# What does this PR do?

To unblock Databricks Finetuning PrPr launch, we need to modify the `get_object_size` function, which is currently used to check object existence. The UC team is working on a long-term fix to support `files.get_status` endpoint, but it is not available yet. The workaround we were recommended to use is to call HEAD on the files api endpoint to fetch metadata on the UC resource.

# Testing
Manually tested in r1z2 by pulling my branch and verifying that this works with my Databricks staging UC object.
```
>>> from composer.utils.object_store.uc_object_store import UCObjectStore as US
>>> o = US('Volumes/main/nancyhung/ift')
>>> o.get_object_size('train.jsonl')
1000000
```
-------
Testing by Jerry with a dogfood volume (but functionally should be same):
```
>>> from composer.utils.object_store import UCObjectStore
>>> o = UCObjectStore('Volumes/main/nancyhung/ift')
>>> o.get_object_size('bruh.jsonl')
Traceback (most recent call last):
  File "/composer/composer/utils/object_store/uc_object_store.py", line 214, in get_object_size
    self.client.api_client.do(method='HEAD',
  File "/usr/lib/python3/dist-packages/databricks/sdk/core.py", line 127, in do
    return retryable(self._perform)(method,
  File "/usr/lib/python3/dist-packages/databricks/sdk/retries.py", line 50, in wrapper
    raise err
  File "/usr/lib/python3/dist-packages/databricks/sdk/retries.py", line 29, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/databricks/sdk/core.py", line 230, in _perform
    raise self._make_nicer_error(response=response, message=message) from None
databricks.sdk.errors.mapping.NotFound: Not Found

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/composer/composer/utils/object_store/uc_object_store.py", line 220, in get_object_size
    _wrap_errors(self.get_uri(object_name), e)
  File "/composer/composer/utils/object_store/uc_object_store.py", line 30, in _wrap_errors
    raise FileNotFoundError(f'Object {uri} not found') from e
FileNotFoundError: Object dbfs:/Volumes/main/nancyhung/ift/bruh.jsonl not found
>>>
>>> o.get_object_size('train.jsonl')
1000000
```

# Before submitting
[...]

PR #2988, opened 2024-02-09
[Bugfix][UC] Add back the full object path

nancyhung:
# What does this PR do?

In previous change, https://github.com/mosaicml/composer/pull/2982 I dropped using the full object path `self._get_object_path(object_name)`. This PR adds it back.

# Testing
Works when modified the code in long-living pod in prod. 
```
>>> llmfoundry.data.finetuning.dataloader._download_remote_hf_dataset('dbfs:/Volumes/finetuning/ift/train_vol', 'train')
'/usr/lib/python3/dist-packages/llmfoundry/.downloaded_finetuning/train'
```
confirmed file is downloaded in the directory
```
# head data/train-00000-of-00001.jsonl
{"prompt": "<|im_start|>system\nA conversation between a user and an LLM-based AI assistant.
...
```

# Before submitting

### DIFF
--- a/composer/utils/object_store/uc_object_store.py
+++ b/composer/utils/object_store/uc_object_store.py
@@ -212,7 +212,7 @@ class UCObjectStore(ObjectStore):
             # does not work. Once fixed, we will call the files API endpoint. We currently only use this
             # function in Composer and LLM-foundry to check the UC object's existence.
             self.client.api_client.do(method='HEAD',
-                                      path=f'{self._UC_VOLUME_FILES_API_ENDPOINT}/{self.prefix}/{object_name}',
+                                      path=f'{self._UC_VOLUME_FILES_API_ENDPOINT}/{self._get_object_path(object_name)}',
                                       headers={'Source': 'mosaicml/composer'})
             return 1000000  # Dummy value, as we don't have a way to get the size of the file
         except DatabricksError as e:


jbloxham (approved):

composer/utils/object_store/uc_object_store.py
```diff
@@ -212,7 +212,7 @@ def get_object_size(self, object_name: str) -> int:
             # does not work. Once fixed, we will call the files API endpoint. We currently only use this
             # function in Composer and LLM-foundry to check the UC object's existence.
             self.client.api_client.do(method='HEAD',
-                                      path=f'{self._UC_VOLUME_FILES_API_ENDPOINT}/{self.prefix}/{object_name}',
+                                      path=f'{self._UC_VOLUME_FILES_API_ENDPOINT}/{self._get_object_path(object_name)}',
```
panchalhp-db, line 215: Can we use os.path.join(..) to avoid bugs with leading and lagging slashes?

jerrychen109 (approved):

composer/utils/object_store/uc_object_store.py
nancyhung: we can't because `os.path.join` because it will ignore previous parts of the input if one of them is an absolute path. `self._get_object_path` returns an absolute path

Spec of `os.path.join`
```
Join two or more pathname components, inserting '/' as needed. If any component is an absolute path, all previous path components will be discarded. An empty last part will result in a path that ends with a separator.
```

Fails in the run if I try `os.path.join`
```
llmfoundry.data.finetuning.dataloader._download_remote_hf_dataset('dbfs:/Volumes/finetuning/ift/train_vol', 'train')
  File "/composer/composer/utils/object_store/uc_object_store.py", line 214, in get_object_size
    self.client.api_client.do(method='HEAD',
  File "/usr/lib/python3/dist-packages/databricks/sdk/core.py", line 127, in do
    return retryable(self._perform)(method,
  File "/usr/lib/python3/dist-packages/databricks/sdk/retries.py", line 50, in wrapper
    raise err
  File "/usr/lib/python3/dist-packages/databricks/sdk/retries.py", line 29, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/databricks/sdk/core.py", line 230, in _perform
    raise self._make_nicer_error(response=response, message=message) from None
databricks.sdk.errors.mapping.NotFound: Not Found
```

We could use pathlib and deal with Path objects, but going to push the fstring fix for now

panchalhp-db: Ah, that's a pain. Thanks for checking! The current code works in that case.

PR #2992, opened 2024-02-11
Fix Profiler schedule skip_first

bigning:
# What does this PR do?

almost same as https://github.com/mosaicml/composer/pull/2986, which was reverted. Copied from that PR:

currently the profiler’s skip_first argument counts batches from the beginning of training and not from the start of the current resumption. This PR make the skip_first count from the start of the last resumption.

This new PR addressed feedback by @mvpatel2000. It moved the `after_load` callback to `Profiler`, instead of `TorchProfiler` by making `Profiler` inheriting `Callback`

# test
pytest tests/profiler/test_profiler.py -k test_skip_first_after_resumption
image

### DIFF
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.19.0'
+__version__ = '0.19.1'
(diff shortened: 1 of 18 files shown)


mvpatel2000 (approved):
LGTM! 

Turning it into a callback triggers a host of tests that run for all callbacks. You'll need to mock inputs for it for these tests -- the infra is in place, you just need to register it. Feel free to ping me if it's not clear how to do that

PR #2996, opened 2024-02-12
Add azure integration test

mvpatel2000:
# What does this PR do?

Add azure integration test. 

Note: We only add one test since under the hood we use libcloud which is tested directly elsewhere

Along for ride, removes a test file which was fully skipped. We added a different file for remote testing with GCS, so this is no longer needed.

Test passes here: https://github.com/mosaicml/composer/actions/runs/7876955247


b-chu (commented):
Why remove GCS test?

PR #2997, opened 2024-02-12
Fix load for non-HSDP device mesh

mvpatel2000:
# What does this PR do?

Currently, we only download on rank 0 if using FSDP + device mesh

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -214,8 +214,8 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
 
     def read_data(self, plan: LoadPlan, planner: LoadPlanner):
         # Download files if not using HSDP or if on first replica with HSDP enabled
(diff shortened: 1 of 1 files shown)


composer/utils/checkpoint.py
```diff
@@ -213,7 +213,9 @@ def __init__(self, source_path: str, destination_path: str, object_store: Union[
         super().__init__(destination_path)
 
     def read_data(self, plan: LoadPlan, planner: LoadPlanner):
-        first_replica = self.device_mesh is None or self.device_mesh.get_local_rank(mesh_dim=0) == 0
+        # Download files if not using HSDP or if on first replica with HSDP enabled
+        first_replica = self.device_mesh is None or self.device_mesh.ndim >= 2 and self.device_mesh.get_local_rank(
```
dakinggg, line 217: can you add a set of parens to make the boolean ordering more clear?

dakinggg (approved):

PR #2998, opened 2024-02-13
Only save RNG on rank 0

mvpatel2000:
# What does this PR do?

Only save RNG on rank 0. This is to avoid dedup warnings we get otherwise.

Manually tested on a 30b save/load and no more logs.

Along for the ride, it seems we have two state variables that have the same value. I've removed one of the duplicate names.


mvpatel2000:
> Is `state.fsdp_elastic_sharded_enabled` getting deprecated? Won't this break for external users?

Restored with deprecation

composer/core/state.py
```diff
@@ -731,6 +731,12 @@ def fsdp_state_dict_type(self):
     def fsdp_sharded_state_dict_enabled(self):
         return self.fsdp_config is not None and self.fsdp_enabled and self.fsdp_state_dict_type == 'sharded'
 
+    @property
+    def fsdp_elastic_sharded_enabled(self):
+        # Deprecation warnign
```
b-chu, line 736:
```suggestion
        # Deprecation warning
```

b-chu (approved):
Thanks for the PR! Just one typo

PR #2999, opened 2024-02-13
Add bias argument to LPLN

mvpatel2000:
# What does this PR do?

Add bias argument to LPLN and forward to LayerNorm parent class

### DIFF
--- a/composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
+++ b/composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
@@ -75,7 +75,15 @@ class LowPrecisionLayerNorm(Algorithm):
 
 class LPLayerNorm(torch.nn.LayerNorm):
 
-    def __init__(self, normalized_shape, eps=1e-05, elementwise_affine=True, bias=True, device=None, dtype=None):
+    def __init__(
+        self,
+        normalized_shape,
+        eps: float = 1e-05,
+        elementwise_affine: bool = True,
+        bias: bool = True,
+        device=None,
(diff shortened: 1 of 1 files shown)


composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
```diff
@@ -75,11 +75,12 @@ def apply(self, event: Event, state: State, logger: Logger) -> Optional[int]:
 
 class LPLayerNorm(torch.nn.LayerNorm):
 
-    def __init__(self, normalized_shape, eps=1e-05, elementwise_affine=True, device=None, dtype=None):
+    def __init__(self, normalized_shape, eps=1e-05, elementwise_affine=True, bias=True, device=None, dtype=None):
```
dakinggg, line 78: Want to add types while you're here?

dakinggg (approved):

composer/algorithms/low_precision_layernorm/low_precision_layernorm.py
dakinggg, line 78: Also probably need to put bias at the end to keep arg order?

mvpatel2000: I'm following torch order

mvpatel2000: I think better to be consistent

vchiley:
only relevant in `torch.__version__ >= 2.1`; should we check that?

PR #3001, opened 2024-02-13
Race condition fix in checkpoint loading util

jessechancy:
# What does this PR do?

With HSDP, there are multiple replicas of a model, and multiple shards per replica. There is a race condition when multiple shards try to download the same file, which causes higher network latency, resulting in a longer time needed for checkpoint resumption.

We fix the issue by deduplicating across the shards in the first replica (only the first replica downloads the files) by first gathering all the required files, and then checking if a lesser ranked shard is already downloading the file.


mvpatel2000 (dismissed):
Nice!! This is super clean :D

composer/utils/checkpoint.py
```diff
@@ -217,17 +217,30 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
         first_replica = self.device_mesh is None or self.device_mesh.ndim == 1 or (
             self.device_mesh.ndim >= 2 and self.device_mesh.get_local_rank(mesh_dim=0) == 0)
 
-        # 1. Download to the destination all files this rank needs if on first replica
+        # 1. Collect the relative paths to download for all ranks for deduplication
+        relative_file_paths = set()
+        for plan_item in plan.items:
+            relative_file_paths.add(self.storage_data[plan_item.storage_index].relative_path)
+        all_file_paths = dist.all_gather_object(relative_file_paths)
+        log.debug(f"Files required for all ranks {all_file_paths}")
```
mvpatel2000, line 225: Note: this does an all-gather across all ranks, but you probably only need what's in the first replica.

However, pulling out the PG conditionally is probably not worth effort given complicated scenarios

composer/utils/checkpoint.py
```diff
@@ -217,17 +217,30 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
         if first_replica:
             log.debug(f'Rank {dist.get_global_rank()} starting to download files.')
+
+            # Download the files for this rank if they haven't been downloaded before
+            rank = dist.get_local_rank()
             for plan_item in plan.items:
-                # Each plan item has a storage index which points to the relative path of the shard file at save time.
                 relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                # Check if the file has already been downloaded by another rank
+                is_downloaded = any(relative_file_path in all_file_paths[i] for i in range(rank))
```
eracah, line 236: Not sure I understand how this works. `all_file_paths[i]` is a list of all the files a rank needs not a list of files the rank already downloaded?

mvpatel2000, line 236: SInce only each node shares a file system, you need to check only if something in the same node is downloading this.

Additionally, this should be using global rank not local rank
```suggestion
                local_rank_0 = dist.get_global_rank() // dist.get_local_world_size() * dist.get_local_world_size() # TODO: cache at top of fn
                is_downloaded = any(relative_file_path in all_file_paths[i] for i in range(dist.get_global_rank()))
```

mvpatel2000: comment/variable name can be more clear. 

Idea is another rank is responsible for downloading it. Eg if rank 9 and 10 both need shard 0, only the lower numbered rank (9) downloads it instead of both

eracah: oh I see. it only loops up to the ranks before the current rank. that's almost too slick

jessechancy: Oh yep I'll update the comment to clarify

composer/utils/checkpoint.py
```diff
@@ -217,17 +217,33 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
         first_replica = self.device_mesh is None or self.device_mesh.ndim == 1 or (
             self.device_mesh.ndim >= 2 and self.device_mesh.get_local_rank(mesh_dim=0) == 0)
 
-        # 1. Download to the destination all files this rank needs if on first replica
+        # 1. Collect the relative paths to download for all ranks for deduplication
+        relative_file_paths = set()
+        for plan_item in plan.items:
+            relative_file_paths.add(self.storage_data[plan_item.storage_index].relative_path)
+        all_file_paths = dist.all_gather_object(relative_file_paths)
+        log.debug(f'Files required for all ranks: {all_file_paths}.')
```
mvpatel2000, line 225:
```suggestion
```
Let's remove this -- for large runs this will become a really really big log

composer/utils/checkpoint.py
```diff
@@ -217,17 +217,33 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+
+        # 2. Download to the destination all files this rank needs if on first replica
         if first_replica:
             log.debug(f'Rank {dist.get_global_rank()} starting to download files.')
+
+            # Get the lowest rank in the current node
+            local_rank_0 = dist.get_global_rank() // dist.get_local_world_size() * dist.get_local_world_size()
```
mvpatel2000, line 232:
```suggestion
            local_rank_0 = dist.get_global_rank() - dist.get_local_rank()
```
mildly cleaner -- my original suggestion was worse. This is useful for variable sized local worlds (not sure why you'd want to do this)

composer/utils/checkpoint.py
```diff
@@ -217,17 +217,33 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+
             for plan_item in plan.items:
-                # Each plan item has a storage index which points to the relative path of the shard file at save time.
                 relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                # Check if the file is scheduled to be downloaded by a lower rank on the same node
+                # i.e. if rank 9 and rank 10 on the same node have the same the same required file, rank 10 should not download it.
```
mvpatel2000, line 237:
```suggestion
                # i.e. if rank 0 and rank 1 on the same node have the same the same required file,
                # only rank 0 should download it and not rank 1.
```
Let's keep example to node 0 for code since it's easier to understand. I gave a rank 1 example for Evan to imply multinode, but for comments likely more confusing

mvpatel2000 (commented):
A few minor nits

mvpatel2000 (approved):

PR #3000, opened 2024-02-13
Update NCCL arg since torch deprecated old one

mvpatel2000:
# What does this PR do?

Update NCCL arg since torch deprecated old one

PR #3004, opened 2024-02-13
Revert "Update NCCL arg since torch deprecated old one"

mvpatel2000:
Reverts mosaicml/composer#3000


bigning:
qq, why this is reverted ? Asking because it's introduced again here https://github.com/mosaicml/composer/pull/3267. 

I suspect PR3267 might cause this issue: https://databricks.slack.com/archives/C02EB3VCYRK/p1716568599038629 cc @j316chuck

PR #3005, opened 2024-02-13
Add debug log before training loop starts

mvpatel2000:
# What does this PR do?

Streaming often takes a while to start first batch. This changes last seen debug message from user to starting training loop. This avoids confusion with users thinking spinning dataloaders is taking a really long time.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -2086,6 +2086,7 @@ class Trainer:
             # asserted to be not None when Trainer.fit() is called
             raise RuntimeError('max_duration must be specified when initializing the Trainer')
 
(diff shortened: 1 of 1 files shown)


bigning (approved):
simple change, lgtm.

mvpatel2000:
> Bonus points: can we make batch 0 print for any `console_log_interval`? So that with an interval of 10 we print after batches {0, 10, 20}, etc instead of {9, 19, 29}, etc? I think this would address 80% of the motivation for the more granular logging I mentioned in the JIRA issue. (May or may not make sense to do in this PR.)

I'll follow up offline. We already do this, maybe you were testing a resumption?

PR #3009, opened 2024-02-13
[EASY] Always log 1st batch when resuming training 

bigning:
# What does this PR do?

composer prints the console logging for 1st batch and every `self.log_interval` batches. but when resuming training, we probably want to print the 1st batch no matter if `cur_batch % self.log_interval == 0`. This PR just forces to print console logging for 1st batch after resumption. 

# test
1. firstly save checkpoint after training 3 batches.
2. load checkpoint and set `console_log_interval=7ba` `composer train/train.py train/yamls/pretrain/mpt-125m.yaml train_loader.dataset.split=train_small  max_duration=10ba eval_interval=0 console_log_interval=7ba load_path=./125m/checkpoints/ep0-ba3-rank0.pt`

## before the change:
first logged batch is 7 
image
## after the change
first logged batch is 4, then 7
image

### DIFF
--- a/composer/loggers/console_logger.py
+++ b/composer/loggers/console_logger.py
@@ -109,7 +109,7 @@ class ConsoleLogger(LoggerDestination):
         cur_epoch = int(state.timestamp.epoch)  # epoch gets incremented right before EPOCH_END
         unit = self.log_interval.unit
 
-        if unit == TimeUnit.EPOCH and (cur_epoch % int(self.log_interval) == 0 or cur_epoch == 1):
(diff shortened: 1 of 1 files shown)


mvpatel2000 (commented):
Should we remove `cur_batch==1` in this case?

Can we please add the same for `def epoch_end`?

bigning:
> Should we remove `cur_batch==1` in this case?
> 
> Can we please add the same for `def epoch_end`?

good catch! updated.

mvpatel2000 (approved):
LGTM! Thanks for the fix.

PR #3014, opened 2024-02-15
Remove "generation_length" in favor of "generation_kwargs"

maxisawesome:
# What does this PR do?

KILLS :bufo-knife: `generation_length` and sets it directly to `max_new_tokens` in `generation_kwargs`


composer/models/huggingface.py
```diff
@@ -469,7 +469,6 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
             self.labels = batch.pop('labels')
             generation = self.generate(batch['input_ids'],
                                        attention_mask=batch['attention_mask'],
```
dakinggg, line 471: this should be done in a backards compatible way. so, get generation length from the batch, if it exists, raise deprecation warning saying to use max_new_tokens instead, and `max_new_tokens = batch.get('generation_kwargs', {}).get('max_new_tokens', batch.get('generation_length', None))` or something like that

maxisawesome: This was never set explicitly in any of our icl datasets. This is for anyone who's written custom icl_datasets, correct?

dakinggg: ya

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -689,14 +689,7 @@ def __init__(
...
         self.max_answer_length = 0
-        static_keys = [
-            'mode',
-            'cot_delimiter',
-            'generation_length',
-            'generation_kwargs',
-            'do_normalization',
-            'stopping_criteria',
-        ]
+        static_keys = ['mode', 'cot_delimiter', 'generation_kwargs', 'do_normalization', 'stopping_criteria']
```
maxisawesome, line 692: if user specifies still need to duplicate generation_length

composer/models/huggingface.py
```diff
@@ -466,13 +466,31 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
...
+            if 'generation_length' in batch:
+                warnings.warn(
+                    ('`generation_length` has been deprecated in favor of passing `max_new_tokens` directly into `generation_kwargs`.'
+                     'It will be removed in v0.21'),
+                    DeprecationWarning,
+                )
+            max_new_tokens = None
+            if 'generation_kwargs' in batch:
+                if 'max_new_tokens' in batch['generation_kwargs']:
+                    max_new_tokens = batch['generation_kwargs']['max_new_tokens']
```
maxisawesome, line 478: set here and remove weird setting behavior

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -690,12 +690,7 @@ def __init__(
         self.has_cot = False
         self.max_answer_length = 0
         static_keys = [
-            'mode',
-            'cot_delimiter',
-            'generation_length',
-            'generation_kwargs',
-            'do_normalization',
-            'stopping_criteria',
+            'mode', 'cot_delimiter', 'generation_length', 'generation_kwargs', 'do_normalization', 'stopping_criteria'
```
dakinggg, line 693: trailing comma back

tests/models/test_hf_model.py
```diff
@@ -1208,6 +1209,67 @@ def test_eval_forward_generate(device, world_size, hf_model, hf_tokenizer, use_f
...
+    if isinstance(hf_tokenizer, transformers.models.gpt2.tokenization_gpt2_fast.GPT2TokenizerFast):
+        hf_tokenizer.padding_side = 'left'
+    input_dict = hf_tokenizer(['hello', 'goodbyes'], return_tensors='pt', padding=True)
+    for k, v in input_dict.items():
+        input_dict[k] = device.tensor_to_device(v)
+    input_dict['mode'] = 'generate'
+    input_dict['generation_kwargs'] = {}
+    input_dict['generation_length'] = 5
+    input_dict['labels'] = [['answer1'], ['answer2']]
+    with pytest.raises(DeprecationWarning):
```
dakinggg, line 1261: this isn't a raise, you want pytest.warns I think. Also I don't think this test needs all the parametrization and FSDP stuff. can simplify it a bunch

dakinggg (approved):

PR #3018, opened 2024-02-15
Use Mosaic constant for GPU file prefix

jjanezhang:
# Use Mosaic constant for GPU file prefix
Swapping out manual coded prefix file for a constant on the Mosaic side. 

Confirmed that this works with logging

image


composer/cli/launcher.py
```diff
@@ -485,7 +486,7 @@ def main():
             MOSAICML_PLATFORM_ENV_VAR,
             'false').lower() == 'true' and str(os.environ.get(MOSAICML_LOG_DIR_ENV_VAR, 'false')).lower() != 'false':
         log.info('Logging all GPU ranks to Mosaic Platform.')
-        log_file_format = f'{os.environ.get(MOSAICML_LOG_DIR_ENV_VAR)}/gpu_{{rank}}.txt'
+        log_file_format = f'{os.environ.get(MOSAICML_LOG_DIR_ENV_VAR)}/{os.environ.get(MOSAICML_GPU_LOG_FILE_PREFIX_ENV_VAR)}{{local_rank}}.txt'
```
j316chuck, line 489: Curious why this is `local_rank` instead of `rank`?

j316chuck (commented):
left a comment. overall lgtm

composer/cli/launcher.py
jjanezhang: The file directory format we use for logging separates node and gpu rank instead of using a combined global rank so this just decreases the amount of parsing we have to do on the mcloud side!

j316chuck (approved):

composer/cli/launcher.py
dakinggg, line 489: @jjanezhang this will error if the env var doesn't exist, did you want that? or should it be more permissive?

PR #3020, opened 2024-02-16
Fall back to normal logging when gpu prefix is not present

jjanezhang:
# Fall back to normal logging when gpu prefix is not present
Re conversation at the end of here: https://github.com/mosaicml/composer/pull/3018

Checking for presence of the gpu prefix env var so we can default to normal logging behavior if we fail to provide a value.

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -23,7 +23,7 @@ import torch
 
 import composer
 from composer.loggers.mosaicml_logger import (MOSAICML_GPU_LOG_FILE_PREFIX_ENV_VAR, MOSAICML_LOG_DIR_ENV_VAR,
-                                              MOSAICML_PLATFORM_ENV_VAR,)
+                                              MOSAICML_PLATFORM_ENV_VAR)
 from composer.utils import get_free_tcp_port
 
 CLEANUP_TIMEOUT = datetime.timedelta(seconds=30)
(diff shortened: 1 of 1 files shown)


composer/cli/launcher.py
```diff
@@ -482,9 +482,9 @@ def main():
         args.stderr = f'{log_tmpdir.name}/rank{{rank}}.stderr.txt'
 
     # If running on the Mosaic platform, log all gpu ranks' stderr and stdout to Mosaic platform
-    if os.environ.get(
-            MOSAICML_PLATFORM_ENV_VAR,
-            'false').lower() == 'true' and str(os.environ.get(MOSAICML_LOG_DIR_ENV_VAR, 'false')).lower() != 'false':
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and str(
+            os.environ.get(MOSAICML_LOG_DIR_ENV_VAR, 'false')).lower() != 'false' and os.environ.get(
+                MOSAICML_GPU_LOG_FILE_PREFIX_ENV_VAR) is not None:
```
mvpatel2000, line 487:
```suggestion
    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and str(
            os.environ.get(MOSAICML_LOG_DIR_ENV_VAR, 'false')).lower() != 'false' and os.environ.get(
                MOSAICML_GPU_LOG_FILE_PREFIX_ENV_VAR, 'false').lower() != 'false':
```

mvpatel2000 (approved):
LGTM. nit to be consistent

PR #3029, opened 2024-02-17
Monkeypatch FSDP+TP+Full State Dict errors in Torch

eracah:
Torch throws errors and warnings if it detects you are using a FSDP + TP and want to materialize a StateDictType=FULL state dict. We monkeypatch them to allow this


eracah:
We will add these monkeypatches later...

PR #3030, opened 2024-02-18
[fix auto-microbatch] FSDP reshard and cleanup after OOM to fix the cuda memory leak

bigning:
# What does this PR do?

There are two issues with the current auto-microbatch.
1. nccl timeout or hang
2. it finds a lower device train micro-batch size or still OOM with device_train_microbatch_size = 1

This PR targets at 2nd issue.  [PR 3016 (WIP)](https://github.com/mosaicml/composer/pull/3016) targets the 1st issue. 

For the 2nd memory issue, it's because if OOM happens during fsdp fwd/bwd after unshard, we need to manually reshard and cleanup. Otherwise, there is memory leaks like this:
image

There is an existing FSDP backward callback [_post_backward_final_callback](https://github.com/pytorch/pytorch/blob/2864a7e161cc107f7e4c00cccdf860a6089c73c3/torch/distributed/fsdp/_runtime_utils.py#L1072-L1074) which tries to [reshard](https://github.com/pytorch/pytorch/blob/2864a7e161cc107f7e4c00cccdf860a6089c73c3/torch/distributed/fsdp/_runtime_utils.py#L1098) and cleanup at the end of fsdp backward. We just call this API after OOM happens. 

# test
## unit test
python -m composer.cli.launcher -n 2 -m pytest -m gpu tests/trainer/test_fsdp.py -k test_fsdp_reshard_after_oom
python -m composer.cli.launcher -n 2 -m pytest -m gpu tests/trainer/test_fsdp.py -k test_fsdp_same_state_after_oom_reshard

### llama2 13B finetuning on 16 A100-40G GPUs with global train batch size = 1024
- baseline: llama2-13b-baseline-new-Ln2FYB OOMed even with device-train-microbatch-size = 1
- test: llama2-13b-test-16-40g-r7z22-IBHkBe run well with device-train-microbatch-size = 2

### llama2 70B finetuning on 32 H100-80G GPUs with global train batch size = 1024
- baseline: llama2-70b-8192-32-h100-80g-baseline-npn6Qj OOMed  even with device-train-microbatch-size = 1
- test: llaama2-70b-8192-32-h100-80g-test-vbLmLL chose  device-train-microbatch-size=2 and run well before evaluation. After evaluation, it has nccl timeout error, which is a different issue. We'll debug and fix the timeout in a separate PR. 
- fixed device-train-microbatch-size=4: llama2-70b-32-h100-80g-fixed-4-KKDQ7I OOMed.


composer/trainer/trainer.py
```diff
@@ -232,6 +234,16 @@ def _is_cuda_oom(e: RuntimeError):
...
 
 
+def _fsdp_reshard_and_cleanup(model: torch.nn.Module):
+    for name, module in model.named_modules():
+        if isinstance(module, FullyShardedDataParallel):
+            if module.check_is_root():
+                try:
+                    _post_backward_final_callback(module, module)
+                except Exception as e:
+                    log.warning(f'Failed to reshard fsdp after oom, error: {e}')
```
mvpatel2000, line 244: Have you ever seen this?

composer/trainer/trainer.py
```diff
@@ -232,6 +234,16 @@ def _is_cuda_oom(e: RuntimeError):
     return False
 
 
+def _fsdp_reshard_and_cleanup(model: torch.nn.Module):
+    for name, module in model.named_modules():
```
mvpatel2000, line 238: can you add a description here for why/when this is necessary?

tests/trainer/test_fsdp.py
```diff
@@ -272,3 +272,59 @@ def test_fsdp_act_ckpt_offload(
...
+
+    def forward(self, x):
+        x = self.fc1(x)
+        x = self.fc2(x)
+        x = self.fc3(x)
+        return x
+
+    def loss(self, outputs, batch):
+        return torch.sum(outputs)
+
```
mvpatel2000, line 300: I think we already have a model you can import from `tests.common`

tests/trainer/test_fsdp.py
```diff
@@ -272,3 +272,59 @@ def test_fsdp_act_ckpt_offload(
...
+    x = torch.rand([2, 128])
+    output = fsdp_model(x)
+    with pytest.raises(Exception):
+        # Backward triggers the fake OOM exception,
+        # which prevents fsdp reshard and cleanup
+        torch.sum(output).backward()
+
+    fc2_flat_param = fsdp_model.fc2._flat_param
+
+    # without cleanup, model.fc2.flat_params is still in unshard state
```
mvpatel2000, line 324:
```suggestion
    # Without cleanup, model.fc2.flat_params is still in unshard state
```
nit

mvpatel2000 (commented):
This is great! A few minor nits.

Would like @cli99 's thoughts too

composer/trainer/trainer.py
bigning: No, i think it should be safe to remove the try catch.

cli99 (approved):
Nice test. Looks good to me.

awgu:
Thanks for calling this out and great catch! I think that calling the `_post_backward_final_callback` sounds good to me.

bigning:
> Thanks for calling this out and great catch! I think that calling the `_post_backward_final_callback` sounds good to me.

Thank you @awgu for the quick response!

composer/trainer/trainer.py
dakinggg: Yeah I think if this fails it should probably just fail the run, not just log?

composer/trainer/trainer.py
```diff
@@ -232,6 +234,16 @@ def _is_cuda_oom(e: RuntimeError):
...
+        if isinstance(module, FullyShardedDataParallel):
+            if module.check_is_root():
```
dakinggg, line 240: Can you add a comment here explaining why to only apply this on root? Since its using a private API :)

tests/trainer/test_fsdp.py
```diff
@@ -272,3 +272,59 @@ def test_fsdp_act_ckpt_offload(
...
+    # the full param is not freed
+    assert fc2_flat_param.data_ptr() != fc2_flat_param._local_shard.data_ptr()
```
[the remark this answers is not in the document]
bigning: > if the param is not freed, the data ptr of the flat param will be different than the data ptr of the local shard

basically yes, looking at the flat_param.reshard(), it has [two high level steps](https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/_flat_param.py#L1705-L1707): 
1. switch the flat_param.data_ptr() from `_full_param_padded` to `_local_shard`
2. free the storage of `full_params_padded`

> _full_param_padded part, is that just empty if the full param has been freed?
I think yes. 
I tried adding `assert fc2_flat_param._full_param_padded.numel() == 0` after the reshard and cleanup, but it failed, because some internal tensor check. i guess the storage is not freed in a normal way. After freeing the storage, it's not a valid tensor anymore, any tensor api call will cause a internal tensor check error.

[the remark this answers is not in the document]
dakinggg: Can we test that second piece via whatever method you've found reliable to tell? like a with pytest.raises on some tensor api call on it. we're relying on internal apis and undocumented behavior here, so would be good to have unit test as thorough as possible.

composer/trainer/trainer.py
```diff
@@ -232,6 +234,25 @@ def _is_cuda_oom(e: RuntimeError):
...
+    Find more information here: https://github.com/mosaicml/composer/pull/3030
+    """
+    for name, module in model.named_modules():
+        if isinstance(module, FullyShardedDataParallel):
+            if module.check_is_root():
+                """
+                Only call _post_backward_final_callback on root module. It will
+                traverse all the FSDP sub-modules and do the reshard and cleanup
+                on all sub-modules
+                """
```
mvpatel2000, line 252:
```suggestion
                # Only call _post_backward_final_callback on root module. It will
                # traverse and reshard all FSDP sub-modules
```
nit: slightly more concise and uses # to be more consistent

composer/trainer/trainer.py
```diff
@@ -232,6 +234,25 @@ def _is_cuda_oom(e: RuntimeError):
...
 
 
+def _fsdp_reshard_and_cleanup(model: torch.nn.Module):
+    """
+    This function manually reshards and cleans-up FSDP model by calling
+    _post_backward_final_callback on FSDP root module. In normal case, _post_backward_final_callback
+    is registered as a backward callback. But when exception happens, like OOM, that callback
+    is skipped. There will be memory leak if we don't call that callback.
+    Find more information here: https://github.com/mosaicml/composer/pull/3030
+    """
```
mvpatel2000, line 244:
```suggestion
    """Manually reshard and clean up FSDP model.
    
    When an exception like OOM happens, _post_backward_final_callback, which
    is registered as a backward callback, will not run. We manually call it to cleanup
    loose memory.
    """
```
Nit: One line summary to be consistent with other comments (not sure why linter isn't complaining...) and slightly more concise description

composer/trainer/trainer.py
```diff
@@ -2479,6 +2501,8 @@ def _train_microbatch(self, use_grad_scaling: bool, current_batch_size: int,
 
             with _get_precision_context(self.state.precision, self.state.precision_config,
                                         self.state.deepspeed_enabled):
+                print(
+                    f'bigning debug batch shape: {self.state.batch[0].shape}, label shape: {self.state.batch[1].shape}')
```
mvpatel2000, line 2508:
```suggestion
```

tests/trainer/test_fsdp.py
```diff
@@ -272,3 +278,114 @@ def test_fsdp_act_ckpt_offload(
...
         else:
             assert not isinstance(trainer.state.model.fc1._fsdp_wrapped_module, CheckpointWrapper)
+
+
+@pytest.mark.gpu
+@world_size(2)
+def test_fsdp_reshard_after_oom(world_size: int):
+    model = SimpleMLP(num_features=128)
+
+    #model.relu._fsdp_wrap = False
```
mvpatel2000, line 298:
```suggestion
```

tests/trainer/test_fsdp.py
```diff
@@ -282,3 +288,114 @@ def test_fsdp_act_ckpt_offload(
...
+    def oom_hook(*args):
+        raise RuntimeError('CUDA out of memory.')
+
+    model.fc2.register_full_backward_hook(oom_hook)
+
+    trainer = Trainer(
+        model=model,
+        fsdp_config={},
+        max_duration='3ba',
+        dist_timeout=20,
```
mvpatel2000, line 309:
```suggestion
```

mvpatel2000 (approved):
This is a really nice PR. Super great catch

dakinggg (approved):

tests/trainer/test_fsdp.py
```diff
@@ -282,3 +284,111 @@ def test_fsdp_act_ckpt_offload(
...
+
+    _fsdp_reshard_and_cleanup(fsdp_model)
+    assert fc2_flat_param.data_ptr() == fc2_flat_param._local_shard.data_ptr()
+    assert fc2_flat_param._full_param_padded._typed_storage()._size() == 0
+
+
+@pytest.mark.gpu
+@world_size(2)
+def test_fsdp_same_state_after_oom_reshard(world_size: int):
+    #Test numerical correctness after continuing to train with smaller batch size after OOM.
```
mvpatel2000, line 329:
```suggestion
    # Test numerical correctness after continuing to train with smaller batch size after OOM.
```

mvpatel2000 (commented):
nit

PR #3036, opened 2024-02-19
Add sorted to logs for checkpoint broadcast

mvpatel2000:
# What does this PR do?

Add sorted to log line for checkpoint broadcast. This ensures it prints consistently with pre-broadcast debug log, making it easier to verify transfer.


eracah (approved):
LGTM

PR #3043, opened 2024-02-21
Download symlink once

mvpatel2000:
# What does this PR do?

Currently, all ranks download symlink. Only global rank 0 should download and instead broadcast to all ranks

manual test:
image

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1431,23 +1431,25 @@ class Trainer:
             assert latest_remote_file_name is not None
             if self.state.fsdp_sharded_state_dict_enabled:
                 ar_object_store = maybe_create_object_store_from_uri(save_folder)
-                # Symlink is on object store.
+                # Symlink is on object store
                 if ar_object_store is not None:
-                    with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -1431,23 +1431,26 @@ def __init__(
...
+                            symlink_file_name = format_name_with_dist(latest_remote_file_name,
+                                                                      self.state.run_name) + '.symlink'
+                            try:
+                                ar_object_store.download_object(symlink_file_name, local_symlink_file)
+                                with open(local_symlink_file, 'r') as f:
+                                    real_path = f.read()
+                                    log.debug(f'Read path {real_path} from symlink file')
+                                autoresume_checkpoint_path = ar_object_store.get_uri(real_path)
+                            except FileNotFoundError:
+                                pass
```
j316chuck, line 1449: should we raise a ValueError here? Else this might broadcast None for `autoresume_checkpoint_path` atm.

j316chuck (approved):

composer/trainer/trainer.py
mvpatel2000: It's intended -- that's if theres no symlink

PR #3044, opened 2024-02-21
Add min size to OCI download

mvpatel2000:
# What does this PR do?

Adds minimum size to OCI download parts to avoid multi-part downloading small files (eg symlink)

Manual test:
image

Also printed number of parts and verified works correctly.

### DIFF
--- a/composer/utils/object_store/oci_object_store.py
+++ b/composer/utils/object_store/oci_object_store.py
@@ -137,6 +137,7 @@ class OCIObjectStore(ObjectStore):
         filename: Union[str, pathlib.Path],
         overwrite: bool = False,
         callback: Optional[Callable[[int, int], None]] = None,
+        min_part_size: int = 128000000,
         num_parts: int = 10,
     ):
         del callback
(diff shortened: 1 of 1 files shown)


bigning (approved):
LGTM!

composer/utils/object_store/oci_object_store.py
```diff
@@ -137,6 +137,7 @@ def download_object(
         filename: Union[str, pathlib.Path],
         overwrite: bool = False,
         callback: Optional[Callable[[int, int], None]] = None,
+        min_part_size: int = 128000000,
         num_parts: int = 10,
```
Skylion007, line 141: As an aside, it also bugs me we just set this at 10 and split the file like this. Doesn't give much opportunity for work stealing and it might be faster if we ensure the requested content matches the chunk_size of the object storage.

Skylion007 (approved):

composer/utils/object_store/oci_object_store.py
mvpatel2000: agree. this is hacky and not super optimized. im hoping OCI can just upstream chunked download -- this really should be captured in an object store SDK. (Fsspec might solve this?)

If we don't migrate to a new soln, I'll definitely look at tweaking

PR #3049, opened 2024-02-22
Add initial support for neuron devices

bfontain:
# What does this PR do?

Add some initial support for neuron devices to composer using the pytorch xla backend.

This PR fixes a few issues:
- Renames is_tpu_installed to is_xla_installed and changes it to store state in a global (vs importing and deleting torch_xla, which could delete a needed import).
- Adds a new device type 'neuron' for training on neuron instances
- Fixes support for AMP on XLA.
- Removes use of xm.optimizer_step (which introduces a second gradient reduce given that the model gets auto wrapped in DDP/FSDP). xm.optimizer_step does 3 things: 1) gradient reduction 2) applying the optimizer 3) xm.mark_step. We already have xm.mark_step in the microbatching loop, so remove the one here improves performance slightly.

Tested using a small model on 2 neuron cores via torchrun. support for composer CLI to come.

# Before submitting


composer/callbacks/speed_monitor.py
```diff
@@ -83,30 +86,47 @@
         'int8': 130e12,
         'int4': 260e12,
     },
+    'trn1': {
```
j316chuck, line 89: Can we add a source sheet from where you got these numbers?

composer/callbacks/speed_monitor.py
```diff
@@ -83,30 +86,47 @@
...
+        'fp32': 190e12 / 2,
+        'tf32': 190e12 / 2,
+        'fp16': 190e12 / 2,
+        'amp_fp16': 190e12 / 2,
```
j316chuck, line 93: Curious if you know why fp16 == fp32?

composer/devices/device_neuron.py
```diff
@@ -0,0 +1,48 @@
...
+    When running on Trn, we automatically set `export PJRT_DEVICE=NEURON`.
+    More details.
+    """
+
+    name = 'neuron'
+    dist_backend = 'xla'
+
+    def __init__(self):
+        import torch_xla.core.xla_model as xm
+        os.environ["NEURON_CC_FLAGS"] = "--auto-cast=none"
```
j316chuck, line 33: Can we please add a comment on where this flag comes from?

composer/trainer/trainer.py
```diff
@@ -1320,7 +1320,7 @@ def __init__(
         if self._train_data_spec is not None:
             self.state.set_dataloader(self._train_data_spec.dataloader, train_dataloader_label,
                                       train_subset_num_batches)
-            if isinstance(self.state.device, DeviceTPU):
+            if self.state.device.dist_backend == 'xla':
```
j316chuck, line 1323:
```suggestion
            if isinstance(self.state.device, DeviceNeuron):
```

j316chuck (commented):
Left a few comments. Thanks for this! Overall LGTM

composer/callbacks/speed_monitor.py
bfontain: Thanks for the catch, I forgot to commit a change, fp32 is actually 4 times slower than fp16.

composer/callbacks/speed_monitor.py
bfontain: Done.

composer/devices/device_neuron.py
bfontain: Done.

composer/trainer/trainer.py
bfontain: I'll like to keep this as self.state.device.dist_backend so it works for all xla based devices (rather than having to add more devices to the isinstance check if another one comes along), what do you think?

j316chuck: Ah got it, makes sense! Thanks

j316chuck (approved):
Thanks for adding this device @bfontain

PR #3050, opened 2024-02-22
Update ruff. Fix PLE & LOG lints

Skylion007:
# What does this PR do?

# Before submitting

Add a PLE code for the PyLint Error Codes (Python code that is very likely to be wrong). I also updated ruff to the latest version and updated the config as well.

### DIFF
--- a/composer/profiler/utils.py
+++ b/composer/profiler/utils.py
@@ -93,5 +93,5 @@ def export_memory_timeline_html(prof: TorchProfile,
 
         with open(path, 'w') as f:
             f.write(html)
-    log.debug('Memory timeline exported to', path, '.')
+    log.debug('Memory timeline exported to %s.', path)
     remove(tmpfile.name)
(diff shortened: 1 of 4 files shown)


composer/profiler/utils.py
```diff
@@ -93,5 +93,5 @@ def export_memory_timeline_html(prof: TorchProfile,
 
         with open(path, 'w') as f:
             f.write(html)
-    log.debug('Memory timeline exported to', path, '.')
+    log.debug('Memory timeline exported to %s.', path)
```
mvpatel2000, line 96:
```suggestion
    log.debug(f'Memory timeline exported to {path}.')
```

mvpatel2000 (approved):
LGTM, ty!

composer/profiler/utils.py
Skylion007: @mvpatel2000 This is actually slower as the fstring will be constructed whether or not the log is in debug mode.  See https://docs.astral.sh/ruff/rules/logging-f-string/ , https://pypi.org/project/flake8-logging-format/ .

PR #3069, opened 2024-02-27
pass prelimiter into ALL ICL datasets

eitanturok:
# What does this PR do?

Currently, we only pass `question_prelimiter` into `InContextLearningQATaskDataset` and `InContextLearningCodeEvalDataset`. If we pass `question_prelimiter` into another ICL dataset such as `InContextLearningMultipleChoiceTaskDataset`, `InContextLearningSchemaTaskDataset`, and `InContextLearningLMTaskDataset`, then `question_prelimiter` is never passed into the actual function and fails silently and goes to the default of `question_prelimiter: " "`. 

To fix this, I just passed the `question_prelimiter` to the other ICL datasets.

NA.

# Before submitting


eitanturok:
As a follow up, I noticed that all the parameters in `build_icl_dataloader()` are passed into each of the ICL task datasets except for `cot_delimiter`, `early_stopping_criteria`, `generations_per_sample`. These parameters are only passed into select ICL task datasets but I think this makes sense because these parameters are only relevant to their select ICL tasks. So this is OK.

maxisawesome:
Can you do an eval run with each of these icl task types with a prelimiter specified to ensure things don't go awry? @eitanturok

ez2rok:
> Can you do an eval run with each of these icl task types with a`question_prelimiter` specified to ensure things don't go awry? @eitanturok

Hi @maxisawesome. Yes, I'll benchmark some evals before and after this PR as a sanity check.

Quick note: In this PR, I added the `question_prelimiter` to the multiple_choice, schema, and language modeling tasks. However, in `tasks_v0.3yaml`, we do not specify a `question_prelimiter` parameter for any of the multiple_choice, schema, and language modeling tasks. So I don't expect anything change when running `tasks_v0.3yaml`, i.e. it should not effect any of the runs that currently use this list of tasks.

maxisawesome:
Awesome, thanks @eitanturok!

ez2rok:
Good catch @maxisawesome. LGTM!

maxisawesome (approved):
add schema tasks

dakinggg (approved):

PR #3075, opened 2024-02-29
Allowing for fractional number of samples per rank

ShashankMosaicML:
# What does this PR do?

This PR makes the composer side changes needed for sequence parallelism. In particular, this PR allows setting fractional device microbatch size and ranks to return fractional samples in the get_num_samples_in_batch function call. Also, I have added code to raise error if we try to use auto microbatching with sequence parallelism.

# Before submitting
- [x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)? Yes
- [x] Is this change a documentation change or typo fix? No
- [x] Was this change discussed/approved in a GitHub issue first? Yes
- [x] Did you update any related docs and document your change? No
- [x] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests)) No
- [x] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites)) Yes


composer/trainer/trainer.py
```diff
@@ -2980,7 +2991,7 @@ def _eval_loop(
                                     rank_num_samples -= 1
                                     num_samples_in_microbatch = data_spec.get_num_samples_in_batch(self.state.batch)
                                     # Skip updating metric if batch is only padded samples
-                                    if num_samples_in_microbatch == 1:
+                                    if num_samples_in_microbatch <= 1 or rank_num_samples <= 0:
```
ShashankMosaicML, line 2994: @mvpatel2000 . In the case of sequence parallelism, rank_num_samples (per gpu) will be less than or equal to 1 (at least right now, until we enable that). So, in that case, is the modification correct? I think for sequence parallelism, the following lines (which are a few lines later) should never be executed.
`                                        self.state.batch = data_spec.split_batch(self.state.batch,
                                                                                 num_samples_in_microbatch - 1)[0]`

composer/trainer/trainer.py
```diff
@@ -2985,7 +2996,7 @@ def _eval_loop(
                                     rank_num_samples -= 1
                                     num_samples_in_microbatch = data_spec.get_num_samples_in_batch(self.state.batch)
                                     # Skip updating metric if batch is only padded samples
-                                    if num_samples_in_microbatch == 1:
+                                    if num_samples_in_microbatch == 1 or rank_num_samples <= 0:
```
ShashankMosaicML, line 2999: @mvpatel2000 . In the case of sequence parallelism, rank_num_samples (per gpu) will be less than or equal to 1 (at least right now, until we enable that). So, in that case, is the modification correct? I think for sequence parallelism, the following lines (which are a few lines later) should never be executed.
`                                        self.state.batch = data_spec.split_batch(self.state.batch,
                                                                                 num_samples_in_microbatch - 1)[0]`

composer/trainer/trainer.py
```diff
@@ -2066,9 +2078,18 @@ def _accumulate_time_across_ranks(
...
+                torch.tensor([num_samples, num_tokens], dtype=torch.float32))
+        else:
+            sample_token_tensor = self.state.device.tensor_to_device(
+                torch.tensor([num_samples, num_tokens], dtype=torch.int))
         dist.all_reduce(sample_token_tensor, reduce_operation='SUM')
+        if isinstance(num_samples, float):
+            sample_token_tensor_int = sample_token_tensor.to(torch.int)
+            if torch.any(torch.ne(sample_token_tensor_int, sample_token_tensor)):
+                raise ValueError('The sums of samples and tokens across ranks should each be integers.')
+            sample_token_tensor = sample_token_tensor_int
```
mvpatel2000, line 2092: Can you put a comment describing the float scenarios here please?

composer/trainer/trainer.py
```diff
@@ -2982,7 +3008,7 @@ def _eval_loop(
                                     rank_num_samples -= 1
                                     num_samples_in_microbatch = data_spec.get_num_samples_in_batch(self.state.batch)
                                     # Skip updating metric if batch is only padded samples
-                                    if num_samples_in_microbatch == 1:
+                                    if num_samples_in_microbatch == 1 or hasattr(data_spec, 'spg_ws'):
```
mvpatel2000, line 3011: Why do we need to do this?

mvpatel2000 (commented):
Can we use a more descriptive name than spg_ws please? Abbreviations should be standardized or we can write out full var name

composer/trainer/trainer.py
ShashankMosaicML: Done, [here](https://github.com/mosaicml/composer/blob/541913d19cd7f3ebfe81add0fd6b9fc9410ab166/composer/trainer/trainer.py#L2082).

composer/trainer/trainer.py
```diff
@@ -2958,7 +2967,9 @@ def _eval_loop(
                 if dist_sampler is not None and drop_last == False and dataset_len is not None:
                     batch_num_samples_tensor = self.state.device.tensor_to_device(torch.tensor(rank_num_samples))
                     dist.all_reduce(batch_num_samples_tensor, reduce_operation='SUM')
-                    batch_num_samples = batch_num_samples_tensor.item()
+                    batch_num_samples = int(batch_num_samples_tensor.item())
+                    if batch_num_samples != batch_num_samples_tensor.item():
+                        raise ValueError('Number of samples in a batch should be integral.')
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

composer/trainer/trainer.py
ShashankMosaicML: For sequence parallelism, we only support microbatch size` 1/seq_parallel_ws`, that is, only one sample per microbatch, which gets distributed over the sequence parallel world. Hence, in that case we cannot reduce the microbatch size. That is why I set `skip_metric_update = True` for that case.

composer/trainer/trainer.py
```diff
@@ -2066,9 +2076,20 @@ def _accumulate_time_across_ranks(
...
+            sample_token_tensor = self.state.device.tensor_to_device(
+                torch.tensor([num_samples, num_tokens], dtype=torch.float32))
+        else:
+            sample_token_tensor = self.state.device.tensor_to_device(
+                torch.tensor([num_samples, num_tokens], dtype=torch.int))
         dist.all_reduce(sample_token_tensor, reduce_operation='SUM')
+        if isinstance(num_samples, float):
+            sample_token_tensor_int = sample_token_tensor.to(torch.int)
+            if torch.any(torch.ne(sample_token_tensor_int, sample_token_tensor)):
+                raise ValueError('The sums of samples and tokens across ranks should each be integers.')
```
mvpatel2000, line 2091: Does this work? What about floating point errors?

composer/trainer/trainer.py
```diff
@@ -2966,7 +2989,9 @@ def _eval_loop(
                 if dist_sampler is not None and drop_last == False and dataset_len is not None:
                     batch_num_samples_tensor = self.state.device.tensor_to_device(torch.tensor(rank_num_samples))
                     dist.all_reduce(batch_num_samples_tensor, reduce_operation='SUM')
-                    batch_num_samples = batch_num_samples_tensor.item()
+                    batch_num_samples = int(batch_num_samples_tensor.item())
+                    if batch_num_samples != batch_num_samples_tensor.item():
+                        raise ValueError('Number of samples in a batch should be an integer.')
```
mvpatel2000, line 2994: is this saying eval doesn't support SP?

composer/trainer/trainer.py
mvpatel2000: What happens if I run sequence parallel + dtms 2? Wouldn't that still work?

mvpatel2000:
Can we rename to something more descriptive than spg_ws? Can we please spell it out since its not a common acronym? See original review.

mvpatel2000:
Test please.

composer/core/data_spec.py
```diff
@@ -171,7 +171,7 @@ def __init__(
         num_tokens: Optional[int] = None,
         device_transforms: Optional[Callable[[Batch], Batch]] = None,
         split_batch: Optional[Callable[[Batch, int], Sequence[Batch]]] = None,
-        get_num_samples_in_batch: Optional[Callable[[Batch], int]] = None,
+        get_num_samples_in_batch: Optional[Callable[[Batch], Union[int, float]]] = None,
```
antoinebrl, line 174: You can leverage the [numerical tower](https://peps.python.org/pep-0484/#the-numeric-tower) as an `int` is a valid `float`.
```suggestion
        get_num_samples_in_batch: Optional[Callable[[Batch], float]] = None,
```

composer/trainer/trainer.py
ShashankMosaicML: dtms has to be 1 for sequence parallelism. The reason being that if you can set dtms = 2, then you can also just halve the sequence parallel world size and set dtms = 1.

composer/core/data_spec.py
mvpatel2000: Is this the recommended path by PEP? I'm personally more in favor of leaving both to be more explicit to user in type hints, but if PEP recommends using numerical tower, maybe it's better to migrate across repo...

composer/trainer/trainer.py
mvpatel2000: Can we enforce this in validation?

composer/trainer/trainer.py
ShashankMosaicML: The `batch_num_samples` here is the sum of `batch_num_samples` across ranks in the sequence parallel group (see [this line](https://github.com/mosaicml/composer/blob/e6b334c908bf11e0935e252329ed577f5aa665e2/composer/trainer/trainer.py#L2991)). The check here only enforces that the sum is an integer.

composer/trainer/trainer.py
ShashankMosaicML: Added checks in composer to raise error if dtms > 1 for seq parallelism.

ShashankMosaicML:
> Test please.

Done!

composer/trainer/trainer.py
ShashankMosaicML: Added logic to allow for some floating point error.

mvpatel2000 (approved):

PR #3076, opened 2024-03-01
Add Iteration related Events

b-chu:
# What does this PR do?
Adds ITERATION_START, ITERATION_END, and ITERATION_CHECKPOINT to Event, Engine, and Callback. Increments iteration during training based on iteration length defined in State. Iteration length is a private variable in State and should have no effect by default.

commit-id:bdbe33f2

---

**Stack**:
- #3087
- #3076 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

# Before submitting

### DIFF
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -353,7 +353,11 @@ class Engine():
 
         # dataloader should be set on all events except INIT/BEFORE_LOAD/AFTER_LOAD/EVAL_STANDALONE_START/EVAL_STANDALONE_END
         if event not in {
-                Event.INIT, Event.BEFORE_LOAD, Event.AFTER_LOAD, Event.EVAL_STANDALONE_START, Event.EVAL_STANDALONE_END
+                Event.INIT,
+                Event.BEFORE_LOAD,
+                Event.AFTER_LOAD,
+                Event.EVAL_STANDALONE_START,
+                Event.EVAL_STANDALONE_END,
         }:
             assert state.dataloader is not None, f'The trainer should have set state.dataloader for event {event}.'
 
@@ -384,15 +388,17 @@ class Engine():
                 exit_code = algorithm.apply(event, self.state, self.logger)
 
             trace_key = f'{algorithm}/{event}'
-            trace[trace_key] = Trace(name=algorithm.__class__.__name__,
-                                     event=event,
-                                     exit_code=exit_code,
-                                     order=order,
-                                     run=True)
+            trace[trace_key] = Trace(
+                name=algorithm.__class__.__name__,
+                event=event,
+                exit_code=exit_code,
+                order=order,
+                run=True,
+            )
 
         if len(trace) > 0:
             self.logger.log_traces(
-                {f'algorithm_traces/{tr.name}/{tr.event}': 1 if tr.run else 0 for _, tr in trace.items()})
+                ({f'algorithm_traces/{tr.name}/{tr.event}': 1 if tr.run else 0 for _, tr in trace.items()}))
 
         return trace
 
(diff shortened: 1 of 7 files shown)


mvpatel2000 (commented):
Can you please add some tests that run multiple iterations? Ideally we have a test that also tries changing iteration length and tests new API in state

mvpatel2000 (approved):

composer/core/engine.py
```diff
@@ -353,7 +353,11 @@ def _assert_dataloader_and_duration_set(state: State, event: Event):
 
         # dataloader should be set on all events except INIT/BEFORE_LOAD/AFTER_LOAD/EVAL_STANDALONE_START/EVAL_STANDALONE_END
         if event not in {
-                Event.INIT, Event.BEFORE_LOAD, Event.AFTER_LOAD, Event.EVAL_STANDALONE_START, Event.EVAL_STANDALONE_END
+                Event.INIT,
```
bcui19, line 356: Just want to confirm these are linting chages in these files?

composer/core/state.py
```diff
@@ -606,6 +608,22 @@ def get_elapsed_duration(self) -> Optional[Time[float]]:
             return None
         return self.timestamp.get(self.max_duration.unit) / self.max_duration
 
+    @property
+    def _iteration_length(self):
+        """The length of an iteration."""
+        return self.__iteration_length
+
+    @_iteration_length.setter
+    def _iteration_length(self, iteration_length: Optional[Union[str, Time[int]]]):
```
bcui19, line 617: Maybe leave some detail as to what an iteration is/why it needs to have time units of an epoch?

composer/trainer/trainer.py
```diff
@@ -2244,6 +2247,14 @@ def _train_loop(self) -> None:
 
                 self.engine.run_event(Event.EPOCH_CHECKPOINT)
 
+                # Increment iteration
+                if (self.state._iteration_length is not None and
+                        self.state.timestamp.epoch_in_iteration == self.state._iteration_length):
+                    self.state.previous_timestamp = self.state.timestamp
+                    self.state.timestamp = self.state.timestamp.to_next_iteration()
```
bcui19, line 2254: Where is `to_next_iteration()` implemented? If it a part of this PR or am I missing sth?

tests/trainer/test_trainer.py
```diff
@@ -1138,6 +1138,25 @@ def test_compile_uncompile_model_weights_trainer_fit(
...
+        trainer = Trainer(
+            model=model,
+            max_duration='5ep',
+            train_dataloader=train_dataloader,
+        )
+        trainer.state._iteration_length = '2ep'
+        trainer.fit()
+
+        assert trainer.state.timestamp.epoch == Time(5, TimeUnit.EPOCH)
+        assert trainer.state.timestamp.iteration == Time(2, TimeUnit.ITERATION)
```
bcui19, line 1158: Just for my sanity, because we're in the middle of the second iteration, this is why the assert passes?

bcui19:
From my perspective, this seems good. Added some comments, but yea it'd be good to add more testing/also include tests on if we have early breaks from an epoch in an iteration.

composer/core/engine.py
b-chu: Yes

composer/trainer/trainer.py
b-chu: It got added in #3017

b-chu:
For early breaks in iteration, this would be implemented by users via callback so wouldn't be tested here

PR #3085, opened 2024-03-04
Improvements to `NeptuneLogger`

AleksanderWWW:
# What does this PR do?

This PR contains the implementation of a couple of suggestions that we received after the release of the `NeptuneLogger` (see the list of planned improvements in the comment below).

# Before submitting


AleksanderWWW:
## Planned improvements

- [ ] Update the parameter name `upload_artifacts` -> `upload_checkpoints`
- [ ] Add support for the `OOM` callback
- [ ] Better handle normalized images logging

composer/loggers/neptune_logger.py
```diff
@@ -232,7 +240,11 @@ def upload_file(
             return
 
         del state  # unused
-        self.base_handler[remote_file_name].upload(str(file_path))
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = os.path.join(temp_dir, str(file_path.name))
+            shutil.copy2(file_path, temp_dir)
+            self.base_handler[remote_file_name].upload(temp_file_path)
```
AleksanderWWW, line 247: As per the `upload_file` docstrings:
```
 After this method returns, training can resume, and the contents of ``file_path`` may change (or be may
  deleted). Thus, if processing the file in the background (as is recommended), it is necessary to first
  copy the file to a temporary directory. Otherwise, the original file may no longer exist, or the logged
  file can be corrupted (e.g., if the logger destination is reading from file while the training loop
  is writing to it).
```

composer/loggers/neptune_logger.py
```diff
@@ -58,7 +60,7 @@ def __init__(
         project: Optional[str] = None,
         api_token: Optional[str] = None,
         rank_zero_only: bool = True,
-        upload_artifacts: bool = False,
+        upload_checkpoints: bool = False,
```
SiddhantSadangi, line 63: Removing `upload_artifacts` would make this a breaking change.
Lets support both `upload_artifacts` and `upload_checkpoint`, and add a deprecation notice to `upload_artifacts`.

composer/loggers/neptune_logger.py
```diff
@@ -232,7 +238,11 @@ def upload_file(
             return
 
         del state  # unused
-        self.base_handler[remote_file_name].upload(str(file_path))
+
+        from neptune.types import File
+
+        with open(str(file_path), 'rb') as fp:
+            self.base_handler[remote_file_name].upload(File.from_content(fp.read(), extension=file_path.suffix))
```
SiddhantSadangi, line 247: Won't this still fail if the file has been deleted before we have uploaded it?

composer/loggers/neptune_logger.py
```diff
@@ -304,4 +324,37 @@ def _validate_image(img: Union[np.ndarray, torch.Tensor], channels_last: bool) -
...
+    from neptune.common.warnings import NeptuneWarning, warn_once
+
+    warn_once(
+        'Image value range is not in the expected range of [0.0, 1.0] or [0, 255]. '
+        'This might be due to the presence of `transforms.Normalize` in the data pipeline. '
+        'Logged images may not display correctly in Neptune.',
+        exception=NeptuneWarning,
+    )
+
+    return _scale_image_to_0_255(img, array_min, array_max)
```
SiddhantSadangi, line 348: If we are already warning that the images might not be displayed correctly, let's not scale them anymore

SiddhantSadangi (changes requested):

composer/loggers/neptune_logger.py
AleksanderWWW: It's still under `1.0.0` version, so I guess breaking changes like that might still be allowed

composer/loggers/neptune_logger.py
AleksanderWWW: We do a similar thing in `neptune-detectron2`, though we use `from_stream`. Not sure if it's a big difference here, but I might be mistaken

composer/loggers/neptune_logger.py
SiddhantSadangi: IMO, that's acceptable. (ideally, we should only `warn_once` unless we can tell exactly which image triggered a warning, but that's a discussion on the client side)

Being warned that the image might not be displayed correctly and then seeing that we've distorted images further by scaling them might be more confusing to the users.

AleksanderWWW: They will anyway be distorted further, because we need to send a png

composer/loggers/neptune_logger.py
SiddhantSadangi: This should be fine.. I was more concerned that the file might have been removed from `file_path` before we can `open()` it, but that should never be the case ideally

composer/loggers/neptune_logger.py
AleksanderWWW: And if we don't scale here, we will get sth like
```
[neptune] [warning] The smallest value in the array is -1.3407506942749023 and the largest value in the array is 1.7876918315887451. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.3407506942749023 and the largest value in the array is 1.7411906719207764. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.8401265144348145 and the largest value in the array is 1.968528151512146. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.6932512521743774 and the largest value in the array is 1.599104881286621. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.9023436307907104 and the largest value in the array is 2.025362491607666. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.752001404762268 and the largest value in the array is 1.911994457244873. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.9023436307907104 and the largest value in the array is 1.7269821166992188. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
[neptune] [warning] The smallest value in the array is -1.3701257705688477 and the largest value in the array is 2.025362491607666. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.
```

SiddhantSadangi: Yeah. This is what I was referring to with a `warn_once`, because multiple such warnings without providing which image triggered the warning are not helpful.

We can go ahead with scaling if the images are going to be distorted for rendering on the app anyway

szysad (approved):

AleksanderWWW:
Probably @normandy7 should also be a part of this PR :smiley:

cli99 (approved):
LGTM. thanks.

mvpatel2000 (commented):
LGTM! I like the refactor for dataclasses -- we're planning on moving to that more and more going forward :) 

Only concern is the breaking change

AleksanderWWW:
@mvpatel2000 thank you! As for the breaking change we can go with @SiddhantSadangi 's suggestion :)

composer/loggers/neptune_logger.py
AleksanderWWW: @SiddhantSadangi done. @normandy7 would you like to review the warning message?

composer/loggers/neptune_logger.py
```diff
@@ -304,4 +330,46 @@ def _validate_image(img: Union[np.ndarray, torch.Tensor], channels_last: bool) -
...
+        if isinstance(callback, OOMObserver):
+            return callback
+    return None
+
+
+def _warn_about_deprecated_upload_artifacts() -> None:
+    from neptune.common.warnings import NeptuneDeprecationWarning, warn_once
+    warn_once(
+        'The \'upload_artifacts\' parameter is deprecated. '
+        'Please use the \'upload_checkpoints\' parameter instead.',
```
SiddhantSadangi, line 373:
```suggestion
        'The \'upload_artifacts\' parameter is deprecated and will be removed in the next version. '
        'Please use the \'upload_checkpoints\' parameter instead.',
```

AleksanderWWW: Technically it's already removed, as we don't do literally anything with a passed value

SiddhantSadangi: Created a PR with the changes I had in mind for supporting `upload_artifacts`: https://github.com/AleksanderWWW/composer/pull/1/files

AleksanderWWW:
@normandy7 doctest failing. Could you take a look?

AleksanderWWW:
@mvpatel2000 Could we ask for a rerun of the checks :smiley:

AleksanderWWW:
@normandy7 doctest still doing this :cry:

docs/source/trainer/file_uploading.rst
```diff
@@ -161,6 +161,30 @@ Weights & Biases Artifacts
...
+    # Configure the Neptune logger
+    logger = NeptuneLogger(
+        upload_checkpoints=True,  # enable logging of checkpoint files
+    )
+
+    # Define the trainer
+    trainer = Trainer(..., loggers=logger)
+
+    # Train
+    trainer.fit()
```
SiddhantSadangi, line 186:
```suggestion
```python
    from composer.loggers import NeptuneLogger
    from composer import Trainer

    # Configure the Neptune logger
    logger = NeptuneLogger(
        upload_checkpoints=True,  # enable logging of checkpoint files
    )

    # Define the trainer
    trainer = Trainer(..., loggers=logger)

    # Train
    trainer.fit()
```    
```

**Review by SiddhantSadangi — CHANGES_REQUESTED** [review]

File: docs/source/trainer/file_uploading.rst (hunk 6 above)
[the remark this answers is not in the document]
Comment (mvpatel2000) [reply docs/source/trainer/file_uploading.rst:186]: Note: it seems to fail as the API token is non-existant:
```
    CODE
    Pass the token to the init_run() function via the api_token argument:
        neptune.init_run(project='WORKSPACE_NAME/PROJECT_NAME', api_token='YOUR_API_TOKEN')

    ENVIRONMENT VARIABLE (Recommended option)
    or export or set an environment variable depending on your operating system:

        Linux/Unix
        In your terminal run:
            export NEPTUNE_API_TOKEN="YOUR_API_TOKEN"

        Windows
        In your CMD run:
            set NEPTUNE_API_TOKEN="YOUR_API_TOKEN"
```
I think you just need to mock the token (though last time there was some discussion on other env vars you had to set to run in offline mode....?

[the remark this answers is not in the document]
Comment (mvpatel2000) [reply docs/source/trainer/file_uploading.rst:186]: Also, you should be able to run it locally

**Comment by normandy7** [comment]
> In the last commit I did two things:
> ...
> Could you both take a look and let me know if my thinking is correct?

No objections from me. Hopefully you discovered the root cause!

**Comment by AleksanderWWW** [comment]
That didn't work, but I know where the issue is - we actually pass the mode in the constructor with the default value of `async`. That's why the env var is not working. I will try to come up with a solution

**Comment by AleksanderWWW** [comment]
Ok, it seems that it finally worked :rocket:

**Review by mvpatel2000 — APPROVED** [review]

File: composer/loggers/neptune_logger.py [hunk 7]
```diff
@@ -299,6 +303,16 @@ def post_close(self) -> None:
...
+        if self._oom_observer:
+            self._log_oom_snapshots()
+
+    def _log_oom_snapshots(self) -> None:
+        if self._oom_observer is None or self._oom_observer.filename_config is None:
+            return
+
+        for file_name in self._oom_observer.filename_config.list_filenames():
+            self.base_handler[f'{NeptuneLogger.oom_snaphot_namespace}/{file_name}'].upload(file_name)
+
```
antoinebrl, line 315: I'm wondering if this is the correct way the callbacks should be connected to the loggers. I was under the impression the loggers were agnostics to the callbacks, and the callbacks would be the one pushing the artifacts. In this specific case, the OOMObserver should push the files when they are generated or on `close`. All loggers would benefit from capturing the files associated to the OOM snapshot, even if they can't render it.

mvpatel2000: This is a good point -- sorry I missed this in review and rushed this.

OOMObserver does log to callbacks:
```
logger.upload_file(remote_file_name=remote_file_name, file_path=f, overwrite=self.overwrite)
```

@AleksanderWWW do you know why this is needed! 

Thank you for flagging @antoinebrl  :)

AleksanderWWW: Hey,
apologies, I might have misunderstood how the callback works. Next week I will open a PR that will address those changes more carefully. We will remove this way of uploading OOM files and instead work on our normal `upload_file` functionality to handle the oom traces better.
Once again, apologies for the confusion.

SiddhantSadangi: Thanks for pointing this out, @antoinebrl ! 🙌 

Since @AleksanderWWW had consulted me internally before creating this PR, I am at least partly to blame for the confusion 🙈 

We'll create a new PR to address these and a few other minor doc changes.

@mvpatel2000 - when is your next release scheduled?

mvpatel2000: We plan to cut next Tuesday and release by Thursday. Is it possible to get this in before? I'm happy to help out. Otherwise, we could also revert this and do a follow-on PR with changes.

We plan to release more frequently (every 2-3 weeks) so it wouldn't be a huge deal to punt

SiddhantSadangi: I'd prefer reverting and creating a new PR if we cannot meet the Tuesday cutoff. Don't want to expose users to this change :)

@AleksanderWWW - wdyt? Can we meet the Tuesday cutoff?

mvpatel2000: Ok, I will revert for now.

PR #3089, opened 2024-03-05
Adding failure reason log wrapper

jjanezhang:
# Adding failure reason log wrapper
Adding failure reason log wrapper to capture user understandable errors in foundry for Mosaic finetuning runs.


eracah:
Can we do some sort of manual test to see how the exception looks when it's logged to mcloud?

mvpatel2000 (commented):
How is log exception called?

eracah (approved):
whoa that manual test example is pretty cool. I like that it shows up really nicely in the describe run output. LGTM. Thanks, @jjanezhang !

PR #3091, opened 2024-03-06
Check FSDP module has _device_mesh before getting it

eracah:
# What does this PR do?
If torch version <2.2.0 then FSDP module won't have `_device_mesh`, so `State.device_mesh` will raise an error. This PR does a check for `_device_mesh` first before getting the attribute


composer/core/state.py
```diff
@@ -764,6 +764,8 @@ def fsdp_device_mesh(self):
         if self.fsdp_enabled:
             if not hasattr(self.model, 'model'):
                 return None
+            if not hasattr(self.model.model, '_device_mesh'):
+                return None
```
mvpatel2000, line 768:
```suggestion
            if not hasattr(self.model, 'model') or not hasattr(self.model.model, '_device_mesh'):
                return None
```
Minor nit (I understand I gave the suggestion :P)

mvpatel2000 (approved):
LGTM!

composer/core/state.py
eracah: lol

eracah: but then CI has to re-run

PR #3093, opened 2024-03-06
Enforce async logging flush in mlflow logger at `post_close` call

chenmoneygithub:
# What does this PR do?

Enforce async logging flush in mlflow logger at `post_close` call so that there is no left over logging data.

# Before submitting

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -437,6 +437,7 @@ class MLFlowLogger(LoggerDestination):
             import mlflow
 
             assert isinstance(self._run_id, str)
+            mlflow.flush_async_logging()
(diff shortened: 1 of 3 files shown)


composer/loggers/mlflow_logger.py
```diff
@@ -437,6 +437,7 @@ def post_close(self):
             import mlflow
 
             assert isinstance(self._run_id, str)
+            mlflow.flush_async_logging()
```
mvpatel2000, line 440: Should this be non-async? because this is in termination...?

mvpatel2000 (approved):

composer/loggers/mlflow_logger.py
chenmoneygithub: yes, this is a blocking call.

PR #3097, opened 2024-03-07
Ability to load sharded checkpoints with remote symlink load_path

eracah:
# What does this PR do?

Adds ability to read from a symlink load_path from object store
Adds unit test to tes this
Along for the ride: S3_BUCKET env var to use for s3 bucket fixture

TODO:
- [x] manual test: `load-symlink-test-8-2Wmq3V`


composer/utils/file_helpers.py
```diff
@@ -52,6 +52,23 @@
 ]
 
 
+def extract_path_from_symlink(
+    source_path: str,
+    object_store: Optional[Union[ObjectStore, LoggerDestination]] = None,
+) -> str:
+    with tempfile.TemporaryDirectory() as tmpdir:
+        if object_store is not None and isinstance(object_store, ObjectStore):
```
dakinggg, line 60: I _think_ you don't want the `isinstance(object_store, ObjectStore)` condition here? shouldn't RUD work here too? The type indicates `LoggerDestination` is valid input as the `object_store`

composer/utils/checkpoint.py
```diff
@@ -354,11 +356,14 @@ def format(self, state: State, is_deepspeed: bool = False, keep_placeholders: bo
...
 
 def is_checkpoint_legacy_sharded(object_store: Optional[ObjectStore], source_path: str):
+    if source_path.endswith('.symlink') or os.path.islink(source_path):
+        source_path = extract_path_from_symlink(source_path, object_store=object_store)
     metadata_path = str(Path(source_path) / Path('.metadata'))
     if object_store is None:
         return not os.path.exists(metadata_path)
     else:
         try:
+            _, _, metadata_path = parse_uri(metadata_path)
```
mvpatel2000, line 366: why is this change needed? Is it because parent calling fn does this but if we read a symlink, we need to redo this? If so, can we do this in the symlink if statement?

composer/utils/file_helpers.py
mvpatel2000: We should just convert RUD to RD

eracah: yeah it RUD could technically work, but only it is never a RUD here unless the user passed the RUD as `load_object_store` to the `Trainer`, which should be deprecated, but. I guess we gotta support it :(

eracah: If you can find me someone that uses `load_object_store` instead of `load_path`, I'll eat my hat

eracah: > We should just convert RUD to RD

you mean RU?

eracah: > yeah it RUD could technically work, but only it is never a RUD here unless the user passed the RUD as `load_object_store` to the `Trainer`, which should be deprecated, but. I guess we gotta support it :(

WandBLogger is actually the use case we really need to make sure we support `LoggerDestination` for

eracah: resolved

composer/utils/checkpoint.py
eracah: good call. fixed

composer/utils/file_helpers.py
```diff
@@ -52,6 +52,23 @@
...
+            _, _, source_path = parse_uri(source_path)
+            symlink_file_path = os.path.join(tmpdir, 'file.symlink')
+            object_store.download_object(object_name=source_path, filename=symlink_file_path)
+        else:
+            symlink_file_path = os.readlink(source_path)
+        with open(symlink_file_path, 'r') as f:
```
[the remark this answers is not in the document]
eracah: good call. fixed

composer/utils/checkpoint.py
eracah: if it's not a symlink it will still have the prefix in it. calling this `_, _, metadata_path = parse_uri(metadata_path)` is like pressing lock on your car key fob. Useful if your car is unlocked and has no effect if you car is already locked, so we should just do it

mvpatel2000 (approved):
LGTM!

PR #3100, opened 2024-03-08
Replace errors with warnings for eval args

mvpatel2000:
# What does this PR do?

Replace errors with warnings for eval args. 

Along for the ride, cleans up some of the seq parallel stuff by centralizing validation

### DIFF
--- a/composer/core/__init__.py
+++ b/composer/core/__init__.py
@@ -12,7 +12,7 @@ from composer.core.algorithm import Algorithm
 from composer.core.callback import Callback
 from composer.core.data_spec import DataSpec, ensure_data_spec
 from composer.core.engine import Engine, Trace
-from composer.core.evaluator import Evaluator, ensure_evaluator, validate_evaluator
+from composer.core.evaluator import Evaluator, ensure_evaluator
 from composer.core.event import Event
 from composer.core.passes import AlgorithmPass
 from composer.core.precision import Precision, get_precision_context
@@ -45,5 +45,4 @@ __all__ = [
     'JSON',
     'MemoryFormat',
     'TrainerMode',
-    'validate_evaluator',
 ]
--- a/composer/core/__init__.py
+++ b/composer/core/__init__.py
@@ -12,7 +12,7 @@ from composer.core.algorithm import Algorithm
 from composer.core.callback import Callback
 from composer.core.data_spec import DataSpec, ensure_data_spec
 from composer.core.engine import Engine, Trace
-from composer.core.evaluator import Evaluator, ensure_evaluator, validate_eval_automicrobatching
+from composer.core.evaluator import Evaluator, ensure_evaluator, validate_evaluator
 from composer.core.event import Event
 from composer.core.passes import AlgorithmPass
 from composer.core.precision import Precision, get_precision_context
@@ -45,5 +45,5 @@ __all__ = [
     'JSON',
     'MemoryFormat',
     'TrainerMode',
-    'validate_eval_automicrobatching',
+    'validate_evaluator',
 ]
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -13,10 +13,9 @@ from composer.core.data_spec import DataSpec, ensure_data_spec
 from composer.core.event import Event
 from composer.core.state import State
 from composer.core.time import Time
-from composer.devices import Device, DeviceGPU
 from composer.utils import create_interval_scheduler
 
-__all__ = ['Evaluator', 'ensure_evaluator', 'validate_evaluator']
+__all__ = ['Evaluator', 'ensure_evaluator']
 
 
 class Evaluator:
--- a/composer/core/evaluator.py
+++ b/composer/core/evaluator.py
@@ -16,7 +16,7 @@ from composer.core.time import Time
 from composer.devices import Device, DeviceGPU
 from composer.utils import create_interval_scheduler
 
-__all__ = ['Evaluator', 'ensure_evaluator', 'validate_eval_automicrobatching']
+__all__ = ['Evaluator', 'ensure_evaluator', 'validate_evaluator']
 
 
 class Evaluator:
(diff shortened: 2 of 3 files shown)


composer/core/__init__.py
```diff
@@ -45,5 +45,5 @@
     'JSON',
     'MemoryFormat',
     'TrainerMode',
-    'validate_eval_automicrobatching',
+    'validate_evaluator',
```
aspfohl, line 48: Why is this included as part of composer.core? This is backwards breaking right?

composer/core/evaluator.py
```diff
@@ -141,17 +141,27 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
         )
 
 
-def validate_eval_automicrobatching(auto_microbatching: bool, device: Device):
```
b-chu, line 144: Doesn't this break the API?

composer/core/evaluator.py
```diff
@@ -141,17 +141,27 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
...
     `_is_auto_microbatching` check because `device` is not available during `Evaluator`
     initialization.
     """
+    auto_microbatching = evaluator.auto_microbatching
     if auto_microbatching and not isinstance(device, DeviceGPU):
         raise ValueError(
             'Can only use adaptive device_eval_microbatch_size on GPU. Please set device_eval_microbatch_size >= 1.',
         )
+    if evaluator.auto_microbatching and hasattr(evaluator.dataloader, 'seq_parallel_world_size'):
+        raise ValueError('Auto microbatching on evaluators is not compatible with sequence parallelism.')
```
aspfohl, line 157: For my understanding: I either need to set a specific evaluator microbatch or turn off sequence parallelism? Could this message be more actionable?

composer/core/evaluator.py
```diff
@@ -141,17 +141,27 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
...
+    if isinstance(evaluator.dataloader.get_num_samples_in_batch, int) and hasattr(
+        evaluator.dataloader,
+        'seq_parallel_world_size',
+    ) and evaluator.dataloader.get_num_samples_in_batch * evaluator.dataloader.seq_parallel_world_size != 1:  # type: ignore
+        raise ValueError(
+            '`Sequence parallelism requires a microbatch size of 1 distributed over the sequence parallel group.',
```
aspfohl, line 163:
```suggestion
            'Sequence parallelism requires a microbatch size of 1 distributed over the sequence parallel group.',
```

composer/core/evaluator.py
```diff
@@ -141,17 +141,27 @@ def ensure_evaluator(evaluator: Union[Evaluator, DataSpec, Iterable, Dict[str, A
...
     if auto_microbatching and not isinstance(device, DeviceGPU):
         raise ValueError(
             'Can only use adaptive device_eval_microbatch_size on GPU. Please set device_eval_microbatch_size >= 1.',
         )
+    if evaluator.auto_microbatching and hasattr(evaluator.dataloader, 'seq_parallel_world_size'):
+        raise ValueError('Auto microbatching on evaluators is not compatible with sequence parallelism.')
+    if isinstance(evaluator.dataloader.get_num_samples_in_batch, int) and hasattr(
+        evaluator.dataloader,
+        'seq_parallel_world_size',
+    ) and evaluator.dataloader.get_num_samples_in_batch * evaluator.dataloader.seq_parallel_world_size != 1:  # type: ignore
```
aspfohl, line 161: 👀 what are you ignoring? Should fix

composer/core/evaluator.py
mvpatel2000: Technically yes, but this is mostly an internal validation function -- given we are pre 1.0, I think it's fine to do something like this which cleans up code and almost certainly is not used externally

composer/trainer/trainer.py
```diff
@@ -2116,22 +2107,21 @@ def fit(
...
-                    raise ValueError('Specifying `eval_subset_num_batches` without an `eval_dataloader` has no effect.')
-                if eval_interval != 1:
-                    raise ValueError('Specifying `eval_interval` without an `eval_dataloader` has no effect.')
+                    warnings.warn(
+                        f'Specifying `eval_subset_num_batches={eval_subset_num_batches}` without an `eval_dataloader` '
+                        'has no effect. If trying to run an evaluator, make sure `eval_dataloader` is specified. '
+                        'Otherwise, set `eval_subset_num_batches` to default value -1.',
+                    )
+                if eval_interval != 0 and eval_interval != 1:
+                    warnings.warn(
```
aspfohl, line 2120: Will this raise multiple warnings?

mvpatel2000: It should raise multiple warnings as the two args are different? Do you want to batch?

composer/core/evaluator.py
mvpatel2000: seq_parallel_world_size is not typed into torch and is outside our control :(

composer/core/evaluator.py
aspfohl: Rather, could we remove this from composer.core if this is the case?

composer/core/__init__.py
mvpatel2000: https://github.com/mosaicml/composer/pull/3100#discussion_r1518223176

composer/core/evaluator.py
mvpatel2000: Yea good suggestion. Moved to trainer as a private fn

composer/trainer/trainer.py
aspfohl: Ah gotcha. Probably easier to keep as is, batching would just complicate the code

dakinggg (approved):

Issue #3094, opened 2024-03-07
Different device error when running evaluation_loop on "mps" device with ComposeClassifier


hyenal:
##  Description

When launching a training using a `Trainer` with "mps" device with `ComposerClassifier` you will encounter the following issue when running evaluation:
```
RuntimeError: Encountered different devices in metric calculation (see stacktrace for details).
```
The error originates from [this block](https://github.com/mosaicml/composer/blob/d3f4f0743ad7593e2eb380692b82637d2bf05938/composer/trainer/trainer.py#L3019) which transfers the output back to the cpu. But the batch remains in its device. 

So if you use the `update_metric` method from `ComposerClassifier` (see [here](https://github.com/mosaicml/composer/blob/dev/composer/models/tasks/classification.py#L120)) you will be using two tensors that are not on the same device.

I would be happy to contribute to a PR that would move the batch alongside the outputs onto the CPU.

##  Environment

``` 
---------------------------------
System Environment Report        
Created: 2024-03-07 13:41:25 CET
---------------------------------

PyTorch information
-------------------
PyTorch version: 2.1.1
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 14.3.1 (arm64)
GCC version: Could not collect
Clang version: 15.0.0 (clang-1500.1.0.2.5)
CMake version: Could not collect
Libc version: N/A

Python version: 3.10.13 (main, Mar  4 2024, 14:06:47) [Clang 15.0.0 (clang-1500.1.0.2.5)] (64-bit runtime)
Python platform: macOS-14.3.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Pro

Versions of relevant libraries:
[pip3] flake8==7.0.0
[pip3] hs-infrastructure-flake8-plugins==0.2.0
[pip3] mypy==1.8.0
[pip3] mypy-extensions==1.0.0
[pip3] numpy==1.26.3
[pip3] pytorch-ranger==0.1.1
[pip3] torch==2.1.1
[pip3] torch-optimizer==0.3.0
[pip3] torchaudio==2.1.1
[pip3] torchmetrics==1.0.3
[...]

PR #3105, opened 2024-03-11
In case of MPS device also copy batch to CPU

hyenal:
# What does this PR do?

This PR fixes an issue when using MPS device and composer's classifier. When computing metrics, `outputs` is copied to the cpu but not `batch` which creates some downstream error.

Fixes https://github.com/mosaicml/composer/issues/3094

# Before submitting

### DIFF
--- a/composer/models/huggingface.py
+++ b/composer/models/huggingface.py
@@ -591,9 +591,9 @@ class HuggingFaceModel(ComposerModel):
 
     def update_metric(self, batch: Any, outputs: Any, metric: Metric) -> Dict:
         if getattr(metric, 'needs_batch', False):
-            metric_result = metric.update(batch=batch, outputs=outputs, labels=self.labels)
+            metric_result = metric.update(batch=batch, outputs=outputs, labels=self.labels.to(metric.device))
         else:
-            metric_result = metric.update(outputs, self.labels)
+            metric_result = metric.update(outputs, self.labels.to(metric.device))
         if metric_result is not None:
             # Add the metric name once for each datapoint in the batch
             metric_result['metric_name'] = [metric.__class__.__name__ for _ in range(0, batch['input_ids'].shape[0])]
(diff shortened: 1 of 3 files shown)


composer/trainer/trainer.py
```diff
@@ -3327,12 +3337,14 @@ def _eval_loop(
                                                 outputs.append(v)
                                     else:
                                         outputs = self.state.outputs.cpu()
+                                    batch = DeviceCPU().batch_to_device(batch)
```
mvpatel2000, line 3340:
```suggestion
                                    batch = DeviceCPU().batch_to_device(self.state.batch)
```

mvpatel2000 (commented):
Thanks for the PR!

I think there's a minor error -- once fixed looks good to me!

composer/trainer/trainer.py
hyenal: Thank you this is now integrated in this PR

mvpatel2000:
Hm... labels can be a complex type so directly calling to seems to fail tests

hyenal:
To ensure compatibility with `HuggingFaceModel` I had to create a new method to shift labels outside of `eval_forward`. 

I hope this is not a breaking change, it felt a bit more natural to modify the labels outside of this method as well but there may be reasons for doing it there that I am unaware of.

In addition `update_metric` will be slightly less efficient although the shift label operation should be very cheap

composer/models/huggingface.py
```diff
@@ -544,26 +543,19 @@ def eval_forward(self, batch, outputs: Optional[Any] = None):
                 ]
 
         if self.use_logits or batch.get('mode', None) == 'icl_task':
-            # pop labels first to avoid computing loss
```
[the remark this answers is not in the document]
hyenal: Instead of `outputs` I took `metric` instead but overall this should be the same. Due to the variable nature of `labels` I used `batch_to_device` to handle multiple cases at once

dakinggg (approved):

PR #3106, opened 2024-03-12
Add analytics logging to `MosaicMLLogger`

angel-ruiz7:
# Add `log_analytics` function to `MosaicMLLogger`

This adds a `log_analytics` function to the `MosaicMLLogger`, which records these metrics about composer runs on the Mosaic platform:
- `fsdp_config`
    - `sharding_strategy`
    - `state_dict_type`
    - `activation_checkpointing`
    - `forward_prefetch`
    - `backward_prefetch`
    - `mixed_precision`
    - `device_mesh`
- `autoresume`
- `algorithms`
- `cloud_providers`
    - `cloud_provider_data`
    - `cloud_provider_checkpoints`
- `save_interval`
- `loggers`
- `precision`
- `optimizer`

By logging these fields from the `MosaicMLLogger` into a run's metadata, we can then parse the metadata in SQL and add these metrics into UC. This will enable anyone on the genai team to create dashboards / queries for analytical purposes and view the patterns in these metrics across runs on the platform. ([see the genai_runs table for more info](https://adb-2548836972759138.18.azuredatabricks.net/explore/data/users/anna_pfohl/genai_runs?o=2548836972759138))

## Output of the [`Quick Start` Example](https://github.com/mosaicml/composer?tab=readme-ov-file#quick-start)
Screenshot 2024-03-21 at 1 40 20 PM


mvpatel2000 (commented):
Can you provide a brief description of what we are trying to enable?

composer/core/engine.py
```diff
@@ -293,6 +294,10 @@ def run_event(
             self._run_loggers(event)
             self._run_nonlogger_callbacks(event)
             traces = self._run_algorithms(event)
+
+            # If a MosaicMLLogger is present, log analytics for the run to metadata.
+            log_run_analytics(self.logger.destinations)
+
```
mvpatel2000, line 300: `LoggerDestination`s are callbacks, which means they run on events. Instead of directly calling in engine.py, you should add `def init` (see other callbacks) to MosaicMLLogger and call your helper function there

composer/utils/analytics_helpers.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Helpers for logging analytics with the MosaicMLLogger."""
```
mvpatel2000, line 4: this file should be in the test mosaicml logger file -- the tests ~try to mirror code structure

composer/core/engine.py
snarayan21: I think putting the `log_run_analytics` call here would make it run at every `Event` instead of at every step -- is that the intended behavior? As Mihir said, having this run at specific `Event`s is a better way to go about it -- see [here](https://docs.mosaicml.com/projects/composer/en/stable/trainer/callbacks.html#:~:text=Callbacks%20provide%20hooks%20that%20run,%2C%20timing%2C%20or%20model%20introspection.) for more details

composer/loggers/mosaicml_logger.py
```diff
@@ -69,10 +73,12 @@ def __init__(
         log_interval: int = 60,
         ignore_keys: Optional[List[str]] = None,
         ignore_exceptions: bool = False,
+        analytics_data: Optional[Dict[str, Any]] = None,
```
snarayan21, line 76: Can you update the docstring as well?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,6 +102,90 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
     def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
         self._log_metadata(metrics)
 
+    def log_analytics(self,) -> None:
+        if self.analytics_data is None:
+            return
+
+        # Fetch / cast metrics that we want to log from self.analytics_data
+        autoresume: bool = self.analytics_data['autoresume']
```
snarayan21, line 110: if we want to extend this down the line, how easy is it to add a new metric? do we just have to add a new field in the `analytics_data` dict?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,6 +102,90 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+            if isinstance(dataloader, torch.utils.data.DataLoader):
+                metrics['composer/eval_loaders'].append(
+                    json.dumps({
+                        'label': evaluator.label,
+                        'num_workers': dataloader.num_workers,
+                    }),
+                )
+
+        metrics['composer/optimizers'] = [
+            json.dumps(optimizer.state_dict(), sort_keys=True) for optimizer in trainer_state.optimizers
```
snarayan21, line 135: I think this optimizer state dict can be very large -- see [here](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.state_dict.html#torch.optim.Optimizer.state_dict) for reference. Calling `optimizer.state_dict()` like this probably gives us the per-parameter state of the optimizer, which will be a ton of data. Has this logging been tested with a decent-sized model, and do we save what you expect? Mind adding a section about testing to the PR description?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,6 +102,90 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        metrics['composer/save_interval'] = save_interval_str
+
+        if trainer_state.fsdp_config:
+            metrics['composer/sharding_strategy'] = trainer_state.fsdp_config.get('sharding_strategy', None)
+            metrics['composer/activation_checkpointing'] = trainer_state.fsdp_config.get(
+                'activation_checkpointing',
+                False,
+            )
+            metrics['composer/forward_prefetch'] = trainer_state.fsdp_config.get('forward_prefetch', False)
+            metrics['composer/backward_prefetch'] = trainer_state.fsdp_config.get('backward_prefetch', None)
```
snarayan21, line 171: Looks like the default for backward prefetch is `BackwardPrefetch.BACKWARD_PRE`, see [here](https://pytorch.org/docs/stable/fsdp.html). Also, if torch changes these defaults then the logged metrics may be wrong. One idea is to initialize a `FullyShardedDataParallel` object on a dummy module with no args and retrieve the default values from that, maybe that'll work

snarayan21 (commented):
some comments!

composer/loggers/mosaicml_logger.py
angel-ruiz7: yup! after adding it to the `analytics_data`, you can log it to metadata however you like!

once it's being logged in the metadata, we can easily pull it out and add it into our [existing table for metrics](https://adb-2548836972759138.18.azuredatabricks.net/explore/data/users/anna_pfohl/genai_runs?o=2548836972759138) of run metrics. 

this is all done from a [notebook](https://github.com/mosaicml/mcloud/pull/3706), pulling out a new value would only be a few lines. Depending on where the metric is pulled from (i.e if it's also in `parameters`), we can backfill these metrics pretty easily as well

composer/loggers/mosaicml_logger.py
angel-ruiz7: Screenshot 2024-03-19 at 5 28 05 PM

here's the logs after running the [quickstart example in the README](https://github.com/mosaicml/composer/tree/angel/add-data-to-metadata-for-analytics?tab=readme-ov-file#quick-start). 

looks like the state dict will get pretty huge -- are there any values that you think will be useful to capture from here? don't want to be logging this unless it's useful.

for `callbacks` maybe we can try to consolidate this too? would it be suitable to just have a list of the `repr` values?

@snarayan21

composer/utils/analytics_helpers.py
angel-ruiz7: [is this good](https://github.com/mosaicml/composer/blob/angel/add-data-to-metadata-for-analytics/tests/loggers/test_mosaicml_logger.py#L82) @mvpatel2000 ? or did you mean something else?

tests/loggers/test_mosaicml_logger.py
```diff
@@ -57,6 +60,53 @@ def _update_metadata(self, run_name, new_metadata):
...
+
+    comet_logger = CometMLLogger()
+    return comet_logger
+
+
+def test_get_logger_type(tmp_path: Path, comet_logger: CometMLLogger):
+    """Test that `get_logger_type` returns the correct logger type."""
+    for logger in LOGGER_TYPES:
+        if logger == CometMLLogger:
+            assert get_logger_type(comet_logger) == 'CometMLLogger'
```
mvpatel2000, line 86: Why is special casing needed for comet logger?

composer/trainer/trainer.py
```diff
@@ -1284,7 +1284,16 @@ def __init__(
             MOSAICML_ACCESS_TOKEN_ENV_VAR,
         ) is not None and not any(isinstance(x, MosaicMLLogger) for x in loggers):
             log.info('Detected run on MosaicML platform. Adding MosaicMLLogger to loggers.')
-            mosaicml_logger = MosaicMLLogger()
+            mosaicml_logger = MosaicMLLogger(
+                analytics_data={
+                    'autoresume': autoresume,
+                    'state': self.state,
+                    'save_interval': save_interval,
+                    'loggers': loggers,
```
mvpatel2000, line 1292: do we need this? seems not as important?

composer/loggers/mosaicml_logger.py
```diff
@@ -62,17 +68,20 @@ class MosaicMLLogger(LoggerDestination):
 
             (default: ``None``)
         ignore_exceptions: Flag to disable logging exceptions. Defaults to False.
+        analytics_data (Dict[str, Any], optional): Analytical metrics to log about the current run. Defaults to ``None``.
```
mvpatel2000, line 71: It seems like a lot more than just this is logged, eg fsdp config?

mvpatel2000 (commented):
I'm a little confused on how we're controlling what we log -- left some comments

composer/loggers/mosaicml_logger.py
mvpatel2000: This is small for SGD and MASSIVE for real optimizers (eg adamw). I would just log the name of the optimizer and call it a day

composer/utils/analytics_helpers.py
```diff
@@ -0,0 +1,42 @@
...
+    CometMLLogger,
+    InMemoryLogger,
+    TensorboardLogger,
+    ProgressBarLogger,
+    RemoteUploaderDownloader,
+    LoggerDestination,
+]
+
+
+def get_logger_type(logger: Any) -> str:
```
mvpatel2000, line 37: type hint should be LoggerDestination

angel-ruiz7:
Added some tests and adjusted the way we display `optimizers`. If `algorithms` is similarly verbose I can do something similar for that as well 👍

composer/loggers/mosaicml_logger.py
```diff
@@ -44,7 +50,25 @@ class MosaicMLLogger(LoggerDestination):
     """Log to the MosaicML platform.
 
     Logs metrics to the MosaicML platform. Logging only happens on rank 0 every ``log_interval``
-    seconds to avoid performance issues.
+    seconds to avoid performance issues. The following metrics are logged upon ``INIT``:
```
mvpatel2000, line 53:
```suggestion
    seconds to avoid performance issues. 
    
    Additionally, the following config metrics are logged upon ``INIT``:
```

composer/loggers/mosaicml_logger.py
```diff
@@ -44,7 +50,25 @@ class MosaicMLLogger(LoggerDestination):
...
+    - ``composer/autoresume``: Whether or not the run can be stopped / resumed during training.
+    - ``composer/precision``: The precision to use for training.
+    - ``composer/train_loader_workers``: The number of workers for the train dataloader.
+    - ``composer/eval_loaders``: A list of dictionaries containing the label and the number of workers for each
+    evaluation dataloader.
+    - ``composer/optimizers``: A list of dictionaries containing the _ for each opimizer.
+    - ``composer/algorithms``: A list of dictionaries containing the _ for each algorithm.
```
mvpatel2000, line 60: what is `_`?

composer/loggers/mosaicml_logger.py
mvpatel2000: Is this in addition to the argument?

mvpatel2000 (commented):
Mostly lgtm! a few minor follow ups.

composer/loggers/mosaicml_logger.py
angel-ruiz7: nope, these are all of the metrics that we parse out of the `analytics_data` variable (the arg) 👍

composer/loggers/mosaicml_logger.py
angel-ruiz7: oops -- placeholder a left in there. just changed this to `A list of dictionaries containing information for each <optimizer / algorithm>`

composer/loggers/mosaicml_logger.py
```diff
@@ -46,6 +52,26 @@ class MosaicMLLogger(LoggerDestination):
     Logs metrics to the MosaicML platform. Logging only happens on rank 0 every ``log_interval``
     seconds to avoid performance issues.
 
+    Additionally, The following metrics are logged upon ``INIT``:
+    - ``composer/autoresume``: Whether or not the run can be stopped / resumed during training.
+    - ``composer/precision``: The precision to use for training.
+    - ``composer/train_loader_workers``: The number of workers for the train dataloader.
```
dakinggg, line 58: num workers isn't really a useful field i think

composer/loggers/mosaicml_logger.py
```diff
@@ -46,6 +52,26 @@ class MosaicMLLogger(LoggerDestination):
...
+    - ``composer/eval_loaders``: A list of dictionaries containing the label and the number of workers for each
+    evaluation dataloader.
+    - ``composer/optimizers``: A list of dictionaries containing information about each opimizer.
```
dakinggg, line 61:
```suggestion
    - ``composer/optimizers``: A list of dictionaries containing information about each optimizer.
```

composer/trainer/trainer.py
```diff
@@ -1284,8 +1284,28 @@ def __init__(
             MOSAICML_ACCESS_TOKEN_ENV_VAR,
         ) is not None and not any(isinstance(x, MosaicMLLogger) for x in loggers):
             log.info('Detected run on MosaicML platform. Adding MosaicMLLogger to loggers.')
-            mosaicml_logger = MosaicMLLogger()
+            mosaicml_logger = MosaicMLLogger(
+                analytics_data={
+                    'autoresume': autoresume,
```
dakinggg, line 1289: would be better to pull out the analytics dictionary instead of having it defined in two places. Also suggest definining a dataclass in the mosaicmllogger file instead of just relying on a dictionary with arbitrary keys

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+
+        self.log_metrics(metrics)
+        self._flush_metadata(force_flush=True)
+
     def log_exception(self, exception: Exception):
         self._log_metadata({'exception': exception_to_json_serializable_dict(exception)})
         self._flush_metadata(force_flush=True)
 
+    def init(self, state: State, logger: Logger) -> None:
+        self.log_analytics()
```
dakinggg, line 224: state should just get passed along here. It does not need to be in the stashed analytics dictionary

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        # we log 'callable' as the save_interval value for analytics.
+        if isinstance(save_interval, Union[str, int]):
+            save_interval_str = str(save_interval)
+        elif isinstance(save_interval, Time):
+            save_interval_str = f'{save_interval._value}{save_interval._unit}'
+        else:
+            save_interval_str = 'callable'
+        metrics['composer/save_interval'] = save_interval_str
+
+        if trainer_state.fsdp_config:
```
dakinggg, line 191: since we have the full fsdp config, can we just log that as a json key instead of unpacking it into individual keys?

composer/utils/analytics_helpers.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Helpers for logging analytics with the MosaicMLLogger."""
+
+from composer.loggers.cometml_logger import CometMLLogger
```
dakinggg, line 6: this is very brittle, and also its better to not have the global utils rely on the library itself. Can we just log the class names no matter what instead of this hardcoded list?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        metrics['composer/loggers'] = [
+            get_logger_type(logger) if not isinstance(logger, MosaicMLLogger) else 'MosaicMLLogger'
+            for logger in loggers
+        ]
+
+        # Take the service provider out of the URI and log it to metadata. If no service provider
+        # is found (i.e. backend = ''), then we assume 'local' for the cloud provider.
+        if load_path is not None:
+            backend, _, _ = parse_uri(load_path)
+            metrics['composer/cloud_provider_data'] = backend if backend else 'local'
```
dakinggg, line 176: Should be `composer/cloud_provided_load_path`?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        if save_folder is not None:
+            backend, _, _ = parse_uri(save_folder)
+            metrics['composer/cloud_provider_checkpoints'] = backend if backend else 'local'
```
dakinggg, line 179:
```suggestion
            metrics['composer/cloud_provider_save_folder'] = backend if backend else 'local'
```

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+                )
+
+        get_optimizer_args = lambda optimizer: {
+            k: v for k, v in optimizer.__dict__.items() if not k.startswith('_') and k != 'param_groups'
+        }
+        metrics['composer/optimizers'] = [{
+            optimizer.__class__.__name__: get_optimizer_args(optimizer),
+        } for optimizer in trainer_state.optimizers]
+        metrics['composer/algorithms'] = [{
+            algorithm.__class__.__name__: algorithm.__dict__,
```
dakinggg, line 164: I think you probably don't want to log `.__dict__`, as there is a lot of stuff in the `.__dict__` of a python object.

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +125,104 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+            dataloader = evaluator.dataloader.dataloader
+            if isinstance(dataloader, torch.utils.data.DataLoader):
+                metrics['composer/eval_loaders'].append(
+                    json.dumps({
+                        'label': evaluator.label,
+                        'num_workers': dataloader.num_workers,
+                    }),
+                )
+
+        get_optimizer_args = lambda optimizer: {
```
dakinggg, line 157: this is a bit dangerous (also please define as a private function in this file rather than a lambda). I suggest you just get `optimizer.defaults`

dakinggg (changes requested):

composer/loggers/mosaicml_logger.py
angel-ruiz7: will just use `optimizer.defaults` instead of `__dict__`, but what do you think is a good alternative for `__dict__` when it comes to subclasses of `Algorithm`? do you think `__repr__` would be a good alternative?

dakinggg: I think its ok to just get the class name

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,63 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
 
+    def log_analytics(self, state: State) -> None:
+        if self.analytics_data is None:
+            return
+
+        metrics: Dict[str, Any] = {
+            'composer/autoresume': self.analytics_data.autoresume,
+            'composer/precision': state.precision,
+        }
+        metrics['composer/eval_loaders'] = []
```
angel-ruiz7, line 126: just logging the `evaluator.label` for the `eval_loaders` field right now. is this sufficient?

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,63 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        for evaluator in state.evaluators:
+            dataloader = evaluator.dataloader.dataloader
+            if isinstance(dataloader, torch.utils.data.DataLoader):
```
dakinggg, line 129: you don't need to pull out the dataloader anymore right? just get `evaluator.label`

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,63 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        # Take the service provider out of the URI and log it to metadata. If no service provider
+        # is found (i.e. backend = ''), then we assume 'local' for the cloud provider.
+        if self.analytics_data.load_path is not None:
+            backend, _, _ = parse_uri(self.analytics_data.load_path)
+            metrics['composer/cloud_provided_load_path'] = backend if backend else 'local'
+        if self.analytics_data.save_folder is not None:
+            backend, _, _ = parse_uri(self.analytics_data.save_folder)
+            metrics['composer/cloud_provided_save_folder'] = backend if backend else 'local'
+
+        # Save interval can be passed in w/ multiple types. If the type is a function, then
```
dakinggg, line 148: Is there some idea for utility of analytics on save interval? Nothing is coming to mind for me

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,63 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+            save_interval_str = f'{self.analytics_data.save_interval._value}{self.analytics_data.save_interval._unit}'
+        else:
+            save_interval_str = 'callable'
+        metrics['composer/save_interval'] = save_interval_str
+
+        if state.fsdp_config:
+            # Keys need to be sorted so they can be parsed consistently in SQL queries
+            metrics['composer/fsdp_config'] = json.dumps(state.fsdp_config, sort_keys=True)
+
+        if state.fsdp_state_dict_type is not None:
```
dakinggg, line 162: this is in the fsdp config, so i think remove here

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,63 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+
+        self.log_metrics(metrics)
+        self._flush_metadata(force_flush=True)
+
     def log_exception(self, exception: Exception):
         self._log_metadata({'exception': exception_to_json_serializable_dict(exception)})
         self._flush_metadata(force_flush=True)
 
+    def init(self, state: State, logger: Logger) -> None:
+        self.log_analytics(state)
```
dakinggg, line 173: can we try/except this? I don't want to crash because we failed to log analytics data

tests/loggers/test_mosaicml_logger.py
```diff
@@ -384,3 +386,52 @@ def test_epoch_zero_no_dataloader_progress_metrics():
     assert training_progress['training_progress'] == '[epoch=1/3]'
     assert 'training_sub_progress' in training_progress
     assert training_progress['training_sub_progress'] == '[batch=1]'
+
+
+@pytest.fixture
```
dakinggg, line 391: why was this comet stuff added?

composer/loggers/mosaicml_logger.py
angel-ruiz7: this was requested in a [Slack thread](https://databricks.slack.com/archives/C02EB3VCYRK/p1709073055904539?thread_ts=1709067237.747639&cid=C02EB3VCYRK) a while back. this included some less-helpful metrics (i.e. `num_workers`) so if it isn't useful i'll take it out.

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,57 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
     def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
         self._log_metadata(metrics)
 
+    def log_analytics(self, state: State, loggers: Tuple[LoggerDestination, ...]) -> None:
```
antoinebrl, line 118: Should this be a callback instead? It seems like all loggers and experiment tracking tools would benefit from having these extra information for navigation or reproducibility purposes.

composer/loggers/mosaicml_logger.py
dakinggg: yeah i think just drop this one

composer/loggers/mosaicml_logger.py
```diff
@@ -96,10 +115,57 @@ def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
...
+        self._flush_metadata(force_flush=True)
+
     def log_exception(self, exception: Exception):
         self._log_metadata({'exception': exception_to_json_serializable_dict(exception)})
         self._flush_metadata(force_flush=True)
 
+    def init(self, state: State, logger: Logger) -> None:
+        try:
+            self.log_analytics(state, logger.destinations)
+        except:
```
dakinggg, line 166: can we log the exception here? We should also probably only log once. if its failing consistently, we don't want to spam the run with warnings

dakinggg (approved):
LGTM! Sorry, lost track of this PR

mvpatel2000 (commented):
Can we just log this all callbacks?

dakinggg:
@mvpatel2000 are you suggesting a new logger function `log_analytics` that broadcasts to all LoggerDestinations? not sure how useful that is. In foundry for example, we already log the full config to the ones where it makes sense (wandb, mlflow).

snarayan21:
@mvpatel2000 @angel-ruiz7 @dakinggg is this one basically done? or are there still concerns?

PR #3107, opened 2024-03-12
Fix error on HYBRID_SHARD misconfiguraiton

abhi-mosaic:
# What does this PR do?

If you use `HYBRID_SHARD` but only specify a single element list for `device_mesh`, it gets to this ValueError, but then the message within the `ValueError` fails, because apparently when `DeviceMesh` is 1-D, `device_mesh.get_group()` returns a `ProcessGroup` object rather than a `List[ProcessGroup]`.

Using `device_mesh.ndim` for the error reporting sidesteps the problem.

### DIFF
--- a/composer/trainer/mosaic_fsdp_utils.py
+++ b/composer/trainer/mosaic_fsdp_utils.py
@@ -772,7 +772,7 @@ if version.parse(torch.__version__) > version.parse('2.2.9') and version.parse(
             else:
                 raise ValueError(
                     'Expected device_mesh to have ndim=2 '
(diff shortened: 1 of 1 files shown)


b-chu (approved):
Thanks for the PR!

mvpatel2000:
Is it not caught earlier? It should be checked here:
https://github.com/mosaicml/composer/blob/f65bb27800810240479c1d4802f124a9b60b2096/composer/trainer/dist_strategy.py#L319-L342

mvpatel2000:
Closing, this has been redone in the TP PR

PR #3109, opened 2024-03-13
Fix rounding issue in interval calculation

dakinggg:
# What does this PR do?
Previously we were implicitly rounding down in the interval calculation because we were multiplying a decimal with a `Time`. The subsequent `math.ceil` therefore did not have the intended effect. This PR remedies that by using `.value` on the time, so the `math.ceil` properly applies to the unrounded decimal. Note: this will slightly change the exact pattern of when the interval scheduler returns true, but it will bring the implementation into line whether your max duration is in epochs or batches. Previously they were different.

The impetus for this is that it resolves a div by zero error if you only have one batch and have a `0.XXdur` interval. Manually tested it is resolved in addition to unit tests.

Along for the ride, fixes some typing that was complaining locally.

### DIFF
--- a/composer/callbacks/eval_output_logging_callback.py
+++ b/composer/callbacks/eval_output_logging_callback.py
@@ -104,12 +104,12 @@ class EvalOutputLogging(Callback):
         self.rows.extend(rows)
 
     def eval_end(self, state: State, logger: Logger) -> None:
+        if not self.name or self.columns is None:
+            return
+
         list_of_rows = all_gather_object(self.rows)
         rows = [row for rows in list_of_rows for row in rows]
         for dest_logger in logger.destinations:
-            assert self.name is not None
-            assert self.columns is not None
-
             if not isinstance(dest_logger, ConsoleLogger):
                 dest_logger.log_table(self.columns, rows, name=self.name, step=state.timestamp.batch.value)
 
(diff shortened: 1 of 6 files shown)


docs/source/trainer/file_uploading.rst
```diff
@@ -169,7 +169,7 @@ Neptune File upload
     The :class:`~composer.loggers.neptune_logger.NeptuneLogger` API Reference.
 
 .. testcode::
-    :skipif: not _NEPTUNE_INSTALLED
+    :skipif: True
```
mvpatel2000, line 172: why?

composer/callbacks/eval_output_logging_callback.py
```diff
@@ -104,6 +104,9 @@ def eval_batch_end(self, state: State, logger: Logger) -> None:
         self.rows.extend(rows)
 
     def eval_end(self, state: State, logger: Logger) -> None:
+        if not self.name or self.columns is None:
+            return
```
mvpatel2000, line 108: self.name is set later in this function -- this check makes callback a noop?

mvpatel2000 (commented):
Primary changes are fine, do not understand other changes.

docs/source/trainer/file_uploading.rst
dakinggg: because the doctests were failing on this...

composer/callbacks/eval_output_logging_callback.py
dakinggg: im pretty sure it gets set in eval_batch_end event, but only if you're processing valid batches for this callback. so you need to also skip the log table calls here if you're not processing valid batches for this callback. i'm gonna just revert this change and do separately though so i can get this pr in.

mvpatel2000 (approved):

PR #3115, opened 2024-03-13
Add checks to schedulers

b-chu:
# What does this PR do?

Changes the error checking in scheduler to check
1. if max_duration is greater than t_max
2. if t_warmup and t_max are different units

Previously it only checked if t_warmup and max_duration were different units which is incorrect because a valid schedule can have t_warmup and max_duration as different units.

# Before submitting

### DIFF
--- a/composer/optim/scheduler.py
+++ b/composer/optim/scheduler.py
@@ -565,7 +565,7 @@ def _raise_if_max_duration_exceeds_t_max(t_max: Union[str, Time], state: State):
 def _raise_if_warmup_and_max_incompatible(t_warmup: Time[int], t_max: Time[int]):
     """Checks that t_warmup and t_max have the same units.
 
-    _convert_time should be called on both before this function is called. As a a result, t_warmup and t_max will not
+    _convert_time should be called on both `t_warmup` and `t_max` before this function is called. As a a result, t_warmup and t_max will not
     be TimeUnit.EPOCH.
     """
     assert t_warmup.unit != TimeUnit.EPOCH and t_max.unit != TimeUnit.EPOCH, 't_warmup and t_max cannot be in units of EPOCH'
--- a/composer/optim/scheduler.py
+++ b/composer/optim/scheduler.py
@@ -521,26 +521,63 @@ class PolynomialScheduler(ComposerScheduler):
         t_max = _convert_time(self.t_max, state, ssr=ssr)
         current_time = state.timestamp.get(t_max.unit)
         frac_of_total = (current_time / t_max).value
+        frac_of_total = min(1.0, frac_of_total)
 
         coeff = (1 - frac_of_total)**self.power
         current_factor = self.alpha_f + coeff * (1.0 - self.alpha_f)
         return current_factor
 
 
-def _raise_if_warmup_and_max_duration_incompatible(t_warmup: Union[str, Time], t_max: Union[str, Time]):
+def _raise_if_max_duration_exceeds_t_max(t_max: Union[str, Time], state: State):
+    assert state.max_duration is not None, 'max_duration should be set whenever schedulers are invoked'
+    max_dur = state.max_duration
+    if isinstance(t_max, str):
+        t_max = Time.from_timestring(t_max)
+    if isinstance(max_dur, str):
+        max_dur = Time.from_timestring(max_dur)
+
+    max_dur_exceeds_t_max = False
+    if t_max.unit == max_dur.unit and t_max.value < max_dur.value:
+        max_dur_exceeds_t_max = True
+    elif (
+        t_max.unit == TimeUnit.BATCH and max_dur.unit == TimeUnit.EPOCH and state.dataloader_len is not None and
+        t_max.value < max_dur.value * int(state.dataloader_len)
+    ):
+        max_dur_exceeds_t_max = True
+    elif (
+        t_max.unit == TimeUnit.EPOCH and max_dur.unit == TimeUnit.BATCH and state.dataloader_len is not None and
+        t_max.value * int(state.dataloader_len) < max_dur.value
+    ):
+        max_dur_exceeds_t_max = True
+    elif t_max.unit != max_dur.unit:
+        log.info(
+            f'Since max_duration {max_dur} with units {max_dur.unit} and t_max {t_max} with units {t_max.unit} are not '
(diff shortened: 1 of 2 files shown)


composer/optim/scheduler.py
```diff
@@ -521,26 +521,63 @@ def __call__(self, state: State, ssr: float = 1.0):
...
+        t_max.value < max_dur.value * int(state.dataloader_len)
+    ):
+        max_dur_exceeds_t_max = True
+    elif (
+        t_max.unit == TimeUnit.EPOCH and max_dur.unit == TimeUnit.BATCH and state.dataloader_len is not None and
+        t_max.value * int(state.dataloader_len) < max_dur.value
+    ):
+        max_dur_exceeds_t_max = True
+    elif t_max.unit != max_dur.unit:
+        log.info(
```
snarayan21, line 553: why not warning here? Also, i think the message could be more explicit. Instead of "make sure that your LR schedule is defined at all points", we probably want to say something that tells the user that max_duration should never exceed t_max, or their LR schedule will be invalid.

composer/optim/scheduler.py
```diff
@@ -521,26 +521,63 @@ def __call__(self, state: State, ssr: float = 1.0):
...
+        raise ValueError(
+            f't_max {t_max} must be greater than or equal to max_duration {max_dur}. Otherwise, the LR schedule will '
+            'not be defined for the entire training duration.',
+        )
+
+
+def _raise_if_warmup_and_max_incompatible(t_warmup: Time[int], t_max: Time[int]):
+    """Checks that t_warmup and t_max have the same units.
+
+    _convert_time should be called on both before this function is called. As a a result, t_warmup and t_max will not
```
snarayan21, line 568:
```suggestion
    _convert_time should be called on both `t_warmup` and `t_max` before this function is called. As a a result, t_warmup and t_max will not
```

tests/optim/test_scheduler.py
```diff
@@ -409,24 +412,62 @@ def test_warmup_schedulers_fail_fast(
...
     state.max_duration = Time.from_timestring(max_duration_str)
     state.timestamp = Timestamp()
     state.set_dataloader([None] * batches_per_epoch, 'train')
 
+    effective_scheduler_max_unit = scheduler_max_unit if scheduler_max_unit != 'dur' else max_duration_unit
+    effective_warmup_duration_unit = warmup_duration_unit if warmup_duration_unit != 'dur' else max_duration_unit
+    max_duration_no_epoch = state.max_duration
+    if max_duration_unit == 'ep':
+        max_duration_no_epoch = Time.from_timestring(max_duration_unit_to_str['ba'])
+    error_context = contextlib.nullcontext()
```
snarayan21, line 443: just for my understanding, what does `nullcontext()` help us do

tests/optim/test_scheduler.py
```diff
@@ -409,24 +412,62 @@ def test_warmup_schedulers_fail_fast(
...
     with error_context:
         for _ in range(num_steps):
             _ = scheduler(state)
             state.timestamp = state.timestamp.to_next_batch(
                 samples=samples_per_batch,
                 tokens=tokens_per_sample * samples_per_batch,
             )
+
+
+def _units_comparable(unit1: str, unit2: str) -> bool:
```
snarayan21, line 466: can you use the `TimeUnit` enums here instead
https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/core/time.py#L32

snarayan21 (commented):
some minor comments!

composer/optim/scheduler.py
b-chu: Every optimizer step will print to the logs and there wouldn't be an easy way to filter it out.

"max_duration should never exceed t_max" Some schedulers have a defined schedule for duration above t_max

tests/optim/test_scheduler.py
b-chu: It basically says that there won't be an error for that test case

tests/optim/test_scheduler.py
b-chu: I'll leave this to a separate PR since pytest params should also use TimeUnit enums if this change is made

composer/optim/scheduler.py
snarayan21: wait really? could you link which ones -- pretty sure that every scheduler that takes in a `t_max` is only supposed to be defined up to that

There are other schedulers with just warmup that don't take t_max, yeah

composer/optim/scheduler.py
```diff
@@ -521,26 +521,63 @@ def __call__(self, state: State, ssr: float = 1.0):
         t_max = _convert_time(self.t_max, state, ssr=ssr)
         current_time = state.timestamp.get(t_max.unit)
         frac_of_total = (current_time / t_max).value
+        frac_of_total = min(1.0, frac_of_total)
```
snarayan21, line 524: shouldn't we raise error if `frac_of_total > 1.0`? As in, if current time exceeds t_max? This is the slow failing behavior we talked about right

snarayan21 (approved):
lgtm, would appreciate clarifiation on questions before merging though.

composer/optim/scheduler.py
b-chu: Linked some examples below

composer/optim/scheduler.py
b-chu: Seems like most schedulers account for this behavior already, but it was missing from this one so adding it to make all schedulers consistent

https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/optim/scheduler.py#L350
https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/optim/scheduler.py#L394
https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/optim/scheduler.py#L747
https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/optim/scheduler.py#L822
https://github.com/mosaicml/composer/blob/2f11230a17bd35eb4f6a5784a18e43c648a56b2f/composer/optim/scheduler.py#L899

PR #3116, opened 2024-03-14
Making default_split_batch public

ShashankMosaicML:
# What does this PR do?

This PR makes the _default_split_batch public. It is already being used outside the file it is defined in (see composer/datasets/in_context_learning_evaluation.py) and it also has unit tests (see tests/test_split_batch.py). This function can be used as a base function to write other batch splitting functions.

### DIFF
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -114,6 +114,9 @@ def _default_split_batch(batch: Any, microbatch_size: int) -> Sequence:
     )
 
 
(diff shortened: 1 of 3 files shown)


mvpatel2000 (commented):
Why? Please add PR description

ShashankMosaicML:
> Why? Please add PR description

Done, PTAL

dakinggg (approved):

Issue #3123, opened 2024-03-18
Set `torch.utils.data.DataLoader.batch_sampler` epoch if defined


Ghelfi:
## 🚀 Feature Request

Current implementation of composer trainer call the `DistributedSampler.set_epoch` method only on the `Dataloader.sampler` attribute but not on the `Dataloader.batch_sampler`, even if it is defined.
One example [here](https://github.com/mosaicml/composer/blob/38ebcd85d152cfaf80c38d847c4ed431f4a73545/composer/trainer/trainer.py#L2267C17-L2267C27).

## Motivation

When doing distributed learning based on batch sampler one might want the epoch to be properly set on the `batch_sampler` since it is usually used to seed new seed over time.
This could be useful for Metric learning where `batch_sampler` can be a worth feature.
For now composer trainer only handle regular sampler.

## Implementation

I'll propose a PR with a technical implementation. The idea is to check if batch_sampler is defined, in which case we can set the epoch properly, else we do it on the regular sampler (always defined in `torch.utils.data.DalaLoader`).

Ghelfi:
Merged by @mvpatel2000. Closing the issue.

PR #3124, opened 2024-03-18
Call `set_epoch` on `Dataloader.batch_sampler` if defined

Ghelfi:
# What does this PR do?

Proposes a technical implementation to `set_epoch` on either `DataLoader.sampler` or `DataLoader.batch_sampler` if it is defined.

- Related to #3123

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -3248,6 +3248,7 @@ class Trainer:
                 # Because evaluation can run on each batch, we use the batch to seed the sampler
                 # so each evaluation will get a proper shuffle.
                 # The epoch provided to `set_epoch` need not be sequential, so this is fine.
+                dist_sampler = sampler
                 sampler.set_epoch(int(self.state.timestamp.batch))
                 drop_last = dataloader.drop_last
                 # Only compute the dataset length if drop_last is False, as otherwise we don't need
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -240,7 +240,7 @@ def _set_evaluator_interval_and_subset_num_batches(
                 )
 
 
-def _is_auto_microbatching(device_train_microbatch_size: Optional[Union[int, str]], device: Device):
+def _is_auto_microbatching(device_train_microbatch_size: Optional[Union[int, float, str]], device: Device):
     if device_train_microbatch_size == 'auto':
         warnings.warn((
             "`device_train_microbatch_size='auto'` may potentially fail with unexpected "
@@ -260,10 +260,10 @@ def _is_auto_microbatching(device_train_microbatch_size: Optional[Union[int, str
 
 
 def _get_initial_device_train_microbatch_size(
-    device_train_microbatch_size: Optional[Union[int, str]],
+    device_train_microbatch_size: Optional[Union[int, float, str]],
     auto_microbatching: bool,
     train_dataloader: Optional[Iterable],
-) -> Optional[int]:
+) -> Optional[Union[int, float]]:
     """Sets initial value of device_train_microbatch_size.
 
     If auto_microbatching, sets initial `device_train_microbatch_size` to per rank batch size. If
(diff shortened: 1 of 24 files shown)


mvpatel2000:
Thanks for the PR! 

Would you mind adding a unit test too please? You can extend `test_checkpoint.py` with `class TestCheckpointResumption:`. I would add an arg to `get_trainer` to add a batch dist sampler and add a new fn that tests that path instead.

If you need help, I'm happy to provide assistance for this!

composer/trainer/trainer.py
```diff
@@ -465,6 +465,16 @@ def _generate_run_name() -> str:
...
 
 
+def _get_sampler(dataloader: DataLoader) -> Sampler | Iterable:
+    """Fetch the sampler from a `dataloader`.
+
+    Returns `dalaoder.batch_sampler` is defined, else `dataloader.sampler` (always defined in `Dataloader.__init__`).
+    """
+    if dataloader.batch_sampler is not None:
+        return dataloader.batch_sampler
+    return dataloader.sampler
```
antoinebrl, line 475: Are the two types of samplers mutually exclusive?

Ghelfi: @antoinebrl  While instantiating yes, `sampler` and `batch_sampler` are mutually exclusive.
However, in `torch.utils.data.DataLoader.__init__`, `self.sampler` is instantiated with default sampler is None is provided, regardless of `batch_sampler` value (see [here](https://github.com/pytorch/pytorch/blob/23a6d74f9352e0afb37750fee300d077c4ba9393/torch/utils/data/dataloader.py#L344)). It will not be used if you provided a `batch_sampler`.

So the process if first to check is `batch_sampler` is available, else return the regular `sampler`.

Ghelfi:
@mvpatel2000 cannot add you as reviewer to ask for a review.
I added a test and ready for a review.
I put the test where you suggested. Will it be better to put it in `test_trainer.py`?

tests/trainer/test_checkpoint.py
```diff
@@ -1257,18 +1259,75 @@ def get_trainer(
...
+                dataset=eval_dataset,
+                batch_sampler=_DistributedBatchSampler(
+                    dataset=eval_dataset,
+                    drop_last=False,
+                    shuffle=False,
+                    num_replicas=dist.get_world_size(),
+                    rank=dist.get_global_rank(),
+                    batch_size=train_batch_size,
+                ),
+            )
```
mvpatel2000, line 1323: Can you construct the sampler first and put a ternary there? Then that can be passed directly into the DataLoader so you avoid the whole construction in the ternary

tests/trainer/test_checkpoint.py
```diff
@@ -1257,18 +1259,75 @@ def get_trainer(
...
+                drop_last=True,
+                shuffle=True,
+                num_replicas=dist.get_world_size(),
+                rank=dist.get_global_rank(),
                 batch_size=train_batch_size,
-                sampler=dist.get_sampler(train_dataset),
             ),
-            eval_dataloader=DataLoader(
+        )
+
```
mvpatel2000, line 1307: Same comment as below!

tests/trainer/test_checkpoint.py
```diff
@@ -1257,18 +1259,75 @@ def get_trainer(
         eval_dataset = RandomClassificationDataset(size=12)
         train_batch_size = 2
 
-        return Trainer(
-            model=model,
-            train_dataloader=DataLoader(
+        class _DistributedBatchSampler(DistributedSampler):
```
mvpatel2000, line 1262: should this be outside the helper fn for cleanliness...? Not strongly opinionated, @dakinggg  thoughts?

mvpatel2000 (commented):
LGTM assuming tests pass! I have some minor style nits

tests/trainer/test_checkpoint.py
Ghelfi: Done

tests/trainer/test_checkpoint.py
Ghelfi: Done

mvpatel2000:
@Ghelfi do you mind if I push directly to this branch? there's a few lint things and refactor in test I would prefer. I'm happy to leave as comments if you'd like, but I figure it's faster for me to directly apply. I don't want to delay you further :)

Ghelfi:
@mvpatel2000 fine by me. Thanks for handling it.

mvpatel2000 (approved):

PR #3125, opened 2024-03-18
Add deprecation warnings for ICL datasets/helper functions/metrics

bmosaicml:
# What does this PR do?

We are migrating ICL tasks from composer to foundry and need to deprecate the existing composer implementations. The migration PR is here: https://github.com/mosaicml/llm-foundry/pull/936

# Before submitting

### DIFF
--- a/composer/datasets/in_context_learning_evaluation.py
+++ b/composer/datasets/in_context_learning_evaluation.py
@@ -8,6 +8,7 @@ import copy
 import json
 import os
 import random
+import warnings
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Union
 
 import torch
@@ -17,6 +18,7 @@ from composer.core import DataSpec
 from composer.core.data_spec import _split_list, default_split_batch
 from composer.datasets.utils import stop_sequences_criteria
 from composer.utils import MissingConditionalImportError, dist, get_file
+from composer.utils.warnings import VersionedDeprecationWarning
 
 if TYPE_CHECKING:
     import transformers
@@ -24,7 +26,6 @@ if TYPE_CHECKING:
 
 # Allow models to have slightly more tokens than were used in the most verbose CoT in the dataset
 _MAX_ANSWER_BUFFER_LENGTH = 10
-
 __all__ = [
     'InContextLearningLMTaskDataset',
     'InContextLearningMultipleChoiceTaskDataset',
@@ -44,6 +45,13 @@ def strip_data(example: Dict) -> Dict:
     Returns:
         dict: The same dictionary with .strip() applied to any value in the dict that is a string
     """
+    warnings.warn(
+            VersionedDeprecationWarning(
+                f'`strip_data` has been deprecated and migrated to MosaicML\'s llm-foundry repo under the llmfoundry.eval.datasets.in_context_learning module: "
+                + "https://github.com/mosaicml/llm-foundry/blob/main/scripts/eval/README.md',
+                remove_version='0.23.0',
+            ),
+        )
     return {k: v.strip() if isinstance(v, str) else v for k, v in example.items()}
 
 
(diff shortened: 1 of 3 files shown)


composer/datasets/in_context_learning_evaluation.py
```diff
@@ -44,6 +45,13 @@ def strip_data(example: Dict) -> Dict:
     Returns:
         dict: The same dictionary with .strip() applied to any value in the dict that is a string
     """
+    warnings.warn(
```
dakinggg, line 48: you don't need to explicitly deprecate all of these helper functions. can just do the classes and the builders

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -293,6 +293,13 @@ def __init__(
         hf_parsing_map: Optional[Dict] = None,
         generation_kwargs: Optional[Dict] = None,
     ):
+        warnings.warn(
```
dakinggg, line 296: you can just put this on the base class right? all instantiation paths end up there

composer/metrics/nlp.py
```diff
@@ -273,7 +281,7 @@ def rename_args(
             raise ValueError('Cannot use both `outputs` and `output_logits`')
         if output_logits is not None:
             warnings.warn(
-                VersionedDeprecationWarning('`output_logits` has been renamed to `outputs`.', remove_version='0.21.0'),
+                VersionedDeprecationWarning('`output_logits` has been renamed to `outputs`.', remove_version='0.23.0'),
```
maxisawesome, line 284: This change is already in the migration. Either we should update the code here in composer now (we're past this version) or just boot it to the migration (given that we don't wanna touch anything in the composer eval flow now).

dakinggg: hm ok, yeah lets just remove this deprecation then, and it will go away when the whole thing goes away. sound good?

composer/datasets/in_context_learning_evaluation.py
```diff
@@ -293,6 +293,14 @@ def __init__(
...
+            VersionedDeprecationWarning(
+                '`InContextLearningDataset`, it\'s subclasses, and eval utility functions have been deprecated and migrated'
+                + ' to MosaicML\'s llm-foundry repo under the llmfoundry.eval.datasets.in_context_learning module: ' +
```
dakinggg, line 299: `in_context_learning_evaluation` I think?

composer/datasets/utils.py
```diff
@@ -78,6 +81,13 @@ def __init__(
             tokenizer: transformers.PreTrainedTokenizerBase,
             batch_size: int,
         ) -> None:
+            warnings.warn(
+                VersionedDeprecationWarning(
+                    '`MultiTokenEOSCriteria` has been deprecated and migrated to MosaicML\'s llm-foundry repo under the llmfoundry.eval.datasets.in_context_learning module: '
```
dakinggg, line 86: same

composer/metrics/nlp.py
```diff
@@ -203,6 +203,14 @@ class InContextLearningMetric(Metric):
     """Base class for In-context learning (ICL) metrics."""
 
     def __init__(self, *args, **kwargs):
+        warnings.warn(
+            VersionedDeprecationWarning(
+                '`InContextLearningMetric` and it\'s subclasses have been deprecated and ' +
```
dakinggg, line 208: same

dakinggg (approved):

composer/metrics/nlp.py
maxisawesome: That would remove the warning that the function signature will change with the migration? My instinct is to just change the deprecation version to tell ppl it'll change at the version that the migration is final.

dakinggg: that works too

DhruvDh:
Not sure I understand why this change was needed. As composer is meant for "composing" training runs, language models, being one of them; and running evals during training is important. Why would ICL evals need to be in llm-foundry?

Is training language models using composer a deprecated use-case?

maxisawesome:
@DhruvDh Committing changes to ICL Datasets and metrics required PRs to both foundry and composer which caused not-insignificant delays and complications in the development process. The ICL tasks are solely designed for LLMs and llm-foundry was always intended as the main place for LLM based eval code. The [migration](https://github.com/mosaicml/llm-foundry/pull/936) of ICL tasks to llm-foundry also helped us streamline code and consolidate defaults, which were two major pain points we had found internally.

DhruvDh:
Thank you for your response.  I did end finding that doing ICL eval with composer/llm-foundry was too painful and up rolling my own. I can see now that it is a bit complicated especially if you want to do distributed evaluation, so this change is probably a good idea.

Though ICL like context-continuation evaluation does not be restricted to natural language and llms. I think you usually end up with better abstractions when you try to be general. I think a "composer" purely for evaluation with a similar events/callback api could be ideal.

PR #3126, opened 2024-03-18
Removed default attrs from exception class in the attrs dict

jjanezhang:
# Removed default attrs from exception class in the attrs dict
`test_exception_to_json` was failing because we left the exception attributes in the default Exception class in the attributes value. Created a copy of a generic Exception and it's default attributes to filter them out.

Test is now fixed.


b-chu (approved):
Thanks!

PR #3129, opened 2024-03-18
Remove rng from save_weights_only ckpt

eracah:
# What does this PR do?
if `save_weights_only` is set then don't include rng in checkpoint


dakinggg (commented):
Was there a reason we had the rng in weights only or was it just an oversight?

mvpatel2000 (approved):
Good catch, thanks!

@dakinggg  discussed offline, this is an oversight

PR #3130, opened 2024-03-20
Only broadcast distcp files

mvpatel2000:
# What does this PR do?

Only broadcasts distcp files (sharded checkpoint files) to avoid propagating temp files from download.

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -287,7 +287,9 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
             receiver = dist.get_global_rank() != rank_in_first_replica
 
             # Send list of files to all ranks
-            file_list = [sorted(os.listdir(self.destination_path))]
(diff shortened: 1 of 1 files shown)


composer/utils/checkpoint.py
```diff
@@ -287,7 +287,9 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
             receiver = dist.get_global_rank() != rank_in_first_replica
 
             # Send list of files to all ranks
-            file_list = [sorted(os.listdir(self.destination_path))]
+            file_list = [
```
dakinggg, line 290: is metadata sent elsewhere?

mvpatel2000: Metadata is downloaded on all ranks and not broadcasted -- it's needed for other things (rng) before broadcast. I removed the line that filters it as well.

dakinggg (approved):

PR #3131, opened 2024-03-20
Explicitly print checkpoint downloading exception

bigning:
# What does this PR do?

PyTorch will hold any exception raised during this [read_data](https://github.com/mosaicml/composer/blob/ee13deadef1935b6054d1458f340971e5af4a635/composer/utils/checkpoint.py#L227), torch code pointer: https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/utils.py#L246-L255

PyTorch only raise the exception after communicate the exception across ranks https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/utils.py#L251, but issue is that communication iteself might fail because comms mis-match across ranks. In this case, it raised NCCL error, and hide the real read_data exceptioin. 

This PR print the exception immediately. It still raise NCCL error, but at least, it prints the checkpoint downloading error. 

# test
on 8-gpu checkpoint load test, i manually raise exception on rank 1 in the for-loop, i can reproduce the `c10::DistBackendError`. With this PR, i can see the logging of the exception.

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -275,6 +275,7 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
                         log.debug(f'Finished downloading {relative_file_path} to {file_destination}.')
             except Exception as e:
                 log.error(f'Exception raised during downloading: {str(e)}')
+                log.error(f'Exception raised during downloading: exception type: {type(e)}')
                 raise e
 
         # 3. Wait for all ranks to finish.
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -244,34 +244,38 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
             # Get the lowest rank in the current node
             local_rank_0 = dist.get_global_rank() - dist.get_local_rank()
 
-            for plan_item in plan.items:
-                relative_file_path = self.storage_data[plan_item.storage_index].relative_path
-                # Check if the file is scheduled to be downloaded by a lower rank on the same node
-                # i.e. if rank 0 and rank 1 on the same node have the same the same required file,
-                # only rank 0 should download it and not rank 1.
-                is_downloaded = any(
-                    relative_file_path in all_file_paths[i] for i in range(local_rank_0, dist.get_global_rank())
-                )
+            try:
+                for plan_item in plan.items:
+                    relative_file_path = self.storage_data[plan_item.storage_index].relative_path
+                    # Check if the file is scheduled to be downloaded by a lower rank on the same node
+                    # i.e. if rank 0 and rank 1 on the same node have the same the same required file,
+                    # only rank 0 should download it and not rank 1.
+                    is_downloaded = any(
+                        relative_file_path in all_file_paths[i] for i in range(local_rank_0, dist.get_global_rank())
+                    )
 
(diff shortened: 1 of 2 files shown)


j316chuck (commented):
@bigning can we add a unit test in `test_checkpoint.py` where download_file is mocked to raises an Exception? This will enforce that our try catch is working as expected consistently.

bigning:
> i'm not sure I understand -- shouldn't this rank already raise the error?

@mvpatel2000 , pytorch captured this error, and then communicate this error, but unfortunuately, it failed to communicate and raised NCCL error. SO it never prints any information about the real error.

This PR just capture the error, print the infromation, and raise it again. So program still crash with nccl error, but at least it prints the real error before crash.

eracah (commented):
Lgtm

composer/utils/checkpoint.py
```diff
@@ -244,34 +244,38 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+                            )
+                        else:
+                            self.object_store.download_file(
+                                remote_file_name=object_name,
+                                destination=file_destination,
+                            )
+                        log.debug(f'Finished downloading {relative_file_path} to {file_destination}.')
+            except Exception as e:
+                log.error(f'Exception raised during downloading: {str(e)}')
+                raise e
```
mvpatel2000, line 278: Can you add comment?

bigning:
> @bigning can we add a unit test in `test_checkpoint.py` where download_file is mocked to raises an Exception? This will enforce that our try catch is working as expected consistently.

@j316chuck do we really need unit test for this PR, it just add a logging. it doesn't change anything.  It's not trivial to set up a test for this. I have to basically write a mockObjectStore, mockDistCPObjectStoreReader, a mockPlan. Then inside mockDistCPObjectStoreReader, i have to create a comlex self.storage_data to work with the mockPlan to make them pass this https://github.com/mosaicml/composer/blob/dev/composer/utils/checkpoint.py#L236-L237. 
Even after all above set-up, it's very easy to break the unit test if we update the `read_data` implementation. 

Or do you have a better idea to test this logging ?

composer/utils/checkpoint.py
```diff
@@ -244,34 +244,42 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
+                            self.object_store.download_file(
+                                remote_file_name=object_name,
+                                destination=file_destination,
+                            )
+                        log.debug(f'Finished downloading {relative_file_path} to {file_destination}.')
+            except Exception as e:
+                # PyTorch will capture any exception of this funciton,
+                # and dist.all_gather_objects(exception) before raising it.
+                # If that all_gather_objects fails, the exception is never visible to user.
+                # We immediately print the exception to avoid that situation.
```
mvpatel2000, line 280:
```suggestion
                # PyTorch will capture any exception of this function,
                # and dist.all_gather_objects(exception) before raising it.
                # If that all_gather_objects fails, the exception is never visible to user.
                # We immediately print the exception to avoid that situation.
```

mvpatel2000 (approved):
LGTM, I'm fine with skipping test

j316chuck:
@bigning sure seems like this is too high lift, let's skip this for now. I think us moving towards Mocks will allow us to simulate ExceptionHandling better in the long run should be the way we go in the future.

PR #3135, opened 2024-03-21
Add ChunkedEncodingError exception handling for OCI

b-chu:
# What does this PR do?

OCI doesn't properly handle their exception `oci._vendor.requests.exceptions.ChunkedEncodingError`. This means that no retries happen following their default retry strategy. The issue is that they have two subclasses of the same base class, but attempt to compare the subclasses. This fix should be upstreamed to OCI, but this patch allows us to use retries while they work on the fix.

Tested with this script
```
from circuitbreaker import CircuitBreakerError
from oci._vendor.requests.exceptions import RequestException as BaseRequestException
from oci._vendor.requests.exceptions import Timeout
from oci._vendor.requests.exceptions import ConnectionError as RequestsConnectionError
from oci.exceptions import ServiceError, RequestException, ConnectTimeout
import threading


def should_retry(self, exception=None, response=None, **kwargs):
    if isinstance(exception, Timeout):
        return True
    elif isinstance(exception, RequestsConnectionError):
        return True
    elif isinstance(exception, RequestException):
        return True
    elif isinstance(exception, BaseRequestException):
        return True
    elif isinstance(exception, ConnectTimeout):
        return True
    elif isinstance(exception, CircuitBreakerError):
        if 'circuit_breaker_callback' in kwargs:
            threading.Thread(target=kwargs['circuit_breaker_callback'],
                             args=(exception, )).start()
        return True
    elif isinstance(exception, ServiceError):
        if exception.status in self.service_error_retry_config:
            codes = self.service_error_retry_config[exception.status]
            if not codes:
                return True
            else:
                return exception.code in codes
        elif self.retry_any_5xx and exception.status >= 500 and exception.status != 501:
            return True
    else:
        # This is inside a try block because ConnectionError exists in Python 3 and not in Python 2
        try:
            if isinstance(exception, ConnectionError):
                return True
        except NameError:
            pass

    return False


import oci

oci.retry.retry_checkers.TimeoutConnectionAndServiceErrorRetryChecker.should_retry = should_retry
retry_strategy = oci.retry.DEFAULT_RETRY_STRATEGY
print(
    retry_strategy.checkers.should_retry(
        exception=oci._vendor.requests.exceptions.ChunkedEncodingError()))

```

# Before submitting

### DIFF
--- a/composer/utils/object_store/oci_object_store_utils.py
+++ b/composer/utils/object_store/oci_object_store_utils.py
@@ -1,0 +1,81 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for monkey patching OCI."""
+
+import logging
+import threading
+
+from composer.utils.import_helpers import MissingConditionalImportError
+
+log = logging.getLogger(__name__)
+
+
+def patch_oci():
+    """Monkey patches the OCI retry strategy to add BaseRequestException.
+
+    This is necessary because the OCI retry strategy does not handle BaseRequestException.
+    """
+    try:
+        import oci
+    except ImportError as e:
+        raise MissingConditionalImportError(
+            conda_package='oci',
+            extra_deps_group='oci',
+            conda_channel='conda-forge',
+        ) from e
+    oci.retry.retry_checkers.TimeoutConnectionAndServiceErrorRetryChecker.should_retry = should_retry
+
+
+def should_retry(self, exception=None, response=None, **kwargs):
+    """Uses OCI retry and adds BaseRequestException.
+
+    https://github.com/oracle/oci-python-sdk/blob/8ba39d63242fd406fb7a97fb6f9acdde81eb6dd9/src/oci/retry/retry_checkers.py#L166
+    """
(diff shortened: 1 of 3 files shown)


mvpatel2000 (commented):
LGTM -- great catch. Will wait for Ning to approve

bigning:
great! let's wait for the e2e run

composer/utils/object_store/oci_object_store_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        from oci.exceptions import ConnectTimeout, RequestException, ServiceError
+    except ImportError as e:
+        raise MissingConditionalImportError(
+            conda_package='oci',
+            extra_deps_group='oci',
+            conda_channel='conda-forge',
+        ) from e
+    print(type(exception))
+    print(str(exception))
+    if isinstance(exception, Timeout):
```
Skylion007, line 30: You can also just group these into a `isinstance(exception, (Timeout....)` for all the valid exception types

b-chu: This is a copy paste from OCI's code so would prefer to keep it as close to the original as possible to get cleaner diffs

PR #3140, opened 2024-03-22
Composer object store download retry

bigning:
# What does this PR do?

add object_store checkpoint downloading retry

# test
bigning-debug-cp-loading-2-J6z0wB  **ConnectionError**
in the logging, there is one rank succeeded at 2nd attempt. 
image

another run: bigning-debug-cp-loading-2-9Lq8am **ChunkedEncodingError**
image

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -299,7 +299,6 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
                     file_name for file_name in sorted(os.listdir(self.destination_path)) if file_name.endswith('.distcp')
                 ]
                 dist.broadcast_object_list(file_list, src=rank_in_first_replica, group=replicate_process_group)
-                file_list = file_list[0]
                 log.debug(f'List of files to broadcast: {file_list}')
 
                 # Send each file to the appropriate rank
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -295,12 +295,13 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
                 receiver = dist.get_global_rank() != rank_in_first_replica
 
                 # Send list of files to all ranks
-                file_list = [
+                file_list = [[
                     file_name for file_name in sorted(os.listdir(self.destination_path)) if file_name.endswith('.distcp')
-                ]
+                ]]
                 dist.broadcast_object_list(file_list, src=rank_in_first_replica, group=replicate_process_group)
+                file_list = file_list[0]
                 log.debug(f'List of files to broadcast: {file_list}')
-
+                
                 # Send each file to the appropriate rank
                 for file_name in file_list:
                     if dist.get_local_rank() == 0:  # Only 1 rank per node needs to transfer file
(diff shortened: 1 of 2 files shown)


b-chu (changes requested):
Some failing tests but overall looks good. Let's clean up the unused import and pyright ignores. This would retry with every exception though, not just the ones that aren't retried from the object store. So you would 5x the number of retries that were already happening.

bigning:
>So you would 5x the number of retries that were already happening.

@b-chu , for the retries that already happening, if the oci internal retry works, then we don't do extra retry. Only if the internal retry fails, we do extra retry (i think that's kind of what we want. :) ).

composer/utils/checkpoint.py
```diff
@@ -188,6 +190,24 @@ def read_metadata(self) -> Metadata:
         return super().read_metadata()
 
 
+@retry(num_attempts=5)
+def download_object_or_file(
+    object_name: str,
+    file_destination: Union[str, pathlib.Path],
```
b-chu, line 196: we already import Path

composer/utils/checkpoint.py
```diff
@@ -188,6 +190,24 @@ def read_metadata(self) -> Metadata:
...
+):
+    if isinstance(object_store, ObjectStore):
+        object_store.download_object(
+            object_name=object_name,
+            filename=file_destination,
+        )
+    else:
+        object_store.download_file(
+            remote_file_name=object_name,
+            destination=cast(str, file_destination),
```
b-chu, line 207: does `str(file_destination)` work?

composer/utils/checkpoint.py
```diff
@@ -228,7 +248,8 @@ def __init__(
     def read_data(self, plan: LoadPlan, planner: LoadPlanner):
         # Download files if not using HSDP or if on first replica with HSDP enabled
         first_replica = self.device_mesh is None or self.device_mesh.ndim == 1 or (
-            self.device_mesh.ndim >= 2 and self.device_mesh.get_local_rank(mesh_dim=0) == 0
+            self.device_mesh.ndim >= 2 and
+            self.device_mesh.get_local_rank(mesh_dim=0) == 0  # pyright: ignore[reportGeneralTypeIssues]
```
b-chu, line 252: Upgrading your local env should fix the pyright issues

b-chu (commented):
Couple of minor comments

b-chu (approved):
Thanks for the cleanup!

composer/utils/checkpoint.py
```diff
@@ -188,6 +189,24 @@ def read_metadata(self) -> Metadata:
...
+    if isinstance(object_store, ObjectStore):
+        object_store.download_object(
+            object_name=object_name,
+            filename=file_destination,
+        )
+    else:
+        object_store.download_file(
+            remote_file_name=object_name,
+            destination=str(file_destination),
+        )
```
mvpatel2000, line 207: can we just add `download_object` as a wrapper fn to LoggerDestination and forward args to `download_file`? 

@eracah thoughts?

bigning: @mvpatel2000  I want to minimize this PR. Can we do that in a follow-up PR if we decide to do so ?

bigning: discussed offline with @mvpatel2000 , not a blocker for this PR.

PR #3118, opened 2024-03-15
More compression options

mbway:
# What does this PR do?

This PR adds the ability to save compressed `.pt` checkpoint files using CLI compression tools available on the system (eg `.pt.lz4` using the `lz4` command).

See the related issue (#3117) for why this is useful. In summary, good compression is achievable with low overhead, but the system must have the required tools available so the default setting to not use compression remains.

## Potential Improvements
- more control over arguments passed to compressors (eg number of threads or compression level)
    - technically a user can modify `KNOWN_COMPRESSORS` and replace compressors with their own `CliCompressor` objects (which gives them full control over the arguments) but this is a bit hacky.
- ability to use CLI compressors with `.tar`, eg `.tar.lz4`

# What issue(s) does this change relate to?

Fixes #3117

# Before submitting
- [X] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [ ] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [X] Did you update any related docs and document your change?
- [X] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [X] Did you run the tests locally to make sure they pass?
- [X] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

PR #3142, opened 2024-03-22
Compressor fixes

mbway:
# What does this PR do?

Fixes the issue encountered here https://github.com/mosaicml/composer/pull/3118#issuecomment-2014108923 where `lzma` is refusing to decompress a file in the docker container.

The version of lzma provided by the lzma package has a bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=700681
where it refuses to decompress if the file extension does not match, causing the test to fail. This could be worked around but the `xz-utils` package already provides a version of `lzma` that does not have this issue, so that can be used instead.

# Before submitting

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,9 +72,9 @@ Composer uses [pytest-codeblocks](https://github.com/nschloe/pytest-codeblocks) 
 
 To test your changes locally, run:
 
-1. `make test`  # run CPU tests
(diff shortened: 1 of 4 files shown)


mbway:
I think the reason for the test failure is that the docker image being used isn't updated until the PR is merged? I can implement the workaround into the test (use the proper file extension for the test file) and that should fix it

mvpatel2000 (approved):
LGTM! Thanks :)

PR #3143, opened 2024-03-24
Don't close the engine if its already closed

dakinggg:
# What does this PR do?
Currently produces duplicate print statements cause it may close both when you call `.close()` and `atexit`


mvpatel2000 (approved):

mvpatel2000 (approved):
How is this possible...?

Issue #3151, opened 2024-03-25
Crash during import when signal handler cannot be set


mbway:
# Problem

Since https://github.com/mosaicml/composer/pull/2125, composer is calling `signal.signal` during import.
This can crash in some situations which prevents composer from being used.

[The documentation](https://docs.python.org/3/library/signal.html#signal.signal) indicates that `signal.signal` may raise a `ValueError`.

The following example crashes with `ValueError: signal only works in main thread of the main interpreter`:
```python
import threading

def worker():
    from composer import Engine

if __name__ == '__main__':
    t = threading.Thread(target=worker)
    t.start()
    t.join()
```

And more importantly, there are some issues using composer with [ray](https://www.ray.io/) because it sets its own signal handler. Ray is using a class called [DeferSigint](https://github.com/ray-project/ray/blob/77ee044a7e3c4d583f94956d7334fb3d6d7ce014/python/ray/_private/utils.py#L1842) which causes a `ValueError` to be raised if `signal.signal` is called while the context manager is active. There is an issue open in the ray repo about this incompatibility with composer as well https://github.com/ray-project/ray/issues/42893 .

Running the following example in pytest crashes with `ValueError: Can't set signal handler for SIGINT while SIGINT is being deferred within a DeferSigint context.`
```python
import ray
from ray.train.data_parallel_trainer import DataParallelTrainer
from composer import Engine

def train():
    pass

def test_train():
    ray.init(address=None, num_cpus=2)
    trainer = DataParallelTrainer(train)
    trainer.fit()
```

## Expected behavior

I would expect to be able to import composer without crashing. My suggestion for an immediate fix/workaround is to put `try/catch` around the calls to `signal.signal` and continue with a warning if they fail. I'm not sure what the ideal fix would be to keep the checkpoint flushing functionality when used with ray if both projects want to hijack the signal handler.

PR #3152, opened 2024-03-25
Prevent crash if signal handler cannot be set

mbway:
# What does this PR do?

This is a workaround for https://github.com/mosaicml/composer/issues/3151 that at least prevents the import from crashing. A better long term solution may be possible.

# Before submitting
- [] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.

### DIFF
--- a/composer/core/engine.py
+++ b/composer/core/engine.py
@@ -123,8 +123,11 @@ def sigterm_handler(signal, frame):
     sys.exit(128 + signal)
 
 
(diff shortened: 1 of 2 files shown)


mvpatel2000 (approved):
Thanks for the detailed writeup and fix! We're planning on potentially revamping this feature (auto checkpointing on SIGKILL), so it's great to know this breaks with ray / if not main process.

PR #3155, opened 2024-03-29
Fix torch monkeypatch version check

dakinggg:


composer/trainer/mosaic_fsdp.py
```diff
@@ -61,7 +61,7 @@ def patch_pytorch():
         from torch.distributed.fsdp import _runtime_utils
         _runtime_utils._validate_and_get_hybrid_shard_state = lambda *args, **kwargs: None
 
-    elif version.parse(torch.__version__) < version.parse('2.2.2'):
+    elif version.parse(torch.__version__) < version.parse('2.3.0'):
         # Monkey patch for torch < 2.2.2 ie torch == 2.2.1
         pass
```
j316chuck, line 66: I don't think there should be a pass here just from pattern matching. I think it should be: 

```
    elif version.parse(torch.__version__) < version.parse('2.3.0'):
        # Monkey patch for torch < 2.2. ie torch == 2.2.1
        # Allow 2D HSDP
        from torch.distributed.fsdp import _runtime_utils
        _runtime_utils._validate_and_get_hybrid_shard_state = lambda *args, **kwargs: None
    elif version.parse(torch.__version__) < version.parse('2.3.1'): ...
```

j316chuck (approved):

PR #3164, opened 2024-04-02
add more retry to checkpoint downloading

bigning:
# what
When run diagnostic on r15z1, we found OCI download error in other two places. Adding the retry to those object store downloading. 

# test
bigning-debug-cp-loading-2-Cv5Bsh


mvpatel2000 (approved):
LGTM!

PR #3166, opened 2024-04-02
Append to gpu rank log files instead of throwing error

jjanezhang:
# What does this PR do?
For runs that call composer commands twice, we would error on creating a new log file because an existing log file had already been created on the first composer command. This changes the behavior to append to the existing or write a new file if doesn't exist already.

# Testing 
Ran a yaml that calls composer twice, verified that logs append
image


mvpatel2000 (approved):
Legend

PR #3167, opened 2024-04-02
[Not for merge] refactor the object_store.download_object retry and add more retry

bigning:
# What
1. searched the repo, replace ALL object_store.download_object() with the retry version to avoid potential downloading error in future.
2. moved the `download_object_or_file` function to `file_helpers.py` to make 1 easy.

# Test
bigning-debug-cp-loading-2-ntUAon
bigning-debug-cp-loading-2-rhJJm0

### DIFF
--- a/composer/utils/file_helpers.py
+++ b/composer/utils/file_helpers.py
@@ -62,15 +62,11 @@ def download_object_or_file(
     progress_bar: bool = False,
 ):
     if isinstance(object_store, ObjectStore):
-        if progress_bar:
-            callback = _get_callback(f'Downloading {object_name}')
-        else:
-            callback = None
         object_store.download_object(
             object_name=object_name,
             filename=file_destination,
             overwrite=overwrite,
-            callback=callback,
+            callback=_get_callback(f'Downloading {object_name}') if progress_bar else None,
         )
     elif isinstance(object_store, LoggerDestination):
         object_store.download_file(
@@ -80,7 +76,7 @@ def download_object_or_file(
             progress_bar=progress_bar,
         )
     else:
-        raise RuntimeError(f'Unrecognized type: {type(object_store)}')
+        raise RuntimeError(f'Unrecognized type: {type(object_store)} when attempting to download object')
 
 
 def extract_path_from_symlink(
--- a/composer/utils/file_helpers.py
+++ b/composer/utils/file_helpers.py
@@ -64,7 +64,7 @@ def download_object_or_file(
     log.debug(f'bigning debug using new code path')
     if isinstance(object_store, ObjectStore):
         if progress_bar:
-            callback = _get_callback(f'Downloading {object_name}') if progress_bar else None
+            callback = _get_callback(f'Downloading {object_name}')
         else:
             callback = None
         object_store.download_object(
--- a/composer/utils/file_helpers.py
+++ b/composer/utils/file_helpers.py
@@ -61,6 +61,7 @@ def download_object_or_file(
     overwrite: bool = False,
     progress_bar: bool = False,
 ):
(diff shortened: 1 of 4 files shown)


composer/utils/file_helpers.py
```diff
@@ -52,6 +53,36 @@
...
+    if isinstance(object_store, ObjectStore):
+        if progress_bar:
+            callback = _get_callback(f'Downloading {object_name}')
+        else:
+            callback = None
+        object_store.download_object(
+            object_name=object_name,
+            filename=file_destination,
+            overwrite=overwrite,
+            callback=callback,
```
mvpatel2000, line 73:
```suggestion
        object_store.download_object(
            object_name=object_name,
            filename=file_destination,
            overwrite=overwrite,
            callback=_get_callback(f'Downloading {object_name}') if progress_bar else None,
```
minor in-lining

composer/utils/file_helpers.py
```diff
@@ -52,6 +53,36 @@
...
+        )
+    elif isinstance(object_store, LoggerDestination):
+        object_store.download_file(
+            remote_file_name=object_name,
+            destination=str(file_destination),
+            overwrite=overwrite,
+            progress_bar=progress_bar,
+        )
+    else:
+        raise RuntimeError(f'Unrecognized type: {type(object_store)}')
```
b-chu, line 83: This will cause retries which is probably not intended behavior

mvpatel2000, line 83:
```suggestion
    else:
        raise RuntimeError(f'Unrecognized type: {type(object_store)} when attempting to download object')
```
nit: a bit more user friendly error?

mvpatel2000 (approved):
LGTM

b-chu (commented):
Thanks for the PR! Only one minor comment in the PR. However, if `get_file` calls `download_object_or_file`, should the rest of the codebase use `get_file` instead of `download_object_or_file`?

bigning:
> Thanks for the PR! Only one minor comment in the PR. However, if `get_file` calls `download_object_or_file`, should the rest of the codebase use `get_file` instead of `download_object_or_file`?

@b-chu , good question! I don't know the background why we were not using get_file() in the beginning. It looks to me `get_file()` will [change the destination_path if it's `symlink` file ](https://github.com/mosaicml/composer/pull/3167/files#diff-7c7f8af4988e41fec242e985aa1e5d894017158082c71a588e393f9a1e97ba5cL577-L592). I guess it's not always preferred ?

b-chu (approved):
Makes sense, we can revisit consolidating functions if it causes a problem in the future

composer/utils/file_helpers.py
bigning: Yes, it will cause retry and raise the RuntimeError after retry. 
If i don't add this, there are some other options:
1. just print warning, it won't retry, but also won't raise any error here, the downstream probably would fails and raise error. 
2. handle this outside of this function. But it needs to add this raise Error to every place that calls this function. 

so I think raising error here might be better, only downside is retry, which should be ok.

bigning:
i will not merge this PR for now, because i just find it regresses the auto-resume. With auto-resume, it will try to `get_file(checkpoint_path)` no matter if checkpoint_path exits or not ([code pointer](https://github.com/mosaicml/composer/blob/688f9d278d75527582f0448dc6d2b85a24ca5275/composer/trainer/trainer.py#L1770-L1784)).  In that case, we should only retry if it's not `FileNotFoundError`.   So we need a retry API like `retry_unless(exceptions)` to handle that.  I can revisit this later after I finshed my other high-pri tasks (SHARD_GRAD_OP failure).  cc @mvpatel2000  @b-chu

Issue #3181, opened 2024-04-05
Safer Checkpoint File Format


mbway:
## 🚀 Feature Request
I have been investigating if it is possible to store checkpoints in a format that does not depend on pickle, as pickles are inherently unsafe as they allow for arbitrary code execution. The `.pt` file format created by `torch.save` uses pickle internally.

Would this functionality be something that you would be interested in? If so I can post a PR with my implementation for consideration.

## Implementation

My solution is a custom `.safe` format that is a `.tar` archive containing:
```
my_checkpoint.safe
├── tensors.safetensors
├── ndarrays.npz
└── other.json
```

To save a state dictionary, I pre-process it so anything not json-serializable gets converted to a dictionary like `{"__safe_obj_type": "tensor", "id": "123"}` for example. The `__safe_obj_type` key is then used during loading to restore the object back to what it was before. The gathered tensors and numpy arrays are stored as `safetensors` and `npz` respectively and anything else is stored in the json file.

At least for the state-dict I was trying to serialize I found that only a few non-json-serializable data types needed to be supported: `torch.tensor`, `np.ndarray`, `datetime.timedelta`, `tuple` Though a few others may be needed if this was implemented beyond a proof of concept.

mvpatel2000:
@mbway Thanks for flagging this -- we're currently actively working on a major redesign of checkpointing that will include migrating to a safer format. We've been hesitant to change how checkpointing works as we deeply care about backwards compatibility, so we have been buffering up a long list of features we will now include in the revamp. You should start seeing things roll out this month.

CC: @eracah it might be good to publicly share a version of the roadmap as a RFC when its ready 

If you already have a PR, please feel free to post it -- it's probably helpful for us to view a design and gain some inspiration. Given the refactor though, we likely wouldn't accept it as this seems like a big change.

PR #3186, opened 2024-04-11
draft: add prototype for safer checkpoint format

mbway:
This is a proof of concept for using a 'safer' checkpoint format in composer, mentioned in https://github.com/mosaicml/composer/issues/3181.
This PR is intended for inspiration and is not expected to be merged.

`pickle` and therefore `.pt` and composer's `.tar` checkpoint formats all have the possibility of executing arbitrary code during loading which is undesirable. The approach taken here to avoid this is to use 'inert' formats like `.npy`, `.safetensors` and `json` to store data.

This implementation patches the composer saving and loading functionality to use the custom format if the checkpoint is given the file extension: `.safe`.

When saving, `state_dict` is traversed and anything not json serializable is extracted and replaced with some state that can be identified and restored during loading such as `{"__safe_storage_obj": "timedelta", "seconds": 123}` for a `timedelta` object, or `{"__safe_storage_obj": "tensor", "id": "foo"}` for a tensor (where the tensor data is stored separately).

This proof of concept has lots of room for improvement:

- tensors could be converted to numpy arrays so there are only two components to a safe checkpoint, 'arrays' and 'other' instead of 'tensors', 'arrays', and 'other. This would eliminate the safetensors dependency. Many of the safetensors benefits such as memory mapping are not utilised since the file is put inside an archive. Note that the inverse would not work: Not all numpy arrays can be converted to tensors because torch doesn't support all dtypes.
- json could be substituted with msgpack or similar
- the implementation is not very efficient as the state dict is copied to replace the non-serializable parts and extra disk space is required as the files have to first be saved, then zipped afterwards.
- in this implementation the supported data types are hard-coded, but a production quality implementation may support more types out of the box and allow the user to register more.

# Before submitting
- [-] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.

### DIFF
--- a/composer/utils/safe_checkpoint.py
+++ b/composer/utils/safe_checkpoint.py
@@ -1,0 +1,211 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""A custom safer checkpoint format that does not allow for arbitrary code execution (unlike pickle)."""
+
+import json
+import shutil
+import tempfile
+import zipfile
+from datetime import timedelta
+from pathlib import Path
+from types import NoneType
+from typing import Any
+
+import numpy as np
+import numpy.typing as npt
+import safetensors.torch
+import torch
+
+import composer.utils.checkpoint
+from composer.loggers import LoggerDestination
+from composer.utils.object_store import ObjectStore
+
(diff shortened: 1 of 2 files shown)


mvpatel2000:
CC: @eracah @dakinggg

PR #3194, opened 2024-04-15
Fix MLFlow Tag Name for Resumption

KuuCi:
https://databricks.atlassian.net/browse/GRT-2836

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -148,7 +148,7 @@ class MLFlowLogger(LoggerDestination):
 
         # Store the Composer run name in the MLFlow run tags so it can be retrieved for autoresume.
         self.tags = self.tags or {}
-        self.tags['run_name'] = state.run_name
+        self.tags['run_name'] = os.environ['RUN_NAME'] if 'RUN_NAME' in os.environ else state.run_name
 
         # Adjust name and group based on `rank_zero_only`.
         if not self._rank_zero_only:
@@ -170,16 +170,6 @@ class MLFlowLogger(LoggerDestination):
                     filter_string=f'tags.run_name = "{state.run_name}"',
                     output_format='list',
                 )
-
-                # Check for the old tag (`composer_run_name`) For backwards compatibility in case a run using the old
-                # tag fails and the run is resumed with a newer version of Composer that uses `run_name` instead of
-                # `composer_run_name`.
-                if len(existing_runs) == 0:
-                    existing_runs = mlflow.search_runs(
-                        experiment_ids=[self._experiment_id],
-                        filter_string=f'tags.composer_run_name = "{state.run_name}"',
-                        output_format='list',
-                    )
 
                 if len(existing_runs) > 0:
                     self._run_id = existing_runs[0].info.run_id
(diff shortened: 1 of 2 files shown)


composer/loggers/mlflow_logger.py
```diff
@@ -171,16 +171,6 @@ def init(self, state: State, logger: Logger) -> None:
                     output_format='list',
                 )
 
-                # Check for the old tag (`composer_run_name`) For backwards compatibility in case a run using the old
```
dakinggg, line 174: wait, we're not turning this off in this PR, right? @jerrychen109 that would just break resumption unless I'm misunderstanding something.

jerrychen109: I think we had talked about removing this shortly after we made the original change to rename the MLflow tag from `composer_run_name` to `run_name`. This was leftover from the original PR that we kept just to avoid breaking resumption for ongoing runs. There should be no runs started that are tagged with `composer_run_name` for the past 2 months, so unless there are 2-month-old runs that are ongoing and need to auto-resume with an updated composer version/commit, this should be safe? I'll leave it to your discretion though depending on how strictly we need to maintain backwards compatibility here, but I would be very very surprised if this breaks an actual user worklfow.

dakinggg: ahh sorry. i misread what was being removed. thanks for clarifying, fine to delete

jerrychen109 (commented):
Can you also update `test_mlflow_logger.py` to unit test the behavior when the env var is set differently from the `state.run_name`? Would also be good to verify the fallback behavior if the env var isn't set. https://github.com/mosaicml/composer/blob/83cb5c14a89fc5dcb69db0b92131e61e6ff4ae8a/tests/loggers/test_mlflow_logger.py#L259-L260

tests/loggers/test_mlflow_logger.py
```diff
@@ -190,6 +190,37 @@ def test_mlflow_experiment_init_existing_composer_run(monkeypatch):
...
+
+    mock_state = MagicMock()
+    mock_state.run_name = 'state-run-name'
+
+    logger = MLFlowLogger()
+    logger.init(state=mock_state, logger=MagicMock())
+    assert logger.tags['run_name'] == 'state-run-name', "Logger should fallback to the state's run name when no environment variable is set."
+
+
 def test_mlflow_experiment_init_existing_composer_run_with_old_tag(monkeypatch):
```
jerrychen109, line 224: Can you also delete this test?

jerrychen109 (approved):
lgtm once you get the code quality/linting to pass - thanks for adding the tests and handling the mocking! (but you'll need to get @dakinggg's approval as code owner)

composer/loggers/mlflow_logger.py
```diff
@@ -148,7 +148,7 @@ def init(self, state: State, logger: Logger) -> None:
 
         # Store the Composer run name in the MLFlow run tags so it can be retrieved for autoresume.
         self.tags = self.tags or {}
-        self.tags['run_name'] = state.run_name
+        self.tags['run_name'] = os.environ['RUN_NAME'] if 'RUN_NAME' in os.environ else state.run_name
```
dakinggg, line 151:
```suggestion
        self.tags['run_name'] = os.environ.get('RUN_NAME', state.run_name)
```

dakinggg (approved):

PR #3201, opened 2024-04-20
Load optimizer separately for all current torch versions

snarayan21:
# What does this PR do?

Addresses a bug when loading optimizer state on resumption with torch >= 2.3.0, with `use_orig_params: False` and activation checkpointing. We currently do not load optimizer state separately when loading checkpoint for torch >=2.3.0, but according to torch's recommendation (see obscure comment [here](https://github.com/pytorch/pytorch/blob/f9fce110af428a600e597291a0ab80e43dc39e93/torch/distributed/checkpoint/optimizer.py#L215)), we should continue to load the optimizer state separately. This PR does that.

Filed an issue with PyTorch [here](https://github.com/pytorch/pytorch/issues/124546).

See Jira issue [here](https://databricks.atlassian.net/browse/GRT-2843).

# Before submitting

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -625,8 +625,10 @@ def load_sharded_checkpoint(
                 state_dict: Dict[str, Any] = {'state': {'model': state.get_model_state_dict()}}
             else:
                 cur_state_dict = state.state_dict()
-                # For older versions of torch, we load optimizer separately.
-                if version.parse(torch.__version__) < version.parse('2.2.9'):
+                # For all currently in-use versions of torch, we load optimizer separately.
+                # Torch recommends loading optimizer separately, see:
+                # See https://github.com/pytorch/pytorch/blob/f9fce110af428a600e597291a0ab80e43dc39e93/torch/distributed/checkpoint/optimizer.py#L215
+                if version.parse(torch.__version__) < version.parse('2.5.0'):
                     cur_state_dict.pop('optimizers')
                 num_rng_ranks = _get_num_ranks_that_saved_rng(storage_reader.read_metadata())
                 state_dict: Dict[str, Any] = {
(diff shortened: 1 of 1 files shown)


mvpatel2000 (commented):
@eracah is there any performance impact?

mvpatel2000 (commented):
Can we please remove the torch gating checks instead and leave as is (separate loading)?

composer/utils/checkpoint.py
```diff
@@ -1067,13 +1064,12 @@ def _save_checkpoint(
...
-        # To load optimizer states with 2.0 <= torch < 2.2.9 , the optimizer state must be at the top
-        # level of the state dict because the load_sharded_optimizer_state_dict function
+        # To load optimizer states with current torch versions, the optimizer state must be at the
+        # top level of the state dict because the load_sharded_optimizer_state_dict function
         # requires a top level state dict key for the optimizer.
         # See https://github.com/pytorch/pytorch/blob/v2.0.1/torch/distributed/checkpoint/optimizer.py#L271
         # for more info.
-        if version.parse(torch.__version__) < version.parse('2.2.9'):
-            state_dict['optimizers'] = state_dict['state'].pop('optimizers')
+        state_dict['optimizers'] = state_dict['state'].pop('optimizers')
```
mvpatel2000, line 1072: Do we need this change where we change state dict? I really dont like optimizers at root level.

eracah: wait what those comments are from 2 years ago

eracah: what makes you think they still hold?

mvpatel2000: Can we try it without?

snarayan21: this bug makes me think it's still valid at least lol. I'll test

snarayan21:
@mvpatel2000 after discussion offline with @eracah, this probably isn't the perferred fix here. It seems that loading the optimizer separately is fixing this issue sort of "on accident" and the real issue is why activation checkpointing is messing up the FQNs. I'm going to close this issue and try to find where the FQNs are getting messed up instead.

PR #3209, opened 2024-04-24
Add torch 2.3 support

mvpatel2000:
# What does this PR do?

- [x] Adds torch 2.3 support
- [x] Removes torch 2.0.1 support (we support 3 rolling versions)
- [x] Updates latest to torch 2.2.1 
- [ ] Test 2.3
- [x] Remove 2.2.9 gating


composer/algorithms/utils/augmentation_common.py
```diff
@@ -43,7 +43,7 @@ def image_as_type(image: _InputImgT, typ: Type[_OutputImgT]) -> _OutputImgT:
         raise TypeError(f'Only typ={{torch.Tensor, Image}} is supported; got {typ}')
 
     if typ is torch.Tensor:
-        return cast(_OutputImgT, torchvision.transforms.functional.to_tensor(image))  # PIL -> Tensor
+        return cast(_OutputImgT, torchvision.transforms.functional.to_tensor(image))  # type: ignore PIL -> Tensor
```
dakinggg, line 46: whats the new type error?

setup.py
```diff
@@ -80,8 +80,8 @@ def package_files(prefix: str, directory: str, extension: str):
     'tqdm>=4.62.3,<5',
     'torchmetrics>=0.10.0,<1.3.3',
     'torch_optimizer>=0.3.0,<0.4',
-    'torchvision>=0.13.1,<0.20',  # TODO: Tighten before release
-    'torch>=2.0.1,<2.3.1',  # TODO: Tighten before release
+    'torchvision>=0.13.1,<0.18.1',
+    'torch>=2.1.2,<2.3.1',
```
dakinggg, line 84: This should be `<2.4` now right?

dakinggg (approved):

composer/algorithms/utils/augmentation_common.py
mvpatel2000: Before they had type Any, now it is actually typed. It's something annoying of how it accepts any nd array (eg numpy) but we might pass in tensors because the input type is more generic. It's just annoyingness with how we make the input type more generic... and it's not worth patching this part of codebase

setup.py
mvpatel2000: Now, we explicitly upgrade torch always

PR #3214, opened 2024-04-25
Use FA's CrossEntropyLoss for metrics calculation

snarayan21:
# What does this PR do?

If available, we should use FA's CrossEntropyLoss for calculating the CE metric in composer. We already use it in foundry, and it's simply faster. Below we can see that the CE loss metric is the same, and there's a nice MFU boost as well.

Pink: using FA's CrossEntropyLoss
Green: using torch's CrossEntropyLoss
Screenshot 2024-04-25 at 4 47 49 PM
Screenshot 2024-04-25 at 4 48 41 PM

# Before submitting

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 # build requirements
 [build-system]
-requires = ["setuptools < 68.0.0"]
+requires = ["setuptools < 68.0.0", "packaging >= 21.3.0, < 24.1"]
 build-backend = "setuptools.build_meta"
 
 # iSort
--- a/setup.py
+++ b/setup.py
@@ -225,7 +225,7 @@ extra_deps['mlflow'] = [
     'mlflow>=2.11.1,<3.0',
 ]
 
-extra_deps['gpu-flash2'] = [
+extra_deps['flash2'] = [
     'flash-attn==2.5.8',
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -225,6 +225,10 @@ extra_deps['mlflow'] = [
     'mlflow>=2.11.1,<3.0',
 ]
 
+extra_deps['gpu-flash2'] = [
+    'flash-attn==2.5.0',
+]
+
 extra_deps['pandas'] = ['pandas>=2.0.0,<3.0']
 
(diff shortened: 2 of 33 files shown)


dakinggg:
https://github.com/mosaicml/composer/pull/2987 this is the third time we've had this PR :joy: once in foundry, once in composer, and now again. @mvpatel2000 last time I thought we should put it in foundry. Do you want to take FA as a Composer dependency?

mvpatel2000 (commented):
Yea Im fine adding it. @snarayan21  please add as an optional dependency

snarayan21:
added as optional dep

b-chu (approved):
Thanks for the PR!

setup.py
```diff
@@ -225,6 +225,10 @@ def package_files(prefix: str, directory: str, extension: str):
     'mlflow>=2.11.1,<3.0',
 ]
 
+extra_deps['gpu-flash2'] = [
+    'flash-attn==2.5.0',
+]
```
mvpatel2000, line 230: why `gpu-` instead of jsut `flash2`?

dakinggg: should just be `flash-attention`

snarayan21: wbout `flash-attn` since its the package name

snarayan21: changed to `flash-attn`

dakinggg: ya sounds good

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 # build requirements
 [build-system]
-requires = ["setuptools < 68.0.0"]
+requires = ["setuptools < 68.0.0", "packaging >= 21.3.0, < 24.1"]
```
mvpatel2000, line 3: why this?

snarayan21: gets around some of the installation errors, but torch is also a dependency, and then we go round and round in this circle trying to get build isolation working correctly. so i'm closing this PR, it should live foundry side.

snarayan21:
Closed this PR since adding `flash-attn` to composer as a dependency makes things very ugly. Will add this to foundry later. 4th time's the charm!

PR #3208, opened 2024-04-24
Create new Mlflow run by default and introduce `run_group`

chenmoneygithub:
# What does this PR do?

Create new Mlflow run by default and introduce `run_group`. Currently the default behavior is to resume the Mlflow run by searching for existing run with the same name. As we discussed we should make the default behavior start a new run. Additionally to mirror the `group` in wandb, we are introducing the `run_group` tag.

# What issue(s) does this change relate to?

# Before submitting
- [x] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [ ] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [ ] Did you update any related docs and document your change?
- [ ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [ ] Did you run the tests locally to make sure they pass?
- [ ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

PR #3215, opened 2024-04-26
Search for proper run_name and not state name

KuuCi:
Make sure getting existing runs pulls from os.environ as well as state run name

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -138,6 +138,7 @@ class MLFlowLogger(LoggerDestination):
                     self._experiment_id = exp_from_name.experiment_id
                 else:
                     self._experiment_id = (self._mlflow_client.create_experiment(name=self.experiment_name))
+            self.run_url = self.get_run_url()
 
     def init(self, state: State, logger: Logger) -> None:
         import mlflow
@@ -165,7 +166,7 @@ class MLFlowLogger(LoggerDestination):
             else:
                 # Search for an existing run tagged with this Composer run.
                 assert self._experiment_id is not None
-                run_name = os.environ.get('RUN_NAME', state.run_name)
+                run_name = self.tags['run_name']
                 existing_runs = mlflow.search_runs(
                     experiment_ids=[self._experiment_id],
                     filter_string=f'tags.run_name = "{run_name}"',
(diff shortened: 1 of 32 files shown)


mvpatel2000 (commented):
Please add PR descriptions! They are your friend

Please wait for https://github.com/mosaicml/composer/pull/3208 to merge first

While ur at it, can u get the mosaicml logger to log the mlflow URL plz

composer/loggers/mlflow_logger.py
```diff
@@ -165,9 +165,10 @@ def init(self, state: State, logger: Logger) -> None:
             else:
                 # Search for an existing run tagged with this Composer run.
                 assert self._experiment_id is not None
+                run_name = os.environ.get('RUN_NAME', state.run_name)
```
dakinggg, line 168: should we just reuse `self.tags['run_name']` so the logic always matches?

composer/loggers/mlflow_logger.py
```diff
@@ -184,6 +185,7 @@ def init(self, state: State, logger: Logger) -> None:
                 tags=self.tags,
                 log_system_metrics=self.log_system_metrics,
             )
+            self.run_url = f'{self.tracking_uri}/#/experiments/{self._experiment_id}/runs/{self._run_id}'
```
dakinggg, line 188: Can you check if mlflow has an API for this? Also please include a manual test run so I can look at the link and make sure it works.

KuuCi: No API replacements to be found :(

mcli logs test-mlflow-logger-pK4ZNT
image

composer/loggers/mosaicml_logger.py
```diff
@@ -113,6 +114,10 @@ def after_load(self, state: State, logger: Logger) -> None:
                     log.debug(f'Logging WandB run URL to metadata: {run_url}')
                 else:
                     log.debug('WandB run URL not found, not logging to metadata')
+            if isinstance(callback, MLFlowLogger) and callback._enabled:
+                run_url = f'{os.environ["DATABRICKS_HOST"]}ml/experiments/{callback._experiment_id}/runs/{callback._run_id}'
```
dakinggg, line 118: shouldn't this be an attr on the logger instead of defined here? Also, please use `posixpath.join` to get reliable joining behavior.

KuuCi:
test-mlflow-logger-EP7BFB

dakinggg (approved):

mvpatel2000 (approved):
LGTM, ty!

PR #3216, opened 2024-04-26
Add a torch.cuda.empty_cache() in utils.save_checkpoint

bfontain:
# What does this PR do?

Adds a empty_cache() to clear the cache right before checkpointing. Some amount of memory is used by NCCL during checkpointing and if we are operating near the total memory of the GPU, empty_cache() can help clear unused memory and allow checkpointing to succeed.

# Before submitting

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -14,6 +14,8 @@ import textwrap
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Union
 
+import torch.cuda
+
 from composer.core import Callback, Event, State, Time, Timestamp
 from composer.loggers import Logger, MLFlowLogger
 from composer.utils import (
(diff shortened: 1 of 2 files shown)


composer/callbacks/checkpoint_saver.py
```diff
@@ -14,6 +14,8 @@
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Union
 
+import torch.cuda
```
mvpatel2000, line 17: can we just do `import torch` to be simple like elsewhere?

mvpatel2000 (commented):
Can we do this in `save_checkpoint` in `checkpoint.py` instead? That way if a user directly calls API, it will also run

composer/callbacks/checkpoint_saver.py
bfontain: No longer needed now that the call has been moved to checkpoint.py

bfontain:
> Can we do this in `save_checkpoint` in `checkpoint.py` instead? That way if a user directly calls API, it will also run

Done!

mvpatel2000 (approved):
Thanks! Might take a bit to merge, CI/CD is broken. I will force merge when tests pass and nothing unusual is failing, please tag me.

dakinggg (approved):

PR #3219, opened 2024-04-28
load_checkpoint defaults to strict weight match

mvpatel2000:
# What does this PR do?

Trainer has an argument `load_strict_model_weights` which defaults to `True`. However, the same argument in `load_checkpoint` defaults to `False`, whereas it should be consistent with the Trainer argument.


dakinggg (approved):

b-chu:
I'm worried this breaks BC

mvpatel2000:
> I'm worried this breaks BC

It is an API behavior change, but I view it as a fix which I think is sufficient. See linked github issue for motivation

b-chu (approved):

eracah (approved):
lgtm

PR #3249, opened 2024-05-02
Fix HF tests

mvpatel2000:
# What does this PR do?

Fix HF tests. bert-base-uncased was moved under google-bert


milocress (approved):

mvpatel2000:
HF is borked, just going to pin it

PR #3250, opened 2024-05-03
[ckpt-rewr] Get Model State Dict Util Function

eracah:
# What does this PR do?
Adds an API for extracting model state dict from a model object.

State dict generation is a necessary operation before the save AND load of a checkpoint.
 Currently in composer it is coupled with the State, and not very readable, hard to extend, hard to test, and hard for users to harness to do custom things. As such, we present a function to generate state_dict for the model decoupled from State as a standalone function. By making an explicit function for the model, it’s easier to test because we have a standalone function (we don’t have to make a dummy State function). Moreover, it’s easier to save each state dict as a separate file  Also, an advanced user can just call these functions themselves if they have a custom, advanced script or callback. 

This state dict generation function enables:
* generating sharded or full state dicts
* generating state dicts of different precision
* specify keys to include
* sprecify keys to exclude

These are all options that will be useful for save and load. Because save and load require state dict generation, we need these options in state dict generation as well


mvpatel2000 (commented):
Is this just pulling out existing code into a helper fn? 

@eracah it would be great to get slightly more description so I know what parts to carefully read over and what is less important

eracah:
> Is this just pulling out existing code into a helper fn?
> 
> @eracah it would be great to get slightly more description so I know what parts to carefully read over and what is less important

It's detailed in the design doc, but I can copy and paste it in if you want

mvpatel2000:
> > Is this just pulling out existing code into a helper fn?
> > @eracah it would be great to get slightly more description so I know what parts to carefully read over and what is less important
> 
> It's detailed in the design doc, but I can copy and paste it in if you want

It's easier when reviewing PRs to either link to the right part of design doc or copy paste description

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,157 @@
...
+
+from composer.models import ComposerModel
+from composer.utils import dist
+
+log = logging.getLogger(__name__)
+
+
+def get_model_state_dict(
+    model: Union[ComposerModel, nn.Module],
+    sharded: bool,
```
bigning, line 24: what if the model is not fsdp but `sharded` is True?  A design question, what's the usecase that the users want to call this function without constructing the trainer object? Or will this API mainly used inside trainer ? I think currently it's not trivial to create a FSDP module outside of the trainer https://github.com/mosaicml/composer/blob/01eec3ad1732074c52cf5da35ef5b8a24531279a/composer/trainer/dist_strategy.py#L212.

eracah: > what if the model is not fsdp but `sharded` is True? 
It will still return non-sharded. I will add an error

>A design question, what's the usecase that the users want to call this function without constructing the trainer object? Or will this API mainly used inside trainer ? I think currently it's not trivial to create a FSDP module outside of the trainer
This will mostly be used internally, but a user can call `get_model_state_dict(trainer.state.model)`. It is still possible to create an FSDP module outside trainer and we will have advanced users that will do it. See the unit tests for how simple it is to make an FSDP module. Also, once fsdp2 is popularized it will be even easier. Also, I think for the load API, we may aim to add some helper functions to materialize the model

eracah:
> > > Is this just pulling out existing code into a helper fn?
> > > @eracah it would be great to get slightly more description so I know what parts to carefully read over and what is less important
> > 
> > 
> > It's detailed in the design doc, but I can copy and paste it in if you want
> 
> It's easier when reviewing PRs to either link to the right part of design doc or copy paste description

Ok added description

tests/checkpoint/test_state_dict.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+from torch.distributed.device_mesh import init_device_mesh
```
mvpatel2000, line 6: this migh tneed to be gated by torch version

tests/checkpoint/test_state_dict.py
```diff
@@ -0,0 +1,202 @@
...
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+
+from composer.checkpoint import get_model_state_dict
```
mvpatel2000, line 9: this migh tneed to be gated by torch version

tests/common/models.py
```diff
@@ -113,6 +113,29 @@ def forward(self, x):
...
+
+
+class SimpleComposerMLP(ComposerClassifier):
+
+    def __init__(self, num_features: int, device: str, num_classes: int = 3):
+        fc1 = torch.nn.Linear(num_features, num_features, device=device, bias=False)
+        fc2 = torch.nn.Linear(num_features, num_features, device=device, bias=False)
+
+        net = torch.nn.Sequential(fc1, torch.nn.ReLU(), fc2)
+        super().__init__(num_classes=num_classes, module=net)
```
mvpatel2000, line 136: why do we need two if both are basically the same? 

Also why do we need this over SimpleMLP?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+
+    log.debug('Finished extracting model state dict')
+    return model_state_dict
+
+
+STR_TO_DTYPE = {
+    'fp32': torch.float32,
+    'fp16': torch.float16,
+    'bf16': torch.bfloat16,
+}
```
mvpatel2000, line 94: nit: can we move to top above fns?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+    if not is_fsdp and sharded:
+        raise ValueError('Sharded state dict can only be generated for FSDP models.')
+    cpu_offload = cpu_offload if cpu_offload is not None else is_fsdp
+
+    log.debug('Extracting model state dict')
+    if version.parse(torch.__version__) >= version.parse('2.3.0') and dist.is_initialized():
+        from torch.distributed.checkpoint.state_dict import StateDictOptions
+        from torch.distributed.checkpoint.state_dict import get_model_state_dict as dcp_get_model_state_dict
+        
+        get_nonsharded_state_dict = not sharded
```
mvpatel2000, line 57:
```suggestion
```
unused?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+
+    is_fsdp = _is_model_fsdp(model)
+    if not is_fsdp and sharded:
+        raise ValueError('Sharded state dict can only be generated for FSDP models.')
+    cpu_offload = cpu_offload if cpu_offload is not None else is_fsdp
+
+    log.debug('Extracting model state dict')
+    if version.parse(torch.__version__) >= version.parse('2.3.0') and dist.is_initialized():
+        from torch.distributed.checkpoint.state_dict import StateDictOptions
+        from torch.distributed.checkpoint.state_dict import get_model_state_dict as dcp_get_model_state_dict
```
mvpatel2000, line 55: Instead of renaming, can we import `state_dict` and call `state_dict.get_model_state_dict`?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+        precision: The precision of the model. Can be specified as a string ('fp32', 'fp16', 'bf16') or a torch.dtype.
+        include_keys: The list of keys to exclusively include in the state dict. If None, all keys are included. Both include_keys and ignore_keys cannot be non-None.
+        ignore_keys: The list of keys to ignore in the state dict. If None, no keys are ignored. Both include_keys and ignore_keys cannot be non-None.
+        cpu_offload: Whether to offload the state dict to CPU. If None, it is set to True if FSDP is enabled, False otherwise.
+
+    Returns:
+        The state dict of the model.
+    """
+    if include_keys is not None and ignore_keys is not None:
+        raise ValueError('Both include_keys and ignore_keys cannot be non-None.')
```
mvpatel2000, line 45:
```suggestion
        raise ValueError(f'Both {include_keys=} and {ignore_keys=} cannot be non-None.')
```

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+
+    is_fsdp = _is_model_fsdp(model)
+    if not is_fsdp and sharded:
+        raise ValueError('Sharded state dict can only be generated for FSDP models.')
+    cpu_offload = cpu_offload if cpu_offload is not None else is_fsdp
```
mvpatel2000, line 50: do we need to CPU offload for sharded ckpts?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,159 @@
...
+    if version.parse(torch.__version__) >= version.parse('2.3.0') and dist.is_initialized():
+        from torch.distributed.checkpoint.state_dict import StateDictOptions
+        from torch.distributed.checkpoint.state_dict import get_model_state_dict as dcp_get_model_state_dict
+        
+        get_nonsharded_state_dict = not sharded
+
+        log.debug('Calling torch get_model_state_dict...')
+        model_state_dict = torch_get_model_state_dict(
+            model=model,
+            submodules=None, # We will handle extracting submodules ourselves down below.
```
mvpatel2000, line 62:
```suggestion
            submodules=None,  # We extract submodules below
```
nit: more concise

mvpatel2000 (commented):
Mostly LGTM! just a few minor nits that should be quick to clean up. Also looks like tests are failing, I think because torch gating is a bit off

composer/checkpoint/state_dict.py
bigning: thanks! this makes sense!

bigning (approved):
Mostly looks good to me! Please address Mihir's comments.

composer/checkpoint/state_dict.py
eracah: oh good point. fixed

tests/common/models.py
eracah: One is ComposrModel and one iss nn.Module. We want to support both

mvpatel2000: Ah got it! Maybe nice to put a docstring on each saying use for (test composer/test pytorch raw) in tests?

tests/checkpoint/test_state_dict.py
```diff
@@ -0,0 +1,234 @@
...
+        sharded_state_dict=True,
+        include_keys=None,
+        ignore_keys=None,
+    )
+    for sharded_tens in model_state_dict.values():
+        local_tensor = sharded_tens.local_tensor() if tensor_type == 'sharded_tensor' else sharded_tens.to_local()
+        assert local_tensor.dtype == precision
+
+
+# TODO test precision
```
mvpatel2000, line 210:
```suggestion
```

tests/checkpoint/test_state_dict.py
```diff
@@ -0,0 +1,234 @@
...
+
+    model_state_dict = get_model_state_dict(model, sharded_state_dict=False, ignore_keys=['module.2*'])
+    assert set(model_state_dict.keys()) == {'module.0.weight'}
+
+
+#TODO add tests for sharded and for precision
+@pytest.mark.gpu
+@world_size(2)
+@pytest.mark.parametrize('tensor_type', ['sharded_tensor', 'dtensor'])
+@pytest.mark.parametrize('use_composer_model', [True, False])
```
mvpatel2000, line 65: Looks like the precision test is missing?

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,161 @@
...
+    precision: Union[str, torch.dtype] = 'fp32',
+    include_keys: Optional[Union[str, Sequence[str]]] = None,
+    ignore_keys: Optional[Union[str, Sequence[str]]] = None,
+    cpu_offload: Optional[bool] = None,
+) -> Dict[str, Any]:
+    """Generate the state dict of the model.
+
+    Args:
+        model: The model to get the state dict from.
+        sharded_state_dict: Whether the model state dict should be sharded or not. If True, every rank returns the state dict of its shards.
```
dakinggg, line 34: any reason not to default this to false? all args optional is nice

composer/checkpoint/state_dict.py
```diff
@@ -0,0 +1,161 @@
...
+            If False, then rank 0 returns the state dict of the entire model.
```
dakinggg, line 35: nit: specify in the docstring what the non0 ranks return in the mono case

composer/checkpoint/state_dict.py
dakinggg: probs can just move this to generic composer utils, isn't really checkpoint specific. nit though, not a strong opinion

tests/common/compare.py
```diff
@@ -20,6 +20,8 @@ def deep_compare(item1: Any, item2: Any, atol: float = 0.0, rtol: float = 0.0):
         item2 (Any): The second item
         atol (bool): Atol tolerance for torch tensors and numpy arrays (default: 0.0)
         rtol (float): Rtol tolerance for torch tensors and numpy arrays (default: 0.0)
+        device_check (bool): Whether to compare the device of torch tensors. If False
```
dakinggg, line 23: i think you just added a docsring for this arg but didn't add the arg?

dakinggg (commented):
few nits, one larger comment/question.

I'm guessing there are a lot of tests that could be simplified by using the functionality that this PR adds. Is that true? If so, is it worth trying to do at least some of that as part of this PR? It would also implicitly test the functionality being added more, since those would be real uses cases.

composer/checkpoint/state_dict.py
eracah: ok renamed as dcpsd because `state_dict` also has scopte conflict

eracah:
> few nits, one larger comment/question.
> 
> I'm guessing there are a lot of tests that could be simplified by using the functionality that this PR adds. Is that true? If so, is it worth trying to do at least some of that as part of this PR? It would also implicitly test the functionality being added more, since those would be real uses cases.

Yes in theory, we would want to do that. However, in this PR we are just adding the API to be used in a script; we aren't actually adding this code to be used in Trainer. So until we swap out state.py state dict generation code for this one, we don't need to change those other tests or add any E2E tests.

PR #3252, opened 2024-05-03
Fix backwards compatibility tests, raise error for torch version mismatch

snarayan21:
# What does this PR do?

Fixes FSDP checkpoint backwards compatibility tests, and raises a helpful error for likely torch version mismatch on checkpoint load. For context, the test that was failing was when trying to load a checkpoint written with torch 2.3.0 when using torch < 2.3.0, since the older torch versions could not correctly read the new metadata file. 

# Before submitting

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -623,11 +623,14 @@ def load_sharded_checkpoint(
             # how to load the optimizer state.
             try:
                 metadata = storage_reader.read_metadata()
-            except AttributeError:
-                raise ValueError(
-                    'Unable to read checkpoint metadata. The checkpoint was likely saved with a '
-                    'newer version of torch. Upgrade your torch version to load this checkpoint.',
-                )
+            except AttributeError as e:
+                if '_MEM_FORMAT_ENCODING' in str(e):
+                    raise ValueError(
+                        'Unable to read checkpoint metadata. The checkpoint was likely saved with a '
+                        'newer version of torch. Upgrade your torch version to load this checkpoint.',
+                    )
+                else:
+                    raise
             # Retrieve all top-level keys of the metadata.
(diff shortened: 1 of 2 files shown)


mvpatel2000 (approved):

composer/utils/checkpoint.py
```diff
@@ -621,7 +621,13 @@ def load_sharded_checkpoint(
             # 1. Load metadata first for backwards compatability check
             # We need to check if the "optimizers" is at the root of the state dict to determine
             # how to load the optimizer state.
-            metadata = storage_reader.read_metadata()
+            try:
+                metadata = storage_reader.read_metadata()
+            except AttributeError:
```
dakinggg, line 626: should we catch the more specific error here? instead of any attribute error?

snarayan21: addressed

snarayan21:
daily tests are passing too

composer/utils/checkpoint.py
```diff
@@ -621,7 +621,16 @@ def load_sharded_checkpoint(
...
+            try:
+                metadata = storage_reader.read_metadata()
+            except AttributeError as e:
+                if '_MEM_FORMAT_ENCODING' in str(e):
+                    raise ValueError(
+                        'Unable to read checkpoint metadata. The checkpoint was likely saved with a '
+                        'newer version of torch. Upgrade your torch version to load this checkpoint.',
+                    )
+                else:
+                    raise
```
b-chu, line 633: Uhh this is a bug right? You need to have `raise e`?

mvpatel2000: you can also do `raise` like this to reraise the error

snarayan21: nah
Screenshot 2024-05-06 at 2 38 43 PM

PR #3259, opened 2024-05-06
Fix: SAM not working with FSDP/DeepSpeed and LR scheduler.

Joqsan:
## Description

I'm tracing this bug back to Composer coming from using LLMFoundry after we noticed that our models didn't learn anything when running distributed training with FSDP alongside SAM and a LR scheduler. If running the same setup but w/o distributed training (on a single GPU), everything works fine.

## Reproducibility

Just checkout the source branch at the commit `d882d8f4`, which contains only the tests (without the fix), and run:

```bash

## All tests pass (no distributed training with 1 GPU)
make test-dist-gpu WORLD_SIZE=1 EXTRA_ARGS="tests/algorithms/test_sam.py::TestSAMParamGroups"

## All tests failed (FSDP and DeepSpeed tests)
make test-dist-gpu WORLD_SIZE=2 EXTRA_ARGS="tests/algorithms/test_sam.py::TestSAMParamGroups"
```

## Problem (FSDP)

- When using SAM with a LR scheduler that resets the starting `param_groups[0]['lr']` to `0.0` (such as  `CosineAnnealingWithWarmupScheduler`),  it varies the learning rate at `SAMOptimizer.param_groups[0]['lr']`,  but not the one at `base_optimizer.param_groups[0]['lr']` (which is also reset to `0.0`).
- So the latter remains equal to `0.0` for the whole training process, which is why we don't see any learning.

### Cause (short answer)
- `SAMOptimizer.param_groups[i]` and `base_optimizer.param_groups[i]` are not referencing the same params because of the way Composer handles the sharding.

<details>
  <summary><h3>Cause (long answer)</h3></summary>
  
  
   * Right when triggering `Event.INIT`, `SAMOptimizer` binds references to the param groups in `base_optimizer`., so up to this point both optimizers reference the same param groups.
   * Later (after initializing optimizers and anything triggered by `Event.INIT`) Composer initializes FSDP ([here](https://github.com/mosaicml/composer/blob/2daa2b6dbf871b0f35aa6482a628d00065807268/composer/trainer/trainer.py#L1561-L1571)), which is weird since the [FSDP docs](https://pytorch.org/docs/stable/fsdp.html) warns about initializing the optimizer **after** doing the sharding, not the other way around.
   * Composer does the sharding by clearing the param groups from the `SAMOptimizer` and bounding to it the new sharded params.
   * At the end of the whole `Trainer` initialization, `SAMOptimizer` and `base_optimizer` reference different param groups:
       * `SAMOptimizer` the sharded ones.
       * `base_optimizer` the original ones.
   * During training, `state.schedulers[0].step()` is done on the `'lr'` param  of `SAMOptimizer`, which are not the ones referenced by `base_optimizer`, but the real optimizer step is done by the `base_optimizer`, not `SAMOptimizer`. So no learning.
</details>

## Problem (DeepSpeed)

- As far as a I debugged, the problem above is just related to PyTorch's FSDP and not DeepSpeed (it handles the sharding process on its own).
- But a problem arises when allocating the internal optimizer state when initializing the `DeepSpeedEngine` ([here]( https://github.com/microsoft/DeepSpeed/blob/f1e4fb0b878e48d6fe33dbb2e0969d791d698e86/deepspeed/runtime/bf16_optimizer.py#L237)).
	- `SAMOptimizer` does the optimizer step iff a closure is passed, while `BF16_Optimizer.initialize_optimizer_states()` needs to do an optimizer step without any closure.

## Solution
- Let any sharding-related initialization be done on the param groups of the _would-be_ `base_optimizer` optimizer instance, and just after all of this wrap the resulting optimizer with a `SAMOptimizer` instance, binding references to the sharded param groups.
- For DeepSpeed this works, even though the underlying `base_optimizer` is a DeepSpeed Zero `BF16_Optimizer` instance, since it uses the same underlying parameter groups as the original PyTorch optimizer (see [here](https://github.com/microsoft/DeepSpeed/blob/fee73135980e78f8be7e1a3ff556751623ef6aaa/deepspeed/runtime/zero/stage_1_and_2.py#L1911-L1917) and [here](https://github.com/microsoft/DeepSpeed/blob/ef17c89570ceae5b26a5f886e9d8cd0941afc0ac/deepspeed/runtime/zero/stage3.py#L2532-L2538)).

This fix is achieved by triggering the `SAMOptimizer` wrapping later in the code. After this, distributed training works as expected.

# Before submitting

### DIFF
--- a/composer/algorithms/sam/sam.py
+++ b/composer/algorithms/sam/sam.py
@@ -176,4 +176,5 @@ class SAM(Algorithm):
             ) for optimizer in ensure_tuple(state.optimizers)
         )
 
+        # Switch to ClosureGradScaler as SAM supports and requires it
         state.scaler = ClosureGradScaler()
--- a/composer/algorithms/sam/sam.py
+++ b/composer/algorithms/sam/sam.py
@@ -121,7 +121,7 @@ class SAM(Algorithm):
     by wrapping an existing optimizer with a :class:`.SAMOptimizer`. SAM can improve model generalization
     and provide robustness to label noise.
 
-    Runs on :attr:`.Event.INIT`.
+    Runs on :attr:`.Event.AFTER_LOAD`.
 
     Args:
         rho (float, optional): The neighborhood size parameter of SAM. Must be greater than 0.
@@ -161,7 +161,7 @@ class SAM(Algorithm):
         self.interval = interval
 
     def match(self, event: Event, state: State) -> bool:
-        return event == Event.INIT
+        return event == Event.AFTER_LOAD
 
     def apply(self, event: Event, state: State, logger: Optional[Logger]) -> Optional[int]:
         assert state.optimizers is not None
--- a/composer/algorithms/sam/sam.py
+++ b/composer/algorithms/sam/sam.py
@@ -13,6 +13,7 @@ import torch
 
 from composer.core import Algorithm, Event, State
 from composer.loggers import Logger
+from composer.trainer._scaler import ClosureGradScaler
 from composer.utils import ensure_tuple
 
 log = logging.getLogger(__name__)
@@ -174,3 +175,5 @@ class SAM(Algorithm):
                 interval=self.interval,
             ) for optimizer in ensure_tuple(state.optimizers)
         )
+
+        state.scaler = ClosureGradScaler()
--- a/tests/algorithms/test_sam.py
+++ b/tests/algorithms/test_sam.py
@@ -1,14 +1,19 @@
 # Copyright 2022 MosaicML Composer authors
 # SPDX-License-Identifier: Apache-2.0
 
+import math
+
 import pytest
+import torch
 from torch.utils.data import DataLoader
 
 from composer import Trainer
 from composer.algorithms import SAM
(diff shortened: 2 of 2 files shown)


mvpatel2000:
Thanks for the PR! Once the tests pass, you can tag me for review.

Also, the PR description is really well done 🙏 -- appreciate the detailed writeup. Fix makes sense to me! Once tests pass I'll approve

Joqsan:
@mvpatel2000 Hi!

I fixed the failing tests.

Now the `SAM` algorithm always set the `state.scaler` to `ClosureGradScaler()`. This is safe since `Trainer._use_grad_scaling()` takes care of enabling/disabling the grad scaling step during training.

Joqsan:
@mvpatel2000 yapf formatting tests are failing, but they are not related to files I changed in this PR.

How should I proceed?

mvpatel2000:
> @mvpatel2000 yapf formatting workflows are failing, but they are not related to files I changed in this PR.
> 
> How should I proceed?

I can help out with linting :) will take a look soon!

mvpatel2000 (approved):
LGTM!

PR #3260, opened 2024-05-06
Add unit test support for full state dict + load_weights_only and save_weights_only

eracah:


mvpatel2000 (commented):
Can we narrow down this to not run the full permutation? to help speed up tests?

eracah:
> Can we narrow down this to not run the full permutation? to help speed up tests?

yeah but I prob won't get around to doing that for another couple days.

mvpatel2000 (approved):

PR #3265, opened 2024-05-08
Add upload timeout patch to mlflow on azure

dakinggg:
# What does this PR do?
We are fixing an occasional hang during artifact upload by patching mlflow to pass a request timeout along, and to set the default socket timeout in the upload worker function. This will be fixed upstream by mlflow in the next release, at which point we can remove the patch.

Manual tests:
- upload without the timeout flag set: `upload-timeout-test-without-2-xlw9bk`
- upload with the timeout flag set: `upload-timeout-test-3-UzoDmx`
- upload with the timeout flag set (and seemingly did something, as this run took longer to complete than the others): `upload-timeout-test-2-sVfbkw`

### DIFF
--- a/composer/utils/object_store/mlflow_object_store.py
+++ b/composer/utils/object_store/mlflow_object_store.py
@@ -132,7 +132,7 @@ def _patch_adls_file_upload_with_timeout(sas_url, local_file, start_byte, size, 
     data = read_chunk(local_file, size, start_byte)
 
     ### Changed here to pass a timeout along to cloud_storage_http_request
-    timeout = int(os.environ['MLFLOW_PATCH_ADLS_FILE_UPLOAD_TIMEOUT'])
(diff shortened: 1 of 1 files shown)


composer/utils/object_store/mlflow_object_store.py
```diff
@@ -127,6 +216,10 @@ def __init__(self, path: str, multipart_upload_chunk_size: int = 100 * 1024 * 10
         except ImportError as e:
             raise MissingConditionalImportError('databricks', conda_package='databricks-sdk>=0.15.0,<1.0') from e
 
+        log.debug('Patching MLflow Azure client to include timeout in ADLS file upload')
```
irenedea, line 219: Can you add a brief comment to describe why we're doing this

irenedea (approved):

composer/utils/object_store/mlflow_object_store.py
dakinggg: done

PR #3269, opened 2024-05-08
Tensor Parallelism Integration

mvpatel2000:
# What does this PR do?

This PR adds Tensor Parallelism (TP) integration. As part of this, we simplify the Trainer interface for parallelism.

Current limitations:
- TP requires FSDP
- TP means no load_monolith_rank0_only
- TP does not support monolith checkpointing
- TP must be all layers with same TP world size

Task List
- [x] Add TensorParallelism integration
- [x] Switch to `parallelism_config` interface
- [x] Code cleanup with respect to where `device_mesh` is tracked
- [x] Make `device_mesh/DTensor` on by default with FSDP
- [x] Docs update
- [x] Move all distributed work to a new folder

Test List
- [x] Add TP train test
- [x] Add TP sharded resumption test
- [x] Add TP full state dict resumption test

Follow-on PR:
- [ ] Switch to Dataclass
- [ ] Load monolith rank0 only support 
- [ ] Train TP only
- [ ] Add TP vs. FSDP determinism test (blocked by no monolith checkpointing)

Note:
- `simulate_tp.py` will be deleted before merge. It is currently used for testing.


composer/core/state.py
```diff
@@ -468,13 +531,32 @@ def __init__(
...
+                raise ValueError(
+                    'Tensor parallelism (TP) currently requires FSDP to be enabled. '
+                    'An empty `fsdp_config` can be specified to enable FSDP with '
+                    'default settings.',
+                )
 
         if self.load_monolith_rank0_only:
-            assert fsdp_config is not None
+            if self.tp_config is not None:
+                raise ValueError('load_fsdp_monolith_rank0_only is not compatible with tensor parallelism (TP).')
```
snarayan21, line 556: Is there a fix the user should use if they hit this?

composer/trainer/dist_strategy.py
```diff
@@ -319,31 +304,26 @@ def sync_hook(*args):
...
-                    f'of size 2 but got device mesh size of {device_mesh_size}.',
-                )
-            from torch.distributed._tensor import init_device_mesh
-            kwargs['device_mesh'] = init_device_mesh(
-                'cuda',
-                tuple([int(x) for x in fsdp_config['device_mesh']]),
+    if version.parse(torch.__version__.split('.dev')[0]) >= version.parse('2.2.0') and 'device_mesh' in fsdp_config:
+        if 'process_group' not in fsdp_config:
+            warnings.warn(
+                'process_group and device_mesh are set for FSDP, so ignoring device_mesh. Please set process_group to None.',
```
snarayan21, line 310: If we're recommending setting process_group to None, and also defaulting to using `DeviceMesh` everywhere, why do we ignore `device_mesh` here?

composer/trainer/mosaic_fsdp_utils.py
```diff
@@ -65,6 +64,51 @@
...
+    if 'device_mesh' in fsdp_config:
+        warnings.warn(VersionedDeprecationWarning('device_mesh is deprecated. Please specify `data_parallel_shard_degree` and `data_parallel_replicate_degree` instead.', remove_version='0.24.0'))
+        if 'data_parallel_shard_degree' in fsdp_config or 'data_parallel_replicate_degree' in fsdp_config:
+            raise ValueError('Cannot specify both `device_mesh` and `data_parallel_shard_degree` or `data_parallel_replicate_degree`. Please remove `device_mesh`.')
+        device_mesh = fsdp_config.pop('device_mesh')
+        fsdp_config['data_parallel_shard_degree'] = device_mesh[0]
+        if len(device_mesh) > 1:
+            fsdp_config['data_parallel_replicate_degree'] = device_mesh[1]
+
+    fsdp_config.setdefault('activation_checkpointing', False)
```
snarayan21, line 81: these will be handled by a dataclass in a future PR right

composer/utils/checkpoint.py
```diff
@@ -1107,12 +1109,13 @@ def _save_checkpoint(
...
         log.debug(f'Saving sharded checkpoints to {save_filename}...')
         process_group = None
-        device_mesh = state.fsdp_device_mesh
+        device_mesh = state.device_mesh
         if device_mesh is not None and device_mesh.ndim == 2:
             # If hybrid shard, only rank in first replica saves
             expect_file = device_mesh.get_local_rank(mesh_dim=0) == 0
             if expect_file:
                 process_group = device_mesh.get_group(1)  # Shard process_group for first replica
+                assert isinstance(process_group, ProcessGroup)
```
snarayan21, line 1118: is this simply for linting or is there another reason

tests/trainer/test_tp.py
```diff
@@ -0,0 +1,53 @@
...
+        ColwiseParallel,
+        RowwiseParallel,
+    )
+
+    model = SimpleModel()
+    dataset = RandomClassificationDataset(size=10)
+    dataloader = DataLoader(dataset, sampler=dist.get_sampler(dataset))
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
+
+    layer_plan = {
```
snarayan21, line 34: Could we test out TP for other layers like norms as well?

snarayan21 (commented):
lookin nice, left some comments

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -743,13 +746,19 @@ def test_fsdp_partitioned_state_dict_load(
...
     load_ignore_keys: Union[list[str], None],
     use_symlink: bool,
+    use_tp: bool,
     use_remote,
     s3_bucket,
     s3_ephemeral_prefix,
     request,
 ):
+    if use_tp:
+        pytest.skip('TP does not work with DP dimension 1')
```
bigning, line 756: the TP test is always skipped ?  We can only run 1-GPU test at this moment, is my understanding correct ?

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -767,6 +776,17 @@ def test_fsdp_partitioned_state_dict_load(
     save_filename = 'ba{batch}-rank{rank}.pt'
 
     fsdp_config = FSDPConfig(state_dict_type='sharded')
+    tp_config = None
+    if use_tp:
+        fsdp_config = FSDPConfig(data_parallel_shard_degree=1)
```
bigning, line 781: IIUC, this is essentially TP only test? Can we also add HSDP + TP, and FSDP + TP? I know we don't support multi-gpu auto-test, but maybe at least add those and run locally for now ?

composer/core/state.py
```diff
@@ -194,6 +195,70 @@ def _ensure_backwards_compatible_checkpointing(state_dict: Dict[str, Any]):
...
+    names.append('dp_shard')
+    if fsdp_config['data_parallel_replicate_degree'] != 1:
+        dims.append(fsdp_config['data_parallel_replicate_degree'])
+        names.append('dp_replicate')
+    if tp_config is not None:
+        dims.append(tp_config['tensor_parallel_degree'])
+        names.append('tp')
+
+    # Fill in the unspecified dimensions
+    product_of_dims = 1
```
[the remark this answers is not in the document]
mihir-db: FSDP defaults to shard dimension being unspecified and replicate dimension being 1. I can put a `log.info` when we autoinfer a dimension

composer/core/state.py
```diff
@@ -194,6 +195,70 @@ def _ensure_backwards_compatible_checkpointing(state_dict: Dict[str, Any]):
...
+        if product_of_dims != dist.get_world_size():
+            raise ValueError(
+                f'World size {dist.get_world_size()} does not equal the product of the specified parallelism degrees '
+                f'{product_of_dims}. Please ensure the product of the specified parallelism degrees matches the world ',
+                f'size. Currently specified degrees are {names=}, {dims=}. One dimension can also be left as -1, which '
+                'will automatically be specified to ensure the product matches the world size.',
+            )
+
+    device_type = device.name
+    if device_type == 'gpu':
```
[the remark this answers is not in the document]
mvpatel2000: Yep!

composer/core/state.py
```diff
@@ -468,13 +531,32 @@ def __init__(
...
+        self.tp_config = parallelism_config.get('tp', None)
+        if self.fsdp_config is not None:
+            # Add an earlier call to patch_pytorch as we require device_mesh slicing before any
+            # model wrapping.
+            from composer.trainer.mosaic_fsdp import patch_pytorch
+
+            patch_pytorch()
+
+        if self.tp_config is not None:
+            if version.parse(torch.__version__.split('.dev')[0]) < version.parse('2.3.0'):
```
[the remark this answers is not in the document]
mvpatel2000: We are trying to standardize to my code since its the most robust, they sometimes have postfixes after dev so we just want to discard all that

composer/core/state.py
```diff
@@ -468,13 +531,32 @@ def __init__(
...
+                raise ValueError('Tensor parallelism (TP) requires torch>=2.3.0.')
+            if self.fsdp_config is None:
+                raise ValueError(
```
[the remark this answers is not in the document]
mvpatel2000: Not right now :( it crashes wtih TP + DDP or TP pure

composer/trainer/mosaic_fsdp_utils.py
mvpatel2000: Yep!

composer/utils/checkpoint.py
mvpatel2000: Will add comment, yes for typechecker

composer/core/state.py
```diff
@@ -194,6 +195,70 @@ def _ensure_backwards_compatible_checkpointing(state_dict: Dict[str, Any]):
...
+    if fsdp_config is None:
+        return None
+
+    # Gather dimensions and names for the device mesh
+    dims: List[int] = []
+    names: List[str] = []
+    dims.append(fsdp_config['data_parallel_shard_degree'])
+    names.append('dp_shard')
+    if fsdp_config['data_parallel_replicate_degree'] != 1:
+        dims.append(fsdp_config['data_parallel_replicate_degree'])
```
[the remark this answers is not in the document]
mvpatel2000: If you replicate first, you would replicate within a node and shard multinode (as an example)

tests/trainer/test_fsdp_checkpoint.py
mvpatel2000: Yea, we need 4 GPU tests to support. @KuuCi  is adding soon!

composer/trainer/dist_strategy.py
mvpatel2000: We autoset device mesh, so if you specify FSDP, it would have both and thus the user specified one is likely process_group... 🤔. It's pretty hacky...

tests/trainer/test_tp.py
mvpatel2000: Any specific reason? it's a bit annoying since we dont have a convenient standard model with norms, but I can add if you think its important

composer/utils/checkpoint.py
```diff
@@ -1107,12 +1109,13 @@ def _save_checkpoint(
 
         log.debug(f'Saving sharded checkpoints to {save_filename}...')
         process_group = None
-        device_mesh = state.fsdp_device_mesh
+        device_mesh = state.device_mesh
         if device_mesh is not None and device_mesh.ndim == 2:
             # If hybrid shard, only rank in first replica saves
             expect_file = device_mesh.get_local_rank(mesh_dim=0) == 0
```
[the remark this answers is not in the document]
mvpatel2000: Yes good catch. @KuuCi  another thing that needs a unit test w 4 GPUs :( I will fix

composer/trainer/trainer.py
```diff
@@ -911,7 +913,21 @@ class Trainer:
...
+                for more details. To use FSDP with default values, set to the empty dictionary ``{}``. To
+                disable FSDP, set to ``None`` or remove the key from the dictionary.
+
+            For `parallelism_config['tp']`, see TODO: ADD DOCS!
+                for more details. To use Tensor Parallelism with default values, set to the empty dictionary ``{}``. To
+                disable Tensor Parallelism, set to ``None`` or remove the key from the dictionary.
+
+            .. note:: This parameter is experimental and subject to change without standard deprecation
+                cycles.
+        tp_config (Dict[str, Any], optional): Configuration for tensor parallelism.
```
dakinggg, line 930: remove?

composer/utils/checkpoint.py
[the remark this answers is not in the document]
bigning: yeah, the 4 GPUs unit test is really important to improve the test coverage. Thank you @KuuCi for working on that!

simulate_tp.py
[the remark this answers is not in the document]
mvpatel2000: Yes see comment

tests/trainer/test_fsdp.py
```diff
@@ -417,3 +402,54 @@ def oom_hook(module, grad_input, grad_ouput):
     output_2 = fsdp_oom_model(x)
 
     assert torch.equal(output_1, output_2)
+
+
+@pytest.mark.gpu
+@world_size(2)
+def test_fsdp_device_mesh(world_size: int):
```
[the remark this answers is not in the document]
mvpatel2000: Yes, device mesh is deprecated

tests/trainer/test_tp.py
```diff
@@ -0,0 +1,53 @@
...
+from tests.common import (
+    RandomClassificationDataset,
+    SimpleModel,
+    world_size,
+)
+
+
+@pytest.mark.gpu
+@world_size(2)
+@pytest.mark.skip  # TP does not work with DP dimension 1
```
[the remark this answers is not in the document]
mvpatel2000: I will change to DP dim 2 when Vincent adds 4 GPU testing

composer/core/state.py
```diff
@@ -468,13 +531,32 @@ def __init__(
...
         self.deepspeed_config = deepspeed_config
-        self.fsdp_config = fsdp_config
-        self.fsdp_auto_wrap = fsdp_auto_wrap
+        parallelism_config = parallelism_config or {}
+        self.fsdp_config = parallelism_config.get('fsdp', None)
+        self.tp_config = parallelism_config.get('tp', None)
+        if self.fsdp_config is not None:
+            # Add an earlier call to patch_pytorch as we require device_mesh slicing before any
+            # model wrapping.
+            from composer.trainer.mosaic_fsdp import patch_pytorch
```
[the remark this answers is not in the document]
mvpatel2000: I would like to avoid patching if you don't need it since it might break things, but not strongly opinionated

composer/core/state.py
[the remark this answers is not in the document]
snarayan21: That would be great, could you point out where you added this log?

composer/core/state.py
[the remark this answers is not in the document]
snarayan21: should we get those fixes in this PR, or is that an item that depends on torch fixes? Can you note this in the PR description as well

composer/core/state.py
snarayan21: We should include unsupported edge cases in docs -- would help decrease support burden and we can always edit down the line

composer/trainer/dist_strategy.py
snarayan21: I thought we were going to be completely moving to `device_mesh` so that we move to DTensor fully. Does using `process_group` go against that

composer/utils/checkpoint.py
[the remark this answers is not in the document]
snarayan21: For my own understanding @bigning @mvpatel2000 :
* with HSDP, device_mesh dim 0 will be the replica process group, and so we want rank 0 in each replica process group (that is, all ranks in the first replica) to save checkpoint shards.
* with HSDP + TP, we still want the same behavior where all ranks in the first HSDP replica saves checkpoint shards, but now mesh_dim 0 is `data_parallel_shard_degree`....why is this changed from normal HSDP?
* with FSDP + TP, we want every rank to save a checkpoint shard still.

tests/trainer/test_tp.py
snarayan21: For example, we're not testing sequence parallel elementwise ops with TP, which is what we'd want to do in a full TP setup. It would be nice to validate that this is working by adding it to the layer_plan and/or making a model class variant.

snarayan21 (commented):
few comments

composer/core/state.py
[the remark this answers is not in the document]
mvpatel2000: https://github.com/mosaicml/composer/pull/3269/files#diff-03f7202bce6ff365b02b9171a328331179a8a5182232cdd5b3a720691f47a98bR249

composer/core/state.py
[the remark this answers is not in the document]
mvpatel2000: Added to description., I think require torch fixes

composer/trainer/dist_strategy.py
mvpatel2000: Yes, but we have to deprecate. we cant just delete :(

composer/utils/checkpoint.py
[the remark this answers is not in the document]
mvpatel2000: @snarayan21 all we want to do here is slice into the first replica on replicate dimension (we dont care about any other dims and how they are sharded)

composer/core/state.py
```diff
@@ -521,6 +608,15 @@ def __init__(
                 ),
             )
 
+        self.device_mesh: Optional[DeviceMesh] = _create_device_mesh(self.device, self.fsdp_config, self.tp_config)
+        if self.fsdp_config is not None and self.device_mesh is not None:
+            fsdp_mesh_dim_names = [ParallelismType.DATA_PARALLEL_SHARD.value]
+            if self.device_mesh.mesh_dim_names is not None and ParallelismType.DATA_PARALLEL_REPLICATE.value in self.device_mesh.mesh_dim_names:
+                fsdp_mesh_dim_names.append(ParallelismType.DATA_PARALLEL_REPLICATE.value)
+            self.fsdp_config['device_mesh'] = self.device_mesh[tuple(fsdp_mesh_dim_names)]  # type: ignore
```
bigning, line 616: I actually think the order here might be not correct ? according to the pytorch doc:
1. 
image
2. [_inter_node_pg to be the outermost dimension, and the process_group(intra_node) to be the innermost dimension](https://github.com/pytorch/pytorch/blob/5325a6de64c9b11a8f35e59e709499135d23579c/torch/distributed/fsdp/_init_utils.py#L163-L166)

My understanding is that the outermost dimension (dim-0) should be replicate, and innermost dimension (dim-1) should be shard dimension. 

cc @snarayan21

bigning: I think we might need 6 GPUs or 8 GPUs (4 X2 or 2 X 4) to test this , lol. 4 GPUs (2 X 2) can't test this order.

mvpatel2000: Yep you're correct! Fixed

bigning: It seems you only fixed the order when constructing the self.device_mesh [here in the beginning of this file](https://github.com/mosaicml/composer/pull/3269/commits/00ce8c654aa0d19a47b308210baa4e67a370e48d) , but here when you construct the device_mesh for fsdp, in `fsdp_mesh_dim_names`, the first one is still `DATA_PARALLEL_SHARD`, not `DATA_PARALLEL_REPLICATE` ?

bigning:
Almost looks good to me. just to clarify, do you want to add unit test for HSDP + TP checkpointing? If so, do you want to add that in this PR or do you want to add those in a follow-up PR?
Since this changed a lot code, can we run an e2e training/resumption with a real model with >=64 GPUs to make sure the existing FSDP and HSDP are still working well ? :)

composer/core/state.py
```diff
@@ -194,6 +195,75 @@ def _ensure_backwards_compatible_checkpointing(state_dict: Dict[str, Any]):
...
+                'parallelism degrees. Please ensure the product of the specified parallelism degrees '
+                'matches the world size.',
+            )
+        for i, dim in enumerate(dims):
+            if dim == -1:
+                dims[i] = remaining_dimension
+                log.info(f'Automatically setting {names[i]} to have parallelization degree {remaining_dimension}.')
+                break
+    else:
+        if product_of_dims != dist.get_world_size():
```
bigning, line 252: sorry, i have one more question here. How will we apply TP ? IIUC, we only apply TP to some of the linear modules, not all the linear modules, right? If it's FSDP + TP with world_size 8, and device mesh is 2 (dp) x 4 (TP), that means, for the linear modules that doesn't apply TP, their parallel degree is only 2, which means there are 4 replicas for non-TP modules. Is that what we want ? Or do we still want those non-TP linear modules sharded across the 8 GPUs ?

mvpatel2000: Hm... this is a good question 🤔 I'm not sure. I will check what happens...

tests/trainer/test_tp.py
mvpatel2000: Discussed offline, because its complicated we're going to push until later when we add to foundry.

mvpatel2000:
> Almost looks good to me. just to clarify, do you want to add unit test for HSDP + TP checkpointing? 

Yep, working on adding requested unit tests. 

> Since this changed a lot code, can we run an e2e training/resumption with a real model with >=64 GPUs to make sure the existing FSDP and HSDP are still working well ? :)

will do once all major issues are resolved!

composer/distributed/dist_strategy.py
```diff
@@ -319,31 +304,26 @@ def sync_hook(*args):
...
-            ] and device_mesh_size != 2:
-                raise ValueError(
-                    f'FSDP sharding strategy {sharding_map_key.upper()} requires a device mesh '
-                    f'of size 2 but got device mesh size of {device_mesh_size}.',
-                )
-            from torch.distributed._tensor import init_device_mesh
-            kwargs['device_mesh'] = init_device_mesh(
-                'cuda',
-                tuple([int(x) for x in fsdp_config['device_mesh']]),
+    if version.parse(torch.__version__.split('.dev')[0]) >= version.parse('2.2.0') and 'device_mesh' in fsdp_config:
```
b-chu, line 307: Is the split part necessary? Only 2.3.0 has .dev currently right?

composer/distributed/dist_strategy.py
```diff
@@ -319,31 +304,26 @@ def sync_hook(*args):
...
+        if fsdp_config['process_group'] is not None:
```
b-chu, line 308: This should be structured so you can't set both

composer/distributed/mosaic_fsdp.py
b-chu, line 1: Why combine mosaic_fsdp and mosaic_fsdp_utils into the same file?

composer/trainer/trainer.py
```diff
@@ -911,7 +914,20 @@ class Trainer:
...
             the module is already wrapped outside, allow the user to disable auto-wrapping.
+        parallelism_config (Dict[str, Any], optional): Configuration for parallelism options.
+            Currently supports fsdp and tensor parallelism, whose respective configs are specified
+            as the keys ``fsdp`` and ``tp``. (default: ``None``)
 
+            For `parallelism_config['fsdp']`, see :doc:`FSDP Documentation </notes/distributed_training>`
+                for more details. To use FSDP with default values, set to the empty dictionary ``{}``. To
+                disable FSDP, set to ``None`` or remove the key from the dictionary.
+
+            For `parallelism_config['tp']`, see TODO: ADD DOCS!
```
b-chu, line 925: Please add docs before merging

composer/trainer/trainer.py
```diff
@@ -911,7 +914,20 @@ class Trainer:
...
+                for more details. To use Tensor Parallelism with default values, set to the empty dictionary ``{}``. To
+                disable Tensor Parallelism, set to ``None`` or remove the key from the dictionary.
+
+            .. note:: This parameter is experimental and subject to change without standard deprecation
```
b-chu, line 929: Use foundry's experimental tag?

composer/trainer/trainer.py
```diff
@@ -1051,10 +1067,11 @@ def __init__(
         # Graceful Resumption
         autoresume: bool = False,
 
-        # DeepSpeed
+        # Parallelism
         deepspeed_config: Optional[Dict[str, Any]] = None,
         fsdp_config: Optional[Dict[str, Any]] = None,
         fsdp_auto_wrap: bool = True,
+        parallelism_config: Optional[Dict[str, Any]] = None,
```
b-chu, line 1074:
```suggestion
        parallelism_config: Optional[dict[str, Any]] = None,
```

composer/trainer/trainer.py
```diff
@@ -1161,7 +1178,40 @@ def __init__(
         assert not isinstance(device_train_microbatch_size, str)
 
         # Distributed
-        if deepspeed_config is not None or fsdp_config is not None or dist.get_world_size() > 1:
+        if fsdp_config is not None:
```
b-chu, line 1181: init function is getting too long. Put this in a helper function init_parallelism_config

composer/trainer/trainer.py
```diff
@@ -1553,12 +1602,22 @@ def __init__(
...
+        if self.state.tp_config is not None:
+            with reproducibility.seed_context(self.state.rank_zero_seed):
+                prepare_tp_module(
+                    model,
+                    self.state.tp_config,
+                )
 
         # FSDP wrap if not using monolith checkpoint on rank 0 only
-        if self.state.fsdp_config is not None and fsdp_auto_wrap and not self.state.load_monolith_rank0_only:
+        if self.state.fsdp_config is not None and self.state.fsdp_config['auto_wrap'
```
b-chu, line 1619: Fix wrapping

composer/trainer/trainer.py
```diff
@@ -1728,7 +1788,8 @@ def __init__(
 
         # FSDP wrap if model is not yet wrapped and FSDP is enabled. This can happen if
         # load_monolith_rank0_only=True but no checkpoint was loaded.
-        if not self.state.fsdp_enabled and self.state.fsdp_config is not None and self.state.fsdp_auto_wrap and self.state.load_monolith_rank0_only:
+        if not self.state.fsdp_enabled and self.state.fsdp_config is not None and self.state.fsdp_config[
```
b-chu, line 1791: Fix wrapping

composer/core/state.py
```diff
@@ -468,13 +536,32 @@ def __init__(
         self.profiler: Optional[Profiler] = None
 
         self.deepspeed_config = deepspeed_config
-        self.fsdp_config = fsdp_config
-        self.fsdp_auto_wrap = fsdp_auto_wrap
+        parallelism_config = parallelism_config or {}
+        self.fsdp_config = parallelism_config.get('fsdp', None)
+        self.tp_config = parallelism_config.get('tp', None)
```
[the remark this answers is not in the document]
mvpatel2000: Follow on PR, see PR description

composer/utils/misc.py
```diff
@@ -39,6 +41,19 @@
 }
 
 
+class ParallelismType(StringEnum):
```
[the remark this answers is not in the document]
mvpatel2000: I think I already only use it for pytorch interface

composer/distributed/dist_strategy.py
mvpatel2000: We should do this everywhere to be consistent -- was decision when working on nightly

composer/distributed/dist_strategy.py
mvpatel2000: Wdym?

composer/distributed/mosaic_fsdp.py
mvpatel2000: It's just a single patch call?

composer/trainer/trainer.py
mvpatel2000: How do I do that?

composer/core/state.py
```diff
@@ -468,13 +536,32 @@ def __init__(
...
+            from composer.distributed import patch_pytorch
+
+            # Add an earlier call to patch_pytorch as we require device_mesh slicing before any
+            # model wrapping.
+            patch_pytorch()
+
+        if self.tp_config is not None:
+            if version.parse(torch.__version__.split('.dev')[0]) < version.parse('2.3.0'):
+                raise ValueError('Tensor parallelism (TP) requires torch>=2.3.0.')
+            if self.fsdp_config is None:
```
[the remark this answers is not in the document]
mvpatel2000: I have tweaked the error message and will leave as is. unfortunately, it seems currently the default values don't work since we require dp_shard>=2 :(

We will change to default on once torch issues are resolved

composer/distributed/dist_strategy.py
mvpatel2000: Discussed offline, will happen in follow-on PR.

composer/core/state.py
```diff
@@ -539,28 +539,72 @@ def __init__(
...
-            patch_pytorch()
+        self._validate_parallelism_configs()
+
+        self.device_mesh: Optional[DeviceMesh] = _create_device_mesh(self.device, self.fsdp_config, self.tp_config)
+        if self.fsdp_config is not None and self.device_mesh is not None:
+            fsdp_mesh_dim_names = []
+            if self.device_mesh.mesh_dim_names is not None and ParallelismType.DATA_PARALLEL_REPLICATE.value in self.device_mesh.mesh_dim_names:
+                fsdp_mesh_dim_names.append(ParallelismType.DATA_PARALLEL_REPLICATE.value)
+            fsdp_mesh_dim_names.append(ParallelismType.DATA_PARALLEL_SHARD.value)
+            self.fsdp_config['device_mesh'] = self.device_mesh[tuple(fsdp_mesh_dim_names)]  # type: ignore
```
b-chu, line 551: Why is type: ignore needed? Can you filter out the specific error instead?

b-chu (approved):
Minor comments, but looks good! Thanks for the PR!

composer/core/state.py
```diff
@@ -539,28 +539,72 @@ def __init__(
...
+            'scaler',
+            'timestamp',
+            'rank_zero_seed',
+            'train_metrics',
+            'eval_metrics',
+            'run_name',
+            'dataset_state',
+        ]
+
+        self.train_metrics: Optional[Dict[str, Metric]] = {}
```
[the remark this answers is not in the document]
mvpatel2000: Follow on PR given this is not from this PR

composer/core/state.py
b-chu: Also interested in knowing what happens. Does this currently apply TP to all linear modules?

PR #3271, opened 2024-05-09
 new dist_cp save planner to fix issue that each rank needs to download all checkpoint files 

bigning:
# what's the issue
PyTorch Issue: https://github.com/pytorch/pytorch/issues/125740
When pytorch 2.3 saves checkpoint files, if there are duplicated tensors, it will do dedup. In the dedup, it tries to balance the storage needed for each rank, so the duplicated tensors are saved on different ranks. The consequest is that, when loading checkpoint, each rank needs to read files saved by different ranks even without resharding. Ideally, each rank_k only needs to download rank_k and rank_0 files. 

# what's the fix
we save the duplicated tensors to the smallest rank only. This is how torch2.1 and torch2.2 do the dedup. 

# test
Tested  on 512 GPUs. Save and load both work.

### DIFF
--- a/composer/trainer/mosaic_fsdp_utils.py
+++ b/composer/trainer/mosaic_fsdp_utils.py
@@ -732,7 +732,7 @@ if version.parse(torch.__version__) >= version.parse('2.3.0') and version.parse(
     from torch.distributed.checkpoint.planner import SavePlan, WriteItem
     from torch.distributed.checkpoint.metadata import MetadataIndex, Metadata
 
-    def dedup_save_plans(all_plans: List[SavePlan]) -> List[SavePlan]:
+    def dedup_save_plans(all_plans: List[SavePlan]) -> List[SavePlan]:  # noqa: D103
         write_item_to_plan_indices: Dict[MetadataIndex, Set[int]] = defaultdict(set)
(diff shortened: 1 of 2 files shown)


composer/trainer/mosaic_fsdp_utils.py
```diff
@@ -717,3 +717,83 @@ def set_optimizer_state_dict(
...
+                # map each write item to its plan
+                write_item_to_plan_indices[write_item.index].add(plan_idx)
+                write_item_idx_to_write_item[write_item.index] = write_item
+
+        # put item in the plan with the smallest size and remove it from the other plan_indices
+        to_remove: List[Set] = [set() for _ in range(len(all_plans))]
+        plan_to_size = [0] * len(all_plans)
+        for write_item_idx, plan_indices in write_item_to_plan_indices.items():
+            # this line is the fix, to keep the duplicated tensors on the same rank
+            select_plan_idx = min(plan_indices, key=lambda plan_idx: plan_idx)
```
bigning, line 754: Most of the code in this files are copied from torch. This line is the only change, to select the lowest rank, instead of selecting the rank with minimum storage.

mvpatel2000 (approved):

PR #3273, opened 2024-05-09
Add option to accumulate train loss over tokens.

aadyotb:
# What does this PR do?

Adds an option to accumulate train loss over the number of tokens in a batch instead of the number of samples.

Currently, losses in the trainer are accumulated in a way that weights each sample in a batch equally. However, for NLP use cases where batches contain padding tokens, it makes more sense to accumulate the loss in a way that instead weights every (non-padding) token equally.

# Before submitting


mvpatel2000:
@dakinggg @irene-dea  can you please look?

I agree we should have an option for this. I'm not sure if it's necessary to pass to Composer vs. check if its an attribute/property on a dataloader

dakinggg:
@mvpatel2000 I think trainer arg is right for this...code looks fine at a glance, would want to test a bit more before merging.

mvpatel2000:
> @mvpatel2000 I think trainer arg is right for this...code looks fine at a glance, would want to test a bit more before merging.

You will own testing?

composer/trainer/trainer.py
```diff
@@ -2733,7 +2737,10 @@ def _train_microbatches(
                         optimizer.zero_grad()
 
             # Tracker for gradient accumulation
-            current_batch_size = sum([self._train_data_spec.get_num_samples_in_batch(batch) for batch in microbatches])
+            if self.accumulate_train_batch_on_tokens:
+                current_batch_size = sum([self._train_data_spec.get_num_tokens_in_batch(b) for b in microbatches])
```
mvpatel2000, line 2741: Can we throw a warning if this is 0? Aka you use a default data spec which fails to parse number of tokens

mvpatel2000 (commented):
Can we add a unit test please?

aadyotb:
> Can we add a unit test please?

Is there a good template I can base this test on? I'm not sure how to isolate the impact of this change from the rest of the trainer.

dakinggg:
@aadyotb Here is a test that exercises the full NLP pipeline (https://github.com/mosaicml/composer/blob/01eec3ad1732074c52cf5da35ef5b8a24531279a/tests/test_full_nlp.py#L231-L338). I think to test this we probably want to (for just the training part of the code that I linked) construct a model that has deterministic loss (based on num padding tokens maybe?) and then test that the results are different in the expected way between sample weighting and token weighting.

dakinggg:
So basically make a trainer with a dummy model and a dummy dataset, and then call it with sample weighting and token weighting (with microbatching on), and assert the losses are different in the expected way.

aadyotb:
@dakinggg I've added a unit test that requires sample-based and token-based weighting result in different outcomes when padding is present.

dakinggg:
@aadyotb awesome, thank you!! Will take a look soon.

tests/test_simple_nlp.py
```diff
@@ -125,3 +126,67 @@ def test_simple_nlp_mlm(tiny_bert_tokenizer, tiny_bert_model):
...
+    # Set up a trainer that accumulates train loss based on token counts, after reloading original state dict
+    model.load_state_dict(state_dict)
+    token_data_spec = DataSpec(
+        dataloader=train_dataloader,
+        get_num_tokens_in_batch=lambda b: (b['input_ids'] != tiny_bert_tokenizer.pad_token_id).sum().item(),
+    )
+    token_trainer = Trainer(
+        model=model,
+        train_dataloader=token_data_spec,
+        accumulate_train_batch_on_tokens=False,
```
dakinggg, line 182: I'm guessing you meant to make this True? And if this test passed as easy I'm guessing we aren't being quite strict enough here and the CEs are randomly different. Trying to think if there is any way we can test more directly than just seeing if the CE is different...

aadyotb: Thanks for catching this. I think the problem is that I'm not controlling the randomness in the data collator. To work around this problem, let me
1. Proactively initialize the training dataset
2. Add `reproducibility.seed_all(42)` before each call to `trainer.fit()`
3. Add an extra sanity check that I can reproduce the CE calculation from the original regular trainer

mvpatel2000:
@aadyotb  Just a heads up Daniel is out this week, and given the subtlety here, I would prefer he finish the review of this PR vs. bringing someone else to review. Please let me know if you need it earlier -- sorry for the delay!

aadyotb:
Thanks @mvpatel2000. For the time being, I've implemented these changes by overriding the `Trainer` class in our local repo, so we will be okay for now. Happy to get further review once Daniel returns.

aadyotb:
@dakinggg bumping this PR for review.

composer/trainer/trainer.py
```diff
@@ -2848,7 +2852,20 @@ def _train_microbatches(
...
+                if current_batch_size == 0:
+                    warnings.warn(
+                        'Requested loss accumulation based on number of tokens in training batch, but zero tokens '
+                        'found (perhaps due to an improper DataSpec).',
+                    )
```
mvpatel2000, line 2861: @dakinggg  shoudl this be an error instead? eg u get all pad tokens? Maybe you need to do this if your dataset is uneven size and you don't want to `drop_last` but using FSDP so all ranks participate?

mvpatel2000 (commented):
LGTM but requiring sign off from @dakinggg  as well

tests/test_simple_nlp.py
```diff
@@ -125,3 +127,92 @@ def test_simple_nlp_mlm(tiny_bert_tokenizer, tiny_bert_model):
...
+
+    train_dataset = RandomTextLMDataset(
+        size=size,
+        vocab_size=vocab_size,
+        sequence_length=sequence_length,
+        use_keys=True,
+        pad_token_id=tiny_bert_tokenizer.pad_token_id,
+    )
+    for i in range(size):  # Proactively load dataset
+        train_dataset[i]
```
mvpatel2000, line 149: why is this necessary? Can we remove?

tests/test_simple_nlp.py
```diff
@@ -125,3 +127,92 @@ def test_simple_nlp_mlm(tiny_bert_tokenizer, tiny_bert_model):
...
+
+    trainer = Trainer(
+        model=model,
+        train_dataloader=data_spec,
+        max_duration='2ep',
+        device_train_microbatch_size=batch_size // 2,
+        accumulate_train_batch_on_tokens=False,
+        device=device,
+    )
+    reproducibility.seed_all(42)
```
mvpatel2000, line 176: compsoer should seed if you pass to it instead

tests/test_simple_nlp.py
```diff
@@ -125,3 +127,92 @@ def test_simple_nlp_mlm(tiny_bert_tokenizer, tiny_bert_model):
...
+    model.load_state_dict(state_dict)
+    token_trainer = Trainer(
+        model=model,
+        train_dataloader=data_spec,
+        max_duration='2ep',
+        device_train_microbatch_size=batch_size // 2,
+        accumulate_train_batch_on_tokens=True,
+        device=device,
+    )
+    reproducibility.seed_all(42)
```
mvpatel2000, line 194: same comment on seeding

tests/test_simple_nlp.py
```diff
@@ -125,3 +127,92 @@ def test_simple_nlp_mlm(tiny_bert_tokenizer, tiny_bert_model):
...
+    model.load_state_dict(state_dict)
+    trainer2 = Trainer(
+        model=model,
+        train_dataloader=data_spec,
+        max_duration='2ep',
+        device_train_microbatch_size=batch_size // 2,
+        accumulate_train_batch_on_tokens=False,
+        device=device,
+    )
+    reproducibility.seed_all(42)
```
mvpatel2000, line 215: same comment on seeding

mvpatel2000 (commented):
will let daniel review the unit test

tests/test_simple_nlp.py
aadyotb: Fixed.

tests/test_simple_nlp.py
aadyotb: This is necessary because the dataset only materializes its random tensors the first time they are invoked. So the first call to the trainer will initialize the dataset and result in a differently advanced random seed for e.g. random MLM masking.

dakinggg:
Hey @aadyotb taking a look now, mostly convincing myself that the math is correct :)

composer/trainer/trainer.py
```diff
@@ -2848,7 +2852,20 @@ def _train_microbatches(
...
+            else:
+                current_batch_size = sum([self._train_data_spec.get_num_samples_in_batch(b) for b in microbatches])
+            # Average the current batch size across ranks, to ensure each rank contributes appropriately
```
dakinggg, line 2864: Ok I think this is not quite correct. If we want to weight each token equally, then we want to scale the microbatch loss as
```
microbatch_loss = microbatch_loss * (microbatch_num_tokens/global_batch_size_tokens)*world_size
```

The current implementation won't be exactly this because its just scaling by the average batch size in tokens. 

Also I would prefer that we leave the scaling code the same for the original samples case, since there is no reason to adjust it.

Let me know if this makes sense or what I am missing....

aadyotb: I'm pretty sure that the two are equivalent: we can write
```
  microbatch_loss * (microbatch_num_tokens / global_batch_size) * world_size
= microbatch_loss * microbatch_num_tokens / (global_batch_size / world_size)
= microbatch_loss * microbatch_num_tokens / all_reduce(current_batch_size, "AVG")
```
I'd be happy to revert for the original samples case, though I do think the updated version is more correct in the event that different ranks have different numbers of samples.

composer/trainer/trainer.py
```diff
@@ -946,6 +946,8 @@ class Trainer:
             .. note:: This is implemented by taking the batch yielded by the ``train_dataloader`` and splitting
                 it into sections of size ``device_train_microbatch_size``. If the batch size of the dataloader
                 is not divisible by ``device_train_microbatch_size``, the last section will be potentially smaller.
+        accumulate_train_batch_on_tokens (bool, optional): Whether training loss is accumulated over the number of tokens in
```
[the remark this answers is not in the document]
aadyotb: Done.

composer/trainer/trainer.py
aadyotb: Happy to emit an exception instead. I chose a warning initially based on https://github.com/mosaicml/composer/pull/3273#discussion_r1600672811

composer/trainer/trainer.py
dakinggg: Ah sorry, I didn't put enough parens, trying again.
```
microbatch_loss = (microbatch_loss * (microbatch_num_tokens/global_batch_size_tokens))*world_size
```

composer/trainer/trainer.py
dakinggg: exception seems reasonable to me

composer/trainer/trainer.py
aadyotb: Not sure if I'm missing something, but does that make a difference? Since it's all multiplications and divisions. The current version is `microbatch_loss * local_batch_size`, which is equivalent only if `local_batch_size` is the same on all all ranks.

For what it's worth, I do find that the version in the code here gives the same result as a `torchmetrics.Metric` I implemented (in our own repo using `composer`) which accumulates the number of tokens and the loss separately, before sum-reducing both and dividing.

composer/trainer/trainer.py
aadyotb: Updated accordingly.

composer/trainer/trainer.py
dakinggg: Ah sorry, found my error, you are correct. @mvpatel2000 what do you think about changing the behavior for the sample weighted case? I agree it is more correct, but its also a pretty subtle change in behavior.

dakinggg (commented):
Other than waiting for Mihir's thought on changing the behavior for the sample weighting case, looks good to me! Thanks so much for the contribution!

composer/trainer/trainer.py
mvpatel2000: I am for changing. It's more correct -- the distributed case should match the non distributed case, and this fixes it in the scenario for uneven samples.

mvpatel2000 (approved):
Approving

dakinggg (approved):

dakinggg:
Ok cool, LGTM. I'm running a before and after loss curve just to double check (in the normal case, with even number of samples per device batch), and will post that graph here when done.

dakinggg:
Screenshot 2024-06-05 at 10 58 34 PM

For posterity, I validated that finetuning behavior is unchanged before and after this PR (for a case with constant samples per device batch), and does change if you specify the new flag.

dakinggg:
@aadyotb I think one more run of precommit should do it, and we should be good to merge.

PR #3275, opened 2024-05-10
Add option to stagger uploads based on local rank

dakinggg:
This PR adds support for adding a stagger based on local rank to the upload workers (i.e. the upload worker will wait by local rank * stagger). We are only exposing this via environment variable, as this is not something most users should be setting themselves.

Manual test:
- `stagger-upload-test-2-IggDEG` (note that each ranks file was uploaded 1 minute apart)
Screenshot 2024-05-10 at 12 16 40 PM

### DIFF
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -676,7 +676,7 @@ def _upload_worker(
         # the concurrency by a factor of num GPUs per node.
         local_rank = dist.get_local_rank()
         local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
-        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank{local_rank}.')
+        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank {local_rank}.')
         time.sleep(local_rank * local_rank_stagger)
 
         # defining as a function-in-function to use decorator notation with num_attempts as an argument
--- a/composer/loggers/remote_uploader_downloader.py
+++ b/composer/loggers/remote_uploader_downloader.py
@@ -660,11 +660,6 @@ def _upload_worker(
     The worker will continuously poll ``file_queue`` for files to upload. Once ``is_finished`` is set, the worker will
     exit once ``file_queue`` is empty.
     """
-    from composer.utils import dist
-    # Stagger uploads by one minute per local rank
-    stagger_wait = dist.get_local_rank() * 60
-    time.sleep(stagger_wait)
-
     remote_backend = _build_remote_backend(remote_backend_name, backend_kwargs)
     while True:
         try:
@@ -675,6 +670,11 @@ def _upload_worker(
             else:
                 continue
         uri = remote_backend.get_uri(remote_file_name)
+
+        if remote_file_name.endswith('.distcp'):
+            filename = remote_file_name.split('/')[-1]
+            rank = filename.split('_')[-2]
+            time.sleep(int(rank % 8) * 60)
 
         # defining as a function-in-function to use decorator notation with num_attempts as an argument
         @retry(ObjectStoreTransientError, num_attempts=num_attempts)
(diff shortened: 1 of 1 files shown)


composer/loggers/remote_uploader_downloader.py
```diff
@@ -671,6 +671,10 @@ def _upload_worker(
                 continue
         uri = remote_backend.get_uri(remote_file_name)
 
+        local_rank = dist.get_local_rank()
```
irenedea, line 674: Please add a comment to describe why we are doing this/ why someone may want to set this

irenedea (approved):

composer/loggers/remote_uploader_downloader.py
```diff
@@ -671,6 +671,10 @@ def _upload_worker(
...
+        local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
```
irenedea, line 675: Can you also add a debug log here

composer/loggers/remote_uploader_downloader.py
dakinggg: done

composer/loggers/remote_uploader_downloader.py
dakinggg: done (although i think this might get swallowed by the subprocess)

composer/loggers/remote_uploader_downloader.py
```diff
@@ -671,6 +671,14 @@ def _upload_worker(
                 continue
         uri = remote_backend.get_uri(remote_file_name)
 
+        # When encountering issues with too much concurrency in uploads, staggering the uploads can help.
+        # This stagger is intended for use when uploading model shards from every rank, and will effectively reduce
+        # the concurrency by a factor of num GPUs per node.
+        local_rank = dist.get_local_rank()
+        local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
+        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on {local_rank} local rank.')
```
dakinggg, line 679:
```suggestion
        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank{local_rank}.')
```

composer/loggers/remote_uploader_downloader.py
```diff
@@ -671,6 +671,14 @@ def _upload_worker(
                 continue
         uri = remote_backend.get_uri(remote_file_name)
 
+        # When encountering issues with too much concurrency in uploads, staggering the uploads can help.
+        # This stagger is intended for use when uploading model shards from every rank, and will effectively reduce
+        # the concurrency by a factor of num GPUs per node.
+        local_rank = dist.get_local_rank()
+        local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
+        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank{local_rank}.')
```
dakinggg, line 679:
```suggestion
        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank {local_rank}.')
```

composer/loggers/remote_uploader_downloader.py
```diff
@@ -671,6 +671,14 @@ def _upload_worker(
                 continue
         uri = remote_backend.get_uri(remote_file_name)
 
+        # When encountering issues with too much concurrency in uploads, staggering the uploads can help.
+        # This stagger is intended for use when uploading model shards from every rank, and will effectively reduce
+        # the concurrency by a factor of num GPUs per node.
+        local_rank = dist.get_local_rank()
+        local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
+        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on local rank {local_rank}.')
+        time.sleep(local_rank * local_rank_stagger)
```
eracah, line 680: nit: can you actually put this logic after the function definition right before `upload_file`. It will make it a but more readable I think

eracah (approved):

composer/loggers/remote_uploader_downloader.py
dakinggg: sure, done

PR #3284, opened 2024-05-13
Fix UCObjectStore.list_objects

dakinggg:
# What does this PR do?
Previously, the databricks SDK did not have `ls` built in, and we discovered a bug where we were not properly handling the truncation of the `ls` results (truncated to 999 elements). This PR fixes that bug by switching over to the built in `ls` in the databricks SDK.

Manual test, ran the convert to MDS script before and after, before it returns 999 elements, after it returns the full correct number (1350 from the directory I was testing)

### DIFF
--- a/composer/utils/object_store/uc_object_store.py
+++ b/composer/utils/object_store/uc_object_store.py
@@ -242,7 +242,6 @@ class UCObjectStore(ObjectStore):
         from databricks.sdk.core import DatabricksError
         try:
             ls_results = [entry.path for entry in self.client.files.list_directory_contents(os.path.join('/', prefix))]
-            print(len(ls_results))
             return ls_results
 
         except DatabricksError as e:
--- a/composer/utils/object_store/uc_object_store.py
+++ b/composer/utils/object_store/uc_object_store.py
@@ -241,7 +241,8 @@ class UCObjectStore(ObjectStore):
 
         from databricks.sdk.core import DatabricksError
         try:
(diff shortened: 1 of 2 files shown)


composer/utils/object_store/uc_object_store.py
```diff
@@ -241,35 +244,28 @@ def list_objects(self, prefix: Optional[str]) -> List[str]:
...
-                    fpath = f['path']
-                    if f['is_dir']:
-                        stack.append(fpath)
+                    if is_directory:
+                        stack.append(path)
                     else:
-                        all_files.append(fpath)
+                        all_files.append(path)
 
+            print(len(all_files))
```
dakinggg, line 268:
```suggestion
```

composer/utils/object_store/uc_object_store.py
```diff
@@ -96,16 +95,20 @@ def validate_path(path: str) -> str:
...
         return os.path.join(*dirs[:4])
 
-    def _get_object_path(self, object_name: str) -> str:
+    def _get_object_path(self, object_name: Optional[str] = None) -> str:
         """Return the absolute Single Path Namespace for the given object_name.
 
         Args:
-            object_name (str): Absolute or relative path of the object w.r.t. the
+            object_name (optional, str): Absolute or relative path of the object w.r.t. the
             UC Volumes root.
```
irenedea, line 103:
```suggestion
            UC Volumes root. If None, the prefix path is returned.
```

irenedea (approved):
lgtm! thanks for the fix!

PR #3300, opened 2024-05-17
Skip compression check with symlink files

mvpatel2000:
# What does this PR do?

Skip compression check with symlink files. Add unit tests.

Fixes an issue where some previous checkpoints where `.pt.symlink`

### DIFF
--- a/composer/utils/compression.py
+++ b/composer/utils/compression.py
@@ -22,7 +22,7 @@ def is_compressed_pt(filename: str) -> bool:
     of a single pt file without a container (like tar).
     """
     parts = filename.split('.')
-    return len(parts) >= 2 and parts[-2] == 'pt'
+    return len(parts) >= 2 and parts[-2] == 'pt' and parts[-1] != 'symlink'
 
 
 class CliCompressor:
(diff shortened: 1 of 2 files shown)


dakinggg (approved):

eracah (commented):
Wow that is a flimsy check. If there are two dots in the filename?! There's gotta be a more robust way

mvpatel2000:
> Wow that is a flimsy check. If there are two dots in the filename?! There's gotta be a more robust way

I agree it's not ideal... maybe it should be an explicit whitelist of compressors. This wouldn't raise an error if you try a compressor that doesn't exist, but it would increase robustness. I'm inclined to leave it as is and if it breaks again we can change?

eracah:
> > Wow that is a flimsy check. If there are two dots in the filename?! There's gotta be a more robust way
> 
> I agree it's not ideal... maybe it should be an explicit whitelist of compressors. This wouldn't raise an error if you try a compressor that doesn't exist, but it would increase robustness. I'm inclined to leave it as is and if it breaks again we can change?

Ok lemme add a JIRA

PR #3303, opened 2024-05-17
[checkpoint v2] add remote uploader class

bigning:
# What does this PR do?

Add a RemoteUploader class
Compared to RemoteUploaderDownloader, it 
1. removes the download API, which is not used
2. uses modern ProcessPoolExecutor which provides high-level interface for asynchronously executing callables
3. greatly simplify the logic, it doesn't use any queues. 
4. it's neither a logger nor a callback, so it's very light weight and standalone component. Instead, it's caller's responsibility to wait the upload to finish
5. much easier to maintain and add more features, since it's simpler

# test
## unit test
pytest tests/utils/test_remote_uploader.py
## e2e test 
see PR https://github.com/mosaicml/composer/pull/3320


eracah:
> To avoid race conditions with symlink and checkpoints, we do `num_workers=1`. Should we just remove multi worker functionality (remove public arg)?
> 
> Otherwise, @eracah does the redesign have a way to fix this?

it will fix this because we will upload the symlink synchronously after the file(s) in the upload_worker

composer/utils/remote_uploader.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # A folder to use for staging uploads
+        self._tempdir = tempfile.TemporaryDirectory()
+        self._upload_staging_folder = self._tempdir.name
+
+        self.num_attempts = num_attempts
+
+        mp_ctx = multiprocessing.get_context('spawn')
+        manager = mp_ctx.Manager()
+        self.file_queue: queue.Queue = manager.Queue()
```
eracah, line 115: why do we need a filequeue? I think instead of the design where we launch the workers once and then continually poll the file queue, it's even simpler if we just call executor.submit every time we have a new upload request

composer/utils/remote_uploader.py
```diff
@@ -0,0 +1,172 @@
...
+            self.workers.append(
+                self.executor.submit(
+                    _upload_worker,
+                    remote_folder=remote_folder,
+                    file_queue=self.file_queue,
+                    is_closed_event=self.is_closed_event,
+                    num_attempts=num_attempts,
+                    parent_process_id=os.getpid(),
+                ),
+            )
```
eracah, line 132: The constructor should just set up of the executor and then each `upload_file_async` call would call an `executor.submit`. This way we can avoid needing a filequeue or an is_closed_event
```suggestion
        self.executor = ProcessPoolExecutor(
            max_workers=num_concurrent_uploads
        )
        self.workers: List[Future] = []
```

composer/utils/remote_uploader.py
```diff
@@ -0,0 +1,172 @@
...
+        file_path: pathlib.Path,
+        overwrite: bool,
+    ):
+        """Async call to submit a job for uploading."""
+        # Copy file to staging folder
+        copied_path = os.path.join(self._upload_staging_folder, str(uuid.uuid4()))
+        os.makedirs(self._upload_staging_folder, exist_ok=True)
+        shutil.copy2(file_path, copied_path)
+
+        self.file_queue.put_nowait((copied_path, remote_file_name, overwrite))
```
eracah, line 146:
```suggestion
    def upload_file_async(
        self,
        remote_file_name: str,
        file_path: pathlib.Path,
        overwrite: bool,
    ):
        """Async call to submit a job for uploading."""
        # Copy file to staging folder
        copied_path = os.path.join(self._upload_staging_folder, str(uuid.uuid4()))
        os.makedirs(self._upload_staging_folder, exist_ok=True)
        shutil.copy2(file_path, copied_path)

        future = self.executor.submit( _upload_worker,
                    remote_folder=remote_folder,
                    num_attempts=num_attempts,
                    parent_process_id=os.getpid(),)
        self.workers.append(future)
```

composer/utils/remote_uploader.py
```diff
@@ -0,0 +1,172 @@
...
+        try:
+            file_path_to_upload, remote_file_name, overwrite = file_queue.get(block=True, timeout=0.5)
+        except queue.Empty:
+            if is_closed_event.is_set():
+                break
+            elif is_parent_process_dead():
+                log.warning(f'Terminate the uploader worker since parent process died')
+                break
+            else:
+                continue
```
eracah, line 82: No need to have this logic if we don't have a file queue

eracah (changes requested):
This is a good start, @bigning ! I think it could be a lot simpler if we move to a design where on each `upload_file_async` call a new executor.submit call is called. This would make it even simpler because we wouldn't need a file queue or an `is_closed_event`. See the design doc implementation section for more details

composer/utils/remote_uploader.py
```diff
@@ -0,0 +1,172 @@
...
+        # When encountering issues with too much concurrency in uploads, staggering the uploads can help.
+        # This stagger is intended for use when uploading model shards from every rank, and will effectively reduce
+        # the concurrency by a factor of num GPUs per node.
+        local_rank = get_local_rank()
+        local_rank_stagger = int(os.environ.get('COMPOSER_LOCAL_RANK_STAGGER_SECONDS', 0))
+        log.debug(f'Staggering uploads by {local_rank * local_rank_stagger} seconds on {local_rank} local rank.')
+        time.sleep(local_rank * local_rank_stagger)
+
+        upload_file(remote_file_name, file_path_to_upload, overwrite)
+
```
eracah, line 93: we also want to the worker to upload the symlink if specified

eracah:
> > > To avoid race conditions with symlink and checkpoints, we do `num_workers=1`. Should we just remove multi worker functionality (remove public arg)?
> > > Otherwise, @eracah does the redesign have a way to fix this?
> > 
> > 
> > it will fix this because we will upload the symlink synchronously after the file(s) in the upload_worker
> 
> This version just does exact same thing as existing remote_uploader_downloader.

Right, but in the design doc, I designed a feature where the symlink gets uploaded synchronously after the file. Is that intended for a follow-up PR?

bigning:
> This is a good start, @bigning ! I think it could be a lot simpler if we move to a design where on each `upload_file_async` call a new executor.submit call is called. This would make it even simpler because we wouldn't need a file queue or an `is_closed_event`. See the design doc implementation section for more details

@eracah , that's actually my initial implementation ([look at this](https://github.com/mosaicml/composer/blob/d57fb67878b7ddf193b1c1ae31cfe4bed92bbc2f/composer/utils/remote_uploader.py#L113-L115)), but it doesn't work unless we create a new object_store everytime we call upload_file. This is because, with mutli-process, it needs to pickle the input to the function, the object_store is not pickable.  Does that make sense ?

composer/utils/remote_uploader.py
bigning: See my another comment. This is because we can't pass object_store to the function that will be executed in a new process. Because you can't pickle an object store. I think that's why the existing remote_uploader_downloader used queues. :)

composer/utils/remote_uploader.py
bigning: this way we need to create an object_store for every file we need to upload, i think we don't want to do that?

composer/utils/remote_uploader.py
bigning: The remote uploader is the basic class to upload any file. The symlink file upload is in the checkpoint_saver in this PR, https://github.com/mosaicml/composer/pull/3320

b-chu:
I'll leave the reviewing to others but adding myself to cc

composer/utils/remote_uploader.py
eracah: I think it's fine to create an object_store for every file. the object store class is pretty lightweight

eracah: the time to create an object store is negligible compared to the time it takes to actually upload. Also, if creating an object store every time we retry (which maybe a lot of the time we don't even retry) is all we need to do to drastically simplify the remoteuploader, then it seems worth it.

bigning:
> > > > To avoid race conditions with symlink and checkpoints, we do `num_workers=1`. Should we just remove multi worker functionality (remove public arg)?
> > > > Otherwise, @eracah does the redesign have a way to fix this?
> > > 
> > > 
> > > it will fix this because we will upload the symlink synchronously after the file(s) in the upload_worker
> > 
> > 
> > This version just does exact same thing as existing remote_uploader_downloader.
> 
> Right, but in the design doc, I designed a feature where the symlink gets uploaded synchronously after the file. Is that intended for a follow-up PR?

discussed offline, this is a separate PR to add the new checkpoint saver

eracah (approved):
LGTM!

PR #3313, opened 2024-05-22
Fix HSDP with ShardDegree < 8 

bigning:
# what
@j316chuck found HSDP with 1 node 2X4 device mesh run can't resume. rank 0 failed [this checkpoint files broadcast](https://github.com/mosaicml/composer/blob/dev/composer/utils/checkpoint.py#L314-L318) . It's because rank 4 (which is in the same replica pg group as rank 0) doesn't join this collective call because of [this if condition](https://github.com/mosaicml/composer/blob/dev/composer/utils/checkpoint.py#L307). 
So the fix is to let all the first rank of shard pg group joins the broadcast , but only local rank 0 writes to file

# test

##  1 node device mesh: 2 X 4
before:
[nccl timeout] mpt-125m-hybrid-resumption-1-node-tYv2bB
after:
 mpt-125m-hybrid-resumption-1-node-pzWqX1
mpt-125m-hybrid-resumption-1-node-RV4ovO

## 2 node device mesh: 2 X 8 still works
mpt-125m-hybrid-resumption-1-node-NlOtP5

## 2 node device mesh: 4 x 4 works:
mpt-125m-hybrid-resumption-1-node-ajvSHb
checked the logging, rank 8 is not downloading any checkpoint files, but it got all files broadcast from rank 0
image

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -290,7 +290,7 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
         if self.device_mesh is not None and self.device_mesh.ndim == 2:
             # Broadcast file to all replicas
             replicate_process_group = self.device_mesh.get_group(0)
-            hsdp_shard_process_group = self.device_mesh.get_group(1)
+            shard_process_group = self.device_mesh.get_group(1)
             shard_size = self.device_mesh.size(1)
             rank_in_first_replica = dist.get_global_rank() % shard_size
             sender = dist.get_global_rank() == rank_in_first_replica
@@ -307,7 +307,7 @@ class DistCPObjectStoreReader(FileSystemReaderWithValidation):
             # Send each file to the appropriate rank
             for file_name in file_list:
                 if dist.get_local_rank() == 0 or (
-                    torch_dist.get_rank(hsdp_shard_process_group) == 0
+                    torch_dist.get_rank(shard_process_group) == 0
                 ):  # Only 1 rank per node needs to transfer file
                     full_path = os.path.join(self.destination_path, file_name)
                     log.debug(f'Transferring {full_path=}')
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
+import torch.distributed as torch_dist
 from packaging import version
 from torch.distributed import checkpoint as dist_cp
 from torch.distributed._tensor import DeviceMesh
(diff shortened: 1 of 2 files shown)


j316chuck (approved):
Tested with : `mpt-125m-hybrid-resumption-1-node-7QStEE`

composer/utils/checkpoint.py
```diff
@@ -289,6 +290,7 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
         if self.device_mesh is not None and self.device_mesh.ndim == 2:
             # Broadcast file to all replicas
             replicate_process_group = self.device_mesh.get_group(0)
+            hsdp_shard_process_group = self.device_mesh.get_group(1)
```
bigning, line 293: @mvpatel2000 , maybe in your TP PR, since you added the mesh name constants, maybe change this to use the name to select the replicate and shard PGs, instead of using index ?

composer/utils/checkpoint.py
```diff
@@ -318,7 +322,7 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
                     )
                     received_file_object = file_object[0]
                     assert received_file_object is not None
-                    if receiver and not os.path.exists(full_path):
+                    if receiver and not os.path.exists(full_path) and dist.get_local_rank() == 0:
```
mvpatel2000, line 325: is `os.path.exists` not a sufficient check for multi GPU write? or would it race?

composer/utils/checkpoint.py
mvpatel2000, line 293:
```suggestion
            shard_process_group = self.device_mesh.get_group(1)
```
replicate group does not say hsdp, to be consistent

composer/utils/checkpoint.py
bigning: Not 100% sure if it would race, but adding this 0 rank condition to be more safe.

composer/utils/checkpoint.py
```diff
@@ -304,7 +306,9 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
 
             # Send each file to the appropriate rank
             for file_name in file_list:
-                if dist.get_local_rank() == 0:  # Only 1 rank per node needs to transfer file
+                if dist.get_local_rank() == 0 or (
+                    torch_dist.get_rank(shard_process_group) == 0
```
mvpatel2000, line 310: @bigning can we instead update our `get_global_rank` function to accept a process group and forward to `get_rank`?

mvpatel2000 (commented):
LGTM! sorry one more request before approval

composer/utils/checkpoint.py
bigning: updated!

mvpatel2000 (approved):
Big Ning stays WinNing

eracah (commented):
Nice! But why would anyone use HSDP with only one node?

bigning:
> Nice! But why would anyone use HSDP with only one node?

I guess at least unit test needs it ? lol

PR #3315, opened 2024-05-23
Improving memory snapshot 

cli99:
# What does this PR do?

This PR uses `_record_memory_history_impl` instead of `_record_memory_history_legacy` (previous) to capture memory snapshot. See https://github.com/pytorch/pytorch/blob/main/torch/cuda/memory.py#L698-L738. With `enabled =all`, this captures all (c++ and python) alloc/free events and gives better memory timeline and stack trace information. 

`_record_memory_history_impl` exists for pytorch 2.1, 2.2, 2.0, no version gating is needed. 

image


cli99:
> Can you check this function exists on torch 2.1/2.2/2.3? If it does't, we would need to gate on torch function.

Checked. The function exists on torch 2.1/2.2/2.3

PR #3316, opened 2024-05-23
Add checkpoint events to mosaicml logger

b-chu:
# What does this PR do?

Adds metadata logging to mosaicml logger for when checkpoint upload starts when a checkpoint is done loading. The information logged is the timestamp and the training duration of the checkpoint.

Tested in `events-test-8ReCjx`

# Before submitting

### DIFF
--- a/composer/loggers/mosaicml_logger.py
+++ b/composer/loggers/mosaicml_logger.py
@@ -92,30 +92,30 @@ class MosaicMLLogger(LoggerDestination):
                 self._enabled = False
 
     def log_hyperparameters(self, hyperparameters: Dict[str, Any]):
-        self._log_metadata(hyperparameters)
+        self.log_metadata(hyperparameters)
 
     def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
-        self._log_metadata(metrics)
+        self.log_metadata(metrics)
 
     def log_exception(self, exception: Exception):
-        self._log_metadata({'exception': exception_to_json_serializable_dict(exception)})
+        self.log_metadata({'exception': exception_to_json_serializable_dict(exception)})
         self._flush_metadata(force_flush=True)
 
     def after_load(self, state: State, logger: Logger) -> None:
         # Log model data downloaded and initialized for run events
         log.debug(f'Logging model initialized time to metadata')
-        self._log_metadata({'model_initialized_time': time.time()})
+        self.log_metadata({'model_initialized_time': time.time()})
         # Log WandB run URL if it exists. Must run on after_load as WandB is setup on event init
         for callback in state.callbacks:
             if isinstance(callback, WandBLogger):
                 run_url = callback.run_url
                 if run_url is not None:
-                    self._log_metadata({'wandb/run_url': run_url})
+                    self.log_metadata({'wandb/run_url': run_url})
                     log.debug(f'Logging WandB run URL to metadata: {run_url}')
                 else:
                     log.debug('WandB run URL not found, not logging to metadata')
             if isinstance(callback, MLFlowLogger) and callback._enabled:
-                self._log_metadata({'mlflow/run_url': callback.run_url})
+                self.log_metadata({'mlflow/run_url': callback.run_url})
                 log.debug(f'Logging MLFlow run URL to metadata: {callback.run_url}')
         self._flush_metadata(force_flush=True)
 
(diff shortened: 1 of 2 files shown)


composer/loggers/mosaicml_logger.py
```diff
@@ -150,6 +150,10 @@ def close(self, state: State, logger: Logger) -> None:
         if self._enabled:
             wait(self._futures)  # Ignore raised errors on close
 
+    def log_checkpoint_upload_start(self, state: State) -> None:
+        self._log_metadata({'checkpoint_upload_start_time': [time.time(), state.timestamp.batch.to_timestring()]})
```
waiwuc, line 154: looks fine, would definitely double check if json is created from iterable properly 

https://github.com/mosaicml/composer/blob/4e1eb8d20a904510c2c03819f4517c1f12fe7b3a/composer/loggers/mosaicml_logger.py#L262

waiwuc: i guess theres a test case already 

https://github.com/mosaicml/composer/blob/4e1eb8d20a904510c2c03819f4517c1f12fe7b3a/tests/loggers/test_mosaicml_logger.py#L71 

Good that this change is so simple :) Dont think we use the checkpoint load event anywhere yet in mcloud, so ok to change to list. But @jjanezhang  should confirm here

waiwuc (approved):

composer/loggers/mosaicml_logger.py
```diff
@@ -104,7 +104,7 @@ def log_exception(self, exception: Exception):
     def after_load(self, state: State, logger: Logger) -> None:
         # Log model data downloaded and initialized for run events
         log.debug(f'Logging model initialized time to metadata')
-        self._log_metadata({'model_initialized_time': time.time()})
+        self._log_metadata({'model_initialized_time': [time.time(), state.timestamp.batch.to_timestring()]})
```
mvpatel2000, line 107: does this work as a list? woah

composer/loggers/mosaicml_logger.py
mvpatel2000, line 154: I don't really like explicitly hard coding a key to mosaicml_logger. Why not just call log_metadata and then flush at end user?

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +545,8 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
             else:
                 if dist.get_global_rank() == 0:
                     shutil.rmtree(prefix_dir)
+
+    def _log_to_mosaicml_logger(self, state: State, logger: Logger):
+        for destination in logger.destinations:
+            if isinstance(destination, MosaicMLLogger):
+                destination.log_checkpoint_upload_start(state)
```
mvpatel2000, line 552: Can we instead log the information directly and make the name of this fn say its only logging save time?

mvpatel2000 (changes requested):
Holding based on discussions

composer/loggers/remote_uploader_downloader.py
```diff
@@ -707,3 +708,7 @@ def upload_file(retry_index: int = 0):
         time.sleep(local_rank * local_rank_stagger)
 
         upload_file()
+
+    for destination in logger.destinations:
+        if isinstance(destination, MosaicMLLogger):
+            destination.log_metadata({'checkpoint_uploaded_time': time.time()})
```
mvpatel2000, line 714: should this have a force flush flag?

mvpatel2000 (commented):
can you include a manual test to verify this works since I don't htink we can easily unit test this?

composer/loggers/remote_uploader_downloader.py
dakinggg, line 714: is this tracking what you are actually trying to track? It will simply overwrite every time a new file is uploaded.

mvpatel2000: we only want last one to measure goodput

b-chu: log_metadata always flushes at the end
https://github.com/mosaicml/composer/blob/4e1eb8d20a904510c2c03819f4517c1f12fe7b3a/composer/loggers/mosaicml_logger.py#L164

mvpatel2000 (commented):
You need to force flush!

mvpatel2000 (approved):

PR #3320, opened 2024-05-24
[WIP] [checkpoint v2] Use remote uploader v2 for checkpointing

bigning:
# what?

Composer change to use the new RemoteUploader for saving checkpoing

# Test
llm-foundry PR: https://github.com/mosaicml/llm-foundry/pull/1237
## 1-node small model OCI test
test-uploader-wfA8Vz
## 1-node small model S3 test
save: test-uploader-10udLt e2e time: 3 min 40s
resume: test-uploader-TDGgTs

use old remote_uploader_downloader `test-uploader-racGuC` e2e time: 3 min 42s

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -29,6 +29,8 @@ from composer.utils import (
     format_name_with_dist_and_time,
     is_model_deepspeed,
     partial_format,
+    RemoteUploader,
+    parse_uri,
 )
 from composer.utils.compression import get_compressor, is_compressed_pt
 from composer.utils.object_store.mlflow_object_store import MLFLOW_EXPERIMENT_ID_FORMAT_KEY, MLFLOW_RUN_ID_FORMAT_KEY
@@ -288,6 +290,9 @@ class CheckpointSaver(Callback):  # noqa: D101
         num_checkpoints_to_keep: int = -1,
         weights_only: bool = False,
         ignore_keys: Optional[Union[List[str], Callable[[Dict], None]]] = None,
+        # RemoteUploaderV2 params
+        use_remote_uploader_v2: bool = False,
+        save_folder: Optional[str] = None,
     ):
         folder = str(folder)
         filename = str(filename)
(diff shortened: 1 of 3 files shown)


composer/checkpoint/state_dict.py
```diff
@@ -152,3 +154,94 @@ def _get_model_state_dict_with_fsdp_context_manager(model: nn.Module, sharded_st
...
         model_state_dict = model.state_dict()
     return model_state_dict
+
+
+def get_metadata_state_dict(
+    model: Optional[Union[ComposerModel, nn.Module]] = None,
+    sharded_state_dict: Optional[bool] = None,
+    precision: Optional[Union[str, torch.dtype]] = None,
+    device: Optional[Device] = None,
+    device_train_microbatch_size: Optional[int] = None,
```
mvpatel2000, line 164: Is this copied from Evan's PR?

mvpatel2000 (commented):
Why do we default to v1 still?

composer/checkpoint/state_dict.py
bigning: oh it got messed up. will clean up

composer/callbacks/checkpoint_saver.py
```diff
@@ -520,11 +549,18 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-                        logger.upload_file(
-                            remote_file_name=symlink_name,
-                            file_path=symlink_filename,
-                            overwrite=True,
-                        )
+                        if self.use_remote_uploader_v2 and self.remote_uploader is not None:
+                            self.remote_uploader.upload_file_async(
+                                remote_file_name=symlink_name,
+                                file_path=symlink_filename,
+                                overwrite=True,
```
eracah, line 556: ok this makes sense for a PR, where we just simplify the remoteuploader and don't change any functionality, but eventually we want to have the remoteuploader manage symlinks. That way we can actually avoid race conditions

bigning:
Discussed with @eracah , we're going to add a new checkpoint saver, instead of changing current one.

PR #3323, opened 2024-05-24
Bugfixes to FSDP + TP

mvpatel2000:
# What does this PR do?

Fixes:
- [x] Fixes checkpointing issue with HSDP + TP where the right process group might not correctly be sliced (as previously, HSDP dimension was hardcoded)
- [x] Fixes torch <2.3 support, which does not support default. Now, it does not autocreate device mesh

Save: `0-16gpu-save-NTsI27`
Load: `0-16gpu-save-AVhYVg`

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -200,6 +200,10 @@ def _create_device_mesh(
     fsdp_config: Optional[Dict[str, Any]],
     tp_config: Optional[Dict[str, Any]],
 ) -> Optional[DeviceMesh]:
+    if version.parse(torch.__version__.split('.dev')[0]) < version.parse('2.3.0'):
+        # Device mesh has correctness issues before torch 2.3.0
+        return None
+
     if fsdp_config is None:
         return None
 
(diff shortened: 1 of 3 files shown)


composer/core/state.py
```diff
@@ -200,6 +200,10 @@ def _create_device_mesh(
     fsdp_config: Optional[Dict[str, Any]],
     tp_config: Optional[Dict[str, Any]],
 ) -> Optional[DeviceMesh]:
+    if version.parse(torch.__version__.split('.dev')[0]) < version.parse('2.3.0'):
+        # Device mesh has correctness issues before torch 2.3.0
```
bigning, line 204: should we raise error if users want to use 2d-parallel with torch version <2.3.0 ?

mvpatel2000: Yep, this happens at in `state._validate_parallelism_configs`

bigning (approved):
LGTM! thanks for the quick fix! It will be great to run multi-node save resume e2e test if you haven't done so.

snarayan21 (approved):
lgtm!

snarayan21:
@mvpatel2000 can you clarify in PR description what the checkpointing issue was, and why this is needed for torch < 2.3?

PR #3324, opened 2024-05-24
[ckpt-rewr] Resumption state dict API

eracah:
Introduces an API for extracting all parts of state necessary for resumption into a state dict. These include:

* timestamp
* schedulers
* run_name
* callbacks
* dataset state


composer/utils/data_helpers.py
```diff
@@ -0,0 +1,38 @@
+from composer.core.data_spec import DataSpec
+from torch.utils.data import DataLoader, Dataset
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Sequence, Union, cast
+
+if TYPE_CHECKING:
+    from composer.core.evaluator import Evaluator
+
+__all__ = ['_dataset_of']
+
+def _dataset_of(self, dataloader: Optional[Union[Evaluator, DataSpec, DataLoader, Iterable]]) -> Optional[Dataset]:
```
mvpatel2000, line 10: should we just put in misc since its only used in one place? I would prefer not to add more files if its only for one fn / not enough to group.

Also should this be a public fn?

composer/checkpoint/state_dict.py
```diff
@@ -152,3 +156,92 @@ def _get_model_state_dict_with_fsdp_context_manager(model: nn.Module, sharded_st
...
+        if hasattr(dataset, 'state_dict'):
+            # Don't save eval sample because we do not checkpoint during eval.
+            obj['eval'][evaluator.label] = dataset.state_dict(0, True)  # pyright: ignore
+
+    return dataset_state_dict
+
+
+# def _get_timestamp_state_dict
+# def _get_scheduler_state_dict,
+# def _dataset_state_state_dict
```
mvpatel2000, line 247: Is this todo?

composer/checkpoint/state_dict.py
```diff
@@ -158,6 +169,82 @@ def _get_model_state_dict_with_fsdp_context_manager(model: nn.Module, sharded_st
...
+    """Generate the state dict for any objects needed for resumption.
+
+    This includes:
+        * timestamp
+        * scheduler
+        * dataset_state
+        * scaler
+        * rank_zero_seed
+        * callbacks
+        * algorithms
```
mvpatel2000, line 182: iirc this list is stored as a constant in `state.py` as serialized atrributes. Is this the same / if so can we use that to avoid changing one part and ignoring another?

We can then test to ensure every attr in the list is included in resumption state dict to avoid discrepancies if someone changes it

mvpatel2000 (commented):
one minor comment, otherwise seems fine

composer/checkpoint/state_dict.py
```diff
@@ -158,6 +169,82 @@ def _get_model_state_dict_with_fsdp_context_manager(model: nn.Module, sharded_st
...
+
+
+def _make_state_dict_for_list_of_objects(objects: Union[Sequence[Any], Any]) -> Dict[str, Any]:
+    object_dict = {}
+    if not isinstance(objects, Sequence):
+        objects = [objects]
+    for obj in objects:
+        if not hasattr(obj, 'state_dict') or obj.state_dict() == {}:
+            continue
+        object_dict[type(obj).__qualname__] = obj.state_dict()
```
bigning, line 224: If there are more than 1 `obj`s in the `objects`,  will the `type(obj).__qualname__` be the same and overwrite the previous one ?

composer/checkpoint/state_dict.py
eracah: No. I think this is way more readable to explicitly put each item

eracah: We will remove that list from state.py once we switch to this API

composer/checkpoint/state_dict.py
eracah: oh interesting edge case. seems rare, but we could use a list perhaps?

bigning: I don't have full context here. I think as long as it's consistent for save and load, we should be good ?

eracah: ok yeah It should be good now

PR #3341, opened 2024-05-29
change ComposerModel output type

hyenal:
# What does this PR do?

This PR changes the output type of the `ComposerModel`. As long as the output type is a tensor, a tensor list, a dictionary of tensors or has a method `cpu()` (to satisfy [this](https://github.com/mosaicml/composer/blob/dev/composer/trainer/trainer.py#L3357), note that if you return a list of non-tensors it will not break the pipeline) we should be able to return any type.

No issue has been opened yet.

# Before submitting

cc: @JAEarly

### DIFF
--- a/composer/models/base.py
+++ b/composer/models/base.py
@@ -90,14 +90,14 @@ class ComposerModel(torch.nn.Module, abc.ABC):
         return state
 
     @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


mvpatel2000 (approved):
Agree it should be more arbitrary. Thanks for the PR!

PR #3343, opened 2024-05-29
Immediately kill the training if checkpoint downloading fails

bigning:
# What does this PR do?
pytorch issue: https://github.com/pytorch/pytorch/issues/122529

Basically, for remote checkpoint downloading, if the download failes, pytorch may fail to raise the correct exception, instead if might just raise nccl timeout. 

Previously, we added a log.error immediately, but in above case, the timeout takes very long time. Here, we just immediately kill the training process if the downloading error happens

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -21,6 +21,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Un
 
 import torch
 from packaging import version
+import sys
 from torch.distributed import checkpoint as dist_cp
 from torch.distributed._tensor import DeviceMesh
 from torch.distributed.checkpoint.metadata import Metadata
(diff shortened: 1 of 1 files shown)


composer/utils/checkpoint.py
```diff
@@ -276,9 +278,14 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner):
...
-                log.error(f'Exception {type(e)} raised during downloading: {str(e)}')
-                raise e
+                # We immediately kill the process and print the exception 
+                import psutil
+                log.warning(f'Exception {type(e)} raised during downloading: {str(e)}, terminating the process')
+                pid = os.getpid()
+                proc = psutil.Process(pid)
+                for child_process in proc.children(recursive=True):
+                    os.kill(child_process.pid, signal.SIGKILL)
+                os.kill(pid, signal.SIGKILL)
```
mvpatel2000, line 288: This kills the current process, but it wont kill other processes which will still NCCL deadlock, right?

This shoudl send a SIGTERM to the launcher process?

bigning: yeah, Irene has a PR to kill all processes https://github.com/mosaicml/composer/pull/3345

PR #3347, opened 2024-05-31
Add return type annotation when __init__ doesn't take any argument

antoinebrl:
According to [mypy documentation](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods):

> if `__init__` has no annotated arguments and no return type annotation, it is considered an untyped method


mvpatel2000 (approved):
Thanks!

Issue #3348, opened 2024-05-31
TypeError: Subscripted generics cannot be used with class and instance checks


kbaltakys:
`isinstance(device_train_microbatch_size, Union[int, float])`
should be
`isinstance(device_train_microbatch_size, (int, float))`
on line
https://github.com/mosaicml/composer/blob/3c0a8179d2c3d60786226a73d8f935bc46306a31/composer/trainer/trainer.py#L298

mvpatel2000:
Good point! Would you mind opening a PR and tagging me for review?

mvpatel2000:
Quick fix just did it. Thanks for flagging!

PR #3349, opened 2024-05-31
Fix type check issue with device train microbatch size

mvpatel2000:
# What does this PR do?

Resolves type issue with device train microbatch size. See https://github.com/mosaicml/composer/issues/3348


snarayan21 (approved):
fire

PR #3354, opened 2024-06-03
Rename metadata keys

mvpatel2000:
# What does this PR do?

Renames keys in `get_composer_env_dict`, following up on naming discussion in https://github.com/mosaicml/composer/pull/3311. Also cleans up code, removing a NamedTuple class which is immediately converted to a dictionary

### DIFF
--- a/composer/utils/collect_env.py
+++ b/composer/utils/collect_env.py
@@ -276,7 +276,7 @@ CPU Model: {cpu_model}
 CPU Count: {cpu_count}
 Number of Nodes: {node_world_size}
 GPU Model: {gpu_model}
-GPUs per Node: {local_world_size}
+GPUs per Node: {num_gpus_per_node}
 GPU Count: {num_gpus}
 CUDA Device Count: {cuda_device_count}
 """.strip()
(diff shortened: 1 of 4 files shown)


composer/utils/collect_env.py
```diff
@@ -282,38 +270,38 @@ def get_torch_env() -> str:
...
-    )._asdict()
-    return mutable_dict
+    return {
+        'composer_version': get_composer_version(),
+        'composer_commit_hash': get_composer_commit_hash(),
+        'cpu_model': get_cpu_model(),
+        'cpu_count': get_cpu_count(),
+        'num_nodes': get_node_world_size(),
+        'gpu_model': get_gpu_model(),
+        'local_world_size': get_local_world_size(),
```
eracah, line 295: should be `num_gpus_per_node`
```suggestion
        'local_world_size': get_local_world_size()
```

eracah (changes requested):

composer/utils/collect_env.py
```diff
@@ -282,38 +270,38 @@ def get_torch_env() -> str:
...
-        composer_commit_hash=get_composer_commit_hash(),
-        host_processor_model_name=get_host_processor_name(),
-        host_processor_core_count=get_host_processor_cores(),
-        node_world_size=get_node_world_size(),
-        accelerator_model_name=get_accel_model_name(),
-        local_world_size=get_local_world_size(),
-        cuda_device_count=get_cuda_device_count(),
-    )._asdict()
-    return mutable_dict
+    return {
```
b-chu, line 288: Can we make this a dataclass?

mvpatel2000: its only used in 2 places, not worth rn imo. we can do if we start using more?

eracah (approved):

PR #3356, opened 2024-06-03
MLFlow better experiment defaults

mvpatel2000:
# What does this PR do?

When running on MAIT or with a databricks secret, default to the databricks workspace. This lets the user skip specifying `tracking_uri`. 

Similarly, if specifying databricks secret, we now autopopulate `/Users/<username>` to experiment id, simplifying user experience.

Exmaple run name: ft-llama-3-8b-r4z14-v0-4roAvl

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -30,8 +30,6 @@ log = logging.getLogger(__name__)
 
 __all__ = ['MLFlowLogger']
 
-DEFAULT_MLFLOW_EXPERIMENT_NAME = 'my-mlflow-experiment'
-
 
 class MLFlowLogger(LoggerDestination):
     """Log to `MLflow <https://www.mlflow.org/docs/latest/index.html>`_.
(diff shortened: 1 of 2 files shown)


eracah:
I actually don't think this PR is necessary anymore. We can just add something loke this:
```
if experiment_name is None:
  experiment_name = 'my_mlflow_experiment'
if os.getenv(DATABRICKS_TOKEN) is not None and not experiment_name.startswith('/'):
  experiment_name = '/' + experiment_name
```

chenmoneygithub (approved):

b-chu:
Will let @eracah and @chenmoneygithub review

eracah (approved):
Whoa this is sick! this is huuuge

milocress (approved):
This looks great, good work!

b-chu (approved):
Great work @mvpatel2000!

dakinggg (approved):
approving the admin part of the change

Issue #3357, opened 2024-06-04
Autoresume and duration mismatch on reload


antoinebrl:
## Description

While experimenting with the `autoresume` feature, I encountered issues related to duration and the scheduler. My error was providing the `duration` argument to the `.fit()` method instead of the `max_duration` argument to the `Trainer constructor`. Since the `.fit()` method can be called multiple times sequentially, each call increases the `max_duration`, as indicated in the [code](https://github.com/mosaicml/composer/blob/a60bf3a16624b0cd3c1026657dfd61c9f74b16be/composer/trainer/trainer.py#L2202). Upon resumption, this offset causes an error in the scheduler because `t_max` becomes smaller than `max_duration`.

Using `max_duration` in the `Trainer` constructor avoids this problem, so I will adopt this approach. Should the scenario described above be detected, and should a warning or error be raised? Essentially, if `autoresume=True`, then `max_duration` should be specified in the `__init__`, and the `.fit()` method should only be called once in the script.

mvpatel2000:
@antoinebrl what if you set `t_max` to `1dur`? This would ensure `t_max = max_duration` always on resumption.

antoinebrl:
Your approach might solve the issue with the scheduler, which is how I identified the issue. However, the training still goes for longer than expected. For example, I do `Trainer(..., autoresume=True).fit(..., duration="10ep")`, the training crashes after 3 epochs, then after a restart the `fit` goes for an en extra 10 epochs. Here, the training lasts for 13 epochs effectively. The more interruptions or the closer they are to end the worse it gets.

mvpatel2000:
Yep... good point. I think it might also break on multiple resumptions. I will add a PR to at least require max_duration on init.

antoinebrl:
Awesome! Thanks for tackling it 💪

For a quick solution on my side I added couple of exceptions with instructions to guide the user toward the right path:
- in the `__init__`, when `autoresume==True` but `max_duration==None`
- in the `.fit`, if `autoresume` was `True` and `duration != None`
- in the `.fit`, if the method is called multiple times when `autoresume==True`

mvpatel2000:
I've added enforcements for the above restrictions. We will look at enabling autoresume + multiple fits, but its lower prioirty.

antoinebrl:
Thanks for your reactivity on this one 💪

mvpatel2000:
[...]

PR #3358, opened 2024-06-04
Autoresume Validation with Max Duration

mvpatel2000:
# What does this PR do?

Additional checks based on https://github.com/mosaicml/composer/issues/3357:
- Autoresume requires max_duration be specified on init and cannot accept `duration` in `fit`. Otherwise, Composer will autoresume from last checkpoint, and then the specified value of `max_duration` from `fit` will be incremented to the loaded max duration. In short, `fit(duration=...)` is not idempotent with autoresume.
- `reset_time` in `fit` does not work with autoresume as it will force a restart in training. Instead, time should be ignored on load, which is used in initial load and not on autoresume load
- Implicitly prevents multiple calls to fit -- if you cannot increment duration in fit or reset time, it is not possible to call fit twice.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1723,10 +1723,11 @@ class Trainer:
         # Load Checkpoint
         self._rng_state = None
         # If autoresume is enabled, first check for existing checkpoints to load
-        if autoresume:
+        self.autoresume = autoresume
+        if self.autoresume:
             log.info('Searching for a previous checkpoint to autoresume')
             error_message = ''
-            if max_duration is not None:
+            if max_duration is None:
                 error_message += 'The `max_duration` must be specified on trainer.__init__ when autoresume is enabled. '
             if save_folder is None:
                 error_message += 'The `save_folder` must be specified when autoresume is enabled. '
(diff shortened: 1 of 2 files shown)


bigning (approved):

eracah:
Do we want to keep the incremented functionality? If I load in a 10ba checkpoint with Trainer(..., max_duration='15ba') and I say fit(duration='12ba'), wouldn't I want it to train for 2 batches inatead of 5. Wouldn't what I specify in fit override what I did in the constructor?

mvpatel2000:
> Do we want to keep the incremented functionality? If I load in a 10ba checkpoint with Trainer(..., max_duration='15ba') and I say fit(duration='12ba'), wouldn't I want it to train for 2 batches instead of 5. Wouldn't what I specify in fit override what I did in the constructor?

Iirc the original behavior was modeled after what other training libraries do... but I agree it's perhaps not ideal. We should revisit this in a follow-on PR

eracah:
> > Do we want to keep the incremented functionality? If I load in a 10ba checkpoint with Trainer(..., max_duration='15ba') and I say fit(duration='12ba'), wouldn't I want it to train for 2 batches instead of 5. Wouldn't what I specify in fit override what I did in the constructor?
> 
> Iirc the original behavior was modeled after what other training libraries do... but I agree it's perhaps not ideal. We should revisit this in a follow-on PR

yeah we should this is very counter-intuitive

mvpatel2000:
> Do we want to keep the incremented functionality? If I load in a 10ba checkpoint with Trainer(..., max_duration='15ba') and I say fit(duration='12ba'), wouldn't I want it to train for 2 batches instead of 5. Wouldn't what I specify in fit override what I did in the constructor?

@antoinebrl curious if you have any input given your recent github issue referencing multiple fit calls

antoinebrl:
I must say this was a very counter intuitive bug to track down. This PR implements some fences which push the user towards the expected behavior but I agree the notion of training duration can be improved.

Regarding the multiple calls to `.fit`, we currently don't have that use-case. However, I remember reading a LLM paper where they do 2.1 epochs: 2 epochs with a very large dataset, and 0.1 for fine-tuning on a high quality dataset tailored for the task.

PR #3361, opened 2024-06-04
Remove unnecessary state dict and load_state_dict functions

eracah:
# What does this PR do?
The following items don't need to implement `state dict` and `load_state_dict` functions:
1. [DeviceGPU](https://github.com/mosaicml/composer/blob/dev/composer/devices/device_gpu.py#L61-L67)
    * returns cuda rng state. This isn't used anywhere since we just use rng state in the [reproducibility module](https://github.com/mosaicml/composer/blob/dev/composer/utils/reproducibility.py#L179-L262)
2. All the other devices 
    *  just return {} for state_dict and error for load_state_dict. Do we even need to have these functions implemented?
3. Loggers
   * just wandb ~and progress_bar~, which are never actually used in practice.


mvpatel2000 (approved):
LGTM

bigning (approved):
thanks for cleanup!

PR #3363, opened 2024-06-05
Generalize PEFT state dict filter

dakinggg:
# What does this PR do?
In the original PEFT integration PR, I copied some of the logic from PEFT for state dict filtering. My logic at the time was a bit flawed (I thought passing the model to this function would be an issue for some reason, but realized its not), so updating now to use the PEFT utility directly.


dakinggg:
Does not work because the built in function removes the adapter names from the state dict

PR #3369, opened 2024-06-06
Pin setuptools in docker

mvpatel2000:
# What does this PR do?

Pin setuptools in docker. Latest release broke features coming from torch which still rely on `pkg_resources`, which is removed in latest release


snarayan21 (commented):
lol this is some chicken and egg situation. the cpu tests are failing. is it possible to run cpu tests, or clone composer, using the docker images built from this? then we can force merge

dakinggg (approved):

PR #3372, opened 2024-06-06
[ckpt-rewr] Save state dict API

eracah:
# What does this PR do?

Implements an API for saving a state dict to disk given a sharded or full state dict:
* `save_state_dict_to_disk`
* `_save_sharded_state_dict_to_disk`
* `_save_full_state_dict_to_disk`

Tests for both CPU and GPU
Also modifies testing infra for comparing DTensors and ShardedTensors, generating sharded test models


composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+) -> Optional[str]:
+    """Saves a state dict to local disk.
+
+    Args:
+        state_dict (Dict[str,Any]): The state dict to save.
+        destination_dir (str): The directory to save the state dict to.
+        filename (str): The name of the file to save the state dict to.
+        overwrite (bool): If True, the file will be overwritten if it exists.
+        save_format (str): The format to save the state dict in. One of 'pt', 'hf', or 'safetensor'.
+        async_save (bool): If True, the save will be done asynchronously and the function will return with the path of where it was going to be saved
```
mvpatel2000, line 37: Does not exist? Can we remove docstring

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Returns:
+        str: The full path to the saved state dict if sharded is false and rank 0 or if sharded is true, otherwise None.
```
mvpatel2000, line 40:
```suggestion
        str: The full path to the saved state dict if (sharded is false and rank 0) or if sharded is true, otherwise None.
```
just a little easier to read :P

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    if state_dict == {}:
+        return None
```
mvpatel2000, line 43: why? can you not save an empty state_dict?

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _save_sharded_state_dict_to_disk(
+    state_dict: Dict[str, Any],
+    destination_file_path: str,
+    overwrite: bool = False,
+    save_format: str = 'pt',  # or safetensor
+) -> Optional[str]:
+
+    if save_format != 'pt':
+        raise NotImplementedError(f'Saving sharded state dict to disk in format {save_format} is not supported.')
```
mvpatel2000, line 65:
```suggestion
        raise NotImplementedError(f"Saving sharded state dict to disk in format {save_format} is not supported. Please choose from ['pt'].")
```
or something to this affect

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+
+    if state_dict == {}:
+        return None
```
mvpatel2000, line 68: same comment, why can't you save an empty state dict?

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # If used specifies filename instead of directory suffixes, strip them and warn.
```
mvpatel2000, line 70:
```suggestion
    # If user specifies filename instead of directory suffixes, strip them and warn
```
nit

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+    if dist.get_global_rank() == 0 and not overwrite and os.path.exists(destination_file_path):
+        raise ValueError(f'Directory {destination_file_path} already exists. Set overwrite=True to overwrite it.')
+
+    log.debug(
+        f'Starting saving of sharded state dict to {destination_file_path}/{_TORCH_DISTRIBUTED_CHECKPOINTS_FILENAME}',
+    )
+
+    # For 2.3.0 and above you can use checkpoint_id, but this version works the best for all versions
+    # of torch (and makes pyright happier) that we support, so we use it for now.
+    DCP.save(state_dict=state_dict, storage_writer=DCP.FileSystemWriter(destination_file_path))
```
mvpatel2000, line 91: no way to supply planner? or process group?

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _save_full_state_dict_to_disk(
+    state_dict: Dict[str, Any],
+    destination_file_path: str,
+    overwrite: bool = False,
+    save_format: str = 'pt',  # or hf, safetensor
+) -> Optional[str]:
+
+    if save_format != 'pt':
+        raise NotImplementedError(f'Saving full state dict to disk in format {save_format} is not supported.')
```
mvpatel2000, line 104:
```suggestion
        raise NotImplementedError(f"Saving sharded state dict to disk in format {save_format} is not supported. Please choose from ['pt'].")
```
same as above

tests/checkpoint/helpers.py
```diff
@@ -0,0 +1,109 @@
...
+):
+    inputs = torch.randn(batch_size, num_features, device=device)
+    targets = torch.randint(low=0, high=num_classes, size=(batch_size,), device=device, dtype=torch.long)
+    batch = (inputs, targets) if use_composer_model else inputs
+    optimizer = adam.Adam(model.parameters())
+    outputs = model(batch)
+    loss = loss_fn(outputs, targets)
+    loss.backward()
+    if take_step:
+        optimizer.step()
```
mvpatel2000, line 108: why do you need to do a step?

tests/checkpoint/helpers.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Dict
+
```
mvpatel2000, line 5: should this be in `common`?

tests/checkpoint/test_save.py
```diff
@@ -0,0 +1,99 @@
...
+@pytest.mark.filterwarnings('ignore:The passed')  # Torch issues a warning for wrapping a CPU model in FSDP
+@pytest.mark.parametrize('sharded_model', [False, True])
+def test_save_full_state_dict_to_disk_cpu(world_size: int, tmp_path: str, sharded_model: bool):
+
+    destination_file_path = os.path.join(tmp_path, 'test.pt')
+    use_fsdp = sharded_model
+    model, _ = init_model(use_fsdp=use_fsdp, device='cpu', sync_module_states=False, cpu_offload=True)
+    state_dict = get_model_state_dict(model, sharded_state_dict=False)
+    path_saved = save_state_dict_to_disk(state_dict, destination_file_path=destination_file_path)
+    time.sleep(1)
```
mvpatel2000, line 55: why ?

composer/checkpoint/save.py
```diff
@@ -0,0 +1,133 @@
...
+    """Determines if the state dict is sharded.
+
+    Args:
+        state_dict (Dict[str, Any]): The state dict to check.
+
+    Returns:
+        bool: Whether the state dict is sharded.
+    """
+    sample_value = next(iter(state_dict.values()))
+    return isinstance(sample_value, ShardedTensor) or isinstance(sample_value, DTensor)
```
bigning, line 125: i think even for sharded state dict, it's possible there is Tensor in state_dict ? So should be better to just pass in `is_sharded` to  `save_state_dict_to_disk`

composer/checkpoint/save.py
eracah: why would we save an empty state dict?

tests/checkpoint/helpers.py
eracah: To initialize the optimizer state. btw this is out of the scope of this PR. These helper functions were introduced in the generate optim state dict PR

composer/checkpoint/save.py
eracah: I don't think that is the case

composer/checkpoint/save.py
eracah: the use case for this is if you generate a full state dict and pass it to the save function then all non-zero-ranks will have an empty state dict, so either:
* we save an empty file for every rank. I can't think of a use case where you'd want that
* we require the user to add a gating by rank if statement (the idea is that the user doesn't have to make any dist.* calls nor do they even have to know about dist.*)

composer/checkpoint/save.py
eracah: we can add that in a future PR. For now (with this tight timeline), we just want a simple mvp for these API's

eracah: I can add them back in if you're ok with no unit tests for that functionality

tests/checkpoint/helpers.py
eracah: they could. I made them specifically for checkpointing tests, but if you think they would be useful in common. However, if putting them in common requires a more time-consuming review then maybe not

tests/checkpoint/test_save.py
eracah: removed

composer/checkpoint/save.py
bigning: I found this case while working on this issue: https://github.com/pytorch/pytorch/issues/125740. That issues is caused by deduping `Tensor` in state_dict to different ranks. Specifically, it's a optimizer scalar tensor in optimizer state

eracah: oh I see yeah ok well we can test all values then maybe? and we can add a unit test with optimizers too?

bigning: > we can test all values

I think that works. but why not just pass in the `is_sharded` ?  that should be much easier and safe ? :)

eracah: would be nice to not have to pass a flag if one could infer it

composer/checkpoint/save.py
```diff
@@ -0,0 +1,139 @@
...
+        state_dict (Dict[str, Any]): The state dict to check.
+
+    Returns:
+        bool: Whether the state dict is sharded.
+    """
+    sample_value = next(iter(state_dict.values()))
+    for value in state_dict.values():
+        if isinstance(value, ShardedTensor) or isinstance(sample_value, DTensor):
+            return True
+
```
bigning, line 131:
```suggestion
    sample_value = next(iter(state_dict.values()))
    for value in state_dict.values():
        if isinstance(value, ShardedTensor) or isinstance(sample_value, DTensor):
            return True
    return False
```

bigning (approved):

bigning:
Please wait for the code freeze

mvpatel2000 (dismissed):
Holding for freeze

composer/checkpoint/save.py
```diff
@@ -0,0 +1,139 @@
...
+        save_format (str): The format to save the state dict in. One of 'pt', 'hf', or 'safetensor'.
+
+    Returns:
+        str: The full path to the saved state dict if (sharded is false and rank 0) or if sharded is true, otherwise None.
+    """
+    if state_dict == {}:
+        return None
+    sharded_state_dict = is_state_dict_sharded(state_dict)
+
+    if sharded_state_dict:
```
mvpatel2000, line 45:
```suggestion
    if is_state_dict_sharded(state_dict):
```

composer/checkpoint/save.py
```diff
@@ -0,0 +1,139 @@
...
+            path_saved = None
+
+    return path_saved
+
+
+def _save_sharded_state_dict_to_disk(
+    state_dict: Dict[str, Any],
+    destination_file_path: str,
+    overwrite: bool = False,
+    save_format: str = 'pt',  # or safetensor
```
mvpatel2000, line 60:
```suggestion
    save_format: str = 'pt',
```
doesnt exist rn

composer/checkpoint/save.py
mvpatel2000: Follow on PR is fine, pro modularity. Note that this blocks switchover due to a few issues which require planner CC: @bigning  on adam

tests/checkpoint/helpers.py
mvpatel2000: I think it's fine for now... though it seems nice to unify eventually.

mvpatel2000 (commented):
LGTM besides hold

composer/checkpoint/save.py
bigning: ah, that's a good point. We definitely need to add [this](https://github.com/mosaicml/composer/blob/ba82cc9ec943e316da59ec6ec0ba540008c9ca1f/composer/utils/checkpoint.py#L1156) in follow on PR before enabling V2 in production.

eracah:
> Holding for freeze

please release

composer/checkpoint/save.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Args:
+        state_dict (Dict[str, Any]): The state dict to check.
+
+    Returns:
+        bool: Whether the state dict is sharded.
+    """
+    sample_value = next(iter(state_dict.values()))
+    for value in state_dict.values():
+        if isinstance(value, ShardedTensor) or isinstance(sample_value, DTensor):
```
bigning, line 132: shouldn't this be `isinstance(value, DTensor)` ?

eracah: oh good catch, @bigning ! fixed

PR #3373, opened 2024-06-06
min/max flagging added to system_metrics_monitor with only non-redundant, necessary gpu metrics logged

JackZ-db:
# What does this PR do?

Added logging for gpu power and removed all gpu metrics covered in other callbacks (used, free, total memory). 

The following GPU metrics pertaining to Straggler Detection are logged:

        gpu_percentage: Occupancy rate, percent of time over the past sampling period during
                        which one or more kernels was executing on the GPU.
        memory_percentage: Percent of time over the past sampling period during which
                        global (device) memory was being read or written.
        gpu_temperature_C: Temperature of device, in Celcius.
        gpu_power_usage_W: Power usage of device, in Watts.

Added clearer documentation for the SystemsMetricMonitor class and removed the user inputted boolean parameter for whether the model uses GPU's or not (gpu_available).

Added a boolean flag for users to select whether to log the min/max values for the GPU metrics listed above and their corresponding their ranks, or log all values for all ranks (log_all_data, default set to false).

# Before submitting


composer/callbacks/system_metrics_monitor.py
```diff
@@ -19,13 +20,56 @@
...
+   'gpu_percentage',
+   'memory_percentage',
+   'gpu_temperature_C',
+   'gpu_power_usage_W',
+]
+
 
 class SystemMetricsMonitor(Callback):
-    """Track system metrics."""
+    """Logs GPU and CPU metrics relevant to straggler detection across all ranks.
```
mvpatel2000, line 32:
```suggestion
    """Logs CPU/GPU system metrics.
```
nit: more concise. it can be used for things besides straggler detection and by default is across all ranks

composer/callbacks/system_metrics_monitor.py
```diff
@@ -83,3 +140,19 @@ def compute_system_metrics(self):
         for k, v in network_usage.items():
             system_metrics[f'network_{k}'] = v
         return system_metrics
+
+    def compute_min_max_metrics(self, all_metrics, model_device):
+        min_max_metrics = {}
+
+        if self.gpu_available:
+            gpu_metrics = _GPU_METRICS
+            for key in gpu_metrics:
```
mvpatel2000, line 149:
```suggestion
            for key in _GPU_METRICS:
```

composer/callbacks/system_metrics_monitor.py
```diff
@@ -46,9 +90,25 @@ def run_event(self, event: Event, state: State, logger: Logger):
         ]:
             local_node_system_metrics = self.compute_system_metrics()
             all_system_metrics = dist.all_gather_object(local_node_system_metrics)
-            system_metrics = {
-                key: value for local_metrics in all_system_metrics for key, value in local_metrics.items()
-            }
+            system_metrics = {}
+            gpu_metrics_set = set(_GPU_METRICS)
```
mvpatel2000, line 94:
```suggestion
```

composer/callbacks/system_metrics_monitor.py
```diff
@@ -46,9 +90,25 @@ def run_event(self, event: Event, state: State, logger: Logger):
...
+            if self.log_all_data:
+                for rank, metrics in enumerate(all_system_metrics):
+                    for key, value in metrics.items():
+                        if key in gpu_metrics_set:
+                            system_metrics[f'Rank_{rank}_{key}'] = value
+                        else:
+                            system_metrics[key] = value
+
+            else:
+                model_device = next(state.model.parameters()).device
```
mvpatel2000, line 105: should use `state.device` instead?

composer/callbacks/system_metrics_monitor.py
```diff
@@ -19,13 +20,56 @@
...
+        ...    eval_dataloader=eval_dataloader,
+        ...    optimizers=optimizer,
+        ...    max_duration='1ep',
+        ...    callbacks=[SystemMetricsMonitor()],
+        ... )
+
+
+    Args:
+    log_all_data (bool, optional): True if user wants to log data for all ranks, not just the min/max.
+    Defaults to False.
```
mvpatel2000, line 66:
```suggestion
    Args:
        log_all_data (bool, optional): True if user wants to log data for all ranks, not just the min/max.
        Defaults to False.
```

composer/callbacks/system_metrics_monitor.py
```diff
@@ -46,9 +90,25 @@ def run_event(self, event: Event, state: State, logger: Logger):
...
+                        else:
+                            system_metrics[key] = value
+
+            else:
+                model_device = next(state.model.parameters()).device
+                system_metrics = self.compute_min_max_metrics(all_system_metrics, model_device)
+                for rank, metrics in enumerate(all_system_metrics):
+                    for key, value in metrics.items():
+                        if key not in gpu_metrics_set:
+                            system_metrics[key] = value
```
mvpatel2000, line 110: why is this if statement needed?

composer/callbacks/system_metrics_monitor.py
```diff
@@ -19,13 +20,56 @@
...
+    GPU Metrics:
+        gpu_percentage: Occupancy rate, percent of time over the past sampling period during
+                        which one or more kernels was executing on the GPU.
+        memory_percentage: Percent of time over the past sampling period during which
+                        global (device) memory was being read or written.
+        gpu_temperature_C: Temperature of device, in Celcius.
+        gpu_power_usage_W: Power usage of device, in Watts.
+
+    If the log_all_data flag is set to false (which is false by default), only the maximum and minimum values
+    for these metrics, alongside their respective ranks in the key names, are logged
```
mvpatel2000, line 43: since default is false, instead you should say if it's set to `True`, here is additional information thats logged

mvpatel2000 (commented):
LGTM, mostly minor style comments :)

composer/callbacks/system_metrics_monitor.py
JackZ-db: I'm following memory_monitor's usage of reduce_value and unit testing, which uses the device of next(state.model.parameters()) - switching to state.device broke unit testing for me

composer/callbacks/system_metrics_monitor.py
JackZ-db: we need to still log the cpu metrics without logging the original gpu metrics without the added prefixes

composer/callbacks/system_metrics_monitor.py
mvpatel2000: Discussed offline, sent API spec

composer/callbacks/system_metrics_monitor.py
mvpatel2000: Discussed offline

composer/callbacks/system_metrics_monitor.py
```diff
@@ -83,3 +136,17 @@ def compute_system_metrics(self):
...
+        min_max_metrics = {}
+
+        if self.gpu_available:
+            for key in _GPU_METRICS:
+                values = torch.tensor([metrics_for_cur_rank[key] for metrics_for_cur_rank in all_metrics],
+                                      device=model_device)
+                min_rank = int(torch.argmin(values).item())
+                max_rank = int(torch.argmax(values).item())
+                min_max_metrics[f'min_{key}/Rank_{min_rank}'] = values[min_rank].item()
+                min_max_metrics[f'max_{key}/Rank_{max_rank}'] = values[max_rank].item()
```
mvpatel2000, line 150:
```suggestion
                min_max_metrics[f'min_{key}_rank_{min_rank}'] = values[min_rank].item()
                min_max_metrics[f'max_{key}_rank_{max_rank}'] = values[max_rank].item()
```
to be consistent with other log

composer/callbacks/system_metrics_monitor.py
```diff
@@ -46,9 +86,25 @@ def run_event(self, event: Event, state: State, logger: Logger):
...
-                key: value for local_metrics in all_system_metrics for key, value in local_metrics.items()
-            }
+            system_metrics = {}
+            gpu_metrics_set = set(_GPU_METRICS)
+
+            if self.log_all_data:
+                for rank, metrics in enumerate(all_system_metrics):
+                    for key, value in metrics.items():
+                        if key in gpu_metrics_set:
+                            system_metrics[f'Rank_{rank}_{key}'] = value
```
mvpatel2000, line 96:
```suggestion
                            system_metrics[f'{key}_rank_{rank}'] = value
```
it's probably easier to view when sorted by key vs. rank and this lets us standardize formatting across the two log calls

composer/callbacks/system_metrics_monitor.py
```diff
@@ -83,3 +136,17 @@ def compute_system_metrics(self):
         for k, v in network_usage.items():
             system_metrics[f'network_{k}'] = v
         return system_metrics
+
+    def compute_min_max_metrics(self, all_metrics, model_device):
```
mvpatel2000, line 140:
```suggestion
    def compute_gpu_min_max_metrics(self, all_metrics, model_device):
```
Can we rename to be a bit more explicit since it ignores the CPU ones?

mvpatel2000 (commented):
Last wave of nits!

mvpatel2000 (approved):
LGTM!

composer/callbacks/system_metrics_monitor.py
```diff
@@ -83,3 +134,17 @@ def compute_system_metrics(self):
         for k, v in network_usage.items():
             system_metrics[f'network_{k}'] = v
         return system_metrics
+
+    def compute_gpu_min_max_metrics(self, all_metrics, state):
```
mvpatel2000, line 138: you could just pass `state.device` here instead of all of `state`. not strongly opinionated

j316chuck (approved):

PR #3376, opened 2024-06-06
[Checkpoint] Fix symlink issue where symlink file uploaded before checkpoint files upload

bigning:
# What does this PR do?
Fix the symlink issues. 
**How**?
[updated]: in the checkpoint saver, on rank-0 which saves the symlink, it all_gather the remote checkpoint file names, and start a new process to check if those remote files finish uploading by calling `object_store.get_object_size`. It only upload symlink file once all the remote file finish uploading. This way:
1. it won't block the training, since it's a separate process
2. it almost has no delay to upload the symlink file (it will sleep 30s if remote file is not there yet)

# Unit test
#  Integration test
## 2-nodes OCI:
save: test-uploader-0Tkv9O
autoresume: test-uploader-yac88U

## 2-nodes mflow:
save:  l38bi-full-sweep-train-bb-1-0e-6-5-VMY2Xo
load: l38bi-full-sweep-train-bb-1-0e-6-5-TyRcPi

## Daily test:
https://github.com/mosaicml/composer/actions/runs/9700144963

## composer regression test:
https://github.com/databricks-mosaic/regression-testing/actions/runs/9700161085

## Perf test (100 batches with 9 batch save interval. The training time varies because of unstable uploading speed, but just want to make sure test didn't regress training time)
64 gpu test: 77b-bs1024-g512-res2-f60-37gSXK  time: 26 minutes
64 GPU baseline: 77b-bs1024-g512-res2-f60-2VZPxB time: more than 40 minutes because rank 29 uploading delay

## In case 1 rank upload fails, it won't hang:
test-uploader-MXcoEp


tests/utils/test_remote_uploader.py
```diff
@@ -38,6 +41,8 @@ def upload_object(
             raise RuntimeError('Raise Error intentionally')
         time.sleep(self.sleep_sec)
         dest_filename = pathlib.Path(self.root) / object_name
+        os.makedirs(os.path.dirname(dest_filename), exist_ok=True)
+        print(f"bigning debug {filename=}, {dest_filename=}")
```
bigning, line 45: will remove this.

pyproject.toml
```diff
@@ -88,7 +88,7 @@ reportUnusedCoroutine = "error"
 # Pytest
 [tool.pytest.ini_options]
 # By default, do not run gpu, vision, docs, notebook, or daily tests
-addopts = "--codeblocks --strict-markers -m 'not gpu and not doctest and not daily and not remote'"
+addopts = "--strict-markers -m 'not gpu and not doctest and not daily and not remote'"
```
mvpatel2000, line 91: why remove codeblocks?

composer/utils/remote_uploader.py
```diff
@@ -97,6 +120,27 @@ def __init__(
...
+            return
+        if get_global_rank() == 0:
+            if self.object_store is None:
+                self.object_store = _build_dbfs_backend(self.path)
+            assert isinstance(self.object_store, MLFlowObjectStore)
+            self.path = self.object_store.get_dbfs_path(self.path)
+        path_list = [self.path]
+        broadcast_object_list(path_list, src=0)
+        self.path = path_list[0]
+        # TODO: add valdation
```
mvpatel2000: this PR or follow on?

composer/callbacks/checkpoint_saver.py
```diff
@@ -321,6 +328,31 @@ def __init__(
 
         self.start_batch = None
 
+        self.remote_uploader = None
+        backend, _, _ = parse_uri(save_folder)
+        self.remote_uploader_futures: List[List[Future]] = []
+        self.symlink_file_tasks: List[tuple[str, str]] = []
+        self.this_rank_saves_remote_symlinks: bool = False
```
mvpatel2000, line 335:
```suggestion
        self.rank_saves_remote_symlinks: bool = False
```
nit: more concise

composer/callbacks/checkpoint_saver.py
```diff
@@ -321,6 +328,31 @@ def __init__(
...
+        self.tmp_dir_for_symlink = tempfile.TemporaryDirectory()
+        self.num_concurrent_uploads = num_concurrent_uploads
+
+        if backend != '':
+            if backend == 'wandb':
+                raise NotImplementedError(
+                    f'There is no implementation for WandB via URI. Please use '
+                    'WandBLogger with log_artifacts set to True',
```
mvpatel2000, line 343:
```suggestion
                    'WandBLogger with log_artifacts set to True.',
```
nit

composer/callbacks/checkpoint_saver.py
```diff
@@ -321,6 +328,31 @@ def __init__(
...
+                )
+            elif backend not in ['s3', 'oci', 'gs', 'azure', 'dbfs']:
+                raise NotImplementedError(
+                    f'There is no implementation for the cloud backend {backend} via URI. Please use '
+                    'one of the supported RemoteUploaderDownloader object stores',
```
mvpatel2000, line 348:
```suggestion
                    'one of the supported RemoteUploaderDownloader object stores.',
```
nit

composer/callbacks/checkpoint_saver.py
```diff
@@ -321,6 +328,31 @@ def __init__(
...
+                )
+            self.remote_uploader = RemoteUploader(
+                remote_folder=save_folder,
+                num_concurrent_uploads=self.num_concurrent_uploads,
+            )
```
mvpatel2000, line 353: wait so... what actualyl is supported as a backend?

composer/callbacks/checkpoint_saver.py
mvpatel2000, line 343: doesn't this not work since you don't call `upload_file`?

composer/callbacks/checkpoint_saver.py
```diff
@@ -510,17 +577,23 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                    self.tmp_dir_for_symlink.name,
+                    f'latest.{self.count}.symlink',
+                )
+                # Sharded checkpoints for torch >2.0 use directories not files for load_paths
+                if state.fsdp_sharded_state_dict_enabled:
+                    src_path = str(pathlib.Path(remote_file_name).parent)
+                else:
+                    src_path = remote_file_name
+                log.debug(f'Creating symlink file {symlink_filename} -> {src_path}')
+                this_rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
```
mvpatel2000, line 590:
```suggestion
                rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
```
nit conciseness. Also why is this different from the class attribute? why does it need to be rechecked?

mvpatel2000 (commented):
So if I specify a remote uploader, it does not call `logger.upload_file`, but if I do not specify one, it does?

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
         if self.num_checkpoints_to_keep >= 0:
             self._rotate_checkpoints(sharding_enabled=state.fsdp_sharded_state_dict_enabled)
 
-    def _rotate_checkpoints(self, sharding_enabled: bool = False):
+    def wait(self) -> None:
+        if self.remote_uploader is None:
+            return
+        # Wait remote uploader futures and start to upload the latest symlink file if necessary
+        if self.this_rank_saves_remote_symlinks:
+            if len(self.remote_uploader_futures) != len(self.symlink_file_tasks):
```
mvpatel2000, line 613: this can be combined to noe statement?

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
         if self.num_checkpoints_to_keep >= 0:
             self._rotate_checkpoints(sharding_enabled=state.fsdp_sharded_state_dict_enabled)
 
-    def _rotate_checkpoints(self, sharding_enabled: bool = False):
+    def wait(self) -> None:
```
mvpatel2000, line 608: can this fn get a doc string given it's importance and subtle logic here?

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+        for i in range(len(self.remote_uploader_futures)):
+            for future in self.remote_uploader_futures[i]:
+                future.result()
+        log.debug(f'Current rank finished existing uploading tasks')
+        self.remote_uploader_futures = []
+
+        t = dist.get_device(None).tensor_to_device(torch.tensor(1))
+        dist.all_reduce(t)
+        if t.item() != dist.get_world_size():
+            raise RuntimeError(f'Some rank failed to upload checkpoint files')
```
mvpatel2000, line 627: Why is this an `all_reduce` and not a barrier? I don't understand this step

composer/callbacks/checkpoint_saver.py
bigning: I just make this same as existing RUD, which doesn't support wandb https://github.com/mosaicml/composer/blob/13f2a4fab92ef33f4c490a31f17c608ddb1a7eee/composer/utils/file_helpers.py#L478-L482

composer/callbacks/checkpoint_saver.py
bigning: it supports `['s3', 'oci', 'gs', 'azure', 'dbfs']`. Did i miss anything?

composer/callbacks/checkpoint_saver.py
bigning: because barrier has known issue that even if fails, it can still run some lines of code below it, which will start to upload the symlink file. Here using a all_reduce and its result to make sure if the all_reduce fails, it will just raise RuntimeError here. Does that make sense ?

composer/callbacks/checkpoint_saver.py
bigning: because we don't know `self.rank_saves_symlinks` is true or not in the __init__. We only know once it runs into this code. I can add a check if `self.rank_saves_symlinks` is set or not, but i don't think it's very necessary.

composer/utils/remote_uploader.py
bigning: THis PR. just added. :)

bigning:
> So if I specify a remote uploader, it does not call `logger.upload_file`, but if I do not specify one, it does?

Yes. So it matches the current behavior:
1. if the save_folder is remote, it will create a new remote_uploader, use that one to upload
2. if the save_folder is local, but users pass in rud in logger, we call `logger.upload_file`

composer/callbacks/checkpoint_saver.py
```diff
@@ -456,12 +496,14 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-            this_rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
-            if this_rank_saves_symlinks:
+            rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
+            if rank_saves_symlinks:
                 os.symlink(os.path.relpath(src_path, os.path.dirname(symlink)), symlink)
 
         # if remote file name provided, upload the checkpoint
         if self.remote_file_name is not None:
+
+            futures: List[Future] = []
```
b-chu, line 506:
```suggestion
            futures: list[Future] = []
```
I'm surprised this doesn't throw a typing error

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
         if self.num_checkpoints_to_keep >= 0:
             self._rotate_checkpoints(sharding_enabled=state.fsdp_sharded_state_dict_enabled)
 
-    def _rotate_checkpoints(self, sharding_enabled: bool = False):
+    def wait(self) -> None:
+        """Wait exsiting upload tasks to finish and start uploading symlink file if necessary."""
```
b-chu, line 609: Grammar

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+        if self.remote_uploader is None:
+            return
+        # Wait remote uploader futures and start to upload the latest symlink file if necessary
```
b-chu, line 612: This is the same as the docstring. Change to a comment that explains the logic

composer/callbacks/checkpoint_saver.py
```diff
@@ -532,8 +605,38 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+            )
+        log.debug('Waiting for previous checkpoint files upload finish')
+        for i in range(len(self.remote_uploader_futures)):
+            for future in self.remote_uploader_futures[i]:
+                future.result()
+        log.debug(f'Current rank finished existing uploading tasks')
+        self.remote_uploader_futures = []
+
+        t = dist.get_device(None).tensor_to_device(torch.tensor(1))
+        dist.all_reduce(t)
```
b-chu, line 625: Why does this not block training during uploading?

composer/utils/remote_uploader.py
```diff
@@ -12,30 +12,49 @@
 import time
 import uuid
 from concurrent.futures import Future, ProcessPoolExecutor
-from typing import List
+from typing import List, Optional
```
b-chu, line 15: @mvpatel2000 I thought we changed all List to list?

b-chu: PR maybe needs rebase?

composer/utils/remote_uploader.py
```diff
@@ -12,30 +12,49 @@
...
+)
+from composer.utils.object_store.uc_object_store import UCObjectStore
 from composer.utils.retrying import retry
 
 log = logging.getLogger(__name__)
 
 __all__ = ['RemoteUploader']
 
 
+def _build_dbfs_backend(path: str) -> ObjectStore:
```
b-chu, line 36: Add this logic to `maybe_create_object_store_from_uri`?

composer/utils/remote_uploader.py
```diff
@@ -12,30 +12,49 @@
...
     overwrite: bool,
     num_attempts: int,
 ) -> int:
-    object_store: ObjectStore = maybe_create_object_store_from_uri(
-        remote_folder,
-    )  # pyright: ignore[reportGeneralTypeIssues]
+    if is_dbfs:
+        object_store: ObjectStore = _build_dbfs_backend(dbfs_path)
+    else:
+        object_store: ObjectStore = maybe_create_object_store_from_uri(
```
b-chu, line 55: If `_build_dbfs_backend` is part of `maybe_create_object_store_from_uri` then you don't need is_dbfs and dbfs_path

b-chu (commented):
Thanks for the PR!

composer/callbacks/checkpoint_saver.py
bigning: the orginal method would block training if the uploading hasn't finished . in the latstet commit, I changed the way to upload the symlink and made it completely async.

composer/utils/remote_uploader.py
bigning: this is not needed in the latest commit.

composer/callbacks/checkpoint_saver.py
```diff
@@ -321,6 +380,41 @@ def __init__(
...
+            self.remote_uploader = RemoteUploader(
+                remote_folder=save_folder,
+                num_concurrent_uploads=self.num_concurrent_uploads,
+            )
+            self.symlink_upload_executor = ProcessPoolExecutor(
+                max_workers=1,
+                mp_context=multiprocessing.get_context('spawn'),
+            )
+
+        self.count = 0
```
mvpatel2000, line 416: can we add a more descriptive name? and maybe move up to w rest of `self.X` var initializations?

composer/utils/remote_uploader.py
```diff
@@ -95,7 +118,31 @@ def __init__(
...
+    def init(self):
+        # If it's dbfs path like: dbfs:/databricks/mlflow-tracking/{mlflow_experiment_id}/{mlflow_run_id}/
+        # We need to fill out the experiment_id and run_id
+
+        if get_global_rank() == 0:
+            retry(
+                ObjectStoreTransientError,
+                self.num_attempts,
+            )(lambda: validate_credentials(self.remote_backend, '.credentials_validated_successfully'))()
+        if self.path.startswith(MLFLOW_DBFS_PATH_PREFIX):
```
mvpatel2000, line 138: maybe split out lambda fn creation into a different line for readability

composer/callbacks/checkpoint_saver.py
```diff
@@ -40,6 +48,54 @@
...
+    for remote_file_name in remote_checkpoint_file_names:
+        while True:
+            # Return if parent process exits
+            try:
+                os.kill(main_process_pid, 0)
+            except OSError:
+                return
+            try:
+                object_store.get_object_size(remote_file_name)
+                break
```
mvpatel2000, line 75: why this fn? can u not just check if it exists w/o checking size?

mvpatel2000 (commented):
LGTM besides minor nits. Woudl like a second review though given subtlety. 

Will likely need a follow-on PR where on load we can safely resume from this symlink with autoresume (because it will crash when it resumes and tries to save next ckpt w some already present shards)

composer/callbacks/checkpoint_saver.py
bigning: do you have any API i can use in your mind? I dont' find a function that can check if a remote object exists [here](https://github.com/mosaicml/composer/blob/dev/composer/utils/object_store/object_store.py) . 

The `get_object_size` is how it's used to check if an object exists or not [here](https://github.com/mosaicml/composer/blob/25b656f4a97606fe1bf7edcd806bf5275311df0f/composer/utils/remote_uploader.py#L44)

bigning:
> Will likely need a follow-on PR where on load we can safely resume from this symlink with autoresume (because it will crash when it resumes and tries to save next ckpt w some already present shards)

Yeah, that's the one of the v2 task: corrupted checkpoint loading

composer/callbacks/checkpoint_saver.py
```diff
@@ -39,6 +47,58 @@
...
+            except Exception as e:
+                if not isinstance(e, FileNotFoundError):
+                    log.debug(f'Got exception {type(e)}: {str(e)} when accessing remote file {remote_file_name}')
+                time.sleep(wait_before_next_try_in_seconds)
+            if time.time() - start_time > max_wait_time_in_seconds:
+                raise RuntimeError(
+                    f'Checkpoint file {remote_file_name} uploading not finished after {max_wait_time_in_seconds} seconds',
+                )
+
+    log.debug(f'Uploading symlink file {remote_symlink_file_name}')
```
b-chu, line 88: Once the symlink file is uploaded, it's guaranteed to be pointing at a complete checkpoint right? How do you handle if the next checkpoint is partially uploaded but the run crashes. If overwrite is False, it'll crash on resumption right?

composer/callbacks/checkpoint_saver.py
```diff
@@ -39,6 +47,58 @@
...
+            try:
+                os.kill(main_process_pid, 0)
+            except OSError:
+                return
+            try:
+                object_store.get_object_size(remote_file_name)
+                break
+            except Exception as e:
+                if not isinstance(e, FileNotFoundError):
+                    log.debug(f'Got exception {type(e)}: {str(e)} when accessing remote file {remote_file_name}')
```
b-chu, line 81: If credentials aren't set properly, will this make it fail slowly?

composer/callbacks/checkpoint_saver.py
```diff
@@ -320,6 +383,43 @@ def __init__(
...
+        if backend != '':
+            if backend == 'wandb':
+                raise NotImplementedError(
+                    f'There is no implementation for WandB via URI. Please use '
+                    'WandBLogger with log_artifacts set to True.',
+                )
+            elif backend not in ['s3', 'oci', 'gs', 'azure', 'dbfs']:
+                raise NotImplementedError(
+                    f'There is no implementation for the cloud backend {backend} via URI. Please use '
+                    'one of the supported object stores.',
```
b-chu, line 410: Can you add `['s3', 'oci', 'gs', 'azure', 'dbfs']` to the error message for clarity to the user?

composer/callbacks/checkpoint_saver.py
```diff
@@ -320,6 +383,43 @@ def __init__(
...
+        self.tmp_dir_for_symlink = tempfile.TemporaryDirectory()
+        self.num_concurrent_uploads = num_concurrent_uploads
+        self.symlink_upload_executor = None
+        self.is_remote_upload_failed = None
+        self.symlink_upload_futures = []
+        self.upload_timeout_in_seconds = upload_timeout_in_seconds
+        # Allow unit test to override this to make it faster
+        self._symlink_upload_wait_before_next_try_in_seconds = 30.0
+        self.pid = os.getpid()
+        self.symlink_count = 0
```
b-chu, line 399: Does this need to get saved in state_dict for resumption?

composer/callbacks/checkpoint_saver.py
```diff
@@ -482,11 +589,19 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                            file_path=pathlib.Path(metadata_local_file_path),
+                            overwrite=self.overwrite,
+                        )
+                        local_remote_filenames.append(metadata_remote_file_name)
+                    else:
+                        logger.upload_file(
+                            remote_file_name=metadata_remote_file_name,
+                            file_path=metadata_local_file_path,
+                            overwrite=self.overwrite,
+                        )
```
b-chu, line 604: This seems very similar to below. Can you helper function it with remote_file_name and local_path?

composer/utils/file_helpers.py
```diff
@@ -737,3 +738,18 @@ def create_symlink_file(
         raise ValueError('The symlink filename must end with .symlink.')
     with open(destination_filename, 'x') as f:
         f.write(existing_path)
+
+
+def validate_credentials(
+    remote_backend: ObjectStore,
+    remote_file_name_to_test: str,
+) -> None:
```
b-chu, line 746:
```suggestion
):
```

composer/utils/file_helpers.py
```diff
@@ -737,3 +738,18 @@ def create_symlink_file(
...
+    """Upload a tiny text file to test if the credentials are setup correctly."""
+    # Validates the credentials by attempting to touch a file in the bucket
+    # raises an error if there was a credentials failure.
+    with tempfile.NamedTemporaryFile('wb') as f:
+        f.write(b'credentials_validated_successfully')
+        remote_backend.upload_object(
```
b-chu, line 752: Delete this file after successful validation?

composer/utils/object_store/utils.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2022 MosaicML Composer authors
```
b-chu, line 1:
```suggestion
# Copyright 2024 MosaicML Composer authors
```

tests/utils/test_remote_uploader.py
```diff
@@ -20,11 +20,14 @@ class DummyObjectStore(ObjectStore):
     """Dummy ObjectStore implementation that is backed by a local directory."""
 
     def __init__(self, **kwargs: Dict[str, Any]) -> None:
-        self.tmp_dir = tempfile.TemporaryDirectory()
+        self.tmp_dir = self.get_tmp_dir()
         self.root = self.tmp_dir.name
         self.sleep_sec = 0
         self.dest_filename = ''
 
+    def get_tmp_dir(self):
```
b-chu, line 28: Why is this necessary instead of just using it directly? Other files (e.g. tests/trainer/test_checkpoint.py) have it as an inline private function

b-chu (commented):
Left some comments, thanks Ning!

composer/callbacks/checkpoint_saver.py
b-chu: I see the validate_credentials function later, does that solve this?

composer/callbacks/checkpoint_saver.py
bigning: yes, there is a checkpoint v2 task: corrupted checkpoint loading, it will do two things:
1. update the symlink to point to the latest complete checkpoint
2. set the overwrite to true after loading the checkpoint.

composer/callbacks/checkpoint_saver.py
bigning: Yes!

composer/callbacks/checkpoint_saver.py
bigning: No, it's just used to distinguish the local symlink file saved in the temporary path.

composer/utils/file_helpers.py
bigning: i'd like to keep it same as current behavior. :)

tests/utils/test_remote_uploader.py
bigning: it's for mock and overwrite the temp path [here](https://github.com/mosaicml/composer/pull/3376/files#diff-17a1d776b94db0772f9f529b7506f222bc77c571df37cf4974099bac5ec0c128R799)

eracah (changes requested):
will follow up with my explanation the changes, but will hold for now

composer/callbacks/checkpoint_saver.py
```diff
@@ -320,6 +383,43 @@ def __init__(
...
+            if backend == 'wandb':
+                raise NotImplementedError(
+                    f'There is no implementation for WandB via URI. Please use '
+                    'WandBLogger with log_artifacts set to True.',
+                )
+            elif backend not in ['s3', 'oci', 'gs', 'azure', 'dbfs']:
+                raise NotImplementedError(
+                    f'There is no implementation for the cloud backend {backend} via URI. Please use '
+                    'one of the supported object stores (s3, oci, gs, azure, dbfs).',
+                )
```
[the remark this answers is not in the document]
bigning: good catch, yeah, these can be skipped.

composer/callbacks/checkpoint_saver.py
b-chu: Is this PR's behavior enabled in this PR? Or will you enable this PR's behavior in the corrupted checkpoint loading PR?

b-chu: Discussed offline, feature will be in production once this is merged

composer/callbacks/checkpoint_saver.py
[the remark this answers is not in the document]
bigning: this is actually required by a unit test to early return if there is unsupported backend. I have removed other upload logic in checkpoint_saver.

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +628,61 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
...
+                continue
+            if check_remote_files_exist_future.done():
+                result = check_remote_files_exist_future.result()
+                if result == RemoteFilesExistingCheckStatus.EXIST:
+                    self.remote_uploader.upload_file_async(
+                        remote_file_name=remote_symlink_file,
+                        file_path=local_symlink_file,
+                        overwrite=True,
+                    )
+                    break
```
eracah, line 656: wait why a break here? don't we want to check on the other futures

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +628,61 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
...
+                else:
+                    raise RuntimeError(f'Failed to check if checkpoint files upload finish: {result}')
+        self.symlink_upload_tasks = undone_symlink_upload_tasks
+
+    def fit_end(self, state: State, logger: Logger) -> None:
+        del state, logger  # unused
+        if self.remote_uploader is None:
+            return
+        log.info('Waiting checkpoint uploading finish')
```
eracah, line 665:
```suggestion
        log.info('Waiting for checkpoint uploading to finish')
```

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +628,61 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
...
+        self.remote_uploader.wait()
+        if self.rank_saves_remote_symlinks and len(self.symlink_upload_tasks) > 0:
+            log.debug('Uploading the last symlink file')
```
eracah, line 668:
```suggestion
            log.debug('Uploading symlink to the latest checkpoint')
```

eracah (commented):
Hey clever solution! I have a bunch of comments/questions

composer/callbacks/checkpoint_saver.py
```diff
@@ -287,6 +290,9 @@ def __init__(
         num_checkpoints_to_keep: int = -1,
         weights_only: bool = False,
         ignore_keys: Optional[Union[list[str], Callable[[dict], None]]] = None,
+        save_folder: str = '',
```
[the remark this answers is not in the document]
bigning: let's say the save_folder is "s3://bucket/path". The folder is just "path", because originally the rud was constructed out side of this class. But with this PR we need to build the backend inside this class, so have to pass in the full remote path with the backend name.

composer/callbacks/checkpoint_saver.py
```diff
@@ -320,6 +326,34 @@ def __init__(
...
+            if backend == 'wandb':
+                raise NotImplementedError(
+                    f'There is no implementation for WandB via URI. Please use '
+                    'WandBLogger with log_artifacts set to True.',
+                )
+            elif backend not in ['s3', 'oci', 'gs', 'azure', 'dbfs']:
+                raise NotImplementedError(
+                    f'There is no implementation for the cloud backend {backend} via URI. Please use '
+                    'one of the supported object stores (s3, oci, gs, azure, dbfs).',
+                )
```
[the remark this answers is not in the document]
bigning: there is a unit test which required to raise these exception, https://github.com/mosaicml/composer/blob/5f8265d88a47f27d3d1340a1fa0f625c51ee270f/tests/trainer/test_checkpoint.py#L336-L339. So just for safety, i added the same exception here.

composer/callbacks/checkpoint_saver.py
```diff
@@ -320,6 +326,34 @@ def __init__(
...
+            self.remote_uploader = RemoteUploader(
+                remote_folder=save_folder,
```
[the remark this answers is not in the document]
bigning: explained above, the save_folder doesn't have the backend information.

composer/callbacks/checkpoint_saver.py
```diff
@@ -432,7 +488,13 @@ def _save_checkpoint(self, state: State, logger: Logger):
         )
         log.debug(f'Checkpoint locally saved to {saved_path}')
 
+        self.symlink_count += 1
```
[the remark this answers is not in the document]
bigning: because now we upload symlink much later after the checkpoint files uploading finish. So there might be multiple local symlink files in the same tmp directory. I need to name them differently, so use the symlink_count to do that.

composer/callbacks/checkpoint_saver.py
```diff
@@ -432,7 +488,13 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+        local_remote_file_names = []
```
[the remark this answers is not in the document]
bigning: lol. here local refers to the local rank

composer/callbacks/checkpoint_saver.py
```diff
@@ -455,8 +517,8 @@ def _save_checkpoint(self, state: State, logger: Logger):
                 src_path = str(pathlib.Path(saved_path).parent)
             else:
                 src_path = saved_path
-            this_rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
-            if this_rank_saves_symlinks:
+            rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
```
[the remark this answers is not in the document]
bigning: yeah, i think for both monolithic and fsdp sharded, that's right, unless we save to local path on a different rank. Not familiar with deepspeed though, i can make this change if you think it's safe.

composer/callbacks/checkpoint_saver.py
```diff
@@ -509,17 +580,33 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                if state.fsdp_sharded_state_dict_enabled:
+                    src_path = str(pathlib.Path(remote_file_name).parent)
+                else:
+                    src_path = remote_file_name
+                log.debug(f'Creating symlink file {symlink_filename} -> {src_path}')
+                rank_saves_symlinks = dist.get_global_rank() == 0 or not state.fsdp_sharded_state_dict_enabled
+                if rank_saves_symlinks:
+                    self.rank_saves_remote_symlinks = True
+                    create_symlink_file(src_path, symlink_filename)
+                    if self.remote_uploader is not None:
```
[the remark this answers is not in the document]
bigning: >  It is set in the constructor

the remote_uploader is None if the save_folder is not a remote path.

composer/callbacks/checkpoint_saver.py
```diff
@@ -509,17 +580,33 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                        remote_checkpoint_file_names = []
+                        for file_names in all_remote_filenames:
+                            remote_checkpoint_file_names += file_names
```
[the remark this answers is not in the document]
bigning: oh because `dist.all_gather_object` returns a list of objects. so `all_remote_filenames` is a list of list, `remote_checkpoint_file_names` is a flatten list.

composer/callbacks/checkpoint_saver.py
bigning: because i traverse the **reverse** list of tasks (line 641). If there are multiple symlink files to upload, and one of them finished the dependency (checkpoint files upload finish), then we only need to upload that symlink file, we don't need to upload symlinks that are earlier than it.

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +628,61 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
...
+
+    def fit_end(self, state: State, logger: Logger) -> None:
+        del state, logger  # unused
+        if self.remote_uploader is None:
+            return
+        log.info('Waiting checkpoint uploading finish')
+        self.remote_uploader.wait()
+        if self.rank_saves_remote_symlinks and len(self.symlink_upload_tasks) > 0:
+            log.debug('Uploading the last symlink file')
+            check_remote_files_exist_future, local_symlink_file, remote_symlink_file = self.symlink_upload_tasks[-1]
```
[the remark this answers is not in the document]
bigning: will do.

composer/callbacks/checkpoint_saver.py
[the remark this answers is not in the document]
bigning: it actually failed the deepspeed test, i changed it back. :)

composer/callbacks/checkpoint_saver.py
```diff
@@ -542,3 +626,62 @@ def _rotate_checkpoints(self, sharding_enabled: bool = False):
...
+
+    def fit_end(self, state: State, logger: Logger) -> None:
+        del state, logger  # unused
+        if self.remote_uploader is None:
+            return
+        log.info('Waiting for checkpoint uploading to finish')
+        self.remote_uploader.wait()
+        if self.rank_saves_symlinks and len(self.symlink_upload_tasks) > 0:
+            log.debug('Uploading symlink to the latest checkpoint')
+            # We only need to upload the latest symlinke file, ignoring the old ones
```
eracah, line 667:
```suggestion
            # We only need to upload a symlink pointing to the latest checkpoint files, so we can ignore successful uploads of older checkpoints.
```

composer/callbacks/checkpoint_saver.py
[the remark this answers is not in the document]
eracah: this logic should be inside the RemoteUploader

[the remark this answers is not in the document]
eracah: the checkpointsaver should be agnostic to the supported backends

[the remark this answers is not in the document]
eracah: please put this code in the `remoteuploader` constructor

composer/utils/remote_uploader.py
```diff
@@ -84,18 +131,69 @@ def __init__(
...
+        elif self.remote_backend_name == 'azure':
+            self.remote_backend_name = 'libcloud'
+            self.backend_kwargs = {
+                'provider': 'AZURE_BLOBS',
+                'container': self.remote_bucket_name,
+                'key_environ': 'AZURE_ACCOUNT_NAME',
+                'secret_environ': 'AZURE_ACCOUNT_ACCESS_KEY',
+            }
+        elif self.remote_backend_name == 'dbfs':
+            self.backend_kwargs['path'] = self.path
```
eracah, line 150: Put this logic from the `CheckpointSaver` constructor here: 
```if backend != '':
            if backend == 'wandb':
                raise NotImplementedError(
                    f'There is no implementation for WandB via URI. Please use '
                    'WandBLogger with log_artifacts set to True.',
                )
            elif backend not in ['s3', 'oci', 'gs', 'azure', 'dbfs']:
                raise NotImplementedError(
                    f'There is no implementation for the cloud backend {backend} via URI. Please use '
                    'one of the supported object stores (s3, oci, gs, azure, dbfs).',
                )
```

eracah (approved):
LGTM

mvpatel2000 (approved):
Approving to unblock

PR #3384, opened 2024-06-07
Remove pydocstyle and configure ruff check for D rules

eracah:
Our pydocstyle did the following:
* used "google" style, which enforces all D rules except D203, D204, D213, D215, D400, D401, D404, D406, D407, D408, D409 and D413
* ignores (adds to the the "google" style ignore list) D102, D105, D107, D401
* selects (aka removes from the "google" style ignore list) D400, D404

To enable the above functionality in ruff, we:
* select all D rules
* ignore: D102, D105, D107, D203, D204, D213, D215, D401,  D406, D407, D408, D409, D413


mvpatel2000 (approved):
LGTM

pyproject.toml
```diff
@@ -27,7 +26,32 @@ ignore = [
...
 ]
+
+[tool.ruff.lint.per-file-ignores]
+".ci*" = ["D"]
+"tests/*" = ["D"]
+"composer/algorithms/*" = ["D"]
+"composer/datasets/*" = ["D"]
+"composer/models/*" = ["D"]
+"composer/trainer/activation_checkpointing.py" = ["D"]
+
```
Skylion007, line 54:
```suggestion
[tool.ruff.lint.pydocstyle]
convention = "google"

```
Should reduce the number of rules you need to ignore

pyproject.toml
```diff
@@ -27,7 +26,32 @@ ignore = [
...
+    "D103"
 ]
+
+[tool.ruff.lint.per-file-ignores]
+".ci*" = ["D"]
+"tests/*" = ["D"]
+"composer/algorithms/*" = ["D"]
+"composer/datasets/*" = ["D"]
+"composer/models/*" = ["D"]
+"composer/trainer/activation_checkpointing.py" = ["D"]
```
Skylion007, line 53: You can also just add a specific noqa comment at the top of the file to disable it directly in the file.

Skylion007:
Our ruff version is really old, we may want to consider updating it. It may include more pydocstyle autofixes in newer versions or reduce any false positives / false negatives

composer/utils/module_surgery.py
```diff
@@ -14,7 +14,7 @@
         instances of the module type have been seen. The function should return a replacement
         :class:`torch.nn.Module` if the module type should be replaced, or ``None`` otherwise.
 
-        Args:
+Args:
             module (torch.nn.Module): Source module
```
b-chu, line 18: Weird indenting here

pyproject.toml
b-chu: Why ignore algorithms/datasets/models?

eracah: I just used the same ignoring that pydocstyle used. See the lines removed from the the pre-commit yaml

pyproject.toml
```diff
@@ -27,7 +26,34 @@ ignore = [
...
+    "D213",
+    "D215",
+    "D401",
+    "D406",
+    "D407",
+    "D408",
+    "D409",
+    "D413",
+    "D417",
+    "D103"
```
eracah, line 44:
```suggestion
```

pyproject.toml
```diff
@@ -15,8 +15,7 @@ split_on_trailing_comma = true
 [tool.ruff.lint]
 select = [
     "C4",
-    # TODO port pydocstyle
-    # "D", # pydocstyle
+    "D",
```
eracah, line 18:
```suggestion
    "D400",
    "D404",
```

eracah (commented):
things to change to make it work with "google" style

composer/utils/module_surgery.py
eracah: is it weird?

b-chu: module is double indented after Args

PR #3389, opened 2024-06-09
Only requires `databricks-sdk` when inside the Databricks platform

antoinebrl:
# What does this PR do?

We would like to use MLFlow outside of the databricks platform. For such scenario, the package `databricks-sdk` don't have to be installed. With this contribution the import of the `databricks.sdk` is gated by the presence of the `DATABRICKS_TOKEN` environment variable, making it possible to reduce the dependency chain.


mvpatel2000 (approved):
LGTM!

PR #3393, opened 2024-06-10
Skip extra dataset state load

mvpatel2000:
# What does this PR do?

Skip extra dataset state load. Previously, if loading a checkpoint with `train_dataloader`, we would first load the dataset_state on load. After loading, if we again set train_dataloader, it would call `load_state_dict` with a state_dict that had a value of `None`. This fixes the check on the setter of `train_dataloader` to properly skip the extra load.

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -708,7 +708,7 @@ class State(Serializable):
         """
         self._train_dataloader = train_dataloader
         # Load dataset state from checkpoint when train_dataloader is set
-        if self.dataset_state:
+        if self.dataset_state is not None and self.dataset_state.get('train') is not None:
             dataset = self._dataset_of(self._train_dataloader)
             if hasattr(dataset, 'load_state_dict'):
                 dataset.load_state_dict(self.dataset_state['train'])  # pyright: ignore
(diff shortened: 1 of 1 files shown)


composer/core/state.py
```diff
@@ -1278,14 +1278,14 @@ def _load_dataset_state(self, obj: dict[str, Any]) -> None:
...
         """
-        self.dataset_state = obj
-
         dataset = self._dataset_of(self.train_dataloader)
         if hasattr(dataset, 'load_state_dict'):
             dataset.load_state_dict(obj['train'])  # pyright: ignore
             obj['train'] = None
             self.dataset_resumption['train'] = True
 
+        self.dataset_state = obj
```
snarayan21, line 1287: I don't think anything changes if we move this line down here, since `obj['train']` will be `None` regardless (changing keys of the same dict, and it's not copied).

snarayan21: Also, just so I'm understanding correctly -- we set `obj['train']` to `None` after loading in the dataset state so that we know that we've already loaded in the dataset state, right?

composer/core/state.py
```diff
@@ -708,7 +708,7 @@ def train_dataloader(self, train_dataloader: Optional[Union[Iterable, DataLoader
         """
         self._train_dataloader = train_dataloader
         # Load dataset state from checkpoint when train_dataloader is set
-        if self.dataset_state:
+        if self.dataset_state is not None and self.dataset_state.get('train') is not None:
```
snarayan21, line 711: Can you add a clarifying comment here why we check for `self.dataset_state.get('train') is not None`?

snarayan21 (commented):
makes sense to me if I'm understanding this right. couple comments. thanks!

composer/core/state.py
mvpatel2000: Yes to understanding.

yes doesn't change anything, I think just easier to follow logic

snarayan21 (approved):
lgtm

PR #3395, opened 2024-06-12
Remove FSDP restriction from PyTorch 1.13

mvpatel2000:
# What does this PR do?

Remove FSDP restriction from PyTorch 1.13. We added this to give a friendlier error for users when PyTorch had a bug, but the corresponding issue is closed and fixed and has been patched for a while since PyTorch 2.

### DIFF
--- a/composer/distributed/dist_strategy.py
+++ b/composer/distributed/dist_strategy.py
@@ -328,7 +328,7 @@ def prepare_fsdp_module(
 
     mixed_precision = fsdp_config.mixed_precision
     keep_low_precision_grads = fsdp_config.keep_low_precision_grads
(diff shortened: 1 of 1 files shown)


snarayan21 (commented):
lgtm, but can you include a run name that successfully runs with torch 1.3 and fp32 param_dtype?

mvpatel2000:
> lgtm, but can you include a run name that successfully runs with torch 1.3 and fp32 param_dtype?

It does not run with 1.13, but we've moved off from that and upgraded

snarayan21 (approved):
ah i misunderstood. lgtm

b-chu (approved):
Thanks!

PR #3396, opened 2024-06-12
Busy wait utils in dist

dakinggg:
# What does this PR do?
This PR adds additional utils for busy waiting on a node using a signal file lock. The main difference between the added utils and the current approach used is that the signal file has a randomly generated identified appended to it, to better support use cases that include multiple runs using a shared file system that may interfere with each other. Future PRs will replace the signal file uses in Composer (and after release Foundry) with these utils, but just want to start with the util implementation to keep PRs small.

Related to https://github.com/mosaicml/llm-foundry/pull/1253

# Before submitting

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -642,9 +642,9 @@ def get_node_signal_file_name(rng: Optional[random.Random] = None) -> str:
         rng = random.Random()
 
     random_string = ''.join(rng.choices(string.ascii_letters + string.digits, k=6))
-    file_name = f'._signal_file_{random_string}'
-    file_name = dist.broadcast_object_list([file_name], src=0)[0]
-    return file_name
+    file_name_list = [f'._signal_file_{random_string}']
+    dist.broadcast_object_list(file_name_list, src=0)
+    return file_name_list[0]
 
 
 def write_signal_file(signal_file_name: str, dir_path: Optional[str] = None) -> str:
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -692,7 +692,7 @@ def busy_wait_for_local_rank_zero(dir_path: Optional[str] = None):
 
     # Wait for the signal file to be created by local rank zero
     with local_rank_zero_download_and_wait(signal_file_path):
-        # Sync up all ranks
+        # Sync all ranks across nodes as busy wait only is within node
         dist.barrier()
 
     # Remove the signal file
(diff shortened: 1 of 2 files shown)


mvpatel2000 (commented):
can you update calls in composer to use these helper fns?

composer/utils/dist.py
```diff
@@ -627,6 +629,76 @@ def get_sampler(
...
+    Returns:
+        str: The name of the file that will be created to signal the end of a node's training.
+    """
+    if rng is None:
+        rng = random.Random()
+
+    random_string = ''.join(rng.choices(string.ascii_letters + string.digits, k=6))
+    file_name_list = [f'._signal_file_{random_string}']
+    dist.broadcast_object_list(file_name_list, src=0)
+    return file_name_list[0]
```
mvpatel2000, line 647: this returns the global rank 0? is that intnded

mvpatel2000: iirc we want different names per node to avoid shared file system cases

composer/utils/dist.py
```diff
@@ -627,6 +629,76 @@ def get_sampler(
...
+def write_signal_file(signal_file_name: str, dir_path: Optional[str] = None) -> str:
+    """Writes a signal file to the specified directory.
+
+    This function creates a signal file in the specified directory. If the directory does
+    Note: Only local rank zero writes the signal file. All other ranks are expected to wait for the signal file.
+
+    Args:
+        signal_file_name (str): The name of the signal file.
+        dir_path (str, optional): The directory in which to create the signal file. If ``None``,
+            the current working directory will be used.
```
mvpatel2000, line 659: should specify its an absolute dir path

composer/utils/dist.py
dakinggg: ah thank you, i waffled on this but couldn't remember if there was a reason why they were supposed to be node specific. will adjust.

dakinggg: fixed

composer/utils/dist.py
dakinggg: done

composer/utils/dist.py
```diff
@@ -627,6 +629,77 @@ def get_sampler(
...
+
+    # All ranks yield execution to allow local rank zero to run the code it needs to
+    yield
+
+    # Local rank zero writes the signal file, all other rank just get the expected path
+    signal_file_path = write_signal_file(signal_file_name=signal_file_name, dir_path=dir_path)
+
+    # Wait for the signal file to be created by local rank zero
+    with local_rank_zero_download_and_wait(signal_file_path):
+        # Sync up all ranks
```
mvpatel2000, line 695:
```suggestion
        # Sync all ranks across nodes as busy wait only is within node
```

mvpatel2000 (approved):
LGTM. Minor comment

PR #3398, opened 2024-06-12
Simplify launcher world size parsing

mvpatel2000:
# What does this PR do?

Simplify launcher world size parsing. Now, `composer -n` correctly enables running on fewer GPUs when iterating on a single machine.

### DIFF
--- a/composer/cli/launcher.py
+++ b/composer/cli/launcher.py
@@ -197,8 +197,13 @@ def _parse_args():
     if args.nproc < 1:
         raise ValueError('The nproc must be 1 or greater')
 
-    if args.world_size is None and 'WORLD_SIZE' in os.environ:
-        args.world_size = int(os.environ['WORLD_SIZE'])
+    if args.world_size is None:
+        if 'WORLD_SIZE' in os.environ and os.environ.get('LOCAL_WORLD_SIZE') != os.environ['WORLD_SIZE']:
+            # Use WORLD_SIZE env var if set and running multinode. Otherwise, default to nproc
+            # to enable easy overriding of number of processes when on a single node.
+            args.world_size = int(os.environ['WORLD_SIZE'])
+        else:
+            args.world_size = args.nproc
 
     if args.base_rank is None and 'BASE_RANK' in os.environ:
         args.base_rank = int(os.environ['BASE_RANK'])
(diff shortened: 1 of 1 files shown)


composer/cli/launcher.py
```diff
@@ -197,8 +197,13 @@ def _parse_args():
     if args.nproc < 1:
         raise ValueError('The nproc must be 1 or greater')
 
-    if args.world_size is None and 'WORLD_SIZE' in os.environ:
-        args.world_size = int(os.environ['WORLD_SIZE'])
+    if args.world_size is None:
+        if 'WORLD_SIZE' in os.environ and os.environ.get('LOCAL_WORLD_SIZE') != os.environ['WORLD_SIZE']:
```
eracah, line 201: tiniest of nits: should store this boolean expression as a variable: `using_multinode`. Only would be useful if there was more logic for multinode vs. not

eracah (approved):
LGTM.

eracah:
Any way we can do a test for this?

composer/cli/launcher.py
mvpatel2000: It's a little tricky since we define multinode variable later on. Technicallyyy you can define WOLRD_SIZE and then manually set `LOCAL_WORLD_SIZE` via nproc, so this wouldn't always work.

mvpatel2000:
> Any way we can do a test for this?

Unfortunately not easily since pytest runs in a single instance and this would require a separate launch. But ill include a manual test

mvpatel2000:
can someone put a request changes hold to block during codefreeze? @dakinggg ?

dakinggg (dismissed):
Holding temporarily

bigning:
I'm curious, why set env_var `WORLD_SIZE` for single node in the first place? It breaks the CPU test `python -m pytest` in interactive if there are mutliple GPUs. Error is missing `RANK` in env_var.  This PR doesn't fix that, because [here](https://github.com/mosaicml/composer/blob/ba82cc9ec943e316da59ec6ec0ba540008c9ca1f/tests/fixtures/autouse_fixtures.py#L75) it's not 1.  I had to unset the `WORLD_SIZE`

mvpatel2000:
> I'm curious, why set env_var `WORLD_SIZE` for single node in the first place? 

was bug in mcloud, it should set all env vars even if on single node

>It breaks the CPU test `python -m pytest` 

will investigate

mvpatel2000:
> > It breaks the CPU test `python -m pytest`
> 
> will investigate

Discussed offline, if you dont use the launcher it will still use the WORLD_SIZE env var which we cannot fix. This break is not related to this PR, you must use `composer -n 1 pytest...`

dakinggg (approved):

PR #3399, opened 2024-06-13
Save checkpoint to disk for API with new save layout

eracah:
# What does this PR do?
1. This PR introduces an API for saving the common objects one would save in a checkpoint:
   * save_model_to_disk
   * save_optim_to_disk
   * save_composer_metadata_to_disk
   * save_resumption_state_to_disk
These functions make sure to save each object into its own directory

2. Introduces a dataclass, `CheckpointSaveOptions` for configuring checkpoint saving
3. Introduces a save to disk workhorse function, `save_checkpoint_to_disk` that handles saving all components of a checkpoint. Uses the API in 1. to save each component into its own separate file

Extras:
* adds an `init_state` helper function for tests to initialize a State object for tests


composer/checkpoint/save.py
```diff
@@ -3,19 +3,293 @@
...
+    overwrite: bool = False
+    save_model: bool = True
+    save_optimizer: bool = True
+    save_resumption_state: bool = True
+    num_checkpoints_to_keep: int = -1
+    save_format: str = 'pt'
+    sharded_checkpoint: bool = False
+    precision: str = 'bf16'
+    # High level objects to save or not save
+    # e.g. 'model', 'optim', 'schedulers', 'rng' etc.
```
mvpatel2000, line 80:
```suggestion
```
docstring covers this?

mvpatel2000 (commented):
LGTM, would be helpful to have someone check though, many moving parts here

bigning (approved):

PR #3413, opened 2024-06-19
Bump version v0.23.3

karan6181:
# What does this PR do?
- Bump version to v0.23.3

### DIFF
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.24.0.dev0'
+__version__ = '0.23.3'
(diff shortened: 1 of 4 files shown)


composer/_version.py
```diff
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.24.0.dev0'
+__version__ = '0.23.3'
```
snarayan21, line 6: isn't this supposed to still be 0.24.dev0, and then only on the release branch you change it to 0.23.0? pretty sure that's what's in the doc

snarayan21 (commented):
need to verify what the version should be on the PR merged to dev

composer/_version.py
bigning: @mvpatel2000 updated the doc to fix the release note auto-generation. Now it needs this change and an extra PR to restore the version to dev

karan6181:
Closing this PR against https://github.com/mosaicml/composer/pull/3414 since the PR source branch must be from the main repo and not from the fork; otherwise, docker credentials won't work.

PR #3431, opened 2024-06-27
Remove save overwrite

mvpatel2000:
# What does this PR do?

Previously, we required `save_overwrite=False` as it was not clear if a user intended to overwrite or resume from a checkpoint. 

However, it turns out this situation is well defined. Autoresume = resume from symlink file. It is a transformation on load_path, and nothing else. So, you can actually set autoresume on and save_overwrite to True, which would be useful for partial checkpoint uploads where the symlink file is not added.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1732,11 +1732,6 @@ class Trainer:
             error_message = ''
             if save_folder is None:
                 error_message += 'The `save_folder` must be specified when autoresume is enabled. '
-            if save_overwrite:
-                error_message += textwrap.dedent(
-                    'The flag `save_overwrite` must be False when autoresume is enabled as autoresume always loads the '
-                    'latest existing checkpoint in `save_folder`. ',
-                )
             if save_latest_filename is None:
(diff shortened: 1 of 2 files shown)


b-chu (approved):
Need approval from either Ning or Evan. Did we not test that save_overwrite=True and autoresume=True threw an error before?

composer/trainer/trainer.py
```diff
@@ -1732,11 +1732,6 @@ def __init__(
             error_message = ''
             if save_folder is None:
                 error_message += 'The `save_folder` must be specified when autoresume is enabled. '
-            if save_overwrite:
-                error_message += textwrap.dedent(
-                    'The flag `save_overwrite` must be False when autoresume is enabled as autoresume always loads the '
-                    'latest existing checkpoint in `save_folder`. ',
-                )
```
eracah, line 1739: maybe add a warning here instead?

eracah (commented):
looks good to me besides maybe adding a warning.

eracah:
oh too late

PR #3434, opened 2024-06-27
Correctly process `parallelism_config['tp']` when it's a dict

snarayan21:
# What does this PR do?
This fixes a bug where if the TP configuration (specified through `parallelism_config['tp']`) was passed in as a dict, it would not be correctly processed into the final `parallelism_config`. This also means that whenever we were specifying the TP config as a dict before, it was not actually being set.

Modified unit tests to make sure the TP config is actually being set.

# Before submitting


mvpatel2000 (dismissed):
oof good catch

snarayan21:
@mvpatel2000 I keep getting errors on 4 gpu tests because it's detecting the `FutureWarning` and counting that as an error. I've tried filtering out these warnings but that doesn't seem to be working, including adding the warning to the pyproject.toml config for pytest. What's the right way to filter these?

mvpatel2000:
> @mvpatel2000 I keep getting errors on 4 gpu tests because it's detecting the `FutureWarning` and counting that as an error. I've tried filtering out these warnings but that doesn't seem to be working, including adding the warning to the pyproject.toml config for pytest. What's the right way to filter these?

pytest.mark.filterwarning?

snarayan21:
@mvpatel2000 I already added this `@pytest.mark.filterwarnings('ignore:.*(TP) is experimental.*:FutureWarning')` on all relevant tests...

mvpatel2000:
> @mvpatel2000 I already added this `@pytest.mark.filterwarnings('ignore:.*(TP) is experimental.*:FutureWarning')` on all relevant tests...

I think its a regex issue, should be `\(`?

snarayan21:
ah that makes sense @mvpatel2000 lemme try

mvpatel2000 (approved):
LGTM

PR #3435, opened 2024-06-27
Relax hf hub pin

dakinggg:
Updates the HF hub pin to permit the breaking changes of 0.23. The simplest way to do this was to also update transformers to permit 0.41.x, which has a minimum hf hub pin of 0.23. Also something changed in bert that prevents random masking (not entirely sure what this did before anyway, but now it produces nan), so changed to the more expected mask that just masks padding at the end of the sequence. The default bert also introduced an op that requires onnx opset 14, so bumped the tests to that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,7 +161,7 @@ filterwarnings = [
     # Ignore save_state_dict / load_state_dict deprecation warnings
     '''ignore:'.*_state_dict' is deprecated and will be removed in future versions.*:UserWarning''',
     # Ignore HF deprecation which affects their own libraries
-    '''ignore:'.*`resume_download` is deprecated and will be removed in version.*:FutureWarning'''
+    '''ignore:'.*Downloads always resume when possible..*:FutureWarning'''
 ]
 
 # Coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -160,8 +160,6 @@ filterwarnings = [
     '''ignore:.*an autograd kernel was not registered to the Autograd key.*:UserWarning''',
     # Ignore save_state_dict / load_state_dict deprecation warnings
     '''ignore:'.*_state_dict' is deprecated and will be removed in future versions.*:UserWarning''',
-    # Ignore HF deprecation which affects their own libraries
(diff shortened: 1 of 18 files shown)


pyproject.toml
```diff
@@ -160,8 +160,6 @@ filterwarnings = [
     '''ignore:.*an autograd kernel was not registered to the Autograd key.*:UserWarning''',
     # Ignore save_state_dict / load_state_dict deprecation warnings
     '''ignore:'.*_state_dict' is deprecated and will be removed in future versions.*:UserWarning''',
-    # Ignore HF deprecation which affects their own libraries
```
snarayan21, line 163: this was just old right? not doing anything anyways

setup.py
```diff
@@ -179,9 +179,9 @@ def package_files(prefix: str, directory: str, extension: str):
 ]
 
 extra_deps['nlp'] = [
-    'transformers>=4.11,!=4.34.0,<4.42',
+    'transformers>=4.11,!=4.34.0,<4.43',
```
snarayan21, line 182: i'm assuming there's some dark reason we don't use 4.34.0...

setup.py
```diff
@@ -179,9 +179,9 @@ def package_files(prefix: str, directory: str, extension: str):
...
     'datasets>=2.4,<3',
-    'huggingface-hub>=0.21.2,<0.23',
+    'huggingface-hub>=0.21.2,<0.24',
```
snarayan21, line 184: just noting, once streaming gets that PR in, people who want to use HF datasets with MDS will use hf 0.23.1 and above. so this is chill

snarayan21 (approved):
lfg. just some minor non blocking comments

pyproject.toml
dakinggg: it wasn't actually working for some reason...but the warnings go away entirely on the newest transformers versions so irrelevant now

setup.py
dakinggg: It was very broken lol https://github.com/mosaicml/composer/pull/2635

mvpatel2000 (approved):

PR #3436, opened 2024-06-28
Lower the system metrics logging frequency to reduce MLflow server's load

chenmoneygithub:
# What does this PR do?

Lower the system metrics logging frequency to reduce MLflow server's load. System metrics logging is using a different async logging queue from normal metrics, so the time buffering doesn't apply to system metrics. This design was by intention so that MLflow can terminate system metrics separately. 30s is the logging frequency of wandb, so there should be no user experience regression.

# Before submitting

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -123,6 +123,13 @@ class MLFlowLogger(LoggerDestination):
         if logging_buffer_seconds:
             os.environ['MLFLOW_ASYNC_LOGGING_BUFFERING_SECONDS'] = str(logging_buffer_seconds)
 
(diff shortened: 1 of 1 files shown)


chenmoneygithub:
@mvpatel2000 please let me know if you prefer to have it be a configurable argument, thanks!

mvpatel2000 (approved):
LGTM. Every 30 seconds seems a bit infrequent... what about 10s?

chenmoneygithub:
@mvpatel2000 Thanks for reviewing! I am shortening the gap to 15s, which matches the wandb behavior: Lower the system metrics logging frequency to reduce MLflow server's load

PR #3447, opened 2024-07-01
[checkpoint v2] Download api

bigning:
adding the download API for monolithic checkpoint


composer/checkpoint/download.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Useful functions for load checkpoints from remote object store or local disk."""
+import logging
+from typing import Optional
```
mvpatel2000, line 6:
```suggestion
"""Useful functions to load checkpoints from remote object store or local disk."""


import logging
from typing import Optional
```
```suggestion
"""Useful functions for load checkpoints from remote object store or local disk."""
import logging
from typing import Optional
```

composer/checkpoint/download.py
```diff
@@ -0,0 +1,84 @@
...
+            object_name=source_path,
+            filename=destination_path,
+        )
+
+    log.debug(f'Downloading {source_path} to {destination_path}')
+    _download()
+    log.debug(f'Finished downloading {source_path} to {destination_path}')
+
+
+def download_monolithic_checkpoint(
```
mvpatel2000, line 60: should this be all local rank 0 or global rank 0?

bigning: global rank 0, updated the PR to make it more clear. I had a discussion with @eracah  on this, we looked the current code, it works like this: for fsdp + monolithic, only global rank 0 load the state dict, then for non-model state dict, global rank0 broadcast to other ranks. For model state dict, it uses the fsdp sync_modules_states to broadcast

mvpatel2000 (approved):
would also like review from @eracah

tests/checkpoint/test_download.py
```diff
@@ -0,0 +1,61 @@
...
+    use_fsdp = False
+    if world_size > 1:
+        use_fsdp = True
+    fsdp_model, _ = init_model(use_fsdp=use_fsdp)
+
+    if use_fsdp:
+        with FSDP.state_dict_type(fsdp_model, StateDictType.FULL_STATE_DICT):
+            state = fsdp_model.state_dict()
+    else:
+        state = fsdp_model.state_dict()
```
eracah, line 35: you can also use `get_model_state_dict` function
```suggestion
    from torch.distributed.checkpoint.state_dict import get_model_state_dict, StateDictOptions
    get_model_state_dict(model, options=StateDictOptions(full_state_dict=True))
```

eracah (approved):
Very slick! LGTM

PR #3449, opened 2024-07-01
Add a monitor of mlflow logger so that it sets run status as failed if main thread exits unexpectedly

chenmoneygithub:
# What does this PR do?

Add a monitor of mlflow logger so that it sets run status as failed if main thread exits unexpectedly

# Before submitting


composer/loggers/mlflow_logger.py
```diff
@@ -33,6 +36,52 @@
...
+        if self.crash_event.is_set():
+            client = MlflowClient(self.mlflow_tracking_uri)
+            client.set_terminated(self.mlflow_run_id, status='FAILED')
+
+    def stop(self):
+        self.exit_event.set()
+
+    def crash(self):
+        self.crash_event.set()
+        self.exit_event.set()
```
mvpatel2000, line 82: @chenmoneygithub why do we need two events?

composer/loggers/mlflow_logger.py
```diff
@@ -528,21 +608,49 @@ def log_images(
 
     def post_close(self):
         if self._enabled:
+            if hasattr(self, 'monitor_process'):
+                # Check if there is an uncaught exception, which means `post_close()` is triggered
+                # due to program crash.
+                if isinstance(self._global_exception_occurred, torch.Tensor):
+                    finish_with_exception = (self._global_exception_occurred == 1).item()
+                else:
+                    finish_with_exception = (self._global_exception_occurred == 1)
```
mvpatel2000, line 617: can we simplify to use a bool flag and get rid of the tensor/move to device stuff

mvpatel2000:
Manual test:
`llama3-8b-2k-lf3dhE` failure
`llama3-8b-2k-Wax2MK` success

composer/loggers/mlflow_logger.py
chenmoneygithub: So `exit_event ` is for normal exit, and `crash_event` is for crash. If the global exception handler finds an error, it will tell the monitor thread that the main process fails and the monitor thread will set the MLflow run status as `FAILED`.

composer/loggers/mlflow_logger.py
chenmoneygithub: Yes I thought about that, right now I am using a `torch.Tensor` because to resolve the single node multi-gpu setup, we need to communicate the `self._global_exception_occurred` across GPUs. If we don't plan to look into that setup in a short term, happy to change it to a bool variable!

PR #3486, opened 2024-07-22
allow listed runners

KevDevSha:
# What does this PR do?

Updating to allow listed runners


.github/workflows/pr-cpu.yaml
```diff
@@ -9,36 +9,39 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/dev' }}
 jobs:
   pytest-cpu:
-    uses: mosaicml/ci-testing/.github/workflows/pytest-cpu.yaml@v0.0.9
+    name: ${{ matrix.name }}
+    runs-on: linux-ubuntu-latest
+    if: github.repository_owner == 'mosaicml'
```
eitanturok, line 14: Do we need this if statement here
```
if: github.repository_owner == 'mosaicml'
```

eitanturok (changes requested):

.github/workflows/pr-cpu.yaml
mvpatel2000: Yes, this is so forks don't run this code as well.

mvpatel2000 (approved):

DinnerMonster:
Will update merge once we get our emu composer repo cofigured for the container pull

KevDevSha:
Will update and merge once our EMU gh registry for composer is created to be used for the job container.

.github/workflows/pr-gpu.yaml
```diff
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
+  pull_request:
```
KevDevSha, line 3: change to pull_request_target before merge

PR #3487, opened 2024-07-22
Load Microbatches instead of Entire Batches to GPU 

JackZ-db:
Prior this PR, when `global batch size` increased, users saw that GPU memory usage increased even if `device_train_microbatch_size`, which is the actual parameter that determines how many samples a GPU sees at once, was held constant. This was caused in part by the fact that entire batches were being put on GPU's in the training loop using `batch_to_device`, instead of microbatches one at a time. 

This PR ensures that global batch size doesn't affect memory usage so long as `device_train_microbatch_size` remains constant. Instead of splitting the batch into microbatches after the entire batch is loaded onto the GPU, the batch is first split in the CPU, before loading the microbatches one at a time onto the GPU.


composer/algorithms/selective_backprop/selective_backprop.py
```diff
@@ -272,6 +272,8 @@ def apply(self, event: Event, state: State, logger: Optional[Logger] = None) ->
                     raise RuntimeError('Model must be of type ComposerModel')
                 self._loss_fn = state.model.loss
             return
+
+        state.batch = state.device.batch_to_device(state.batch)
```
mvpatel2000, line 276: why is this necessary?

composer/devices/device.py
```diff
@@ -90,6 +90,20 @@ def optimizer_to_device(self, optimizer: Optimizer) -> Optimizer:
                     state[k] = self.tensor_to_device(v)
         return optimizer
 
+    def decode_batch_to_dict(self, batch: T_Batch) -> T_Batch:
```
mvpatel2000, line 93: why is this necessary?

composer/algorithms/selective_backprop/selective_backprop.py
JackZ-db: selective backprop forwards the whole batch to determine what fraction of the minibatch to use for gradient calc - unit tests also break otherwise

composer/devices/device.py
JackZ-db: this method doesn't change the content or shape or anything, but without this method, the batch type is sometimes `'transformers.tokenization_utils_base.BatchEncoding'>`, which `get_num_samples_in_batch` and `get_num_tokens_in_batch` can't handle. this just converts it to a dict mapping.

composer/trainer/trainer.py
```diff
@@ -3461,6 +3462,8 @@ def _eval_loop(
                     try:
                         microbatches = data_spec.split_batch(device_batch, evaluator.device_eval_microbatch_size)
                         for i, self.state.batch in enumerate(microbatches):
+                            self.state.batch = self.state.device.batch_to_device(self.state.batch)
+                            self.state.batch = data_spec.device_transforms(self.state.batch)
```
mvpatel2000, line 3466: isnt this already called in the outer for loop and therefore is not needed here? it should probably only be called once right

composer/devices/device.py
mvpatel2000: Can you elaborate and/or give an example transformation? I still don't follow

JackZ-db: synced offline

tldr `get_num_samples_in_batch` and `get_num_tokens_in_batch` throw NotImplemented Errors when they are called before a batch is loaded onto a gpu bc they don't recognize `'transformers.tokenization_utils_base.BatchEncoding'>`

composer/trainer/trainer.py
JackZ-db: aren't some non-default, custom `device_transforms` methods device dependent? So doing it on GPU vs CPU would matter? I think it makes sense to get rid of the first `device_transforms` outside of the for loop

mvpatel2000: Agree, lets do that

composer/core/data_spec.py
```diff
@@ -246,6 +248,9 @@ def _default_get_num_samples_in_batch(self, batch: Batch) -> int:
         if isinstance(batch, torch.Tensor):
             return batch.shape[0]
 
+        if isinstance(batch, BatchEncoding):
+            batch = dict(batch)
```
mvpatel2000, line 252: can we do if transformer is installed, check here? You can do somethin glike:

```
try:
  from transformers.tokenization_utils_base import BatchEncoding
  _TRANSFORMERS_INSTALLED = True
except:
  _TRANSFORMERS_INSTALLED = False
```
at top and then gate check on that if statement

JackZ-db: done

mvpatel2000 (approved):
@JackZ-db lets make sure to call this out in release notes -- it's behavior changing that full batch is not on device which may affect workloads

PR #3488, opened 2024-07-23
[Checkpoint V2] Upload API

bigning:
Adding upload API, [design doc](https://docs.google.com/document/d/14ruvosCcVgqDMJGPPmi6HRm6YvJO3BKUiZFN0KhBGR4/edit#bookmark=id.qzhhpmitypba). This API uploads the input files to remote object store, interface design:

```
def upload_file(
    dest_dir: str,
    source_path: Optional[str]=None,
    symlink_granularity: Optional[str]=None, # file, dir, or None
    symlink_name: Optional[str]='latest.symlink',
    async_upload: bool = True,
    state: Optional[State] = None,
    overwrite: bool = False,
):
    """Standalone function for uploading a checkpoint file.
    This function does not actually upload the checkpoint; it initiates the RemoteUploader's uploading of it
    Args:
        source_path (str): The path to the file to upload.
        dest_dir (str): The directory/uri to upload the file to.
        symlink_granularity (Optional[str]): The granularity to use for symlinking. One of 'file', 'dir', or None.
            if None: no symlink uploaded
            if 'file': command remoteuploader to wait until the file (specificied by source_path) is uploaded and then uploads a symlink pointing to the uploaded file
            if 'dir': command remoteuploader  to wait until all files across all ranks are uploaded to dest_dir and then uploads a symlink
                pointing to the remote directory (prefix in object_store terminology).
        symlink_name (Optional[str]): The name to use for the symlink. Defaults to 'latest.symlink'.
        async_upload (bool): If True, the uploads will be done asynchronously via the RemoteUploader and this function will return immediately.
        state (Optional[State]): If async_upload is True, then state must be specified so that the remote_uploader can be
            either extracted from state.callbacks or initialized and added to state.callbacks.
        overwrite (bool): If allow overwrite existing remote checkpoint files
    """ 
```

# unit test
python3   -m composer.cli.launcher -n 2 --master_port 26000 -m pytest -v --durations=20 -m 'not daily and not remote and gpu and (doctest or not doctest)' tests/checkpoint/test_upload.py

### DIFF
--- a/composer/checkpoint/upload.py
+++ b/composer/checkpoint/upload.py
@@ -82,6 +82,7 @@ class CheckpointUploadCallback(Callback):
             else:
                 result = check_remote_files_exist_future.result()
                 if result == RemoteFilesExistingCheckStatus.EXIST:
+                    assert remote_symlink_file is not None
                     self.remote_uploader.upload_file_async(
                         remote_file_name=remote_symlink_file,
                         file_path=pathlib.Path(local_symlink_file),
@@ -102,6 +103,7 @@ class CheckpointUploadCallback(Callback):
             check_remote_files_exist_future, local_symlink_file, remote_symlink_file = self.symlink_upload_tasks[-1]
             result = check_remote_files_exist_future.result()
             if result == RemoteFilesExistingCheckStatus.EXIST:
+                assert remote_symlink_file is not None
                 symlink_upload_future = self.remote_uploader.upload_file_async(
                     remote_file_name=remote_symlink_file,
                     file_path=pathlib.Path(local_symlink_file),
--- a/composer/checkpoint/upload.py
+++ b/composer/checkpoint/upload.py
@@ -37,6 +37,10 @@ class CheckpointUploadCallback(Callback):
         else:
             self.rank_saves_symlinks = False
 
+        # Each Tuple contains information of:
+        # the future of checking if checkpoint files upload finish
+        # local symlink file name
+        # remote symlink file name
         self.symlink_upload_tasks: list[tuple[Future[RemoteFilesExistingCheckStatus], str, Optional[str]]] = []
         self.upload_timeout_in_seconds: int = upload_timeout_in_seconds
 
--- a/composer/checkpoint/upload.py
+++ b/composer/checkpoint/upload.py
@@ -1,0 +1,207 @@
+# Copyright 2024 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Useful functions for uploading checkpoints to remote object store."""
+
+import logging
+import os
+import pathlib
+import tempfile
+from concurrent.futures import Future
+from typing import Optional
+
+import composer.utils.dist as dist
+from composer.core import Callback, State
+from composer.loggers import Logger, MosaicMLLogger
+from composer.utils import (
+    RemoteUploader,
+    create_symlink_file,
+    parse_uri,
+    RemoteFilesExistingCheckStatus,
+)
+
+log = logging.getLogger(__name__)
+
+
+class CheckpointUploadCallback(Callback):
+
+    def __init__(
+        self,
+        remote_uploader: RemoteUploader,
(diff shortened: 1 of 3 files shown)


composer/checkpoint/upload.py
```diff
@@ -0,0 +1,209 @@
...
+        remote_uploader: RemoteUploader,
+        upload_timeout_in_seconds: int = 3600,
+    ):
+        self.remote_uploader = remote_uploader
+        if dist.get_global_rank() == 0:
+            self.rank_saves_symlinks = True
+        else:
+            self.rank_saves_symlinks = False
+
+        self.symlink_upload_tasks: list[tuple[Future[RemoteFilesExistingCheckStatus], str, Optional[str]]] = []
```
eracah, line 40: can you put a comment describing what each field in the tuple is?

composer/checkpoint/upload.py
```diff
@@ -0,0 +1,209 @@
...
+    ):
+        check_remote_files_exist_future = None
+        if self.rank_saves_symlinks and symlink_file_name is not None:
+            check_remote_files_exist_future = self.remote_uploader.check_remote_files_exist_async(
+                remote_checkpoint_file_names=remote_file_names,
+                max_wait_time_in_seconds=self.upload_timeout_in_seconds,
+                wait_before_next_try_in_seconds=self._symlink_upload_wait_before_next_try_in_seconds,
+            )
+            self.symlink_upload_tasks.append(
+                (check_remote_files_exist_future, symlink_file_name, symlink_remote_file_name),
```
eracah, line 60: Could also use `deque` and `appendleft` instead of having to reverse the list.

composer/checkpoint/upload.py
```diff
@@ -0,0 +1,209 @@
...
+        if not self.rank_saves_symlinks:
+            return
+        undone_symlink_upload_tasks = []
+        for (check_remote_files_exist_future, local_symlink_file,
+             remote_symlink_file) in reversed(self.symlink_upload_tasks):
+            if not check_remote_files_exist_future.done():
+                undone_symlink_upload_tasks.insert(
+                    0,
+                    (check_remote_files_exist_future, local_symlink_file, remote_symlink_file),
+                )
```
eracah, line 80: wait once we upload the symlink for the last thing in the list then we don't need to upload the other ones? right?

eracah (commented):
This is really slick! Nice work, @bigning ! I will take a look at the unit test tmrw! One request: can we now remove the symlink upload logic from `CheckpointSaver` now? We can just have `CheckpointSaver` use `CheckpointUploadCallback`?

composer/checkpoint/upload.py
bigning: yes, we could. I personally just prefer to use a plain list with chronological order. otherwise, it might be easy to mess the order when using it

composer/checkpoint/upload.py
bigning: we still upload the checkpoint files. But for symlink file, once we upload one symlink, we don't need the previous ones, right?

karan6181:
@bigning, please do not attach the doc link, as not everyone would have access to it. Please add detail description on what would this PR enables and how did you test it

composer/checkpoint/upload.py
eracah: right so why do we still insert the remaining tasks into the list?

eracah (approved):

composer/checkpoint/upload.py
bigning: oh, we reversely traverse the list and insert all the tasks that not finished. Once we find a finished one, we stop the traverse.

bigning:
> @bigning, please do not attach the doc link, as not everyone would have access to it. Please add detail description on what would this PR enables and how did you test it

@karan6181  , added the interface in the description, and also keep the design doc in the description, so reviewers could have the whole picture of all the APIs. 

> how did you test it

This was already in the description.

PR #3491, opened 2024-07-24
Patch conf.py for readthedocs sphinx injection deprecation.

mvpatel2000:
# What does this PR do?

Patch conf.py for readthedocs sphinx injection deprecation. As build time injection is deprecated, we apply the patch recommended by readthedocs.


mvpatel2000:
@bandish-shah @dakinggg bumping for review in case it slipped through

dakinggg (approved):
🙈

Issue #3493, opened 2024-07-25
FSDP Wrapping Alters Optimizer's Parameter Tracking Behavior


DavidBert:
Hi!
There appears to be an inconsistency in the behavior of the optimizer before and after wrapping with Fully Sharded Data Parallel (FSDP).

When FSDP wraps the optimizer, it seems to modify the set of parameters being tracked by the optimizer. Specifically, after wrapping, all model parameters are tracked by the optimizer.

The optimizer should maintain its original parameter tracking behavior after FSDP wrapping. For example, if the optimizer was initially tracking only parameters from specific submodules, it should continue to do so after FSDP wrapping.

This behavior change could lead to unexpected results, especially in cases where users have intentionally set up their optimizers to track only specific parts of the model.
https://github.com/mosaicml/composer/blob/55f0b7d1880caaf218a56212e86f174bbc463d12/composer/distributed/dist_strategy.py#L719C1-L725C51

mvpatel2000:
Hm... yes I think it might be not properly saving the subset of parameters if the group is not all params. 

Code to save: https://github.com/mosaicml/composer/blob/55f0b7d1880caaf218a56212e86f174bbc463d12/composer/distributed/dist_strategy.py#L265-L295

@sashaDoubov what do you think? iirc you added this bit

@DavidBert do you have a repro we can use here?

DavidBert:
Thanks for the quick answer!
This code should demonstrate the undesired behavior:

```python
import torch
import torch.nn as nn
import torch.nn.functional as F
from torchvision import datasets, transforms
from torch.utils.data import DataLoader
# optimizer
from torch.optim import Adam

from composer import Trainer
from composer.models import ComposerClassifier
from composer.utils import dist
import copy


class Model(nn.Module):
    """Toy convolutional neural network architecture in pytorch for MNIST."""

    def __init__(self, num_classes: int = 10):
        super().__init__()

        self.num_classes = num_classes

        self.conv1 = nn.Conv2d(1, 16, (3, 3), padding=0)
        self.conv2 = nn.Conv2d(16, 32, (3, 3), padding=0)
        self.bn = nn.BatchNorm2d(32)
        self.fc1 = nn.Linear(32 * 16, 32)
        self.fc2 = nn.Linear(32, num_classes)

    def forward(self, x):
        out = self.conv1(x)
        out = F.relu(out)
        out = self.conv2(out)
        out = self.bn(out)
        out = F.relu(out)
[...]

PR #3502, opened 2024-07-30
Fix bug when subset of model parameters is passed into optimizer with FSDP

sashaDoubov:
# What does this PR do?

This PR addresses https://github.com/mosaicml/composer/issues/3493, where we were adding all model parameters into the optimizer state after FSDP wrapping. This change uses the previous flow for param_groups > 1, where we:
1. Get the mapping between params in the optimizer group and their names by using the pointers (found with `id`) of the tensors
2. Store the extra param group info, such as LR, wd, etc. in a separate dict
3. Iterate over the fsdp wrapped parameters, and populates the param_groups based on the names of the fsdp wrapped tensor and the mapping from before.

This change relies on `use_orig_params=True`, so an error is thrown if a subset of parameters is passed into the param_groups.
 
**NOTE**: at the moment, tensor parallelism breaks with multiple param_groups (as the ptrs of TP'd model.parameters() != ptrs of elements in optimizer param_groups), so a special case was added to handle this.

Fixes https://github.com/mosaicml/composer/issues/3493

# Before submitting

### DIFF
--- a/composer/distributed/dist_strategy.py
+++ b/composer/distributed/dist_strategy.py
@@ -186,9 +186,23 @@ def _recreate_fsdp_param_groups_from_unwrapped_opt_info(
 
 def prepare_tp_module(
     model: torch.nn.Module,
+    optimizers: Optional[Union[torch.optim.Optimizer, Sequence[torch.optim.Optimizer]]],
     tp_config: TPConfig,
 ) -> None:
     """Prepare a module (assumed ComposerModel) for use with tensor parallel."""
+    optimizers_tuple = ensure_tuple(optimizers)
+    if len(optimizers_tuple) != 1:
+        raise NotImplementedError(f'Only one optimizer is supported; found {len(optimizers_tuple)} optimizers')
+
+    optim = optimizers_tuple[0]
+    if len(optim.param_groups) > 1:
+        raise RuntimeError('Multiple optimizer groups are not supported with tensor parallelism.',)
+
+    if len(optim.param_groups[0]['params']) != len(list(model.parameters())):
+        raise ValueError(
+            'Passing in a subset of model parameters to the optimizer is not supported with tensor parallelism.',
+        )
+
     from torch.distributed.tensor.parallel import parallelize_module
 
     device_mesh = tp_config.device_mesh
@@ -207,7 +221,6 @@ def prepare_fsdp_module(
     precision: Precision,
     device: Device,
     auto_microbatching: bool,
-    using_tp: bool = False,
     te_rng_seed: int = 1234,
 ) -> None:
     """Prepare a module (assumed ComposerModel) and optimizer for use with :class:`torch.distributed.fsdp.FullyShardedDataParallel`.
@@ -219,7 +232,6 @@ def prepare_fsdp_module(
         precision: (Precision): The precision being used by the Trainer, used to fill in defaults for FSDP `mixed_precision` settings.
         device (Device): The device being used by the Trainer.
         auto_microbatching (bool, optional): Whether or not auto microbatching is enabled.
-        using_tp (bool, optional): Whether the model has been wrapped with Tensor Parallelism, in which case only a single optimizer param group is supported.
         te_rng_seed(int): The seed to use for the Transformer Engine activation checkpointing RNG. Defaults to 1234.
     """
     # Check sync_module_states is True for mixed initialization or HSDP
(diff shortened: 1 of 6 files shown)


composer/distributed/dist_strategy.py
```diff
@@ -261,37 +267,41 @@ def sync_hook(*args):
...
                 optimizer_specific_group_info = {
                     k: v for k, v in optim.param_groups[group_num].items() if k != 'params'
                 }
-                group_num_to_param_group_info[group_num] = optimizer_specific_group_info
+                group_num_to_opt_group_info[group_num] = optimizer_specific_group_info
         else:
+            if len(optim.param_groups) > 1:
+                raise RuntimeError(
+                    'Multiple optimizer groups with FSDP are not supported with tensor parallelism and/or use_orig_params=False.',
+                )
```
sashaDoubov, line 298: would like feedback if we want this error message broken out into separate conditions for TP vs. use_orig_params=False

composer/distributed/dist_strategy.py
```diff
@@ -213,6 +219,7 @@ def prepare_fsdp_module(
         precision: (Precision): The precision being used by the Trainer, used to fill in defaults for FSDP `mixed_precision` settings.
         device (Device): The device being used by the Trainer.
         auto_microbatching (bool, optional): Whether or not auto microbatching is enabled.
+        using_tp (bool, optional): Whether the model has been wrapped with Tensor Parallelism, in which case only a single optimizer param group is supported.
```
sashaDoubov, line 222: new arg, perhaps there's an easier way to determine if using tp, but this was what made sense to me. would also appreciate feedback regarding this.

composer/distributed/dist_strategy.py
```diff
@@ -171,9 +171,14 @@ def _recreate_fsdp_param_groups_from_unwrapped_opt_info(
     for fsdp_name, param in fsdp_wrapped_named_params:
 
         unwrapped_name = clean_tensor_name(fsdp_name)
-        # need to have a 1:1 mapping between a fsdp param name and the non-wrapped vanilla param name
-        retrieved_group_num = non_wrapped_param_names_to_group_num[unwrapped_name]
-        group_num_to_optimizer_info[retrieved_group_num]['params'].append(param)
+
+        # since we are iterating over all model.named_parameters() after fsdp wrapping, we need to check
```
mvpatel2000, line 175:
```suggestion
        # Since we are iterating over all model.named_parameters() after fsdp wrapping, we need to check
```
capitalize start of comment

composer/distributed/dist_strategy.py
```diff
@@ -171,9 +171,14 @@ def _recreate_fsdp_param_groups_from_unwrapped_opt_info(
...
+        # if the parameter was included in the optimizer param_group pre fsdp wrapping, in order to support
+        # passing a subset of model params in the optimizer
+        if unwrapped_name in non_wrapped_param_names_to_group_num:
+            # need to have a 1:1 mapping between a fsdp param name and the non-wrapped vanilla param name
```
mvpatel2000, line 179:
```suggestion
            # Need to have a 1:1 mapping between a fsdp param name and the non-wrapped vanilla param name
```

composer/distributed/dist_strategy.py
mvpatel2000: can we just check the param groups in `prepare_tp_module`

composer/distributed/dist_strategy.py
sashaDoubov: I've now moved this to TP as recommended by @mvpatel2000

tests/trainer/test_tp.py
```diff
@@ -47,3 +49,80 @@ def test_tp_train(world_size: int):
...
+            parallelism_config={
+                'tp': {
+                    'layer_plan': layer_plan,
+                    'tensor_parallel_degree': 2,
+                },
+                'fsdp': {},
+            },
+            max_duration='3ba',
+        )
+    gc.collect()
```
mvpatel2000, line 128: why is this necessary?

tests/trainer/test_tp.py
```diff
@@ -47,3 +49,80 @@ def test_tp_train(world_size: int):
...
```
mvpatel2000, line 93: same as below

mvpatel2000 (commented):
LGTM! 

why are gc collect necessary? will approve after that

tests/trainer/test_tp.py
sashaDoubov: I was following my previous impl here: https://github.com/mosaicml/composer/blob/98a87f3845aa19d0dee53767c448533ac93287b6/tests/trainer/test_fsdp_param_groups.py#L51, but I'll try removing and see if the test pass in that case.

mvpatel2000 (approved):

PR #3510, opened 2024-07-31
Remove OOM-Driven FSDP Deadlocks and Increase Throughput of Automicrobatching

JackZ-db:
Prior to this PR, automicrobatching suffered from both low reliability, in the form of consistent deadlocks when using FSDP, and decreased throughput, compared to if `device_train_microbatch_size` were manually set to the value auto found. This PR addresses both those issues, removing the sources of OOM-driven FSDP deadlocks and implementing a more intelligent sync hook adding/dropping method that allows auto to perform as well as when `dtms` is manually set.

**Reliability:**

Hooks: 

With FSDP, there are 4 different sources of deadlocks, which become consistent especially when model size grows to 30b:

1. Before the forward of an FSDP module, some ranks may OOM and some ranks may not, leading to the OOMing ranks all_reducing when they are trying to see if other ranks are OOMing and the non-OOMing ranks all_gathering when they are unsharding and resharding as they continue into the FSDP modules
   a) Solution: Register `module.register_forward_pre_hook(sync_hook, prepend=True)` on FSDP modules

2. Before the backwards of an FSDP module, some ranks may OOM and some ranks may not, leading to the OOMing ranks all_reducing when they are trying to see if other ranks are OOMing and the non-OOMing ranks all_gathering when they are unsharding and resharding as they continue into the FSDP modules
   a) Solution: Register `module.register_full_backward_pre_hook(sync_hook, prepend=True)` on FSDP modules

3. In the FSDP post-backward-hook, [post_backward_reshard](https://github.com/pytorch/pytorch/blob/d4a79d4a7c746068d25fe5cf9333495561f4ce1f/torch/distributed/fsdp/_runtime_utils.py#L796) prefetches unshard, moving the unshard from the beginning of forward to the end of backward
   a) Solution: Register `module.register_full_backward_hook(sync_hook)` on the non-FSDP original modules

4. Within the FSDP post-backward-hook, [the call to unshard reallocates](https://github.com/pytorch/pytorch/blob/ddde9dd25cdfd7a3f7f8e7ba76ad096f50cf352f/torch/distributed/fsdp/_flat_param.py#L1311C9-L1311C73) memory, causing some ranks to OOM and others to not OOM. Since there is no syncing within the FSDP native hook, this can lead to deadlock when half the ranks OOM during that realloc and half don’t
   a) Solution: Monkeypatch a sync hook right before the `realloc`; proof of adaptive patch working: `test-patch-mpt-125m-chinchilla-regression-LfuGDp` 

Thrashing:

Thrashing occurs when we are close to the GPU's maximum memory, and `alloc_retries` consistently occur, leading to lower throughput and a risk of OOMing. If we detect `alloc_retries` for two consecutive batches, we consider it thrashing, and search downwards for a smaller microbatch size, treating it as if it were an OOM.

Run with thrashing check: `mpt-30b-auto-fix-egrtom`

- Samples/Sec: 7.25
- dtms: 4

Run without thrashing check: `mpt-30b-auto-fix-ZQmoZp `

- Samples/Sec: 6.9
- dtms: 8

**Throughput**

We turn all hooks on when we are searching for a microbatch size for the first time, right after we finish an eval in case of a memory spike, when we detect thrashing, and if we hit even a single OOM. We leave these hooks on until we can successfully run batches with the selected microbatch size for 3 consecutive batches. Then, to increase throughput, we drop all hooks until we hit one of the events listed above. 

This allows us to achieve the same throughput as if we had manually set the `dtms` to what auto found.

**Experiments**

Improvement on throughput depends on whether the bottleneck of training is GPU/CPU bound, which depends on model size/architecture.

MPT-125M, 1 Node: 
- [Auto Pre-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118539/runs/b445f2fb27434474a3126d689fb53484?o=3360802220363900) (dtms = 32): 835 Samples/Sec
- [Auto Post-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118540/runs/699536691ffc41bcbbb3584be3ffdd94?o=3360802220363900) (dtms = 32): 990 Samples/Sec
- [Manual on Dev](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/1017967647092209/runs/84a12f6766164591a8f0d53e8324a4b5/model-metrics?o=3360802220363900) (dtms = 32): 990 Samples/Sec

MPT-1B, 1 Node:
- [Auto Pre-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118539/runs/de27c22cdfb245a7b274ab8df94c7327?o=3360802220363900) (dtms = 64): 152 Samples/Sec
- [Auto Post-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118540/runs/84bc1e19e0cd431dafd59083ef0816af/model-metrics?o=3360802220363900) (dtms = 64): 157 Samples/Sec
- [Manual on Dev](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/1017967647092209/runs/40eb203d991744c89725b00f82445ac5/model-metrics?o=3360802220363900) (dtms = 64):157 Samples/Sec

MPT-7B, 1 Node:
- [Auto Pre-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118539/runs/60df79a832a64d2eb4de38e59bea30c2?o=3360802220363900) (dtms = 32): 74 Samples/Sec
- [Auto Post-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118540/runs/0c3fc313f1364771bd7fa7922787384b?o=3360802220363900) (dtms = 32): 75.6 Samples/Sec
- [Manual on Dev](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/1017967647092209/runs/34064392219d44ff967f16cb009689ab/model-metrics?o=3360802220363900) (dtms = 32): 76 Samples/Sec

MPT-13B, 2 Node:
- [Auto Pre-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118539/runs/dc6b755cee474234bb17c3356a760951?o=3360802220363900) (dtms = 32): 41.25 Samples/Sec
- [Auto Post-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118540/runs/d790f52c3d9d43088b7f209c00ae2e7d?o=3360802220363900) (dtms = 32): 41.85 Samples/Sec
- [Manual on Dev](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/1017967647092209/runs/176a705c2442409f9821179287fa1a84/model-metrics?o=3360802220363900) (dtms = 32): 41.62 Samples/Sec

MPT-30B, 2 Node:
- [Auto Pre-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118539/runs/427e1e4642c849c58360a53ff35d85d3?o=3360802220363900) (dtms = 16): 17.9 Samples/Sec
- [Auto Post-PR](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2995546623118540/runs/58d7daafc5544dbd85d8a5215c9331b0?o=3360802220363900)(dtms = 16): 18.26 Samples/Sec
- [Manual on Dev](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/1017967647092209/runs/25d9368372274ee7b4bd6e22b7d9c535/model-metrics?o=3360802220363900) (dtms = 16): 18.25 Samples/Sec

Credits: One unit test, `test_automicrobatching_fsdp`, borrows classes introduced in an earlier [WIP PR](https://github.com/mosaicml/composer/blob/efd7bf02b4c52b0efee439b9ddd503b417d998eb/tests/trainer/test_fsdp.py) by @bigning.

Successful regression test: mcli logs -f llm-foundry-regression-tests-runner-PW7pJe

- Same loss, higher throughput, lower memory usage (due to removal of hooks)


composer/trainer/trainer.py
```diff
@@ -1251,6 +1334,7 @@ def __init__(
         if parallelism_config is not None:
             # Patch PyTorch to fix distributed bugs
             patch_pytorch()
+            patch_unshard_for_automicrobatching(False)
```
bigning, line 1338: Can we only patch if auto_microbatching is True?

composer/trainer/trainer.py
```diff
@@ -2801,13 +2901,56 @@ def _train_batch(self, use_grad_scaling: bool) -> dict[str, torch.Tensor]:
...
+                    handle.remove()
+                self.automicrobatch_fsdp_hook_handles.clear()
+            if torch.cuda.is_available():
+                memory_stats = torch.cuda.memory_stats()
+                self.cumulative_alloc_retries = memory_stats['num_alloc_retries']
             self.logger.log_metrics({'trainer/device_train_microbatch_size': self.state.device_train_microbatch_size})
             self.first_batch_complete = True
+            print(len(
+                self.automicrobatch_fsdp_hook_handles
+            ))
```
bigning: remove this

composer/trainer/trainer.py
```diff
@@ -400,6 +411,72 @@ def _adjust_device_eval_microbatch_size(evaluator: Evaluator):
...
+        dist.all_reduce(found_cuda_oom_tensor, reduce_operation='MAX')
+        found_cuda_oom = found_cuda_oom_tensor.item()
+        # Signal current rank is still in batch
+        all_ranks_finished_tensor = state.device.tensor_to_device(torch.tensor([0], dtype=torch.uint8))
+        dist.all_reduce(all_ranks_finished_tensor, reduce_operation='MIN')
+
+        if found_cuda_oom == 1:
+            raise RuntimeError('CUDA out of memory encountered on a different rank')
+
+    return sync_hook
```
bigning, line 462: why do we need to wrap the sync_hook func with `_create_sync_hook` ?

composer/trainer/trainer.py
```diff
@@ -2801,11 +2902,51 @@ def _train_batch(self, use_grad_scaling: bool) -> dict[str, torch.Tensor]:
...
+                    RuntimeWarning(
+                        'Automicrobatching changed the microbatch size from '
+                        f'{original_microbatch_size} -> {self.state.device_train_microbatch_size}.',
+                    ),
+                )
+            self.num_consecutive_non_OOM_batches += 1
+            if self.state.fsdp_enabled and len(
+                self.automicrobatch_fsdp_hook_handles,
+            ) > 0 and self.num_consecutive_non_OOM_batches >= 3:
+                print('remove hooks from batch completion')
```
bigning, line 2942: print -> log.debug ?

tests/trainer/test_fsdp.py
```diff
@@ -206,6 +206,78 @@ def test_fsdp_prefetch_limit(forward_prefetch_limit: int, backward_prefetch_limi
...
+            'forward_prefetch_limit': 1,
+            'backward_prefetch_limit': 1,
+        },
+        max_duration='1ba',
+        device='gpu',
+        device_train_microbatch_size='auto',
+        dist_timeout=20,
+    )
+    trainer.fit()
+
```
bigning, line 279: can we test the code of  removing hook then readding hook ?

bigning (commented):
Great write-up in the PR description!

composer/trainer/_patch_pytorch.py
```diff
@@ -1043,3 +1053,71 @@ def device_mesh__getitem__(
...
+            unsharded_flat_param = (
+                self._get_padded_unsharded_flat_param()
+                if self.uses_sharded_strategy
+                else self.flat_param
+            )
+            self._use_unsharded_flat_param(unsharded_flat_param)
+            return
+        unsharded_flat_param = self._alloc_padded_unsharded_flat_param()
+        padded_unsharded_flat_param = self._all_gather_flat_param(unsharded_flat_param)
+        self._use_unsharded_flat_param(padded_unsharded_flat_param)
```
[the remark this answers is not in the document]
JackZ-db: got it, I'll import unshard in a conditional in patch_unshard_for_automicrobatching

composer/trainer/trainer.py
JackZ-db: good catch

tests/trainer/test_fsdp.py
JackZ-db: yep, also citing you in PR description bc this just reminded me

composer/trainer/trainer.py
JackZ-db: i think to register a hook you need to pass in an actual method, and the only other way I can think of is redefining sync_hook explicitly in _train_batch, which I think is less readable? What are you thinking of?

composer/trainer/trainer.py
```diff
@@ -2801,11 +2902,51 @@ def _train_batch(self, use_grad_scaling: bool) -> dict[str, torch.Tensor]:
...
+                    # Readd sync hooks if they were previously turned off
+                    if self.state.fsdp_enabled and len(self.automicrobatch_fsdp_hook_handles) == 0:
+                        self.automicrobatch_fsdp_hook_handles = _readd_fsdp_sync_hooks(self.fsdp_modules, sync_hook)
+                    _adjust_device_train_microbatch_size(self.state)
+                    self.num_consecutive_thrashes = 0
+                    continue
+
             # Log microbatch and return loss if we've completed without OOMing.
             assert self.state.device_train_microbatch_size is not None
+            if original_microbatch_size != self.state.device_train_microbatch_size:
```
mvpatel2000, line 2931: why is this a warning? should be log at most right?

composer/trainer/trainer.py
```diff
@@ -2801,11 +2902,51 @@ def _train_batch(self, use_grad_scaling: bool) -> dict[str, torch.Tensor]:
                     dist.all_reduce(all_ranks_finished_tensor, reduce_operation='MIN')
                     all_ranks_finished = all_ranks_finished_tensor.item() == 1
                 if found_cuda_oom == 1:
+                    # Readd sync hooks if they were previously turned off
+                    if self.state.fsdp_enabled and len(self.automicrobatch_fsdp_hook_handles) == 0:
```
mvpatel2000, line 2906: could we clean to more helper fns?

composer/trainer/trainer.py
```diff
@@ -1251,6 +1335,7 @@ def __init__(
         if parallelism_config is not None:
             # Patch PyTorch to fix distributed bugs
             patch_pytorch()
+            patch_unshard_for_automicrobatching(False)
```
mvpatel2000, line 1338: why does this patch need to be separate? can we not just include in pytorch patch? concerned this adds more maintenance overhead.

composer/trainer/trainer.py
```diff
@@ -400,6 +411,73 @@ def _adjust_device_eval_microbatch_size(evaluator: Evaluator):
...
+
+    # Propagate across all ranks if any rank had alloc retries this batch
+    alloc_retry_tensor = state.device.tensor_to_device(torch.tensor([alloc_retry_this_batch], dtype=torch.uint8),)
+    dist.all_reduce(alloc_retry_tensor, reduce_operation='MAX')
+    alloc_retry_this_batch = alloc_retry_tensor.item() == 1
+    if alloc_retry_this_batch:
+        num_consecutive_thrashes += 1
+    else:
+        num_consecutive_thrashes = 0
+    print(num_consecutive_thrashes)
```
mvpatel2000, line 438: delete

composer/trainer/trainer.py
mvpatel2000: +1 add comment for why it's needed

mvpatel2000 (commented):
first pass. Need to think through this carefully after comments are addressed, but I think its right. 

This is a huge PR 🎉

composer/trainer/trainer.py
JackZ-db: changing to log

composer/trainer/trainer.py
JackZ-db: added a comment

composer/trainer/trainer.py
JackZ-db: resolved offline

composer/trainer/trainer.py
JackZ-db: this patch is dynamic, since it flips between the original FSDP implementation of unshard and our patched unshard with the sync live during the training loop - it wouldn't make sense to call patch_pytorch() and not change 90% of the methods. in terms of maintenance overhead, regardless of whether it's in `patch_pytorch` or `patch_unshard_for_automicrobatching`, theres still going to be exactly one method that needs to be maintained (`unshard_with_sync`)

composer/trainer/trainer.py
JackZ-db: resolved

composer/trainer/trainer.py
JackZ-db: resolved

composer/trainer/trainer.py
```diff
@@ -400,6 +411,75 @@ def _adjust_device_eval_microbatch_size(evaluator: Evaluator):
...
+        # Check if any other rank hit an OOM
+        found_cuda_oom_tensor = state.device.tensor_to_device(torch.tensor([0], dtype=torch.uint8))
+        dist.all_reduce(found_cuda_oom_tensor, reduce_operation='MAX')
+        found_cuda_oom = found_cuda_oom_tensor.item()
+        # Signal current rank is still in batch
+        all_ranks_finished_tensor = state.device.tensor_to_device(torch.tensor([0], dtype=torch.uint8))
+        dist.all_reduce(all_ranks_finished_tensor, reduce_operation='MIN')
+
+        if found_cuda_oom == 1:
+            raise RuntimeError('CUDA out of memory encountered on a different rank')
```
bigning, line 462: nit: since this string is used later to detect if it's OOM exception, might be better to define a constant for this string.

composer/distributed/dist_strategy.py
```diff
@@ -588,6 +586,22 @@ def _auto_wrap_policy_new(module: torch.nn.Module, recurse: bool, nonwrapped_num
...
+            #   forward_pre_hook: before forwards of FSDP modules
+            #   full_backward_pre_hook: before backwards of FSDP modules
+            #   full_backward_hook: before a prefetched unshard called by FSDP's `post_backward_reshard`
+            if auto_microbatching:
+                for _, module in fsdp_obj.named_modules():
+                    if isinstance(module, FullyShardedDataParallel):
+                        hook_handles.append(module.register_forward_pre_hook(sync_hook, prepend=True))
+                        hook_handles.append(module.register_full_backward_pre_hook(sync_hook, prepend=True))
+                    else:
+                        hook_handles.append(module.register_full_backward_hook(sync_hook))
```
bigning, line 602: why do we need to return a dict `fsdp_obj_named_modules` ? we can just return fsdp_obj? So that we can put these into a helper function that can be re-used at [1] ?

composer/trainer/trainer.py
```diff
@@ -400,6 +411,75 @@ def _adjust_device_eval_microbatch_size(evaluator: Evaluator):
...
+    Called when preparing to search for or searching for new microbatch size during automicrobatching.
+    """
+    automicrobatch_fsdp_hook_handles = []
+    patch_unshard_for_automicrobatching(auto_microbatch_size_found=False)
+    for module in fsdp_modules.values():
+        if isinstance(module, FullyShardedDataParallel):
+            automicrobatch_fsdp_hook_handles.append(module.register_forward_pre_hook(sync_hook, prepend=True))
+            automicrobatch_fsdp_hook_handles.append(module.register_full_backward_pre_hook(sync_hook, prepend=True))
+        else:
+            automicrobatch_fsdp_hook_handles.append(module.register_full_backward_hook(sync_hook))
```
bigning, line 479: [1]

composer/distributed/dist_strategy.py
JackZ-db: since we're iterating through potentially multiple objects in this method, we want to aggregate all of the modules we added hooks to across all objects. Otherwise, we would just be overwriting all of that with the last fsdp_obj.

composer/trainer/trainer.py
JackZ-db: got it

composer/trainer/trainer.py
JackZ-db: resol,ved

bigning (approved):
Could you please also add the before/after training loss comparison. And make sure the loss is on-par before merge.

composer/trainer/trainer.py
```diff
@@ -2803,11 +2909,48 @@ def _train_batch(self, use_grad_scaling: bool) -> dict[str, torch.Tensor]:
...
                     # Skip return and rerun after handling oom
                     continue
+                if not oom_found_this_batch and torch.cuda.is_available():
+                    # Sync across all ranks to check if any rank had additional alloc retries this batch
+                    self.num_consecutive_thrashes = _update_num_consecutive_thrashes(
+                        self.state,
+                        self.num_consecutive_thrashes,
+                        self.cumulative_alloc_retries,
+                    )
+                if self.num_consecutive_thrashes >= 2:
```
dakinggg, line 2928: @mvpatel2000 @JackZ-db A small number of alloc retries consistently increasing, or a large number of alloc retries in the first couple batches might be ok right and we wouldn't want to decrease dtms?

j316chuck:
Thanks for the clean PR description and nice debugging work! Just for my understanding: 

1) On the thrashing section, why is auto throughput samples/sec lower after fix? 

2) Throughput: iiuc, we can:  turn on hooks in train forward -> find micro batch size A ->  turn off hooks -> oom in eval forward ->  turn on hooks again in eval forward -> find microbatch size B -> turn off hooks in eval forward? If so, then the first train batch will != second train batch size

composer/trainer/trainer.py
JackZ-db: correct me if I'm wrong, but allocations should stabilize after 2 batches? I agree that it makes sense to make this threshold more relaxed though - I'm thinking we can increase the number of consecutive thrashes needed or add an additional check saying alloc retries can't go up by `x` amt? @mvpatel2000

JackZ-db:
> Thanks for the clean PR description and nice debugging work! Two qq:
> 
> 1. On the thrashing section, why is auto throughput samples/sec lower after fix?
> 2. Throughput: iiuc, we can:  turn on hooks in train forward -> find micro batch size A ->  turn off hooks -> oom in eval forward ->  turn on hooks again in eval forward -> find microbatch size B -> turn off hooks in eval forward? If so, then the first train batch will != second train batch size. Is this expected?
> 
> One high level comment is it might be good to split up the PR into 4 different PR's each containing the four different sources of deadlock each with a repro example + throughput improvement. Would be good for posterity when we are reviewing/thinking of understanding each of the components of deadlocks.

1 is a good catch, i just switched the two headings - thrashing decreases dtms but still increases throughput.

2. The hook readding isn't meant for OOMs in evaluation - it's meant to account for a memory spike that we see directly after the first evaluation (future evaluations don't introduce new memory allocations), which then may cause the first train batch after evaluation to OOM.

composer/trainer/trainer.py
mvpatel2000: I've only seen consistently increasing + safe behavior with MoEs with dropless routing. Given auto does not work for MoEs as is, I think it's fine. 

I view this as a more conservative approach (decreasing DTMS too much => 10% slowdown, thrashing => 2x slowdown), and I think that's the right place to start -- if we discover later it's too cautious, we can always tune it.

dakinggg: I believe we have also seen this for huge dense models.

mvpatel2000: Have an example you can dig up?

mvpatel2000 (approved):
Looks good as a v1. 

I agree with @j316chuck in general we should cut this up into smaller PRs, but I think it's fine at this point to merge given we've reviewed and signed off.

@dakinggg feel free to block if you're concerned about thrashing, but I'm pro merging with the cautious approach and we can revisit in follow-on PR

composer/trainer/trainer.py
dakinggg: @snarayan21 can you share on slack?

dakinggg:
@mvpatel2000 yeah i think its ok to merge as is and revisit as needed

mvpatel2000:
> Just wanted to check - has this been tested for scenarios where you are starting from "older" composer checkpoints?

@jacobfulano it should be independent of checkpoint loading

JackZ-db:
> Just wanted to check - has this been tested for scenarios where you are starting from "older" composer checkpoints?

afaik, checkpointing saves automicrobatching info, but once its loaded, its never used, so idt this should be a problem? let me know though if you're referring to something else and I'll take a look!

PR #3513, opened 2024-08-01
Monitor mlflow exit status

mvpatel2000:
Add a monitor process for mlflow logger so that it sets run status as failed if main thread exits unexpectedly


composer/loggers/mlflow_logger.py
```diff
@@ -33,6 +36,52 @@
...
 
+class MlflowMonitorProcess(multiprocessing.Process):
+
+    def __init__(self, main_pid, mlflow_run_id, mlflow_tracking_uri):
+        super().__init__()
+        self.main_pid = main_pid
+        self.mlflow_run_id = mlflow_run_id
+        self.mlflow_tracking_uri = mlflow_tracking_uri
+        self.exit_event = multiprocessing.Event()
+        self.crash_event = multiprocessing.Event()
```
mvpatel2000, line 47: @chenmoneygithub why do we need two events?

dakinggg (commented):
I assume this is pretty tough to unit test, so can you please provide manual testing before I review?

composer/loggers/mlflow_logger.py
```diff
@@ -528,21 +608,49 @@ def log_images(
...
     def post_close(self):
         if self._enabled:
+            if hasattr(self, 'monitor_process'):
+                # Check if there is an uncaught exception, which means `post_close()` is triggered
+                # due to program crash.
+                if isinstance(self._global_exception_occurred, torch.Tensor):
+                    finish_with_exception = (self._global_exception_occurred == 1).item()
+                else:
+                    finish_with_exception = (self._global_exception_occurred == 1)
+                if finish_with_exception:
```
mvpatel2000, line 618: @chenmoneygithub I think we can remove the torch tensor things and just use a bool for this var

PR #3521, opened 2024-08-05
Check if automicrobatching is on before using FSDP modules in Eval

JackZ-db:
Runs resuming from pre-sync-dropping Composer broke due to FSDP modules not existing in the Eval Loop. This checks for automicrobatching before calling `self.fsdp_modules`.

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -3753,10 +3753,11 @@ class Trainer:
             self.state.dataloader_len = original_num_batches
 
         # If training occurs after evaluation, readd hooks in case of memory spike
-        sync_hook = _create_sync_hook(self.state)
-        if self.state.fsdp_enabled and len(self.automicrobatch_fsdp_hook_handles) == 0:
-            self.automicrobatch_fsdp_hook_handles = _readd_fsdp_sync_hooks(self.fsdp_modules, sync_hook)
-        self.num_consecutive_non_OOM_batches = 0
+        if self.state.auto_microbatching:
+            sync_hook = _create_sync_hook(self.state)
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -3753,10 +3753,11 @@ def _eval_loop(
...
         # If training occurs after evaluation, readd hooks in case of memory spike
-        sync_hook = _create_sync_hook(self.state)
-        if self.state.fsdp_enabled and len(self.automicrobatch_fsdp_hook_handles) == 0:
-            self.automicrobatch_fsdp_hook_handles = _readd_fsdp_sync_hooks(self.fsdp_modules, sync_hook)
-        self.num_consecutive_non_OOM_batches = 0
+        if self.state.auto_microbatching:
+            sync_hook = _create_sync_hook(self.state)
+            if self.state.fsdp_enabled and len(self.automicrobatch_fsdp_hook_handles) == 0:
+                self.automicrobatch_fsdp_hook_handles = _readd_fsdp_sync_hooks(self.fsdp_modules, sync_hook)
+            self.num_consecutive_non_OOM_batches = 0
```
mvpatel2000, line 3760: I don't think this solves the issue -- if you resume with automicrobatching on, then it will still fail. Instead, the top level if statement should also check if attr `fsdp_modules` exists. 

Can you verify also this works with Jacob's run?

mvpatel2000: I also don't understand -- shouldn't `self.fsdp_modules` be set during FSDP wrap on init?

JackZ-db: yea it doesnt the issue is in checkpointing, where prepare_fsdp_module is called and doesnt return anything

PR #3522, opened 2024-08-05
Move hooks and fsdp modules onto state rather than trainer

JackZ-db:
Eval'ing after resuming from checkpointing broke since calls to `prepare_fsdp_module` in `state.py` didn't modify the handles and FSDP modules in modules. We fix this by moving `automicrobatch_fsdp_hook_handles` and `fsdp_modules` into `State` instance variables instead of `Trainer` instance variables.

Fixed regression test that previously broken: `mpt-7b-composer-eval-regression-KkY29t`

### DIFF
--- a/composer/checkpoint/load.py
+++ b/composer/checkpoint/load.py
@@ -139,7 +139,8 @@ def load_checkpoint(
         assert model is not None
         assert model_child_path is not None
         model_load_path = os.path.join(load_path, model_child_path)
-        state.automicrobatch_fsdp_hook_handles, state.fsdp_modules = load_model_checkpoint(model, load_path=model_load_path, load_options=load_options)
+        if state:
+            state.automicrobatch_fsdp_hook_handles, state.fsdp_modules = load_model_checkpoint(model, load_path=model_load_path, load_options=load_options)
 
     if load_options.load_optimizer:
         assert optim_child_path is not None
--- a/composer/checkpoint/load.py
+++ b/composer/checkpoint/load.py
@@ -11,7 +11,7 @@ import tempfile
 import textwrap
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, Optional, Sequence, Union, Tuple
+from typing import Any, Dict, Optional, Sequence, Tuple, Union
 
 import torch
 import torch.distributed.checkpoint as DCP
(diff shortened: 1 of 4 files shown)


tests/checkpoint/test_load.py
```diff
@@ -333,3 +338,102 @@ def test_load_checkpoint(
             deep_compare(original_model_state_dict, new_state_dict)
             deep_compare(original_optim_state_dict, new_optim_state_dict)
             deep_compare(original_resumption_state, new_resumption_state, ignore_keys=['rng', 'run_name'])
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize(
+    'world_size,sharded_model,sharded_checkpoint,shard_as_needed_during_load',
+    [
+        #Loading an unsharded checkpoint into an unsharded model on a single GPU (not sharding after)
```
mvpatel2000, line 347:
```suggestion
        # Loading an unsharded checkpoint into an unsharded model on a single GPU (not sharding after)
```

tests/checkpoint/test_load.py
```diff
@@ -333,3 +338,102 @@ def test_load_checkpoint(
...
+
+        # Loading a sharded checkpoint into a sharded model in distributed setting
+        pytest.param(2, True, True, False, marks=pytest.mark.world_size(2)),
+
+        # SHOULD FAIL: Loading an unsharded checkpoint into a sharded model
+        pytest.param(2, True, False, False, marks=pytest.mark.world_size(2)),
+
+        # SHOULD FAIL: Attempting to load a sharded checkpoint into an unsharded model without sharding
+        pytest.param(2, False, True, False, marks=pytest.mark.world_size(2)),
+
```
mvpatel2000, line 358:
```suggestion
```
no point in testing scenarios where loading fails

composer/checkpoint/load.py
```diff
@@ -139,7 +139,10 @@ def load_checkpoint(
         assert model is not None
         assert model_child_path is not None
         model_load_path = os.path.join(load_path, model_child_path)
-        load_model_checkpoint(model, load_path=model_load_path, load_options=load_options)
+        if state:
```
mvpatel2000, line 142:
```suggestion
        if state is not None:
```
I don't think this is correct -- you should set return values if state is passed in, but you should still call `load_model_checkpoint` even if state is not specified

JackZ-db: let me know if my fix changes things

mvpatel2000 (approved):
LGTM, please ensure manual testing works

JackZ-db:
> LGTM, please ensure manual testing works

Requested rerun from research

bigning (approved):
LGTM!

PR #3530, opened 2024-08-08
Fix FSDP Config Validation

mvpatel2000:
# What does this PR do?

Previously, FSDP config validation only checked the state dict type if FSDP was enabled. As FSDP can be enabled later on after this validation check, e.g. via load_monolith_rank0_only=True or by manual wrapping in a callback, we should instead check the true value of state dict type in the config.


snarayan21 (approved):
code lgtm, do you have a manual run (or is there a unit test) that uses `load_monolith_rank0_only=True`?

approving but would like that before merge

dakinggg (approved):

PR #3531, opened 2024-08-08
hasattr check for Wandb 0.17.6

mvpatel2000:
# What does this PR do?

Wandb 0.17.6 changed disabled mode in this PR: https://github.com/wandb/wandb/pull/8037, which breaks wandb logger when running with disabled mode as certain attributes (entity, project) are not set. This wraps these attributes in a check to see if they exist.

This should not break any internal workflows as we have churned from WandB completely in favor of MLFlow, but it may affect open source users so we should prioritize including in patch release.

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -77,13 +77,10 @@ jobs:
       pytest-command: ${{ matrix.pytest_command }}
       pytest-markers: ${{ matrix.markers }}
       composer_package_name: ${{ matrix.composer_package_name }}
-      pytest-wandb-entity: "mosaicml-public-integration-tests"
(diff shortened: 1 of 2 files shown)


bigning (approved):

dakinggg (commented):
would it be safe to just make the setting of `self._enabled` conditional on what mode wandb is in?

mvpatel2000:
> would it be safer to just make the setting of `self._enabled` conditional on what mode wandb is in?

While possible, I would prefer to rely on wandb's handling of disabled as it's less code surface for us to own and maintain. This is especially true since we now don't use wandb so it's harder to test things like mocking disabled. 

If more things break, we can do that, but for now, I'd prefer to leave as is and consider this code as frozen as possible

dakinggg (approved):

PR #3538, opened 2024-08-08
restore google cloud object store test

bigning:
We've got the new google cloud project, service account and keys with help from security team, now we can restore the google cloud object store test. 
daily test: https://github.com/mosaicml/composer/actions/runs/10310134721
image


snarayan21 (approved):
lgtm, thanks!

j316chuck (approved):

PR #3540, opened 2024-08-09
Bump to Ubuntu 22.04

mvpatel2000:
# What does this PR do?

Bump to Ubuntu 22.04. 20.04 will be EOLed in 8 months. It's [Jammy Jellyfish](https://releases.ubuntu.com/jammy/) time


dakinggg (commented):
will wait for builds to pass before reviewing. Also, can we please verify we can still build foundry images on top of it too?

PR #3549, opened 2024-08-12
Add Torch 2.4 Tests

mvpatel2000:
# What does this PR do?

Add Torch 2.4 Tests. Includes
- Updating error messages to latest versions in PyTorch 2.4
- Zero grad must be now called at end of fit w new ckpting API for stateless optimizers, https://github.com/pytorch/pytorch/issues/133415

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - name: gpu-3.11-2.342
+        - name: gpu-3.11-2.4-2
           container: mosaicml/pytorch:2.4.0_cu121-python3.11-ubuntu20.04
           markers: not daily and not remote and gpu and (doctest or not doctest)
           pytest_command: coverage run -m pytest
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
         include:
-        - name: gpu-3.11-2.3-1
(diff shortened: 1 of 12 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -39,8 +39,8 @@ jobs:
     strategy:
       matrix:
         include:
-        - name: gpu-3.11-2.3-2
-          container: mosaicml/pytorch:2.3.1_cu121-python3.11-ubuntu20.04
+        - name: gpu-3.11-2.342
```
dakinggg, line 42: typo here?

docs/source/doctest_fixtures.py
```diff
@@ -117,7 +124,7 @@
 
 model = SimpleModel(num_channels, num_classes)
 
-optimizer = torch.optim.SGD(model.parameters(), lr=0.001)
+optimizer = DecoupledSGDW(model.parameters(), lr=0.001)
```
dakinggg, line 127: why this change?

dakinggg (approved):

docs/source/doctest_fixtures.py
mvpatel2000: Due to https://github.com/pytorch/pytorch/issues/133415. Stateless optimizer causes problems

dakinggg: does that imply composer is broken with torch.optim.SGD?

mvpatel2000: Currently, yes 😬. We have workaround though (decoupledSGDW) and hoping to get into torch2.4.1 for a fix

PR #3344, opened 2024-05-30
Migrate Type Hints for PEP 585

mvpatel2000:
# What does this PR do?

Update type hints following [PEP 585](https://peps.python.org/pep-0585/).

PR #2697, opened 2023-11-08
Update types to follow PEP 585

b-chu:
# What does this PR do?
commit-id:ecb54855

---

**Stack**:
- #2698
- #2697 ⬅
- #2696
- #2695

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

# What issue(s) does this change relate to?

# Before submitting
- [ ] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [ ] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [ ] Did you update any related docs and document your change?
- [ ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#running-tests))
- [ ] Did you run the tests locally to make sure they pass?
- [ ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites))

PR #3551, opened 2024-08-14
Implement ruff rules enforcing pep 585

snarayan21:
# What does this PR do?

Builds on #2697 and #3344 to make sure that we actually enforce native python type hints via ruff.

# Before submitting

### DIFF
--- a/composer/__init__.py
+++ b/composer/__init__.py
@@ -3,6 +3,7 @@
 
 """Some key classes are available directly in the ``composer`` namespace."""
 
+from composer._version import __version__
 from composer.core import Algorithm, Callback, DataSpec, Engine, Evaluator, Event, State, Time, Timestamp, TimeUnit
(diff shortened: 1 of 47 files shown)


mvpatel2000 (approved):
yes chef cook chef

b-chu (approved):
Thanks!

snarayan21:
@mvpatel2000 to get tests to pass, did 2 things:
* filtered a bunch of FutureWarnings -- are these things we should be addressing? a few were from Deepspeed (which we will be deprecating anyways) but most were from FSDP deprecating some state dict methods. See failed gpu tests here:
* Made sure that `unshard_with_sync` was defined for torch versions >= 2.3.1 (incl. 2.4.0) so that the auto microbatching patch still works

lmk -- could be better to put these in a separate PR as well. happy to do that

PR #3563, opened 2024-08-20
Change iteration timestamp for old checkpoints

b-chu:
# What does this PR do?

When an old checkpoint is loaded, any timestamp attributes associated with iteration are 0. This means those attributes are inaccurate and can't be correctly used. This fixes old checkpoints the first time they're loaded to set those attributes to the correct value.

# Before submitting

### DIFF
--- a/composer/trainer/trainer.py
+++ b/composer/trainer/trainer.py
@@ -1943,6 +1943,18 @@ class Trainer:
                     auto_microbatching,
                 )
 
+        # Set the iteration timestamp to the overall timestamp if loading from a checkpoint that was created before
+        # iteration was introduced in Composer v0.19.1. This is necessary to ensure that the iteration timestamp is
+        # accurate for callbacks and backwards compatibility for checkpoints.
+        if (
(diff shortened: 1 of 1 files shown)


composer/trainer/trainer.py
```diff
@@ -1943,6 +1943,16 @@ def __init__(
                     auto_microbatching,
                 )
 
+        # Set the iteration timestamp to the overall timestamp if loading from a checkpoint that was created before
+        # iteration was introduced in Composer v0.19.1. This is necessary to ensure that the iteration timestamp is
+        # accurate for callbacks and backwards compatibility for checkpoints.
+        if (
+            self.state.timestamp.iteration == 0 and self.state.timestamp.token_in_iteration == 0 and
+            self.state.timestamp.epoch_in_iteration == 0
```
mvpatel2000, line 1951: why does epcoh_in_iteration need to be 0?

b-chu: If the checkpoint is old, then both tokens and epoch should be 0

mvpatel2000 (approved):

PR #3564, opened 2024-08-20
[torch2.4] Fix sharded checkpointing backward compatibility issue

bigning:
torch 2.4 breaks the sharded checkpointing backward compatibility. It changed how the save_planner and load_planner flatten the state dict keys. Torch issue: https://github.com/pytorch/pytorch/issues/133923 . So the new load planner can't load the checkpointing saved with old save_planner. 

This PR monkey patches the load_planner if it failed to load an old checkpoint, then it removes the patch after loading the checkpointing.


bigning:
closing this since it's a forked repo and can't start a daily test.

PR #3566, opened 2024-08-21
Fix typo in `composer_collect_env`

dakinggg:
Previous PR introduced a typo into `composer_collect_env` that caused it to crash. This PR fixes that and adds a simple test that makes sure it runs without error.


mvpatel2000 (approved):
Good catch!

PR #3567, opened 2024-08-21
Microbatch Device Movement

mvpatel2000:
# What does this PR do?

Instead of moving the entire batch to device at once, we now move each microbatch to device. This saves memory for large inputs, e.g. multimodal data, when training with many microbatches. 

This change may affect certain callbacks which run operations on the batch which require it to be moved to an accelerator ahead of time, such as the two changed in this PR. There shouldn't be too many of these callbacks, so we anticipate this change will be relatively safe.

Manual test showing same loss curves:
image


composer/core/data_spec.py
```diff
@@ -246,6 +252,9 @@ def _default_get_num_samples_in_batch(self, batch: Batch) -> int:
         if isinstance(batch, torch.Tensor):
             return batch.shape[0]
 
+        if _TRANSFORMERS_INSTALLED and isinstance(batch, BatchEncoding):
```
dakinggg, line 255: sorry i still don't get the need for this. `BatchEncoding` supports dict like access, and instead of checking `isinstance(batch, dict)` below, I think we just want to update to be `isinstance(batch, Mapping)`

dakinggg (approved):

PR #3568, opened 2024-08-21
Add default value to get_device()

coryMosaicML:
Adds a default `None` value to `get_device()` such that `get_device()` works as the docstring implies.


composer/utils/device.py
```diff
@@ -15,7 +15,7 @@
 _is_xla_installed = None
 
 
-def get_device(device: Optional[Union[str, 'Device']]) -> 'Device':
```
dakinggg, line 18: can you add `Default: None` to the docstring for the `device` arg?

dakinggg (approved):

PR #3570, opened 2024-08-23
Bump version to v0.24.0 + deprecations

snarayan21:
# What does this PR do?

Bumps to 0.24.0 and deprecates some code paths, notably passing `process_group` and `device_mesh` directly in the fsdp config object.

Dailies succeeded: https://github.com/mosaicml/composer/actions/runs/10528606007

Continuation of #3527 

# What issue(s) does this change relate to?

# Before submitting
- [ ] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md)?
- [ ] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.
- [ ] Was this change discussed/approved in a GitHub issue first? It is much more likely to be merged if so.
- [ ] Did you update any related docs and document your change?
- [ ] Did you update any related tests and add any new tests related to your change? (see [testing](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md#running-tests))
- [ ] Did you run the tests locally to make sure they pass?
- [ ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md#prerequisites))

PR #3580, opened 2024-08-27
Error if device mesh specified in fsdp config

snarayan21:
# What does this PR do?

Ensures that we throw value error if device mesh is directly specified in the fsdp config. We want users to more to the more descriptive, less confusing `data_parallel_shard_degree` and `data_parallel_replicate_degree` instead. Adding this error prevents users from accidentally configuring their parallelism differently than the device mesh would indicate -- we won't allow for directly specifying device mesh at all.

Basically changes the warning / warning test removed in #3570 to error / error test.

# Before submitting

### DIFF
--- a/composer/utils/parallelism.py
+++ b/composer/utils/parallelism.py
@@ -9,8 +9,8 @@ from typing import Any, Optional
 from torch.distributed._tensor.device_mesh import DeviceMesh
 
 
-@dataclass
-class _FSDPConfig:
+@dataclass(kw_only=True)
+class FSDPConfig:
     """Configuration for Fully Sharded Data Parallelism (FSDP)."""
     activation_checkpointing: bool = False
     activation_checkpointing_reentrant: bool = True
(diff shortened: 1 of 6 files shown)


dakinggg (approved):
this seems fine to me, so approving. But wondering if it would be a bit easier to make `FSDPConfig` take in a `device_mesh` arg and just raise if it is specified. That way you don't have to put the error in multiple places

snarayan21:
@dakinggg I wanted to only error in the case where we used to throw the deprecation warning so this won't break users workflows. For example, if someone explicitly creates an FSDPConfig object and passes that to the trainer, that wasn't marked as deprecated and we shouldn't error. We should only error for the deprecation path, which was for dict fsdp config + dict parallelism config (this covers foundry case). lmk if you disagree

taking out the error on `_shard_with_fsdp`, mistakenly left that in

mvpatel2000 (commented):
We only want to set device mesh in `state.py`, so I agree with Daniel lets just check in constructor?

composer/utils/parallelism.py
```diff
@@ -41,6 +40,24 @@ class FSDPConfig:
     use_orig_params: bool = True
     verbose: bool = False
 
+    _device_mesh: Optional[DeviceMesh] = field(default=None, init=False, repr=False)
+
+    def __init__(self, **kwargs):
+        if 'device_mesh' in kwargs or '_device_mesh' in kwargs:
+            raise ValueError(
+                f'Directly specifying device mesh for FSDP was deprecated in Composer version 0.24.0. Please specify \'data_parallel_shard_degree\' and/or \'data_parallel_replicate_degree\' instead.',
+            )
```
mvpatel2000, line 49:
```suggestion
                f"Directly specifying device mesh for FSDP was deprecated in Composer version 0.24.0. Please specify 'data_parallel_shard_degree' and/or 'data_parallel_replicate_degree' instead.",
            )
```
nit: mildly easier

mvpatel2000 (approved):

composer/utils/parallelism.py
snarayan21: long line with this error msg was bothering me so resolved this as part of that

bigning (approved):

mvpatel2000 (approved):

PR #3585, opened 2024-08-29
Init Dist Default None

mvpatel2000:
# What does this PR do?

Defaults `dist.initialize_dist` to `device=None`, which lets `get_device` autodetect hardware.


b-chu (approved):
Thanks!

snarayan21 (approved):
nice

PR #3586, opened 2024-08-29
Explicit None Check in get_device

mvpatel2000:
# What does this PR do?

Explicit None Check in get_device


snarayan21 (approved):
i mean, functionally the same, but is more explicit yea

PR #3590, opened 2024-08-31
Update ci-testing to 0.2.2

dakinggg:
Updates the ci-testing to 0.2.2, which includes the switch to uv from pip. It also adds a numpy<2 pin to our deepspeed dependency group, as the version of deepspeed we have pinned requires numpy<2

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
+  pull_request:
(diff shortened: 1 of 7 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
```
dakinggg, line 3: dont merge

mvpatel2000 (approved):

PR #3598, opened 2024-09-04
Roll back ci-testing for daillies

mvpatel2000:
# What does this PR do?

Dailies are broken with `uv` due to mellanox versioning. Rolls back to older version with `pip` for now.

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
+  pull_request:
```
dakinggg, line 3: will approve after this is put back

dakinggg (approved):

PR #3612, opened 2024-09-10
fix 2.4.1 test

bigning:
Fix the torch 2.4.1 2-gpu tests. It missed a auto-microbatch torch patch.

### DIFF
--- a/composer/trainer/_patch_pytorch.py
+++ b/composer/trainer/_patch_pytorch.py
@@ -943,7 +943,9 @@ if version.parse(torch.__version__) >= version.parse('2.3.0') and version.parse(
         self._use_unsharded_flat_param(padded_unsharded_flat_param)
 
 
-if version.parse(torch.__version__) == version.parse('2.4.0'):
+if version.parse(torch.__version__) >= version.parse('2.4.0') and version.parse(
+        torch.__version__,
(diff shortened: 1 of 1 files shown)


composer/trainer/_patch_pytorch.py
```diff
@@ -943,54 +943,8 @@ def unshard_with_sync(self):
...
-        all_ranks_finished_tensor = torch.tensor([0], dtype=torch.uint8).to(self.device, non_blocking=True)
-
-        dist.all_reduce(all_ranks_finished_tensor, reduce_operation='MIN')
-
-        if found_cuda_oom == 1:
-            raise RuntimeError('CUDA out of memory encountered on a different rank')
-        padded_unsharded_flat_param = self._all_gather_flat_param(unsharded_flat_param)
-        self._use_unsharded_flat_param(padded_unsharded_flat_param)
-
+if version.parse(torch.__version__) == version.parse('2.4.0'):
```
mvpatel2000, line 946:
```suggestion
if version.parse(torch.__version__) >= version.parse('2.4.0') and version.parse(
        torch.__version__,
) < version.parse('2.4.1'):
```
this is preferred as it enables nightly which ahve different version

mvpatel2000 (approved):

PR #3628, opened 2024-09-23
Fix CPU dailies

mvpatel2000:
# What does this PR do?

Fix CPU dailies. Runs wandb in offline mode


dakinggg (approved):

snarayan21 (approved):
oh that's v useful

b-chu:
Shouldn't dailies also test integrations with third party providers? We already have some wandb tests as offline

mvpatel2000:
> Shouldn't dailies also test integrations with third party providers? We already have some wandb tests as offline

We don't have keys anymore as we've switched to MLFlow

b-chu:
We should get keys for our team to make sure we can keep supporting wandb integrations

PR #3629, opened 2024-09-23
fix 2.4.1ckpt

bigning:
daily test is broken: https://github.com/mosaicml/composer/actions/runs/10969280313/job/30461744791
We thought 2.4.1 included the checkpointing backward compatibility fix, but it turned out not. So we add the patch to 2.4.1 in this PR. 

daily test: https://github.com/mosaicml/composer/actions/runs/11000779231

image


mvpatel2000 (approved):
Legend

dakinggg (approved):

PR #3631, opened 2024-09-23
Remove Legacy Checkpointing

mvpatel2000:
# What does this PR do?

Remove legacy checkpointing. This codepath is incredibly brittle and buggy and has caused numerous issues. As it was only available for a brief period, we do not think it will affect end-users. We will provide an upgrade script to users who do have these checkpoints.

### DIFF
--- a/tests/trainer/test_fsdp_checkpoint.py
+++ b/tests/trainer/test_fsdp_checkpoint.py
@@ -537,6 +537,9 @@ def test_fsdp_load_old_checkpoint(
         pytest.skip('Current torch version is older than torch version that checkpoint was written with.')
 
     if composer_version in ['0.13.5', '0.14.0', '0.14.1', '0.15.1']:
+        if state_dict_type == 'sharded':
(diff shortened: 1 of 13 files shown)


tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -537,6 +537,9 @@ def test_fsdp_load_old_checkpoint(
         pytest.skip('Current torch version is older than torch version that checkpoint was written with.')
 
     if composer_version in ['0.13.5', '0.14.0', '0.14.1', '0.15.1']:
+        if state_dict_type == 'sharded':
+            pytest.mark.skip('Loading legacy sharded checkpoints are not supported after v0.25.0.')
+
         rank = 0 if state_dict_type == 'full' else '{rank}'
```
dakinggg, line 543: don't need the checkpoint type casing here anymore, yeah?

dakinggg (approved):
LGTM, but can you please run the regression and daily tests against this PR before merging?

PR #3633, opened 2024-09-23
Bump version 25

dakinggg:
# What does this PR do?
Bumps Composer version in preparation for 0.25.0 release.

### DIFF
--- a/composer/_version.py
+++ b/composer/_version.py
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.25.0.dev0'
+__version__ = '0.26.0.dev0'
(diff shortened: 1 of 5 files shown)


composer/_version.py
```diff
@@ -3,4 +3,4 @@
 
 """The Composer Version."""
 
-__version__ = '0.25.0.dev0'
+__version__ = '0.26.0.dev0'
```
mvpatel2000, line 6: this should be 
```suggestion
__version__ = '0.25.0'
```
so whne you use release branch (which will fork off this) its the right version and the tagged version is on this commit (which is on main branch)

mvpatel2000 (approved):
LGTM!

PR #3641, opened 2024-10-01
Create callback to load checkpoint

irenedea:
# What does this PR do?

Adds a new callback to load a checkpoint at a specified event. This enables use cases like loading model base weights with peft.

## Manual Test

Ran two lora finetuning runs.
`test-load-checkpoint-callback-baseline-K0m4fq`: Initialized from Huggingface checkpoint and ran for 100 batches
`test-load-checkpoint-callback-resume-GS8oWj`: Initialized using this callback and the baseline checkpoint at ba50.

We see that the losses match up.
image

Also confirmed that the adapter-only checkpoint is smaller than a full checkpoint as expected.

# Before submitting

### DIFF
--- a/composer/callbacks/load_checkpoint.py
+++ b/composer/callbacks/load_checkpoint.py
@@ -10,6 +10,7 @@ from composer.core.event import Event
 from composer.loggers import Logger
 from composer.models.huggingface import HuggingFaceModel
 from composer.utils.checkpoint import load_checkpoint
+from composer.utils.file_helpers import maybe_create_object_store_from_uri, parse_uri
 
 log = logging.getLogger(__name__)
 
@@ -33,6 +34,9 @@ class LoadCheckpoint(Callback):
     ):
         super().__init__()
         self.load_path = load_path
+        self.load_object_store = maybe_create_object_store_from_uri(load_path)
+        _, _, self.parsed_path = parse_uri(load_path)
+
         self.load_weights_only = load_weights_only
         self.strict_model_weights = strict_model_weights
         self.ignore_keys = ignore_keys
@@ -58,9 +62,10 @@ class LoadCheckpoint(Callback):
             model.should_save_peft_only = False
 
         load_checkpoint(
-            path=self.load_path,
+            path=self.parsed_path,
             state=state,
             logger=logger,
+            object_store=self.load_object_store,
             strict_model_weights=self.strict_model_weights,
             ignore_keys=self.ignore_keys,
             load_weights_only=self.load_weights_only,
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -169,6 +169,8 @@ filterwarnings = [
     # Ignore mlflow warnings about transformers versions,
     '''ignore:The 'transformers' MLflow Models integration.*:UserWarning''',
     # Ignore our own deprecation warnings,
+    '''ignore:To use flash-attn v3*:UserWarning''',
```
dakinggg, line 172: 🙏

composer/callbacks/load_checkpoint.py
```diff
@@ -0,0 +1,76 @@
...
+
+        return super().run_event(event, state, logger)
+
+    def _load(self, state: State, logger: Logger) -> None:
+
+        # We need to temporarily disable the `should_save_peft_only` flag on the model
+        # so that we can have access to the full model weights for loading.
+        model = state.model
+        original_should_save_peft_only = False
+        if isinstance(model, HuggingFaceModel):
```
dakinggg, line 60: I _think_ it might be better in this case to just check `hasattr(model, 'should_save_peft_only')`. no strong opinion, could go either way.

dakinggg (approved):
looks right to me

pyproject.toml
```diff
@@ -169,6 +169,8 @@ filterwarnings = [
...
+    # Ignore the flash v3 warnings from transformer engine
```
irenedea, line 173:
```suggestion
    # Ignore the flash v3 warnings from transformer engine
    '''ignore:To use flash-attn v3*:UserWarning''',
```

composer/callbacks/load_checkpoint.py
mvpatel2000: +1

composer/callbacks/load_checkpoint.py
```diff
@@ -0,0 +1,76 @@
...
+            path=self.parsed_path,
+            state=state,
+            logger=logger,
+            object_store=self.load_object_store,
+            strict_model_weights=self.strict_model_weights,
+            ignore_keys=self.ignore_keys,
+            load_weights_only=self.load_weights_only,
+        )
+
+        # Restore the original `should_save_peft_only` flag on the model
```
mvpatel2000, line 74: why does load_checkpoint remove this?

composer/callbacks/load_checkpoint.py
irenedea: I'd rather check `isinstance` here since the state dict filtering is implemented specifically on this class.

composer/callbacks/load_checkpoint.py
irenedea: If `should_save_peft_only` is set to True on HuggingFaceModel, it will filter out adapters from the state_dict. We want it to be set to False during loading, so we can load in any weights we want.

PR #3643, opened 2024-10-04
Switch to log.info for deterministic mode

mvpatel2000:
# What does this PR do?

Switch to `log.info` as this isn't something going wrong that requires user intervention. If a user intentionally sets this, it should not result in a warning.

Most frequently, this results in spurious warnings in our Ci/CD

### DIFF
--- a/composer/utils/reproducibility.py
+++ b/composer/utils/reproducibility.py
@@ -127,7 +127,7 @@ def configure_deterministic_mode():
     # See https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html
     # and https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
     os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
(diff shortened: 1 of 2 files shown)


composer/utils/reproducibility.py
```diff
@@ -127,7 +127,7 @@ def configure_deterministic_mode():
     # See https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html
     # and https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
     os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
-    warnings.warn('Deterministic mode is activated. This will negatively impact performance.', category=UserWarning)
+    log.warning('Deterministic mode is activated. This will negatively impact performance.', category=UserWarning)
```
snarayan21, line 130: log.info?

snarayan21 (commented):
according to PR description, likely log.info right?

composer/utils/reproducibility.py
mvpatel2000: Oops forgot to repush

snarayan21 (approved):
lgtm

PR #3654, opened 2024-10-14
Replace list with Sequence

KuuCi:
# What does this PR do?

This PR is to replace some of the type checking. Currently, If we have `RandomClass(BaseClass)`, if we have pyright check for `list[BaseClass]`, it will throw a pyright error if user is passing in `list[RandomClass]`. Instead, we should replace with `Sequence[BaseClass]`
https://github.com/microsoft/pyright/issues/130

Specifically in our case, if we were to make a wrapper around `torchmetrics.Metrics`, a pyright error will be thrown because it will expect list[Metrics], instead of whatever class is inheriting `torchmetrics.Metrics`


irenedea (approved):
lgtm! Thank you!

PR #3672, opened 2024-10-21
Don't overwrite existing excepthook in mlflow logger

dakinggg:
# What does this PR do?
The [previous PR](https://github.com/mosaicml/composer/pull/3449) that added exception hook handling to mlflow just blindly overwrites the existing excepthook. This PR changes that to _use_ the existing excepthook and just _add_ what the mlflow logger needs.

Before: `llama-7b-commits-6-D3ipHw` (exception is not rich formatted, because systemwide excepthook is not triggered)
After: `llama-7b-commits-11-3KdS8L` (exception is rich formatted, because systemwide excepthook is triggered)

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -304,7 +304,7 @@ class MLFlowLogger(LoggerDestination):
     def _global_exception_handler(self, original_excepthook, exc_type, exc_value, exc_traceback):
         """Catch global exception."""
         self._global_exception_occurred += 1
-        sys.excepthook(exc_type, exc_value, exc_traceback)
+        original_excepthook(exc_type, exc_value, exc_traceback)
 
     def init(self, state: State, logger: Logger) -> None:
         del logger  # unused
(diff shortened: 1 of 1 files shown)


siriuslee (approved):
🎉

irenedea (approved):
LGTM, good catch!

composer/loggers/mlflow_logger.py
```diff
@@ -322,7 +322,13 @@ def init(self, state: State, logger: Logger) -> None:
             self.run_name += f'-rank{dist.get_global_rank()}'
 
         # Register the global exception handler so that uncaught exception is tracked.
-        sys.excepthook = self._global_exception_handler
+        original_excepthook = sys.excepthook
```
irenedea, line 325: alternatively you could save `original_excepthook` on self (i.e. `self.original_excepthook`), so you don't need to use a lambda below

dakinggg: yeah i think lambda is ok. could've also done functools.partial. don't feel particularly strongly between any of those options

mvpatel2000 (approved):
LGTM, good catch

dakinggg:
Closing in favor of #3675

PR #3673, opened 2024-10-21
Tmpdir checkpoints

irenedea:
# What does this PR do?

# Before submitting

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -299,7 +299,8 @@ class CheckpointSaver(Callback):  # noqa: D101
         if local_folder == '':
             local_folder = '.'
 
-        if backend != '':  # If we are uploading to a remote path, we use a temporary directory to save local checkpoints.
+        is_remote_folder = backend != ''
+        if is_remote_folder:  # If we are uploading to a remote path, we use a temporary directory to save local checkpoints.
             local_folder = os.path.join(tempfile.mkdtemp(), local_folder)
 
         filename = str(filename)
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -294,9 +294,13 @@ class CheckpointSaver(Callback):  # noqa: D101
         num_concurrent_uploads: int = 1,
         upload_timeout_in_seconds: int = 3600,
     ):
+
         backend, _, local_folder = parse_uri(str(folder))
         if local_folder == '':
             local_folder = '.'
+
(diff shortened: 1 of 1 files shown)


composer/callbacks/checkpoint_saver.py
```diff
@@ -294,10 +294,15 @@ def __init__(
         num_concurrent_uploads: int = 1,
         upload_timeout_in_seconds: int = 3600,
     ):
+
         backend, _, local_folder = parse_uri(str(folder))
         if local_folder == '':
             local_folder = '.'
 
+        is_remote_folder = backend != ''
```
snarayan21, line 302: nit: is there a more robust check we could have for this?

snarayan21 (commented):
looks good, but want someone on checkpointing side to have a look

composer/callbacks/checkpoint_saver.py
```diff
@@ -294,10 +294,15 @@ def __init__(
...
+        if is_remote_folder:  # If we are uploading to a remote path, we use a temporary directory to save local checkpoints.
```
mvpatel2000, line 303:
```suggestion
        if is_remote_folder:  # If uploading to a remote path, use a temporary directory to save local checkpoints.
```
nit: more concise

mvpatel2000 (commented):
I'm not 100% sure this is OK. `/tmp` folders may not have enough space to store all ckpts, depending on machine...

@willgleich would this have worked on azure cluster? Any general takes?

irenedea:
@mvpatel2000 Another benefit of using a tmpdir is that we can configure it via environment variable, so we can redirect it easily if needed to (which we'll need to do for notebooks)

mvpatel2000 (approved):
We should monitor this closely. Probably fine.

composer/callbacks/checkpoint_saver.py
mvpatel2000: https://github.com/mosaicml/composer/pull/3673/files#diff-b3be1fd7d6bb329e7dfcb406b7996e3ef170416faf5524caf50a451333b5f29eR348-R351 

We do it here, so I think it's fine

willgleich:
Yeah this seems fine to me, the only worry I see here is checkpoints filling up the disk. /tmp is special in azure as its one of the two mount points for the nvme. On our other clusters we leverage ephemeral storage so /tmp would be included in that.

That being said, [tempfile.mkdtemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) requires manual clean up for the files and directory. Something to consider here if you are hoping for these to be taken care of. 

IIRC there is a context manager that can handle this in a clean way if you do want automatic clean up.

PR #3677, opened 2024-10-22
Loss gen tokens

dakinggg:
# What does this PR do?
This PR allows differentiation between loss generating tokens and total tokens, by allowing `get_num_tokens_in_batch` to return a dict with `total` and `loss_generating` keys. This will only have an impact if you are using the `accumulate_batch_on_tokens` flag. This does not break or change any existing functionality, as by default, if you are just returning a single number from `get_num_tokens_in_batch`, we will accumulate using that value. Things will only change if the user adjusts their `get_num_tokens_in_batch` to be a dictionary.

Related PR in LLM Foundry: https://github.com/mosaicml/llm-foundry/pull/1610

Manual testing (using Foundry draft PR):

Before with all tokens loss generating, microbatching is deterministic:
Screenshot 2024-10-22 at 8 15 40 PM

Before with not all tokens loss generating, microbatching is not deterministic:
Screenshot 2024-10-22 at 8 15 56 PM

After with not all tokens loss generating, microbatching is deterministic:
Screenshot 2024-10-22 at 8 16 13 PM

### DIFF
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -192,7 +192,7 @@ class DataSpec:
         self.device_transforms = self._default_device_transforms if device_transforms is None else device_transforms
         self.split_batch = default_split_batch if split_batch is None else split_batch
         self.get_num_samples_in_batch = self._default_get_num_samples_in_batch if get_num_samples_in_batch is None else get_num_samples_in_batch
-        self.get_num_tokens_in_batch = self._default_get_num_tokens_in_batch if get_num_tokens_in_batch is None else get_num_tokens_in_batch
+        self._get_num_tokens_in_batch = self._default_get_num_tokens_in_batch if get_num_tokens_in_batch is None else get_num_tokens_in_batch
 
         if num_samples is not None:
             self.num_samples = num_samples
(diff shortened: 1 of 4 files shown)


composer/core/data_spec.py
```diff
@@ -295,6 +298,23 @@ def _default_get_num_tokens_in_batch(self, batch: Batch) -> int:
...
         return 0
 
+    def get_num_tokens_in_batch(self, batch: Batch, token_type: str = 'total') -> int:
+        num_tokens = self._get_num_tokens_in_batch(batch)
+
+        if isinstance(num_tokens, int):
+            if token_type != 'total':
+                log.warning(
+                    f'get_num_tokens_in_batch returned an int, but token_type is {token_type}. ' +
+                    'Returning the total number of tokens in the batch.',
```
mvpatel2000, line 308:
```suggestion
            if token_type != 'total':
                log.warning(
                    f'get_num_tokens_in_batch returned an int, but token_type was specified as {token_type}. ' +
                    'Returning the total number of tokens in the batch.',
```
nit: make more clear that this was user set

composer/core/data_spec.py
```diff
@@ -295,6 +298,23 @@ def _default_get_num_tokens_in_batch(self, batch: Batch) -> int:
             return self.dataloader.dataset.max_seq_len * samples_per_batch  # type: ignore
         return 0
 
+    def get_num_tokens_in_batch(self, batch: Batch, token_type: str = 'total') -> int:
+        num_tokens = self._get_num_tokens_in_batch(batch)
+
+        if isinstance(num_tokens, int):
+            if token_type != 'total':
+                log.warning(
+                    f'get_num_tokens_in_batch returned an int, but token_type is {token_type}. ' +
```
mvpatel2000, line 307: this is going to occur every step which is really annoying. It should only be emitted once imo... or I would rather not emit it otherwise

mvpatel2000 (commented):
Some concern on warning spam, otherwise fine

composer/core/data_spec.py
dakinggg: ah yeah, good point, will remove

mvpatel2000 (approved):

PR #3678, opened 2024-10-23
Metrics dedup for MLflow logger

chenmoneygithub:
# What does this PR do?

Context: MLflow has receiving lots of data... so we want to save some budget.

In this PR, if we detect that a metric has the same value as the previous step, we don't log it unless it meets one of the following 2 criteria:

- There has been `log_duplicated_metric_every_n_steps` skipped since last step logging, default to 100.
- There has been `log_duplicated_metric_every_n_millis` passed since last step logging, default to 600000 (10 minutes).

We put this exception for sampling purpose, otherwise it will be challenging to capture the duplicated values during sampling, despite the duplicated value occupies a significant portion of the metric history.

Tested with LLM foundry quickstart, and 3 setup: 1) no dedupping 2) log duplicated steps per 2 steps 3) log duplicated steps per 10 steps. No problem spotted. See testing result below:

image

image

image

# Before submitting

### DIFF
--- a/composer/loggers/mlflow_logger.py
+++ b/composer/loggers/mlflow_logger.py
@@ -34,6 +34,8 @@ log = logging.getLogger(__name__)
 __all__ = ['MLFlowLogger']
 
 DEFAULT_MLFLOW_EXPERIMENT_NAME = 'my-mlflow-experiment'
+LOG_DUPLICATED_METRIC_VALUE_PER_N_STEPS = 100
+LOG_DUPLICATED_METRIC_VALUE_PER_N_MILLIS = 600000
 
 
 class MlflowMonitorProcess(multiprocessing.Process):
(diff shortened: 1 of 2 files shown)


mvpatel2000 (commented):
Please add manual test + link to run/plots!

composer/loggers/mlflow_logger.py
```diff
@@ -118,6 +120,11 @@ class MLFlowLogger(LoggerDestination):
         logging_buffer_seconds (int, optional): The amount of time, in seconds, that MLflow
             waits before sending logs to the MLflow tracking server. Metrics/params/tags logged
             within this buffer time will be grouped in batches before being sent to the backend.
+        log_duplicated_metric_every_n_steps (int, optional): The number of steps to wait before
+            logging the duplicated metric value. Duplicated metric value means the new step has the
+            same value as the previous step. (default: ``100``)
+        log_duplicated_metric_every_n_millis (int, optional): The number of milliseconds to wait
+            before logging the duplicated metric value. (default: ``600000``)
```
mvpatel2000, line 127: honestly this is probably fine to remove, I can't think of a good use case for this and it'll simplify things

chenmoneygithub: sg!

mvpatel2000 (approved):
LGTM. But do we need to log every `n` steps if its a duplicated value? why not just never log it?

PR #3682, opened 2024-10-25
Don't error if some batch slice has no loss generating tokens

dakinggg:
# What does this PR do?
Allow microbatches to have 0 loss generating tokens


mvpatel2000 (commented):
emit warning

mvpatel2000 (approved):
nvm, too much spam from warning. I think its fine

PR #3688, opened 2024-10-28
Unwrap mlflow 401/403 errors

mattyding:
# What does this PR do?

Mlflow wraps any object store 401/403 as a 500 InternalException which is bad for error handling. We can unwrap by checking the associated message.

# Before submitting

### DIFF
--- a/composer/utils/object_store/mlflow_object_store.py
+++ b/composer/utils/object_store/mlflow_object_store.py
@@ -61,9 +61,16 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
     ]
     retryable_client_codes = [ErrorCode.Name(code) for code in [ABORTED, REQUEST_LIMIT_EXCEEDED, RESOURCE_EXHAUSTED]]
     not_found_codes = [ErrorCode.Name(code) for code in [RESOURCE_DOES_NOT_EXIST, NOT_FOUND, ENDPOINT_NOT_FOUND]]
-
+    # MLflow wraps Azure data exceptions as INTERNAL_ERROR. Need to unwrap and check msg for the specific error.
+    non_retryable_internal_error_codes = [
+        '401',
+        '403',
+    ]
     if isinstance(e, MlflowException):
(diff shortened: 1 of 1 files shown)


composer/utils/object_store/mlflow_object_store.py
```diff
@@ -61,9 +61,16 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
...
+    # MLflow wraps Azure data exceptions as INTERNAL_ERROR. Need to unwrap and check msg for the specific error.
+    non_retryable_internal_error_codes = [
+        '401',
+        '403',
+    ]
     if isinstance(e, MlflowException):
         error_code = e.error_code  # pyright: ignore
+        if error_code == ErrorCode.Name(INTERNAL_ERROR):
+            if any(e.message.startswith(code) for code in non_retryable_internal_error_codes):
+                raise e
```
nancyhung, line 73: why not raise a the 401 or 403 here instead of getting swallowed in the original error message?

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -61,9 +61,16 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
...
-
+    # MLflow wraps Azure data exceptions as INTERNAL_ERROR. Need to unwrap and check msg for the specific error.
+    non_retryable_internal_error_codes = [
+        '401',
+        '403',
+    ]
     if isinstance(e, MlflowException):
         error_code = e.error_code  # pyright: ignore
+        if error_code == ErrorCode.Name(INTERNAL_ERROR):
+            if any(e.message.startswith(code) for code in non_retryable_internal_error_codes):
```
nancyhung, line 72: do these error messages always start with `401`??

mattyding: I don't think there are any example of 401 private links. I was just including it because there was a comment at some point about wrapping both 401 and 403 errors

PR #3689, opened 2024-10-29
Private link error handling

nancyhung:
# What does this PR do?

Throw permission error if swallowed in mlflow exception stacktrace.

# Before submitting

### DIFF
--- a/composer/utils/object_store/mlflow_object_store.py
+++ b/composer/utils/object_store/mlflow_object_store.py
@@ -72,7 +72,7 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
         if error_code == ErrorCode.Name(INTERNAL_ERROR):
             error_message = e.message # pyright: ignore
             if any(code in error_message for code in non_retryable_internal_error_codes):
-                raise PermissionError(error_message)
+                raise PermissionError(f'Permission denied for object {uri} from the data provider. Details: {error_message}') from e
         elif error_code in retryable_server_codes or error_code in retryable_client_codes:
             raise ObjectStoreTransientError(error_code) from e
         elif error_code in not_found_codes:
--- a/composer/utils/object_store/mlflow_object_store.py
+++ b/composer/utils/object_store/mlflow_object_store.py
@@ -71,7 +71,7 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
         error_code = e.error_code  # pyright: ignore
         if error_code == ErrorCode.Name(INTERNAL_ERROR):
             error_message = e.message # pyright: ignore
-            if any(code in error_message for code in non_retryable_internal_error_codes):
+            if any(f'{code} Client Error' in error_message for code in non_retryable_internal_error_codes):
                 raise PermissionError(f'Permission denied for object {uri} from the data provider. Details: {error_message}') from e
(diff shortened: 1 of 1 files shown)


composer/utils/object_store/mlflow_object_store.py
```diff
@@ -53,7 +53,6 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
     retryable_server_codes = [
         ErrorCode.Name(code) for code in [
             DATA_LOSS,
-            INTERNAL_ERROR,
```
nancyhung, line 56: why were we retrying on 500s before? @jerrychen109

dakinggg: because 500s can be many things, since mlflow is wrapping all server errors as 500. some of them might be retryable, so we retry

irenedea: I think we still want to retry on those? There may be other internal errors that we would need to retry on

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -62,9 +61,19 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
...
+        '401',
+        '403',
+    ]
+
     if isinstance(e, MlflowException):
         error_code = e.error_code  # pyright: ignore
-        if error_code in retryable_server_codes or error_code in retryable_client_codes:
+        if error_code == ErrorCode.Name(INTERNAL_ERROR):
+            error_message = e.message # pyright: ignore
+            if any(code in error_message for code in non_retryable_internal_error_codes):
```
mattyding, line 74: I don't know if this is safe from false positives because the error message contains a long URL with many numbers in it.
```
requests.exceptions.HTTPError: 403 Client Error: This request is not authorized to perform this operation. for url: https:<long url with lots of numbers in it>
```

mattyding:
```suggestion
            if any(f'{code} Client Error' in error_message for code in non_retryable_internal_error_codes):
```

composer/utils/object_store/mlflow_object_store.py
nancyhung: hm ok that's still sus to me though

composer/utils/object_store/mlflow_object_store.py
```diff
@@ -62,9 +62,19 @@ def _wrap_mlflow_exceptions(uri: str, e: Exception):
...
+        '403',
+    ]
+
     if isinstance(e, MlflowException):
         error_code = e.error_code  # pyright: ignore
-        if error_code in retryable_server_codes or error_code in retryable_client_codes:
+        if error_code == ErrorCode.Name(INTERNAL_ERROR):
+            error_message = e.message # pyright: ignore
+            if any(f'{code} Client Error' in error_message for code in non_retryable_internal_error_codes):
+                raise PermissionError(f'Permission denied for object {uri} from the data provider. Details: {error_message}') from e
```
dakinggg, line 76: since we're just raising permissions error here, can we rename the list of codes variable name to something indicating that?

dakinggg (approved):
need to run pre commit for ci to pass

PR #3691, opened 2024-10-29
Lower min torchmetrics version

mvpatel2000:
# What does this PR do?

Lower min torchmetrics version. https://github.com/mosaicml/composer/pull/3460/ raised it and we instead intend to stay no >1.0


dakinggg (approved):
Approving, cause if CI passes than lgtm, but CI is not passing

PR #3693, opened 2024-10-31
add time units to throughput metrics and also secs_per_step metric.

ekelsen:
# What does this PR do?

The time units didn't have units, this makes the units explicit and adds a new secs_per_step metric.

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -258,6 +258,7 @@ class SpeedMonitor(Callback):
         self.history_flops: deque[float] = deque(maxlen=window_size + 1)
 
         self.gpu_flops_available = gpu_flops_available
+        self.time_unit = time_unit
 
         self.divider = 1
         if time_unit == 'seconds':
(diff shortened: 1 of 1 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -360,10 +361,18 @@ def batch_end(self, state: State, logger: Logger):
...
+        if len(self.history_wct) > 1:
+            secs_per_step = self.history_wct[-1] - self.history_wct[-2]
+
         logger.log_metrics({
             'time/train': train_wct / self.divider,
             'time/val': self.total_eval_wct / self.divider,
             'time/total': (train_wct + self.total_eval_wct) / self.divider,
+            f'time_{self.time_unit}/train': train_wct / self.divider,
+            f'time_{self.time_unit}/val': self.total_eval_wct / self.divider,
+            f'time_{self.time_unit}/total': (train_wct + self.total_eval_wct) / self.divider,
```
mvpatel2000, line 374: Let's just rename `time/train` -> `time/train/hours`? 

Changing log keys is a little annoying to end user... but duplicate logging is more annoying imo.

composer/callbacks/speed_monitor.py
```diff
@@ -360,10 +361,18 @@ def batch_end(self, state: State, logger: Logger):
         # Log the time
         # `state.timestamp` excludes any time spent in evaluation
         train_wct = state.timestamp.total_wct.total_seconds()
+        secs_per_step = 0
+        if len(self.history_wct) > 1:
+            secs_per_step = self.history_wct[-1] - self.history_wct[-2]
```
mvpatel2000, line 366: This should probably average over history_wct as all metrics respect `window_size`. Something like:
`elapsed_wct = (self.history_wct[-1] - self.history_wct[0]) / (len(history-wct)-1)`

mvpatel2000 (commented):
Any specific use case for secs_per_step?

composer/callbacks/speed_monitor.py
dakinggg: @mvpatel2000 i was a bit worried this is a breaking change for any downstream pipelines that process metrics.

mvpatel2000: Discussed offline, it's better to avoid overwhelming loggers

Issue #3699, opened 2024-11-04
`DataSpec.device_transforms` do not run on device


Ghelfi:
## Description

Since release 0.25.0 `DataSpec.device_transforms` do not run on device.

2 successive PR changed where batches / micro_batches are moved to device and where `device_transforms` are applied in an inconsistent way:
* [First PR](https://github.com/mosaicml/composer/pull/3487) moved `device_transforms` and copy at the microbatch level
* [Second PR](https://github.com/mosaicml/composer/pull/3567) brought back the `device_transforms` before moving the batch to device.

Hence, `device_transforms` are applied by batch but systematically on the `cpu`.

Ghelfi:
Here is a snippet to reporduce the error:
```python
from collections.abc import Callable
from typing import cast

import torch.nn.functional as F
from composer import Trainer
from composer.core import DataSpec
from composer.models import ComposerClassifier
from torch import Tensor, nn
from torch.utils.data import DataLoader
from torchvision import datasets, transforms


class Model(nn.Module):
    def __init__(self, num_classes: int = 10):
        super().__init__()

        self.num_classes = num_classes
        self.conv = nn.Conv2d(1, 16, (3, 3), padding=0)
        self.fc = nn.Linear(16, num_classes)

    def forward(self, x: Tensor) -> Tensor:
        out = self.conv(x)
        out = F.relu(out)
        return cast(Tensor, self.fc(out))


ef get_device_transform(device: str) -> Callable[[tuple[Tensor, Tensor]], tuple[Tensor, Tensor]]:
    def _transform(batch: tuple[Tensor, Tensor]) -> tuple[Tensor, Tensor]:
        """Dummy on device transform to check `batch` is actally on the right device."""
        assert (
            batch[0].device.type == device
        ), f"found data on device {batch[0].device.type} while expecting device_transform to run on {device}"
        assert (
            batch[1].device.type == device
        ), f"found data on device {batch[1].device.type} while expecting device_transform to run on {device}"

        return batch

    return _transform


transform = transforms.Compose([transforms.ToTensor()])
dataset = datasets.MNIST("data", train=True, download=True, transform=transform)
train_dataloader = DataSpec(
    dataloader=DataLoader(dataset, batch_size=16),
    device_transforms=get_device_transform(device="cuda"),
    get_num_samples_in_batch=lambda batch: len(batch[0]),
)

trainer = Trainer(
    model=ComposerClassifier(module=Model(), num_classes=10),
[...]

PR #3703, opened 2024-11-05
Add batch/microbatch transforms

mvpatel2000:
# What does this PR do?

https://github.com/mosaicml/composer/pull/3566 moved device transfers into the microbatch loop. This results in device transformations occurring on CPU, which can be slow as pointed out in https://github.com/mosaicml/composer/issues/3699.

To remedy this, we deprecate `device_transforms` (which was poorly named anyways) and instead add `batch_transforms` and `microbatch_transforms` to dataspec. Batch transforms happens on CPU, and microbatch transforms happens on GPU after the device transfer.

Additionally, if users want to do batch level transforms on GPU, they can add a transform that moves to device.

### DIFF
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -215,7 +215,7 @@ class DataSpec:
                     'The `device_transforms` argument is deprecated. Please use `batch_transforms` for batch level '
                     'transformations on CPU and `microbatch_transforms` for microbatch level transformations on target '
                     'device.',
-                    'v0.28.0',
+                    'v0.29.0',
                 ),
             )
             self.batch_transforms = device_transforms
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -216,7 +216,7 @@ class DataSpec:
                     'transformations on CPU and `microbatch_transforms` for microbatch level transformations on target '
                     'device.',
                     'v0.28.0',
-                )
+                ),
             )
             self.batch_transforms = device_transforms
         self.batch_transforms = self._default_transforms if batch_transforms is None else batch_transforms
--- a/composer/core/data_spec.py
+++ b/composer/core/data_spec.py
@@ -126,16 +126,16 @@ default_split_batch = _default_split_batch
 class DataSpec:
     """Specifications for operating and training on data.
 
-    An example of constructing a :class:`DataSpec` object with a ``device_transforms``
+    An example of constructing a :class:`DataSpec` object with a ``batch_transforms``
     callable and then using it with :class:`~.Trainer`:
 
     .. doctest::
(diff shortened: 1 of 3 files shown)


composer/core/data_spec.py
```diff
@@ -155,11 +155,20 @@ class DataSpec:
         num_tokens (int, optional): The total number of tokens in an epoch. This field is used by the
             :class:`.Timestamp` (training progress tracker).
 
-        device_transforms ((Batch) -> Batch, optional): Function called by the :class:`.Trainer` to modify the
-            batch once it has been moved onto the device. For example, this function can be used for GPU-based
+        device_transforms ((Batch) -> Batch, optional): Deprecated argument. Please use ``batch_transforms`` for batch
+            level transformations on CPU and ``microbatch_transforms`` for microbatch level transformations on target
+            device.
```
antoinebrl, line 160: Should `device_transforms` be kept but applied to the micro-batches. The name remains valid because the transformation is performed after the micro-batch is moved to the requested device. This would eliminate the need for deprecating this argument.

antoinebrl:
Thanks @mvpatel2000 ! Will there be a new release following this merge? The regression introduced in `0.25.0` is a blocker for us as our trainings are 10x slower.

composer/core/data_spec.py
mvpatel2000: I'm for renaming it to make it more clear where it runs. I'm concerned about subtle logic bugs that may emerge otherwise as people do not realize it does not run at batch level.

mvpatel2000:
> Thanks @mvpatel2000 ! Will there be a new release following this merge? The regression introduced in `0.25.0` is a blocker for us as our trainings are 10x slower.

We're still figuring out when it would be released. To unblock, I recommend adding a transform that just moves batch to device.

antoinebrl:
> To unblock, I recommend adding a transform that just moves batch to device.

Thanks! That is exactly the patch we have right now 👍

composer/trainer/trainer.py
```diff
@@ -3310,7 +3311,8 @@ def predict_batch_end(self, state: State, logger: Logger) -> None:
                 self.state.batch = self.state.device.batch_to_device(self.state.batch)
 
                 # Perform any device transforms
-                self.state.batch = data_spec.device_transforms(self.state.batch)
+                self.state.batch = data_spec.batch_transforms(self.state.batch)
```
dakinggg, line 3314: shouldn't the batch transform be before the device move for consistency with other parts of the code?

antoinebrl (approved):
LGTM! Thanks

dakinggg (approved):

PR #3710, opened 2024-11-13
Add torch 2.5.1 composer tests

j316chuck:
# What does this PR do?
Add torch 2.5.1 composer tests

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -31,12 +31,12 @@ jobs:
           pytest_command: coverage run -m pytest
           composer_package_name: mosaicml
         - name: cpu-3.11-2.5
-          container: mosaicml/pytorch:2.5.0_cpu-python3.11-ubuntu20.04
+          container: mosaicml/pytorch:2.5.1_cpu-python3.11-ubuntu20.04
           markers: not daily and (remote or not remote) and not gpu and not doctest
           pytest_command: coverage run -m pytest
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -262,7 +262,7 @@ RUN if [ -n "$MOFED_VERSION" ] ; then \
 RUN if [ -n "$CUDA_VERSION" ] ; then \
         pip${PYTHON_VERSION} install --upgrade --no-cache-dir ninja==1.11.1 && \
         pip${PYTHON_VERSION} install --upgrade --no-cache-dir --force-reinstall packaging==22.0 && \
-        MAX_JOBS=1 pip${PYTHON_VERSION} install --no-cache-dir --no-build-isolation flash-attn==2.6.3; \
+        pip${PYTHON_VERSION} install flash-attn==2.6.3; \
```
mvpatel2000, line 265: why these changes?

mvpatel2000 (approved):

docker/Dockerfile
j316chuck: just testing if we can not spend 2 hours building flash attention .

dakinggg:
btw, FA has pre built wheels that we could probably use instead of pip instal: https://github.com/Dao-AILab/flash-attention/releases/tag/v2.7.0.post2 (see "assets" under each release)

PR #3712, opened 2024-11-14
Fix composer gpu daily test to use torch 2.5.1

j316chuck:
# What does this PR do?
Fix composer gpu daily test to use torch 2.5.1


mvpatel2000 (approved):
moon

PR #3719, opened 2024-11-20
mean_resizing = True does not work with mixed/meta initialization

jdchang1:
# What does this PR do?

Transformers recently added in `mean_resizing` to `resize_token_embeddings`. This is breaking with mixed initialization in downstream training tasks that requires adding tokens to Composer Huggingface Models.  This PR sets this value to False for now rather than defaulting to True.


dakinggg (commented):
For posterity/if we want to fix this, could you explain why the mean_resizing doesn't work with meta but the old version does?

mvpatel2000 (commented):
LGTM besides Daniel's comment which will fix tests

PR #3724, opened 2024-11-26
Expose `DistributedSampler` RNG seed argument

### LINKED ISSUE #3723: DistributedSampler seeding
[The `DistributedSampler` is constructed with its default constant `seed=0` even when the `Trainer` is given a different seed.](https://github.com/mosaicml/composer/blob/b8cc29e27ddc3955fb445e5acd4a47ee444182cf/composer/utils/dist.py#L617-L623) (The sampler's `seed` attribute is also never set before its iteration, which would also work.) To me, this is unexpected, as I would expect the data distribution to also change when I change the global seed. As an alternative option to including the global seed when setting up the data sampler, it would be feasible to allow setting a separate `data_seed` with a default value of 0 (which would also preserve full backward compatibility).

** Environment **
Composer `main` (b8cc29e27ddc3955fb445e5acd4a47ee444182cf).

## Expected behavior

My personal expectation is that a changed seed would vary the data sampler's seed as well. However, this may conflict with your views on what changing the seed should change. Thus the proposal for a separate seed to control for data reproducibility.

janEbert:
This way, it becomes easier to randomly vary the order of the data.

# What does this PR do?

Exposes the `DistributedSampler`'s `seed` argument in the `get_sampler` utility function. Since the default value is already `seed=0`, https://github.com/mosaicml/composer/issues/3723. 

Fixes #3723

# Before submitting
- [X] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md)?
- [X] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md#prerequisites))

@mvpatel2000

### DIFF
--- a/composer/utils/dist.py
+++ b/composer/utils/dist.py
@@ -592,6 +592,7 @@ def get_sampler(
     shuffle: bool = False,
     num_replicas: Optional[int] = None,
     rank: Optional[int] = None,
+    seed: int = 0,
 ):
     """Constructs a :class:`~torch.utils.data.distributed.DistributedSampler` for a dataset.
 
@@ -620,6 +621,7 @@ def get_sampler(
         shuffle=shuffle,
         num_replicas=get_world_size() if num_replicas is None else num_replicas,
         rank=get_global_rank() if rank is None else rank,
+        seed=seed,
     )
 
 


dakinggg (approved):
Thanks!

mvpatel2000 (approved):

PR #3757, opened 2025-01-30
Update cpu check to have better warning

KuuCi:
# What does this PR do?

# Before submitting

### DIFF
--- a/composer/callbacks/memory_monitor.py
+++ b/composer/callbacks/memory_monitor.py
@@ -133,7 +133,7 @@ class MemoryMonitor(Callback):
         # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
         model_device = next(state.model.parameters()).device
 
-        if model_device.type not in ('cuda', 'meta'):
+        if model_device.type == 'cpu':
             warnings.warn(f'The memory monitor only works on CUDA devices, but the model is on {model_device.type}.')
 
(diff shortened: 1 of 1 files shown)


irenedea:
I think this will still give a warning for mixed init right? Because mixed is a concept in foundry only that means the rank 0 gets init on cpu while the others are meta, so rank 0 will still give this warning.

irenedea:
I'm seeing 
```
/usr/lib/python3/dist-packages/composer/callbacks/memory_monitor.py:137: UserWarning: The memory monitor only works on CUDA devices, but the model is on cpu.
  warnings.warn(f'The memory monitor only works on CUDA devices, but the model is on {model_device.type}.')
``` 
in `mcli logs embedding-ft-Lf9M5o`

composer/callbacks/memory_monitor.py
```diff
@@ -133,7 +133,8 @@ def init(self, state: State, logger: Logger) -> None:
         # Not relying on `torch.cuda.is_available()` since the model could be on CPU.
         model_device = next(state.model.parameters()).device
 
-        if model_device.type not in ('cuda', 'meta'):
+        print('----UNGA BUNGA', model_device.type)
+        if model_device.type == 'cpu':
```
irenedea, line 137: Can you try moving the original warning to after_load? At that point, I think the model should be on GPU only if using mixed init

PR #3763, opened 2025-02-08
Bump pytorch to 2.6.0

rithwik-db:
# What does this PR do?

Upgrading to Torch 2.6

Run of daily tests: https://github.com/mosaicml/composer/actions/runs/13463042795

# Before submitting
- [x] Is this change a documentation change or typo fix? If so, skip the rest of this checklist.


.github/workflows/pr-gpu.yaml
```diff
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
+  pull_request:
```
bigning, line 3: IIRC, we need to change it back before merging?

bigning (commented):
LGTM, please manually trigger a daily test on this branch https://github.com/mosaicml/composer/actions/workflows/daily.yaml . The currently daily test is broken (cc oncall @milocress ), but you can temporarily remove that failed tests.

.github/workflows/pr-gpu.yaml
rithwik-db: Yeah will plan on doing that once the `daily` tests are done and I will add a screenshot of all of the working tests here before I make a final commit to change this.

composer/algorithms/factorize/factorize_modules.py
```diff
@@ -96,7 +96,7 @@ def _apply_solution_to_module_parameters(
 
     with torch.no_grad():
         # first op always has no bias since adds no expressivity
-        if module0.bias is not None:
```
dakinggg, line 99: did i see you say somewhere that this check will fail..?

composer/metrics/metrics.py
```diff
@@ -53,8 +53,8 @@ def update(self, logits: Tensor, targets: Tensor):
             area_prediction = torch.histc(pred.float(), bins=self.num_classes, min=0, max=self.num_classes - 1)
             area_target = torch.histc(target.float(), bins=self.num_classes, min=0, max=self.num_classes - 1)
 
-            self.total_intersect += area_intersect
-            self.total_union += area_prediction + area_target - area_intersect
+            self.total_intersect += area_intersect  # type: ignore
```
dakinggg, line 56: there are a lot of type ignores added. should i be worried about that?

composer/algorithms/factorize/factorize_modules.py
rithwik-db: It's not that this check will fail, it's that this check is meaningless since bias can't be `None` anymore, the type hint says it can only be `Tensor | Module`

rithwik-db: Since the tests should cover this (called in the `set_rank` function which is tested a few times in `test_factorized_modules.py`), I am assuming that this will be fine (same with other cases of this type of `type: ignore`)

composer/metrics/metrics.py
rithwik-db: I can add some type assertions in here instead:
```
assert isinstance(self.total_intersect, Tensor)
assert isinstance(self.total_union, Tensor)
self.total_intersect += area_intersect
self.total_union += area_prediction + area_target - area_intersect
```
instead since the issue that's reported is 
```
Operator "+" not supported for types "Module" and "Tensor" (reportGeneralTypeIssues)
```

dakinggg: yes, throughout, if you _know_ the type of the thing and just need to tell the type checker, assertion is preferred over type ignore, because its more clear for the reader and validates the invariant rather than ignoring it.

docker/generate_build_matrix.py
```diff
@@ -45,12 +45,12 @@ def _get_cuda_version(pytorch_version: str, use_cuda: bool):
     # From https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/
     if not use_cuda:
         return ''
+    if pytorch_version == '2.6.0':
+        return '12.4.1'
```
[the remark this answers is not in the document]
rithwik-db: We decided to stick with 12.4 (discussed [here](https://databricks.slack.com/archives/C08AE4S78MV/p1738782361323509) since it's the default cuda version for 12.4)

composer/algorithms/factorize/factorize_modules.py
rithwik-db: Actually this should be fine. We have code [here](https://github.com/pytorch/pytorch/blame/v2.6.0/torch/nn/modules/module.py#L613-L614) that supports parameters with `None` and `torch.nn.Linear` still follows this functionality [here](https://github.com/pytorch/pytorch/blob/v2.6.0/torch/nn/modules/linear.py#L108-L111). I think that they should have added `None` (alongside other supported types) as another valid type instead of just sticking with `[Tensor, Module]` since their [PR](https://github.com/pytorch/pytorch/pull/141240) is about fixing just the type safety and doesn't seem like a functional change... In this situation, it might just be best to stick with the `None` checks and add `# type: ignore`. Does this sound good to you? Also please feel free to let me know if there's something that I'm missing :)

composer/metrics/metrics.py
rithwik-db: Added this fix to most of the `# type: ignore` changes, but I left it for changes that seem a bit esoteric/confusing to read or invalid (e.g. the case of the remaining `test_fsdp.py` ignores and also the ignores related to module parameters equalling `None`)

composer/algorithms/factorize/factorize_modules.py
dakinggg: oh yeah i dont think we should be removing the bias none checks.

composer/trainer/_patch_pytorch.py
```diff
@@ -1101,3 +1106,52 @@ def unshard_with_sync(self):
             raise RuntimeError('CUDA out of memory encountered on a different rank')
         padded_unsharded_flat_param = self._all_gather_flat_param(unsharded_flat_param)
         self._use_unsharded_flat_param(padded_unsharded_flat_param)
+
+if version.parse(torch.__version__) >= version.parse('2.6.0') and version.parse(
```
dakinggg, line 1110: is this an exact copy of the patch above? should the range of that patch just be expanded to include 2.6?

dakinggg (commented):
lgtm, but won't approve until the pull request target is reverted back

composer/trainer/_patch_pytorch.py
rithwik-db: Oh yeah that's right, I just misread the code from earlier.

dakinggg (approved):

antoinebrl:
Thanks for adopting torch 2.6 🙏 ! Would this merge trigger a new release of Composer?

PR #3775, opened 2025-03-03
Mlflow rocm error

KuuCi:
# What does this PR do?

Ignores PytestUnraisableExceptionWarning that is caused by new Mlflow version, causing tests to break. Also adds another round of patches for Mlflow tests so that mlflow runs properly are ended before starting again. Finally, patching some directories since Mlflow is throwing errors stating `.trash` directory is not found.

# Before submitting

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,7 +186,7 @@ filterwarnings = [
     # Ignore _get_pg_default_device deprecation warning, it's used by torch internal, not composer
     '''ignore:.*`_get_pg_default_device` will be deprecated.*:UserWarning''',
     # Ignore ROCMMonitor warnings
-    '''ignore:Exception ignored in: <function ROCMMonitor.__del__:*''',
+    '''ignore:.*`Exception ignored in: <function ROCMMonitor.__del__:*''',
 ]
 
 # Coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,7 +186,7 @@ filterwarnings = [
     # Ignore _get_pg_default_device deprecation warning, it's used by torch internal, not composer
     '''ignore:.*`_get_pg_default_device` will be deprecated.*:UserWarning''',
     # Ignore ROCMMonitor warnings
-    '''ignore:.*ROCMMonitor.__del__''',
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -185,6 +185,8 @@ filterwarnings = [
     '''ignore:.*`clean_up_tokenization_spaces` was not set.*:FutureWarning''',
     # Ignore _get_pg_default_device deprecation warning, it's used by torch internal, not composer
     '''ignore:.*`_get_pg_default_device` will be deprecated.*:UserWarning''',
+    # Ignore ROCMMonitor warnings
+    '''ignore::pytest.PytestUnraisableExceptionWarning''',
```
dakinggg, line 189: can this be more scoped to the specific mlflow error here by string regex? or is there not more detail in the message?

tests/callbacks/test_callbacks.py
```diff
@@ -19,6 +19,18 @@
 from tests.common import EventCounterCallback
 
 
+@pytest.fixture(autouse=True)
```
dakinggg, line 22: I don't think this should be autoused. or if it is, it should case on mlflow not being installed since mlflow is an optional dep.

tests/callbacks/test_callbacks.py
```diff
@@ -162,23 +221,70 @@ def test_trains(self, cb_cls: type[Callback], device_train_microbatch_size: int,
...
+        Tests that training with multiple fits complete. Note: future functional tests should test for idempotency (e.g functionally)
         """
         del _remote  # unused. `_remote` must be passed through to parameterize the test markers.
         cb_kwargs = get_cb_kwargs(cb_cls)
         cb = cb_cls(**cb_kwargs)
 
         maybe_patch_context = get_cb_patches(cb_cls)
 
+        # For MLFlowLogger, add a patch for the file_utils.is_directory function
+        if cb_cls.__name__ == 'MLFlowLogger':
```
dakinggg, line 262: is it possible to centralize this logic (e.g. with `get_cb_patches`)` instead of copying it for each test?

pyproject.toml
KuuCi: There isn't anything more specific, ignoring ROCMMonitor won't work

PR #3784, opened 2025-03-19
Make it possible to define the root temporary directory for composer

mshtelma:
Make it possible to define the root temporary directory for composer

# What does this PR do?

Checkpoints can become very large, and sometimes the temporary file system doesn't have enough disk space to accommodate them. This PR introduces the ability to use the TMPDIR environment variable to specify the root temporary directory. Additionally, if TMPDIR is not set and the /local_disk0 directory exists, the system will default to using /local_disk0/temp as the temporary directory.

# Before submitting
- [x ] Have you read the [contributor guidelines](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md)?
- [x ] Did you run `pre-commit` on your change? (see the `pre-commit` section of [prerequisites](https://github.com/mosaicml/composer/blob/main/CONTRIBUTING.md#prerequisites))

### DIFF
--- a/composer/callbacks/checkpoint_saver.py
+++ b/composer/callbacks/checkpoint_saver.py
@@ -277,7 +277,7 @@ class CheckpointSaver(Callback):  # noqa: D101
             root_temp_folder = None
             if os.environ.get('TMPDIR') is not None:
                 root_temp_folder = os.environ.get('TMPDIR')
-            elif pathlib.Path('/local_disk0').exists():  # Probably we are on MLR, so we have to use /local_disk0
+            elif os.path.exists('/local_disk0/'):  # Probably we are on MLR, so we have to use /local_disk0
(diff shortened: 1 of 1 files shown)


composer/callbacks/checkpoint_saver.py
```diff
@@ -274,7 +274,13 @@ def __init__(
 
         is_remote_folder = backend != ''
         if is_remote_folder:  # If uploading to a remote path, use a temporary directory to save local checkpoints.
-            local_folder = os.path.join(tempfile.mkdtemp(), local_folder)
+            root_temp_folder = None
```
dakinggg, line 277: I believe this is already how mkdtemp works? `https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp`. The first place it should look to decide the tmpdir folder is whatever is specified by the `TMPDIR` env var.

PR #3792, opened 2025-03-26
Fix checkpoint loading in Pytorch 2.6.0 for ckpts exported before Pytorch 2.1.0

ethantang-db:
# What does this PR do?

This PR fixes a bug in Pytorch 2.6 where when we are loading checkpoints in a distributed manner that are created by Pytorch 2.1.0 or earlier, we would have an exception due to mismatching sizes of `rng.cuda`. This is because Pytorch now starts to check for the dimensions of the state_dict tensor against the dimensions that are in the checkpoint (see https://github.com/pytorch/pytorch/pull/141571).

This PR also updates our CI package in our tests which extends the timeout of submitting runs remotely for execution.

this fixes our daily tests. Example of passing: https://github.com/mosaicml/composer/actions/runs/14120989576

# Before submitting

### DIFF
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -574,7 +574,7 @@ def dist_cp_load(
                 # load a ckpt created before that version of Pytorch.
                 for idx in range(len(state_dict['rng'])):
                     state_dict['rng'][idx]['cuda'] = torch.zeros(816, dtype=torch.uint8)
-                
+
                 dist_cp.load(
                     state_dict=state_dict,
                     storage_reader=storage_reader,
--- a/composer/utils/checkpoint.py
+++ b/composer/utils/checkpoint.py
@@ -557,7 +557,6 @@ def dist_cp_load(
     storage_reader: StorageReader,
     load_planner: Optional[LoadPlanner] = None,
 ):
-    print(state_dict)
     if version.parse(torch.__version__) >= version.parse('2.4.0'):
         from torch.distributed.checkpoint.utils import CheckpointException
         try:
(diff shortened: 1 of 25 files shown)


composer/utils/checkpoint.py
```diff
@@ -566,6 +566,7 @@ def dist_cp_load(
                 planner=load_planner,
             )
         except CheckpointException as e:
+            raise e
```
dakinggg, line 569: is this intentional?

dakinggg (commented):
PR title says do not merge, is this actually ready for review?

composer/utils/checkpoint.py
```diff
@@ -566,6 +566,21 @@ def dist_cp_load(
                 planner=load_planner,
             )
         except CheckpointException as e:
+            if "torch.Size([816])" in str(e) and "rng" in str(e) and "cuda" in str(e):
+                # In torch 2.6, we are strictly enforcing the sizes of the state dict values vs
```
dakinggg, line 570: instead of "we", "PyTorch" (throughout)

composer/utils/checkpoint.py
```diff
@@ -566,6 +566,21 @@ def dist_cp_load(
                 planner=load_planner,
             )
         except CheckpointException as e:
+            if "torch.Size([816])" in str(e) and "rng" in str(e) and "cuda" in str(e):
```
dakinggg, line 569: I'd match on a bit more of the string. For reference the full error is `Size mismatch between saved torch.Size([816]) and current: torch.Size([16]) for rng.0.cuda`. like regex * torch.Size([816]) * torch.Size([16]) * rng.\d.cuda (not a valid regex, just getting the point across)

dakinggg (commented):
Please update the PR description with full explanation for both of the issues (helps when we forget why we did this and want to come look at the PR again)

composer/utils/checkpoint.py
```diff
@@ -566,6 +567,23 @@ def dist_cp_load(
                 planner=load_planner,
             )
         except CheckpointException as e:
+            if re.search(r'Size mismatch.+torch.Size([816]).+rng.+cuda', str(e)) is not None:
+                # In torch 2.6, Pytorch (throughout) is strictly enforcing the sizes of the
```
dakinggg, line 571: lol the "throughout" meant that you should change "we" to "PyTorch" in this whole comment. Not meant "throughout" as part of the comment

composer/utils/checkpoint.py
```diff
@@ -566,6 +567,23 @@ def dist_cp_load(
                 planner=load_planner,
             )
         except CheckpointException as e:
+            if re.search(r'Size mismatch.+torch.Size([816]).+rng.+cuda', str(e)) is not None:
```
dakinggg, line 570: don't think this regex is right. think this is `r'Size mismatch.+torch\.Size\(\[816\]\).+rng\.\d+\.cuda'`

dakinggg (approved):

PR #3822, opened 2025-04-16
Update optimizer params for fsdp2

rithwik-db:
# What does this PR do?

PR to update optimizer params to point to dtensors instead of regular model tensors. This should work fine for checkpointing as well since we fully shard before updating the optimizer's state (which is referenced in `update_optimizer_modules`)

### DIFF
--- a/composer/distributed/fsdp2.py
+++ b/composer/distributed/fsdp2.py
@@ -3,11 +3,12 @@
 
 """Helpers for FSDP2."""
 
+import warnings
+from typing import Optional
+
 import torch
 import torch.nn as nn
 from torch.distributed.fsdp._fully_shard import fully_shard
-from typing import Optional
-import warnings
 
 from composer.utils.parallelism import FSDP2Config
 
@@ -97,7 +98,7 @@ def legalize_param_sharing_between_modules(model: nn.Module, modules_to_shard: l
 def update_optimizer_modules(
     optimizer: torch.optim.Optimizer,
     model: nn.Module,
-    orig_param_id_to_name: dict[int, str],
+    orig_param_id_to_name: dict[torch.nn.Parameter, str],
 ) -> None:
     """Updates the optimizer's parameter groups to use the sharded model parameters.
     Assumes no training has occurred yet and the optimizer state is empty. If the optimizer state is not empty,
@@ -110,11 +111,11 @@ def update_optimizer_modules(
         orig_param_id_to_name (dict[int, str]): Mapping from original parameter IDs to their names.
     """
     # Check if the optimizer state is empty
-
+    # If not, clear it and warn the user
     if optimizer.state:
         warnings.warn(
-            "FSDP2 wrapping assumes the optimizer state is empty (i.e., training has not started). "
-            "but non-empty optimizer state was found. Optimizer state will be cleared."
+            'FSDP2 wrapping assumes the optimizer state is empty (i.e., training has not started). '
+            'but non-empty optimizer state was found. Optimizer state will be cleared.',
         )
         optimizer.state.clear()
 
@@ -131,7 +132,7 @@ def update_optimizer_modules(
             if param_name is not None and param_name in name_to_sharded_param:
                 old_to_new_param[param] = name_to_sharded_param[param_name]
             else:
-                raise ValueError(f"The same model must be passed to the optimizer and trainer.")
+                raise ValueError(f'The same model must be passed to the optimizer and trainer.')
 
     # Update param groups with new parameters
     new_param_groups = []
--- a/composer/distributed/fsdp2.py
+++ b/composer/distributed/fsdp2.py
@@ -98,9 +98,10 @@ def legalize_param_sharing_between_modules(model: nn.Module, modules_to_shard: l
 def update_optimizer_modules(
     optimizer: torch.optim.Optimizer,
     model: nn.Module,
-    orig_param_id_to_name: dict[torch.nn.Parameter, str],
+    orig_param_to_name: dict[torch.nn.Parameter, str],
 ) -> None:
     """Updates the optimizer's parameter groups to use the sharded model parameters.
+
     Assumes no training has occurred yet and the optimizer state is empty. If the optimizer state is not empty,
     it will be cleared with a warning.
 
@@ -108,7 +109,7 @@ def update_optimizer_modules(
         optimizer (Optimizer): The optimizer to update.
         modules_to_shard (list[nn.Module]): The modules that will be sharded.
         model (nn.Module): The parent model that is also sharded.
-        orig_param_id_to_name (dict[int, str]): Mapping from original parameter IDs to their names.
+        orig_param_to_name (dict[torch.nn.Parameter, str]): Mapping from original parameters to their names.
     """
     # Check if the optimizer state is empty
     # If not, clear it and warn the user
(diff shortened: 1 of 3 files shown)


composer/distributed/fsdp2.py
```diff
@@ -190,4 +245,14 @@ def prepare_fully_shard(
         None
     """
     modules_to_shard, _ = get_standalone_and_tied_modules(list(model.children()))
+
+    # Build the parameter ID to name mapping (with no duplicates)
+    orig_param_id_to_name = {id(param): name for name, param in model.named_parameters()}
+    for module in modules_to_shard:
+        orig_param_id_to_name.update({id(param): name for name, param in module.named_parameters()})
```
bowenyang008, line 252: nit: is there any reason to use id(param), I think direct hash of param should work

composer/distributed/fsdp2.py
```diff
@@ -119,6 +121,58 @@ def _check_param_sharing(module: nn.Module):
...
+        modules_to_shard (list[nn.Module]): The modules that will be sharded.
+        model (nn.Module): The parent model that is also sharded.
+        orig_param_id_to_name (dict[int, str]): Mapping from original parameter IDs to their names.
+    """
+    # Using the same logic as in FSDP1 to address multiple optimizers
+    optimizers_tuple = ensure_tuple(optimizers)
+    if len(optimizers_tuple) != 1:
+        raise NotImplementedError(f'Only one optimizer is supported; found {len(optimizers_tuple)} optimizers')
+
+    optimizer = optimizers_tuple[0]
```
bowenyang008, line 144: if FSDP1 only takes a single optimizer, I think we can start with a simple interface with just Optimizer instead of sequence of optimizers. We can extend this interface as we see fit.

composer/distributed/fsdp2.py
```diff
@@ -119,6 +121,58 @@ def _check_param_sharing(module: nn.Module):
...
+
+    # Build a mapping from parameter name to sharded parameter (after sharding)
+    name_to_sharded_param = dict(model.named_parameters())
+    for module in modules_to_shard:
+        name_to_sharded_param.update(dict(module.named_parameters()))
```
bowenyang008, line 149: can we simplify it by just taking the root module? module.named_paramters(recurse=True) should just give you all the params so you don't have to iterate over sharded modules again

composer/distributed/fsdp2.py
```diff
@@ -119,6 +121,58 @@ def _check_param_sharing(module: nn.Module):
...
     _check_param_sharing(model)
 
+def update_optimizer_modules(
+    optimizers: Optional[Union[torch.optim.Optimizer, Sequence[torch.optim.Optimizer]]],
+    modules_to_shard: list[nn.Module],
+    model: nn.Module,
+    orig_param_id_to_name: dict[int, str],
+) -> None:
+    """Updates the optimizer's parameter groups to use the sharded model parameters.
+    Assumes no training has occurred yet and the optimizer state is empty.
```
bowenyang008, line 131: can we legalize against this?

composer/distributed/fsdp2.py
dakinggg: the multi optimizers is something that someone put in composer a long time ago, but composer mostly does not actually support. we should drop it from everything, but haven't  bothered with the deprecations. feel free to not support multiple optimizers here.

composer/distributed/fsdp2.py
rithwik-db: I was taking the same route as we were with FSDP1 (which uses id(...)), either should work fine I believe.

tests/trainer/test_fsdp2.py
```diff
@@ -52,7 +52,8 @@ def test_fsdp2_initialization_with_tied_params(
         mp_policy=None,
         offload_policy=None,
     )
-    prepare_fully_shard(model=model.module, fsdp2_config=fsdp2_config)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
+    prepare_fully_shard(model=model.module, optimizers=optimizer, fsdp2_config=fsdp2_config)
```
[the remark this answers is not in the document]
rithwik-db: Yeah that's fair, I can remove this change

tests/trainer/test_fsdp2.py
```diff
@@ -86,3 +87,82 @@ def test_fsdp2_initialization_with_tied_params(
...
     assert (model.mlp.fc1.weight is model.mlp.fc2.weight)
     assert (torch.equal(weight_1, weight_2))
+
+
+@pytest.mark.parametrize('case', ['all_params_one_group', 'subset_one_group', 'multiple_groups'])
+@pytest.mark.parametrize('device', _INIT_DEVICES)
+@world_size(2)
+@pytest.mark.gpu
+@pytest.mark.filterwarnings('ignore:FSDP2 Config/APIs are experimental*:UserWarning')
+@pytest.mark.skipif(SKIP_TEST, reason='FSDP2 is not available in torch < 2.6.0')
```
dakinggg, line 97: use the test helper that bowen just merged?

composer/distributed/fsdp2.py
bowenyang008: IIUC python basically hash based on id unless Parameter has its own hasher and it does not work for our case

tests/trainer/test_fsdp2.py
```diff
@@ -86,3 +87,82 @@ def test_fsdp2_initialization_with_tied_params(
...
+    def validate_optimizer_state(current_optimizer: torch.optim.Optimizer, stage: str):
+        assert len(current_optimizer.param_groups) == len(optimizer_input), \
+            f"[{case}/{stage}] Group count mismatch. Expected {len(optimizer_input)}, Got {len(current_optimizer.param_groups)}"
+        for i, group in enumerate(current_optimizer.param_groups):
+            opt_params = group['params']
+            assert len(opt_params) == len(optimizer_input[i]['params']), \
+                f"[{case}/{stage}] Group {i}: Param count mismatch. Expected {len(optimizer_input[i]['params'])}, Got {len(opt_params)}"
+            assert all(isinstance(p, DTensor) for p in opt_params), \
+                f"[{case}/{stage}] Group {i}: Not all parameters are DTensors"
+            assert group['lr'] == optimizer_input[i]['lr'], \
```
dakinggg, line 151: should probably check that all keys match? lr might be the only key, but the code has the intention os preserving all so we should check all i think

composer/distributed/fsdp2.py
```diff
@@ -190,4 +245,14 @@ def prepare_fully_shard(
         None
     """
     modules_to_shard, _ = get_standalone_and_tied_modules(list(model.children()))
+
+    # Build the parameter ID to name mapping (with no duplicates)
+    orig_param_id_to_name = {id(param): name for name, param in model.named_parameters()}
+    for module in modules_to_shard:
```
dakinggg, line 251: I think you can just make the original call with `recurse=True` and avoid the extra for loop?

dakinggg: oh bowen left this comment somewhere else too

composer/distributed/fsdp2.py
```diff
@@ -91,6 +95,59 @@ def _check_param_sharing(module: nn.Module):
...
+    # Note: if params are tied and the same parameter is in multiple groups, pytorch will raise an error
+    old_to_new_param = {}
+    for group in optimizer.param_groups:
+        for param in group['params']:
+            param_name = orig_param_id_to_name.get(param, None)
+            # Note: the names of the parameters stay the same after sharding so we can do the following.
+            if param_name is not None and param_name in name_to_sharded_param:
+                old_to_new_param[param] = name_to_sharded_param[param_name]
+            else:
+                raise ValueError(f'The same model must be passed to the optimizer and trainer.')
```
bowenyang008, line 135: nit: can we print the offensive param_name for debugging purposes?

bowenyang008 (approved):
LGTM, thanks @rithwik-db!

composer/distributed/fsdp2.py
bowenyang008: Oh, I simply meant `ValueError(f'The same model must be passed to the optimizer and trainer. Found an offensive param: {param_name}')
` But your updated solution is definitely more verbose/clear than my initial suggest

composer/distributed/fsdp2.py
```diff
@@ -91,6 +95,78 @@ def _check_param_sharing(module: nn.Module):
...
+            else:
+                old_to_new_param[param] = name_to_sharded_param[param_name]
+
+    # Raise an error with all the parameters that were not found in the sharded model
+    if len(unseen_params) > 0:
+        raise ValueError(
+            f'The same model must be passed to the optimizer and trainer but the '
+            f'following parameters were not found in the sharded model: {list(unseen_params)}.'
+            'All parameters prefixed with "optimizer.param_id" imply that the optimizer has the wrong model.'
+            'All parameters prefixed with "model.param_name" imply a significant issue where sharding '
```
bowenyang008, line 152: are the params prefixed with something like "optimizer.param_id(name)" if the params don't belong to this model like the example test you wrote? Is torch naming them to something prefixed with "optimizer"?

composer/distributed/fsdp2.py
rithwik-db: We would only have `param_name` set to a value if our sharding algorithm is broken, when there are optimizer issues `param_name` is just `None`, so this won't really provide much insight to customers.

composer/distributed/fsdp2.py
rithwik-db: No I am naming it myself, I don't really have a better way of naming them since we could either return the full parameter that's causing the issue (which prints out something pretty ugly for users), or we could return the id of the parameter, which is more readable relative to the rest of the error message.

composer/distributed/fsdp2.py
```diff
@@ -91,6 +95,78 @@ def _check_param_sharing(module: nn.Module):
...
+    orig_param_to_name: dict[torch.nn.Parameter, str],
+) -> None:
+    """Updates the optimizer's parameter groups to use the sharded model parameters.
+
+    Assumes no training has occurred yet and the optimizer state is empty. If the optimizer state is not empty,
+    it will be cleared with a warning.
+
+    Args:
+        optimizer (Optimizer): The optimizer to update.
+        modules_to_shard (list[nn.Module]): The modules that will be sharded.
```
dakinggg, line 110: old docstring?

dakinggg (approved):
LGTM

sashaDoubov:
LGTM as well.

dakinggg:
Just FYI, I'm pretty sure the new code does not have this issue, but one quirk of the FSDP1 impl: https://github.com/mosaicml/composer/pull/2320

PR #3826, opened 2025-04-17
Support submodule wrapping for FSDP2 according to model definition (with `_fsdp_wrap` and `fsdp_wrap_fn`)

rithwik-db:
# What does this PR do?

Modified existing code to support `_fsdp_wrap` and `_fsdp_wrap_fn` similar to how FSDP1 handles it (with the slight caveat that we allow `None` to be a valid input for skipping the current module and checking descendants).

Reworked module wrapping so that we do a recursive check where we pre-order check the legalization and validation of params to make sure nothing that shouldn't be weight tied is actually tied and then we do a post-order fully_shard on valid parameters.


composer/distributed/fsdp2.py
```diff
@@ -145,6 +145,49 @@ def apply_fully_shard(
...
+    """
+    modules_to_wrap = []
+    visited_modules = set()
+
+    def find_modules_to_wrap_recursive(module: nn.Module):
+        if module in visited_modules:
+            return
+        visited_modules.add(module)
+
+        fsdp_wrap_setting = getattr(module, '_fsdp_wrap', None)
```
dakinggg, line 172: The FSDP1 support is a bit more complicated than just the _fsdp_wrap attr. See https://github.com/mosaicml/composer/blob/03ac462c55040c56aca4604cc4ba92bb02543b94/composer/distributed/dist_strategy.py#L456-L464

dakinggg: btw here is pytorch's post order traversal from fsdp1: https://github.com/pytorch/pytorch/blob/24262587893517028b169e72e81e8e1c01585249/torch/distributed/fsdp/wrap.py#L31

bowenyang008: Re: post order. We need a similar post order to apply fully_shard, but also need a pre order check to legalize there are no weight tying since once we apply fully_shard, we lost the info if two weights are tied between two submodules.

composer/distributed/fsdp2.py
```diff
@@ -3,12 +3,109 @@
...
+            raise ValueError(
+                f'Detected tied parameters between modules designated for FSDP wrapping within {type(module).__name__}. '
+                f'FSDP cannot wrap modules with tied parameters independently at the same level. '
+                f'Please adjust the auto_wrap_policy to ensure no parameter sharing exists between modules to be sharded.',
+            )
+
+        # Check for tying between candidates and the rest of the model (using root_module)
+        legalize_param_sharing_between_modules(root_module, standalone_child_candidates)
+
+    # 3. Legalize all submodules for the modules where _fsdp_wrap is False (none of the descendants should be candidates for sharding)
```
dakinggg, line 93: I don't think this is how this works. FSDP1 would apply the auto wrap policy check to all modules (https://github.com/pytorch/pytorch/blob/aa285e6512bae860981ffeb2547b85946fde0027/torch/distributed/fsdp/wrap.py#L247), regardless of what happens in the rest of the tree. so for us, any module that has ._fsdp_wrap==True or fsdp_wrap_fn(module)==True/dict, would get put into its own FSDP block (possibly with children that are also in their own FSDP block. lmk if you think I'm misunderstanding, or happy to talk it through live.

composer/distributed/fsdp2.py
```diff
@@ -3,12 +3,109 @@
...
 
+from typing import Callable, Optional, Union
+
 import torch.nn as nn
 from torch.distributed.fsdp._fully_shard import fully_shard
 
 from composer.utils.parallelism import FSDP2Config
 
 
+def _generate_default_policy(parent_model: nn.Module) -> Callable[[nn.Module], Union[bool, dict, None]]:
```
dakinggg, line 14: should we use the `CustomPolicy` object from pytorch that is used for FSDP1?

composer/distributed/fsdp2.py
```diff
@@ -3,12 +3,109 @@
...
+            candidate_kwargs[child] = policy_result
+    return candidates, candidate_kwargs
+
+
+def legalize_fsdp_wrap_policy(
+    module: nn.Module,
+    auto_wrap_policy: Callable[[nn.Module], Union[bool, dict, None]],
+) -> None:
+    """Legalizes the FSDP wrap policy by ensuring that no submodule has _fsdp_wrap set to True if if the ancestor has _fsdp_wrap set to False."""
+    assert auto_wrap_policy(module) is False, 'The root module must not be wrapped'
```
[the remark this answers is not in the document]
rithwik-db: Oops this was a miscommunication, we do wrap the root module, we just don't wrap this specific module since `_fsdp_wrap` was set to `False` for this, let me remove the error message

composer/distributed/fsdp2.py
```diff
@@ -3,12 +3,109 @@
...
+    for submodule in module.modules():
+        submodule_policy = auto_wrap_policy(submodule)
+        if submodule_policy is True or isinstance(submodule_policy, dict):
+            raise ValueError(
```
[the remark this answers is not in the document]
rithwik-db: I would agree, but I tested this out [here](https://dbc-559ffd80-2bfc.cloud.databricks.com/editor/notebooks/2455365426683650?o=7395834863327820#command/2455365426683967) and if I set `_fsdp_wrap` of the ancestor to `False`, it doesn't check the descendants to see if `_fsdp_wrap` is set to `True`

composer/distributed/fsdp2.py
dakinggg: Ah I forgot this existed, but this is explicitly true for the first level (https://github.com/mosaicml/composer/blob/635d9a73f870d3dafea328a36689cfcf5a2e9c3e/composer/distributed/dist_strategy.py#L397) within the `ComposerModel`, but not true recursively. Will revisit my review in light of this.

composer/distributed/fsdp2.py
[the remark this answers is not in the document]
dakinggg: Ah I forgot this existed, but this is explicitly true for the first level (https://github.com/mosaicml/composer/blob/635d9a73f870d3dafea328a36689cfcf5a2e9c3e/composer/distributed/dist_strategy.py#L397) within the `ComposerModel`, but not true recursively. Will revisit my review in light of this.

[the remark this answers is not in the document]
rithwik-db: Ah yeah just tried this without considering the top level module and it works as you mentioned ([here](https://dbc-559ffd80-2bfc.cloud.databricks.com/editor/notebooks/2455365426683650?o=7395834863327820#command/2455365426717451)), will update the code accordingly.

composer/distributed/fsdp2.py
```diff
@@ -168,79 +279,64 @@ def update_optimizer_modules(
...
-    fully_shard(independent_submodules, **fully_shard_kwargs)
-    # Apply fully_shard to the parent model to ensure all parameters are sharded
-    fully_shard(model, **fully_shard_kwargs)
+    # We default to wrapping the root module as long as its wrapping policy is not explicitly set to False
+    if root_policy_decision is True or root_policy_decision is None:
+        fully_shard(model, **fully_shard_kwargs)
+    elif isinstance(root_policy_decision, dict):
+        kwargs = root_policy_decision
+        fully_shard(model, **kwargs)
+    else:
```
dakinggg, line 312: should the else be an error? if not, can just delete, don't need an else that just has a pass

composer/distributed/fsdp2.py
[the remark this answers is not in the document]
dakinggg: ok sg, lmk when ready for another look

composer/distributed/fsdp2.py
rithwik-db: Updated this to follow the same mechanics as `FSDP1` and added a comment about `dist_strategy.py:L367`

composer/distributed/fsdp2.py
rithwik-db: Not sure I follow...  Is the premise just to support user-provided `CustomPolicy` to `prepare_fully_shard` and then just take the lambda function inside and treat that as the `auto_wrap_policy`? I assumed that we wouldn't need to go about this since we define the policy ourselves within `dist_strategy.py`, right?

composer/distributed/fsdp2.py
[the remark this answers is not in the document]
rithwik-db: Updated the code, thanks for catching this!

composer/distributed/fsdp2.py
bowenyang008: FSDP1 actually supports both Union[Policy, Callable], and Policy has a relative complicated path down in torch. Though we used Policy in FSDP1, the logic seems simple enough to be replaced by a Callable. So to avoid rebuild FSDP1 with FSDP2 APIs, my suggest to @rithwik-db was:
1. let's start with something simpler to get things rolling first
2. If we found Policy based useful, we can always add it back
3. there is also a chance torch will add back autowrap support, so we can keep our interface simple first for easier future upgrade/swap

dakinggg: sounds good! was only suggesting it if it simplified our implementation - if not, no need

composer/distributed/fsdp2.py
```diff
@@ -168,79 +256,60 @@ def update_optimizer_modules(
...
+        auto_wrap_policy (Callable[[nn.Module], Union[bool, Dict, None]]): The policy to apply to the model.
 
     Returns:
         None
     """
-    fully_shard_kwargs = {'mesh': fsdp2_config.device_mesh, 'reshard_after_forward': fsdp2_config.reshard_after_forward}
+    # Skip wrapping the model if the top level has _fsdp_wrap set to False
+    # Following the same approach as dist_strategy.py:L397 (FSDP1 implementation)
+    if hasattr(model, '_fsdp_wrap') and not bool(model._fsdp_wrap):
+        return
```
bowenyang008, line 277: I think we would like to isolate `_fsdp_wrap` from this function, so the logic purely depends on auto_wrap_policy

rithwik-db: I am not opposed, I just took the same route as the FSPD1 implementation, not sure what the original reasoning was for the route that was taken in `dist_strategy.py:L397`... cc: @dakinggg any context that you remember about this?

composer/distributed/fsdp2.py
```diff
@@ -4,15 +4,77 @@
...
+        standalone_child_candidates, tied_children = get_standalone_and_tied_modules(child_candidates)
+        if tied_children:
+            raise ValueError(
+                f'Detected tied parameters between modules designated for FSDP wrapping within {type(module).__name__}. '
+                f'FSDP cannot wrap modules with tied parameters independently at the same level. '
+                f'Please adjust the auto_wrap_policy to ensure no parameter sharing exists between modules to be sharded.',
+            )
+
+        # Check for tying between candidates and the rest of the model (using root_module)
+        legalize_param_sharing_between_modules(root_module, standalone_child_candidates)
```
bowenyang008, line 65: can you leave a comment as we discussed yesterday, like:
"NOTE: right now we don't allow weight sharing between fsdp and non-fsdp submodules, even if the parent module is not FSDP either. we may consider relax this when we see fit"

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+
+    fsdp2_config = FSDP2Config()
+
+    # M2 and M5 are candidates for sharding with no weight tying so this should work fine
+    m1 = DeepNestedModel()
+    m1.m2._fsdp_wrap = True  # type: ignore
+    m1.m5._fsdp_wrap = True  # type: ignore
+    opt = torch.optim.Adam(m1.parameters(), lr=0.01)
+    prepare_fully_shard(m1, opt, fsdp2_config)
+    assert isinstance(m1.m2.m3.weight, DTensor), 'm1.m2.m3.weight should be a DTensor'  # type: ignore
```
dakinggg, line 390: should check m4 and m7 here too?

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+    assert str(e.value).startswith('Parameter sharing'), str(e.value)
+
+    # Testing M1 has _fsdp_wrap set to True and M3 and M6 have tied weights
+    # This shouldn't return an error and all tensors are expected to be DTensors
+    m1 = DeepNestedModel()
+    m1._fsdp_wrap = True  # type: ignore
+    m1.m2.m3.weight = m1.m5.m6.weight  # type: ignore
+    opt = torch.optim.Adam(m1.parameters(), lr=0.01)
+    prepare_fully_shard(m1, opt, fsdp2_config)
+    assert isinstance(m1.m2.m3.weight, DTensor), 'm1.m2.m3.weight should be a DTensor'  # type: ignore
```
dakinggg, line 448: just loop over all weights and check they are dtensor?

composer/distributed/fsdp2.py
```diff
@@ -168,79 +246,71 @@ def update_optimizer_modules(
...
+
+    # Apply fully_shard to the model itself if it's set to be wrapped
+    # NOTE: There is a slight caveat with this approach. View the following example:
+    # M0 (_fsdp_wrap = False)
+    # ├── M1 (_fsdp_wrap not set)
+    # └── M2 (_fsdp_wrap = True)
+    # and M1 and M2 have tied weights.
+    # In this situation, we will raise an error even though this is perfectly valid and duplicate wrapping
+    # is not done.
+    # TODO: Address this caveat if it does become an issue for customers.
```
bowenyang008, line 285: oh, didn't see you have it here, then lgtm

bowenyang008: maybe worth to move it into _recursive_apply_fully_shard where the error out logic resides though

composer/distributed/fsdp2.py
```diff
@@ -64,6 +126,22 @@ def legalize_param_sharing_between_modules(model: nn.Module, modules_to_shard: l
...
+    #  self.child1
+    #  self.child2
+    #    ├── self.child1
+    #    └── grandchild
+    # We can fully_shard self.child2 however if we call fully_shard on `model` again, then
+    # due to `child1` is not a FSDPModule, it will try to fully_shard params in `child1` which is already sharded by `child2`,
+    # and it errors out with misleading error as it thinks it is applying FSDP on top of another parallelism.
+
+    # Currently identify_shardable_modules avoids this case through generally handling weight tying so that only parent of child1 and child2
+    # is sharded. However if we allow users to call this function directly with custom modules_to_shard, we need to:
```
bowenyang008, line 141: we can remove these two lines as they are not relevant anymore

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+    m1 = DeepNestedModel()
+    m1.m2._fsdp_wrap = True  # type: ignore
+    m1.m2.weight = m1.m2.m3.weight  # type: ignore
+    m1.m2.m3.weight = m1.m5.m6.weight  # type: ignore
+    opt = torch.optim.Adam(m1.parameters(), lr=0.01)
+    with pytest.raises(ValueError) as e:
+        prepare_fully_shard(m1, opt, fsdp2_config)
+    assert str(e.value).startswith('Parameter sharing'), str(e.value)
+
+    # NOTE!! EDGE CASE THAT ERRORS OUT WHEN IT SHOULD NOT:
```
dakinggg, line 512: can you explain why this _should_ be allowed? I would think that since m6 is in its own fsdp block, it is not allowed to share weights with m3, which is outside of its fsdp block

composer/distributed/fsdp2.py
```diff
@@ -4,15 +4,77 @@
...
+            ret = bool(current_module._fsdp_wrap)
+        elif hasattr(parent_model, 'fsdp_wrap_fn') and isinstance(parent_model.fsdp_wrap_fn, Callable):
+            # There are certain situations where _fsdp_wrap for the parent model is not set, but we wrap submodules
+            # with _fsdp_wrap_fn (e.g. wrapping all GPTBlocks). In those situations, we generally also want to wrap
+            # the parent model, so we have an additional check here.
+            if current_module == parent_model:
+                return True
+            ret = parent_model.fsdp_wrap_fn(current_module)
+            if isinstance(ret, dict):
+                assert set(ret.keys()) == {'mesh', 'reshard_after_forward'} \
```
dakinggg, line 31: Can we make this so you don't have to specify all args? Like it should inherit the default config and this only applies anything in `ret` as an override

tests/common/models.py
```diff
@@ -141,6 +141,10 @@ def __init__(self, num_features: int, device: Union[str, torch.device], num_clas
         net = torch.nn.Sequential(fc1, torch.nn.ReLU(), fc2)
         super().__init__(num_classes=num_classes, module=net)
 
+    def add_fsdp_wrap_attribute_to_children(self):
+        for child in self.module.children():
+            child._fsdp_wrap = True  # type: ignore
```
bowenyang008, line 146: do we have some test also test: fsdp_wrap_fn? actually this might be a question for @dakinggg, if most of code base just rely on fsdp_wrap_fn, should we just remove support of using _fsdp_wrap?

composer/distributed/fsdp2.py
dakinggg: also seems like we still need to add a test for using fsdp_wrap_fn and returning kwargs?

tests/common/models.py
dakinggg: we use both in LLM Foundry (although of course could switch to only using one or the other)

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+        setattr(self, f'm{parent_num+2}', nn.Linear(10, 10))
+
+
+class DeepNestedModel(nn.Module):
+    """A model with a deep nested structure."""
+
+    def __init__(self):
+        super().__init__()
+        self.m2 = NestedModule(parent_num=2)
+        self.m5 = NestedModule(parent_num=5)
```
bowenyang008, line 363: since parent_num is hardcoded, why not just set via self.xxx = ... instead of setattr?

tests/trainer/test_fsdp2.py
bowenyang008: so maybe we should have a generic legalizer in fsdp2, regardless how we wrap, the params that are shared before FSDP must be shared the same way after FSDP?

composer/distributed/fsdp2.py
rithwik-db: As a sidenote, we did end up going with `Policy` since it does make things a bit simpler

composer/distributed/fsdp2.py
```diff
@@ -4,15 +4,77 @@
...
+
+    # 3. Recurse on module's children for downstream sharding
+    for child in module.children():
+        _recursive_apply_fully_shard(root_module, child, target_modules_to_kwargs)
+
+    # 4. Apply fully_shard to the standalone children identified earlier (Post-order application)
+    # Note that all modules that we fully_shard will have kwargs in target_modules_to_kwargs
+    for child in standalone_child_candidates:
+        kwargs = target_modules_to_kwargs[child]
+        fully_shard(child, **kwargs)
```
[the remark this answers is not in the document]
rithwik-db: Can do

composer/distributed/fsdp2.py
rithwik-db: Yeah it's better there

tests/trainer/test_fsdp2.py
rithwik-db: Yeah I can use a conditional to make the intellisense work better.

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+    assert isinstance(m1.m5.m6.weight, DTensor), 'm1.m5.m6.weight should be a DTensor'  # type: ignore
+
+    error_msg = 'm1.m2.m3.weight and m1.m5.m6.weight should be the same object'
+    assert id(m1.m2.m3.weight) == id(m1.m5.m6.weight), error_msg  # type: ignore
+
+    # Testing the case when submodule has `_fsdp_wrap` set to True but ancestor has `_fsdp_wrap` set to False
+    # This should wrap the model fine, we have to make sure that m1.fsdp_wrap is set to False to not allow
+    # for general FSDP wrapping to happen.
+    m1 = DeepNestedModel()
+    m1._fsdp_wrap = False  # type: ignore
```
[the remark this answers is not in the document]
rithwik-db: Yeah this is what FSDP1 does, but @bowenyang008 mentioned that we should go for a policy that we still recurse even if the root module has this attribute set to `false` and it seems to make sense in situations where the user would not want all tensors to be dtensors but a mix of tensors and dtensors (in those situations, we can't `fully_shard(root_module)`. To address this, we added the code:
```
elif current_module == parent_model:
            # Unless the user specifically sets the _fsdp_wrap attribute to False for the parent model,
            # we default to wrapping the parent model.
            ret = True
```
So unless the user specifically sets the attribute `_fsdp_wrap=False` on the root module, we still end up wrapping the root model at the end.

tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +343,181 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+    # when it can't because it shares weights with M2 (which also requests to be wrapped)
+    m1 = DeepNestedModel()
+    m1.m2.m3._fsdp_wrap = True  # type: ignore
+    m1.m2.weight = m1.m2.m3.weight  # type: ignore
+    opt = torch.optim.Adam(m1.parameters(), lr=0.01)
+    with pytest.raises(ValueError) as e:
+        prepare_fully_shard(m1, opt, fsdp2_config)
+    assert str(e.value).startswith('Parameter sharing'), str(e.value)
+
+    # Testing parent (M2) sharing weights with child (M3) and other branch (M5) - this should work since parent has _fsdp_wrap=True
```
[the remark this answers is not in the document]
rithwik-db: Yep this is a comment issue, this correctly raises an error

tests/trainer/test_fsdp2.py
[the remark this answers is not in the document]
bowenyang008: yup, so if the root module and submodules all have _fsdp_wrap = False, the behavior should be the same, yet if root module has _fsdp_wrap = True but submodules has _fsdp_wrap = False (which should have never occurred)  we can respect it with FSDP2 wrapper

tests/trainer/test_fsdp2.py
rithwik-db: A couple of things:
1. Oops, forgot to add `m1._fsdp_wrap = False`
2. @bowenyang008 can you explain the logic from wednesday again please? Let's say theres:

M1
- fully_shard(M2) - {W0, W1}
- M3 - {W1, W2}

Where W1 is tied between M2 and M3 and M2 is fully sharded. It makes sense why a user would do something like this, for instance if W0 needs to be sharded and W2 is small enough to not be sharded. But in this situation, this should still be an error right? Since If we do something like this, W1 wouldn't be tied anymore and would cause errors. I think I'm missing something from our discussion...

bowenyang008: correct, that's why I said I was wrong on Weds. So this case should error out, not SHOULD NOT. And to avoid getting lost in these corner cases, I suggest build a generic util to check the fully qualified names (FQN) of params should maintain their tying relation pre and post wrapping, e.g.,
```
with check_param_typing(model):
    apply_fully_shard(model)
```

composer/distributed/fsdp2_utils.py
```diff
@@ -0,0 +1,271 @@
...
+
+def _get_param_tying_groups(model: nn.Module) -> list[set[str]]:
+    """Identifies groups of tied parameters within a model based on object identity.
+
+    A parameter is considered tied if the same nn.Parameter object appears multiple
+    times when iterating through model.named_parameters().
+
+    NOTE: We take the recursive approach since .named_parameters() has a weird behavior where if you do
+    m1.m2.m3.weight = m1.m2.m4.weight and then call m1.named_parameters(), it will only return the FQN for m1.m2.m3.weight
+    but not m1.m2.m4.weight.
```
bowenyang008, line 122: this is expected as torch only returns params w/o duplication, so maybe we can simply compare set(m.named_parameters()) is the same before and after the context?

bowenyang008: your current implementation is also valid, so either way is fine

composer/distributed/fsdp2_utils.py
```diff
@@ -0,0 +1,271 @@
...
+
+import torch
+import torch.nn as nn
+from torch.distributed.fsdp.wrap import CustomPolicy
+
+from composer.utils.parallelism import FSDP2Config
+
+# FSDP2 Weight Tying Functions
+# TODO: These functions are all relatively similar to each other, we should consider
+# refactoring them in the future to be simpler.
```
bowenyang008, line 18: agree, maybe also worth putting the weight tying related ones into a new file as well in a following up PR. We may need a dir for FSDP utils instead

composer/distributed/fsdp2_utils.py
```diff
@@ -0,0 +1,271 @@
...
+        # Recursively process child modules
+        for child_name, child in module.named_children():
+            child_prefix = f'{prefix}.{child_name}' if prefix else child_name
+            _recursive_get_params(child, child_prefix)
+
+    _recursive_get_params(model)
+
+    # Filter to keep only groups where the same parameter object has multiple FQNs
+    tying_groups = [fqns for fqns in param_object_to_fqns.values() if len(fqns) > 1]
+    return tying_groups
```
bowenyang008, line 144: we can probably just compare all groups, e.g., we can also avoid the case the case one param was not tied before but became tied afterwards for some reason

tests/trainer/test_fsdp2_wrap_fsdp_attribute.py
```diff
@@ -0,0 +1,365 @@
...
+
+    m1.fsdp_wrap_fn = wrap_fn  # type: ignore
+    opt = torch.optim.Adam(m1.parameters(), lr=0.01)
+    prepare_fully_shard(m1, opt, fsdp2_config)
+
+    def check_reshard_after_forward(module: nn.Module):
+        fsdp_state = module._get_fsdp_state()  # type: ignore
+        param_group = fsdp_state._fsdp_param_group
+        assert param_group.post_forward_mesh_info is None, \
+            f'reshard_after_forward should be False, but got {param_group.post_forward_mesh_info}'
```
bowenyang008, line 329: nice test, though I am just wondering if we can directly check if forward_hooks are empty (assuming they are empty if reshard_after_forward is False) so we don't depend on these internal states prefixed with _xxx so subject to change

tests/trainer/test_fsdp2_wrap_fsdp_attribute.py
```diff
@@ -0,0 +1,365 @@
...
+
+@fsdp2_context
+@world_size(2)
+def test_fsdp_wrap_fn_invalid_keys(world_size: int):
+    """Test that an error is raised if the wrap function returns a dict with invalid keys."""
+    fsdp2_config = FSDP2Config()
+    m1 = DeepNestedModel()
+
+    def wrap_fn(module: nn.Module):
+        return {'tacos': False}
```
bowenyang008, line 266: ❤️ 🌮

tests/common/models.py
bowenyang008: @rithwik-db I just discussed with @dakinggg, we can keep supporting both for the moment, but let's raise a Deprecation Warning for use of _fsdp_wrap in generate_default_policy, so we will only support fsdp_wrap_fn in the future.

bowenyang008 (approved):
Amazing work, thanks @rithwik-db!

dakinggg (approved):

composer/distributed/fsdp2_utils.py
rithwik-db: Added a note

tests/trainer/test_fsdp2_wrap_fsdp_attribute.py
rithwik-db: Seems like `module._forward_hooks` is `1` regardless of whether `reshard_after_forward` is `True` or `Not` as the existing `_post_foward` hook does more than just handle resharding, so the current approach might be a more targetted solution

bowenyang008: ic, sg, maybe we can create an additional GPU test that runs fwd and check params are still full_tensor before backprop, in a followup PR

composer/distributed/fsdp2_utils.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+    attribute_warning_raised = False
+
+    def lambda_fn(current_module: nn.Module) -> Union[bool, dict[str, Any]]:
+        nonlocal attribute_warning_raised
+
+        ret = False
+        if hasattr(current_module, '_fsdp_wrap'):
+            if not attribute_warning_raised:
+                warnings.warn(
```
bowenyang008, line 263: you can probably use: warnings.filterwarnings('once', category= DeprecationWarning) instead of capture it yourself

rithwik-db: Thanks! Never knew this

PR #3832, opened 2025-04-25
Activation Checkpointing and Offloading for FSDP2

rithwik-db:
# What does this PR do?

Adding activation checkpointing and activation cpu offloading using the same method as TorchTitan in their examples (which aligns with our previous approach using `checkpoint_wrapper`)

Added tests to check if checkpointing and cpu offloading is working as expected.

### DIFF
--- a/composer/distributed/activation_checkpointing.py
+++ b/composer/distributed/activation_checkpointing.py
@@ -1,0 +1,60 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Helpers for activation checkpointing. Note that while this is orthogonal to FSDP2, it is implemented in the distributed directory because it is closely related to FSDP2."""
+
+from typing import Callable
+
+import torch
+import torch.nn as nn
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    CheckpointImpl,
+    apply_activation_checkpointing,
+    checkpoint_wrapper,
+    offload_wrapper,
+)
+
+from composer.utils.parallelism import FSDP2Config
+
+
+def apply_ac(model: nn.Module, fsdp2_config: FSDP2Config) -> None:
+    """Apply activation checkpointing to the model. This is orthogonal to FSDP2 so it can be applied pre-sharding or post-sharding.
+
+    This method follows the same logic as FSDP1 as well as TorchTitan's AC example. FSDP1 doesn't support op-based
+    selective activation checkpointing, so that will not be implemented unless we get sufficient demand.
+
+    Args:
+        model (nn.Module): The model to apply activation checkpointing to.
+        fsdp2_config (FSDP2Config): The FSDP2 configuration.
+    """
+    activation_checkpointing = fsdp2_config.activation_checkpointing
+    activation_cpu_offload = fsdp2_config.activation_cpu_offload
+
+    # Create the base checkpointing wrapper using no_reentrant checkpointing by default as
+    # PyTorch notes that reentrant checkpointing is deprecated and will be removed in a future release
+    opt_checkpoint_wrapper = lambda m: checkpoint_wrapper(
+        m,
+        checkpoint_impl=CheckpointImpl.NO_REENTRANT,
+    ) if activation_checkpointing else (lambda module: module)
+    # Create the combined wrapper which takes cpu offloading into consideration
+    opt_combined_wrapper = (
+        lambda module: offload_wrapper(
+            opt_checkpoint_wrapper(module)
+            if activation_checkpointing else module,  # type: ignore reportGeneralTypeIssues
+        )
+    ) if activation_cpu_offload else opt_checkpoint_wrapper
+
+    # Create the check function to determine if a module should be checkpointed
+    def _check_fn(module: torch.nn.Module) -> bool:
+        if hasattr(module, '_activation_checkpointing'):
+            return bool(module._activation_checkpointing)
+        if hasattr(
+            model,
+            'activation_checkpointing_fn',
+        ) and isinstance(model.activation_checkpointing_fn, Callable):
+            return model.activation_checkpointing_fn(module)
+        return False
+
+    # Apply the activation checkpointing on the model, this uses _recursive_wrap to apply the wrapper all submodules
+    # but doesn't apply the wrapper to the root module
+    apply_activation_checkpointing(model, opt_combined_wrapper, _check_fn)  # type: ignore
--- a/composer/distributed/activation_checkpointing.py
+++ b/composer/distributed/activation_checkpointing.py
@@ -8,102 +8,11 @@ from typing import Callable, Optional
 import torch
 import torch.nn as nn
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
-    _CHECKPOINT_WRAPPED_MODULE,
-    ActivationWrapper,
     CheckpointImpl,
-    OffloadWrapper,
     apply_activation_checkpointing,
     checkpoint_wrapper,
     offload_wrapper,
 )
-
-from composer.utils.parallelism import FSDP2Config
-
-
-def validate_activation_wrapper(
-    module: torch.nn.Module,
-    is_activation_checkpoint_enabled: bool,
-    is_activation_offload_enabled: bool,
-    checkpoint_fn: Optional[Callable] = None,
-) -> None:
-    """Verify that activation checkpointing and offload wrappers exist where expected based on the boolean flags.
-
-    Raises ValueError if validation fails, listing the offending module names.
-
-    Args:
-        module (torch.nn.Module): The root model module to inspect.
-        is_activation_checkpoint_enabled (bool): Whether activation checkpointing wrappers (`ActivationWrapper`) are expected.
-        is_activation_offload_enabled (bool): Whether activation offload wrappers (`OffloadWrapper`) are expected.
-        checkpoint_fn (Optional[Callable]): An optional function to determine if a module should be checkpointed.
-    """
-    offenders: list[str] = []
-
-    def check_module_needs_wrapping(inner_module: torch.nn.Module) -> bool:
-        if checkpoint_fn is not None and checkpoint_fn(inner_module):
-            return True
-        return hasattr(inner_module, '_activation_checkpointing') and bool(inner_module._activation_checkpointing)
-
-    def _walk(mod: torch.nn.Module, prefix: str) -> None:
-        inner_mod = mod
-        actual_is_offloaded = isinstance(mod, OffloadWrapper)
-        actual_is_checkpointed = False
-
-        if actual_is_offloaded:
-            offload_inner = getattr(mod, _CHECKPOINT_WRAPPED_MODULE, None)
-            if offload_inner is None:
-                # We shouldn't ever see this, but it's good to have the check
-                offenders.append(f'{prefix}: OffloadWrapper missing inner module')
-                return
-
-            actual_is_checkpointed = isinstance(offload_inner, ActivationWrapper)
-            if actual_is_checkpointed:
-                inner_mod = getattr(offload_inner, _CHECKPOINT_WRAPPED_MODULE, None)
-                if inner_mod is None:
-                    # We shouldn't ever see this, but it's good to have the check
-                    offenders.append(f'{prefix}: ActivationWrapper missing inner module')
-                    return
-            else:
-                inner_mod = offload_inner
-        elif isinstance(mod, ActivationWrapper):
-            actual_is_checkpointed = True
-            inner_mod = getattr(mod, _CHECKPOINT_WRAPPED_MODULE, None)
-            if inner_mod is None:
-                # We shouldn't ever see this, but it's good to have the check
-                offenders.append(f'{prefix}: ActivationWrapper missing inner module')
-                return
-
-        module_needs_wrapping = check_module_needs_wrapping(inner_mod)
-
-        validation_failed = False
-        error_reasons = []
-        if module_needs_wrapping:
-            if actual_is_checkpointed != is_activation_checkpoint_enabled:
-                validation_failed = True
-                error_reasons.append(
-                    f'Expected checkpointed={is_activation_checkpoint_enabled}, Got={actual_is_checkpointed}',
-                )
-            if actual_is_offloaded != is_activation_offload_enabled:
-                validation_failed = True
-                error_reasons.append(f'Expected offloaded={is_activation_offload_enabled}, Got={actual_is_offloaded}')
-        else:
-            if actual_is_checkpointed:
-                validation_failed = True
-                error_reasons.append('Should not be checkpointed, but is')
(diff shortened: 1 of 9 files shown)


tests/trainer/test_fsdp2.py
```diff
@@ -332,3 +339,71 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+    # Set the activation checkpointing attribute for the fc1 module
+    model.module[0]._activation_checkpointing = True
+
+    trainer = create_trainer_with_model(model=model, num_classes=10, use_fsdp2=True, activation_checkpointing=True)
+    trainer.fit()
+    validate_activation_checkpointing(trainer.state.model.module)
+
+    # TODO: Test with `activation_checkpointing_fn`
+    # TODO: Test with `activation_cpu_offload`
+    # TODO: Test with selective activation checkpointing
```
rithwik-db, line 408: Do we want to support SAC? It might be best to do this on a case-by-case basis once we start using FSDP2 on models in llmfoundry?

bowenyang008: I forgot if composer/FSDP1 supports SAC, if so I think we can support it as well?

composer/distributed/fsdp2.py
```diff
@@ -4,15 +4,64 @@
...
+    apply_activation_checkpointing,
+    checkpoint_wrapper,
+    offload_wrapper,
+)
 from torch.distributed.fsdp._fully_shard import fully_shard
 
 from composer.utils.parallelism import FSDP2Config
 
 
+def apply_ac(model: nn.Module, fsdp2_config: FSDP2Config) -> None:
```
bowenyang008, line 22: since ac works with fsdp2 in an orthogonal or arbitrary order way I think we should put the AC utils in a separate file, and create a new function, say prepare_distributed to handle both fsdp2 and AC, so things can stay modularized and isolated

tests/trainer/test_fsdp2.py
```diff
@@ -340,3 +362,217 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
     # We check with `optimizer.param_id.` (with the period) since `optimizer.param_id` exists
     # by default in the error message's legend
     assert 'optimizer.param_id.' in str(e.value)
+
+
+# Testing activation checkpointing
+
+
+def validate_activation_wrapper(
```
rithwik-db, line 370: I should probably create a test_utils.py file? Or shall we move the activation checkpointing tests to another file? Something like `test_activation_checkpointing.py`?

tests/trainer/test_fsdp2.py
rithwik-db: As discussed in the slack thread, we can support op-based SAC when we see fit

composer/distributed/fsdp2.py
rithwik-db: Moved to a different file, should I update FSDP1 to use this code? I guess the question boils down to whether we want to support `transformer_engine` at this time or if we should do that in the future.

tests/trainer/test_fsdp2.py
```diff
@@ -340,3 +362,217 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+        (True, 2, False),
+        (False, 1, True),
+        (False, 1, False),
+    ],
+)
+def test_fsdp2_activation_checkpointing_attribute(
+    world_size: int,
+    activation_checkpointing: bool,
+    expected_forward_count: int,
+    activation_cpu_offload: bool,
```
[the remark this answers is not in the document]
bowenyang008: did you find similar test in our codebase for FSDP1? and what was your observed behavior in that commit?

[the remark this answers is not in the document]
rithwik-db: Nope they only have this https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md

tests/trainer/test_fsdp2.py
rithwik-db: Also is this a bit too overkill? I just realized that the [existing test for FSDP1](https://sourcegraph.com/github.com/mosaicml/composer/-/blob/tests/trainer/test_fsdp.py?L411-455) does a much simpler check

tests/trainer/test_fsdp2.py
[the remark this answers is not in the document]
bowenyang008: I meant in composer codebase is there any test or test utils we can reuse

tests/trainer/test_fsdp2.py
bowenyang008: I think it is a bit overkill for a test util, but it would be good to publish this as a legalizer in composer/distributed/activation_checkpointing.py instead and use it after AC construction

composer/distributed/activation_checkpointing.py
```diff
@@ -0,0 +1,60 @@
...
+    CheckpointImpl,
+    apply_activation_checkpointing,
+    checkpoint_wrapper,
+    offload_wrapper,
+)
+
+from composer.utils.parallelism import FSDP2Config
+
+
+def apply_ac(model: nn.Module, fsdp2_config: FSDP2Config) -> None:
```
bowenyang008, line 20: I think we should decouple this main API from FSDP2 Config, similar to FSDP2 API, let's provide a generic check_fn: Callable[[nn.Module], bool] = _check_fn as the interface and default value, so users are not locked to the hardcoded _check_fn

composer/distributed/fsdp2.py
```diff
@@ -124,3 +125,7 @@ def prepare_fully_shard(
     # If the optimizer is provided, update the optimizer's parameter groups to use the sharded model's DTensor parameters
     if optimizer is not None:
         update_optimizer_modules(optimizer, model, orig_param_to_name)
+
+    # Apply activation checkpointing
+    if fsdp2_config.activation_checkpointing or fsdp2_config.activation_cpu_offload:
+        apply_ac(model, fsdp2_config)
```
bowenyang008, line 131: instead of coupling it to prepare_fully_shard, let's just create a new file, say prepare_distributed.py and create a new function say parallelize_modle that does the fsdp2 wrap and ac wrap, and we will support tp/ep wrap etc in it, just like torchtitan's parallelize_llama

tests/trainer/test_fsdp2.py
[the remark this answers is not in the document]
bowenyang008: I think I figured out what's wrong with your earlier approach, basically the inner context (and its pack/unpack) override the outer ones, in this case the inner context is the cpu_offload wrapper, check https://github.com/mosaicml/composer/blob/boweny/playground/test_saved_tensor_hooks.py

tests/trainer/test_fsdp2.py
```diff
@@ -340,3 +362,217 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
     # We check with `optimizer.param_id.` (with the period) since `optimizer.param_id` exists
     # by default in the error message's legend
     assert 'optimizer.param_id.' in str(e.value)
+
+
+# Testing activation checkpointing
```
bowenyang008, line 367: I would suggest create a test like this: https://github.com/mosaicml/composer/blob/boweny/playground/test_cuda_memory_usage.py

tests/trainer/test_fsdp2.py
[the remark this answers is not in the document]
rithwik-db: Since we decided to just use cuda memory to define whether or not our checkpointing works, I added a new test that checks things more granularly.

composer/distributed/prepare_distributed.py
```diff
@@ -0,0 +1,41 @@
...
+        config (FSDP2Config | FSDPConfig): The configuration for distributed training. Currently only FSDP2Config is supported.
+        optimizer (Optional[torch.optim.Optimizer]): The optimizer to use for distributed training.
+        auto_wrap_policy (Optional[CustomPolicy]): The auto wrap policy to use for distributed training.
+        activation_checkpointing_check_fn (Optional[Callable]): The function to use to check if a module's activations should be checkpointed or offloaded.
+    """
+    if isinstance(config, FSDPConfig):
+        raise ValueError('FSDPConfig is not supported for now, use FSDP2Config instead')
+
+    if activation_checkpointing_check_fn is not None:
+        assert config.activation_checkpointing or config.activation_cpu_offload, 'Activation checkpointing or offloading must be enabled if activation_checkpointing_check_fn is provided'
```
dakinggg, line 36: raise error instead of assert

composer/distributed/activation_checkpointing.py
```diff
@@ -0,0 +1,162 @@
...
+            return model.activation_checkpointing_fn(module)
+        return False
+
+    return _check_fn
+
+
+def apply_ac(model: nn.Module, fsdp2_config: FSDP2Config, check_fn: Optional[Callable] = None) -> None:
+    """Apply activation checkpointing to the model. This is orthogonal to FSDP2 so it can be applied pre-sharding or post-sharding.
+
+    This method follows the same logic as FSDP1 as well as TorchTitan's AC example. FSDP1 doesn't support op-based
```
dakinggg, line 128: you can delete the comment about op-based from the docstring

composer/distributed/activation_checkpointing.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Create the check function to determine if a module should be checkpointed
+    if check_fn is None:
+        check_fn = generate_default_check_fn(model)
+
+    # Apply the activation checkpointing on the model, this uses _recursive_wrap to apply the wrapper to all submodules
+    # but doesn't apply the wrapper to the root module
+    apply_activation_checkpointing(model, opt_combined_wrapper, check_fn)  # type: ignore
+
+    # Validate that activation checkpointing is applied correctly
```
dakinggg, line 161: Do we need this code? It seems like this would be covered by unit testing and doesn't really need to be in the code path itself? It seems like the validation basically reimplements the traversal part of apply_activation_checkpointing to test whether torch's apply_activation_checkpointing function is correct which seems a bit overkill to me and can be handled by unit tests. lmk if im missing something

dakinggg (approved):
comment about whether the validation is really necessary, otherwise lgtm

composer/distributed/activation_checkpointing.py
rithwik-db: I guess we were considering keeping this as a legalizer function, but I am not really opposed to removing this since it doesn't really add much value other than what we expect (cc: @bowenyang008 what do you think)

composer/distributed/activation_checkpointing.py
```diff
@@ -0,0 +1,162 @@
...
+            model,
+            'activation_checkpointing_fn',
+        ) and isinstance(model.activation_checkpointing_fn, Callable):
+            return model.activation_checkpointing_fn(module)
+        return False
+
+    return _check_fn
+
+
+def apply_ac(model: nn.Module, fsdp2_config: FSDP2Config, check_fn: Optional[Callable] = None) -> None:
```
bowenyang008, line 125: let's just use activation_checkpointing and activation_cpu_offload instead of a FSDP2Config

composer/distributed/activation_checkpointing.py
bowenyang008: @rithwik-db used to have it in the unit test and we thought it was an overkill for unit test so I suggested moving it here. I am open to remove it to save codebase complexity

composer/distributed/prepare_distributed.py
```diff
@@ -0,0 +1,41 @@
...
+from composer.distributed.activation_checkpointing import apply_ac
+from composer.distributed.fsdp2 import prepare_fully_shard
+from composer.utils.parallelism import FSDP2Config, FSDPConfig
+
+
+def parallelize_model(
+    model: torch.nn.Module,
+    config: FSDP2Config | FSDPConfig,
+    optimizer: Optional[torch.optim.Optimizer] = None,
+    auto_wrap_policy: Optional[CustomPolicy] = None,
```
bowenyang008, line 20: can we name it to fsdp_wrap_policy instead to signal it is for FSDP2

composer/distributed/activation_checkpointing.py
dakinggg: ah i see. yeah i think its ok to remove. don't really need to test that the torch function works and the unit tests cover our integration well

composer/distributed/prepare_distributed.py
bowenyang008, line 36: I would like to leave to a TODO that if we find AC is indeed decoupled from FSDP in our large scale experiments, we can decouple these two args from FSDP2 config

composer/distributed/prepare_distributed.py
```diff
@@ -0,0 +1,41 @@
...
+    if isinstance(config, FSDPConfig):
+        raise ValueError('FSDPConfig is not supported for now, use FSDP2Config instead')
+
+    if activation_checkpointing_check_fn is not None:
+        assert config.activation_checkpointing or config.activation_cpu_offload, 'Activation checkpointing or offloading must be enabled if activation_checkpointing_check_fn is provided'
+
+    prepare_fully_shard(model, optimizer, config, auto_wrap_policy)
+
+    if config.activation_checkpointing or config.activation_cpu_offload:
+        apply_ac(model, config, activation_checkpointing_check_fn)
```
bowenyang008, line 41: as we discussed, let's move apply_ac before fsdp, in consistent with torchtitan

composer/distributed/activation_checkpointing.py
rithwik-db: Sounds good, will remove

composer/distributed/prepare_distributed.py
rithwik-db: `fsdp2_wrap_policy` or just `fsdp_wrap_policy`?

composer/distributed/prepare_distributed.py
rithwik-db: Added a `TODO` in the `FSDP2Config` class in `parallelism.py`

tests/trainer/test_activation_checkpointing.py
```diff
@@ -0,0 +1,318 @@
...
+        # only final layer activation remains on GPU everything else is offloaded
+        # 4B * 1024 = 4KB
+        assert delta == pytest.approx(elem_size * num_inputs, abs=1000)
+    elif activation_checkpointing and not activation_cpu_offload:
+        # Saved segment input + output = 2 * (4B * 1024) = 8MB
+        assert delta == pytest.approx(2 * elem_size * num_inputs, abs=1000)
+    else:
+        # Same as offload only case
+        # 4B * 1024 = 4KB
+        assert delta == pytest.approx(elem_size * num_inputs, abs=1000)
```
bowenyang008, line 316: cool!

composer/distributed/fsdp2.py
rithwik-db: We will support transformer engine on a future PR

PR #3833, opened 2025-04-25
Eliminate dead code before torch version 2.4  

bowenyang008:
# What does this PR do?

Remove dead code which is only effective before torch 2.4, which we don't support anymore.

1. Pave the way to support and release torch 2.7 and deprecate torch 2.4/2.5
2. Test Cursor on trivial SDE task

# Changes by AI coder
Most changes are created by Cursor, yet reviewed by myself. Workflow:
Codebase wise one shot change is still hard for Cursor, so current flow is:
```
1. Use agent mode in Cursor
2. Use Claude 3.7 non-thinking
3. 
For file in target_files:
    prompt("currently we support torch versions between 2.4 and 2.6, can you help me apply necessary changes to remove or clean up code related to torch version less than 2.4 in this @file? Please don't modify other comments or docstrings. When you need to delete a function please consider if it is used by other files in @composer as well.")
```

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -11,7 +11,6 @@ import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Optional, Sequence, Union, cast
-from unittest.mock import MagicMock
 
 import numpy as np
 import torch
@@ -31,10 +30,7 @@ from torch.optim.lr_scheduler import LRScheduler
 from torch.utils.data import DataLoader, Dataset
 from torchmetrics import Metric
 
-if version.parse(torch.__version__) >= version.parse('2.3.0'):
-    from torch.amp.grad_scaler import GradScaler  # type: ignore
-else:
-    from torch.cuda.amp.grad_scaler import GradScaler  # type: ignore
+from torch.amp.grad_scaler import GradScaler
 
 from composer.core.data_spec import DataSpec
 from composer.core.event import Event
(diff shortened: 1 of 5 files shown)


composer/utils/checkpoint.py
```diff
@@ -558,64 +559,28 @@ def dist_cp_load(
...
-                dist_cp.load(
-                    state_dict=state_dict,
-                    storage_reader=storage_reader,
-                    planner=load_planner,
-                )
-                # Revert the override
-                nested_dict.traverse_state_dict = traverse_2_4_0
-                sharded_tensor_util.traverse_state_dict = traverse_2_4_0
-            else:
-                raise e
```
bowenyang008, line 611: do we still have checkpoints created before 2.4? Can we safely remove this piece of code?

dakinggg: we should maintain support for old checkpoints.

composer/core/state.py
```diff
@@ -73,6 +75,10 @@
 def fsdp_state_dict_type_context(module: torch.nn.Module, state_dict_type: str = 'full'):
     """Context manager for materializing or loading an fsdp module's state dict.
 
+    .. warning::
```
dakinggg, line 78: can we commit to a version for these and use versioneddeprecationwarning?

composer/utils/checkpoint.py
```diff
@@ -1080,13 +1039,7 @@ def _save_checkpoint(
...
         if dist.get_global_rank() > 0:
             state_dict.pop('rng')
-        # To load optimizer states with 2.0 <= torch < 2.2.3 , the optimizer state must be at the top
-        # level of the state dict because the load_sharded_optimizer_state_dict function
-        # requires a top level state dict key for the optimizer.
-        # See https://github.com/pytorch/pytorch/blob/v2.0.1/torch/distributed/checkpoint/optimizer.py#L271
-        # for more info.
-        if version.parse(torch.__version__) < version.parse('2.2.3'):
-            state_dict['optimizers'] = state_dict['state'].pop('optimizers')
+        # For older versions of PyTorch, we don't need this special handling anymore
```
dakinggg, line 1042: comment doesn't refer to anything?

bowenyang008: good catch 🤦

composer/utils/checkpoint.py
bowenyang008: reverted

dakinggg (approved):

PR #3840, opened 2025-04-30
Upgrade yapf version

dakinggg:
# What does this PR do?
Should resolve issues related to python3.12 support in pre commit.

All changes other than the version upgrade itself are simply the result of changes in yapf.

### DIFF
--- a/tests/utils/test_remote_uploader.py
+++ b/tests/utils/test_remote_uploader.py
@@ -118,7 +118,9 @@ def test_overwrite(overwrite: bool):
     with patch('composer.utils.object_store.utils.S3ObjectStore', DummyObjectStore):
         with patch('tempfile.TemporaryDirectory', _get_tmp_dir):
             with patch('composer.utils.remote_uploader.multiprocessing.get_context', lambda _: fork_context):
-                remote_uploader = RemoteUploader(remote_folder='S3://whatever/path',)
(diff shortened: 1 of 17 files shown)


tests/utils/test_remote_uploader.py
```diff
@@ -118,7 +118,9 @@ def _get_tmp_dir():
     with patch('composer.utils.object_store.utils.S3ObjectStore', DummyObjectStore):
         with patch('tempfile.TemporaryDirectory', _get_tmp_dir):
             with patch('composer.utils.remote_uploader.multiprocessing.get_context', lambda _: fork_context):
-                remote_uploader = RemoteUploader(remote_folder='S3://whatever/path',)
+                remote_uploader = RemoteUploader(
+                    remote_folder='S3://whatever/path',
+                )
```
antoinebrl, line 123: If you remove some trailing commas, yapf would inline some of these statements. For example here:
```suggestion
                remote_uploader = RemoteUploader(remote_folder='S3://whatever/path')
```
I know that `black` as an option for remove trailing commas if possible, not sure about `yapf`.

dakinggg: Thanks, I'm open to that change, but just going to let yapf do what it does with current settings for now

irenedea (approved):

PR #3841, opened 2025-05-01
Add two legacy torch images to the container build matrix

asfandyarq:
# What does this PR do?

This PR adds two new torch 2.6 images to the container build matrix. These are needed to support older production platforms that are stuck on Ubuntu 20.04 and Python 3.11.


KuuCi (commented):
I think you have to also make ubuntu version passable here:
https://github.com/mosaicml/composer/blob/main/docker/Dockerfile#L252-L258
Right now the error you are running into is it's trying to install MOFED for 22.04, not the legacy version, otherwise lgtm

docker/generate_build_matrix.py
```diff
@@ -37,12 +38,14 @@ def _version_geq(v1: str, v2: str):
     return packaging.version.parse(v1) >= packaging.version.parse(v2)
 
 
-def _get_base_image(cuda_version: str):
+def _get_base_image(cuda_version: str, ubuntu_version: str = ''):
```
dakinggg, line 41: reason to not put the default value in the params itself?

dakinggg (commented):
looks fine to me, lmk when the images successfully build and you've validate them and ill approve. I think you need to add the ubuntu version arg to the pr-docker.yaml github action as well.

docker/generate_build_matrix.py
asfandyarq: done. was following how `cuda_version` is handled.

dakinggg (approved):

PR #3843, opened 2025-05-02
Added hook handles for FSDP2 to address automicrobatching

rithwik-db:
Added hook handles for FSDP2 to prevent deadlocks during automicrobatching (+ example of how this works since the actual code can be pretty hard to understand)


composer/distributed/fsdp2.py
```diff
@@ -3,28 +3,104 @@
...
+    Here's an example of why this is needed using a simple 2-GPU setup and how it handles OOM issues during auto microbatching:
+
+    - Rank 0: Layer 1 works fine
+    - Rank 1: Layer 1 works fine
+    - Rank 0: Layer 2 works fine
+    - Rank 1: Layer 2 OOMs
+        - Rank 1 raises an error _is_cuda_oom() [[trainer.py:2756]]
+        - Rank 1 sets found_cuda_oom to 1 [[trainer.py:2758]]
+        - Rank 1 creates found_cuda_oom_tensor = [1] and calls all_reduce on it with reduce_operation='MAX' [[trainer.py:2773]]
+    - Rank 2: Layer 2 works fine until a hook handle is hit
```
bowenyang008, line 36: if there are two GPUs, is this Rank 2 a typo, would only expect 0 and 1

tests/common/models.py
```diff
@@ -180,6 +181,29 @@ def __init__(
...
+    def __init__(self, num_layers: int, num_classes: int, device: Union[str, torch.device], always_fail: bool = False):
+        module = torch.nn.Sequential(
+            *[torch.nn.Linear(num_classes, num_classes, device=device) for _ in range(num_layers)],
+        )
+        super().__init__(
+            num_classes=num_classes,
+            module=module,
+        )
+        self.module = module
+        self.rank = dist.get_global_rank()
```
bowenyang008, line 196: seems no need to save this? why not just call it in the fwd?

tests/common/models.py
```diff
@@ -180,6 +181,29 @@ def __init__(
...
+        self.always_fail = always_fail
+
+    def forward(self, batch: tuple[torch.Tensor, Any]) -> torch.Tensor:
+        inputs, _ = batch
+        outputs = self.module(inputs)
+        if self.rank == 1 and (self.always_fail or inputs.shape[0] >= 64):
```
bowenyang008, line 202: any comments on why inputs.shape[0] >= 64?

tests/trainer/test_fsdp2.py
```diff
@@ -354,3 +365,109 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(world_si
...
+
+@world_size(2)
+@pytest.mark.gpu
+@fsdp2_context
+@pytest.mark.filterwarnings("ignore:`device_train_microbatch_size='auto'` may potentially fail with unexpected.*")
+@pytest.mark.parametrize(
+    'use_alternate,num_layers,expected_num_hooks',
+    [
+        # 3 children modules wrapped * 2 hook handles per module
+        # + 1 hook handle for the root module
```
bowenyang008, line 378: is this added regardless of FSDP wrapping? I am curious why this root module hook is necessary if you figured out

composer/distributed/fsdp2.py
```diff
@@ -3,28 +3,104 @@
...
+        - Rank 2 exits the while loop and adjusts the device_train_microbatch_size to half of the previous value [[trainer.py:2790]]
+
+    Args:
+        device (torch.device): The device to check for OOM.
+
+    Returns:
+        Callable: The hook that checks if any other rank hit an OOM.
+    """
+
+    def sync_hook(*args):
```
dakinggg, line 70: is this code exact duplicate of the existing code? if yes, can we factor it out and share?

bowenyang008: this reminds me that we might have duplicated the code of the AC wrapper as well. The latter one can be addressed in a follow up as well

tests/common/models.py
rithwik-db: Mentioned here: https://github.com/mosaicml/composer/pull/3843/files#diff-28ba25a244aea4b7eaaa0b7e71bdc381ea544fae353afd0ee21c1b6ede04ffa7R461, I guess I can add a comment to this part of the code as well.

tests/trainer/test_fsdp2.py
rithwik-db: I left a comment in `shared_utils.py` saying we should investigate that further (will focus on that in a separate PR if I have time after the B200 work) but in any case, it shouldn't hurt.

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+    The following sync hooks are added to prevent FSDP deadlocks that are caused when some ranks OOM
+    and other ranks do not OOM, leading to OOMing ranks calling all_reduce to wait on the non-OOMing
+    ranks and the non-OOMing ranks calling all_gatherbase to continue with FSDP training:
+
+    forward_pre_hook: before forwards of FSDP modules
+    full_backward_pre_hook: before backwards of FSDP modules
+    full_backward_hook: before a prefetched unshard called by FSDP's `post_backward_reshard`
+
+    View https://github.com/mosaicml/composer/pull/3510 for more details.
```
bowenyang008, line 109: actually after reading it, I’m still not sure I follow.
1. IIUC, to prevent hang it basically needs to add a hook check before any FSDP derived collective communications, and for FSDP1, there are potentially 3 places for AG: pre-forward/backward, post-backward
2. then why is it only registered to post backward_hook of non-FSDP module, if it is post-backward prefetch, shouldn't it also apply to FSDP modules as well?
3. why is not prepend=True, if False, it would be executed after the prefetch and still hangs?
4. what about reduce-scatter? is it possible some ranks would OOM before SC and some are not? or is it because SC is a less memory intensive op than AG, so it is not a big issue?
5. if FSDP2 only has AG at two places aka pre-fwd/bwd hooks, does it mean the else is useless in FSDP2?

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    We check if other ranks OOMed after forward/backward pass when using auto microbatching. This
+    may happen when close to memory limit or with uneven memory usage across ranks. Since we
+    need to do this before the model weights are gathered for the next FSDP(1/2) block, we wrap every
+    FSDP(1/2) block with a hook that checks if any other rank OOMed.
+
+    Here's an example of why this is needed using a simple 2-GPU setup and how it handles OOM issues during auto microbatching:
+
+    - Rank 0: Layer 1 works fine
+    - Rank 1: Layer 1 works fine
+    - Rank 0: Layer 2 works fine
```
bowenyang008, line 44: remove?

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    - Rank 1: Layer 2 OOMs
```
bowenyang008, line 45: Rank 0?

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    if device is None:
+        device = get_device()
+    hook = generate_oom_hook(device)
+    assert fsdp_module_type in get_valid_fsdp_module_types(), f'Invalid FSDP module type: {fsdp_module_type}'
+    for module in model.modules():
+        if isinstance(module, fsdp_module_type):
+            hook_handles.append(module.register_forward_pre_hook(hook, prepend=True))  # type: ignore
+            hook_handles.append(module.register_full_backward_pre_hook(hook, prepend=True))  # type: ignore
+        else:
+            hook_handles.append(module.register_full_backward_hook(hook))  # type: ignore
```
bowenyang008, line 129: So I am not sure if this else is necessary for FSDP2 at all, I would suggest we start w/o it and ran some experiments to see if it is necessary

tests/trainer/test_fsdp2.py
bowenyang008: can we run an experiment w/o the post-fwd-hook for FSDP2 to see if auto microbatching still works?

composer/distributed/shared_utils.py
rithwik-db: thanks for catching this

tests/trainer/test_fsdp2.py
rithwik-db: We don't have a `post-fwd-hook` right? We only have `pre-forward`, `pre-backward`, and `full-backward`

composer/distributed/shared_utils.py
rithwik-db: Maybe my intuition is wrong, but my understanding is as long as we have one hook (pre-forward, pre-backward, full-backward), things should work fine for FSDP1 and FSDP2 due to the non-blocking nature of the collectives. It seems that all of these additional hooks are added just to be more thorough in the checks. Since we remove these handles if we succeed >= 3 times, it seems fine to keep this last check as well. If you feel strongly about this, I am not opposed to removing this.

bowenyang008: I don't have strong opinion whether we should keep it, I am just more curious how it works so we don't leave code we don't understand behind

composer/distributed/shared_utils.py
rithwik-db: I took a deeper look into the FSDP2 code and I'll answer these questions in a different order

Sidenote: this [file](https://sourcegraph.com/github.com/pytorch/pytorch/-/blob/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py) has pretty much all of the context.

4. Yes, exactly. Reduce scatter could in theory run into OOMs, but it's much less likely as all gather complexity scales with the number of GPUs in the shard group whereas for reduce scatter, memory usage maxes out at the size of the input tensor being RS'd.

5. In the FSDP2 code, the only times where `unshard` is called is in `FSDPState._pre_forward()` and `FSDPState._pre_backward()`. Note that `FSDP._pre_backward()` also calls `_backward_prefetch()` which `unshard`s the next FSDP param, so unlike FSDP1, we don't have to worry about `backward_prefetch=BACKWARD_POST` (which is also mentioned [here](https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md)). Therefore, I will gate the `register_full_backward_hook` based on whether we're doing FSDP1)

1. Yep, context is in answer to 5 (also this [code](https://github.com/pytorch/pytorch/blob/d4a79d4a7c746068d25fe5cf9333495561f4ce1f/torch/distributed/fsdp/_runtime_utils.py#L1241-L1245))

2. This, I don't get, the backward handles only gets registered on previous FSDP1 modules to prefetch the next FSDP1 module so it doesn't make sense to me why we need this conditional in the first place. Maybe there's something I'm missing...

3. Assuming we are only adding to non-FSDP modules, prepend doesn't need to exist since there's no competition for hooks (no other hooks exist on these modules)

bowenyang008: RE: 3. I forgot why I left such a comment, cuz the code is indeed prepend=True... we can ignore this point

rithwik-db: Actually on second thought, it generally doesn't make sense to me why don't consider after the forward (`register_forward_hook`) since OOMs could cause the `reduce_scatter` to hang since we now have the full params + activations on memory and also the `register_full_backward_hook` situation since params + gradients are now in memory. If we want to be thorough with deadlocks, why not just add all of these to mitigate issues with all of the blocking collective operations?

rithwik-db: > RE: 3. I forgot why I left such a comment, cuz the code is indeed prepend=True... we can ignore this point
Prepend is not true for the `register_full_backward_hook`)...

rithwik-db: > In the FSDP post-backward-hook, [post_backward_reshard](https://github.com/pytorch/pytorch/blob/d4a79d4a7c746068d25fe5cf9333495561f4ce1f/torch/distributed/fsdp/_runtime_utils.py#L796) prefetches unshard, moving the unshard from the beginning of the next backward to the end of this backward

Also, if we use the idea that: FSDP module successfully prefetches -> OOM -> next non-FSDP modules backward happens, it kind of makes sense to include this here but it doesn't actually prevent the deadlock if the allgather is what caused the OOM...

bowenyang008: 🤦 so I would suggest we start from something simple and just for FSDP2, e.g., just add the hooks to FSDP2 modules on pre_fwd/bwd, we will see if we need the extra one when we run more experiments

composer/distributed/prepare_distributed.py
```diff
@@ -81,17 +88,20 @@ def parallelize_model(
...
 
+    return hook_handles, named_modules
+
 
 def parallelize_composer_model(
     composer_model: ComposerModel,
     optimizer: Optional[torch.optim.Optimizer],
     config: FSDP2Config,
-):
+    auto_microbatching: bool = False,
```
bowenyang008, line 103: I would prefer handling the named_modules in this function if possible as this function is designed to handle ComposerModel specific logic and prepare_fully_shard is for more general torch Modules

composer/distributed/prepare_distributed.py
```diff
@@ -81,17 +88,20 @@ def parallelize_model(
     # Use the context manager for optimizer synchronization if optimizer is provided
     with sync_optimizer_and_model_params(optimizer, model) if optimizer is not None else nullcontext():
         with log_execution_time(log, 'Prepare FSDP2'):
-            prepare_fully_shard(model, config, fsdp_wrap_policy)
+            hook_handles, named_modules = prepare_fully_shard(model, config, fsdp_wrap_policy, auto_microbatching)
```
bowenyang008, line 91: similarly I think we can move hook_handle registration to this function instead of prepare_fully_shard so:
1. prepare_fully_shard can stay focused on sharding modules and this function is responsible for general distributed training, e.g., we may these hooks for TP/SP as well in the future
2. we don't have cascaded argument passing between functions

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        list: List of root modules from a ComposerModel.
+    """
+    assert isinstance(model, ComposerModel)
+    root_modules = []
+    for child in model.children():
+        if isinstance(child, (Metric, MetricCollection)):
+            continue
+        root_modules.append(child)
+
+    return root_modules
```
bowenyang008, line 49: the name of this func is a bit confusing, I would expect a name like: _get_direct_children_from_composer_model, and since it is likely internal use only I add an _

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def generate_oom_hook(device: Device) -> Callable:
```
bowenyang008, line 52: nit: seems we can simplify this with a partial , e.g.,
```
def sync_hook(*args, device): ...
...
hook = partial(sync_hook, device=device)
```

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,170 @@
...
+        root_modules_for_hooks = get_root_modules_from_composer_model(model)
+    else:
+        root_modules_for_hooks.append(model)
+
+    # In FSDP2, we don't support backward_prefetch=BACKWARD_POST, so we only need to add the OOM hooks to the
+    # forward pre and backward pre hooks.
+    # TODO: In FSDP1, we might not need the non-FSDP wrapped backward hook either, but we'll keep it for now until further investigation.
+    # TODO: If we want to reduce as many potential deadlocks as possible, we may need to add hooks before all blocking collectives:
+    #   - register_forward_pre_hook (before blocking all_gather)
+    #   - register_forward_hook (before blocking reduce_scatter)
```
bowenyang008, line 156: why is there a RS during fwd?

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,170 @@
...
+    valid_types: dict[int, type] = {1: FullyShardedDataParallel}
+
+    if version.parse(torch.__version__) >= version.parse('2.6.0'):
+        from torch.distributed.fsdp._fully_shard import FSDPModule
+        valid_types[2] = FSDPModule
+
+    return valid_types
+
+
+def get_root_modules_from_composer_model(model: ComposerModel):
```
bowenyang008, line 34: return types?

composer/distributed/shared_utils.py
```diff
@@ -0,0 +1,170 @@
...
+from packaging import version
+from torch.utils.hooks import RemovableHandle
+from torchmetrics import Metric, MetricCollection
+
+from composer.devices import Device
+from composer.models import ComposerModel
+from composer.utils import dist, get_device
+
+
+def get_valid_fsdp_module_types():
```
bowenyang008, line 18: nit: seems an overkill to return a dict, a util with an if-else or match/case seems enough?

composer/distributed/shared_utils.py
rithwik-db: That was a brainfart, they just free the memory that's not being used anymore ([here](https://sourcegraph.com/github.com/pytorch/pytorch@335c89c6f100bd23a7b46c815f0cecbfea733174/-/blob/torch/distributed/fsdp/_fully_shard/_fsdp_param.py?L539-542))

tests/trainer/fsdp2_context.py
```diff
@@ -10,6 +10,8 @@
 SKIP_TEST = version.parse(torch.__version__) < version.parse('2.6.0')
 if not SKIP_TEST:
     # TODO (FSDP2) move this to top once we deprecate torch 2.5
+    from torch.distributed import fsdp
```
bowenyang008, line 13: why not just just use FSDPModule directly as it is available over all versions?

tests/trainer/test_fsdp2.py
```diff
@@ -8,35 +8,39 @@
...
-from tests.trainer.fsdp2_context import fsdp2_context
+from tests.trainer.fsdp2_context import FSDPModule, fsdp2_context
 
 _INIT_DEVICES = ['cuda', 'meta']
 
 
 def create_trainer_with_model(
     model: ComposerClassifier,
     num_classes: int = 10,
+    size: int = 2,
```
bowenyang008, line 33: nit: dataset_size?

composer/distributed/prepare_distributed.py
bowenyang008: I am just wondering if we can have state.automicrobatch_fsdp_hook_handles automatically look for microbatching handles instead of passing it throughout function calls

tests/trainer/test_fsdp2.py
```diff
@@ -54,8 +58,10 @@ def create_trainer_with_model(
         train_dataloader=dataloader,
         max_duration=max_duration,
         parallelism_config=parallelism_config,
+        device_train_microbatch_size='auto' if auto_microbatching else None,
     )
-    return trainer
+
+    return trainer, trainer.state.automicrobatch_fsdp_hook_handles
```
bowenyang008, line 64: seems unnecessary if you can access  automicrobatch_fsdp_hook_handles from trainer

tests/trainer/test_fsdp2.py
```diff
@@ -309,3 +315,109 @@ def test_fsdp2_optimizer_raises_error_when_optimizer_modules_dont_match(
...
+
+    # Assert number of hooks on each module
+    # Note: reshard_after_forward doesn't change the number of backward_hooks, it just changes the existing hooks do so the numbers
+    # below are the same for both reshard_after_forward = True and False.
+    error_msg = 'Expected {} forward pre hooks on module {}, but got {}'
+    for child in model.module.children():
+        if isinstance(child, FSDPModule):  # type: ignore
+            # Hooks exist for FSDP wrapped modules
+            assert len(child._forward_pre_hooks) == 2, error_msg.format(2, child, len(child._forward_pre_hooks))
+            assert len(child._backward_pre_hooks) == 1, error_msg.format(1, child, len(child._backward_pre_hooks))
```
bowenyang008, line 375: wait, shouldn't this also have two hooks?

composer/distributed/shared_utils.py
bowenyang008: but freeing the memory should be just a reshard/slice of the unsharded weight right? why is there a RS?

bowenyang008:
can we run an e2e test to verify it works?

tests/trainer/test_fsdp2.py
rithwik-db: As discussed, these hooks are added during runtime by the forward hooks.

composer/distributed/prepare_distributed.py
rithwik-db: Since this is tied to FSDP2, I am going to add a helper function in `fsdp2.py` and used it in `prepare_distributed.py` since this helps us get the best of both worlds.

composer/distributed/fsdp2.py
```diff
@@ -140,3 +145,20 @@ def prepare_fully_shard(
...
+        hook_handles = add_fsdp_oom_hooks(model, fsdp_config_version=2)
+
+    # Get the right named modules given that the input can be a ComposerModel
+    named_modules = {}
+    if isinstance(model, ComposerModel):
+        root_modules = get_root_modules_from_composer_model(model)
+        for root_module in root_modules:
+            named_modules.update(dict(root_module.named_modules()))
+    else:
+        named_modules = dict(model.named_modules())
```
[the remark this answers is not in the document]
rithwik-db: thanks for catching this, I updated this to only be updated when we are automicrobatching

composer/distributed/shared_utils.py
rithwik-db: I updated the name as per your suggestion, but since this function isn't internal, I didn't add a `_`

rithwik-db:
> can we run an e2e test to verify it works?

It seems hard to curate an e2e test to catch this failure that would be more informative than our current unit tests. We have these two unit tests:
1. [FSDP1](https://sourcegraph.com/github.com/mosaicml/composer@ee8f74f5e755439d2bb3c99b705e118318ffe69a/-/blob/tests/trainer/test_fsdp.py?L253-273)
2. [FSDP2](https://github.com/mosaicml/composer/pull/3843/files#diff-28ba25a244aea4b7eaaa0b7e71bdc381ea544fae353afd0ee21c1b6ede04ffa7R392)

I guess in theory, we could create a larger example where a certain MPT module raises a CUDA OOM error at a certain epoch given a specific batch size or we could use a MPT module with a massive hidden layer in the FFN that will run into OOM for one batch size but not 1/2 of it...

composer/distributed/prepare_distributed.py
bowenyang008: we can discuss the 2nd comment offline, just figured out there might be a better way to handle this

bowenyang008:
> > can we run an e2e test to verify it works?
> 
> It seems hard to curate an e2e test to catch this failure that would be more informative than our current unit tests. We have these two unit tests:
> 
> 1. [FSDP1](https://sourcegraph.com/github.com/mosaicml/composer@ee8f74f5e755439d2bb3c99b705e118318ffe69a/-/blob/tests/trainer/test_fsdp.py?L253-273)
> 2. [FSDP2](https://github.com/mosaicml/composer/pull/3843/files#diff-28ba25a244aea4b7eaaa0b7e71bdc381ea544fae353afd0ee21c1b6ede04ffa7R392)
> 
> I guess in theory, we could create a larger example where a certain MPT module raises a CUDA OOM error at a certain epoch given a specific batch size or we could use a MPT module with a massive hidden layer in the FFN that will run into OOM for one batch size but not 1/2 of it...

I meant just a general e2e test, does not have to trigger OOM

rithwik-db:
Tested here: mpt-7b-fsdp2-GkKdt7 @bowenyang008

composer/distributed/fsdp2.py
```diff
@@ -140,3 +139,23 @@ def prepare_fully_shard(
...
+        dict[str, nn.Module]: A dictionary of valid named modules in the ComposerModel.
+    """
+    assert isinstance(model, ComposerModel), f'{type(model)} is not a ComposerModel'
+    hook_handles = add_fsdp_oom_hooks(model, fsdp_config_version=2)
+
+    named_modules = {}
+    direct_children = get_direct_children_from_composer_model(model)
+    for child in direct_children:
+        named_modules.update(dict(child.named_modules()))
+    return hook_handles, named_modules
```
bowenyang008, line 161: actually I am not sure why FSDP1 one do it in such a way, isn't a loop over root_model.modules return the same values? Note the trainer only needs modules instead of named_modules so keys are not useful at all

composer/trainer/trainer.py
```diff
@@ -3824,7 +3807,7 @@ def _parse_parallelism_config(
                 if isinstance(parallelism_config['fsdp'], FSDPConfig | FSDP2Config):
                     parallelism_config_args['fsdp'] = parallelism_config['fsdp']
                 elif os.environ.get('FSDP_VERSION', '1') == '2':
-                    parallelism_config_args['fsdp'] = FSDP2Config.from_compatible_attrs(parallelism_config['fsdp'])
+                    parallelism_config_args['fsdp2'] = FSDP2Config.from_compatible_attrs(parallelism_config['fsdp'])
```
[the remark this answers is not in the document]
bowenyang008: good catch, I think this is indeed a bug but it was shadowed by https://github.com/mosaicml/composer/blob/main/composer/core/state.py#L871 so FSDP2 is still in effect. @rithwik-db can you run an e2e test (in regression-test) to validate it has no effect on end result?

composer/trainer/trainer.py
```diff
@@ -1859,6 +1838,7 @@ def _wrap_model_for_distributed(
                             self.state.seed,
                         )
                     case 2:
+                        assert not auto_microbatching
```
bowenyang008, line 1841: can we make this a bit more verbose with a message?

bowenyang008 (approved):
left a few minor comments, LGTM

composer/trainer/trainer.py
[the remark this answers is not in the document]
rithwik-db: Can do

PR #3844, opened 2025-05-02
Update EFA installer version

dakinggg:
# What does this PR do?
Updates to a more recent EFA installer version. This now includes AWS OFI NCCL plugin without us having to build it ourselves.

Before: `llama70b-8192-ofi-old-image-aws-1-0g5Lef`
After: `llama70b-8192-ofi-new-vendored-aws-6-Ra7thp`

throughput is slightly higher on the new image.

Note: we were originally going to go back to using `latest` for the EFA installer version, but 1.40 is producing an installation an error, so I'm pinning that to 1.39 (previous version). This is a known issue [here](https://github.com/aws/aws-ofi-nccl/issues/872) with the debug install, so should be fixed soonish.


docker/build_matrix.yaml
```diff
@@ -139,7 +139,7 @@
   TARGET: pytorch_stage
   TORCHVISION_VERSION: 0.19.1
   UBUNTU_VERSION: '22.04'
-- AWS_OFI_NCCL_VERSION: v1.11.0-aws
+- AWS_OFI_NCCL_VERSION: v1.14.2
```
rajachan, line 142: FYI, the aws-ofi-nccl plugin is now vended as part of the EFA installer itself starting installer v1.39 (which came with 1.14.1)... Installer v1.39 ships with aws-ofi-nccl plugin v1.14.2.

dakinggg: oh nice, thanks for the heads up!

willgleich (approved):
lgtm thanks for the info regarding the EFA INSTALLER @rajachan

dakinggg:
@willgleich @irenedea this is ready for review again. I removed the separate ofi nccl build since that is included in efa installer now.

@irenedea if you can admin stamp, it should be good. I'll also wait for @willgleich to take a look.

irenedea (approved):
Seems reasonable, stamp for admin approval. will let @willgleich take another pass

docker/Dockerfile
```diff
@@ -202,18 +202,17 @@ RUN if [ -z "$PYTORCH_NIGHTLY_URL" ] ; then \
...
-#####################################
+#############
+# Install EFA
+#############
 
-ARG EFA_INSTALLER_VERSION=1.38.1
-ARG AWS_OFI_NCCL_VERSION
+ARG EFA_INSTALLER_VERSION
 
 ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/opt/amazon/efa/lib:/opt/aws-ofi-nccl/install/lib:$LD_LIBRARY_PATH
```
rajachan, line 211: Did you mean to update the LD_LIBRARY_PATH to drop `/opt/aws-ofi-nccl/install/lib` in favor of the path from the installer as well, given you dropped the custom build now? Ref [doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start-nccl.html#nccl-start-base-enable), it is in `/opt/amazon/ofi-nccl`

dakinggg: ah yes, thanks for the catch!

dakinggg:
Retested after latest changes, still lgtm. Merging.

PR #3850, opened 2025-05-06
Updating composer to support Torch 2.7

rithwik-db:
# What does this PR do?

Updating to support Torch 2.7 (using CUDA 12.6.3)


tests/test_precision.py
```diff
@@ -78,7 +78,7 @@ def predict_and_measure_memory(precision) -> int:
 def test_train_precision_memory(precision: Precision):
     memory_fp32 = fit_and_measure_memory(Precision.FP32)
     memory_half = fit_and_measure_memory(precision)
-    assert memory_half < 0.87 * memory_fp32
+    assert memory_half < 0.89 * memory_fp32
```
rithwik-db, line 81: Torch 2.7 passes when I set a slightly higher value; this seems empirical by nature so I don't think this should be too big of an issue?

tests/test_simple_nlp.py
```diff
@@ -269,7 +269,7 @@ def test_simple_nlp_mlm_loss_gen_token_batch(tiny_bert_tokenizer, device):
         model=model,
         seed=42,
         train_dataloader=data_spec,
-        max_duration='2ep',
+        max_duration='3ep',
```
rithwik-db, line 272: Using 2 epochs makes the `cross_entropy` and `token_cross_entropy` the same for torch 2.7 but since even in earlier torch versions, these values are really similar, I added an extra epoch to exaggerate this loss slightly to make the tests pass.

tests/test_precision.py
irenedea: yeah agreed, ok with this one. Seems reasonable that there are variations in memory with torch/cuda bump

tests/test_simple_nlp.py
irenedea: A little worried about the loss values changing post torch bump -- not expecting changes here. Are they the same loss as before? Or different values all together? (I'm seeing `10.4344` on torch 2.6)

irenedea: Maybe reasonable due to cuda bump...

rithwik-db: Issue happens on non-CUDA bump as well :( (downgraded to CUDA 12.6.3)

rithwik-db: Will reopen new threads

.github/workflows/daily.yaml
```diff
@@ -20,11 +20,6 @@ jobs:
     strategy:
       matrix:
         include:
-        - name: cpu-3.12-2.4
-          container: mosaicml/pytorch:2.4.1_cpu-python3.12-ubuntu22.04
-          markers: not daily and (remote or not remote) and not gpu and not doctest
-          pytest_command: coverage run -m pytest
-          composer_package_name: mosaicml
         - name: cpu-3.12-2.5
```
rithwik-db, line 23: @bowenyang008 @dakinggg, just to clarify, are we okay with removing 2.4 and 2.5? Just wanted to make sure.

dakinggg: yes

.github/workflows/pr-gpu.yaml
```diff
@@ -1,6 +1,6 @@
 name: PR GPU tests
 on:
-  pull_request_target:
+  pull_request:
```
dakinggg, line 3: leaving comment to not merge

dakinggg (commented):
Wondering if we should use cuda 12.8 in prep for blackwell? Any particular reason not to?

tests/test_simple_nlp.py
dakinggg: Could you try exaggerating this by making the dtms smaller instead of increasing epochs?

tests/test_precision.py
```diff
@@ -78,7 +78,7 @@ def predict_and_measure_memory(precision) -> int:
 def test_train_precision_memory(precision: Precision):
     memory_fp32 = fit_and_measure_memory(Precision.FP32)
     memory_half = fit_and_measure_memory(precision)
-    assert memory_half < 0.87 * memory_fp32
+    assert memory_half < 0.92 * memory_fp32
```
dakinggg, line 81: just to make sure we're not missing something here...can you try increasing the model size and see if this threshold can be made lower? may or may not merge depending on how long it makes the test take, but would like to know

rithwik-db: It seems that something is weird here; I can increase the model size to `resnet50`, which gets us below this threshold for training, but for eval and predict, the memory usage is somehow actually higher than with FP32???

rithwik-db: jk just added some garbage collection and the test seem fine now

docker/generate_build_matrix.py
```diff
@@ -211,9 +208,10 @@ def _cross_product_extra_cuda(
 
 
 def _main():
-    python_pytorch_versions = [('3.12', '2.6.0'), ('3.12', '2.5.1'), ('3.12', '2.4.1')]
+    python_pytorch_versions = [('3.12', '2.7.0'), ('3.12', '2.6.0')]
     pytorch_cuda_variants_extra = {
         '2.6.0': ['12.6.3'],
+        '2.7.0': ['12.8.0'],
```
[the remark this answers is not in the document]
dakinggg: yeah that seems reasonable to me

tests/test_simple_nlp.py
rithwik-db: Yeah that worked

tests/test_precision.py
```diff
@@ -77,23 +78,29 @@ def predict_and_measure_memory(precision) -> int:
 @pytest.mark.filterwarnings(r'ignore:.*Plan failed with a cudnnException.*:UserWarning')  # Torch 2.3 regression
 def test_train_precision_memory(precision: Precision):
     memory_fp32 = fit_and_measure_memory(Precision.FP32)
+    gc.collect()
```
bowenyang008, line 81: we can put these gc in fit_and_measure_memory, and instead of gc, can you try trainer.close?

tests/trainer/test_checkpoint.py
```diff
@@ -118,6 +118,14 @@ def _assert_checkpoints_equivalent(file1, file2, atol=0.0, rtol=0.0):
     del checkpoint_1['state']['callbacks']['CheckpointSaver']['all_saved_checkpoints_to_timestamp']
     del checkpoint_2['state']['callbacks']['CheckpointSaver']['all_saved_checkpoints_to_timestamp']
 
+    # Remove verbose label in state/schedulers if exists
```
dakinggg, line 121: what is this?

rithwik-db: Was trying to get rid of this error: https://github.com/mosaicml/composer/actions/runs/15181682096/job/42692335913 but this doesn't seem to resolve it; will figure out a better solution today

rithwik-db: Moved it to a better location

rithwik-db:
PR Docker build is taking a long time to build flash attention so that fails: https://github.com/mosaicml/composer/actions/runs/15199825422/job/42751811513

.github/workflows/docker-configure-build-push.yaml
```diff
@@ -37,6 +37,7 @@ on:
 jobs:
   configure-build-push:
     runs-on: mosaic-4wide
+    timeout-minutes: 600
```
rithwik-db, line 40: TODO: remove this change

composer/utils/checkpoint.py
```diff
@@ -566,6 +566,22 @@ def dist_cp_load(
...
+            #
+            # we use a DefaultLoadPlanner with allow_partial_load=True to allow for missing keys,
+            # added here: https://github.com/pytorch/pytorch/commit/c7338f457cb5fcf1707d44ee9891ed80adfac916
+            # This should be fine since the load planner will ignore the missing key and set
+            # decoupled_weight_decay to False.
+            try:
+                dist_cp.load(
+                    state_dict=state_dict,
+                    storage_reader=storage_reader,
+                    planner=DefaultLoadPlanner(allow_partial_load=True),
```
dakinggg, line 580: nice, didn't know about this!

dakinggg (approved):

PR #3856, opened 2025-05-19
FSDP2 time and verbose logging

bowenyang008:
# What does this PR do?

Support logging of FSDP2 execution time, wrapped model and verbose config

# Test
Tested manually, see: mpt-7b-fsdp2-G2eJiP

### DIFF
--- a/composer/distributed/fsdp2.py
+++ b/composer/distributed/fsdp2.py
@@ -3,6 +3,7 @@
 
 """Helpers for FSDP2."""
 
+import logging
 from typing import Optional
 
 import torch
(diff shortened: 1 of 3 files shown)


dakinggg:
@bowenyang008 Composer loggers need to be used if you want to log to an experiment tracker (mlflow, wandb, etc). We also have a custom logger for the MCLI platform, and custom loggers for progress bar logging and "pretty" training progress console logging. But for general debugging/info statements, just use python logging.

bowenyang008:
> LGTM, please leave MCLI run name in the "manual testing" section of the PR description so it is easy for reviewer to double check and refer back to. Unless run name includes sensitive info, in which case you can DM it to reviewer.

Oh, I just tested locally on my dev machine as the PR is simple enough so I just need to make sure it logs the info to console correctly. But here is a MCLI run I just kicked off: mpt-7b-fsdp2-4Ofhj8

dakinggg:
Ah fair enough. Yeah its nice, to launch a run for posterity (although this is obviously a very minor change)

PR #3863, opened 2025-05-27
Cleanup version gating pre-2.6.0

rithwik-db:
# What does this PR do?

There's a lot of gating that's no longer necessary with the latest support for PyTorch 2.6 and 2.7, and we should clean all of that up.

Continuation of: https://github.com/mosaicml/composer/pull/3833#issue-3021029341


tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -847,13 +835,8 @@ def test_fsdp_partitioned_state_dict_load(
     s3_ephemeral_prefix,
     request,
 ):
-    if use_tp:
```
dakinggg, line 850: this is probably still true

composer/utils/checkpoint.py
```diff
@@ -596,34 +596,7 @@ def dist_cp_load(
                 planner=load_planner,
             )
             return
-
-        checkpoint_metadata = storage_reader.read_metadata().state_dict_metadata
```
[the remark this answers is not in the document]
rithwik-db: Yeah that's what I am thinking, but this code depends on this `traverse_state_dict` that is gated for ~2.4.0, so I am running a daily test to figure out if the failures still occur in the first place since we shouldn't even be able to access this function

rithwik-db:
Daily tests: https://github.com/mosaicml/composer/actions/runs/15286666232/job/42998282228

tests/trainer/test_fsdp_checkpoint.py
rithwik-db: T'was the case

tests/trainer/fsdp2_context.py
```diff
@@ -4,32 +4,19 @@
...
-    check_param_tying = lambda *args, **kwargs: None
+from composer.distributed import activation_checkpointing, fsdp2, fsdp2_utils, prepare_distributed
+
+apply_ac = activation_checkpointing.apply_ac
+parallelize_model = prepare_distributed.parallelize_model
+legalize_param_sharing_between_modules = fsdp2_utils.legalize_param_sharing_between_modules
+get_standalone_and_tied_modules = fsdp2_utils.get_standalone_and_tied_modules
+_recursive_apply_fully_shard = fsdp2._recursive_apply_fully_shard
+_generate_default_policy = fsdp2_utils.generate_default_policy
+check_param_tying = fsdp2_utils.check_param_tying
```
bowenyang008, line 16: should we just import these directly in relevant files instead of still keeping such an indirection?

rithwik-db: Also @bowenyang008 I am planning on moving the FSDP2 experimental warning check in `@fsdp2_context` to `pyproject.toml` so that I can just delete `fsdp2_context.py` entirely for simplicity

bowenyang008: cool, didn't know such a way

PR #3864, opened 2025-05-28
Bumping dev version to 0.32.0.dev0 and deprecated some features

rithwik-db:
# What does this PR do?

Bump to new dev version and deprecated 1) LocalEvalClient 2) LambdaEvalClient and 3) composer.core.state.fsdp_get_optim_state_dict


rithwik-db:
Oops, forgot that there's two rounds of deprecation checks

composer/utils/eval_client/eval_client.py
```diff
@@ -12,7 +12,7 @@
 
 
 class EvalClient(abc.ABC):
```
dakinggg, line 14: should just delete the base class. the deprecations of the subclasses was intended to indicate deprecation of the whole concept, without adding a constructor to the abstract base class

rithwik-db: sure can do

dakinggg (approved):

PR #3867, opened 2025-05-29
Using cu128 instead of cu126 for pr-gpu and daily tests

rithwik-db:
# What does this PR do?

Just updated the CUDA version to use the most modern image (CU128 instead of CU126)


rithwik-db:
Tests pass for this: https://github.com/mosaicml/composer/pull/3867/commits/ed030f45bdf0195e1209d0f9680ed129e6b5491c

rithwik-db:
Running daily tests here: https://github.com/mosaicml/composer/actions/runs/15334900129

PR #3869, opened 2025-05-30
Removed most s3 bucket based tests (replaced with UC Volumes)

rithwik-db:
# What does this PR do?

Essentially, all usages of `s3_bucket` has been removed outside of `test_object_store.py` and `test_s3_object_store.py` since we have the ephemeral and read_only paths in UC Volumes now.

### DIFF
--- a/composer/utils/reproducibility.py
+++ b/composer/utils/reproducibility.py
@@ -232,6 +232,16 @@ def load_rng_state(rng_state_dicts: list[dict[str, Any]]):
         log.debug('Restoring the RNG state')
 
         if is_cuda_available and has_cuda_rng_state:
+            # PyTorch 2.6 strictly enforces the size of the state dict values vs the size in the checkpoint.
+            # PyTorch 2.1 has moved to a RNG of size 16 instead of 816
+            # Therefore, errors occur if we load a ckpt at or after 2.6 if the ckpt was saved before 2.1.
+            # To avoid this, we zero out the CUDA RNG state if the size is incorrect.
+            if rng_state_dict['cuda'].shape[0] != 16:
+                warnings.warn(
+                    'The CUDA RNG state was saved with a different version of PyTorch than the current version.'
+                    'CUDA RNG state will be zeroed out instead.',
+                )
+                rng_state_dict['cuda'] = torch.zeros(16, dtype=torch.uint8)
             try:
                 torch.cuda.set_rng_state(rng_state_dict['cuda'])
             except RuntimeError as e:
--- a/composer/utils/reproducibility.py
+++ b/composer/utils/reproducibility.py
@@ -232,16 +232,6 @@ def load_rng_state(rng_state_dicts: list[dict[str, Any]]):
         log.debug('Restoring the RNG state')
 
         if is_cuda_available and has_cuda_rng_state:
-            # PyTorch 2.6 strictly enforces the size of the state dict values vs the size in the checkpoint.
-            # PyTorch 2.1 has moved to a RNG of size 16 instead of 816
-            # Therefore, errors occur if we load a ckpt at or after 2.6 if the ckpt was saved before 2.1.
-            # To avoid this, we zero out the CUDA RNG state if the size is incorrect.
-            if rng_state_dict['cuda'].shape[0] != 16:
-                warnings.warn(
-                    'The CUDA RNG state was saved with a different version of PyTorch than the current version.'
-                    'CUDA RNG state will be zeroed out instead.',
-                )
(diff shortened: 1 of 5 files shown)


tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -971,19 +970,18 @@ def test_elastic_resumption(
...
     if autoresume:
         run_name = 'my-autoresume-run'
     else:
         run_name = None
 
-    base_path = (
-        f's3://{s3_bucket}/{s3_read_only_prefix}/elastic_test/'
-        f'{sharding_strategy.lower()}_sharded_{precision}_'
-        f'{num_shards}/'
+    base_path = os.path.join(
```
rithwik-db, line 981: Same issue with `.metadata` not existing (daily tests that fail: https://github.com/mosaicml/composer/actions/runs/15356296375/job/43215949886) cc: @ethantang-db, can you lmk when you move this files? Thanks!

dakinggg:
Ah I think its because dist is initialized with gpu (i.e. nccl) rather than cpu (i.e gloo) when running the gpu tests. its probably not worth adjusting test set up to support this, and just let the test run on gpu instead of cpu

composer/utils/reproducibility.py
```diff
@@ -232,6 +232,16 @@ def load_rng_state(rng_state_dicts: list[dict[str, Any]]):
...
+            # PyTorch 2.6 strictly enforces the size of the state dict values vs the size in the checkpoint.
+            # PyTorch 2.1 has moved to a RNG of size 16 instead of 816
+            # Therefore, errors occur if we load a ckpt at or after 2.6 if the ckpt was saved before 2.1.
+            # To avoid this, we zero out the CUDA RNG state if the size is incorrect.
+            if rng_state_dict['cuda'].shape[0] != 16:
+                warnings.warn(
+                    'The CUDA RNG state was saved with a different version of PyTorch than the current version.'
+                    'CUDA RNG state will be zeroed out instead.',
+                )
+                rng_state_dict['cuda'] = torch.zeros(16, dtype=torch.uint8)
```
rithwik-db, line 244: It seems that since we updated to use gpu, the `rng_state_dict` is relevant now and this is causing issues with equality in the metrics/models. I reduced the tolerance to get the tests to pass on the daily workflow but the current tolerances work fine on my remote instance. Not sure what I am missing...

dakinggg:
I'm ok just adjusting the tolerance for 2.6 if thats sufficient for the test to pass. Not super critical to keep exact numerical determinism across torch versions - and in fact on GPU I'd guess this might not be possible.

tests/trainer/test_checkpoint.py
```diff
@@ -1154,42 +1155,49 @@ def test_save_ignore_keys(self, save_ignore_keys, weights_equal, callbacks_equal
...
     @pytest.mark.parametrize(
-        'remote_checkpoint_uri, remote_checkpoint_name, continue_training_dur, final_checkpoint_name',
+        'remote_checkpoint_name, continue_training_dur, final_checkpoint_name',
         [
-            ['backwards_compatibility/trained_ckpt_cpu_ep2.pt', 'ep2.pt', '3ep', 'ep3.pt'],
+            ['ep2.pt', '3ep', 'ep3.pt'],
         ],
     )
     @pytest.mark.filterwarnings('ignore:.*The checkpoint included CUDA RNG state.*')
+    @pytest.mark.filterwarnings('ignore:.*The CUDA RNG state could not be loaded from the checkpoint.*')
```
dakinggg, line 1167: don't need to filter this anymore, right?

tests/trainer/test_fsdp_checkpoint.py
```diff
@@ -525,13 +525,14 @@ def test_fsdp_mixed_with_sync(
 @pytest.mark.filterwarnings(r'ignore:.*The CUDA RNG state could not be loaded.*:UserWarning')
 @pytest.mark.filterwarnings(r'ignore:.*ShardedTensor.to only move tensor to its current device.*:UserWarning')
 @pytest.mark.filterwarnings(r'ignore:.*\(TP\) is experimental.*:FutureWarning')
+@pytest.mark.filterwarnings(r'ignore:.*The CUDA RNG state was saved with a different version of PyTorch.*')
```
dakinggg, line 528: why was this change needed? we didn't change the checkpoints being used here right?

dakinggg (approved):

rithwik-db:
Reran daily tests to make sure with latest fixes: https://github.com/mosaicml/composer/actions/runs/15500553934

PR #3872, opened 2025-06-03
Supporting Mixed Init on FSDP2

rithwik-db:
# What does this PR do?

Supporting mixed init on FSDP2.

### DIFF
--- a/composer/distributed/fsdp2.py
+++ b/composer/distributed/fsdp2.py
@@ -10,6 +10,7 @@ import torch
 import torch.nn as nn
 from torch.distributed.fsdp import MixedPrecisionPolicy, fully_shard
 from torch.distributed.fsdp.wrap import CustomPolicy
+from torch.distributed.tensor import DTensor
 
 from composer.distributed.fsdp2_utils import (
     check_param_tying,
@@ -17,6 +18,7 @@ from composer.distributed.fsdp2_utils import (
     get_standalone_and_tied_modules,
     legalize_param_sharing_between_modules,
 )
+from composer.utils import dist
 from composer.utils.parallelism import FSDP2Config
 
 log = logging.getLogger(__name__)
@@ -140,3 +142,41 @@ def prepare_fully_shard(
             if attr == 'verbose':
                 continue
             log.info(f'FSDP2: {attr}: {getattr(fsdp2_config, attr)}')
+
+
+def sync_module_states(model: nn.Module, full_state_dict: dict) -> None:
+    """Syncs the module states of the model across all ranks using the full state dict.
+
+    This function synchronizes the module states of the model across all ranks using the full state dict.
+    This handles mixed initialization scenarios where different ranks have parameters on different devices.
+
+    Args:
+        model (nn.Module): The FSDP2-wrapped model to synchronize.
+        full_state_dict (dict): The full state dict to synchronize. This is only fully populated on rank 0. The other ranks will receive a partial state dict.
+
+    Returns:
+        None
+    """
+    from torch.distributed.checkpoint.state_dict import StateDictOptions, set_model_state_dict
+
+    # In cases where you want to FSDP2 on CPU (although not recommended),
+    device = torch.cuda.current_device() if torch.cuda.is_available() else 'cpu'
+
+    if dist.get_global_rank() == 0:
+        model = model.to(device=device, non_blocking=True)
+    else:
+        model = model.to_empty(device=device)
+
+    # strict=False is required for when we have multiple references to the same module
+    # since otherwise, we would have to load the parameters of the same module multiple
+    # times using different names for the weights and that's not viable with the internal
+    # logic of set_model_state_dict.
+    options = StateDictOptions(full_state_dict=True, broadcast_from_rank0=True, strict=False)
+
+    # Sync parameters and buffers
+    set_model_state_dict(model, full_state_dict, options=options)
+
+    # Sync additional buffers that may not be in the state_dict
+    for _, buffer in model.named_buffers():
+        assert not isinstance(buffer, DTensor), 'Buffers should not be DTensor'
+        dist.broadcast(buffer, src=0)
--- a/composer/distributed/prepare_distributed.py
+++ b/composer/distributed/prepare_distributed.py
@@ -15,7 +15,7 @@ from composer.distributed.activation_checkpointing import apply_ac, generate_com
 from composer.distributed.fsdp2 import prepare_fully_shard, sync_module_states
 from composer.distributed.fsdp2_utils import generate_composer_model_policy, sync_optimizer_and_model_params
 from composer.distributed.param_init import meta_init
-from composer.distributed.shared_utils import validate_model_requires_state_sync
+from composer.distributed.shared_utils import update_sync_module_states_if_needed
 from composer.models import ComposerModel
 from composer.utils import dist
 from composer.utils.parallelism import FSDP2Config
@@ -69,7 +69,7 @@ def _parallelize_model_helper(
     2. With sync_module_states: param_init on rank 0 first, then fully_shard, then broadcast the
        initialized state to all other ranks. This makes sure that all ranks have rank 0's model state.
     """
-    validate_model_requires_state_sync(model, config)
+    update_sync_module_states_if_needed(model, config)
 
     if config.sync_module_states:
         # If we are syncing module states, we assume that rank 0 has the model on CPU/GPU
--- a/composer/distributed/prepare_distributed.py
+++ b/composer/distributed/prepare_distributed.py
@@ -12,10 +12,12 @@ import torch
 from torch.distributed.fsdp.wrap import CustomPolicy
 
 from composer.distributed.activation_checkpointing import apply_ac, generate_composer_model_check_fn
-from composer.distributed.fsdp2 import prepare_fully_shard
+from composer.distributed.fsdp2 import prepare_fully_shard, sync_module_states
 from composer.distributed.fsdp2_utils import generate_composer_model_policy, sync_optimizer_and_model_params
 from composer.distributed.param_init import meta_init
+from composer.distributed.shared_utils import validate_model_requires_state_sync
 from composer.models import ComposerModel
+from composer.utils import dist
 from composer.utils.parallelism import FSDP2Config
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


tests/common/models.py
```diff
@@ -205,20 +205,17 @@ class SimpleWeightTiedModel(ComposerClassifier):
     def __init__(self, num_features: int = 1, device: str = 'cpu') -> None:
         self.num_features = num_features
 
-        mlp = SimpleMLP(num_features, device)
-
+        fc1 = torch.nn.Linear(num_features, num_features, device=device, bias=False)
```
rithwik-db, line 208: I had to remove the redundant modules since this keeps throwing errors. Will look into this in more detail tomorrow though

composer/distributed/param_init.py
```diff
@@ -6,10 +6,15 @@
 from typing import Callable, Optional
 
 import torch.nn as nn
-from torch.distributed.tensor import DTensor
 
+from composer.utils import dist
 
-def meta_init(module: nn.Module, param_init_fn: Optional[Callable[[nn.Module], None]] = None) -> None:
+
+def meta_init(
```
rithwik-db, line 13: TODO: Change the name and the description of this function

rithwik-db:
@bowenyang008 @dakinggg this is still a WIP but wanted to get some early feedback if you have time; there's one bug that I need to address (https://github.com/mosaicml/composer/pull/3872#discussion_r2125214457) but ideally that should be it?

composer/distributed/fsdp2.py
```diff
@@ -140,3 +141,27 @@ def prepare_fully_shard(
...
+        None
+    """
+    from torch.distributed.checkpoint.state_dict import StateDictOptions, set_model_state_dict
+
+    if dist.get_global_rank() == 0:
+        model = model.to(device=torch.cuda.current_device(), non_blocking=True)
+    else:
+        model = model.to_empty(device=torch.cuda.current_device())
+
+    options = StateDictOptions(full_state_dict=True, broadcast_from_rank0=True, strict=False)
```
dakinggg, line 166: wondering - is strict=False needed? shouldn't all the keys be present in the rank0 state dict?

composer/distributed/fsdp2.py
```diff
@@ -140,3 +141,27 @@ def prepare_fully_shard(
...
+    Args:
+        model (nn.Module): The FSDP2-wrapped model to synchronize.
+        full_state_dict (dict): The full state dict to synchronize. This is only fully populated on rank 0. The other ranks will receive a partial state dict.
+
+    Returns:
+        None
+    """
+    from torch.distributed.checkpoint.state_dict import StateDictOptions, set_model_state_dict
+
+    if dist.get_global_rank() == 0:
```
dakinggg, line 161: why rank 0 non blocking but other ranks blocking?

dakinggg: ah misread, nvm

composer/distributed/param_init.py
```diff
@@ -6,10 +6,15 @@
...
+    module: nn.Module,
+    param_init_fn: Optional[Callable[[nn.Module], None]] = None,
+    sync_module_states: bool = False,
+) -> None:
```
dakinggg, line 17: i'm not sure i really understand the logical flow here any more. don't we just want to do something like

```
if not sync_module_states:
    same logic as before calling this meta_init function
else:
    call sync_module_states function, which will call to_empty on all ranks with their current device. the model will have already been sharded, so this should be safe?

File: composer/distributed/fsdp2.py [hunk 6]
```diff
@@ -140,3 +141,27 @@ def prepare_fully_shard(
...
+        model = model.to(device=torch.cuda.current_device(), non_blocking=True)
+    else:
+        model = model.to_empty(device=torch.cuda.current_device())
+
+    options = StateDictOptions(full_state_dict=True, broadcast_from_rank0=True, strict=False)
+    set_model_state_dict(model, full_state_dict, options=options)
```
dakinggg, line 167: we probably need to broadcast buffers here too?

rithwik-db: the buffers would exist as part of the `state_dict`, right? I guess https://github.com/volcengine/verl/blob/main/verl/utils/fsdp_utils.py#L416-L418 mentions that there are some buffers that aren't part of the `state_dict` that would be best to broadcast as well though so I will add it.

composer/utils/parallelism.py
```diff
@@ -80,6 +80,7 @@ class FSDP2Config:
     #       in most of our use cases, we can decouple these two attributes from the FSDP2Config class.
     activation_checkpointing: bool = False
     activation_cpu_offload: bool = False
+    sync_module_states: bool = False
```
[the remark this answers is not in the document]
rithwik-db: Essentially instead of using `validate_model_device_arrangement`, we just set it ourselves depending on the "outcome" of that function?

composer/distributed/fsdp2.py
rithwik-db: Added a comment on this

composer/distributed/prepare_distributed.py
```diff
@@ -33,6 +35,59 @@ def log_execution_time(logger: logging.Logger, operation_name: str):
...
 
+def _filter_state_dict_duplicates(model: torch.nn.Module, state_dict: dict) -> dict:
+    """Filter out duplicate params in the state dict.
+
+    This is necessary when sync_module_states is enabled as multiple references
+    to the same module will lead to errors since state_dict() will have multiple references
+    but module.children() will not have multiple references.
+    """
+    filtered_dict = {}
+    valid_params = list(dict(model.named_parameters()).keys())
```
bowenyang008, line 46: why using list not set?

composer/distributed/prepare_distributed.py
```diff
@@ -33,6 +35,59 @@ def log_execution_time(logger: logging.Logger, operation_name: str):
         logger.info(f'{operation_name} took {end_time - start_time:.2f} seconds')
 
 
+def _filter_state_dict_duplicates(model: torch.nn.Module, state_dict: dict) -> dict:
```
bowenyang008, line 38: seems we can just take a model and state_dict is basically model.state_dict()? I would also add a bit more explantion that model.named_parameter() dont' return same references to the same tensor/param while state_dict does

composer/distributed/fsdp2.py
```diff
@@ -140,3 +141,37 @@ def prepare_fully_shard(
...
+    # since otherwise, we would have to load the parameters of the same module multiple
+    # times using different names for the weights and that's not viable with the internal
+    # logic of set_model_state_dict.
+    options = StateDictOptions(full_state_dict=True, broadcast_from_rank0=True, strict=False)
+
+    # Sync parameters and buffers
+    set_model_state_dict(model, full_state_dict, options=options)
+
+    # Sync additional buffers that may not be in the state_dict
+    for _, buffer in model.named_buffers():
```
bowenyang008, line 176: can we assert buffer is not a Dtensor, otherwise it is not correct

tests/trainer/test_fsdp2.py
```diff
@@ -321,3 +323,124 @@ def test_fsdp2_auto_microbatching_raises_error(
...
+@pytest.mark.gpu
+def test_fsdp2_syncs_module_states_when_multiple_devices_are_used(
+    world_size: int,
+):
+    """Test that FSDP2 syncs module states when multiple devices are used."""
+    del world_size
+    device = 'cuda' if dist.get_local_rank() == 0 else 'meta'
+    model = SimpleComposerMLP(num_features=10, device=device, num_classes=10)
+
+    # create a dummy param_init_fn that initializes the weights and biases to 1 and 2 respectively
```
bowenyang008, line 364: didn't see bias init

tests/trainer/test_fsdp2.py
```diff
@@ -321,3 +323,124 @@ def test_fsdp2_auto_microbatching_raises_error(
...
+
+
+@world_size(2)
+@pytest.mark.gpu
+def test_fsdp2_syncs_module_states_when_multiple_devices_are_used(
+    world_size: int,
+):
+    """Test that FSDP2 syncs module states when multiple devices are used."""
+    del world_size
+    device = 'cuda' if dist.get_local_rank() == 0 else 'meta'
```
bowenyang008, line 361: do you have a test with rank0 on cpu as well?

tests/trainer/test_fsdp2.py
```diff
@@ -321,3 +323,124 @@ def test_fsdp2_auto_microbatching_raises_error(
...
+        num_classes=10,
+        use_fsdp2=True,
+        optimizer=optimizer,
+        sync_module_states=True,
+    )
+    trainer.fit()
+
+    assert torch.equal(
+        model.mlp.fc1.weight.full_tensor(),  # type: ignore
+        optimizer.param_groups[0]['params'][0].full_tensor(),
```
bowenyang008, line 441: I think you can just test the local tensor instead of the full_tensor

tests/trainer/test_fsdp2.py
bowenyang008: and we should have a test to test 'mixed' device type as well right? or is this PR to support sync_module_state first and have a follow up PR to connect 'mixed' to use sync_module_state?

bowenyang008: and when we are ready to test real 'mixed' init I would like to see tests with 'mixed' init using both FSDP1 and FSDP2 converge.

composer/distributed/prepare_distributed.py
rithwik-db: no reason, will update

composer/utils/parallelism.py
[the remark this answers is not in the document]
bowenyang008: then should we remove this from the field and expose it only as a settable attr?

composer/distributed/shared_utils.py
```diff
@@ -155,3 +158,46 @@ def add_fsdp_oom_hooks(model: torch.nn.Module, device: Optional[Device] = None)
...
+    a mix where some ranks have parameters on meta device and others have parameters on
+    cpu/gpu, it raises a ValueError since this can lead to random weight initialization
+    when loading checkpoints.
+
+    Args:
+        model (nn.Module): The model to validate.
+        fsdp_config (FSDP2Config | FSDPConfig): The FSDP2 configuration containing sync_module_states setting.
+
+    Raises:
+        ValueError: If sync_module_states is False and there's mixed initialization across ranks.
```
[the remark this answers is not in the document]
bowenyang008: I think it is fine to use warning if we don't expose it as a field, see my comment in that thread. Though I would prefer another check that if sync_module_states is True, assert only rank0 is non-meta and all the other ranks are meta, otherwise our logic is incorrect.

composer/utils/parallelism.py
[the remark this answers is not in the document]
rithwik-db: We should only include this as a gettable property? Not sure why we want this to be settable if we're removing this entirely...

composer/distributed/shared_utils.py
```diff
@@ -155,3 +157,40 @@ def add_fsdp_oom_hooks(model: torch.nn.Module, device: Optional[Device] = None)
...
+    a mix where some ranks have parameters on meta device and others have parameters on
+    cpu/gpu, it raises a ValueError since this can lead to random weight initialization
+    when loading checkpoints.
+
+    Args:
+        model (nn.Module): The model to validate.
+        fsdp_config (FSDP2Config | FSDPConfig): The FSDP2 configuration containing sync_module_states setting.
+
+    Raises:
+        ValueError: If invalid sync_module_states configuration is detected.
```
bowenyang008, line 176: update it?

composer/distributed/shared_utils.py
```diff
@@ -155,3 +157,40 @@ def add_fsdp_oom_hooks(model: torch.nn.Module, device: Optional[Device] = None)
                 hook_handles.append(module.register_full_backward_hook(hook))  # type: ignore
 
     return hook_handles
+
+
+def validate_model_requires_state_sync(model: nn.Module, fsdp_config: FSDP2Config | FSDPConfig) -> None:
```
bowenyang008: should you change the name back to "udpate_xxx"?

composer/distributed/shared_utils.py
```diff
@@ -155,3 +157,40 @@ def add_fsdp_oom_hooks(model: torch.nn.Module, device: Optional[Device] = None)
...
+        model (nn.Module): The model to validate.
+        fsdp_config (FSDP2Config | FSDPConfig): The FSDP2 configuration containing sync_module_states setting.
+
+    Raises:
+        ValueError: If invalid sync_module_states configuration is detected.
+    """
+    device = get_device()
+    requires_sync = False
+
+    rank_on_meta = 1 if next(model.parameters()).device.type == 'meta' else 0
```
bowenyang008, line 177: should we use any instead of next?

composer/distributed/shared_utils.py
```diff
@@ -155,3 +157,40 @@ def add_fsdp_oom_hooks(model: torch.nn.Module, device: Optional[Device] = None)
...
+
+    if not fsdp_config.sync_module_states and requires_sync:
+        fsdp_config.sync_module_states = True
+
+    # Asserts that the rank setup is valid
+    if fsdp_config.sync_module_states:
+        if dist.get_global_rank() == 0:
+            assert rank_on_meta == 0
+        else:
+            assert rank_on_meta == 1
```
bowenyang008: I think we need a similar all_reduce to notify other ranks that a rank is invalid, otherwise we may hit NCCL hang/timeout

tests/trainer/test_fsdp2.py
```diff
@@ -321,3 +325,231 @@ def test_fsdp2_auto_microbatching_raises_error(
...
+            fsdp1_weights.values(),
+            fsdp2_weights.values(),
+        ):
+
+            assert fsdp1_weight.shape == fsdp2_weight.shape, \
+                f'Shape mismatch for {name}: FSDP1 {fsdp1_weight.shape} vs FSDP2 {fsdp2_weight.shape}'
+
+            diff = torch.abs(fsdp1_weight - fsdp2_weight).max().item()
+            assert diff < tolerance, \
+                f'Weight difference for {name} exceeds tolerance: {diff} > {tolerance}.'
```
bowenyang008, line 410: better use torch.allclose?

bowenyang008 (approved):
Nice work, thanks @rithwik-db!

composer/distributed/shared_utils.py
rithwik-db: Added a new test for this too

PR #3876, opened 2025-06-11
Monolithic checkpointing

rithwik-db:
Added monolithic checkpointing for FSDP2

Tested test runs in this [comment](https://github.com/mosaicml/composer/pull/3876#issuecomment-3002699822) + added unit tests that check tied weights (although in general, tied modules is invalid and we raise an error for that)

### DIFF
--- a/composer/core/state.py
+++ b/composer/core/state.py
@@ -453,6 +453,8 @@ class State(Serializable):
         precision_config: Optional[dict[str, Any]] = None,
 
         # optimizers
+        # TODO: Deprecate optimizers and support `optimizer` instead since we
+        # don't support multiple optimizers
         optimizers: Optional[Union[Optimizer, Sequence[Optimizer]]] = None,
 
         # scaler
@@ -591,7 +593,8 @@ class State(Serializable):
                 raise ValueError('load_monolith_rank0_only is not compatible with tensor parallelism (TP).')
             assert self.fsdp_config is not None
             error_message = ''
-            if self.fsdp_config.sync_module_states == False:
+            # FSDP2 automatically syncs module states, so we don't need to check for it
+            if isinstance(self.fsdp_config, FSDPConfig) and self.fsdp_config.sync_module_states == False:
                 error_message += textwrap.dedent(
                     "load_monolith_rank0_only requires parallelism_config['fsdp']['sync_module_states'] to be True. "
                     "Either set parallelism_config['fsdp']['sync_module_states'] = True or set load_monolith_rank0_only = False.",
@@ -911,10 +914,14 @@ class State(Serializable):
 
     @property
     def load_monolith_rank0_only(self):
-        return (
-            self.fsdp_config is not None and self.fsdp_config.auto_wrap and
-            self.fsdp_config.state_dict_type == 'full' and self.fsdp_config.load_monolith_rank0_only == True
+        should_load_monolith_rank0_only = (
+            self.fsdp_config is not None and self.fsdp_config.state_dict_type == 'full' and
+            self.fsdp_config.load_monolith_rank0_only == True
         )
+        # TODO: Only FSDP1 has auto_wrap; if this is a legacy config, we should remove this check
+        if isinstance(self.fsdp_config, FSDPConfig):
+            should_load_monolith_rank0_only = should_load_monolith_rank0_only and self.fsdp_config.auto_wrap
+        return should_load_monolith_rank0_only
 
     def _get_integrations_state_dict(self) -> dict[str, Any]:
         """Gets a dictionary of information about integrations to store in the state dict.
@@ -1325,14 +1332,14 @@ class State(Serializable):
         if self.load_monolith_rank0_only:
             assert self.fsdp_config is not None
             log.info('Wrapping model with FSDP after loading model_state.')
-            with reproducibility.seed_context(self.rank_zero_seed):
-                from composer.distributed import prepare_fsdp_module
+            self._apply_fsdp()
+            log.debug('Finished wrapping model with FSDP.')
 
-                # TODO (FSDP2): support calling FSDP2 wrapper depending on the config type
-                assert isinstance(
-                    self.fsdp_config,
-                    FSDPConfig,
-                ), f'prepare_fsdp_module requires FSDPConfig, got: {type(self.fsdp_config)}'
+    def _apply_fsdp(self):
+        # Init with globally fixed seed so all FSDP/HSDP replicas have the same initial weights
+        with reproducibility.seed_context(self.rank_zero_seed):
+            if isinstance(self.fsdp_config, FSDPConfig):
+                from composer.distributed import prepare_fsdp_module
                 self.automicrobatch_fsdp_hook_handles, self.fsdp_modules = prepare_fsdp_module(
                     self.model,
                     self.optimizers,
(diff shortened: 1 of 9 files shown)


composer/core/state.py
```diff
@@ -1370,15 +1387,24 @@ def load_optim_state(self, state_dict: dict[str, Any], strict: bool = True):
...
+                broadcast_from_rank0=broadcast_from_rank0,
+                cpu_offload=cpu_offload,
+                strict=strict,
+            )
+
             set_optimizer_state_dict(
                 model=self.model,
                 optimizers=optimizer,
                 optim_state_dict=optim_state_dict,  # type: ignore
-                options=StateDictOptions(
```
rithwik-db, line 1377: this part isn't really clear to me on how it actually worked lol, need to dig into how `broadcast_from_rank0` was bypassed in the past

composer/core/state.py
```diff
@@ -1326,21 +1331,33 @@ def load_model_state(
...
+                        self.model,
+                        self.optimizers,
+                        self.fsdp_config,
+                        self.precision,
+                        self.device,
+                        self.auto_microbatching,
+                    )
+                elif isinstance(self.fsdp_config, FSDP2Config):
+                    from composer.distributed.prepare_distributed import parallelize_composer_model
+                    from composer.models import ComposerModel
```
bowenyang008, line 1346: we should be able to move this to top of file, correct?

composer/core/state.py
```diff
@@ -1326,21 +1331,33 @@ def load_model_state(
...
+
+                    # FSDP2 doesn't support auto_microbatching (checked earlier, just validating here to be safe)
+                    assert not self.auto_microbatching, 'auto_microbatching is not supported with FSDP2'
+
+                    # FSDP2 requires a ComposerModel
+                    assert isinstance(self.model, ComposerModel), 'FSDP2 requires a ComposerModel'
```
bowenyang008, line 1352: type check should be good for this case

composer/core/state.py
```diff
@@ -1326,21 +1331,33 @@ def load_model_state(
...
+
+                    parallelize_composer_model(
+                        self.model,
+                        self.optimizers[0] if self.optimizers else None,
```
bowenyang008, line 1356: this just reminds me in the funtion _wrap_model_for_distributed we probably should change the arg name from "optimizer(s)" to "optimizer" given its type

composer/core/state.py
bowenyang008: interesting 🤦

composer/trainer/trainer.py
```diff
@@ -1731,25 +1731,32 @@ def __init__(
...
-            assert isinstance(
-                self.state.fsdp_config,
-                FSDPConfig,
-            ), f'prepare_fsdp_module requires FSDPConfig, got: {type(self.state.fsdp_config)}'
+        should_load_monolith = (
+            not self.state.fsdp_enabled and self.state.fsdp_config is not None and self.state.load_monolith_rank0_only
+        )
+        if isinstance(self.state.fsdp_config, FSDPConfig):
+            should_load_monolith = should_load_monolith and self.state.fsdp_config.auto_wrap
+        if should_load_monolith:
```
bowenyang008, line 1739: is this just the same logic as state.load_monolith_rank0_only?

bowenyang008 (commented):
did you have a chance to run an e2e test with finetuning?

composer/trainer/trainer.py
rithwik-db: This is running into some weird side effects if we consolidate, will look into this further.

composer/core/state.py
rithwik-db: This might be better to do in a separate PR

rithwik-db:
@bowenyang008  image <- `mpt-125m-fsdp2-monolithic-resumption-J6JwLr` looks like resumption is working correctly and the llama finetune regression went as expected as well (`llama2-finetune-regression-LbPI17`)

composer/core/state.py
```diff
@@ -1341,7 +1348,22 @@ def load_model_state(
                     self.device,
                     self.auto_microbatching,
                 )
-            log.debug('Finished wrapping model with FSDP.')
+            elif isinstance(self.fsdp_config, FSDP2Config):
+                from composer import ComposerModel
+                from composer.distributed.prepare_distributed import parallelize_composer_model
```
bowenyang008, line 1353: forgot if I asked before but can we move these imports to top now or is there still import dependency issue

rithwik-db: circular dependency issue :(

composer/core/state.py
```diff
@@ -1370,15 +1392,27 @@ def load_optim_state(self, state_dict: dict[str, Any], strict: bool = True):
 
             optim_state_dict = serialized_value[type(optimizer).__qualname__] if serialized_value is not None else None
 
+            # Note: 'broadcast_from_rank0' is only supported for FSDP2.
+            # - In `set_optimizer_state_dict`, FSDP1 follows a different code path where it detects FSDP modules and handles FlatParameters differently.
+            #   Furthermore, it requires `cpu_offload` to be set to True when loading the optimizer state in a monolithic checkpoint.
+            # - In FSDP2, setting `broadcast_from_rank0` will cause `set_optimizer_state_dict` to set the optim_state_dict (which is on CPU)
+            #   to be broadcasted from rank0's `full_state_dict` to all ranks' `local_state_dict` (moving them to DTensors on GPU)
```
bowenyang008, line 1399: what if broadcast_from_rank0 is False for FSDP2?

rithwik-db: well we wouldn't really have monolithic checkpointing right? It would just hang since it would just run `distribute_tensor` on all tensors in `full_state_dict` and since only rank 0 has elements in `full_state_dict`, it would call `distribute_tensor` and wait for all other ranks which would've skipped this stop and consequently, never call `distribute_tensor`.

composer/utils/parallelism.py
```diff
@@ -162,18 +161,19 @@ def data_parallel_shard_degree(self) -> int:
...
-        return 'sharded'
-
     @property
     def use_orig_params(self) -> bool:
         return True
 
     def __post_init__(self):
         warnings.warn('FSDP2 Config/APIs are experimental and subject to heavy changes', UserWarning)
 
+        if self.load_monolith_rank0_only and self.state_dict_type != 'full':
```
bowenyang008, line 171: can we just have state_dict_type property determined by load_monolith_rank0_only for backward compatibility instead of adding a new field?

bowenyang008 (approved):
Just left a few questions and LGTM!

composer/core/state.py
```diff
@@ -1370,15 +1392,27 @@ def load_optim_state(self, state_dict: dict[str, Any], strict: bool = True):
 
             optim_state_dict = serialized_value[type(optimizer).__qualname__] if serialized_value is not None else None
 
+            # Note: 'broadcast_from_rank0' is only supported for FSDP2.
+            # - In `set_optimizer_state_dict`, FSDP1 follows a different code path where it detects FSDP modules and handles FlatParameters differently.
+            #   Furthermore, it requires `cpu_offload` to be set to True when loading the optimizer state in a monolithic checkpoint.
```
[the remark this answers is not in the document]
rithwik-db: setting `broadcast_from_rank0=True` should lead to the same results as FSDP2 as it would for FSDP1 (regardless of what to what the state of `cpu_offload` is), but if `broadcast_from_rank0` is False, then `cpu_offload` has to be `True` otherwise, ranks with an empty state dict will fail (ranks >0) in the following check:
```
            if (
                not optim_state_dict
                and not (info.cpu_offload and info.full_state_dict)
                and (not info.broadcast_from_rank0)
            ):
````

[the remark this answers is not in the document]
Comment (rithwik-db) [reply composer/core/state.py]: Added more clarifications to the comment

File: composer/utils/parallelism.py (hunk 8 above)
Comment (rithwik-db) [reply composer/utils/parallelism.py:174]: Added a comment with a TODO on this to resolve later

File: tests/trainer/test_fsdp2.py [hunk 10]
```diff
@@ -605,3 +622,83 @@ def test_fsdp2_sync_module_states_mixed_init_weight_equivalence(
...
+    BATCH_SIZE = 2
+    DATASET_SIZE = 16  # 8 batches, 2 batches per rank, 2 ranks = 32 samples
+
+    # Use tmp_path from all ranks to ensure consistency
+    tmp_paths = dist.all_gather_object(os.path.abspath(tmp_path))
+    save_folder = tmp_paths[0]
+
+    save_interval = '1ba'
+    save_filename = 'ba{batch}-rank{rank}.pt'
+    resume_file = 'ba1-rank{rank}.pt'
```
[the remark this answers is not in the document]
rithwik-db: trains from ba1 -> ba8 (not ba1 -> ba9) so the output should be the same

PR #3884, opened 2025-06-27
Mixed Precision for FSDP2

rithwik-db:
# What does this PR do?

Added mixed precision for FSDP2. It is a bit hard to test the `reduce_dtype` so I added a workaround for that test, but the regular `param_dtype` validation is as expected.

### DIFF
--- a/composer/distributed/prepare_distributed.py
+++ b/composer/distributed/prepare_distributed.py
@@ -24,6 +24,9 @@ from composer.models import ComposerModel
 from composer.utils import dist
 from composer.utils.parallelism import FSDP2Config
 
+from composer.core.precision import Precision
+
+
 log = logging.getLogger(__name__)
 
 
@@ -68,6 +71,7 @@ def _check_duplicate_modules(model: torch.nn.Module) -> None:
 def _parallelize_model_helper(
     model: torch.nn.Module,
     config: FSDP2Config,
+    precision: Precision,
     fsdp_wrap_policy: Optional[CustomPolicy] = None,
     param_init_fn: Callable[[torch.nn.Module], None] = lambda m: None,
 ) -> None:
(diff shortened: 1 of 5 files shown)


rithwik-db:
@bowenyang008 this is a [quick test](https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/2514300354561999/runs?o=3360802220363900&searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D&compareRunsMode=CHART) that I ran for the various mixed precision options, the peak reserved memory looks reasonable and the loss curves are identical. There's a deviation in MFU though but I am not sure if that's super relevant?

composer/distributed/prepare_distributed.py
```diff
@@ -140,6 +147,12 @@ def parallelize_model(
         ValueError: If the config is not an FSDP2Config or if activation_checkpointing_check_fn is provided
             but neither activation_checkpointing nor activation_cpu_offload is enabled in the config.
     """
+    if precision is None:
+        precision = Precision.AMP_FP16 if isinstance(get_device(), DeviceGPU) else Precision.FP32
+    elif isinstance(precision, str):
+        precision = Precision(precision)
+    _validate_precision(precision, get_device())
```
bowenyang008, line 154: seems we call this twice, once here and once in fsdp2.py?

rithwik-db: just for testing really (to make sure that `test_fsdp2_wrap_fsdp_attribute.py` doesn't run into an invalid num argument error). I can work around that though if this is confusing

tests/trainer/test_fsdp2.py
```diff
@@ -702,3 +707,145 @@ def test_fsdp2_monolithic_checkpoint_save_and_load(
...
+            actual_grad_dtype = module._get_fsdp_state()._fsdp_param_group._reduce_dtype  # type: ignore
+            new_validation = {
+                'module': module.__class__.__name__,
+                'expected': expected_reduce_dtype,
+                'actual': actual_grad_dtype,
+                'matches': actual_grad_dtype == expected_reduce_dtype,
+            }
+            grad_dtype_validations.append(new_validation)
+        except Exception:
+            pass
```
bowenyang008, line 813: do we still need to check this as a hook? my understand is module._get_fsdp_state()._fsdp_param_group._reduce_dtype is static, so independent of whether you have run a fwd or not

bowenyang008 (approved):
Thanks, @rithwik-db !

young-agi:
Hey folks! Are you gonna merge this soon?

tests/trainer/test_fsdp2.py
rithwik-db: Yeah we don't really need this as a hook, I'll move it outside

composer/distributed/prepare_distributed.py
rithwik-db: I guess I will instead just add it as part of the `test_fsdp2_wrap_fsdp_attribute.py` tests since that will make it less confusing

rithwik-db: Oh wait sorry, was reading the wrong thing, yeah we can remove the `_validate_precision()` call in `fsdp.py`

PR #3889, opened 2025-07-01
Add h200 to flops dict

dakinggg:
# What does this PR do?
As title. Assumes SXM because I didn't see an easy way to differentiate between SXM and PCIe variants - it is no longer specified in the device name. afaict we would have to check bandwidth, power limits, bus ids, or something like that. I also don't have an H200 PCIe to check against.

Tested manually what device name is returned.

### DIFF
--- a/composer/callbacks/speed_monitor.py
+++ b/composer/callbacks/speed_monitor.py
@@ -21,6 +21,19 @@ if is_xla_installed():
 __all__ = ['SpeedMonitor']
 
 GPU_AVAILABLE_FLOPS = {
+    # source: https://resources.nvidia.com/en-us-data-center-overview-mc/en-us-data-center-overview/hpc-datasheet-sc23-h200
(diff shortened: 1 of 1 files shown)


composer/callbacks/speed_monitor.py
```diff
@@ -21,6 +21,19 @@
...
+        'fp32': 67e12,
+        'tf32': 989e12 / 2,
+        'fp16': 1.979e15 / 2,
+        'amp_fp16': 1.979e15 / 2,
+        'bf16': 1.979e15 / 2,
+        'amp_bf16': 1.979e15 / 2,
+        'fp8': 3.958e15 / 2,
+        'amp_fp8': 3.958e15 / 2,
+        'int8': 3.958e15 / 2,
+    },
```
bowenyang008, line 36: nit: can we share the same value with h100 since they are the same flops-wise?

PR #3897, opened 2025-07-11
Supporting peft for FSDP2

rithwik-db:
# What does this PR do?

FSDP2 mixed init requires the full state dict and peft models, when `should_save_peft_only: true` has an updated `state_dict` function that only returns the lora modules instead of the full modules. Therefore, when we encounter this, we use a context manager to switch the variable to get the full state dict.

Ran a full test on `dpo-rl-fsdp2-ESou5r`.

### DIFF
--- a/composer/distributed/prepare_distributed.py
+++ b/composer/distributed/prepare_distributed.py
@@ -42,6 +42,25 @@ def log_execution_time(logger: logging.Logger, operation_name: str):
         logger.info(f'{operation_name} took {end_time - start_time:.2f} seconds')
 
 
+@contextmanager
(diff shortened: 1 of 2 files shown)


bowenyang008 (approved):
LGTM, in your experiments did you notice if the lora weights are also sharded, if so we might want to log a TODO to provide interfaces to avoid sharding lora weights as they may be inefficien to shard

rithwik-db:
> LGTM, in your experiments did you notice if the lora weights are also sharded, if so we might want to log a TODO to provide interfaces to avoid sharding lora weights as they may be inefficien to shard

Yeah definitely makes sense, added a TODO above

PR #3907, opened 2025-07-22
Adding `support_full_params` functionality for FSDP2

rithwik-db:
# What does this PR do?

Added functionality and some tests.


composer/distributed/fsdp2_utils.py
```diff
@@ -378,3 +380,184 @@ def lambda_fn(current_module: nn.Module) -> bool | dict[str, Any]:
...
+
+    We currently don't support rank0_only, offload_to_cpu, and with_grads.
+    """
+    # TODO: We want to support these arguments in the future.
+    if any([rank0_only, offload_to_cpu, with_grads]):
+        warnings.warn(
+            'rank0_only, offload_to_cpu, and with_grads are not supported for FSDP2 models. '
+            'These arguments will be ignored. The defaults supported are: '
+            'rank0_only=False, offload_to_cpu=False, with_grads=False.',
+        )
```
bowenyang008, line 457: I think we should error out the provided is not the same as default

composer/distributed/fsdp2_utils.py
```diff
@@ -378,3 +380,184 @@ def lambda_fn(current_module: nn.Module) -> bool | dict[str, Any]:
...
+    for name, dtensor_param in dtensor_params.items():
+        tensor_hash = hash(dtensor_param)
+        if tensor_hash not in tensor_to_names:
+            tensor_to_names[tensor_hash] = []
+        tensor_to_names[tensor_hash].append(name)
+
+    # Process parameters, handling tied parameters correctly
+    processed_tensors = set()
+    for name, dtensor_param in dtensor_params.items():
+        tensor_hash = hash(dtensor_param)
```
bowenyang008, line 490: I am not sure why do we still need to handle tied params here per our earlier discussion. based on our FSDP2 construction logic, there can't be any tied params in any two FSDP2 (or even non-FSDP2) modules, so we should be safe to skip this logic

composer/distributed/fsdp2_utils.py
```diff
@@ -378,3 +380,184 @@ def lambda_fn(current_module: nn.Module) -> bool | dict[str, Any]:
...
+        for part in parts[:-1]:
+            parent_module = getattr(parent_module, part)
+        return parent_module, parts[-1]
+
+    # Group parameters by their underlying tensor to handle tied parameters
+    tensor_to_names = {}
+    for name, dtensor_param in dtensor_params.items():
+        tensor_hash = hash(dtensor_param)
+        if tensor_hash not in tensor_to_names:
+            tensor_to_names[tensor_hash] = []
```
bowenyang008, line 484: I think you can just use the dtensor_param as the direct key, instead of this double hashing otherwise you may theoretically run into hash collision

bowenyang008 (approved):
LGTM, thanks @rithwik-db

composer/distributed/fsdp2_utils.py
rithwik-db: Yeah that makes sense

gupta-abhay (approved):
lgtm

composer/distributed/prepare_distributed.py
```diff
@@ -121,6 +121,12 @@ def _parallelize_model_helper(
             )
             with get_full_state_dict(model):
                 full_state_dict = get_model_state_dict(model, options=options)
+            # Also move the model to CPU to avoid GPU 0 memory spike
+            # fully_shard(module) will handle nodes to the correct device
+            # based on the available accelerators in the process group.
+            # If we don't do this, we can run out of GPU memory for large models.
+            model = model.cpu()
+            torch.cuda.empty_cache()
```
gupta-abhay, line 129: is this happening explicitly with fsdp2?

rithwik-db: Yep, not something I noticed originally, maybe because the model sizes were so small (and I can't see my 7B param run anymore since for some reason I am not assigned to the `mosaic-flow` workspace where that is saved), but here's what the logs look like:
image

Since I was using H200s before, also didn't notice this.

Seems we are allocating beyond what is needed, but something that doesn't make sense to me is why we're allocating 142GB, what is the recommended way to profile this with our stack?

gupta-abhay: yeah we have the Profiler callback - https://github.com/mosaicml/composer/blob/main/composer/profiler/profiler.py which I've used previously.

PR #3908, opened 2025-07-25
Reverting automicrobatching changes

rithwik-db:
# What does this PR do?

Haven't investigated fully yet (my mcli is broken and need to fix it tomorrow), but it seems that we are getting overall lower (token?) throughput in some smaller-model runs. Reverting this change seems to have fixed the issue but it's unclear what the actual root cause is. We should probably only merge this change in when we understand what changed.

### DIFF
--- a/composer/distributed/dist_strategy.py
+++ b/composer/distributed/dist_strategy.py
@@ -31,7 +31,7 @@ from composer.distributed.mosaic_parallelism import (
     get_mixed_precision,
     set_custom_fsdp_module_kwargs,
 )
-from composer.distributed.shared_utils import add_fsdp_oom_hooks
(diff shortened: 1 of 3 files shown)


composer/distributed/dist_strategy.py
```diff
@@ -31,7 +31,6 @@
     get_mixed_precision,
     set_custom_fsdp_module_kwargs,
 )
-from composer.distributed.shared_utils import add_fsdp_oom_hooks
```
rithwik-db, line 34: WIP since we need to remove the functions in `shared_utils.py` that we're not using (need to get some `mcli` stuff fixed before that)

rithwik-db:
this was fixed in another PR

PR #3915, opened 2025-07-28
Log gradient clipping norm and frequency

josejg:
## Context

Currently `torch.nn.utils.clip_grad_norm_` and `FSDP.clip_grad_norm_` apply the gradient normalization in place but also return the pre-clip gradient norm value, however the value is not capture nor logged anywhere. 

We can't change the API for all gradient clipping methods since some don't have top level scalar, but we can for gradient norm clipping, the most frequent one we use. 

This PR propagates the value outside of the helper and into the algorithm where it can be logged. Since clipping is fairly bursty, we also compute the rolling window over `clipping_frequency_window` samples to provide a more parseable metric.

## Known caveats:
- `_clipping_history` is not persisted so the metric will change slightly upon resumption
- If gradient clipping is not enabled, the gradient norm won't be logged. For logging without clipping the best solution is to set clipping to really high threshold, e.g. `clipping_threshold: 100`

## Experiments

Couple of example experiments that showcase the functionality with SFT and GRPO are `2025-07-21-debug-gradient-clipping` and `2025-07-25-math-rlvr-grpo` respectively

### DIFF
--- a/composer/algorithms/gradient_clipping/gradient_clipping.py
+++ b/composer/algorithms/gradient_clipping/gradient_clipping.py
@@ -25,7 +25,7 @@ def apply_gradient_clipping(
     clipping_type: str,
     clipping_threshold: float,
     fsdp_enabled: bool,
-):
+) -> Union[torch.Tensor, None]:
     """Clips all gradients in model based on specified clipping_type.
 
     Args:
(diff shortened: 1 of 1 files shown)


composer/algorithms/gradient_clipping/gradient_clipping.py
```diff
@@ -41,12 +41,16 @@ def apply_gradient_clipping(
             threshold by which if grad_norm / weight_norm is greater than this threshold then
             scale gradients by this threshold * (weight_norm / grad_norm) (for 'adaptive').
         fsdp_enabled (bool): Bool of if the model is a FSDP model or not.
+
+    Returns:
+        Union[torch.Tensor, None]: The total gradient norm before clipping for 'norm' clipping type,
```
irenedea, line 46: Let's just always return the result, not just for norm. 

It's a weird contract for a separate function downstream to know this behavior

irenedea (approved):

composer/algorithms/gradient_clipping/gradient_clipping.py
irenedea: approved, not ideal, but makes sense that this is the best we can do

PR #3931, opened 2026-02-11
fix: Progressive Resizing logs height twice instead of width

Mr-Neutr0n:
## Bug
Progressive Resizing logs the height value for both height and width, so the width is never correctly reported in logs.

In `resize_batch()`, the debug log for new input dimensions uses `X_sized.shape[2]` (height) twice instead of `X_sized.shape[2]` for height and `X_sized.shape[3]` for width.

## Fix
Changed the second `X_sized.shape[2]` to `X_sized.shape[3]` so the width dimension is correctly logged.

### DIFF
--- a/composer/algorithms/cutmix/cutmix.py
+++ b/composer/algorithms/cutmix/cutmix.py
@@ -333,8 +333,8 @@ def _gen_cutmix_coef(alpha: float) -> float:
 
 
 def _rand_bbox(
-    W: int,
     H: int,
+    W: int,
     cutmix_lambda: float,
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #3932, opened 2026-02-11
fix: correct CutMix bounding box dimension swap

Mr-Neutr0n:
## Bug
In CutMix, the bounding box computation swaps height and width dimensions — height is computed from the width extent and vice versa, leading to incorrectly shaped cutout regions.

The `_rand_bbox` function declares parameters as `(W, H)` but callers pass `(input.shape[2], input.shape[3])` which is `(H, W)`. This causes `cut_w` to be computed from the height dimension and `cut_h` from the width dimension, and the clipping bounds to use the wrong spatial extent.

## Fix
Corrected the parameter order in `_rand_bbox` from `(W, H)` to `(H, W)` so that it matches the call sites. Also fixed the corresponding keyword arguments in the test.

### DIFF
--- a/composer/algorithms/cutmix/cutmix.py
+++ b/composer/algorithms/cutmix/cutmix.py
@@ -333,8 +333,8 @@ def _gen_cutmix_coef(alpha: float) -> float:
 
 
 def _rand_bbox(
-    W: int,
     H: int,
+    W: int,
     cutmix_lambda: float,
     cx: Optional[int] = None,
     cy: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

